PR #196, opened 2020-01-28
Added Fetch and Load Methods

### LINKED ISSUE #195: [TASK]🗃️ Fetch and Load Tipsets for Syncer
**Task summary**

Be able to load tipset from the chain store and call blocksync to fetch a tipset.

**Specification reference**

- [LoadAndFetchTipSets](https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L349)

dutterbutter:
**Summary of changes**
* Added `fetch_tipsets` method which returns `FullTipset` either from store or requesting via blocksync (todo)
* Added `load_fts` which retrieves and constructs full tipset if keys exist in key-value store
* Added method for returning Unsigned and Signed Messages from key-value store
* Added `TipSetKey` method for returning a reference to cid keys
* Added comments

Closes  #195 

The full completion of #195 is dependent on the AMT/HAMT and blocksync. See comments in `syncer.rs` and `store.rs` for more details. Once this PR is reviewed, I can make a separate task based on the above mentioned TODOS.

- [Lotus Reference Syncer.rs](https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L357)
- [Lotus Reference Store.rs](https://github.com/filecoin-project/lotus/blob/master/chain/store/store.go#L764)


blockchain/blocks/src/header.rs
```diff
@@ -396,7 +396,7 @@ mod tests {
         assert_eq!(&header_bytes, &HEADER_BYTES)
     }
 
-    fn build_header() -> BlockHeader {
+    pub fn build_header() -> BlockHeader {
```
austinabell: doesn't matter because it's in the unexported module, but why this change?

ipld/amt/src/lib.rs
```diff
@@ -8,7 +8,7 @@
 
 mod amt;
 mod bitmap;
-mod block_store;
+pub mod block_store;
```
austinabell: BlockStore is already being exported? Why export the module here?

blockchain/sync_manager/src/sync.rs
```diff
@@ -78,32 +74,25 @@ impl<'a> Syncer<'a> {
...
-
-        let mut secp_cids = Vec::new();
-        for secp_msg in secp_msgs {
-            secp_cids.push(secp_msg.cid());
-        }
-
-        let mut amt = AMT::new(&self.chain_store.db);
+        // collect bls and secp cids
+        let bls_cids = bls_msgs.iter().map(|x| x.cid()).collect::<Vec<_>>();
+        let secp_cids = secp_msgs.iter().map(|x| x.cid()).collect::<Vec<_>>();
```
austinabell: You're actually creating an AMT of results, should just use the cids_from_messages helper function you have in this file. How did this function go unused when it was previously used?

Should look something like:
```
        let bls_cids = cids_from_messages(block.bls_msgs())?;
        let secp_cids = cids_from_messages(block.secp_msgs())?;
```

blockchain/sync_manager/src/sync.rs
```diff
@@ -78,32 +74,25 @@ impl<'a> Syncer<'a> {
...
+        // generate AMT and batch set message values
         let bls_root = AMT::new_from_slice(&self.chain_store.db, &bls_cids)?;
         let secp_root = AMT::new_from_slice(&self.chain_store.db, &secp_cids)?;
 
         let meta = MsgMeta {
             bls_message_root: bls_root,
             secp_message_root: secp_root,
         };
+        // store message roots and receive meta_root
+        let meta_root = self.chain_store.db.put(&meta)?;
```
austinabell: Just a comment to consider changing the db type to an implementation of the BlockStore trait, since that is the only way it is used, but I can explore this in a seperate PR

ipld/cid/src/error.rs
```diff
@@ -4,10 +4,10 @@
 use encoding::Error as EncodingError;
 use multibase;
 use multihash;
+use serde::{Deserialize, Serialize};
 use std::{error, fmt, io};
-
 /// Error types
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
```
austinabell: remove serialize and deserialize, you only have this because you are trying to serialize this in the amt

node/db/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 
 mod errors;
 mod memory;
-mod rocks;
+pub mod rocks;
```
austinabell: no need to change this

vm/message/src/lib.rs
```diff
@@ -31,3 +33,9 @@ pub trait Message {
     /// Returns the gas limit for the message
     fn gas_limit(&self) -> &BigUint;
 }
+
+#[derive(Serialize, Deserialize)]
```
austinabell: Note that if you derive this will be in map format, when it seems it's serialized as a tuple in Lotus like: `(bls_message_root, secp_message_root)` 

Probably worth implementing here, or in another PR where all of this is tested against Lotus

blockchain/sync_manager/src/sync.rs
austinabell: you removed the function with: https://github.com/ChainSafe/forest/pull/196/commits/9155c84212b43d3c43bf098dc50dd73649c130f7, but you probably want it as it will make your life easier

blockchain/blocks/src/header.rs
dutterbutter: I was using it while I was writing unit tests, will revert back until then.

ipld/amt/src/lib.rs
dutterbutter: As mentioned, there are compile errors when trying to use the BlockStore trait without the module being public.

node/db/src/lib.rs
dutterbutter: As mentioned, there are compile errors when trying to use the BlockStore trait for Rocksdb without the module being public.

blockchain/blocks/src/tipset.rs
```diff
@@ -38,6 +37,10 @@ impl TipSetKeys {
         }
         true
     }
+    /// Returns tipset keys
+    pub fn tipset_keys(&self) -> &Vec<Cid> {
```
austinabell:
```suggestion
    pub fn tipset_keys(&self) -> &[Cid] {
```
Is there a reason to have it this way?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -17,7 +18,7 @@ pub struct ChainStore<'a> {
     // TODO add a pubsub channel that publishes an event every time the head changes.
 
     // key-value datastore
-    db: Blockstore,
+    pub db: Blockstore,
```
austinabell: you have the function `blockstore()` already, so why do you need to make this public?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -106,4 +107,53 @@ impl<'a> ChainStore<'a> {
     pub fn blockstore(&self) -> &Blockstore {
         &self.db
     }
+    /// Returns Tipset from key-value store from provided keys
+    pub fn tipset(&self, keys: &[Cid]) -> Result<Tipset, Error> {
+        let mut block_headers = Vec::new();
+        for k in keys {
+            let raw_header = self.db.read(k.key())?;
+            if let Some(ref x) = raw_header {
```
austinabell:
```suggestion
            if let Some(x) = raw_header {
```
nit, rust will dereference it anyway but there is no need here

blockchain/chain/src/store/chain_store.rs
```diff
@@ -106,4 +107,53 @@ impl<'a> ChainStore<'a> {
     pub fn blockstore(&self) -> &Blockstore {
         &self.db
     }
+    /// Returns Tipset from key-value store from provided keys
+    pub fn tipset(&self, keys: &[Cid]) -> Result<Tipset, Error> {
```
austinabell: Optional, to maybe have `keys` be `cids` since it makes more sense than referring to it as the key value for the db. Also kinda confusing below to call k.key()

blockchain/chain/src/store/chain_store.rs
```diff
@@ -106,4 +107,53 @@ impl<'a> ChainStore<'a> {
...
+
+    /// Returns a Tuple of bls messages of type UnsignedMessage and secp messages
+    /// of type SignedMessage
+    pub fn messages(
+        &self,
+        _bh: &BlockHeader,
+    ) -> Result<(Vec<UnsignedMessage>, Vec<SignedMessage>), Error> {
+        // TODO read_msg_cids from message root; returns bls_cids and secp_cids
+        // let (blscids, secpkcids) = read_msg_cids(bh.messages())
+        // temporarily using vec!(bh.message_receipts() until read_msg_cids is completed with HAMT
```
austinabell: Optional to update or remove this comment to not confuse anything

ipld/amt/src/lib.rs
austinabell: Nah you don't need this just refer to it as `amt::BlockStore` and not `amt::blockstore::BlockStore`

node/db/src/lib.rs
austinabell: Not needed, just tested

vm/message/src/lib.rs
```diff
@@ -10,7 +10,13 @@ pub use signed_message::*;
 pub use unsigned_message::*;
 
 use address::Address;
+use cid::Cid;
+use encoding::{
+    de::{self, Deserializer},
+    ser::{self, Serializer},
+};
 use num_bigint::BigUint;
+use serde::Deserialize;
```
austinabell: nit: Use the one in `encoding::de::Deserialize` (Weird that the two are referenced in the same file, although they are basically reexports of the same thing)

blockchain/chain/src/store/chain_store.rs
dutterbutter: I need access to the underlying DB type, for 

` let bls_root = AMT::new_from_slice(&self.chain_store.db, &bls_cids)?;`

If I made use of `blockstore()`, such as 

`let bls_root = AMT::new_from_slice(&self.chain_store.blockstore().db()?, &bls_cids)?;`

`AMT::new_from_slice` complains about the reference and error type.

austinabell: yeah, because it was behind two references (I don't know where you got that function) use:
```
        let bls_root = AMT::new_from_slice(self.chain_store.blockstore(), &bls_cids)?;
        let secp_root = AMT::new_from_slice(self.chain_store.blockstore(), &secp_cids)?;
```

and you don't need to make the db variable public in two different ways

PR #197, opened 2020-01-29
AMT implementation

### LINKED ISSUE #188: Implement AMT structure
Array (ordered) version of #187

austinabell:
**Summary of changes**
- Implements AMT IPLD structure #188 
  - Allows it to be used in #196 and other PR/ changes using an AMT
  - Will try to make the code more readible in another PR, wanted to get this in ASAP to be able to be used (Code isn't janky it just isn't very readible)
  - Will open an issue to benchmark this
  - Cids of nodes will not match Lotus, as they are using Blake2b256 which isn't available in rust multihash (but I will fork and PR in other changes)

In summary, this is a sharded array, which underlying nodes get cbor serialized and stored to the underlying database. The BlockStore trait here is a wrapper around our DB traits to abstract the need to serialize and generate a cid specifically every root and node.

A node always has 8 items, but can either be a leaf node where all items are values or a link node where the values can be empty, a cid to a node (which can be pulled from the blockstore when needed), or a cached node (heap allocated). To not require every change in the AMT to be persisted to the DB, the AMT should only be flushed (remove cached nodes) when the Cid or the AMT needs to be persisted.

Closes #188 


ipld/amt/tests/amt_tests.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+#[test]
+fn delete() {
+    let db = db::MemoryDB::default();
+    let mut a = AMT::new(&db);
+    a.set(0, &"ferret").unwrap();
+    a.set(1, &"ferret").unwrap();
+    a.set(2, &"ferret").unwrap();
+    a.set(3, &"ferret").unwrap();
```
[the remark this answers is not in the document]
ec2: easter eggs haha

austinabell:
Ended up going with the generic and typesafe version because the other way was actually incompatible, so sorry for making changes again but the way it was is going to be janky to serialize and deserialize correctly. Also this fixes https://github.com/ChainSafe/forest/issues/203 because I'm using a fork of multihash where I added this functionality and also checked the cid for AMTs against ones generated through Lotus

blockchain/blocks/Cargo.toml
```diff
@@ -10,7 +10,7 @@ crypto = { path = "../../crypto" }
 message = { package = "forest_message", path = "../../vm/message" }
 clock = { path = "../../node/clock" }
 cid = { package = "forest_cid", path = "../../ipld/cid" }
-multihash = "0.9.3"
+multihash = { git = "https://github.com/austinabell/rust-multihash", rev = "56a8304b1b47697660dba7252d214be9829e137d" }
```
dutterbutter: Assuming we can update these back given latest merge

ipld/amt/src/bitmap.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl<'de> de::Deserialize<'de> for BitMap {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        let bz: Vec<u8> = serde_bytes::Deserialize::deserialize(deserializer)?;
+
+        // Get bitmap byte from serialized bytes
+        let bmap: BitMap = bz
```
ec2: Do you need to bind to bmap here? Can't just return?

austinabell: Rust will actually compile each identically, this is just for readability

ipld/amt/src/root.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::Node;
+use encoding::{
+    de::{self, Deserialize},
+    ser::{self, Serialize},
+};
+
+/// Root of an AMT vector, can be serialized and keeps track of height and count
+#[derive(PartialEq, Debug)]
+pub(super) struct Root<V> {
+    pub(super) height: u32,
```
ec2: Just a thought. But these fields can be `pub` instead of `pub(super)` since your struct is `pub(super)`, right? Doesn't matter though, just a though :~)

austinabell: This is more explicit for if the Root visibility changes in future. All of this is internal so no need for that <|:)

Issue #142, opened 2020-01-07
Migrate to Stable Futures + async-std


ec2:
Libp2p has merged in their efforts to move to stable futures. Gossipsub has a branch that is pretty much in review. So migrating now is a good time

amerameen:
This is complete, right? @ec2

ec2:
Yeah closing.

PR #209, opened 2020-02-03
Migration to Stable Futures and Network Refactor

ec2:
**Summary of changes**
- Remove the need for the `network` crate
- Migrate to Stable Futures and async-std instead of Tokio

Closes 
#142


node/forest_libp2p/src/service.rs
```diff
@@ -63,57 +83,76 @@ impl Libp2pService {
...
-}
-
-impl Stream for Libp2pService {
-    type Item = NetworkEvent;
-    type Error = ();
 
-    /// Continuously polls the Libp2p swarm to get events
-    fn poll(&mut self) -> Result<Async<Option<Self::Item>>, Self::Error> {
+    /// Starts the `Libp2pService` networking stack. This Future resolves when shutdown occurs.
+    pub async fn run(self) -> Result<(), ()> {
```
austinabell: Is there a reason the return type is as such? I don't see the function returning anything?

node/forest_libp2p/src/service.rs
```diff
@@ -63,57 +83,76 @@ impl Libp2pService {
...
+                        self.pubsub_sender_out.send(NetworkEvent::PubsubMessage {
                             source,
                             topics,
-                            message,
-                        })));
+                            message
+                        }).await;
                     }
+                    }
+                    Some(MergeEvent::Pubsub(_)) => error!(self.log, "Should not happen"),
```
austinabell: Can you log a more detailed message? Would it be worth passing through the merge event to log it?

node/forest_libp2p/src/service.rs
```diff
@@ -63,57 +83,76 @@ impl Libp2pService {
...
                 },
-                Ok(Async::Ready(None)) => break,
-                Ok(Async::NotReady) => break,
-                _ => break,
-            }
+                pubsub_event = pubsub_stream.next() => match pubsub_event {
+                    Some(MergeEvent::Pubsub(payload)) => {
+                       swarm_stream.get_mut().get_mut().publish(&payload.0, payload.1);
+                    }
+                    Some(MergeEvent::Swarm(_)) => error!(self.log, "Should not happen"),
```
austinabell: Same as above

node/forest_libp2p/src/service.rs
```diff
@@ -63,57 +83,76 @@ impl Libp2pService {
...
-        topics: Vec<TopicHash>,
-        message: Vec<u8>,
-    },
+    /// Returns a `Sender` allowing you to send messages over GossipSub
+    pub fn pubsub_sender(&self) -> Sender<(Topic, Vec<u8>)> {
+        self.pubsub_sender_in.clone()
+    }
+
+    /// Returns a `Receiver` to listen to GossipSub messages
+    pub fn pubsub_receiver(&self) -> Receiver<NetworkEvent> {
```
austinabell: is there a reason you don't pass a reference for these?

node/forest_libp2p/src/service.rs
```diff
@@ -63,57 +83,76 @@ impl Libp2pService {
...
+        self.pubsub_receiver_out.clone()
+    }
 }
 
+/// Builds the transport stack that LibP2P will communicate over
 pub fn build_transport(local_key: Keypair) -> Boxed<(PeerId, StreamMuxerBox), Error> {
     let transport = libp2p::tcp::TcpConfig::new().nodelay(true);
-    let transport = libp2p::dns::DnsConfig::new(transport);
+    let transport = libp2p::dns::DnsConfig::new(transport).unwrap();
```
austinabell: Can you avoid unwrapping by handling the error?

node/src/main.rs
```diff
@@ -5,30 +5,28 @@ mod cli;
...
-    let rt = Runtime::new().unwrap();
+    let logger = log.clone();
 
-    // Create the channel so we can receive messages from NetworkService
-    let (tx, _rx) = mpsc::unbounded_channel::<NetworkEvent>();
-    // Create the default libp2p config
-    // Start the NetworkService. Returns net_tx so  you can pass messages in.
-    let (_network_service, _net_tx, _exit_tx) =
-        NetworkService::new(&config.network, &log, tx, &rt.executor());
+    let lp2p_service = Libp2pService::new(logger, &config.network);
```
austinabell: just clone the logger in the function when you call `::new`

node/src/main.rs
```diff
@@ -5,30 +5,28 @@ mod cli;
...
+use async_std::task;
+use forest_libp2p::service::Libp2pService;
 use slog::info;
-use tokio::runtime::Runtime;
-use tokio::sync::mpsc;
+use std::error::Error;
 
-fn main() {
+#[async_std::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
austinabell: This doesn't need to have a result return? You aren't handling any errors in this function? Are you planning on doing this later?

node/src/main.rs
```diff
@@ -5,30 +5,28 @@ mod cli;
...
-    // Create the channel so we can receive messages from NetworkService
-    let (tx, _rx) = mpsc::unbounded_channel::<NetworkEvent>();
-    // Create the default libp2p config
-    // Start the NetworkService. Returns net_tx so  you can pass messages in.
-    let (_network_service, _net_tx, _exit_tx) =
-        NetworkService::new(&config.network, &log, tx, &rt.executor());
+    let lp2p_service = Libp2pService::new(logger, &config.network);
+
+    task::block_on(async move {
+        let _ = lp2p_service.run().await;
```
austinabell:
```suggestion
        lp2p_service.run().await;
```
Why do you need the `let _ =` ?

node/forest_libp2p/src/behaviour.rs
```diff
@@ -12,13 +12,13 @@ use libp2p::ping::{
...
 use libp2p::swarm::{NetworkBehaviourAction, NetworkBehaviourEventProcess};
-use libp2p::tokio_io::{AsyncRead, AsyncWrite};
 use libp2p::NetworkBehaviour;
 use slog::{debug, Logger};
+use std::{task::Context, task::Poll};
 
 #[derive(NetworkBehaviour)]
 #[behaviour(out_event = "MyBehaviourEvent", poll_method = "poll")]
-pub struct MyBehaviour<TSubstream: AsyncRead + AsyncWrite> {
+pub struct MyBehaviour<TSubstream: AsyncRead + AsyncWrite + Unpin + Send + 'static> {
```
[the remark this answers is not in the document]
ec2: Good idea

node/forest_libp2p/src/service.rs
```diff
@@ -63,57 +83,76 @@ impl Libp2pService {
             swarm.subscribe(topic);
         }
 
-        Libp2pService { swarm }
+        let (pubsub_sender_in, pubsub_receiver_in) = channel(20);
+        let (pubsub_sender_out, pubsub_receiver_out) = channel(20);
```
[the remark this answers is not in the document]
ec2: We had channels before. We're now using async-std channels which are bounded. Before we used UnboundedChannels from futures.rs. 

20 is chosen arbitrarily and can be adjusted when we do some benchmarks.

node/forest_libp2p/src/service.rs
```diff
@@ -63,57 +83,76 @@ impl Libp2pService {
...
-    type Error = ();
 
-    /// Continuously polls the Libp2p swarm to get events
-    fn poll(&mut self) -> Result<Async<Option<Self::Item>>, Self::Error> {
+    /// Starts the `Libp2pService` networking stack. This Future resolves when shutdown occurs.
+    pub async fn run(self) -> Result<(), ()> {
+        enum MergeEvent {
+            Swarm(MyBehaviourEvent),
+            Pubsub((Topic, Vec<u8>)),
+        };
```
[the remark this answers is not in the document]
ec2: Its solely used to map the the two streams to the same type. We need that because the `select!` macro requires that the futures it selects on resolve to the same type.

node/forest_libp2p/src/service.rs
ec2: You mean return a reference? These are multiple producer, multiple consumer channels. So we can clone these and pass them around

node/forest_libp2p/src/service.rs
ec2: I figured that if the transport cant be built, might as well crash the program on start? Thoughts?

node/forest_libp2p/src/behaviour.rs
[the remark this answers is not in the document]
ec2: fixed

node/forest_libp2p/src/behaviour.rs
```diff
@@ -29,6 +29,7 @@ pub struct MyBehaviour<TSubstream: AsyncRead + AsyncWrite> {
     log: Logger,
 }
 
+#[derive(Debug)]
 pub enum MyBehaviourEvent {
```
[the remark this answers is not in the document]
ec2: fixed

node/forest_libp2p/src/service.rs
ec2: I changed it to unreachable

node/forest_libp2p/src/service.rs
ec2: I changed it to unreachable

node/src/main.rs
ec2: I'll remove it for now

node/src/main.rs
ec2: I needed it because run returns a result, but i just removed it.

node/forest_libp2p/src/service.rs
ec2: I had it in there before for some reason. Ill remove it

blockchain/chain/Cargo.toml
```diff
@@ -5,9 +5,9 @@ authors = ["ChainSafe Systems <info@chainsafe.io>"]
 edition = "2018"
 
 [dependencies]
+forest_libp2p = {path = "../../node/forest_libp2p"}
```
dutterbutter: Nit: Fix spacing `{ path = "../../node/forest_libp2p" }`

node/forest_libp2p/Cargo.toml
```diff
@@ -5,10 +5,11 @@ authors = ["ChainSafe Systems <info@chainsafe.io>"]
...
-libp2p =  { git = "https://github.com/SigP/rust-libp2p", rev = "776d13ef046358964c7d64cda3295a3a3cb24743" }
 utils = { path = "../utils" }
-tokio = "0.1.22"
-futures = "0.1.29"
+libp2p = "0.15.0"
+futures = "0.3.1"
+futures-util = "0.3.1"
 log = "0.4.8"
 slog = "2.5.2"
+async-std = {version  = "1.4.0", features = ["unstable"]}
```
dutterbutter: Nit: Fix spacing `{ version  = "1.4.0", features = ["unstable"] }`

node/forest_libp2p/src/service.rs
```diff
@@ -1,42 +1,68 @@
...
 
         info!(log, "Local peer id: {:?}", peer_id);
 
         let transport = build_transport(net_keypair.clone());
 
         let mut swarm = {
-            let be = MyBehaviour::new(log.clone(), &net_keypair);
+            let be = ForestBehaviour::new(log.clone(), &net_keypair);
             Swarm::new(transport, be, peer_id)
         };
```
dutterbutter: Although not changed in this PR, should we avoid unwrapping on `Swarm::liston_on` on LN 86 (below). 

I don't see why Github does not allow to comment on code unchanged in pr reviews, saw it being requested [here as a feature](https://github.com/isaacs/github/issues/284) interestingly.

ec2: I feel like yall commented on it last time. And I don’t see how you can handle that kinda error. If you cant even listen_on, we shouldn’t handle that

node/forest_libp2p/src/service.rs
austinabell: But isn't that case technically reachable? Unreachable macro should only be used for code that is literally impossible to reach, this doesn't seem like something we want to panic on

node/forest_libp2p/src/service.rs
austinabell: Yeah that is what I meant, but yeah okay if cloning won't cause data inconsistencies and overhead. Just more wondering the use case for this because I don't see why you would need more than a reference to this (this function isn't used yet)

node/forest_libp2p/src/service.rs
austinabell: I just think it would be better to handle the error anyway instead of panicing. I see the error is handled and boxed below, so I don't see why not here also. 

Also what is the point of setting up the tcp transport when you replace it with the dns transport this line?

node/src/main.rs
austinabell: But you weren't handling the result? You needed to do something with the result or pass the error up to the main function return with `?`

node/forest_libp2p/src/service.rs
ec2: This is literally unreachable. If you look at l110, I map all the Stream items to MergeEvent::Swarm

node/forest_libp2p/src/service.rs
ec2: Cool, I'll do it 👌

node/forest_libp2p/Cargo.toml
ec2: Wish  they had a linter for cargo files 😂

austinabell: There is a TOML formatter though! I use one with VSCode, but it's not perfect

ec2: fixed

blockchain/chain/Cargo.toml
ec2: fixed

PR #213, opened 2020-02-04
Add audit to CI

austinabell:
**Summary of changes**
- Adds audit to CI to check for vulnerable dependencies 
  - Only runs on changes to any cargo file (don't want to slow the CI by running on any change)
- Updates libsecp256k1 dependency because of vulnerability

Closes 

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,14 @@
+name: Security audit
+on:
+  push:
+    paths: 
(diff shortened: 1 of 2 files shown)


ansermino:
We should make this a required check

austinabell:
> We should make this a required check

I would rather not make it required because an audit could be unrelated to changes, they can just be fixed in parallel and not hold up the CI for PRs coming in.

PR #217, opened 2020-02-07
Remove TODO rename old file to keypair.old(?)

hskang9:
**Summary of changes**
- I add a utility function `count_files` to count files in a directory
- ? willl be the number of old keystore files in the libp2p directory
- when the key does not work, the program will label the old keystore file and make new one

### DIFF
--- a/node/forest_libp2p/src/service.rs
+++ b/node/forest_libp2p/src/service.rs
@@ -16,9 +16,10 @@ use libp2p::{
     mplex, secio, yamux, PeerId, Swarm, Transport,
 };
 use slog::{debug, error, info, trace, Logger};
+use std::fs;
 use std::io::{Error, ErrorKind};
 use std::time::Duration;
-use utils::{get_home_dir, read_file_to_vec, write_to_file};
+use utils::{count_files, get_home_dir, read_file_to_vec, write_to_file};
 
 type Libp2pStream = Boxed<(PeerId, StreamMuxerBox), Error>;
 type Libp2pBehaviour = ForestBehaviour<Substream<StreamMuxerBox>>;
@@ -179,7 +180,7 @@ fn get_keypair(log: &Logger) -> Keypair {
         }
         Ok(mut vec) => {
             // If decoding fails, generate new peer id
-            // TODO rename old file to keypair.old(?)
+            // rename old file to keypair.old(number of old files in the directory)
             match ed25519::Keypair::decode(&mut vec) {
                 Ok(kp) => {
                     info!(log, "Recovered keystore from {:?}", &path_to_keystore);
@@ -188,6 +189,22 @@ fn get_keypair(log: &Logger) -> Keypair {
                 Err(e) => {
                     info!(log, "Could not decode networking keystore!");
                     trace!(log, "Error {:?}", e);
+                    // Get current number of old files
+                    let file_count =
+                        count_files(path_to_keystore[0..path_to_keystore.len() - 8].to_string());
+                    match file_count {
+                        Err(e) => {
+                            info!(log, "Error {:?}", &e);
+                        }
+                        Ok(v) => {
+                            fs::rename(
+                                path_to_keystore.clone(),
+                                path_to_keystore.clone() + &format!(".old({:})", v),
+                            )
+                            .expect("label the file as old key");
+                        }
+                    }
+                    // Generate new one
                     return generate_new_peer_id(log);
                 }
             }
--- a/node/forest_libp2p/src/service.rs
+++ b/node/forest_libp2p/src/service.rs
@@ -16,10 +16,9 @@ use libp2p::{
     mplex, secio, yamux, PeerId, Swarm, Transport,
 };
 use slog::{debug, error, info, trace, Logger};
-use std::fs;
 use std::io::{Error, ErrorKind};
 use std::time::Duration;
-use utils::{count_files, get_home_dir, read_file_to_vec, write_to_file};
+use utils::{count_files, get_home_dir, read_file_to_vec, rename_file, write_to_file};
 
 type Libp2pStream = Boxed<(PeerId, StreamMuxerBox), Error>;
 type Libp2pBehaviour = ForestBehaviour<Substream<StreamMuxerBox>>;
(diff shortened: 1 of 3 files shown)


node/utils/src/lib.rs
```diff
@@ -66,3 +66,29 @@ where
...
+/// let file_count = count_files(path_to_keystore[0..path_to_keystore.len()-8].to_string());
+/// assert_eq!(file_count, 1);
+/// ```
+pub fn count_files(dir: String) -> Result<u32> {
+    let mut file_count = 0;
+    for _i in read_dir(dir).expect("read files in the directory") {
+        file_count += 1;
+    }
+    Ok(file_count)
+}
```
austinabell: The return type is a result, but you're not passing up the error and instead panicing. The dir parameter does not need to be a String (heap allocated) and can instead be a `&str`. No need for for loop when only taking count.

Consider possibly something like: 
```
pub fn count_files(dir: &str) -> Result<usize> {
    Ok(read_dir(dir)?.enumerate().count())
}
```

node/utils/src/lib.rs
```diff
@@ -66,3 +66,29 @@ where
...
+/// - The process lacks permissions to view the contents.
+/// - The path points at a non-directory file.
+///
+/// # Example
+/// ```
+/// use utils::count_files;
+/// use utils::get_home_dir;
+///
+/// let path_to_keystore = get_home_dir() + "/.forest/libp2p/keypair";
+/// let file_count = count_files(path_to_keystore[0..path_to_keystore.len()-8].to_string());
```
austinabell: why are you indexing to length - 8?

node/utils/src/lib.rs
```diff
@@ -66,3 +66,29 @@ where
...
+/// assert_eq!(file_count, 1);
```
austinabell: file_count here is a result, should unwrap in example in previous line or this

node/utils/src/lib.rs
```diff
@@ -66,3 +66,29 @@ where
...
+/// Will return the number of files in the directory
+///
+/// # Panics
+/// It will panic if:
+/// - The provided path doesn't exist.
+/// - The process lacks permissions to view the contents.
+/// - The path points at a non-directory file.
+///
+/// # Example
+/// ```
```
austinabell:
```suggestion
/// ```no_run
```

Shouldn't have a test that uses non-controlled local file system

node/utils/src/lib.rs
```diff
@@ -66,3 +66,29 @@ where
...
 }
+
+/// Get file count in a certain directory
+/// Will return the number of files in the directory
+///
+/// # Panics
+/// It will panic if:
+/// - The provided path doesn't exist.
+/// - The process lacks permissions to view the contents.
+/// - The path points at a non-directory file.
```
austinabell: Remove this when panic removed in function

node/forest_libp2p/src/service.rs
```diff
@@ -188,6 +189,22 @@ fn get_keypair(log: &Logger) -> Keypair {
                 Err(e) => {
                     info!(log, "Could not decode networking keystore!");
                     trace!(log, "Error {:?}", e);
+                    // Get current number of old files
+                    let file_count =
+                        count_files(path_to_keystore[0..path_to_keystore.len() - 8].to_string());
```
austinabell: why the -8 indexing here also?

node/forest_libp2p/src/service.rs
```diff
@@ -188,6 +189,22 @@ fn get_keypair(log: &Logger) -> Keypair {
...
+                    match file_count {
+                        Err(e) => {
+                            info!(log, "Error {:?}", &e);
+                        }
+                        Ok(v) => {
+                            fs::rename(
+                                path_to_keystore.clone(),
+                                path_to_keystore.clone() + &format!(".old({:})", v),
+                            )
+                            .expect("label the file as old key");
```
austinabell: don't panic if this fails, should be recoverable

node/utils/tests/files.rs
```diff
@@ -116,3 +116,23 @@ fn read_from_toml() {
...
 }
+
+#[test]
+fn count_files_in_a_dir() {
+    let msg = "Hello World!";
+    let path = "./test_string_read_file/";
+    let file_name = "out.keystore";
+    if let Err(e) = write_to_file(&msg.as_bytes().to_vec(), &path, file_name) {
+        panic!(e);
+    }
```
austinabell: Can you change this to:

```
    write_to_file(msg.as_bytes(), &path, file_name).unwrap();
```

Sorry that it's set up wrong above in this file, I can make a pass through after to fix, but there is no need to heap allocate the message or manually handle the error in this test

node/utils/src/lib.rs
hskang9: Removing the filename for now. I can replace it like "/keypair".len().

node/utils/src/lib.rs
hskang9: Nice catch. I forgot to update documentation

node/forest_libp2p/src/service.rs
hskang9: ok

node/utils/src/lib.rs
hskang9: ok

node/forest_libp2p/src/service.rs
hskang9: ok

node/utils/tests/files.rs
hskang9: ok

node/utils/src/lib.rs
hskang9: ok

austinabell:
Hi @hskang9 sorry to do this but I had to refactor some files referenced in this PR for the work we are doing now. Those files are unlikely to change again in the near future if you do intend on finishing this PR. 

Not a problem if not, but to finish you should run `make lint` to add the license headers to missing files and to fix those clippy errors and run `cargo test` to clear up those test failures.

hskang9:
It is okay. It is good to see that the repository is active. I will make changes today.

austinabell:
going to close this for now because I'm assuming you want to abandon this, if you do intend to finish this feel free to open up again :)

PR #219, opened 2020-02-09
Initial Validation Checks - Message, Timestamp and Block Sig

### LINKED ISSUE #212: Implement Message Check
**Task summary**

Part of block validation requires a check of the messages, which includes and not limited to, verifying bls aggregate, balance and nonce check of actor, signature verification and root check. Although some of the functionality exists (e.g. signature verification), much of it is not implemented. 

**Specification reference**

- [Message portion of validation](https://filecoin-project.github.io/specs/#block-syntax-validation)
- [Messages](https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md)

- [Lotus checkBlockMessages](https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L705)
- Used with #78 , specifically for messages. 
- Verify bls aggregate has its own issue #44

dutterbutter:
**Summary of changes**
- Added initial block validation checks
- Specifically, messages validation check to ensure valid state transitions, timestamp checks, and block signature checks
- Added `required_funds` method to `Message` type
- Added `check_block_signature` method to `BlockHeader` type

Closes #212 

Validation links used for reference:
- [Lotus Check Message Check](https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L710:23)
- [Spec Validation](https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md)
- [Block Validations Spec](https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/mining.md#block-validation)


blockchain/sync_manager/src/sync.rs
```diff
@@ -122,6 +146,141 @@ impl<'a> Syncer<'a> {
...
+        }
+
+        Ok(())
+    }
+
+    /// Validates block semantically according to https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md
+    pub fn validate(&self, block: Block) -> Result<(), Error> {
+        // get header from full block
+        let header = block.to_header();
+        let base_tipset = self.load_fts(TipSetKeys::new((*header.parents().cids).to_vec()))?;
```
austinabell:
```suggestion
        let base_tipset = self.load_fts(TipSetKeys::new(header.parents().cids.clone()))?;
```
Just slightly confusing

austinabell: Can maybe move this retrieval of the tipset to below when you actually need it (avoid retrieval if following checks fail)

blockchain/sync_manager/src/sync.rs
```diff
@@ -122,6 +146,141 @@ impl<'a> Syncer<'a> {
...
+        let allowable_clock_drift: u64 = 1;
+        let time_now = SystemTime::now().duration_since(UNIX_EPOCH)?;
+        if header.timestamp() > time_now.as_secs() + allowable_clock_drift {
+            return Err(Error::Validation("Block was from the future".to_string()));
+        }
+        if header.timestamp() > time_now.as_secs() {
+            return Err(Error::Validation(
+                "Got block from the future, but within threshold".to_string(),
+            ));
+        }
```
austinabell: I'm confused, why do we need to handle these seperately if they return the same error variant?

blockchain/sync_manager/src/sync.rs
```diff
@@ -122,6 +146,141 @@ impl<'a> Syncer<'a> {
...
+        // check if block has been signed
+        if header.signature().bytes().is_empty() {
+            return Err(Error::Validation("Signature is nil in header".to_string()));
+        }
+
+        // first check that it is not in the future; see https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md
+        // allowing for some small grace period to deal with small asynchrony
+        // using allowable_clock_drift from Lotus; see https://github.com/filecoin-project/lotus/blob/master/build/params_shared.go#L34:7
+        let allowable_clock_drift: u64 = 1;
+        let time_now = SystemTime::now().duration_since(UNIX_EPOCH)?;
```
austinabell: Just want to confirm you did verify that the header timestamp is time from unix epoch? The spec was very vague

blockchain/sync_manager/src/sync.rs
```diff
@@ -28,6 +34,25 @@ pub struct Syncer<'a> {
...
 
+/// Message data used to ensure valid state transition
+struct MsgMetaData {
+    balance: BigUint,
+    sequence: u64,
+}
+
+struct MsgCheck {
+    metadata: HashMap<Address, MsgMetaData>,
+}
```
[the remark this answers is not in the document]
dutterbutter: No plans for future use, however having the value be a struct with the two fields need for message validations makes it easier for get and put operations. What would be your suggestion here.

blockchain/sync_manager/src/sync.rs
dutterbutter: Yes, as mentioned [here](https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/mining.md#block-validation) under the timestamp bullet point.

blockchain/blocks/src/header.rs
```diff
@@ -279,6 +279,22 @@ impl BlockHeader {
         self.cached_cid = Cid::from_bytes_default(&self.cached_bytes).map_err(|e| e.to_string())?;
         Ok(())
     }
+    /// Check to ensure block signature is valid
+    pub fn check_block_signature(&self, addr: &Address) -> Result<(), Error> {
```
ansermino: Is there any reason to make the address a param? Aren't we always validating against the miner?

blockchain/sync_manager/src/sync.rs
```diff
@@ -122,6 +146,142 @@ impl<'a> Syncer<'a> {
...
+
+        // check if block has been signed
+        if header.signature().bytes().is_empty() {
+            return Err(Error::Validation("Signature is nil in header".to_string()));
+        }
+
+        // first check that it is not in the future; see https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md
+        // allowing for some small grace period to deal with small asynchrony
+        // using ALLOWABLE_CLOCK_DRIFT from Lotus; see https://github.com/filecoin-project/lotus/blob/master/build/params_shared.go#L34:7
+        const ALLOWABLE_CLOCK_DRIFT: u64 = 1;
```
ansermino: This time logic is quite involved, perhaps move to a sub method?

blockchain/sync_manager/src/sync.rs
```diff
@@ -122,6 +146,142 @@ impl<'a> Syncer<'a> {
...
+        let sm_root = self.compute_msg_data(&block)?;
+        if block.to_header().messages() != &sm_root {
+            return Err(Error::InvalidRoots);
+        }
+
+        Ok(())
+    }
+
+    /// Validates block semantically according to https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md
+    pub fn validate(&self, block: Block) -> Result<(), Error> {
```
ansermino: What's the rationale for this being a method on Syncer? At first glance is feels like this logic should be associated with `Block`

blockchain/sync_manager/src/sync.rs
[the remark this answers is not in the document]
austinabell: Just to use the hashmap where you used the MsgCheck struct, then each usage doesn't have to be like this
```
msg_meta_data
                .metadata
                .<command>()
```

and you can just remove all metadata referencing and just call:
```
msg_meta_data.<command>()
```
I had made those changes locally to make sure there was no issues

blockchain/sync_manager/src/sync.rs
dutterbutter: A few reasons, avoiding circular dependencies as the validation requires a lot of pieces from other crates (especially when we complete all checks) of which those crates are used already through the sync crate, the `syncer.rs` file provides the functionality required for syncing which validation is a part of whereas, block.rs (at least right now) just defines the Block structures, the Block of which we are validating will be passed through the syncer via `inform_new_head`, and it's also how Lotus has it in place.

blockchain/blocks/src/header.rs
dutterbutter: You need the address in order to validate the signature

ansermino: `addr == self.miner_address`, doesn't seem like it needs to be passed in

dutterbutter: Different addresses, when we check_block_signature we are passing in the raw worker miner address not the address included in the header. See https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L611

blockchain/sync_manager/src/sync.rs
dutterbutter: Moved to its own method on BlockHeader

blockchain/sync_manager/src/sync.rs
dutterbutter: As discussed, given our current crate design we run into circular dependencies related to the `chain_store` if this function is moved to `block.rs`. If we want to change how our crates are organized to allow for the movement of this function including all the remaining validation checks (e.g. state which all requires access to chain_store) I think that should be put forth in a separate PR.

blockchain/sync_manager/src/sync.rs
```diff
@@ -102,7 +112,7 @@ impl<'a> Syncer<'a> {
         Ok(fts)
     }
     /// Returns a reconstructed FullTipset from store if keys exist
-    fn load_fts(&self, keys: TipSetKeys) -> Result<FullTipset, Error> {
+    pub fn load_fts(&self, keys: TipSetKeys) -> Result<FullTipset, Error> {
```
austinabell: why does this need to be public?

vm/actor/src/lib.rs
```diff
@@ -40,6 +40,14 @@ impl ActorState {
...
         }
     }
+    /// Returns sequence from actor state
+    pub fn sequence(&self) -> &u64 {
+        &self.sequence
+    }
+    /// Returns balance from actor state
+    pub fn balance(&self) -> &BigUint {
+        &self.balance
+    }
```
austinabell: can just delete these since the fields are not protected (and can make state and code_id public also for consistency later but doesn't matter)

blockchain/sync_manager/src/sync.rs
austinabell: I am fine with this now as well because it relies on the syncer to retrieve the full tipset, so would in fact introduce a circular dep for blocks to use the sync_manager crate (I didn't notice this when we chatted this morning)

PR #221, opened 2020-02-11
Refactor keypair retrieval and saving

austinabell:
**Summary of changes**
- Issues found when working with network tests, keypair should not be automatically persisted when a keypair is not found. There was some janky logic that I wanted to clean up as well.
  - Essentially the problem is that a keypair is automatically pulled from keystore and used when generating new Libp2pService
- Cleaned up tests that unnecessarily heap allocated bytes and handled errors specifically in the test (found in #217) 

Closes 

### DIFF
--- a/node/forest_libp2p/src/service.rs
+++ b/node/forest_libp2p/src/service.rs
@@ -167,27 +167,25 @@ pub fn build_transport(local_key: Keypair) -> Boxed<(PeerId, StreamMuxerBox), Er
         .boxed()
 }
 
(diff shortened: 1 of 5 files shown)


node/forest_libp2p/src/service.rs
```diff
@@ -169,46 +168,26 @@ pub fn build_transport(local_key: Keypair) -> Boxed<(PeerId, StreamMuxerBox), Er
 }
 
 /// Fetch keypair from disk, or generate a new one if its not available
```
ec2: Remove this comment

Issue #78, opened 2019-12-07
Add Block Validation Check Methods


ansermino:
While these exact parameters are likely to change we can start to include some these checks and document those that aren't yet implementable.

Must meet the criteria outlined here:
- [Validation.md](https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md)

In the spec:
- [Syntactical Validation](https://filecoin-project.github.io/specs/#block-syntax-validation)
- [Semantic Validation](https://filecoin-project.github.io/specs/#block-semantic-validation)

Reference from existing implementations:
- [Lotus Block Validations](https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L482)
- [Go-Filecoin Block Validations](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/consensus/block_validation.go)

**Tasks required to complete this task:**
- [ ] Block validation trait(s)
- [ ] Semantic validations
- [ ] Syntactic validations

amerameen:
On Ice for now due to dependancies on VM and HAMT

austinabell:
Some of this has been done, but should revisit to make sure we complete all or document what is missing, shouldn't be on ice for any reason I don't think

amerameen:
@nannick please update the issue with what our implementation is missing.

Issue #216, opened 2020-02-06
Consolidate blockchain errors and investigate error handling


austinabell:
**Issue summary**

Error types in the blockchain crates are duplicated in a few places and can be consolidated to remove the need to convert from multiple enums when interacting with multiple crates within the blockchain system. 

This could be a good opportunity to look into an alternative for error handling that could be less explicit and tedious than we currently do if not needed. There are many different directions we could go with this so would love to get opinions for what should be a standard.

The main reasons for updating this is for:
- Ease of use not having to provide a `From` implementation to another error or mapping in the function call
- More succinct error messages (Each time an error conversion happens now, it wraps the error message in another, which will get very annoying to parse when this happens a few times)
- More accurate error messages
  - Sometimes the `From` implementation will be not correct and convert implicitly to an incorrect error

It makes sense to have an Error type for a specific crate that will be deployed and used specifically, but for internal types within systems, it doesn't make much sense.

**Other information and links**

Error types referenced:
https://github.com/ChainSafe/forest/blob/6d4304e292857343ef614785be8f272c4aa7eef1/blockchain/chain/src/store/errors.rs
https://github.com/ChainSafe/forest/blob/6d4304e292857343ef614785be8f272c4aa7eef1/blockchain/sync_manager/src/errors.rs
https://github.com/ChainSafe/forest/blob/6d4304e292857343ef614785be8f272c4aa7eef1/blockchain/blocks/src/errors.rs

PR #224, opened 2020-02-12
State Manager and initial miner retrieval methods 

dutterbutter:
**Summary of changes**
- Added `StateManager` and retrieval methods for state objects 
- Added `StorageMinerActor`, `MinerInfo` and `StorageMinerActorState` structures used to retrieve data about miners
- Included the retrieval of `slashed_at` and `miner_sector_size` to block validation method
- Updated comments 

It does not close any particular task although it pushes #78 forward.

Block validation will remain incomplete until we implement missing VM components required to retrieve storage power, get a raw miner worker address and can validate miner.


blockchain/sync_manager/src/sync.rs
```diff
@@ -15,14 +15,16 @@ use libp2p::core::PeerId;
...
+/// Syncer updates the key-value store based on series of validation checks adhering to consensus rules, can query
+/// the network for blocks and assists in informing the network of incoming blocks
 pub struct Syncer<'a> {
     // TODO add ability to send msg to all subscribers indicating incoming blocks
-    // TODO add state manager
     // TODO add block sync
-
+    state_manager: StateManager<'a>,
     // manages sync buckets
     sync_manager: SyncManager<'a>,
```
austinabell: Should assign a different lifetime for these variables (unless they have the same lifetime for sure)

vm/state_manager/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+use address::Address;
+use amt::BlockStore;
+use chain::ChainStore;
+use encoding::de::DeserializeOwned;
+use state_tree::{HamtStateTree, StateTree};
+
+/// Intermediary for retrieving state objects and updating actor states
+pub struct StateManager<'a> {
+    cs: &'a ChainStore<'a>,
+    tree: HamtStateTree,
```
austinabell: Ideal if this variable is generic, something like:
```
pub struct StateManager<'a, T: StateTree> {
// ...
    tree: T,
}
```

vm/state_manager/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+use actor::{MinerInfo, StorageMinerActorState};
+use address::Address;
+use amt::BlockStore;
+use chain::ChainStore;
+use encoding::de::DeserializeOwned;
+use state_tree::{HamtStateTree, StateTree};
+
+/// Intermediary for retrieving state objects and updating actor states
+pub struct StateManager<'a> {
+    cs: &'a ChainStore<'a>,
```
austinabell: My worry is that the chainstore will have to be mutable (updating heaviest tipset) and requiring an immutable reference for this type will inhibit this, thoughts?

vm/state_manager/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    tree: HamtStateTree,
+}
+
+impl<'a> StateManager<'a> {
+    /// constructor
+    pub fn new(cs: &'a ChainStore, tree: HamtStateTree) -> Self {
```
austinabell: Same point as before, would be better if state tree type was not predefined and made generic (would have to constrain the type in the `impl`

vm/state_manager/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+        Self { cs, tree }
+    }
+    /// Loads actor state from IPLD Store
+    fn load_actor_state<T>(&self, addr: &Address) -> Result<T, Error>
```
austinabell: Why is the actor state generic type? Won't this always be the same struct?

vm/state_manager/src/errors.rs
```diff
@@ -0,0 +1,28 @@
...
+    State(String),
+    /// Error originating from amt
+    AMT(String),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::State(msg) => write!(f, "Error state data: {}", msg),
+            Error::AMT(msg) => write!(f, "Error originating from the AMT: {}", msg),
```
austinabell: Here is another example of appending strings to error messages, maybe consider not adding the prefix to the error message for these to remove redundant messages?

vm/actor/src/builtin/miner_actor.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn slashed_at(&self) -> &u64 {
+        &self.slashed_at
+    }
+    /// Returns cid that can be used to retrieve static information regarding this miner
+    pub fn info(&self) -> &Cid {
+        &self.info
+    }
+}
+
+impl MinerInfo {
```
austinabell: can you just move this to below the definition for readability?

vm/actor/src/builtin/miner_actor.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Container representing storage miner actor state
+pub struct StorageMinerActorState {
+    // TODO add proving_set, post_state
+
+    // The height at which this miner was slashed at.
+    slashed_at: u64,
+    // Sectors this miner has committed
+    sectors: Cid,
+    // Contains static info about this miner
+    info: Cid,
```
austinabell: Can you please make all of the comments `///` for cargo documentation (other places in code but I won't comment on all)

vm/actor/src/builtin/errors.rs
```diff
@@ -0,0 +1,21 @@
...
+    Encoding(String),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::State(msg) => write!(f, "Error state data: {}", msg),
+            Error::Encoding(msg) => write!(f, "Error originating from Encoding type: {}", msg),
+        }
+    }
```
austinabell: Same as previous (appending additional string message to error string) if the errors don't need to be checked for specifically, consider just returning String for the error type

blockchain/sync_manager/src/sync.rs
```diff
@@ -15,14 +15,16 @@ use libp2p::core::PeerId;
 use message::{Message, MsgMeta};
 use num_bigint::BigUint;
 use raw_block::RawBlock;
-use state::{HamtStateTree, StateTree};
+use state_manager::StateManager;
+use state_tree::{HamtStateTree, StateTree};
```
austinabell: side note the HamtStateTree usage here can be removed by referencing the state tree as the trait and not the specific implementation

vm/state_manager/src/lib.rs
dutterbutter: No, in its current usage it could be the `StorageMinerActorState ` or `MinerInfo` struct

vm/state_manager/src/lib.rs
dutterbutter: Yeah, I agree with your concern, will look to improve.

vm/actor/src/builtin/miner_actor.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Container representing storage miner actor state
+pub struct StorageMinerActorState {
+    // TODO add proving_set, post_state
+    /// The height at which this miner was slashed at.
+    slashed_at: u64,
+    /// Sectors this miner has committed
+    sectors: Cid,
+    /// Contains static info about this miner
+    info: Cid,
+}
```
austinabell: What is the specific reason to protect these fields? For something like this it may make sense to just have the fields public (should any interaction be protected?)

vm/actor/src/builtin/miner_actor.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Worker account for this miner
+    /// This will be the key that is used to sign blocks created by this miner, and
+    /// sign messages sent on behalf of this miner to commit sectors, submit PoSts, and
+    /// other day to day miner activities
+    worker_address: Address,
+    /// Libp2p identity that should be used when connecting to this miner
+    peer_id: PeerId,
+    /// Amount of space in each sector committed to the network by this miner
+    sector_size: u64,
+}
```
austinabell: This struct makes more sense to not make fields public, but there is no way to construct it currently so I'm wondering how this will be actually used.

Would it make sense to just prefix the unused with _ and not allow unused code at start to easily identify which are unused?

vm/actor/src/builtin/miner_actor.rs
```diff
@@ -0,0 +1,95 @@
...
+            .map_err(|_| de::Error::custom("Error parsing PeerId type to String"))?;
+
+        Ok(Self {
+            owner,
+            worker_address,
+            peer_id,
+            sector_size,
+        })
+    }
+}
```
austinabell: So optional but when using `from_str` you can actually just use `.parse()` and not need to import the FromStr trait (was stabilized). You also don't need that intermediate struct as all types are bound when you generate MinerInfo from it. You could do something like this if you wanted:

```
impl<'de> de::Deserialize<'de> for MinerInfo {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: de::Deserializer<'de>,
    {
        let (owner, worker_address, string_peer_id, sector_size): (_, _, String, _) =
            Deserialize::deserialize(deserializer)?;

        Ok(Self {
            owner,
            worker_address,
            sector_size,
            peer_id: string_peer_id
                .parse()
                .map_err(|_| de::Error::custom("Error parsing PeerId type from String"))?,
        })
    }
}

```

Where you only need to constrain the string type on the tuple deserialize (not sure why, since it should be constrained when using `.parse()`)

vm/actor/src/code.rs
```diff
@@ -2,10 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use cid::Cid;
+use serde::{Deserialize, Serialize};
 
 /// CodeID is the reference to the code which is attached to the Actor state.
 /// There are builtin IDs and the option for custom code with a Cid
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
```
austinabell: I don't care, but just to note: enum default serialization is not very efficient so if the codeID is going to be serialized and deserialized a lot, this should probably be avoided.

vm/actor/src/lib.rs
```diff
@@ -50,3 +49,33 @@ pub trait Actor {
...
+    {
+        let (code_id, state, sequence, balance) = Deserialize::deserialize(deserializer)?;
+        Ok(Self {
+            code_id,
+            state,
+            sequence,
+            balance,
+        })
+    }
+}
```
austinabell: I don't see why this is needed (extension of serializing CodeID) I removed from this PR with no issue

blockchain/sync_manager/src/sync.rs
```diff
@@ -15,21 +15,23 @@ use libp2p::core::PeerId;
...
-pub struct Syncer<'a> {
+/// Syncer updates the key-value store based on series of validation checks adhering to consensus rules, can query
+/// the network for blocks and assists in informing the network of incoming blocks
+pub struct Syncer<'a, 'b, T: StateTree> {
     // TODO add ability to send msg to all subscribers indicating incoming blocks
-    // TODO add state manager
     // TODO add block sync
-
-    // manages sync buckets
+    state_manager: StateManager<'a, 'b, T>,
```
austinabell: doc comment on this would be nice

vm/actor/src/builtin/miner_actor.rs
dutterbutter: No particular reason just defaulted to protecting, will make pub and remove getter methods to retrieve fields

blockchain/sync_manager/src/errors.rs
```diff
@@ -37,20 +38,14 @@ impl fmt::Display for Error {
...
-            Error::State(msg) => write!(f, "Error originating from the State: {}", msg),
-            Error::Validation(msg) => write!(f, "Error validating data: {}", msg),
+            Error::Blockchain(msg) => write!(f, "{}", msg),
+            Error::KeyValueStore(msg) => write!(f, "{}", msg),
+            Error::Encoding(msg) => write!(f, "{}", msg),
+            Error::InvalidCid(msg) => write!(f, "{}", msg),
+            Error::Store(msg) => write!(f, "{}", msg),
+            Error::AMT(msg) => write!(f, "{}", msg),
+            Error::State(msg) => write!(f, "{}", msg),
+            Error::Validation(msg) => write!(f, "{}", msg),
```
ansermino: Shouldn't the error type be specified as part of the returned message? `msg` may be ambiguous

blockchain/sync_manager/src/sync.rs
```diff
@@ -233,6 +239,12 @@ impl<'a> Syncer<'a> {
         // see https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L611
         header.check_block_signature(header.miner_address())?;
 
+        // TODO: incomplete, still need to retrieve power in order to ensure ticket is the winner
+        let _slash = self.state_manager.miner_slashed(header.miner_address())?;
```
ansermino: Why do we check this? Can't correlate with spec

vm/state_manager/src/errors.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use amt::Error as AmtErr;
+use std::fmt;
+
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    /// Error orginating from state
+    State(String),
+    /// Error originating from amt
+    AMT(String),
```
ansermino: Whats the difference between these? Aren't AMT errors also state errors?

blockchain/sync_manager/src/errors.rs
dutterbutter: Discussed error messages with @austinabell last week and decided to remove the extra message as it looked quite messy with the included `msg`. Also, the error variant type would indicate the error type. Nonetheless, I am in favor of determining a standard for our error.rs files moving forward.

blockchain/sync_manager/src/sync.rs
dutterbutter: https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md

Under Semantical Validation: must be from a valid miner, i.e. has not been slashed

vm/state_manager/src/errors.rs
dutterbutter: State error originates from `state_manager`, whereas, AMT originates from the AMT. Happy to consolidate if that's the group's consensus.

blockchain/sync_manager/src/errors.rs
austinabell: Ya the reason another message isn't appended to these are they are essentially forwarded error messages from another type. Can rework how we handle errors or in this case have the error it is forwarding be included as the variant type instead of `String`

austinabell: #216

vm/actor/src/builtin/errors.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::State(msg) => write!(f, "{}", msg),
+            Error::Encoding(msg) => write!(f, "{}", msg),
+        }
+    }
+}
```
austinabell: This is never used. Wouldn't make sense to have a builtin actor specific errors anyway

I think for error handling in cases like this where it would just be forwarding the error as a wrapped string, probably makes sense to just return a `String` as the error return, as it shouldn't matter what variant the error was, you would just need a string to log to console explaining.

vm/state_manager/src/errors.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use amt::Error as AmtErr;
+use std::fmt;
+
+#[derive(Debug, PartialEq)]
+pub enum Error {
```
austinabell: If this type is just forwarding errors, can just change the Error variant to string and get rid of this. Only preference for this though, can keep if you'd like

Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
     "vm/message",
     "vm/runtime",
     "vm/state_tree",
+    "vm/state_manager",
```
austinabell: doesn't make much sense to me for the state manager to be inside of the vm folder, since the state manager is external to the vm? Thoughts from anyone else?

blockchain/sync_manager/src/sync.rs
ansermino: That doesn't appear in the current spec

vm/state_manager/src/errors.rs
ansermino: I feel like the distinction would be lost on most people, we also don't really expect AMT errors as they would probably be fatal.

Cargo.toml
dutterbutter: Yeah that works for me, I put it there because of the state managers interactions with the state tree, and actors but could definitely be moved into blockchain component.

austinabell: Obviously no functional change, but organizationally it makes more sense to me

blockchain/sync_manager/src/sync.rs
dutterbutter: The spec uses the term 'plausible miner' which to me is the equivalent to 'valid miner' the term given in the link I posted above which uses slashing as an example of what makes a 'valid miner', and in lotus, it's being used for their block validation. I am happy to remove but I want to make sure I provide relevant information as to why I included it.

vm/state_manager/src/errors.rs
dutterbutter: Yeah that makes sense to me! Will change.

vm/actor/src/code.rs
austinabell:
```suggestion
#[derive(PartialEq, Eq, Clone, Debug)]
```

Sorry one last quick change and good with me, just want to not be confusing that the CodeID will be serialized and that the format is checked

vm/actor/src/code.rs
```diff
@@ -2,10 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use cid::Cid;
+use serde::{Deserialize, Serialize};
```
austinabell:
```suggestion
```
This will need to be removed with previous

vm/actor/src/builtin/miner_actor.rs
```diff
@@ -0,0 +1,78 @@
...
+
+impl<'de> de::Deserialize<'de> for MinerInfo {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        let (_owner, _worker_address, string_peer_id, sector_size): (_, _, String, _) =
```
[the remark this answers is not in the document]
austinabell: Because only String needs to be explicitly defined, others are inferred based on usage below

PR #228, opened 2020-02-13
Fix cbor serialization formats and cleanup

### LINKED ISSUE #143: CBOR Encoding

austinabell:
**Summary of changes**
- Implement/ fix TxMeta (or message root) serialization to tuple serialization
- Implement serialization of signed messages (will be important for signatures soon)
- Cleaned up usages of a few small things for consistency
- Adds cid function from block to access cid for block easier from header

These things were bugging me a bit and wanted to do now as we would need soon

Closes #143

### DIFF
--- a/blockchain/blocks/src/block.rs
+++ b/blockchain/blocks/src/block.rs
@@ -47,10 +47,9 @@ impl Block {
 }
 
 /// Tracks the merkleroots of both secp and bls messages separately
-#[derive(Clone, Debug, PartialEq, Default)]
 pub struct TxMeta {
-    pub bls_messages: Cid,
-    pub secp_messages: Cid,
+    pub bls_message_root: Cid,
+    pub secp_message_root: Cid,
 }
 
 impl Serialize for TxMeta {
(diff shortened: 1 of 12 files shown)


blockchain/blocks/src/block.rs
```diff
@@ -29,31 +29,52 @@ pub struct Block {
...
+    {
+        let (bls_messages, secp_messages) = Deserialize::deserialize(deserializer)?;
+        Ok(TxMeta {
+            bls_messages,
+            secp_messages,
+        })
+    }
+}
 
 /// ElectionPoStVerifyInfo seems to be connected to VRF
```
dutterbutter: You can also remove this struct, pretty sure its duplicated in ticket.rs

austinabell: It was duplicated in the message crate, I removed that one because it should live here and just cleaned up references to it.

dutterbutter: I am referring to `ElectionPoStVerifyInfo` sorry

austinabell: ah gotcha, unrelated to this PR but I will remove because I'm on a spree of cutting code 🔪

PR #229, opened 2020-02-13
Libp2p RPC protocol and Blocksync

ec2:
**Summary of changes**
- Created the framework for libp2p RPC
- Created the libp2p protocol for Blocksync

Closes 


node/forest_libp2p/src/rpc/codec.rs
```diff
@@ -0,0 +1,106 @@
...
+    type Error = RPCError;
+    type Item = RPCRequest;
+
+    fn decode(&mut self, bz: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if bz.is_empty() {
+            return Ok(None);
+        }
+
+        Ok(Some(RPCRequest::Blocksync(
+            // Reaplce map
```
dutterbutter: spelling mistake

node/forest_libp2p/src/rpc/behaviour.rs
```diff
@@ -0,0 +1,107 @@
...
+use libp2p::swarm::{
+    protocols_handler::ProtocolsHandler, NetworkBehaviour, NetworkBehaviourAction, PollParameters,
+};
+use libp2p::{Multiaddr, PeerId};
+use std::marker::PhantomData;
+
+pub struct RPC<TSubstream> {
+    /// Queue of events to processed.
+    events: Vec<NetworkBehaviourAction<RPCEvent, RPCMessage>>,
+    /// Pins the generic substream.
```
dutterbutter: I don't believe these comments will be visible in our docs as the `RPC` struct is not commented on. If the intent was to have them visible to the docs I think you will need to also document the struct.

node/forest_libp2p/tests/rpc_test.rs
```diff
@@ -0,0 +1,117 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+#![cfg(test)]
+
+use async_std::task;
+use forest_libp2p::rpc::{Message, RPCEvent, RPCMessage, RPCRequest, RPCResponse, Response, RPC};
+// use forest_libp2p::{ForestBehaviourEvent, Libp2pConfig, Libp2pService};
```
dutterbutter: remove

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub options: u64,
+}
+
+impl Serialize for Message {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let value = (self.start.clone(), self.request_len, self.options);
+        ser::Serialize::serialize(&value, serializer)
```
austinabell: serialize using references to the data:
```
(&self.start, &self.request_len, &self.options).serialize(serializer)
```

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub message: String,
+}
+
+impl Serialize for Response {
+    fn serialize<S>(&self, serializer: S) -> Result<<S as Serializer>::Ok, <S as Serializer>::Error>
+    where
+        S: Serializer,
+    {
+        let value = (self.chain.clone(), self.status, self.message.clone());
+        Serialize::serialize(&value, serializer)
```
austinabell: same as prev:
```
(&self.chain, &self.status, &self.message).serialize(serializer)
```

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,113 @@
...
+        S: Serializer,
+    {
+        let value = (
+            self.blocks.clone(),
+            self.secp_msgs.clone(),
+            self.secp_msg_includes.clone(),
+            self.bls_msgs.clone(),
+            self.bls_msg_includes.clone(),
+        );
+        Serialize::serialize(&value, serializer)
```
austinabell: same as previous:

```
        (
            &self.blocks,
            &self.secp_msgs,
            &self.secp_msg_includes,
            &self.bls_msgs,
            &self.bls_msg_includes,
        )
            .serialize(serializer)
```

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<<S as Serializer>::Ok, <S as Serializer>::Error>
+    where
+        S: Serializer,
+    {
+        let value = (self.chain.clone(), self.status, self.message.clone());
+        Serialize::serialize(&value, serializer)
+    }
+}
+impl<'de> Deserialize<'de> for Response {
+    fn deserialize<D>(deserializer: D) -> Result<Self, <D as Deserializer<'de>>::Error>
```
austinabell:
```suggestion
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
```

Can you clean these up as well (on all of the serialize and deserialize fns for consistency)?

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,113 @@
...
+            message,
+        })
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct TipSetBundle {
+    pub blocks: Vec<BlockHeader>,
+    pub secp_msgs: Vec<UnsignedMessage>,
+    pub secp_msg_includes: Vec<Vec<u64>>,
```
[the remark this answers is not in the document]
austinabell: it is the indexes of which messages are included in which block, index of outer vec corresponds to index of blocks, inner index corresponds to secp_msgs index

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,113 @@
...
+use forest_blocks::BlockHeader;
+use forest_cid::Cid;
+use forest_encoding::{
+    de::{self, Deserialize, Deserializer},
+    ser::{self, Serialize, Serializer},
+};
+use forest_message::{SignedMessage, UnsignedMessage};
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Message {
```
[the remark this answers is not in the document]
ec2: fixed

node/forest_libp2p/src/behaviour.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use super::rpc::{RPCEvent, RPC};
+use crate::rpc::RPCMessage;
```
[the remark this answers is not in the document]
ec2: fixed

node/forest_libp2p/src/rpc/blocksync_message.rs
ec2: fixed

node/forest_libp2p/src/rpc/blocksync_message.rs
ec2: fixed

node/forest_libp2p/src/rpc/blocksync_message.rs
ec2: fixed

ec2: fixed

node/forest_libp2p/src/rpc/blocksync_message.rs
ec2: fixed

node/forest_libp2p/src/rpc/blocksync_message.rs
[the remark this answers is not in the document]
ec2: > Generally I feel like this could use more docs.

Where would you like to see more docs? A vast majority of this PR is implementing different traits, which if I understand correctly from @austinabell , comments won't show up in the docs.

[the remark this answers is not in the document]
austinabell: > > Generally I feel like this could use more docs.
> 
> Where would you like to see more docs? A vast majority of this PR is implementing different traits, which if I understand correctly from @austinabell , comments won't show up in the docs.

Yeah not the trait impls but there is a lot undocumented, just go through everything public and check to make sure it has some description, even if very brief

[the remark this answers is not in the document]
ec2: I've done a little more docs. Is there any that i missed?

node/forest_libp2p/src/rpc/codec.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn from(err: std::io::Error) -> Self {
+        Self::Custom(err.to_string())
+    }
+}
+
+impl From<EncodingError> for RPCError {
+    fn from(_: EncodingError) -> Self {
+        Self::Codec
+    }
+}
```
ansermino: Related to above, doesn't this mean we discard the message from the endcoding error?

ec2: youre right. fixed. missed that

node/forest_libp2p/src/rpc/codec.rs
```diff
@@ -0,0 +1,106 @@
...
+use forest_encoding::{error::Error as EncodingError, from_slice, to_vec};
+use futures_codec::{Decoder, Encoder};
+use std::fmt;
+
+pub struct InboundCodec;
+pub struct OutboundCodec;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum RPCError {
+    Codec,
```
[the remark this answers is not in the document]
ec2: youre right. fixed. missed that

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,125 @@
...
+            chain,
+            status,
+            message,
+        })
+    }
+}
+
+/// Contains the blocks and messages in a particular tipset
+#[derive(Clone, Debug, PartialEq)]
+pub struct TipSetBundle {
```
dutterbutter: Slightly confused with this type...could we not utilize the pre-existing `FullTipSet` which has a `Vec<Blocks>` that contains the bls and secp messages and would describe the block each message belongs to? 

Further, based on the spec and our existing implementation `blocks` should only contain a single header not a `Vec<BlockHeader>`?

austinabell: nah, this type defines specifically what is sent via the blocksync protocol. They use the tipsetbundle to remove redundant data passed over the wire. 

Also what type are you saying to only include one header? for even a single tipset there should always be a vector of headers (or blocks)

node/forest_libp2p/src/rpc/blocksync_message.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/// The response to a BlockSync request.
+#[derive(Clone, Debug, PartialEq)]
+pub struct BlockSyncResponse {
+    /// The tipsets requested
+    pub chain: Vec<TipSetBundle>,
+    /// Error code
+    pub status: u64,
+    pub message: String,
```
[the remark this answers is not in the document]
austinabell: Just a note about this, the message isn't in the spec so it's possible it will be removed in the future, but it exists in Lotus so we are including for now

node/forest_libp2p/src/rpc/blocksync_message.rs
dutterbutter: I see that makes sense, although not completely clear how its a compressed Full Tipset as it contains the same data fields (e.g. blockheader fields and messages) and even adds an additional field with `msg_includes`? Could you not still use message vectors from a FullTipset?

austinabell: It's "compressed" because a tipset could have duplicate messages in the tipset in different blocks, the FullTipset is a vector of full blocks with messages, and the TipSetBundle has a dump of all headers, unique messages, and indexes for headers to messages

node/forest_libp2p/src/rpc/handler.rs
```diff
@@ -0,0 +1,362 @@
...
+use libp2p::{InboundUpgrade, OutboundUpgrade};
+use smallvec::SmallVec;
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+    time::{Duration, Instant},
+};
+
+/// The time (in seconds) before a substream that is awaiting a response from the user times out.
+pub const RESPONSE_TIMEOUT: u64 = 10;
```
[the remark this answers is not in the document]
austinabell: yeah it was arbitrary, can tweak it later but 10 seconds seemed like a reasonable timeout for an RPC response

PR #230, opened 2020-02-13
Refactor blockstore and Cid for usage

austinabell:
**Summary of changes**
- Moved blockstore into own crate to be used more easily by crates now and in the near future
  - Removes need to import AMT to use just that and also removes need to have duplicates inside of HAMT or otherwise 
- Refactored cid crate for easier upstream and non-forest specific logic
  - Having default cid function is nice, but removing for now to be specific about which hashes are actually being used. (default was 2b512 before when 256 wasn't available and wouldn't run into issues until we generated our own Cids to compare with Lotus)
  - Makes easier to upstream or share changes with rust-ipld in future to not have opinionated functions

This shouldn't cause any actual conflicts, maybe only minor if you were referencing blockstore in AMT crate or generating a cid from bytes default (I don't think that's the case for any of you guys rn)


blockchain/sync_manager/Cargo.toml
```diff
@@ -10,16 +10,17 @@ amt = { package = "ipld_amt", path = "../../ipld/amt" }
 blocks = { package = "forest_blocks", path = "../blocks" }
 db = { path = "../../node/db" }
 encoding = { package = "forest_encoding", path = "../../encoding" }
-libp2p =  { git = "https://github.com/SigP/rust-libp2p", rev = "776d13ef046358964c7d64cda3295a3a3cb24743" }
+libp2p = { git = "https://github.com/SigP/rust-libp2p", rev = "776d13ef046358964c7d64cda3295a3a3cb24743" }
```
dutterbutter: Shouldn't this be `libp2p = "0.15.0"`?

austinabell: yeah I didn't notice, this was auto formatted. Will swap. This must have come in after the swap to libp2p base

austinabell:
Re-requested review since I also removed raw-block crate by updating the Cbor trait we use in encoding. 

Also removed dependency of Cid crate on our local encoding utility and seperated the serde stuff into a feature so now the Cid crate can very easily be upstreamed as is.

PR #231, opened 2020-02-15
Local bigint serialization

austinabell:
**Summary of changes**
- Implemented local serialization for big int types
- Updated dependency references from using my bigint fork

So since the serde macros don't call the associated variant of serialize, you can use a module in place of the type to serialize (similar to how serde_bytes works). This is important in future to not have to deploy/use a forked bigint dependency, cleaner for prod and makes easier to update dependency in the future. I have all bigint dependencies routing through this crate I created.

Downside to this is having to define a tuple type to annotate the serialization method. Another option is to later add a type to be able to convert to/from which could act as a proxy for serialization similar to how it is done currently when types need to be converted for serialization (I think the usage added in this PR is cleaner though).

Closes 


math/bigint/src/biguint_ser.rs
```diff
@@ -0,0 +1,41 @@
...
+use serde::Serialize;
+
+pub fn serialize<S>(int: &BigUint, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    let mut bz = int.to_bytes_be();
+
+    // Insert positive sign byte at start of encoded bytes if non-zero
+    if bz == vec![0] {
```
ec2: Wonder if its cheaper to check  if `bz` == 1 and `bz[0]` == 0 instead of allocating a new vec and comparing...

ec2: oops, i meant check if the length of bz == 1

austinabell: can just compare the vec with an array at same effect, which is what I just pushed

PR #232, opened 2020-02-18
Global async logging

ec2:
**Summary of changes**
- Replace slog with async-log + pretty_env_logger
- No more need to pass around a logger. There is now a global logger. 

Closes 


forest/src/logger/mod.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+pub(crate) fn setup_logger() {
+    let logger = pretty_env_logger::formatted_timed_builder()
+        .filter(None, log::LevelFilter::Trace)
+        .build();
+    async_log::Logger::wrap(logger, || 0)
+        .start(log::LevelFilter::Trace)
```
austinabell: Info should probably be default?

```suggestion
        .start(log::LevelFilter::Info)
```

forest/src/logger/mod.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+pub(crate) fn setup_logger() {
+    let logger = pretty_env_logger::formatted_timed_builder()
```
austinabell: Also, thoughts on changing the timestamp format
```
2020-02-18T19:35:07.045
```

to something like
```
2020-02-18 19:35:07
```
Nit, but the timestamp is a lot to look at right now lol. Can change in another PR or something, this may just be me

vmx:
I've good experience with https://crates.io/crates/flexi_logger.

There is also https://github.com/filecoin-project/rust-fil-logger which is used for the filecoin proofs stuff that has a flag to create JSON output that is compatible to the one Lotus is using.

PR #236, opened 2020-02-20
Setup Forest execution threads

austinabell:
**Summary of changes**

Currently, the networking service was running in the main thread. This was always planned to run in seperate thread and this needs to change for the changes I'm working on. PRing in now because it can be isolated from the other changes I'm making but have testing running multiple threads and the SIGINT handler functions as intended.

To summarize the SIGINT handler, pressing once will drop all threads (and when services need to be carefully shut down that logic can be added) and a second press will force exit the main thread. With thread/task sleeps you can test this yourself.

Also the `process::exit()` will not call destructors on any thread and will exit the process immediately, docs: https://doc.rust-lang.org/std/process/fn.exit.html and maybe should change the exit code to a non-zero one? Makes sense to me but I'm okay with changing

Closes 


forest/src/main.rs
```diff
@@ -36,11 +39,32 @@ async fn main() {
...
+            println!("Got interrupt, shutting down...");
+        } else {
+            process::exit(0);
+        }
+    })
+    .expect("Error setting Ctrl-C handler");
 
-    task::block_on(async move {
+    // Start libp2p service
+    let lp2p_service = Libp2pService::new(logger, &config.network, net_keypair);
```
GregTheGreek: I would just call it p2p_service

austinabell: unchanged in this PR, but changed it haha

forest/src/main.rs
```diff
@@ -36,11 +39,32 @@ async fn main() {
...
+    let p2p_service = Libp2pService::new(logger, &config.network, net_keypair);
+    let lp2p_thread = task::spawn(async {
+        p2p_service.run().await;
     });
 
+    loop {
+        if running.load(Ordering::SeqCst) > 0 {
+            // TODO change dropping threads to gracefully shutting down services
+            // or implement drop on components with sensitive shutdown
+            drop(lp2p_thread);
```
GregTheGreek: nit: p2p_thread

austinabell:
@ansermino Added a thread sleep on the SIGINT variable check to avoid that busy loop, sorry to make change after review

forest/src/main.rs
```diff
@@ -8,8 +8,35 @@ use async_std::task;
...
+// Blocks current thread until ctrl-c is received
+fn block_until_sigint() {
+    let (ctrlc_send, ctrlc_oneshot) = futures::channel::oneshot::channel();
+    let ctrlc_send_c = RefCell::new(Some(ctrlc_send));
+
+    let running = Arc::new(AtomicUsize::new(0));
+    ctrlc::set_handler(move || {
+        let prev = running.fetch_add(1, Ordering::SeqCst);
+        if prev == 0 {
+            println!("Got interrupt, shutting down...");
```
dutterbutter: nit: "Got interrupt**ed**, shutting down..."

austinabell: I think the noun version makes more sense in this context, anyone else have a preference? Idrc happy to change

ec2: I'm okay with either. They're both clear enough. Maybe "Shutdown signal received... Shutting down" might be better? Again, indifferent though

austinabell: haha to avoid going down this rabbit hole I'm just going to pull in as is, I'll approve any change if anyone cares that much :)

PR #238, opened 2020-02-20
Refactor SyncManager to have ownership over tipsets

austinabell:
**Summary of changes**
- I wrote the syncmanager assuming that it would just use references to the tipsets kept in memory (optimistic thinking, I know) but in fact how it needs to happen is tipsets will be pulled from the network and kept in memory only for the sync manager. Because of this, the tipsets need to be heap allocated using an Rc (https://doc.rust-lang.org/std/rc/struct.Rc.html) and clone references to it to use in the sync manager.

Don't be worried about the clone, it only creates a new pointer to the heap allocated tipset and increments the reference count. Also don't need to use the non-associative `Rc::clone(&other)` in my usages here because the type in enforced by the function usage, but be careful when cloning a tipset into the sync manager not to create a duplicate of it.

Closes 

### DIFF
--- a/blockchain/sync_manager/src/bucket.rs
+++ b/blockchain/sync_manager/src/bucket.rs
@@ -2,26 +2,27 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use blocks::Tipset;
+use std::rc::Rc;
 
 /// SyncBucket defines a bucket of tipsets to sync
 #[derive(Clone, Default)]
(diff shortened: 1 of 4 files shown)


austinabell:
> Should the same approach used here using Rc be applied to the heaviest tipset in the `ChainStore`?

I have changed it on my local branch working changes, but depending on usage that tipset may only need to be owned only by the chain store. (Will know for sure once I finish implementing something)

PR #240, opened 2020-02-23
Adding Verification Function for Aggregate BLS Signatures

DragonMural:
**Summary of changes**
- Added a basic verification function for BLS aggregate signatures 
- Added a simple test 

Issues with pull Request :
-  This function can be made more convenient to use if it takes in the CIDs as a argument instead of the actual data. I was advised by Austin to do this, but I wasn't sure exactly how to do so.


crypto/src/signature.rs
```diff
@@ -134,6 +139,33 @@ pub(crate) fn verify_bls_sig(data: &[u8], pub_k: &[u8], sig: &Signature) -> bool
...
+    // For loop produces the hashed data and gets the public keys from bytes.
+    for x in 0 .. num_sigs{
+        let pk = match BlsPubKey::from_bytes(pub_keys[x]) {
+            Ok(v) => v,
+            Err(_) => return false,
+        };
+        pks.push(pk);
+        let h_data =   bls_hash(data[x]);
+        hashed_data.push(h_data);
+    }
```
austinabell: These two vectors do not need to be handled in the same function, this can be made more easy to read and more efficient if you just map the values to the vector types expected (can explain in person if this is confusing)

crypto/src/signature.rs
```diff
@@ -4,8 +4,10 @@
 use super::errors::Error;
 use address::{Address, Protocol};
 use bls_signatures::{
-    hash as bls_hash, verify, PublicKey as BlsPubKey, Serialize, Signature as BlsSignature,
+    hash as bls_hash, verify,  PublicKey as BlsPubKey, Serialize, Signature as BlsSignature,  paired::bls12_381::G2 as G2
 };
+
+
```
austinabell: nit: remove these extra line breaks

crypto/src/signature.rs
```diff
@@ -109,6 +111,9 @@ pub fn is_valid_signature(data: &[u8], addr: &Address, sig: &Signature) -> bool
     }
 }
 
+
+
+
```
austinabell: same, remove these (but will probably be done automatically from cargo fmt)

crypto/src/signature.rs
```diff
@@ -134,6 +139,33 @@ pub(crate) fn verify_bls_sig(data: &[u8], pub_k: &[u8], sig: &Signature) -> bool
     verify(&sig, &[hashed], &[pk])
 }
 
+pub(crate) fn verify_agg_bls_sig(data: &[&[u8]], pub_keys : &[&[u8]], aggregate_sig: &BlsSignature) -> bool{
```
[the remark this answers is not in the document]
austinabell: Also, change the aggregate signature type to our Signature type (so the BlsSignature type can be opaque to anything outside of this crate)

crypto/src/signature.rs
```diff
@@ -212,4 +244,37 @@ mod tests {
...
+        let rng = &mut XorShiftRng::from_seed([0x3dbe6259, 0x8d313d76, 0x3237db17, 0xe5bc0654]);
+
+        let msg = (0..message_length).map(|_| rng.gen()).collect::<Vec<u8>>();
+        let data: Vec<&[u8]> = (0..num_sigs).map(|x| &msg[x * 64..(x + 1) * 64]).collect();
+
+        let private_keys: Vec<PrivateKey> =
+            (0..num_sigs).map(|_| PrivateKey::generate(rng)).collect();
+        let public_keys: Vec<_> = (0..num_sigs)
+            .map(|x| private_keys[x].public_key().as_bytes())
+            .collect();
```
austinabell:
```
        let public_keys: Vec<_> = private_keys
            .iter()
            .map(|x| x.public_key().as_bytes())
            .collect();
```

to remove need to iterate and index (defeats purpose of map to use x as an index into the vector)

crypto/src/signature.rs
```diff
@@ -212,4 +244,37 @@ mod tests {
...
+
+        let signatures: Vec<BlsSignature> = (0..num_sigs)
+            .map(|x| private_keys[x].sign(data[x]))
+            .collect();
```
austinabell: can do same as above

PR #243, opened 2020-02-27
ChainSync framework

### LINKED ISSUE #223: Connect sync manager with Libp2p RPC service
**Task summary**

- Once the Blocksync RPC service is done with #101 this needs to be connected with the sync manager to be able to send RPC requests to retrieve tipsets to sync. How this should be architected is opinionated, so I won't go into detail in this issue. 

This issue is seperate from handling incoming Blocksync requests, as this will be ignored for now.

The sync manager, however defined, will basically just handle sending the RPC request for Tipsets, then saving to our blockstore.

**Specification reference**

- 

### LINKED ISSUE #227: Implement parsing of blocksync response into more usable types
**Task summary**

Instead of each user having to decode the response manually to see if it's an error or usable, there should be some conversion into a usable result or even `enum` to handle the error status codes and do any necessary logging before returning the data back to the sender.

Currently the response looks something like:
```
pub struct Response {
    pub chain: Vec<TipSetBundle>,
    pub status: u64,
    pub message: String,
}
```
but it would be ideal to maybe return a `Result<Vec<TipSetBundle>, SomeError>` so it can more easily be used. Other alternative would be to seperate the success and error into `enum` variants, but I personally like the former better.

austinabell:
**Summary of changes**
- Sets up chainsync thread and network polling
- Implements #223 by giving the `ChainSyncer` a channel from the networking thread to this thread
- Implements #227 By allowing the conversion from a `BlockSyncResponse` including the tipset bundle and status code into a result with ok type being a full tipset
  - Error type can be improved if needed to handle the network errors differently, but no need for that yet. Error type is currently String but should in future be an enum matching status codes
- Created `NetworkContext` struct to be able to handle RPC requests to the network and manage the request IDs
- Refactored existing types and functions to be more usable/compatible

In this PR I've setup the ChainSyncer so that the underlying database and channels can be replaced from using a live db and p2p network to an in memory db and local channels for testing or replaced by anything else generic in the future. I will wait to write the actual integrated tests until the full structure is done, but I setup a test to make it easier to set that up if someone else does this.

There is some stuff still TODO, but as discussed I would open this PR as interim to hopefully allow others to build ontop of

Closes #223 
Closes #227


blockchain/chain/src/store/chain_store.rs
```diff
@@ -88,28 +91,33 @@ impl<'a> ChainStore<'a> {
         }
         Ok(())
     }
+
     /// Returns genesis blockheader from blockstore
     pub fn genesis(&self) -> Result<BlockHeader, Error> {
-        let bz = self.db.read(self.genesis.key())?;
+        // TODO change data store for pulling genesis
+        let bz = self.db.read("gen_block")?;
```
dutterbutter: If we are not having the genesis cid key within the `ChainStore` struct and are storing the genesis with the key `gen_block` then the function `set_genesis` needs to be updated to store the genesis with that same key.

blockchain/chain_sync/src/sync.rs
```diff
@@ -0,0 +1,493 @@
...
+/// Message data used to ensure valid state transition
+struct MsgMetaData {
+    balance: BigUint,
+    sequence: u64,
+}
+
+impl<'db, DB> ChainSyncer<'db, DB, HamtStateTree>
+where
+    DB: BlockStore,
+    // BS: BlockSyncProvider,
```
dutterbutter: Can this be removed?

blockchain/chain_sync/src/sync.rs
```diff
@@ -0,0 +1,493 @@
...
+    /// Starts syncing process
+    pub async fn sync(&mut self) -> Result<(), Error> {
+        let mut nw = self.network_rx.clone().fuse();
+        loop {
+            select! {
+                network_msg = nw.next().fuse() => match network_msg {
+                    // TODO remove this comment (testing use)
+                    // Some(NetworkEvent::PubsubMessage{source, ..}) => {
+                    //     let req_id = self.network.blocksync_headers("QmUbiXETSptmbSYNtvCWEBYqXDKGwiBCCFWgJn1ebGiFiV".parse().unwrap(), self._genesis.key(), 10).await;
+                    // },
```
dutterbutter: Can this be removed?

blockchain/chain_sync/src/sync.rs
```diff
@@ -0,0 +1,493 @@
...
+    pub fn inform_new_head(&self, from: PeerId, fts: FullTipset) -> Result<(), Error> {
+        // check if full block is nil and if so return error
+        if fts.blocks().is_empty() {
+            return Err(Error::NoBlocks);
+        }
+        // validate message data
+        for block in fts.blocks() {
+            self.validate_msg_data(block)?;
+        }
+        // TODO send pubsub message indicating incoming blocks
```
dutterbutter: I might be mistaken but since `network_rx` was added in this PR could we complete this TODO by informing the network of the incoming block?

blockchain/chain_sync/src/sync.rs
```diff
@@ -0,0 +1,493 @@
...
+        Ok(())
+    }
+
+    async fn sync_fork(&mut self, _head: &Tipset, _to: &Tipset) -> Result<Vec<Arc<Tipset>>, Error> {
+        // TODO sync fork until tipsets are equal or reaches genesis
+        todo!()
+    }
+
+    // Persists headers from tipset slice to chain store
+    fn persist_headers(&self, tipsets: &[Tipset]) -> Result<(), DBError> {
```
dutterbutter: This seems like a duplicate from `persist_headers` that exists in the ChainStore?

blockchain/chain/src/store/chain_store.rs
austinabell: The difference is the genesis block needs to be pulled from storage from a given static string, not by a cid. When starting the node, there is no way to know what the genesis Cid is to pull from db

blockchain/chain_sync/src/sync.rs
austinabell: ya will do, had kept in because I wanted to make the provider generic but probably not needed anymore with just channels passed in to handle requests

blockchain/chain/src/store/chain_store.rs
dutterbutter: Which is fine but `set_genesis` will still need to be updated to reflect the static string as the key no?

blockchain/chain_sync/src/sync.rs
austinabell: it's just mapping over a vector of tipsets and calling that fn

blockchain/chain/src/store/chain_store.rs
austinabell: oh yes, I forgot to add a todo or mention it. We basically need some seperate data store seperate from the chain store (or part of that type) to save genesis as. I just haven't thought about how exactly this will be done

blockchain/chain_sync/src/sync.rs
austinabell: hmm yes good idea, I had assumed it would happen outside of this, but the point of it happening here is because it has to be validated first? Seems redundant to repropogate all blocks in a validated tipset when those would have already been propogated per block

blockchain/chain/src/store/chain_store.rs
austinabell: Updated and added test, let me know if changes work and I'll resolve this

blockchain/chain_sync/src/sync.rs
dutterbutter: Yeah, I see that now.

blockchain/chain_sync/src/manager.rs
```diff
@@ -4,22 +4,22 @@
...
+use std::sync::Arc;
 
 #[derive(Default)]
 pub struct SyncManager {
     sync_queue: SyncBucketSet,
 }
 
 impl SyncManager {
-    pub fn schedule_tipset(&mut self, tipset: Rc<Tipset>) {
+    pub fn schedule_tipset(&mut self, tipset: Arc<Tipset>) {
```
ec2: comments

ec2: same for the 2 other funcs below

blockchain/chain_sync/src/network_context.rs
```diff
@@ -0,0 +1,78 @@
...
+use async_std::sync::Sender;
+use blocks::TipSetKeys;
+use forest_libp2p::{
+    rpc::{BlockSyncRequest, RPCEvent, RPCRequest, RequestId},
+    NetworkMessage,
+};
+use libp2p::core::PeerId;
+use log::trace;
+
+/// Context used in chain sync to handle sequential
```
ec2: sequential what?

austinabell: I had it fixed on my local wip changes lol

blockchain/chain_sync/src/sync.rs
```diff
@@ -0,0 +1,497 @@
...
+        };
+        // store message roots and receive meta_root
+        let meta_root = self.chain_store.blockstore().put(&meta)?;
+
+        Ok(meta_root)
+    }
+    /// Returns FullTipset from store if TipSetKeys exist in key-value store otherwise requests FullTipset
+    /// from block sync
+    pub fn fetch_tipset(&self, _peer_id: PeerId, tsk: &TipSetKeys) -> Result<FullTipset, Error> {
+        let fts = match self.load_fts(tsk) {
```
ec2: cant you just use `?` instead of match? Just a nit dun matter

austinabell: not changed in this PR but the idea is leaving like this until the TODO is done (I commented the same thing on Dustin's PR)

PR #245, opened 2020-02-28
Hello RPC Protocol

ec2:
**Summary of changes**
Changes introduced in this pull request:
- On dial, send Hello message

**Reference issue to close (if applicable)**

Closes 
- #194 

**Other information and links**

PR #246, opened 2020-02-29
Refactor RPC and implement hello protocol

### LINKED ISSUE #194: [TASK]🗃️ Updates from Spec | Hello Protocol
**Task summary**

Last week Filecoin spec team updated the `Hello Protocol` under Nodes/Network. In particular, the protocol string and flow. See the below links for detailed changes.

**Specification reference**

- [Hello Protocol PR](https://github.com/filecoin-project/specs/pull/847/files)
- [Specification Section Updated](https://filecoin-project.github.io/specs/#hello-spec)

austinabell:
**Summary of changes**
- Refactors RPC to allow for more protocols
  - I started doing a bigger refactor, but it was very extensive and didn't think it was worth the time since it works like this. Can revisit though once we update libp2p dependency
- Implements hello protocol
  - Needed to add another protocol to test so this is an alternative PR to #245 
- Updated RPC tests to more efficiently use futures and not waste CPU polling unnecessarily

The hello protocol matches the spec definition but does not match lotus, lotus has only one Hello message type and has optional map fields that they serialize for both the request and response. It was a huge pain to try to retrieve their exact format so I will leave this for now but will probably have to switch it to match lotus to test our syncing functionality.

Closes #194 


node/forest_libp2p/src/rpc/mod.rs
```diff
@@ -51,3 +50,36 @@ impl std::fmt::Display for RPCEvent {
         }
     }
 }
+
```
dutterbutter: For consistency should this be its own file `error.rs`?

node/forest_libp2p/src/rpc/protocol.rs
```diff
@@ -10,6 +12,13 @@ use libp2p::core::{Negotiated, UpgradeInfo};
...
 use std::pin::Pin;
 
+/// RPCResponse payloads for request/response calls
+#[derive(Debug, Clone, PartialEq)]
+pub enum RPCResponse {
+    Blocksync(BlockSyncResponse),
+    Hello(HelloResponse),
+}
+
 /// Protocol upgrade for inbound RPC requests. Currently supports Blocksync.
```
dutterbutter: Comment should be updated to include mention of hello protocol?

node/forest_libp2p/src/rpc/protocol.rs
```diff
@@ -35,34 +44,53 @@ where
...
     fn protocol_info(&self) -> Self::InfoIter {
-        vec![b"/fil/sync/blk/0.0.1"]
+        self.supported_protocols()
+    }
+}
+
+impl RPCRequest {
+    pub fn supported_protocols(&self) -> Vec<&'static [u8]> {
+        match self {
+            // add more protocols when versions/encodings are supported
```
dutterbutter: maybe add a TODO here

node/forest_libp2p/src/service.rs
```diff
@@ -145,15 +152,19 @@ impl Libp2pService {
...
+                                RPCEvent::Request(req_id, RPCRequest::Blocksync(r)) => {
+                                    // TODO implement handling incoming blocksync requests
                                     swarm_stream.get_mut().send_rpc(peer_id, RPCEvent::Response(1, RPCResponse::Blocksync(BlockSyncResponse {
                                         chain: vec![],
                                         status: 203,
                                         message: "handling requests not implemented".to_owned(),
                                     })));
                                 }
+                                RPCEvent::Request(req_id, RPCRequest::Hello(message)) => {
+
```
dutterbutter: nit: remove spacing

dutterbutter:
This should also close #194

austinabell:
> This should also close #194

It has that in the PR description already :)

PR #247, opened 2020-03-02
Sync fork in ChainSyncer

dutterbutter:
**Summary of changes**
- Added the method `sync_fork` within the ChainSyncer
- Adds forks into `return_set`in `sync_headers_reverse`
- Adds an equality check to `tipset` type for comparisons

Closes n/a

### DIFF
--- a/blockchain/blocks/src/tipset.rs
+++ b/blockchain/blocks/src/tipset.rs
@@ -202,24 +202,6 @@ impl Tipset {
     pub fn tip_epoch(&self) -> &ChainEpoch {
         self.blocks[0].epoch()
     }
-    /// Returns true if tipsets are equal
-    pub fn equals(&self, ots: Tipset) -> bool {
-        if self.blocks().is_empty() && ots.blocks().is_empty() {
-            return true;
-        }
-        if self.blocks.is_empty() || ots.blocks().is_empty() {
-            return false;
-        }
-        if self.blocks().len() != ots.blocks().len() {
-            return false;
-        }
-        for (i, b) in (self.blocks()).iter().enumerate() {
-            if b.cid() != ots.blocks()[i].cid() {
-                return false;
-            }
-        }
-        true
-    }
 }
 
 /// FullTipSet is an expanded version of the TipSet that contains all the blocks and messages
--- a/blockchain/blocks/src/tipset.rs
+++ b/blockchain/blocks/src/tipset.rs
@@ -202,6 +202,24 @@ impl Tipset {
     pub fn tip_epoch(&self) -> &ChainEpoch {
         self.blocks[0].epoch()
     }
+    /// Returns true if tipsets are equal
+    pub fn equals(&self, ots: Tipset) -> bool {
+        if self.blocks().is_empty() && ots.blocks().is_empty() {
+            return true;
+        }
+        if self.blocks.is_empty() || ots.blocks().is_empty() {
+            return false;
+        }
+        if self.blocks().len() != ots.blocks().len() {
+            return false;
+        }
+        for (i, b) in (self.blocks()).iter().enumerate() {
+            if b.cid() != ots.blocks()[i].cid() {
+                return false;
+            }
+        }
+        true
+    }
 }
 
 /// FullTipSet is an expanded version of the TipSet that contains all the blocks and messages
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -433,8 +433,9 @@ where
                 // This removes need to sync fork
                 return Ok(return_set);
             }
-            // TODO add fork to return set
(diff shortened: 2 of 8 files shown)


blockchain/blocks/src/tipset.rs
```diff
@@ -202,6 +202,24 @@ impl Tipset {
     pub fn tip_epoch(&self) -> &ChainEpoch {
         self.blocks[0].epoch()
     }
+    /// Returns true if tipsets are equal
+    pub fn equals(&self, ots: Tipset) -> bool {
+        if self.blocks().is_empty() && ots.blocks().is_empty() {
+            return true;
+        }
```
austinabell: A valid tipset should not have empty blocks, so weird to me that this check exists

blockchain/blocks/src/tipset.rs
```diff
@@ -202,6 +202,24 @@ impl Tipset {
...
+        if self.blocks.is_empty() || ots.blocks().is_empty() {
+            return false;
+        }
```
austinabell: This is unnecessary, would be caught by the next block

blockchain/blocks/src/tipset.rs
```diff
@@ -202,6 +202,24 @@ impl Tipset {
...
+        if self.blocks().len() != ots.blocks().len() {
+            return false;
+        }
+        for (i, b) in (self.blocks()).iter().enumerate() {
+            if b.cid() != ots.blocks()[i].cid() {
+                return false;
+            }
+        }
+        true
+    }
```
austinabell: Why do you need this over just deriving PartialEq and/or Eq ? Logic seems to be identical and would enable more easy usage and more compact?

blockchain/chain_sync/src/sync.rs
```diff
@@ -478,13 +461,53 @@ where
...
+                        return Err(Error::Other(
+                            "Chain is linked back to a different genesis (bad genesis)".to_string(),
+                        ));
+                    }
+                }
+                return Err(Error::Other(
+                    "Synced chain forked at genesis, refusing to sync".to_string(),
+                ));
+            }
+            if ts.equals(tips[i].clone()) {
```
austinabell: Shouldn't need to clone to check equality, if you do decide to keep that function, you should check equality with a ref to the other tipset

blockchain/chain_sync/src/sync.rs
```diff
@@ -478,13 +461,53 @@ where
...
+        };
 
-    async fn sync_fork(&mut self, _head: &Tipset, _to: &Tipset) -> Result<Vec<Arc<Tipset>>, Error> {
-        // TODO sync fork until tipsets are equal or reaches genesis
-        todo!()
+        let mut ts = self.chain_store.tipset_from_keys(to.parents())?;
+
+        for i in 0..tips.len() {
+            if *ts.tip_epoch().chain_epoch() == 0 {
+                if let Some(gen) = self.chain_store.genesis()? {
```
austinabell: you have access to the genesis on the ChainSyncer, don't need to query storage for this. Also, if there was none found, this check would just be skipped

blockchain/chain_sync/src/sync.rs
```diff
@@ -478,13 +461,53 @@ where
...
+        let tips: Vec<Tipset> = match self
+            .network
+            .blocksync_headers(peer_id.clone(), head.parents(), FORK_LENGTH_THRESHOLD)
+            .await
+        {
+            Ok(ts) => ts,
+            Err(_e) => {
+                return Err(Error::Other("Failed to retrieve tipset".to_string()));
+            }
+        };
```
austinabell: No need to match on this, can just use `?` and map the error if you need to

blockchain/chain_sync/src/sync.rs
```diff
@@ -478,13 +461,53 @@ where
...
 
-    async fn sync_fork(&mut self, _head: &Tipset, _to: &Tipset) -> Result<Vec<Arc<Tipset>>, Error> {
-        // TODO sync fork until tipsets are equal or reaches genesis
-        todo!()
+        let mut ts = self.chain_store.tipset_from_keys(to.parents())?;
+
+        for i in 0..tips.len() {
+            if *ts.tip_epoch().chain_epoch() == 0 {
+                if let Some(gen) = self.chain_store.genesis()? {
+                    if gen != ts.blocks()[0] {
```
austinabell: Should be checking the genesis tipset against the retrieved one, probably can assume genesis is always one block and that no future tipset would have an equivalent first block, but prob best to future proof

blockchain/blocks/src/tipset.rs
austinabell: Actually it's already being deriven right now, but if you need to change the logic of it (maybe ignoring the keys for whatever reason) can just manually implement it, better to be able to use `==` syntax

PR #248, opened 2020-03-02
Implement sync fork

dutterbutter:
**Summary of changes**
- Added the method sync_fork within the ChainSyncer
- Adds forks into return_set in sync_headers_reverse
- Removed redundant method on tipset 

Closes N/A

### DIFF
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -1,6 +1,8 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+#![allow(unused_variables)]
+
 use super::network_handler::NetworkHandler;
 use super::{Error, SyncManager, SyncNetworkContext};
 use address::Address;
@@ -478,8 +480,8 @@ where
         let mut ts = self.chain_store.tipset_from_keys(to.parents())?;
 
         for i in 0..tips.len() {
-            if *ts.tip_epoch().chain_epoch() == 0 {
-                if self.genesis.blocks()[0] != ts.blocks()[0] {
+            if *ts.epoch().to_u64() == 0 {
+                if self.genesis.blocks() != ts.blocks() {
                     return Err(Error::Other(
                         "Chain is linked back to a different genesis (bad genesis)".to_string(),
                     ));
@@ -491,10 +493,18 @@ where
             if ts == tips[i] {
                 return Ok(tips[0..=i].to_vec());
             }
-            if ts.epoch() > tips[i].epoch() {
+        }
+
+        let mut n = 1;
+
+        while n < tips.len() {
+            if ts.epoch() > tips[n].epoch() {
                 ts = self.chain_store.tipset_from_keys(ts.parents())?;
+            } else {
+                n += 1;
             }
         }
+
         Err(Error::Other(
             "Fork longer than threshold finality of 500".to_string(),
         ))
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -1,8 +1,6 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-#![allow(unused_variables)]
-
 use super::network_handler::NetworkHandler;
 use super::{Error, SyncManager, SyncNetworkContext};
 use address::Address;
@@ -61,7 +59,7 @@ pub struct ChainSyncer<'db, DB, ST> {
     network: SyncNetworkContext,
 
     /// the known genesis tipset
-    genesis: Tipset,
+    _genesis: Tipset,
 
     /// Bad blocks cache, updates based on invalid state transitions.
     /// Will mark any invalid blocks and all childen as bad in this bounded cache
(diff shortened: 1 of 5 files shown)


blockchain/chain_sync/src/sync.rs
```diff
@@ -461,13 +461,46 @@ where
...
+        let peer_id = PeerId::random();
+        // pulled from Lotus: https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L996
+        const FORK_LENGTH_THRESHOLD: u64 = 500;
+
+        // Load blocks from network using blocksync
+        let tips: Vec<Tipset> = self
+            .network
+            .blocksync_headers(peer_id.clone(), head.parents(), FORK_LENGTH_THRESHOLD)
+            .await
+            .map_err(|_| Error::Other("Could not retrieve tipset".to_string()))?;
```
austinabell: the problem I have with this is that if this one request to the one peer fails, this will exit the execution of the node. The way the sync function is called and handled is likely to change, but I don't think any network error should automatically fail the sync

blockchain/chain_sync/src/sync.rs
```diff
@@ -461,13 +461,46 @@ where
...
+
+        let mut ts = self.chain_store.tipset_from_keys(to.parents())?;
+
+        for i in 0..tips.len() {
+            if *ts.tip_epoch().chain_epoch() == 0 {
```
austinabell: the chain_epoch function name is confusing, since it is going from a chain_epoch to a u64, also the `tip_epoch()` could have just been `epoch()` since it's operating on a tipset (nits aside don't need to change in this PR, just thought I'd give thoughts on API)

blockchain/chain_sync/src/sync.rs
```diff
@@ -461,13 +461,46 @@ where
...
+                if self.genesis.blocks()[0] != ts.blocks()[0] {
```
austinabell:
```suggestion
                if self.genesis.blocks() != ts.blocks() {
```
Reason why comparison of only first header?

blockchain/chain_sync/src/sync.rs
```diff
@@ -461,13 +461,46 @@ where
...
+                return Err(Error::Other(
+                    "Synced chain forked at genesis, refusing to sync".to_string(),
+                ));
+            }
+            if ts == tips[i] {
+                return Ok(tips[0..=i].to_vec());
+            }
+            if ts.epoch() > tips[i].epoch() {
+                ts = self.chain_store.tipset_from_keys(ts.parents())?;
+            }
```
austinabell: should this be a loop while? What if the epoch is more than one ahead of current? Is this guaranteed some other way?

blockchain/chain_sync/src/sync.rs
dutterbutter: I figured a genesis will only be one header but the above would work fine.

blockchain/chain_sync/src/sync.rs
dutterbutter: That's a good point. I could move this statement outside the existing loop into a while loop.

```
let mut n = 1;

while n < tips.len() {
    if ts.epoch() > tips[n].epoch() {
        ts = self.chain_store.tipset_from_keys(ts.parents())?; 
    } else {
        n += 1;
    }
}
```

blockchain/chain_sync/src/sync.rs
austinabell: idk it's just safer to make sure also that the tipset you're checking also doesn't have any other headers. Extremely unlikely edge case but the logic seems incorrect, maybe just to me?

dutterbutter: Your above suggestion covers the edge case and its actual purpose. Change made.

node/clock/src/lib.rs
```diff
@@ -106,8 +106,8 @@ impl ChainEpoch {
     pub fn new(timestamp: i64) -> Result<ChainEpoch, TryFromIntError> {
         Ok(ChainEpoch(timestamp.try_into()?))
     }
-    // Returns chain epoch
-    pub fn chain_epoch(&self) -> &u64 {
+    // Returns chain epoch as u64
+    pub fn to_u64(&self) -> &u64 {
```
austinabell: nit:
```suggestion
    pub fn as_u64(&self) -> &u64 {
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -461,13 +463,54 @@ where
...
+                n += 1;
+            }
+        }
 
-    async fn sync_fork(&mut self, _head: &Tipset, _to: &Tipset) -> Result<Vec<Arc<Tipset>>, Error> {
-        // TODO sync fork until tipsets are equal or reaches genesis
-        todo!()
+        Err(Error::Other(
+            "Fork longer than threshold finality of 500".to_string(),
+        ))
```
austinabell: I'm so confused on what this is doing, you are returning Error here regardless, why are you doing this loop? What, at a high level, are you trying to do here?

blockchain/chain_sync/src/sync.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+#![allow(unused_variables)]
```
austinabell: remove please, not doing anything anyway

node/clock/src/lib.rs
dutterbutter: updated

blockchain/chain_sync/src/sync.rs
dutterbutter: updated

blockchain/chain_sync/src/sync.rs
dutterbutter: updated

PR #252, opened 2020-03-03
Implement basic peer manager

### LINKED ISSUE #250: PeerManager

austinabell:
**Summary of changes**
- Implements a peer manager, which is basically just a threadsafe hashmap wrapped in an Arc for concurrent access across the network polling thread and the syncing thread
  - Can definitely be improved to handle the invalid responses or disconnect from the peer entirely under certain circumstances, but I decided to not get too opinionated or in depth in this PR
- Updates config file usage to make it easier and more consolidated to swap out the data directory for the db and network keypair
- Switched basically all references to the db to Arc pointers to allow these structs to be more easily moved across threads and allow concurrent access to the db. The database only requires immutable access because every implementation including in memory uses locks (idk about rocksdb iternally actually) to get mutable access

easiest way to test current usage is running:
```
cargo run -p forest -- -c config.toml
```
in multiple proccesses and changing the config file for each where the data dir config is overriden like so:
```
data_dir = "test_dir"
```
or whatever directory you want the chain data placed.

Closes #250 

### DIFF
--- a/blockchain/chain_sync/src/peer_manager.rs
+++ b/blockchain/chain_sync/src/peer_manager.rs
@@ -1,0 +1,33 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use libp2p::core::PeerId;
+use log::debug;
+use parking_lot::RwLock;
+use std::collections::HashSet;
+
+/// Thread safe peer manager
+#[derive(Default)]
+pub struct PeerManager {
+    /// Hash set of full peers available
+    full_peers: RwLock<HashSet<PeerId>>,
+}
+
+impl PeerManager {
+    /// Adds a PeerId to the set of managed peers
+    pub fn _add_peer(&self, peer_id: PeerId) {
+        debug!("Added PeerId to full peers list: {}", &peer_id);
+        self.full_peers.write().insert(peer_id);
+    }
+
+    /// Returns true if peer set is empty
+    pub fn is_empty(&self) -> bool {
+        self.full_peers.read().is_empty()
+    }
+
+    /// Retrieves a cloned PeerId to be used to send network request
+    pub fn get_peer(&self) -> Option<PeerId> {
+        // TODO replace this with a shuffled or more random sample
+        self.full_peers.read().iter().next().cloned()
+    }
+}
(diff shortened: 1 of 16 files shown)


blockchain/chain_sync/src/peer_manager.rs
```diff
@@ -0,0 +1,44 @@
...
+impl PeerManager {
+    /// Adds a PeerId to the set of managed peers
+    pub fn add_peer(&self, peer_id: PeerId) {
+        debug!("Added PeerId to full peers list: {}", &peer_id);
+        self.full_peers.write().insert(peer_id);
+    }
+
+    /// Returns true if peer set is empty
+    pub fn is_empty(&self) -> bool {
+        self.full_peers.read().is_empty()
```
dignifiedquire: Using sth like parking_lot here will make the executor block, if this is called from an `async` function. I suggest using `async_std::sync::RwLock` instead, which avoids this, as it integrates with the executor.

austinabell: hmm interesting, will fix now. thanks :)

austinabell: https://github.com/ChainSafe/forest/pull/252/commits/43f56ab7a90da01cff787853a2ff9893bf5f0ac1 sufficient? Thanks so much for this review by the way !

dignifiedquire: yes, looks good.

blockchain/chain_sync/src/sync/peer_test.rs
```diff
@@ -0,0 +1,44 @@
...
+            .await;
+
+        // Would be ideal to not have to sleep here and have it deterministic
+        task::sleep(Duration::from_millis(50)).await;
+
+        assert_eq!(peer_manager.len().await, 1);
+        assert_eq!(peer_manager.get_peer().await, Some(source_clone));
+    });
+
+    drop(cs_thread);
```
ec2: Seems unnecessary?

austinabell: yeah it is, was just being cautious and making it more clear what is happening, remove it?

ec2: Well nothing really happens when you drop it. It just "detaches" the task when you drop the handler. I'm ok with or without it. Up to you.

austinabell: hmm yeah for the test and our bin execution the thread is closed but I will look more in depth with interactions later as this could probably be handled better

PR #254, opened 2020-03-06
Read CAR Files

### LINKED ISSUE #222: Implement CAR file utility
**Task summary**

A car file utility will be needed to import the state for genesis for the current testnet. Both Lotus and go-filecoin use the same [utility](https://github.com/ipfs/go-car) which isn't exactly documented. There is [a spec repo](https://github.com/ipfs/archive-format) which is WIP and [is mentioned that the go-car may not satisfy requirements](https://github.com/ipfs/archive-format/issues/13#issuecomment-444995978).

This will be needed because Lotus generates their genesis and then exports the current state in this CAR file format, which is used for their [devnet genesis](https://github.com/filecoin-project/lotus/blob/master/build/genesis/devnet.car). This will require the ability to store the dag state into our state tree to actually use to load the state from the car utility, but this utility can be done before then.

This is not needed for initial syncing, but will be required once we are performing the state transition and updating the state tree

**Specification reference**

- https://github.com/ipfs/archive-format (wip spec repo as referenced before)

ec2:
**Summary of changes**
- Reads car files

Closes #222 


ipld/car/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "forest_car"
+version = "0.1.0"
+authors = ["Eric Tu <tu.eric@hotmail.com>"]
```
dutterbutter: This should be ChainSafe Systems <info@chainsafe.io>

ipld/car/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
dutterbutter: Prob can remove

ipld/car/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    data: Vec<u8>,
+}
+
+/// Loads a CAR buffer into a BlockStore
+pub fn load_car<R: Read, B: BlockStore>(s: &mut B, buf_reader: BufReader<R>) -> Result<(), Error> {
+    let mut car_reader = CarReader::new(buf_reader)?;
+
+    while !car_reader.buf_reader.buffer().is_empty() {
+        let block = car_reader.next_block()?;
+        let check_cid = Cid::new_from_prefix(&block.cid.prefix(), &block.data).unwrap();
```
dutterbutter: Could use `?` instead of `unwrap()`

ipld/car/Cargo.toml
ec2: Whoops

ipld/car/Cargo.toml
ec2: Done

ipld/car/src/lib.rs
ec2: It shouldnt fail ever. I'm getting a prefix from the Cid struct
@austinabell

austinabell: Only fails if the multihash generated in the Cid is not a built-in one. Probably not a case we will ever need, but might as well handle the error to future proof and not panic if that's the case

ec2: Hmmmmm i guess that makes sense... I dont know why, but i just love `.unwrap()`s

ec2: Done

ipld/car/src/util.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let cid: Cid = Cid::new(
+        cid::Codec::from(codec)?,
+        cid::Version::from(version)?,
+        cid::multihash::Multihash::from_bytes(
+            multihash_with_data[0..=(len as usize + hashcode_len_diff)].to_vec(),
+        )?,
+    );
+    let len = cid.to_bytes().len() as u64;
+    Ok((cid, len))
```
austinabell: It doesn't make sense to me why you convert the reader into a slice, then to use it here as a reader. This seems like logic that should be attached to the Cid crate. Thoughts on making this take in a reader and attaching it to Cid crate? You wouldn't need that ld_read function or all this indexing that could panic the program on a bad input

ipld/car/src/error.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        use self::Error::*;
+
+        match self {
+            ParsingError(_) => "Failed to parse CAR file",
+            Error::InvalidFile(_) => "Invalid CAR file",
+            Other(_) => "Other Cid Error",
```
austinabell: why the use if you are going to specify the enum prefix? Also, this is consuming the error messages, what is the point of this error type? If it doesn't matter then just remove this type and make the error type `&'static str`

ipld/car/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// Creates a new CarReader and parses the CarHeader
+    pub fn new(mut buf_reader: BufReader<R>) -> Result<Self, Error> {
+        let (_len, buf) = ld_read(&mut buf_reader)?;
+        let header: CarHeader = from_slice(&buf).map_err(|e| Error::ParsingError(e.to_string()))?;
+        if header.roots.is_empty() {
+            return Err(Error::ParsingError("empty CAR file".to_owned()));
+        }
+        if header.version != 1 {
+            return Err(Error::InvalidFile("CAR file version must be 1".to_owned()));
+        }
```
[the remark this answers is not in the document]
ec2: Should be in the reader constructor, I'd say. This allows us to parse and verify the header to fail early if its not a valid header

ipld/car/src/util.rs
```diff
@@ -0,0 +1,47 @@
...
+    let mut buf = Vec::with_capacity(l as usize);
+    buf_reader
+        .take(l)
+        .read_to_end(&mut buf)
+        .map_err(|e| Error::Other(e.to_string()))?;
+    Ok((l, buf))
+}
+
+pub(crate) fn read_node<R: Read>(buf_reader: &mut R) -> Result<(Cid, Vec<u8>), Error> {
+    let (_len, buf) = ld_read(buf_reader)?;
```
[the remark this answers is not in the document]
ec2: actually ld_read doesnt need to return len. So i will remove

ipld/car/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// Loads a CAR buffer into a BlockStore
+pub fn load_car<R: Read, B: BlockStore>(s: &mut B, buf_reader: BufReader<R>) -> Result<(), Error> {
+    let mut car_reader = CarReader::new(buf_reader)?;
+
+    while !car_reader.buf_reader.buffer().is_empty() {
+        let block = car_reader.next_block()?;
+        let check_cid = Cid::new_from_prefix(&block.cid.prefix(), &block.data)
+            .map_err(|e| Error::Other(e.to_string()))?;
+        assert_eq!(&check_cid, &block.cid);
```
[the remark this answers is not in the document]
ec2: fixed. forgot to remove

ipld/car/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/// IPLD Block
+pub struct Block {
+    cid: Cid,
+    data: Vec<u8>,
+}
+
+/// Loads a CAR buffer into a BlockStore
+pub fn load_car<R: Read, B: BlockStore>(s: &mut B, buf_reader: BufReader<R>) -> Result<(), Error> {
```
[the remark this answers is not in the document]
ec2: True. I just assumed you need it as mut becuase we're writing, but that is in fact not the case. fixed

ipld/car/src/util.rs
ec2: Added a todo for read_Cid using a bufreader

ipld/car/src/error.rs
ec2: True

ipld/car/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// Reads CAR files that are in a BufReader
+pub struct CarReader<R> {
+    pub buf_reader: BufReader<R>,
+    pub header: CarHeader,
+}
+
+impl<R> CarReader<R>
+where
+    R: std::io::Read,
```
austinabell: nit: move this to the constructor, no need to constrain the whole impl (future proof)

ipld/car/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+use cid::Cid;
+use forest_encoding::from_slice;
+use serde::{Deserialize, Serialize};
+use std::io::{BufReader, Read};
+
+mod error;
+mod util;
+use crate::util::read_node;
+use error::*;
+use util::ld_read;
```
austinabell: nit: can you move the mods to the top and group all uses together?

ipld/car/src/error.rs
```diff
@@ -0,0 +1,34 @@
...
+    InvalidFile(String),
+    Other(String),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Error::ParsingError(err) => write!(f, "Failed to parse CAR file: {}", err.clone()),
+            Error::InvalidFile(err) => write!(f, "Invalid CAR file: {}", err.clone()),
+            Error::Other(err) => write!(f, "Other cid Error: {}", err.clone()),
```
austinabell: nit, this error message was probably copied from the Cid error ?
```suggestion
            Error::Other(err) => write!(f, "CAR Error: {}", err.clone()),
```
or something like that, idk these error variants don't make much sense to me as they are anyway

PR #256, opened 2020-03-10
Update runtime to new spec/ impl

### LINKED ISSUE #180: Update Runtime Interface based on spec changes
Some new methods were added and some were changed.

See spec and also refer to this commit: https://github.com/filecoin-project/specs/commit/c666c4deadfc462acbe89eb6fd10e424986a42f0

austinabell:
**Summary of changes**
- Update runtime interface
  - Has changed a lot since I touched last lol. Hard to guess the exact usage or when the runtime will need to be mutable, and I will think a bit more into the exact usage but this will probably just act as a template to be updated when needed

The actors spec/ shared module is A LOT better than the spec was before, and although it might be painful to switch over now, the work can be very contained and easily worked on in parallel with coordination

Please give opinions on the API, I will think through more but this was just a quick pass so we can have something to quickly work off of

Closes #180 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
     "vm/runtime",
     "vm/state_tree",
     "vm/interpreter",
+    "vm/abi",
     "node",
     "node/clock",
     "crypto",
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -2,11 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 mod errors;
+mod randomness;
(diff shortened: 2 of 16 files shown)


ec2:
LGTM! Looks like you missed some of the methods in the runtime though such as `ValidateImmediateCallerAcceptAnyOfTypes`

austinabell:
> LGTM! Looks like you missed some of the methods in the runtime though such as `ValidateImmediateCallerAcceptAnyOfTypes`

Chatted in person but to answer I am basing off of the specs-actors, which this doesn't exist anymore

Issue #257, opened 2020-03-11
Update VM exit codes to match new spec


austinabell:
**Task summary**

Exit codes have changed, should be easy to update, I will likely do very soon so ping me if you want to do

**Specification reference**

- https://github.com/filecoin-project/specs-actors/tree/master/actors/runtime/exitcode

**Other information and links**

ec2:
I can do it

austinabell:
cool, make sure that you have the correct integer value mapping, I assume it only need to be from `ExitCode` -> `u64` or whatever but look into this to see what exactly is needed

PR #270, opened 2020-03-11
Update exit codes

ec2:
**Summary of changes**
- Update exit codes to specs-actor latest version

Closes 
#257 

### DIFF
--- a/vm/src/exit_code.rs
+++ b/vm/src/exit_code.rs
@@ -2,93 +2,91 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 /// ExitCode defines the exit code from the VM execution
-#[derive(PartialEq, Eq, Debug, Clone)]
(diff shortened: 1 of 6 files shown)


vm/src/exit_code.rs
```diff
@@ -2,93 +2,91 @@
...
-    /// RuntimeAssertFailure is returned when an actor method invocation calls
-    /// rt.Assert with a false condition.
-    RuntimeAssertFailure,
-
-    /// MethodSubcallError is returned when an actor method's Send call has
-    /// returned with a failure error code (and the Send call did not specify
-    /// to ignore errors).
-    MethodSubcallError,
-}
+    // Indicates failure to find an actor in the state tree.
```
dutterbutter: These should `///` rather than `//`

ec2: good catch, thanks

PR #273, opened 2020-03-12
State tree full implementation and refactor IPLD

### LINKED ISSUE #189: Include hamt in state tree implementation
Currently the data structure for state tree is setup here: https://github.com/ChainSafe/ferret/blob/master/vm/state_tree/src/lib.rs but only includes the cache part of the store. This issue is to include functionality needed with hamt once #187 is implemented and include the hamt in the HamtStateTree struct.

This will allow for persistent storage and cid generation for the tree

austinabell:
**Summary of changes**
- Includes hamt in state tree
- Changed State tree interface
  - Spec is still very incomplete so used Lotus as ref
  - I used a different snapshotting method since the vector and indexing way they have seems bad
- Refactor HAMT
  - So since the HAMT cannot be constrained by one value type for our usage, instead the generic value type is replaced with our IPLD object and the get function for the hamt is now generic where the type implements deserialize and gets converted at that level so the interface is the same
- Refactor IPLD and implement serialize and conversion functions
  - Because of previous where the type cannot be constrained, there needed to be a way to go from a struct to IPLD and from IPLD to a struct where the IPLD can be serialized and stored in the underlying store.
  - IPLD -> struct is done inefficiently by going through a serialized buffer, but it's fine for now

Closes #189 


ipld/hamt/src/hamt.rs
```diff
@@ -17,25 +18,24 @@ use std::borrow::Borrow;
...
+/// assert_eq!(map.delete(&1).unwrap(), true);
+/// assert_eq!(map.get::<_, String>(&1).unwrap(), None);
 /// let cid = map.flush().unwrap();
 /// ```
 #[derive(Debug)]
-pub struct Hamt<'a, K, V, S> {
-    root: Node<K, V>,
-    store: &'a S,
+pub struct Hamt<'a, K, BS> {
+    root: Node<K>,
```
ec2: Nice haha

vm/actor/src/singletons.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use address::Address;
+use vm::ActorID;
+
+lazy_static! {
```
ec2: this is cool. first time seeing this macro

Issue #258, opened 2020-03-11
Update database interface


austinabell:
**Issue summary**

- Functions should be async now working with new futures and the parameters really don't need to be moving the values, it's unnecessary
- I'm also going to just make it one trait if no one objects, annoying to have Read + Write (also conflicting with std::io::{Read, Write})

**Other information and links**

austinabell:
Started to port over, huge headache switching to async because of all the recursive async calls with the AMT/HAMT. Wip here: https://github.com/ChainSafe/forest/commit/7930afa7ae52e13c6379e2cb834534f22a85e0db

Async part is not extremely useful yet since RocksDB isn't async yet

PR #274, opened 2020-03-13
Remove unnecessary db trait

austinabell:
**Summary of changes**
- 1/2 of #258 wanted to make this change so the more painful switch to async can be done easier and wanted to fix this now
- Also cleaned up Error usage relating to the db Errors because it was a mess

Closes 

### DIFF
--- a/node/db/src/memory.rs
+++ b/node/db/src/memory.rs
@@ -1,8 +1,7 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use super::Error;
-use super::{DatabaseService, Read, Write};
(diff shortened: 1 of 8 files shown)


node/db/src/memory.rs
```diff
@@ -42,7 +41,7 @@ impl Default for MemoryDB {
 
 impl DatabaseService for MemoryDB {}
 
-impl Write for MemoryDB {
+impl Store for MemoryDB {
```
ec2: Whats the reason for collapsing Write and Read into Store?

austinabell: Annoying to reference both, no benefit for them being separated, conflicting trait name with io read and write

PR #282, opened 2020-03-16
Init actor implementation

### LINKED ISSUE #263: Implement updated init Actor
**Issue summary**

https://github.com/filecoin-project/specs-actors/tree/master/actors/builtin/init

austinabell:
**Summary of changes**
- Closes #263 
- Made some slight changes to the runtime
- State testing through state tree test


vm/actor/src/builtin/init/state.rs
```diff
@@ -34,21 +36,38 @@ impl State {
...
+
+        let map: Hamt<String, _> =
+            Hamt::load_with_bit_width(&self.address_map, store, HAMT_BIT_WIDTH)?;
+
+        let actor_id: ActorID = match map.get(&addr.hash_key())? {
+            Some(id) => id,
+            None => return Err(HamtError::Custom("address not found")),
+        };
+
+        Ok(Address::new_id(actor_id.0).unwrap())
```
dutterbutter: Any particular reason for the `unwrap()` usage here?

vm/actor/src/builtin/init/state.rs
```diff
@@ -34,21 +36,38 @@ impl State {
...
+        if addr.protocol() == Protocol::ID {
+            return Ok(addr.clone());
+        }
+
+        let map: Hamt<String, _> =
+            Hamt::load_with_bit_width(&self.address_map, store, HAMT_BIT_WIDTH)?;
+
+        let actor_id: ActorID = match map.get(&addr.hash_key())? {
+            Some(id) => id,
+            None => return Err(HamtError::Custom("address not found")),
```
dutterbutter: Small nit: are we capitalizing error messages? Whichever it is I think we should keep consistent.

vm/actor/src/builtin/init/state.rs
austinabell: because that can't realistically fail and I didn't want to change the error type, but I'll change

PR #285, opened 2020-03-18
Implement balance table

### LINKED ISSUE #278: Implement balance table
**Task summary**

Wrapper around a hamt with the value types being `TokenAmount` and some other basic logic. Used in market actor #264 and power actor #161 

Definitely optional, but may make life easier to just port over because of lack of spec tests and detailed tests (just assuming from scanning over the two actors that use this)

**Specification reference**

- https://github.com/filecoin-project/specs-actors/blob/master/actors/util/adt/balancetable.go

austinabell:
**Summary of changes**
- Implements balance table, which is needed for market and power actors to more easily match spec
- Adds a `for_each` function on the hamt, which allows passing a function which reads kvs and performs the function on all of them while mutating state outside of the function.
- Update TokenAmount traits for more clean usage 

Closes #278 


ipld/hamt/src/hamt.rs
```diff
@@ -201,4 +201,35 @@ where
...
+    ///
+    /// let mut total = 0;
+    /// map.for_each(&mut |_, v: u64| {
+    ///    total += v;
+    ///    Ok(())
+    /// }).unwrap();
+    /// assert_eq!(total, 3);
+    /// ```
+    #[inline]
+    pub fn for_each<F, V>(&self, f: &mut F) -> Result<(), String>
```
ec2: couldn't impl the Iterator trait instead?

ipld/hamt/src/node.rs
```diff
@@ -129,6 +128,29 @@ where
         self.pointers.is_empty()
     }
 
+    pub(crate) fn for_each<V, S, F>(&self, store: &S, f: &mut F) -> Result<(), String>
```
ec2: couldn't impl the Iterator trait instead?

ipld/hamt/src/hamt.rs
austinabell: no because all of the elements are not in memory. Also this would be done on top of this, the reason for this is to be able to more closely match the spec.

PR #293, opened 2020-03-20
Initial chainsync process

### LINKED ISSUE #249: Implement Basic Sync Manager
**Task summary**

- Build out the necessary methods to handle incoming tipsets, queues, targets, update sync states and determine sync results

**Specification reference**

- Somewhat follows [chain sync procedure](https://filecoin-project.github.io/specs/#systems__filecoin_blockchain__chainsync)

- [Lotus sync manager reference](https://github.com/filecoin-project/lotus/blob/master/chain/sync_manager.go)

dutterbutter:
**Summary of changes**
- Added sync entry from main
- Updating sync states
- Initial logic for follow up sync state in place
- Fetch full tipset for initial checkpoint and begins syncing headers reverse from there
- Sync messages from blocksync functionality added
- Moved common chain structure construction for tests into its own module
- Added unit tests to sync.rs

Closes #249 #298 


blockchain/chain_sync/src/bucket.rs
```diff
@@ -57,9 +60,18 @@ impl SyncBucketSet {
         }
         self.buckets.push(SyncBucket::new(vec![tipset]))
     }
-    pub(crate) fn heaviest(&self) -> Option<Arc<Tipset>> {
+    /// Removes the SyncBucket with heaviest weighted Tipset from SyncBucketSet
+    pub(crate) fn _pop(&mut self) -> Option<SyncBucket> {
```
austinabell:
```suggestion
    pub(crate) fn pop(&mut self) -> Option<SyncBucket> {
```

blockchain/chain_sync/src/bucket.rs
```diff
@@ -68,6 +80,36 @@ impl SyncBucketSet {
         // Return the heaviest tipset bucket
         vals.iter().max_by_key(|b| b.weight()).cloned()
     }
+    /// Updates SyncBucketSet by removing specified SyncBucket
+    pub(crate) fn _remove(&mut self, ts_bucket: &SyncBucket) {
```
austinabell:
```suggestion
    pub(crate) fn remove(&mut self, ts_bucket: &SyncBucket) {
```

(same with below, I won't comment on each)

blockchain/chain_sync/src/bucket.rs
```diff
@@ -68,6 +80,36 @@ impl SyncBucketSet {
...
+    /// Removes SyncBucket specified by provided Tipset
+    pub(crate) fn _pop_related(&mut self, ts: Tipset) {
+        for b in self.buckets() {
+            if b.clone().same_chain_as(&ts.clone()) {
+                self.clone()._remove(b)
+            }
+        }
+    }
+    /// Returns a vector of SyncBuckets
+    pub(crate) fn buckets(&self) -> &Vec<SyncBucket> {
```
austinabell:
```suggestion
    pub(crate) fn buckets(&self) -> &[SyncBucket] {
```

blockchain/chain_sync/src/network_context.rs
```diff
@@ -70,6 +70,27 @@ impl SyncNetworkContext {
...
+        tsk: &TipSetKeys,
+        count: u64,
+    ) -> Result<FullTipset, String> {
+        let bs_res = self
+            .blocksync_request(
+                peer_id,
+                BlockSyncRequest {
+                    start: tsk.cids().to_vec(),
+                    request_len: count,
+                    options: 1,
```
austinabell: This doesn't request messages as is, so I don't see the point of this. Also would be preferable to have just one function with this logic and have the headers and fts functions just call with the options (remove duplicate logic)
```suggestion
                    options: 3,
```

blockchain/chain_sync/src/peer_manager.rs
```diff
@@ -2,15 +2,19 @@
...
+use std::collections::HashMap;
 use std::collections::HashSet;
 
 /// Thread safe peer manager
 #[derive(Default)]
 pub struct PeerManager {
     /// Hash set of full peers available
     full_peers: RwLock<HashSet<PeerId>>,
+    /// Represents peers and proposed tipsets from the network
+    pub peer_heads: RwLock<HashMap<PeerId, Tipset>>,
```
austinabell: why would you need both? peer_heads seems to be all you need (seems messy to handle update each seperately)

blockchain/chain_sync/src/sync.rs
```diff
@@ -134,12 +140,39 @@ impl<DB> ChainSyncer<DB>
...
+                    .fetch_tipset(
+                        source.clone(),
+                        &TipSetKeys::new(message.heaviest_tip_set.clone()),
+                        1,
+                    )
+                    .await
+                {
+                    if self.inform_new_head(&source.clone(), &fts).await.is_err() {
+                        warn!(
+                            "failed to sync with provided tipset cid: {:?}",
```
austinabell: just to note, this will debug print an array of all block cids for the tipset (what is the benefit for seeing this?)

blockchain/chain_sync/src/sync.rs
```diff
@@ -134,12 +140,39 @@ impl<DB> ChainSyncer<DB>
...
+                            message.heaviest_tip_set.clone()
```
austinabell: Should never need to clone to log
```suggestion
                            &message.heaviest_tip_set
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -134,12 +140,39 @@ impl<DB> ChainSyncer<DB>
...
+                        );
+                    };
+                } else {
+                    warn!(
+                        "failed to fetch full tipset from peer: {}, cid: {:?} from storage or network",
+                        source,
+                        message.heaviest_tip_set.clone()
```
austinabell:
```suggestion
                        &message.heaviest_tip_set
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -188,15 +230,82 @@ where
...
+            .await;
+
+        // initial sync
+        if self.get_state() == &SyncState::Init {
+            if let Some(best_target) = self.select_sync_target().await {
+                self.sync(best_target).await?;
+                return Ok(());
+            }
+        }
+        self.schedule_tipset(ts.clone()).await?;
```
austinabell: Not used after, can just move to remove unnecessary clone
```suggestion
        self.schedule_tipset(ts).await?;
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -188,15 +230,82 @@ where
...
+        Ok(())
+    }
+    /// Retrieves the heaviest tipset in the sync queue; considered best target head
+    pub async fn select_sync_target(&mut self) -> Option<Tipset> {
+        let heads = Vec::new();
+        for (_, ts) in self.peer_manager.peer_heads.read().await.iter() {
+            heads.clone().push(ts);
```
austinabell: This does nothing as is? it's adding tipsets to the cloned vector?
```suggestion
            heads.push(ts);
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -188,15 +230,82 @@ where
...
+                self.sync(best_target).await?;
+                return Ok(());
+            }
+        }
+        self.schedule_tipset(ts.clone()).await?;
+        Ok(())
+    }
+    /// Retrieves the heaviest tipset in the sync queue; considered best target head
+    pub async fn select_sync_target(&mut self) -> Option<Tipset> {
+        let heads = Vec::new();
```
austinabell: you can tell it's not doing anything because heads is not mutable

blockchain/chain_sync/src/sync.rs
```diff
@@ -188,15 +230,82 @@ where
...
+        for (_, ts) in self.peer_manager.peer_heads.read().await.iter() {
+            heads.clone().push(ts);
+        }
+        // sort tipsets by epoch
+        heads.clone().sort_by_key(|header| (*header.epoch()));
```
austinabell: continuation
```suggestion
        heads.sort_by_key(|header| (*header.epoch()));
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -188,15 +230,82 @@ where
...
+        self.sync_queue.heaviest()
+    }
+    /// Schedules a new tipset to be handled by the sync manager
+    async fn schedule_tipset(&mut self, tipset: Tipset) -> Result<(), Error> {
+        info!("scheduling incoming tipset to sync: {:?}", tipset.cids());
+
+        // check sync status if indicates tipsets are ready to be synced
+        if self.get_state() == &SyncState::Catchup {
+            // send tipsets to be synced
+            self.sync(tipset.clone()).await?;
```
austinabell:
```suggestion
            self.sync(tipset).await?;
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -230,12 +339,17 @@ where
     }
     /// Returns FullTipset from store if TipSetKeys exist in key-value store otherwise requests FullTipset
     /// from block sync
-    pub fn fetch_tipset(&self, _peer_id: PeerId, tsk: &TipSetKeys) -> Result<FullTipset, Error> {
+    async fn fetch_tipset(
+        &mut self,
+        peer_id: PeerId,
+        tsk: &TipSetKeys,
+        count: u64,
```
austinabell: isn't this always going to be 1? you are checking if a singular tipset exists in storage then requesting if not

blockchain/chain_sync/src/sync.rs
```diff
@@ -536,12 +656,25 @@ where
         ))
     }
 
+    /// iterate over headers to make full tipsets and check state accordingly
+    async fn _sync_msgs_check_state(&self, _headers: Vec<BlockHeader>) {
+        todo!()
+    }
```
austinabell: shouldn't this take in a Tipset instead of a vector of headers? Is this going to return a full tipset? Shouldn't this be a slice passed in? Why does it need to be a moved vector?

blockchain/chain_sync/src/peer_manager.rs
dutterbutter: Can you elaborate on what you mean by both? Are referencing having both peerId and tipsets?

blockchain/chain_sync/src/sync.rs
dutterbutter: Will refactor

blockchain/chain_sync/src/peer_manager.rs
austinabell: Nah I just mean you should only need the variable you created, the hash set isn't necessary (two copies of the peerids that have to be managed)

dutterbutter: I think I am going to remove this altogether from peer_manager, I only actually need the tipsets I believe

dutterbutter:
> Interop testing status? Hard to check specific logic

I was not able to get the lotus node working myself, I have raised the question in the fil-help slack channel. I will follow up and update the status once i hear back.

Makefile
```diff
@@ -68,4 +68,7 @@ license:
 docs:
 	cargo doc --no-deps --all-features
 
+run:
+	cargo run -p forest -- --config config.toml
```
austinabell: was this just meant as a temp for testing? Seems weird to use a config file that doesn't exist by default

blockchain/blocks/src/tipset.rs
```diff
@@ -64,9 +65,21 @@ impl<'de> de::Deserialize<'de> for TipSetKeys {
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.cids.partial_cmp(&other.cids)
+    }
+}
+
+impl Ord for TipSetKeys {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.cids.cmp(&other.cids)
+    }
+}
```
austinabell: This is just the default Ord/ Partial Ord impl? Why explicit?

blockchain/chain_sync/src/bucket.rs
```diff
@@ -68,6 +80,36 @@ impl SyncBucketSet {
...
+                self.clone().remove(b)
+            }
+        }
+    }
+    /// Returns a vector of SyncBuckets
+    pub(crate) fn buckets(&self) -> &[SyncBucket] {
+        &self.buckets
+    }
+    /// Returns true if SyncBucket is empty
+    pub(crate) fn _is_empty(&self) -> bool {
```
austinabell:
```suggestion
    pub(crate) fn is_empty(&self) -> bool {
```

blockchain/chain_sync/src/bucket.rs
```diff
@@ -68,6 +80,36 @@ impl SyncBucketSet {
         // Return the heaviest tipset bucket
         vals.iter().max_by_key(|b| b.weight()).cloned()
     }
+    /// Updates SyncBucketSet by removing specified SyncBucket
+    pub(crate) fn remove(&mut self, ts_bucket: &SyncBucket) {
+        let vals: Vec<SyncBucket> = self
+            .buckets
+            .clone()
```
austinabell:
```suggestion
```

blockchain/chain_sync/src/bucket.rs
```diff
@@ -57,9 +60,18 @@ impl SyncBucketSet {
         }
         self.buckets.push(SyncBucket::new(vec![tipset]))
     }
-    pub(crate) fn heaviest(&self) -> Option<Arc<Tipset>> {
+    /// Removes the SyncBucket with heaviest weighted Tipset from SyncBucketSet
+    pub(crate) fn pop(&mut self) -> Option<SyncBucket> {
+        if let Some(heaviest_bucket) = self.buckets().iter().max_by_key(|b| b.heaviest_tipset()) {
+            self.clone().remove(heaviest_bucket);
```
austinabell:
```suggestion
            self.remove(heaviest_bucket);
```

blockchain/chain_sync/src/network_context.rs
```diff
@@ -70,6 +70,26 @@ impl SyncNetworkContext {
...
+        peer_id: PeerId,
+        tsk: &TipSetKeys,
+    ) -> Result<FullTipset, String> {
+        let bs_res = self
+            .blocksync_request(
+                peer_id,
+                BlockSyncRequest {
+                    start: tsk.cids().to_vec(),
+                    request_len: 1,
+                    options: 3,
```
austinabell: not needed to change now, but for this and the one above, options should be constants and options set based on those, example:
```
BLOCKS = 1
MESSAGES = 2
..
options: BLOCKS | MESSAGES
```
something like that for readability/ future proofing, can do in this PR if you want, or I can make the change later

blockchain/chain_sync/src/network_context.rs
```diff
@@ -70,6 +70,26 @@ impl SyncNetworkContext {
...
+                },
+            )
+            .await?;
+
+        let fts = bs_res.into_result()?;
+        Ok(fts[0].clone())
```
austinabell:
```suggestion
        Ok(fts[0])
```

node/forest_libp2p/src/config.rs
```diff
@@ -18,8 +18,8 @@ impl Default for Libp2pConfig {
         Libp2pConfig {
             listening_multiaddr: "/ip4/0.0.0.0/tcp/0".to_owned(),
             pubsub_topics: vec![
-                Topic::new("/fil/blocks".to_owned()),
-                Topic::new("/fil/messages".to_owned()),
+                Topic::new("/fil/blocks/interop".to_owned()),
+                Topic::new("/fil/msgs/interop".to_owned()),
```
austinabell: Should maybe add a TODO since this will for sure not be the final name, and is just being used for testing

Makefile
dutterbutter: Yeah I will remove

blockchain/chain_sync/src/bucket.rs
```diff
@@ -2,29 +2,28 @@
...
+    tips: Vec<Tipset>,
 }
 
 impl SyncBucket {
     /// Constructor for tipset bucket
-    fn new(tips: Vec<Arc<Tipset>>) -> SyncBucket {
+    fn new(tips: Vec<Tipset>) -> SyncBucket {
         Self { tips }
     }
     /// heaviest_tipset returns the tipset with the max weight
```
ec2: no need to mention function name here.

blockchain/chain_sync/src/bucket.rs
dutterbutter: Need to clone, the compiler complains otherwise.

blockchain/chain_sync/src/bucket.rs
dutterbutter: Need to clone, the compiler complains otherwise.

blockchain/chain_sync/src/network_context.rs
dutterbutter: Need to clone, the compiler complains otherwise.

blockchain/chain_sync/src/bucket.rs
austinabell: yeah didn't see all you were doing, replace the function with this instead:
```
        self.buckets.retain(|b| b != ts_bucket);
```
instead of cloning and full tipset and filtering

blockchain/chain_sync/src/bucket.rs
austinabell: That's because the logic of the function is flawed, you need to be mutating self to remove the actual value or this will just create a copy of the removed item. Also the `Some()` return doesn't need to clone if you keep similar logic

blockchain/chain_sync/src/network_context.rs
austinabell: you're right, for some reason thought it could be moved. One point about this, can you change it to a checked indexing (`get(0)`) or make sure the length is 1 before this so that a bad response won't crash the node?

blockchain/chain_sync/src/bucket.rs
ec2: you can do this instead: 
```
        if let Some(heaviest_bucket) = self.buckets().iter().max_by_key(|b| b.heaviest_tipset()).clone() {
            let heaviest_bucket = heaviest_bucket.clone();
            self.remove(&heaviest_bucket);
            Some(heaviest_bucket)
        } else {
            None
        }
```

austinabell: That still requires a clone of a bucket of tipsets, better to not rely on this

ec2: oh youre right... i didnt realize it was a whole bunch of tipsets

austinabell: One option, as I mentioned:
```
        if let Some((i, _)) = self
            .buckets()
            .iter()
            .enumerate()
            .max_by_key(|(_, b)| b.heaviest_tipset())
        {
            let ts = self.buckets.remove(i);
            Some(ts)
        } else {
            None
        }
```
you are already doing all of the checks for finding the max, seems redundant to have the remove function anyway. This would obviously have to be tested, just wanted to put my thought into something tangible. 

I don't think the max_by_key is terribly efficient, especially since the logic that was changed in this PR has heaviest_tipset cloning that heaviest tipset, (sub thought is that you don't clone the tipset and order based on that as a key as that may not even order correctly since it compares the tipset not the weight)

blockchain/chain_sync/src/sync.rs
```diff
@@ -134,12 +140,38 @@ impl<DB> ChainSyncer<DB>
 where
     DB: BlockStore,
 {
-    /// Starts syncing process
-    pub async fn sync(mut self) -> Result<(), Error> {
+    pub async fn start(mut self) -> Result<(), Error> {
         self.net_handler.spawn(Arc::clone(&self.peer_manager));
```
ec2: I guess i didn't notice this last time, but why didnt you just do `self.peer_manager.clone()` seems to work fine when I try it...

blockchain/chain_sync/src/sync.rs
```diff
@@ -74,7 +81,7 @@ pub struct ChainSyncer<DB> {
     net_handler: NetworkHandler,
 
     /// Peer manager to handle full peers to send ChainSync requests to
-    peer_manager: Arc<PeerManager>,
+    pub peer_manager: Arc<PeerManager>,
```
ec2: Doesn't look like it needs to be public

blockchain/chain_sync/src/sync.rs
austinabell: In this case it's the same, because the type is inferred, it's just recommended with Arc and Rc to explicitly clone like that so that you don't accidentally clone the underlying type and create a duplicate. Can switch if you'd like, but safer in future to keep imo

ec2: That's cool. I didn't know this was convention. It's funny I think i saw someone proposing that they change the name of the clone method on Arc Rc the other day. Makes sense

blockchain/chain_sync/src/sync.rs
```diff
@@ -153,33 +186,157 @@ where
...
+    ) -> Result<FullTipset, Error> {
+        // see https://github.com/filecoin-project/lotus/blob/master/build/params_shared.go#L109 for block message limit
+        const BLOCK_MESSAGE_LIMIT: usize = 512;
+        let mut blocks: Vec<Block> = Vec::with_capacity(ts.blocks().len());
+
+        if ts.blocks().len() != secp_msg_len || ts.blocks().len() != bls_msg_len {
+            return Err(Error::Other(
+                "Msg included length didnt match tipset size".to_string(),
+            ));
+        }
```
austinabell, line 298: This check makes no sense, please explain

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +797,339 @@ where
...
+        let ts = construct_tipset(1);
+        let (bls, secp) = construct_messages();
+
+        let fts: FullTipset = cs
+            .tipset_msgs(
+                ts.clone(),
+                &[bls],
+                &[secp],
+                ts.blocks().len(),
+                ts.blocks().len(),
```
austinabell, line 1084: This is supposed to be the bls and msgs lengths, which I don't see why that should ever be parameters, because if that was the length that it was supposed to actually be, you have the two slices of bls and secp messages, length is redundant. You just created this test to follow the logic of the function, I don't think this test is very good

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +797,339 @@ where
...
+
+        // ensure response is on channel prior to spawning test thread
+        task::block_on(async {
+            event_sender
+                .send(NetworkEvent::RPCResponse {
+                    req_id: 0,
+                    response: c_response.clone(),
+                })
+                .await;
+        });
```
austinabell, line 1123: another case of the duplicate logic and redundant clones

utils/test_utils/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "test_utils"
+version = "0.1.0"
+authors = ["ChainSafe Systems <info@chainsafe.io>"]
+edition = "2018"
```
austinabell, line 5: Would be great if all of this logic is separated by a feature that is only used in dev-dependencies. By default when we compile, or run any tests, this will have to be compiled

utils/test_utils/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+address = { package = "forest_address", path = "../../vm/address" }
+blocks = { package = "forest_blocks", path = "../../blockchain/blocks/" }
+cid = { package = "forest_cid", path = "../../ipld/cid" }
+ipld_blockstore = { path = "../../ipld/blockstore" }
+chain_sync = { path = "../../blockchain/chain_sync/"}
```
austinabell, line 12:
```suggestion
```

blockchain/chain/src/store/chain_store.rs
```diff
@@ -22,11 +23,10 @@ pub struct ChainStore<DB> {
     pub db: Arc<DB>,
 
     // Tipset at the head of the best-known chain.
-    // TODO revisit if this should be pointer to tipset on heap
-    heaviest: Arc<Tipset>,
+    heaviest: Tipset,
```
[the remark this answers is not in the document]
dutterbutter: Yeah this will likely change when the task #327 is picked up

blockchain/chain/src/store/chain_store.rs
```diff
@@ -36,14 +36,20 @@ where
...
+        let heaviest = Tipset::new(vec![BlockHeader::default()]).unwrap();
         Self {
             db,
             tip_index: TipIndex::new(),
             heaviest,
         }
     }
 
+    /// Sets heaviest tipset within ChainStore
+    // TODO
```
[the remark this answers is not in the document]
dutterbutter: Potentially could change when #327 gets picked up, I can remove in the meantime

blockchain/chain_sync/src/sync.rs
```diff
@@ -153,33 +186,157 @@ where
...
         Ok(())
     }
 
+    /// Syncs messages by first checking state for message existence otherwise fetches messages from blocksync
+    async fn sync_messages_check_state(&mut self, ts: &[Tipset]) -> Result<(), Error> {
+        // see https://github.com/filecoin-project/lotus/blob/master/build/params_shared.go#L109 for request window size
+        const REQUEST_WINDOW: i64 = 200;
+
+        // set i to the length of provided tipsets
+        let mut i: i64 = i64::try_from(ts.len())? - 1;
```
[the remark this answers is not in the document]
dutterbutter: Yeah, I agree. I would be open to alternative approaches here. The usize seems required to index the tipsets, and i64 is required for the math operations.

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +797,339 @@ where
...
+
+    #[test]
+    fn chainsync_constructor() {
+        let db = MemoryDB::default();
+        let (local_sender, _test_receiver) = channel(20);
+        let (_event_sender, event_receiver) = channel(20);
+
+        // Test just makes sure that the chain syncer can be created without using a live database or
+        // p2p network (local channels to simulate network messages and responses)
+        let _chain_syncer = ChainSyncer::new(Arc::new(db), local_sender, event_receiver).unwrap();
```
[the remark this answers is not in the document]
dutterbutter: Why did it need to be in its own test directory by itself?

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +797,339 @@ where
...
+
+    pub fn chain_syncer_setup<Db>(db: Arc<Db>) -> ChainSyncer<Db>
+    where
+        Db: BlockStore,
+    {
+        let (local_sender, _test_receiver) = channel(20);
+        let (_event_sender, event_receiver) = channel(20);
+
+        ChainSyncer::new(db, local_sender, event_receiver).unwrap()
+    }
```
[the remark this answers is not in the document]
dutterbutter: I need the ChainSyncer itself not just the sender and receiver?

blockchain/chain_sync/src/sync.rs
dutterbutter: See lotus: https://github.com/filecoin-project/lotus/tree/master/https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L287

blockchain/chain_sync/src/sync.rs
```diff
@@ -380,6 +614,20 @@ where
...
     }
+    /// validates tipsets and adds header data to tipset tracker
+    fn validate_tipsets(&mut self, fts: FullTipset) -> Result<(), Error> {
+        if fts.tipset()? == self._genesis {
+            return Ok(());
+        }
+
+        for b in fts.blocks() {
+            if let Err(e) = self.validate(b) {
+                self.bad_blocks.put(b.cid().clone(), e.to_string());
```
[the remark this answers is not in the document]
dutterbutter: Will update.

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +797,339 @@ where
...
+        task::spawn(async move {
+            cs.schedule_tipset(ts.clone())
+                .await
+                .expect("error scheduling tipsets");
+            assert_eq!(cs.next_sync_target._tipsets(), &[ts]);
+
+            cs.schedule_tipset(ts_copy)
+                .await
+                .expect("error scheduling tipsets");
+            assert_eq!(cs.next_sync_target._tipsets().len(), 2);
```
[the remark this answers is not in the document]
dutterbutter: Yes, that is correct. I will update the test to make them different tipsets. The reason this did not fail is because the logic for identifying and handling related tipsets is a TODO.

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +797,339 @@ where
...
+    }
+
+    #[test]
+    fn compute_msg_data_given_msgs_test() {
+        let (bls, secp) = construct_messages();
+
+        let db = MemoryDB::default();
+        let cs = chain_syncer_setup(Arc::new(db));
+        let expected_root =
+            Cid::from_raw_cid("bafy2bzaced5inutkibck2wagtnggbvjpbr65ghdncivs3gpagx67s3xs3i5wa")
```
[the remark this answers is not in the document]
dutterbutter: No that's the CID generated from using the roots of the `construct_messages()` function. The only way I could think of having a deterministic root that I can check against.

blockchain/chain_sync/src/sync.rs
```diff
@@ -134,12 +141,38 @@ impl<DB> ChainSyncer<DB>
...
+                        source,
+                    );
+                }
+            }
+        }
+        Ok(())
+    }
 
+    /// Performs syncing process
+    async fn sync(&mut self, head: Tipset) -> Result<(), Error> {
```
ec2, line 175: I don't see why we need to pass ownership of head here. I just tried it with a reference instead, and it works perfectly

blockchain/chain_sync/src/sync.rs
dutterbutter: Will remove.

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +797,339 @@ where
...
+
+    #[test]
+    fn validate_msg_data_given_msgs_test() {
+        let block = construct_block();
+
+        let db = MemoryDB::default();
+        let cs = chain_syncer_setup(Arc::new(db));
+
+        cs.validate_msg_data(&block).expect("invalid msg data");
+    }
```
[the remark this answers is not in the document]
dutterbutter: Will remove.

blockchain/chain/src/store/chain_store.rs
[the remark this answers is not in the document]
austinabell: Then what is the reason for changing it here?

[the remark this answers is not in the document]
dutterbutter: Yeah agreed. I changed it back.

blockchain/chain_sync/src/sync.rs
austinabell: yeah? those last two parameters are the 2d arrays for message includes indexes, you are assigning ALL messages to ALL blocks, when you should be using the indexes to attach the propoer messages. That check you are doing here makes no sense

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +791,178 @@ where
...
+    #[test]
+    fn select_target_test() {
+        let ts_1 = construct_tipset(1, 5);
+        let ts_2 = construct_tipset(2, 10);
+        let ts_3 = construct_tipset(3, 7);
+
+        let db = MemoryDB::default();
+        let mut cs = chain_syncer_setup(Arc::new(db));
+
+        task::spawn(async move {
```
austinabell, line 941:
```suggestion
        task::block_on(async move {
```

utils/test_utils/src/chain_structures.rs
```diff
@@ -0,0 +1,220 @@
...
+use chain::TipSetMetadata;
+use cid::{multihash::Blake2b256, Cid};
+use crypto::{Signature, Signer, VRFResult};
+use encoding::to_vec;
+use forest_libp2p::blocksync::{BlockSyncResponse, TipSetBundle};
+use forest_libp2p::rpc::RPCResponse;
+
+use message::{SignedMessage, UnsignedMessage};
+use num_bigint::BigUint;
+use std::error::Error;
```
austinabell, line 18: This doesn't compile by default, could just add the `#![cfg(feature = "test_constructors")]` tag at the top of this file and may also have to add `#[cfg(feature = "test_constructors")]` to the lib.rs reference to this module. No need to add the feature tag to everything specifically, so can remove those

blockchain/blocks/tests/tipset_tests.rs
```diff
@@ -0,0 +1,70 @@
...
+#[test]
+fn equals_test() {
+    let tipset_keys = TipSetKeys {
+        cids: construct_keys().clone(),
+    };
+    let tipset_keys2 = TipSetKeys {
+        cids: construct_keys().clone(),
+    };
+    assert_eq!(tipset_keys.equals(&tipset_keys2), true);
+}
```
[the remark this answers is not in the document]
dutterbutter: You want me to remove this file then?

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +791,178 @@ where
...
+            event_sender
+                .send(NetworkEvent::Hello {
+                    message: HelloMessage::default(),
+                    source,
+                })
+                .await;
+
+            task::sleep(Duration::from_millis(50)).await;
+        });
+    }
```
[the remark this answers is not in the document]
dutterbutter: Will remove.

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +791,178 @@ where
...
+        // send HELLO message to channel
+        send_hello_msg();
+        // send blocksync response to channel
+        send_blocksync_response();
+
+        task::spawn(async {
+            cs.start()
+                .await
+                .expect("error starting sync process thread");
+        });
```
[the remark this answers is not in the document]
dutterbutter: Will remove.

blockchain/chain_sync/src/sync.rs
```diff
@@ -542,8 +791,178 @@ where
...
+        let fts = construct_full_tipset();
+
+        let db = MemoryDB::default();
+        let mut cs = chain_syncer_setup(Arc::new(db));
+
+        task::spawn(async move {
+            cs.inform_new_head(&fts)
+                .await
+                .expect("error with informing of new head");
+        });
```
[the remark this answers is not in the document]
dutterbutter: Will remove. It hangs because this file is still incomplete in certain areas (e.g. validation).

blockchain/chain_sync/src/bucket.rs
```diff
@@ -34,29 +35,50 @@ impl SyncBucket {
...
 pub(crate) struct SyncBucketSet {
     buckets: Vec<SyncBucket>,
 }
 
 impl SyncBucketSet {
+    /// Inserts a tipset into a bucket
     pub(crate) fn insert(&mut self, tipset: Arc<Tipset>) {
         for b in self.buckets.iter_mut() {
-            if b.same_chain_as(&tipset) {
+            if b.same_chain_as(tipset.clone()) {
```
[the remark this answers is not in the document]
dutterbutter: this clone is required.

blockchain/blocks/tests/tipset_tests.rs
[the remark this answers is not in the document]
austinabell: Don't have to if you think there is a good reason, I'm just trying to understand that reason

blockchain/chain_sync/src/bucket.rs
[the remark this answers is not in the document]
austinabell: Oh, didn't see the change in the same_chain_as function, why is that function signature changed to arc as a parameter? you should only need the reference to the TS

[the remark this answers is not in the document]
dutterbutter: In your previous review, you mentioned that these operations within `bucket.rs` are now cloning full tipsets instead of arc references....

https://github.com/ChainSafe/forest/pull/293#discussion_r402426818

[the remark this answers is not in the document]
austinabell: That comment was referring to the change of the underlying type and how you were cloning the whole tipset when the change was made, for functions like this there is no reason to have to pass an Arc<Tipset> if you only need the reference to that tipset for a check (basically just keeping it how it was before). Sorry if I worded poorly

blockchain/blocks/tests/tipset_tests.rs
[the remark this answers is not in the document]
dutterbutter: These tests were made months ago. Will remove.

PR #294, opened 2020-03-20
Sector types

### LINKED ISSUE #277: Create sector types
**Task summary**

Types needed for miner actor #265 and power actor #161 

**Specification reference**

- https://github.com/filecoin-project/specs-actors/blob/master/actors/abi/sector.go

austinabell:
**Summary of changes**

I'm not going to say that this was the most boring PR to make, but I also don't want to lie to you

Closes #277 


vm/src/sector/post.rs
```diff
@@ -0,0 +1,66 @@
...
+
+// TODO docs
+pub struct SectorInfo {
+    /// Used when sealing - needs to be mapped to PoSt registered proof when used to verify a PoSt
+    pub proof: RegisteredProof,
+    pub sector_number: SectorNumber,
+    pub sealed_cid: Cid,
+}
+
+// TODO docs
```
dutterbutter: I notice we have similar types in `ticket.rs` can we remove them there or are they used differently?

austinabell: yeah, I had a look, either they are different types or the ones that existed are now out of date, either way we will find out and remove if there is the connection (inevitably will not be missed later)

PR #295, opened 2020-03-22
Refactor annotated serializations

austinabell:
**Summary of changes**
- Thought of a cleaner way to override serialization through changes in #294 so I'm making it here to clean up the standard usage and remove the need to declare temporary types

Idk I think it's so cool what you can do with rust and serde here to make things very clean and efficient

Closes 


austinabell:
> looked more boring than the other one lol

Noo because it's doing new things rather than creating a bunch of types. Plus destructuring that wrapper inside of the tuple destructure to deserialize in oneline without the interim struct is pretty neato if you ask me !

PR #308, opened 2020-03-26
Storage Power actor

### LINKED ISSUE #161: Implement Basic Storage Power Actor
A basic storage power actor framework has already been implemented however the defined task below provides additional details of what is required. 

**Storage Power Actor Definition:**
- https://filecoin-project.github.io/specs/#storagepoweractor-interface

**Lotus implementation:**
- https://github.com/filecoin-project/lotus/blob/master/chain/actors/actor_storagepower.go

**Data structures required:**
- [ ] `StoragePowerActorState`
- [ ] `CreateStorageMinerParams`
- [ ] `PowerTableHAMT`
- [ ] `MinerEventsHAMT`

**Methods required defined via spec:**
- `_slashPledgeCollateral`
- `_minerNominalPowerMeetsConsensusMinimum`
- `_selectMinersToSurprise`
- `_addClaimedPowerForSector` 
- `_deductClaimedPowerForSectorAssert`
- `_updatePowerEntriesFromClaimedPower`
- `_getCurrPledgeForMiner`
- `_getPowerTotalForMiner`
- `_getPledgeSlashForConsensusFault`
- `_setClaimedPowerEntryInternal`
- `_setNominalPowerEntryInternal`
- `_setPowerEntryInternal`

**Lotus methods defined:**
- https://github.com/filecoin-project/lotus/blob/master/chain/actors/actor_storagepower.go#L24

austinabell:
**Summary of changes**
- Implements storage power actor

Sorry for big PR, not much I could do to break up lol

Some logic is changed a bit with this due to the reliance on negative token amounts for logic within the go/spec impl. I also had to clean up some other small auxiliary things, I tried to keep very contained and they should be very unopinionated.

If you notice the deref and ref of the lazy static items, that is because the intended way to access the underlying value is through the deref (cleaner to denote specifically what is happening instead of an implicit deref through the let binding, I should have read the docs before using lol)

Closes #161


vm/actor/src/builtin/power/policy.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub static ref PLEDGE_FACTOR: BigInt = BigInt::from(3); // PARAM_FINISH
+
+    /// Total expected block reward per epoch (per-winner reward * expected winners), as input to pledge requirement.
+    pub static ref EPOCH_TOTAL_EXPECTED_REWARD: BigInt = reward::BLOCK_REWARD_TARGET.clone() * 5; // PARAM_FINISH
+
+    /// Minimum power of an individual miner to meet the threshold for leader election.
+    pub static ref CONSENSUS_MINER_MIN_POWER: StoragePower = StoragePower::from(2 << 30); // placeholder
+
+}
+
```
dutterbutter: Add comment: "Penalty to pledge collateral for the termination of an individual sector."

vm/actor/src/builtin/power/policy.rs
```diff
@@ -0,0 +1,104 @@
...
+    TokenAmount::new(0)
+}
+
+/// Penalty to pledge collateral for a consensus fault.
+pub(super) fn pledge_penalty_for_consensus_fault(
+    pledge: TokenAmount,
+    fault_type: ConsensusFaultType,
+) -> TokenAmount {
+    // PARAM_FINISH: always penalise the entire pledge.
+    match fault_type {
```
dutterbutter: Would we need a catch-all statement here if nothing is hit?

vm/actor/src/builtin/power/policy.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+pub(super) fn pledge_penalty_for_sector_termination(
+    _pledge: &TokenAmount,
+    _term_type: SectorTermination,
+) -> TokenAmount {
+    // PARAM_FINISH
+    TokenAmount::new(0)
+}
+
```
dutterbutter: Seems `pledgePenaltyForWindowedPoStFailure`is missing? See here in [policy.go.](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/power/policy.go#L44)

vm/actor/src/builtin/power/types.rs
```diff
@@ -0,0 +1,410 @@
...
+
+pub type SectorTermination = i64;
+
+/// Implicit termination after all deals expire
+pub const SECTOR_TERMINATION_EXPIRED: SectorTermination = 0;
+/// Unscheduled explicit termination by the miner
+pub const SECTOR_TERMINATION_MANUAL: SectorTermination = 1;
+
+#[derive(Clone)]
+pub struct SectorStorageWeightDesc {
```
dutterbutter: It might be worth adding comments to each of these types, I'll leave that to your judgment.

vm/actor/src/builtin/power/mod.rs
```diff
@@ -43,15 +58,692 @@ impl Method {
...
+                let mut table = BalanceTable::from_root(rt.store(), &st.escrow_table)?;
+                let balance = table.remove(miner)?;
+
+                st.escrow_table = table.root()?;
+
+                Ok(balance)
+            })
+            .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e))?;
+
+        // TODO switch 6 to OnDeleteMiner on miner actor impl
```
dutterbutter: Can't you do this now?

vm/actor/src/builtin/power/policy.rs
austinabell: no, because that would fail in serialization. All patterns are matched (rust doesn't let otherwise)

vm/actor/src/builtin/power/policy.rs
austinabell: good catch ! I mistakenly referenced the other temporary function

vm/actor/src/builtin/power/types.rs
austinabell: I just don't know enough about them yet to add comments on all of them, I will make a pass through later when more specifics are ironed out

vm/actor/src/builtin/power/mod.rs
austinabell: I want to keep changes contained, those method definitions don't exist in the miner actor (it's empty rn)

vm/actor/src/builtin/power/policy.rs
```diff
@@ -0,0 +1,114 @@
...
+pub(super) fn reward_for_consensus_slash_report(
+    elapsed_epoch: ChainEpoch,
+    collateral: TokenAmount,
+) -> TokenAmount {
+    // PARAM_FINISH
+    // var growthRate = SLASHER_SHARE_GROWTH_RATE_NUM / SLASHER_SHARE_GROWTH_RATE_DENOM
+    // var multiplier = growthRate^elapsed_epoch
+    // var slasherProportion = min(INITIAL_SLASHER_SHARE * multiplier, 1.0)
```
ec2: Did you forget to remove these comments?

vm/actor/src/builtin/power/policy.rs
```diff
@@ -0,0 +1,114 @@
...
+) -> TokenAmount {
+    // PARAM_FINISH: always penalise the entire pledge.
+    match fault_type {
+        ConsensusFaultType::DoubleForkMining => pledge,
+        ConsensusFaultType::ParentGrinding => pledge,
+        ConsensusFaultType::TimeOffsetMining => pledge,
+    }
+}
+
+lazy_static! {
```
ec2: Is there a reason why these arent just declared inside the function? Doesnt look like youre using them anywhere else. Do you think they will in the future?

austinabell: Better to have them in a static context and as variables that can more easily be modified in future, are you saying function referring to `reward_for_consensus_slash_report` ?

vm/actor/src/builtin/power/policy.rs
austinabell: Was just to give context based on the spec impl, but I removed

vm/actor/src/builtin/power/policy.rs
ec2: Yeah, thats the function I'm referring to. I'm ok with them being static, though. Doesn't have much impact, and you might be right about future modifications.

PR #309, opened 2020-03-27
Change ChainEpoch and TokenAmount types

austinabell:
**Summary of changes**
- Since no methods on these types or any need for it, changes to aliases to not require to manually implement traits and wrap underlying types

Closes 


vm/src/token.rs
```diff
@@ -1,103 +1,6 @@
...
-use num_bigint::{biguint_ser, BigInt, BigUint, ParseBigIntError};
-use num_traits::CheckedSub;
-use serde::{Deserialize, Serialize};
-use std::convert::TryFrom;
-use std::fmt;
-use std::ops::{Add, AddAssign, Mul, Sub};
-use std::str::FromStr;
+use num_bigint::BigUint;
 
-/// Wrapper around a big int variable to handle token specific functionality
```
ec2: This serializes properly? Having a biguint wrapped in a tuple struct is serialized differently... Though i dont know if we have any interop test wrt to that

austinabell: Although I have no idea what specifically you are talking about because there are no new serializations that are added in this PR, I'll assume you mean the BigInt serialization wrappers that were introduced in #295. Yes, of course there are interop serialization tests, such as the message serialization vectors which includes two `TokenAmount`s and the test you wrote that tests the serialization of that one blockheader which includes uses this same wrapper type to name just a few.

This literally doesn't matter if it's wrapped because the serialization implementation is custom, but even if it was, like in the case of TokenAmount with the annotation before this change, then it still serializes as the parameter if there is only one parameter. Remember when you had to serialize VRFProof in blockheader as a 1 length array because of this?

PR #330, opened 2020-04-03
Market actor state

### LINKED ISSUE #320: Market Actor State

austinabell:
**Summary of changes**
- Everything but the actor implementation

Closes #320 


vm/actor/src/builtin/market/state.rs
```diff
@@ -0,0 +1,567 @@
...
+        BS: BlockStore,
+    {
+        assert!(state.sector_start_epoch.is_some());
+
+        self.unlock_balance(store, &deal.provider, &deal.provider_collateral)?;
+        self.unlock_balance(store, &deal.client, &deal.client_collateral)?;
+
+        self.delete_deal(store, deal_id, deal)
+    }
+
```
dutterbutter, line 453: Although I don't see it used anywhere, [`generateStorageDealID`](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_state.go#L220) is not included here.

austinabell: it's only used in the market actor, which is why I think I left it out initially. I added it now anyway though

vm/actor/src/builtin/market/state.rs
```diff
@@ -0,0 +1,573 @@
...
+            if epoch > deal.start_epoch {
+                return self.process_deal_init_timed_out(store, deal_id, deal, state);
+            }
+            return Ok(TokenAmount::zero());
+        }
+
+        assert!(deal.start_epoch <= epoch);
+
+        let deal_end = if ever_slashed {
+            assert!(state.slash_epoch.unwrap() <= deal.end_epoch);
```
ec2, line 323: Do we want to leave assets in the code?

ec2: asserts*

austinabell: no reason not to I don't think, if it fails it is a bug or something malicious, and would be caught. Not really a good reason to just pass back an error that could be missed

ec2: True. Would it make sense to fail with a message then? Or do asserts give you the line number?

austinabell: They do, I can add a message anyway I guess, this isn't something we expect to ever hit though, just a sanity check

Issue #324, opened 2020-03-31
Resolve actor hash key impls


austinabell:
**Issue summary**

Currently, there are placeholders for the hash keys for the map key types in the actors, but this needs to be updated and tested to match the reference implementation (specs-actors). 

The issue with our current impls are there are lossy generation of a String only because in go they are specifically coercing bytes to string as they have it hard coded to have a string key for their Hamt. Reason for lossy conversion in rust is because it requires unsafe code to convert non-utf8 bytes to string, and it would come with the pain of ensuring the keys are not used in any unsafe way on any retrieval. This shouldn't be an issue because bytes should hash the same as a string of the same bytes, but this will have to be tested against their implementation and within our client.

Ideal is that bytes and string hashing matches in hamt and keys are just used as bytes within the actors code, and hopefully no compromises will need to be made.

I am planning on doing this issue after my current, so ping me if you would like to take this on.

**Other information and links**

PR #333, opened 2020-04-06
Key hashing compatibility

austinabell:
**Summary of changes**
- In go impls, they use a string as the key for the hamt despite the fact that it is really non utf-8 bytes. To solve this, there is the custom type introduced in this PR (`BytesKey` but open to a better name) for the type of the key to match. 
  -The constraints on this type is that it has to hash the same as go implementation and that it serializes the same. str and [u8] hash the same with our impl so it's fine to use bytes as the hash type and that matches, and to get the serialization to match the bytes need to be unsafely converting to a string before serialization.
  - To clarify, a custom type would have to be created anyway even if go impls switch to using a byte slice as keys because a u8 slice serializes as an array of u8 by default, so this isn't all wasted work

Edit: go impl changed to serialize keys as bytes, so this update changes key type used within Actors

Closes 
#324 


vm/actor/src/util/balance_table.rs
```diff
@@ -55,22 +55,22 @@ where
...
 
     /// Adds an amount to a balance. Creates entry if not exists
     pub fn add_create(&mut self, key: &Address, value: TokenAmount) -> Result<(), String> {
-        let new_val = match self.0.get::<_, BigUintDe>(&key.hash_key())? {
+        let new_val = match self.0.get::<_, BigUintDe>(&key.to_bytes())? {
             Some(v) => v.0 + value,
             None => value,
         };
-        Ok(self.0.set(key.hash_key(), BigUintDe(new_val))?)
+        Ok(self.0.set(key.to_bytes().into(), BigUintDe(new_val))?)
```
dignifiedquire, line 73: you could just implement `impl<T: AsRef<str>> From<T> for BytesKey` and avoid all these `to_bytes` calls

vm/actor/src/lib.rs
```diff
@@ -36,17 +41,91 @@ fn check_empty_params(params: &Serialized) -> Result<(), EncodingError> {
...
+        &self.0
+    }
+}
+
+impl Serialize for BytesKey {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        unsafe { str::from_utf8_unchecked(&self.0) }.serialize(serializer)
```
dignifiedquire, line 101: Something sounds wrong here, quoting from the Cbor RFC:

> a text string, specifically a string of Unicode characters that is encoded as UTF-8

So either the go implementation is generating invalid cbor, or storing its strings as bytes. If they do use bytes, you can just use bytes as well and avoid the unsafe. If they generate invalid cbor, that should be fixed and not worked around.

austinabell: ah yeah, I mentioned over a call to others that this was a temporary measure and I indicate it in the TODO, I will bring it up as an issue in the go implementations, since they are encoding Hamt keys as strings when they are actually non utf8 bytes. I hate having to do this, but need compatibility for now until this is resolved

vm/actor/src/util/balance_table.rs
austinabell: hmm, I don't think this helps, I'm not sure if you're seeing that the key is of the address type, it is being converted into a `Vec<u8>`  and then the `into()` is putting it as the wrapper type used as a key here (`BytesKey`) which is specifically needed to ensure the string serialization of those bytes

dignifiedquire: Ah sorry, `.to_bytes()` made me think it was a string,

austinabell:
Gonna let this PR hang for a bit, was indicated that the key serialization is going to be fixed in go impls, so might as well put this on ice to not have to revisit

austinabell:
bs_ask

PR #336, opened 2020-04-08
Refactor error handling

austinabell:
**Summary of changes**
- Been bugging me for a while now, and now seemed like a good time to change because less conflicts
  - Switches all errors to derive std::error::Error through the `thiserror` crate (https://github.com/dtolnay/thiserror) which makes it more extensible and clean/quick to use

There are still things I think could be improved, but I tried to keep these changes unopinionated without making any functional changes.

Only functional changes are two things I suspect to be bugs [here](https://github.com/ChainSafe/forest/compare/austin/errorrefactor?expand=1#diff-84b58b922d122fcf1338a9609b10fb68R95) if that doesn't resolve, the change(s) in `chain_store.rs` @dutterbutter can you confirm there isn't a specific reason it should not change

Closes 
#216 

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -92,7 +92,7 @@ where
             let key = m.cid()?.key();
             let value = &m.marshal_cbor()?;
             if self.db.exists(&key)? {
-                return Err(Error::KeyValueStore("Keys exist".to_string()));
+                return Ok(());
             }
             self.db.write(&key, value)?
         }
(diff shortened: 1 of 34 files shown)


dutterbutter:
@austinabell yes that should return `Ok(())` rather than an error. Nice catch!

blockchain/chain/src/store/chain_store.rs
```diff
@@ -92,7 +92,7 @@ where
             let key = m.cid()?.key();
             let value = &m.marshal_cbor()?;
             if self.db.exists(&key)? {
-                return Err(Error::KeyValueStore("Keys exist".to_string()));
+                return Ok(());
```
ec2, line 95: Why did this change from an Err to Ok?

austinabell: Because it was a bug, re: conversation above

PR #337, opened 2020-04-09
Update block header serialization

austinabell:
**Summary of changes**
- Either some things have changed or there was some discrepancies in the port from go, They are using a pointer for signatures and the nil value gets serialized as null in Cbor, which is equivalent to `Option` or `None` in rust. They were checking the nil if bytes are empty, which doesn't translate to Rust or how we had our types setup
  - Also found a bug in the deserialization of Signature that I did a while ago, if bytes are empty it will panic on `.remove(0)` because that impl expects a valid serialization

Edit: Also updates EPoStProof serialization as discussed, and removed the tests that were statically setup for the old serialization

Edit2: Added test runner for block header serialization vectors, but the vectors were a little broken so I switched the submodule to be my fork for now

Closes 


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "tests/serialization-vectors"]
 	path = tests/serialization-vectors
-	url = https://github.com/filecoin-project/serialization-vectors
+	url = https://github.com/austinabell/serialization-vectors
```
dutterbutter, line 3: Should prob move this under chainsafe

PR #338, opened 2020-04-09
Market actor implementation

### LINKED ISSUE #264: Implement updated market actor
**Issue summary**

https://github.com/filecoin-project/specs-actors/tree/master/actors/builtin/market

dutterbutter:
**Summary of changes**
- Added market actor and its types
- Updated transaction method params

Closes #264 

- [Market actor spec](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go#L32)


vm/src/error.rs
```diff
@@ -49,3 +51,33 @@ impl From<EncodingError> for ActorError {
...
+
+impl From<String> for ActorError {
+    fn from(e: String) -> Self {
+        Self {
+            fatal: false,
+            exit_code: ExitCode::ErrSerialization,
+            msg: e,
+        }
+    }
+}
```
austinabell, line 83: Probably not good to just forward all errors from this. Message receipts are very specific, and exit codes are specifically used in consensus, so forwarding all ExitCodes to a serialized error is probably incorrect. I haven't been as diligent in the other impls to make sure until the spec is finalized, but it should definitely be attempted to match error codes 1:1 for each specific case

vm/actor/src/builtin/paych/state.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
```
austinabell, line 6: duplicate header

vm/actor/src/builtin/market/types.rs
```diff
@@ -0,0 +1,184 @@
...
+
+impl<'de> Deserialize<'de> for OnMinerSectorsTerminateParams {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let deal_ids = Deserialize::deserialize(deserializer)?;
+        Ok(Self { deal_ids })
+    }
+}
```
austinabell, line 59: This doesn't serialize correctly, need to serialize as 1 length array

vm/actor/src/builtin/market/types.rs
```diff
@@ -0,0 +1,184 @@
...
+
+impl<'de> Deserialize<'de> for HandleExpiredDealsParams {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let deal_ids = Deserialize::deserialize(deserializer)?;
+        Ok(Self { deal_ids })
+    }
+}
```
austinabell, line 82: same as above

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,652 @@
...
+                format!("signature proposal invalid: {}", e),
+            )
+        })?;
+
+    Ok(())
+}
+
+// Resolves a provider or client address to the canonical form against which a balance should be held, and
+// the designated recipient address of withdrawals (which is the same, for simple account parties).
+pub fn escrow_address<BS, RT>(rt: &RT, addr: &Address) -> Result<(Address, Address), ActorError>
```
austinabell, line 586: doesn't need to be public?

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,652 @@
...
+                Self::withdraw_balance(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
+            }
+            Some(Method::HandleExpiredDeals) => {
+                Self::handle_expired_deal(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
+            }
+            Some(Method::PublishStorageDeals) => {
+                Self::publish_storage_deals(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
```
austinabell, line 638: This should return the serialized result from the function, not default

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,652 @@
...
+            }
+            Some(Method::VerifyDealsOnSectorProveCommit) => {
+                Self::verify_deals_on_sector_prove_commit(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
```
austinabell, line 642: same as above

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,652 @@
...
+            }
+            Some(Method::OnMinerSectorsTerminate) => {
+                Self::on_miners_sector_terminate(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
+            }
+            Some(Method::ComputeDataCommitment) => {
+                Self::compute_data_commitment(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
```
austinabell, line 650: saa

vm/actor/src/builtin/market/state.rs
```diff
@@ -245,7 +245,6 @@ impl State {
     // Deal state operations
     ////////////////////////////////////////////////////////////////////////////////
 
-    #[allow(dead_code)]
```
austinabell, line 248: See you removed some of these in this file, all should be removed in this whole folder, I just added them until this came in to avoid linting issues (can remove the TODO's on two of the files too about this)

vm/src/error.rs
dutterbutter: Yeah, that makes sense. I'll remove and just map the err where needed.

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+        params: WithdrawBalanceParams,
+    ) -> Result<(), ActorError>
+    where
+        BS: BlockStore,
+        RT: Runtime<BS>,
+    {
+        let (nominal, recipient) = escrow_address(rt, &params.provider_or_client)?;
+
+        let mut amount_slashed_total = TokenAmount::zero();
+        let mut amount_extracted = TokenAmount::zero();
```
ec2, line 102: What does this do? It gets assigned at line 125 and that's it. It never gets used...

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+                "Empty deals parameter.".to_owned(),
+            ));
+        }
+
+        // All deals should have the same provider so get worker once
+        let provider_raw = &params.deals[0].proposal.provider;
+        let provider = rt.resolve_address(&provider_raw)?;
+
+        let (_, worker) = request_miner_control_addrs(rt, &provider)?;
+        if worker != rt.message().from().clone() {
```
ec2, line 204: you can borrow worker instead of cloning rt.message().from()

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+            return Err(ActorError::new(
+                ExitCode::ErrForbidden,
+                format!("Caller is not provider {}", worker),
+            ));
+        }
+
+        // All deals should have the same provider so get worker once
+        let provider_raw = &params.deals[0].clone().proposal.provider;
```
ec2, line 212: I think you can clone at the end instead of cloning the whole deal `&params.deals[0].proposal.provider.clone();`

austinabell:
CI is broken because you reverted the submodule changes after you merged in commit af4dd79 

To fix either undo that commit, or just cd into that submodule and do a git pull (pretty sure the latter should work if you commit the change)

vm/actor/src/builtin/market/mod.rs
austinabell: ^
```suggestion
        if &worker != rt.message().from() {
```

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+        let provider = rt.resolve_address(&provider_raw)?;
```
austinabell, line 213: nit but can you take the reference from either one of these lines, whichever makes more sense (confusing to what type provider_raw is)

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+
+        let mut new_deal_ids: Vec<DealID> = Vec::new();
+        rt.transaction::<State, Result<(), ActorError>, _>(|st: &mut State, rt| {
+            let mut prop = Amt::load(&st.proposals, rt.store())
+                .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?;
+            let mut dbp = SetMultimap::from_root(rt.store(), &st.deal_ids_by_party)
+                .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?;
```
austinabell, line 220: Errors should be ErrIllegalState

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+                format!("Caller is not provider {}", worker),
+            ));
+        }
+
+        // All deals should have the same provider so get worker once
+        let provider_raw = &params.deals[0].clone().proposal.provider;
+        let provider = rt.resolve_address(&provider_raw)?;
+
+        let mut new_deal_ids: Vec<DealID> = Vec::new();
+        rt.transaction::<State, Result<(), ActorError>, _>(|st: &mut State, rt| {
```
austinabell, line 216:
```suggestion
        rt.transaction(|st: &mut State, rt| {
```

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+            }
+            Ok(())
+        })??;
+
+        let commd = rt
+            .syscalls()
+            .compute_unsealed_sector_cid(params.sector_type, &pieces)
+            .map_err(|e| {
+                ActorError::new(
+                    ExitCode::ErrIllegalArgument,
```
austinabell, line 402:
```suggestion
                    ExitCode::SysErrorIllegalArgument,
```

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+        RT: Runtime<BS>,
+    {
+        rt.validate_immediate_caller_type(std::iter::once(&*MINER_ACTOR_CODE_ID))?;
+        let miner_addr = rt.message().from().clone();
+
+        rt.transaction::<State, Result<(), ActorError>, _>(|st, rt| {
+            let prop = Amt::load(&st.proposals, rt.store())
+                .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?;
+            let mut states = Amt::load(&st.states, rt.store())
+                .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?;
```
austinabell, line 427: State exit codes

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+                    .get(id)
+                    .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?
+                    .ok_or_else(|| {
+                        ActorError::new(ExitCode::ErrIllegalState, "Get deal error".to_owned())
+                    })?;
+                assert_eq!(deal.provider, miner_addr);
+
+                let mut state: DealState = states
+                    .get(id)
+                    .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?
```
austinabell, line 440: all should be state exit codes

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+        rt.transaction::<State, Result<(), ActorError>, _>(|st, rt| {
+            let prop = Amt::load(&st.proposals, rt.store())
+                .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?;
+            let mut states = Amt::load(&st.states, rt.store())
+                .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?;
+
+            for id in params.deal_ids {
+                let deal: DealProposal = prop
+                    .get(id)
+                    .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e.into()))?
```
austinabell, line 432:
```suggestion
                    .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e.into()))?
```

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+
+            st.states = states
+                .flush()
+                .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e.into()))?;
+            Ok(())
+        })??;
+        Ok(())
+    }
+
+    fn handle_expired_deal<BS, RT>(
```
austinabell, line 463:
```suggestion
    fn handle_expired_deals<BS, RT>(
```

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+        BS: BlockStore,
+        RT: Runtime<BS>,
+    {
+        rt.validate_immediate_caller_type(CALLER_TYPES_SIGNABLE.iter())?;
+        let mut slashed = TokenAmount::zero();
+        let curr_epoch = rt.curr_epoch();
+        rt.transaction::<State, Result<(), ActorError>, _>(|st: &mut State, rt| {
+            slashed = st.update_pending_deal_states(rt.store(), params.deal_ids, curr_epoch)?;
+            Ok(())
+        })??;
```
austinabell, line 477:
```suggestion
        let curr_epoch = rt.curr_epoch();
        let slashed = rt.transaction(|st: &mut State, rt| {
            st.update_pending_deal_states(rt.store(), params.deal_ids, curr_epoch)
        })??;
```

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+}
+////////////////////////////////////////////////////////////////////////////////
+// Checks
+////////////////////////////////////////////////////////////////////////////////
+fn validate_deal_can_activate(
+    curr_epoch: ChainEpoch,
+    miner_addr: &Address,
+    sector_exp: ChainEpoch,
+    proposal: &DealProposal,
+) -> Result<(), ActorError> {
```
austinabell, line 498: You're missing the deal state check for if undefined (in our case `None`)

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+
+// Resolves a provider or client address to the canonical form against which a balance should be held, and
+// the designated recipient address of withdrawals (which is the same, for simple account parties).
+fn escrow_address<BS, RT>(rt: &RT, addr: &Address) -> Result<(Address, Address), ActorError>
+where
+    BS: BlockStore,
+    RT: Runtime<BS>,
+{
+    // Resolve the provided address to the canonical form against which the balance is held.
+    let nominal = rt.resolve_address(addr)?;
```
austinabell, line 627: need to map this error, the resolve by default returns a placeholder exit code (This should probably be reworked in our runtime interface to not do this and potentially cause a conflict, thoughts @ec2 ?)

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+
+    let code_id = rt.get_actor_code_cid(&nominal)?;
```
austinabell, line 629: Same here, error returned is a `SysErrInternal` which is going to cause problems like the previous.

cc: @ec2

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+
+    if code_id != *MINER_ACTOR_CODE_ID {
+        // Ordinary account-style actor entry; funds recipient is just the entry address itself.
+        rt.validate_immediate_caller_type(CALLER_TYPES_SIGNABLE.iter())?;
```
austinabell, line 633: Can you add the TODO to for the `RequestMinerControlAddrs` call equivalent that is left out so this doesn't get missed (maybe you left out because it is connected to Miner Actor?)

dutterbutter: That is there, I just changed the conditional blocks.

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,687 @@
...
+                .ok_or_else(|| {
+                    ActorError::new(
+                        ExitCode::ErrIllegalArgument,
+                        format!(
+                            "invalid sector expiry epoch {:?} ahead of current {:?}",
+                            params.sector_expiry,
+                            rt.curr_epoch()
+                        ),
+                    )
+                })?;
```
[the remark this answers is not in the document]
austinabell: I know this may sound weird, but you actually need to do 
```
.checked_sub(..).ok_or(1u8)
```
lol

I thought through the logic and their impl doesn't error when this value is negative, so this needs to succeed, divide 0 by a valid number and return successfully and/or panic on a divide by 0

https://github.com/filecoin-project/specs-actors/issues/280

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,710 @@
...
+            let sector_space_time: BigInt = BigInt::from(params.sector_size as u64 * epoch_value);
```
austinabell, line 367: You need to wrap the `params.sector_size` in a big int before doing the multiplication, this did not prevent the overflow I mentioned before

vm/actor/src/builtin/market/mod.rs
```diff
@@ -99,32 +99,31 @@ impl Actor {
...
 
         let mut amount_slashed_total = TokenAmount::zero();
-        let mut amount_extracted = TokenAmount::zero();
-        rt.transaction::<State, Result<(), ActorError>, _>(|st: &mut State, rt| {
-            // Before any operations that check the balance tables for funds, execute all deferred
-            // deal state updates.
-            amount_slashed_total +=
-                st.update_pending_deal_states_for_party(rt.store(), rt.curr_epoch(), &nominal)?;
+        let amount_extracted =
+            rt.transaction::<State, Result<TokenAmount, ActorError>, _>(|st: &mut State, rt| {
```
austinabell, line 103: optional: you can remove the duplicate `State` type definition

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,710 @@
...
+    proposal: &DealProposal,
+) -> Result<(), ActorError> {
+    if &proposal.provider != miner_addr {
+        return Err(ActorError::new(
+            ExitCode::ErrIllegalArgument,
+            "Deal has incorrect miner as its provider.".to_owned(),
+        ));
+    };
+
+    if deal.sector_start_epoch != OptionalEpoch(None) {
```
austinabell, line 512:
```suggestion
    if deal.sector_start_epoch.is_some() {
```
Can use optional epoch type as `Option<ChainEpoch>` because I impl deref on it

vm/actor/src/builtin/market/mod.rs
```diff
@@ -4,7 +4,713 @@
...
+                        format!(
+                            "invalid sector expiry epoch {:?} ahead of current {:?}",
+                            params.sector_expiry,
+                            rt.curr_epoch()
+                        ),
+                    )
+                })?;
+
+            let param_sec_size = BigInt::from(params.sector_size as u64);
+            let sector_space_time = param_sec_size * epoch_value;
```
austinabell, line 370: for future reference, could have done (just had to wrap the first parameter):
```suggestion
           let sector_space_time: BigInt = BigInt::from(params.sector_size as u64) * epoch_value;
```

vm/actor/src/builtin/market/mod.rs
austinabell: oops I missed this, you didn't change this. To be precise, it needs to be:
```
let epoch_value = params
                .sector_expiry
                .checked_sub(rt.curr_epoch())
                .unwrap_or(1u8.into());
```

dutterbutter: This is changed? 

```
let epoch_value = params
                .sector_expiry
                .checked_sub(rt.curr_epoch())
                .ok_or(1u8) 
                .map_err(|_| {
                    ActorError::new(
                        ExitCode::ErrIllegalArgument,
                        format!(
                            "invalid sector expiry epoch {:?} ahead of current {:?}",
                            params.sector_expiry,
                            rt.curr_epoch()
                        ),
                    )
                })?;
```

austinabell: nah, what you're doing there is converting the `Option<ChainEpoch>` to `Result<ChainEpoch, u8>` then unwrapping that and ignoring the u8 value to throw the error anyway

austinabell: sorry if my previous comment was misleading, it was just meant as pseudocode

PR #342, opened 2020-04-10
Connect remaining Actor invocations to runtime and cleanup

austinabell:
**Summary of changes**
- #338 came in and didn't notice until after it wasn't connected to runtime so did that and also connected a base miner actor so on that implementation the changes are more contained
- Cleaned up things I noticed along the way and fixed some exit codes I noticed
  - Noticed other things I wanted to fix, but changes ended up being too extensive and breaking so will leave for another PR and left a few TODOs of for sure broken things

Closes 

### DIFF
--- a/vm/actor/src/builtin/market/mod.rs
+++ b/vm/actor/src/builtin/market/mod.rs
@@ -45,12 +45,7 @@ pub enum Method {
     OnMinerSectorsTerminate = 7,
     ComputeDataCommitment = 8,
 }
-impl Method {
-    /// Converts a method number into a Method enum
-    fn from_method_num(m: MethodNum) -> Option<Method> {
-        FromPrimitive::from_u64(m)
-    }
-}
+
 /// Market Actor
 pub struct Actor;
 impl Actor {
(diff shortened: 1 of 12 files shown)


vm/actor/src/builtin/market/mod.rs
```diff
@@ -86,9 +81,42 @@ impl Actor {
         rt.create(&st)?;
         Ok(())
     }
+
```
dutterbutter, line 84: What's the reason for moving these market methods around?

vm/default_runtime/src/lib.rs
```diff
@@ -410,42 +412,38 @@ pub fn internal_send<ST: StateTree, DB: BlockStore>(
...
                 }
                 x if x == *REWARD_ACTOR_CODE_ID => {
-                    actor::cron::Actor.invoke_method(&mut *runtime, *method_num, msg.params())
+                    actor::cron::Actor.invoke_method(runtime, *method_num, msg.params())
                 }
                 _ => Err(ActorError::new(
-                    ExitCode::SysErrForbidden,
-                    "invalid method id".to_owned(),
+                    ExitCode::SysErrorIllegalActor,
+                    format!("no code for actor at address {}", msg.to()),
```
dutterbutter, line 446: Since you are updating the exit code here, I think the `transfer` map_err exit code on ~LN 401, near the start of this fn should be changed from `ExitCode::SysErrInternal` to `ExitCode::SysErrActorNotFound`.

vm/actor/src/builtin/market/mod.rs
austinabell: They weren't in order, it was bugging me a bit to follow some of them lol, can swap back if you guys would like

dutterbutter: No, not a problem. I just was curious if it was required.

vm/default_runtime/src/lib.rs
austinabell: Also updated the names of the exit codes to match 1:1, not sure if changed since eric updated

vm/actor/src/builtin/market/mod.rs
austinabell: Yeah no functional change there, mb should have just noticed when it was PRed or not changed again but yeah just easier to jump to implementations

PR #350, opened 2020-04-14
VM gas usage implementation and refactor

### LINKED ISSUE #332: VM Gas metering and usage
- Need to implement gas table and consumption of gas through the runtime execution

austinabell:
**Summary of changes**
- Implements gas consumption with explicit calls and implicit through `GasBlockStore` and `GasSyscalls` #332 
  - GasTracker is the shared object which is shared between those two objects and the runtime itself. This is achieved by wrapping it in a `Rc<RefCell<>>` which can have a runtime panic if a mutable borrow is attempted when any other borrow exists so make sure if the `GasTracker` is borrowed that it is dropped as soon as possible

- Removes error from Cid generation from Cbor bytes
  - Could have done before, but wanted to keep API as similar to upstream, but that isn't a concern anymore

- Refactored some error handling, downcasting errors in future is going to be needed to match the patterns in go to interop 1:1 with exit codes and this is a step toward that 

Closes #332


vm/default_runtime/src/lib.rs
```diff
@@ -16,55 +21,98 @@ use forest_encoding::Cbor;
...
+        self.gas_tracker.borrow_mut().charge_gas(to_use)
+    }
+
+    /// Returns gas used by runtime
+    /// * Will borrow gas tracker RefCell, do not call if a mutable reference exists
+    pub fn gas_used(&self) -> i64 {
+        self.gas_tracker.borrow().gas_used()
+    }
+
+    fn gas_available(&self) -> i64 {
```
dutterbutter, line 109: Nit: add comment

vm/src/gas/price_list.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.create_actor_base + self.create_actor_extra
+    }
+    /// Returns the gas required for deleting an actor
+    #[inline]
+    pub fn on_delete_actor(&self) -> i64 {
+        self.delete_actor
+    }
+    /// Returns gas required for signature verification
+    #[inline]
+    pub fn on_verify_signature(&self, sig_type: SignatureType, plain_text_size: usize) -> i64 {
```
dutterbutter, line 123: I notice in the lotus implementation that they are passing back a `SysErrInternal` exit code if the retrieval of the `costFn` fails. Do we need to match this exit code?

Also, I would include the following TODO as mentioned [here](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/vm/gas.go#L59).

`this cost is not persistable, create a LinearCost{a,b} struct that has a .Cost(x) -> ax + b`

vm/src/gas/price_list.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.hashing_base + data_size as i64 * self.hashing_per_byte
+    }
+    /// Returns gas required for computing unsealed sector Cid
+    #[inline]
+    pub fn on_compute_unsealed_sector_cid(
+        &self,
+        _proof: RegisteredProof,
+        _pieces: &[PieceInfo],
+    ) -> i64 {
+        self.compute_unsealed_sector_cid_base
```
dutterbutter, line 141: Nit: add a TODO here

vm/src/gas/price_list.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+    /// Returns gas required for seal verification
+    #[inline]
+    pub fn on_verify_seal(&self, _info: &SealVerifyInfo) -> i64 {
```
dutterbutter, line 145: Nit: add a TODO here

vm/src/gas/price_list.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.verify_seal_base
+    }
+    /// Returns gas required for PoSt verification
+    #[inline]
+    pub fn on_verify_post(&self, _info: &PoStVerifyInfo) -> i64 {
```
dutterbutter, line 150: Nit: add a TODO here

vm/src/gas/price_list.rs
```diff
@@ -0,0 +1,192 @@
...
+    hashing_per_byte: 2,
+    compute_unsealed_sector_cid_base: 100,
+    verify_seal_base: 2000,
+    verify_post_base: 700,
+    verify_consensus_fault: 10,
+};
+
+/// Returns gas price list by Epoch for gas consumption
+pub fn price_list_by_epoch(_epoch: ChainEpoch) -> PriceList {
+    // In future will match on epoch and select matching price lists when config options allowed
```
dutterbutter, line 190: Nit add TODO.

vm/default_runtime/src/lib.rs
austinabell: private method but sure I'll add one

vm/src/gas/price_list.rs
austinabell: nah, not needed, they are using a map to retrieve the function, we are just matching on all signature types, impossible for us to hit this

vm/src/gas/price_list.rs
austinabell: There is nothing else TODO for all of these, but I will add a comment to revisit as this probably will change in future

dutterbutter: I figured a TODO must be required since we are not using the params?

vm/src/gas/price_list.rs
austinabell: There is nothing to do here, the protocol is not supporting different price tables yet (unless I am mistaken), this is just a feature

Issue #301, opened 2020-03-25
verify signature syscall


austinabell:
Worth scoping the proofs library for their signature verification, but could use the one in our crypto crate.

PR #351, opened 2020-04-15
Implement verify signature syscall and cleanup

austinabell:
**Summary of changes**
- Implements verify sig syscall #301 
  - Removes duplicate error when syscalls are called in actors
    - (I don't actually think the out of gas exit code can be hit within the actors and maybe that is intentional)
- Crypto crate cleanup
  - A lot was setup to try to match spec 1:1 and old implementation when possible, so now more closely matches impls and cleaner interface in rust
  - Signature verification now also returns result instead of just a boolean, for better error reporting

Closes 
#301

### DIFF
--- a/blockchain/blocks/src/header.rs
+++ b/blockchain/blocks/src/header.rs
@@ -5,7 +5,7 @@ use super::{EPostProof, Error, FullTipset, Ticket, TipSetKeys};
 use address::Address;
 use cid::{multihash::Blake2b256, Cid};
 use clock::ChainEpoch;
-use crypto::{is_valid_signature, Signature};
+use crypto::Signature;
 use derive_builder::Builder;
 use encoding::{
     de::{self, Deserializer},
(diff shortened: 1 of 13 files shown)


blockchain/chain_sync/src/sync.rs
```diff
@@ -574,14 +573,12 @@ where
...
         if header.signature().is_none() {
-            return Err(Error::Validation("Signature is nil in header"));
+            return Err(Error::Validation("Signature is nil in header".to_owned()));
         }
 
         let base_tipset = self.load_fts(&TipSetKeys::new(header.parents().cids.clone()))?;
-        println!("base tipset {:?}", base_tipset);
         let parent_tipset = base_tipset.tipset()?;
 
-        println!("BEFORE TIME STAMP VALIDATIONS");
```
dutterbutter, line 584: hehehe

crypto/src/signature.rs
```diff
@@ -102,40 +102,53 @@ impl Signature {
...
+
+        // BLS verify hash against key
+        if verify(&sig, &[hashed], &[pk]) {
+            Ok(())
+        } else {
+            Err("bls signature verification failed".to_owned())
+        }
+    }
+
+    /// Returns true if a secp256k1 signature is valid
```
dutterbutter, line 135: Misleading comment as it does not return `bool`?

blockchain/blocks/src/header.rs
```diff
@@ -268,11 +268,11 @@ impl BlockHeader {
         let signature = self
             .signature()
             .as_ref()
-            .ok_or(Error::InvalidSignature("Signature is nil in header"))?;
+            .ok_or_else(|| Error::InvalidSignature("Signature is nil in header".to_owned()))?;
```
ec2, line 271: There's nothing more beautiful in the world than a big fat method chain

austinabell: ????????

PR #353, opened 2020-04-15
Implements apply_message and apply_implicit_message

### LINKED ISSUE #340: Complete apply messages in VM
**Task summary**

Need to complete the remaining logic for `apply_message` in the VM. This is used for valid state transitions. 

Additionally, `apply_tipset_messages` logic to manage message ordering needs to be completed.

**Specification reference**

- [Apply Message definition](https://filecoin-project.github.io/specs/#vminterpreter-implementation)
- [Lotus ApplyMessage implementation](https://github.com/filecoin-project/lotus/blob/master/chain/vm/vm.go#L445)

dutterbutter:
**Summary of changes**
- Added missing logic for `apply_message`
- Added `apply_implicit_message`
- Added/updated transfer gas functions
- Modified `send` to return a tuple 
- Added `syscalls` to VM struct

I noticed in the spec they are applying a `network_tx_fee` to `total_cost` which is checked against the `from_actors` balance to see if there are sufficient funds. I added a TODO for this, however, I am not sure what that `network_tx_fee` is comprised of. 

Additionally, I added `apply_implicit_message` as it exists in the lotus implementation, but I do not see `apply_tipset_messages` in the implementation anymore. I did not remove it from our implementation as it seems that it will be required?

Closes #340 

- [Lotus-testnet/3 ApplyMessage reference](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/vm/vm.go#L260)
- [Spec reference](https://filecoin-project.github.io/specs/#vminterpreter-implementation)


vm/default_runtime/src/lib.rs
```diff
@@ -531,33 +531,64 @@ fn transfer<ST: StateTree>(
...
+        return Err("not enough funds".to_owned());
+    }
+    act.balance -= amt;
+    Ok(())
+}
+/// Deposits funds to an Actor
+fn deposit_funds(act: &mut ActorState, amt: &TokenAmount) -> Result<(), String> {
+    act.balance += amt;
+    Ok(())
+}
```
austinabell, line 594:
```suggestion
fn deposit_funds(act: &mut ActorState, amt: &TokenAmount) {
    act.balance += amt;
}
```

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,318 @@
...
-    ) -> Result<MessageReceipt, String> {
-        check_message(&msg)?;
+    /// Returns ApplyRet structure which contains the message receipt and some meta data.
+    fn apply_message(&mut self, msg: &UnsignedMessage) -> Result<ApplyRet, String> {
+        check_message(msg)?;
 
-        let snapshot = self.state.snapshot()?;
+        let pl = price_list_by_epoch(*self.epoch());
+        let ser_msg = &msg.marshal_cbor().map_err(|e| e.to_string())?;
+        let msg_gas_cost = pl.on_chain_message(ser_msg.len() as i64);
```
austinabell, line 107:
```suggestion
        let msg_gas_cost = pl.on_chain_message(ser_msg.len()) as u64;
```
Also might make sense to convert to u64 once, instead of doing the conversion throughout this function?

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,318 @@
...
+                    return Err(format!("Revert state failed: {}", state_err));
+                };
+            }
+            warn!("Send actor error: from:{}, to:{}", msg.from(), msg.to());
+        }
+        // TODO ask if this is correct as this could give a free tx if gas used was a negative value(e.g. delete_miner)
+        // see: https://github.com/filecoin-project/lotus/blob/testnet/3/chain/vm/vm.go#L388
+        if gas_used < 0 {
+            gas_used = 0;
+        }
```
austinabell, line 221: consider maybe something like
```suggestion
        let gas_used = if gas_used < 0 { 0 } else { gas_used as u64 };
```
so you don't have to do a bunch of conversions below

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,318 @@
...
-                    return state_err;
-                }
-                e.to_string()
-            })?
+                    return Err(format!("Revert state failed: {}", state_err));
+                };
+            }
+            warn!("Send actor error: from:{}, to:{}", msg.from(), msg.to());
+        }
+        // TODO ask if this is correct as this could give a free tx if gas used was a negative value(e.g. delete_miner)
```
austinabell, line 217: I think the free tx isn't a problem, just want to confirm that the actor should not be refunded if they don't spend the delete actor refund gas

vm/default_runtime/src/lib.rs
```diff
@@ -531,33 +531,63 @@ fn transfer<ST: StateTree>(
...
-        act.balance += amt;
+    if value < &0u8.into() {
+        return Err("Attempted to transfer negative value to gas holder".to_owned());
+    }
+    state.mutate_actor(addr, |act| {
+        deduct_funds(gas_holder, &value)?;
+        deposit_funds(act, &value);
         Ok(())
     })
 }
```
austinabell, line 580: I personally do not like the idea of having these functions for this, doesn't make any sense why you would create an Actor state and these extra functions just to track the gas moved. Changes made here for reference https://github.com/ChainSafe/forest/commit/29416e6037dd9bfc6e98351a8ebb8b662d2a5a5f

Definitely more opinionated so happy to chat through a solution, but I don't like following lotus here

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,318 @@
...
-
+use num_traits::Zero;
+use runtime::Syscalls;
+use vm::{price_list_by_epoch, ActorError, ExitCode, Serialized, StateTree};
 /// Interpreter which handles execution of state transitioning messages and returns receipts
 /// from the vm execution.
-pub struct VM<'a, ST: StateTree, DB: BlockStore> {
+pub struct VM<'a, ST: StateTree, DB: BlockStore, SYS: Syscalls>
+where
+    SYS: Copy,
```
austinabell, line 23: just move the syscalls trait definition below or put both above, this is hard to follow (trait definitions in the where seem to make more sense to me personally)

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,318 @@
...
+    }
+
     /// Apply all messages from a tipset
     /// Returns the receipts from the transactions.
     pub fn apply_tip_set_messages(
         &mut self,
         _tipset: &Tipset,
         msgs: &TipSetMessages,
-    ) -> Result<Vec<MessageReceipt>, String> {
+    ) -> Result<Vec<ApplyRet>, String> {
```
austinabell, line 56: Can you reference to why this extra data is needed to be returned?

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,318 @@
...
+        let ser = match internal_send(&mut rt, msg, gas_cost) {
+            Ok(ser) => ser,
+            Err(actor_err) => return (Serialized::default(), rt, Some(actor_err)),
+        };
+        (ser, rt, None)
+    }
+}
+
+// TODO remove allow dead_code
+#[allow(dead_code)]
```
austinabell, line 280: Remove and add a doc comment? or is still dead code because fields are not read?

vm/src/actor_state.rs
```diff
@@ -48,6 +50,17 @@ impl<'de> Deserialize<'de> for ActorState {
...
+impl Default for ActorState {
+    fn default() -> Self {
+        Self {
+            code: Cid::default(),
+            state: Cid::default(),
+            balance: BigUint::zero(),
+            sequence: 0,
+        }
+    }
+}
```
austinabell, line 62: I don't personally like having the actor state have default because it would allow the "undefined" cids to potentially be used without noticing. If previous suggestion is done, there isn't a need for this

vm/default_runtime/src/lib.rs
```diff
@@ -531,33 +531,63 @@ fn transfer<ST: StateTree>(
...
+    if &act.balance < amt {
+        return Err("not enough funds".to_owned());
+    }
+    act.balance -= amt;
+    Ok(())
+}
+/// Deposits funds to an Actor
+fn deposit_funds(act: &mut ActorState, amt: &TokenAmount) {
+    act.balance += amt;
+}
```
austinabell, line 593: Almost seems like these functions should just be an implementation on ActorState, thoughts?

dutterbutter: Added them to ActorState, I think its better but review and lmk.

vm/interpreter/src/lib.rs
dutterbutter: Still, fields are not read.

vm/interpreter/src/lib.rs
dutterbutter: The `penalty` and `actor_err` are being used in state manager component, for example in [ApplyBlocks](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/stmgr/stmgr.go#L170).

If it's determined that these fields can be derived elsewhere or they are not needed in state manager I would be happy to switch back to returning `MessageReceipt`, would look much cleaner.

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,319 @@
...
+use runtime::Syscalls;
+use vm::{price_list_by_epoch, ActorError, ExitCode, Serialized, StateTree};
 /// Interpreter which handles execution of state transitioning messages and returns receipts
 /// from the vm execution.
-pub struct VM<'a, ST: StateTree, DB: BlockStore> {
+pub struct VM<'a, ST, DB, SYS>
+where
+    ST: StateTree,
+    DB: BlockStore,
+    SYS: Syscalls + Copy,
```
austinabell, line 23: Sorry I missed this on a previous pass, but none of this is necessary (better to just constrain the generics when they are actually needed, not at the highest level)

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,319 @@
...
             store,
             epoch,
+            syscalls,
         }
     }
 
+    /// Returns ChainEpoch
+    fn epoch(&self) -> &ChainEpoch {
+        &self.epoch
+    }
```
austinabell, line 50:
```suggestion
    /// Returns ChainEpoch
    fn epoch(&self) -> ChainEpoch {
        self.epoch
    }
```
ChainEpoch is just a u64, any reason specifically for returning a reference?

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,319 @@
...
-                e.to_string()
-            })?
+                    return Err(format!("Revert state failed: {}", state_err));
+                };
+            }
+            warn!("Send actor error: from:{}, to:{}", msg.from(), msg.to());
+        }
+        let gas_used = if gas_used < 0 { 0 } else { gas_used as u64 };
+        // refund unused gas
+        let refund = (msg.gas_limit().checked_sub(gas_used).unwrap_or(1u64)) * msg.gas_price();
```
austinabell, line 223:
```suggestion
        let refund = (msg.gas_limit().checked_sub(gas_used).unwrap_or(0u64)) * msg.gas_price();
```
Is checked sub actually necessary here? Don't we have a guarantee that gas used cannot be greater than the gas limit?

vm/interpreter/src/lib.rs
dutterbutter: Not just a habit. I will remove.

vm/interpreter/src/lib.rs
dutterbutter: I cannot find this guarantee anywhere. Do you have a reference?

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,314 @@
...
             for msg in &block.secp_messages {
-                receipts.push(self.apply_message(&msg.message(), &block.miner)?);
+                receipts.push(self.apply_message(msg.message())?);
             }
         }
 
         Ok(receipts)
     }
 
+    fn _apply_implicit_message(&mut self, msg: &UnsignedMessage) -> Result<ApplyRet, String> {
```
ec2, line 71: I'm confused as to why this is a Result. You already put the error into the ApplyRet struct and never return an Err via Result

vm/interpreter/src/lib.rs
```diff
@@ -1,130 +1,314 @@
...
     /// Instantiates a new Runtime, and calls internal_send to do the execution.
-    fn send(&mut self, msg: &UnsignedMessage, gas_cost: i64) -> Result<Serialized, ActorError> {
+    fn send<'m>(
+        &mut self,
+        msg: &'m UnsignedMessage,
+        gas_cost: i64,
+    ) -> (
+        Serialized,
+        DefaultRuntime<'_, 'm, '_, ST, DB, SYS>,
+        Option<ActorError>,
```
ec2, line 252: Don't understand why this is an option? If internal_send is successful, we can return ExitCode::Ok, otherwise return actor_err, no?

dutterbutter: It seems weird to construct an error upon a successful call. More preference I suppose.

Issue #341, opened 2020-04-10
Update sector types for protocol change


austinabell:
**Issue summary**

- This issue should probably not be done now and wait for other changes, just seemed significant enough to open an issue to revisit so it doesn't get missed.

changes: https://github.com/filecoin-project/specs-actors/commit/e569f4a2f50403eeaef22c410d36d2ed30c57e37

**Other information and links**

PR #357, opened 2020-04-17
Update to new PoSt sector types

austinabell:
**Summary of changes**
- Based on changes in https://github.com/filecoin-project/specs-actors/commit/e569f4a2f50403eeaef22c410d36d2ed30c57e37
  - Needed for updates to PoSt proofs in future, these changes target ahead of testnet/3, but will reduce amount of duplicate work
  - Changes are needed for the proof work I'm doing now to remove this duplicate work (no point targeting the current testnet)

Closes 
#341 

### DIFF
--- a/vm/src/sector/registered_proof.rs
+++ b/vm/src/sector/registered_proof.rs
@@ -11,13 +11,25 @@ use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 #[repr(u8)]
 pub enum RegisteredProof {
     StackedDRG32GiBSeal = 1,
-    StackedDRG32GiBPoSt = 2,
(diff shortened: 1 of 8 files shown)


vm/src/sector/registered_proof.rs
```diff
@@ -11,13 +11,25 @@ use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 #[repr(u8)]
 pub enum RegisteredProof {
     StackedDRG32GiBSeal = 1,
-    StackedDRG32GiBPoSt = 2,
+    StackedDRG32GiBPoSt = 2, // unused (revisit if being removed)
```
ec2, line 14: Maybe add "TODO" to the comment. Makes it easier to search for all TODOs when we need to

Issue #327, opened 2020-04-01
Implement change in heaviest tipset in ChainStore


dutterbutter:
**Task summary**

On start, we need to be able to `Load()` the `ChainStore` and pull the `heaviest_tipset` from storage and set the heaviest field in the ChainStore struct. Further, we need the logic required to make the determination of the `heaviest_tipset` during the **catchup** and **follow** sync stages and then store it accordingly under the `chain_head_key`. 

**Specification reference**

- [Lotus Load function (where we set heaviest tipset for ChainStore structure)](https://github.com/filecoin-project/lotus/blob/master/chain/store/store.go#L118)
- [Lotus write heaviest tipset to storage](https://github.com/filecoin-project/lotus/blob/master/chain/store/store.go#L143)
- [Lotus heaviest determination logic](https://github.com/filecoin-project/lotus/blob/master/chain/store/store.go#L243)

This is required in order to successfully sync.

**Other information and links**

dutterbutter:
Closed with #359

PR #359, opened 2020-04-17
Implement Weight method + st/gt for heaviest ts

### LINKED ISSUE #118: Implement Chain Selection
Chain Selection is an integral part of EC but needs to be accessible within the `ChainManager` and have the heaviest tipset value available in the `Store` struct. As per the spec chain selection definition can be found [here](https://filecoin-project.github.io/specs/#algorithms__expected_consensus__chain_selection) and may look like the following:

```
// ChainSelector chooses the heaviest between chains.
type ChainSelector interface {
	// IsHeavier returns true if tipset a is heavier than tipset b and false if
	// tipset b is heavier than tipset a.
	IsHeavier(ctx context.Context, a, b block.TipSet, aStateID, bStateID cid.Cid) (bool, error)
	// Weight returns the weight of a tipset after the upgrade to version 1
	Weight(ctx context.Context, ts block.TipSet, stRoot cid.Cid) (uint64, error)
}
```

Where the algorithm for determining the **weight** can be found:
- [Spec Definition ](https://filecoin-project.github.io/specs/#chain-weighting)
- [Go-Filecoin Weight Function](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/consensus/chain_selector.go#L49)
- [Lotus Weight Function](https://github.com/filecoin-project/lotus/blob/master/chain/store/weight.go#L16)

Where the algorithm for determining the **IsHeavier** can be found:
- [Spec Definition ](https://filecoin-project.github.io/specs/#chain-weighting)
- [Go-Filecoin IsHeavier Function](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/consensus/chain_selector.go#L97)
- [Lotus Heaviest Function](https://github.com/filecoin-project/lotus/blob/master/chain/sync_manager.go#L174)

**Tasks Required for completion:**
- [ ] `ChainSelection` trait
- [ ] `Weight` fn
- [ ] `Heaviest` fn

dutterbutter:
**Summary of changes**
- Implements the` Weight` method
- Adds `load_heaviest_tipset` to be used when starting the node
- Adds `set_heaviest_tipset`
- Changes `StoragePower` to BigUint

Closes #118 #327 

- [Weight fn calculation](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/store/weight.go#L20)


blockchain/blocks/src/tipset.rs
```diff
@@ -204,6 +205,32 @@ impl Tipset {
...
+impl<'de> de::Deserialize<'de> for Tipset {
+    fn deserialize<D>(deserializer: D) -> Result<Self, <D as Deserializer<'de>>::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let (blocks, key) = Deserialize::deserialize(deserializer)?;
+
+        Ok(Tipset { blocks, key })
+    }
+}
```
austinabell, line 232: This will create two different impls, the only reason marshal_cbor allows for overriding is for caching reasons, having two different impls will cause an issue later for sure

blockchain/chain/src/store/chain_store.rs
```diff
@@ -59,21 +71,16 @@ where
...
     }
-    /// weight
-    pub fn weight(&self, _ts: &Tipset) -> Result<BigUint, Error> {
-        // TODO
-        Ok(BigUint::from(0 as u32))
-    }
 
     /// Writes genesis to blockstore
-    pub fn set_genesis(&self, header: BlockHeader) -> Result<(), Error> {
+    pub fn set_genesis(&mut self, header: BlockHeader) -> Result<(), Error> {
```
austinabell, line 76: why do you need to change to mutable?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -210,6 +235,51 @@ where
...
+        let curr_weight = self.weight(&self.heaviest)?;
+
+        if new_weight > curr_weight {
+            // TODO potentially need to deal with re-orgs here
+            info!("New heaviest tipset");
+            self.set_heaviest_tipset(Arc::new(ts.clone()))?;
+        }
+
+        Ok(())
+    }
```
austinabell, line 251: Maybe change the name of this to update_heaviest() or something, confusing the actual function of this

blockchain/chain_sync/src/sync.rs
```diff
@@ -780,7 +780,7 @@ where
     }
 
     /// Persists headers from tipset slice to chain store
-    fn persist_headers(&self, tipsets: &[Tipset]) -> Result<(), Error> {
+    fn persist_headers(&mut self, tipsets: &[Tipset]) -> Result<(), Error> {
```
austinabell, line 783: why mut?

blockchain/chain/src/store/chain_store.rs
dutterbutter: Because `persist_headers` needs a mutable reference.

blockchain/chain_sync/src/sync.rs
dutterbutter: Because `ChainStore::persist_headers` needs a mutable reference.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -210,31 +235,49 @@ where
...
-            .messages(Cid::new_from_cbor(&[], Identity))
-            .message_receipts(Cid::new_from_cbor(&[], Identity))
-            .state_root(Cid::new_from_cbor(&[], Identity))
-            .miner_address(Address::new_id(0).unwrap())
-            .build_and_validate()
-            .unwrap();
-
-        assert_eq!(cs.genesis().unwrap(), None);
-        cs.set_genesis(gen_block.clone()).unwrap();
-        assert_eq!(cs.genesis().unwrap(), Some(gen_block));
```
austinabell, line 238: Can you explain why exactly this test needs to be removed? Seems like at a high level that function should be able to be called without anything else, why can this not happen?

dutterbutter: The test needs more setup as it fails during the `weight()` method called from `update_heaviest()`.  I could follow lotus more precisely and have a `put_tipset()` wrapper over `persist_headers()` where `update_heaviest()` would live and have `set_genesis()` only call `persist_headers()` since it only takes in a blockheader anyway. Otherwise, I will refactor the test to account for new required setup.

austinabell: what is the setup that's missing though? that genesis block not existing in the blockstore? seems intuitive that this should not require other setup, but can look more into specifics later

dutterbutter: No, the setup that's missing is `STORAGE_POWER_ACTOR_ADDR` not being available in the `weight` method. I am adding a `put_tipset` method, similar to lotus; see [here](https://github.com/filecoin-project/lotus/blob/master/chain/store/store.go#L224). As it seems we should not be calling `update_heaviest()` upon initializing the node with genesis.

PR #360, opened 2020-04-19
Compute unsealed sector CID syscall

austinabell:
**Summary of changes**

- Implements the compute_unsealed_sector_cid syscall in the runtime
  - Interfaces through the filecoin-proofs-api
- Updates commcid interface to take in references (was no need to move variables before, but assumed they wouldn't be needed after)

Dependent on #357 so I will remove other commits when that comes into mater and open this PR up

Closes 
#303 


vm/Cargo.toml
```diff
@@ -15,3 +15,9 @@ num-derive = "0.3.0"
 clock = { path = "../node/clock" }
 thiserror = "1.0.11"
 crypto = { path = "../crypto" }
+commcid = { path = "../utils/commcid" }
+
+[dependencies.filecoin-proofs-api]
```
ec2, line 20: Why the departure of our standard way of declaring dependencies?

ec2: Also, don't understand why we need to specify a version if we already specify a rev.

vm/runtime/Cargo.toml
```diff
@@ -12,4 +12,10 @@ message = { package = "forest_message", path = "../message" }
 cid = { package = "forest_cid", path = "../../ipld/cid" }
 ipld_blockstore = { path = "../../ipld/blockstore" }
 clock = { path = "../../node/clock" }
-forest_encoding = { path = "../../encoding"}
\ No newline at end of file
+forest_encoding = { path = "../../encoding" }
+commcid = { path = "../../utils/commcid" }
+
+[dependencies.filecoin-proofs-api]
```
ec2, line 18: Same as the comment in vm/Cargo.toml

vm/runtime/src/lib.rs
```diff
@@ -150,11 +153,42 @@ pub trait Syscalls {
...
-        // TODO
-        todo!()
+        let mut sum = 0u64;
+        for p in pieces {
+            sum += p.size.0;
+        }
+
+        let ssize = proof_type.sector_size() as u64;
+
+        let mut fcp_pieces: Vec<filecoin_proofs_api::PieceInfo> = pieces
```
ec2, line 166:
```suggestion
        let mut fcp_pieces: Vec<PieceInfo> = pieces
```

ec2: Oops, Nvm... Apparently there are two different PieceInfos....

vm/runtime/src/lib.rs
```diff
@@ -150,11 +153,42 @@ pub trait Syscalls {
...
+            .iter()
+            .map(filecoin_proofs_api::PieceInfo::try_from)
+            .collect::<Result<_, &'static str>>()
```
ec2, line 169:
```suggestion
                        .collect::<Result<Vec<filecoin_proofs_api::PieceInfo>, &'static str>>()
```
Then you can remove the type annotation on fcp_pieces. This is a little more understandable to me, BUT I could be in the minority... Up to you :)

ec2:
line 23 in vm/src/piece/mod.rs
`pub fn validate(self) -> Result<(), &’static str> {`
Can probably take a ref. I couldnt comment on that line so posting here

vm/Cargo.toml
austinabell: just for safety, and the reason I used this format is just to make updating commit easier and more visible, I'll change to other format tho

vm/runtime/src/lib.rs
austinabell: Idk I thought this way is more readible because it's more clear what comes out of it

austinabell:
> line 23 in vm/src/piece/mod.rs
> `pub fn validate(self) -> Result<(), &’static str> {`
> Can probably take a ref. I couldnt comment on that line so posting here

Nah it's just a u64 so better to just move in. Pretty sure clippy warns on this

vm/runtime/src/lib.rs
ec2: That sounds pretty logical!

PR #369, opened 2020-04-21
Stmgr retrieval methods + is_ticket_winner calc for block validation

### LINKED ISSUE #361: Ticket winner determination
**Task summary**

The calculation for determining if a ticket is a winner. Used for block validation and consensus.

**Specification reference**

- [Lotus isTicketWinner reference](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/types/blockheader.go#L191)

In our codebase, this method will be used during the syncing process as its part of block validation. Specifically, it will likely be called from the [ChainSyncer component](https://github.com/ChainSafe/forest/blob/master/blockchain/chain_sync/src/sync.rs#L570) and should probably live as a method under the [`BlockHeader component`](https://github.com/ChainSafe/forest/blob/master/blockchain/blocks/src/header.rs#L50).

### LINKED ISSUE #362: Retrieve network raw power for block validation
**Task summary**

Determining if a ticket is a winner requires the `claimed power` and `total network power`. We need the ability to retrieve these values during block validation to be used for determining a ticket winner. 

**Specification reference**

- [Lotus GetPower reference](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/stmgr/utils.go#L80)
- [Where its used in block validation](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/sync.go#L613)

In our codebase, the retrieval of these two power params would likely be called from the [`Chainsyncer component`](https://github.com/ChainSafe/forest/blob/master/blockchain/chain_sync/src/sync.rs#L570) during the validation process. And would likely live under the StateManager component.

### LINKED ISSUE #365: Retrieve raw miner address for block validation
**Task summary**

The winner check and block signature check within block validation require the raw miner work address. Specifically, the raw miner address is used a parameter for verifying VRF and verifying the block signatures.
 
**Specification reference**

- [Lotus GetMinerRawAddress reference ](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/stmgr/utils.go#L44)
- [Lotus block validation GetMinerRawAddress usage](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/sync.go#L578)

In our codebase, we would be utilizing the raw miner address during the block validation process in the [`ChainSyncer component here`](https://github.com/ChainSafe/forest/blob/master/blockchain/chain_sync/src/sync.rs#L591).

dutterbutter:
**Summary of changes**
- Added state retrieval methods to `StateManager` component which are used during block validation in `ChainSyncer`
- Added `is_ticket_winner` calc to `BlockHeader`  which is used during block validation in `ChainSyncer`
- Moved `resolve_to_key_addr` outside of `DefaultRuntime`

Closes #361 Closes #362 Closes #365 

- [is_ticket_winner calc reference](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/types/blockheader.go#L191)


blockchain/blocks/src/header.rs
```diff
@@ -301,6 +305,31 @@ impl BlockHeader {
...
+        let mut lhs = BigUint::from_bytes_le(&h);
+        lhs *= net_pow;
+
+        // rhs = sectorSize * 2^256
+        // rhs = sectorSize << 256
+        let mut rhs = mpow << SHA_256_BITS;
+        rhs *= BigUint::from(BLOCKS_PER_EPOCH);
+
+        // h(vrfout) * totalPower < e * sectorSize * 2^256
+        lhs.cmp(&rhs) < Ordering::Equal
```
austinabell, line 331:
```suggestion
        lhs < rhs
```
lol

blockchain/chain_sync/src/sync.rs
```diff
@@ -585,23 +584,32 @@ where
         // check messages to ensure valid state transitions
         self.check_blk_msgs(block.clone(), &parent_tipset)?;
 
+        // TODO use state_root instead of parent_tipset.parent_state()
```
austinabell, line 587: can you explain this?

blockchain/state_manager/src/lib.rs
```diff
@@ -41,23 +42,45 @@ where
...
+            Some(_) => Ok(false),
+            None => Ok(true),
+        }
     }
-    pub fn get_actor(&self, addr: &Address, ts: &Tipset) -> Result<Option<ActorState>, Error> {
-        let state = HamtStateTree::new_from_root(self.bs.as_ref(), ts.parent_state())
-            .map_err(Error::State)?;
-        state.get_actor(addr).map_err(Error::State)
+    /// Returns raw address of a miner
+    pub fn get_miner_wrk_addr(&self, state_cid: &Cid, addr: &Address) -> Result<Address, Error> {
```
austinabell, line 64: can we not use short form for at least `work`? I don't see the need for this to be that shortened

blockchain/state_manager/src/lib.rs
```diff
@@ -41,23 +42,45 @@ where
...
+        let state =
+            HamtStateTree::new_from_root(self.bs.as_ref(), state_cid).map_err(Error::State)?;
+        // Note: miner::State info likely to be changed to CID
+        let addr = resolve_to_key_addr(&state, self.bs.as_ref(), &ms.info.worker)
+            .map_err(|e| Error::Other(format!("Failed to resolve key address; error: {}", e)))?;
+        Ok(addr)
+    }
+    /// Returns specified actor's claimed power and total network power as a tuple
+    pub fn get_power(&self, state_cid: &Cid, addr: &Address) -> Result<(BigUint, BigUint), Error> {
+        let ps: power::State = self.load_actor_state(addr, state_cid)?;
```
austinabell, line 76: Almost certain you want to use the builtin power actor address here, pretty sure this will never be pulled from an address being passed in? Might want to double check this (miner address state will not be in the form of a power actor state)

vm/src/error.rs
```diff
@@ -27,10 +27,10 @@ impl ActorError {
         }
     }
 
-    pub fn new_fatal(msg: String) -> Self {
+    pub fn new_fatal(exit_code: ExitCode, msg: String) -> Self {
```
austinabell, line 30:
```suggestion
    pub fn new_fatal(msg: String) -> Self {
```
Exit code should never matter for a fatal error, and there would be nothing to match against the go client here
(I think we will move to this exit code being optional very soon)

blockchain/chain_sync/src/sync.rs
dutterbutter: Rather than using the parent state root, use the computed state root returned (when its implemented) [here](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/stmgr/stmgr.go#L63:25).

blockchain/blocks/src/header.rs
ec2: 🤔

blockchain/state_manager/src/lib.rs
dutterbutter: lol just to clarify, you do not want the use of short-form verbiage? I have no objections to that and I am happy to change. However, we do have many function signatures that have similar short-form usages so we should probably enforce it across the board?

blockchain/blocks/src/header.rs
dutterbutter: haha I always like doing things the hard way 🙃

blockchain/state_manager/src/lib.rs
ec2: I think what @austinabell means here specfically is that were only shortening it by 1 char. Doesn't make much sense to shorten it. Might as well shorten miner to mnr 😂  go full startup company name (dropping all the vowels)

dutterbutter: We're trendy haha

austinabell: yeah guess I meant to prefix the comment with optional, was just confusing for me to read

vm/default_runtime/src/lib.rs
```diff
@@ -546,3 +513,54 @@ fn transfer<ST: StateTree>(
 
     Ok(())
 }
+
+/// Returns public address of the specified actor address
+pub fn resolve_to_key_addr<'st, 'bs, ST, BS>(
+    st: &'st ST,
+    store: &'bs BS,
```
austinabell, line 520: Just a thought: Seems weird that there needs to be two references to the blockstore here, I guess way around this is to not have the state tree generic and use HamtStateTree specifically which isn't necessarily better and this is probably better for future proofing, but just figured I'd mention

vm/default_runtime/src/lib.rs
```diff
@@ -546,3 +513,54 @@ fn transfer<ST: StateTree>(
...
+        .map_err(|e| ActorError::new(ExitCode::SysErrInternal, e))?
+    {
+        Some(act) => act,
+        None => {
+            return Err(ActorError::new(
+                ExitCode::SysErrInternal,
+                format!("Failed to retrieve actor: {}", addr),
+            ))
+        }
+    };
```
austinabell, line 542: can you change this to be `.ok_or_else(|| ActorError::new( ...` instead of matching (as you have below), could make the message more descriptive that the actor was not found (optional though, this is fine)

blockchain/chain_sync/src/sync.rs
```diff
@@ -613,23 +612,32 @@ where
         // check messages to ensure valid state transitions
         self.check_blk_msgs(block.clone(), &parent_tipset)?;
 
+        // TODO use computed state_root instead of parent_tipset.parent_state()
+        let wrk_addr = self
```
ec2, line 616: Hey dstn 😂

Issue #363, opened 2020-04-20
Apply Blocks from StateManager


dutterbutter:
**Task summary**

ApplyBlocks is used to update the state to reflect validated inbound blocks. It looks to me that all the VM pieces required to complete this task are ready.

**Specification reference**

- [Lotus ApplyBlocks reference](https://github.com/filecoin-project/lotus/blob/testnet/3/chain/stmgr/stmgr.go#L151)

**Other information and links**

Issue #373, opened 2020-04-24
Update StateTree to not flush on snapshot


austinabell:
**Issue summary**

- State tree was changed recently to have a similar snapshotting/ state change mechanism as geth, Making this change on our end will probably be worth it for when we are using a purely persisted blockstore for this. Currently, snapshotting is done by flushing the tree and returning the Cid root, but this does have the cost of having to flush the hamt on each snapshot. This shouldn't matter much for the end data stored as the data transfered from a memory store to cold storage will eventually be done based on the root (no unnecessary writes) so this seems a bit redundant anyway.

I don't really see the need as long as we are using an in memory store while interacting with the state tree, but to make sure of no discrepancies as well, this should probably be changed.

I am refactoring the state tree as is in current changes, and will link that PR when I open as this issue will be blocked on that coming in.

**Other information and links**

PR #374, opened 2020-04-24
Apply Blocks and refactor

austinabell:
**Summary of changes**
- Implements apply blocks method from syncer. #363 
  - Main logic is moved to inside of the VM since applying the tipset messages made sense to not be attached to the syncer
- Setup buffered blockstore
  - As to not commit all changes to the underlying store (especially in the case of a failure) and will be expanded in future to limit writes based on what's connected to the root

Refactor and cleanup:
- Replaced state tree trait with just the structure itself
  - There wasn't a great reason why this was the way it is other than future proofing and there is no need to swap out implementation for testing
  - There was a bunch of changes in state tree, so I opened issue detailing potential changes #373 
- Refactored the conversions between Tipset and FullTipset
  - Isn't ideal the way it's setup now, so I'm hoping to make a dedicated pass through since this will be used a lot and still has unnecessary clones
- Update and fix how peer heads are managed (was a bug previously to only use the tipset passed in the function and the tipset was never connected to the respective peer)
  - Still not in an ideal place, but since this will be replaced with GraphSync I won't refactor it too much (only enough to test)

Closes 
#363 


vm/interpreter/src/lib.rs
```diff
@@ -49,35 +65,117 @@ where
...
         }
 
+        // TODO same as above, unnecessary state retrieval
+        let sys_act = self
+            .state
+            .get_actor(&*SYSTEM_ACTOR_ADDR)?
+            .ok_or_else(|| "Failed to query system actor".to_string())?;
+
+        let cron_msg = UnsignedMessage::builder()
+            .to(SYSTEM_ACTOR_ADDR.clone())
```
dutterbutter, line 151: Shouldn't this be to: `CRON_ACTOR_ADDR` from: `SYSTEM_ACTOR_ADDR`?

blockchain/blocks/src/tipset.rs
```diff
@@ -218,12 +222,28 @@ impl FullTipset {
...
+    /// Returns reference to all blocks in a full tipset
     pub fn blocks(&self) -> &[Block] {
         &self.blocks
     }
+    /// Returns all blocks in a full tipset
+    pub fn into_blocks(self) -> Vec<Block> {
+        self.blocks
+    }
+    // TODO: conversions from full to regular tipset should not return a result
+    // and should be validated on creation instead
```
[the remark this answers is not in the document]
austinabell: The `Tipset` does yeah, but the `FullTipset` is sometimes constructed without validation (why the conversion to Tipset is error prone

vm/interpreter/src/lib.rs
austinabell: Oops I swapped the order and confused myself lmao

ipld/blockstore/src/buffered.rs
```diff
@@ -0,0 +1,131 @@
...
+use super::BlockStore;
+use cid::{multihash::MultihashDigest, Cid};
+use db::{Error, Store};
+use encoding::{ser::Serialize, to_vec};
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::error::Error as StdError;
+
+/// Wrapper around `BlockStore` to limit and have control over when values are written.
+/// This type is not threadsafe and can only be used in syncronous contexts.
```
ec2, line 13: TYPO: synchronous

ipld/blockstore/src/buffered.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Debug)]
+pub struct BufferedBlockStore<'bs, BS> {
+    base: &'bs BS,
+    write: RefCell<HashMap<Cid, Vec<u8>>>,
+}
+
+impl<'bs, BS> BufferedBlockStore<'bs, BS>
+where
+    BS: BlockStore + Store,
```
ec2, line 22: Curious what are the consequences of removing the requirement of Store on BS, since BlockStore inherets the Store trait... I removed it and it works.

austinabell: functionally the same, but yeah should be more succinct, I'll fix

ec2: Cool. Just wanted to make sure i was on the same page. But yeah, removing is best

PR #385, opened 2020-04-29
BlockHeader Update

ec2:
**Summary of changes**
- Updates BlockHeader definition to match Lotus interopnet branch.

Closes 


ec2:
Not entirely sure where to put PostProof, so its in blockheader.rs for now

blockchain/beacon/src/beacon_entries.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use encoding::{
+    de::{self, Deserializer},
+    ser::{self, Serializer},
```
austinabell, line 6:
```suggestion
    de::{Serialize, Deserializer},
    ser::{Deserialize, Serializer},
```
Why prefix if only using those?

blockchain/beacon/src/beacon_entries.rs
```diff
@@ -0,0 +1,63 @@
...
+        (&self.round, serde_bytes::Bytes::new(&self.data)).serialize(serializer)
+    }
+}
+
+impl<'de> de::Deserialize<'de> for BeaconEntry {
+    fn deserialize<D>(deserializer: D) -> Result<Self, <D as Deserializer<'de>>::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let (round, data): (u64, serde_bytes::ByteBuf) = Deserialize::deserialize(deserializer)?;
```
austinabell, line 55: Optional: Could use the `BytesSer` and `BytesDe` from the same encoding package for a bit cleaner deserialization

blockchain/beacon/src/beacon_entries.rs
```diff
@@ -0,0 +1,63 @@
...
+    serde_bytes,
+};
+use serde::Deserialize;
+
+/// The result from getting an entry from Drand.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct BeaconEntry {
+    round: u64,
+    data: Vec<u8>,
+    prev_round: u64,
```
austinabell, line 16: Why have this as a variable? Not needed when prev_round can just give you this?

blockchain/blocks/src/header.rs
```diff
@@ -25,6 +26,36 @@ use std::time::{SystemTime, UNIX_EPOCH};
...
+}
+impl ser::Serialize for PostProof {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        (self.registered_proof, &self.proof_bytes).serialize(serializer)
+    }
+}
+impl<'de> de::Deserialize<'de> for PostProof {
```
austinabell, line 43: Optional: Could do the same as previous, there is no need for this in this file

blockchain/blocks/src/header.rs
```diff
@@ -25,6 +26,36 @@ use std::time::{SystemTime, UNIX_EPOCH};
 const SHA_256_BITS: usize = 256;
 const BLOCKS_PER_EPOCH: u64 = 5;
 
+/// Proof of spacetime proof
+#[derive(Eq, PartialEq, Debug, Clone, Default)]
+pub struct PostProof {
+    registered_proof: u64,
+    proof_bytes: Vec<u8>,
+}
```
austinabell, line 34: This type exists in `vm/src/sector/post.rs` Should probably use that especially for registered_proof consistency

blockchain/blocks/src/header.rs
```diff
@@ -216,6 +261,14 @@ impl BlockHeader {
     pub fn epoch(&self) -> ChainEpoch {
         self.epoch
     }
+    /// Getter for Drand BeaconEntry
+    pub fn beacon_entries(&self) -> &[BeaconEntry] {
+        &self.beacon_entries
+    }
+    /// Getter for window PoSt proof
+    pub fn win_post_proof(&self) -> &Vec<PostProof> {
```
austinabell, line 269:
```suggestion
    pub fn win_post_proof(&self) -> &[PostProof] {
```

blockchain/blocks/src/header.rs
ec2: Oh shit, nice find. I didn't see that...

blockchain/beacon/src/beacon_entries.rs
ec2: It will be used later in an upcoming PR.

dutterbutter:
Just for reference: https://github.com/filecoin-project/specs/issues/918

blockchain/beacon/src/beacon_entries.rs
austinabell: but can always just call the function for this instead of having it as a field?

PR #386, opened 2020-04-29
Implements verify consensus fault syscall and reorg vm crates

### LINKED ISSUE #306: verify_consensus_fault syscall

dutterbutter:
**Summary of changes**
- implements verify consensus fault syscall
- moves default runtime into interpreter crate due to dependency issues involving statetree and actors

Closes #306 


vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+            let bh_3: BlockHeader = from_slice(extra).map_err(|e| {
+                ActorError::new(
+                    ExitCode::ErrPlaceholder,
+                    format!("cannot decode extra {}", e.to_string()),
+                )
+            })?;
+            if bh_1.parents() != bh_3.parents()
+                && bh_1.epoch() != bh_3.epoch()
+                && contains_cid(bh_2.parents().cids(), bh_3.cid())
+                && !contains_cid(bh_2.parents().cids(), bh_1.cid())
```
austinabell, line 123: There is a function for this! Can remove the function definition below as well
```suggestion
                && bh_2.parents().cids().contains(bh_3.cid())
                && !bh_2.parents().cids().contains(bh_1.cid())
```

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+                Ok(None)
+            }
+        } else {
+            // (4) expensive final checks
+
+            // check blocks are properly signed by their respective miner
+            // note we do not need to check extra's: it is a parent to block b
+            // which itself is signed, so it was willingly included by the miner
+            self.verify_block_signature(&bh_1)?;
+            self.verify_block_signature(&bh_2)?;
```
austinabell, line 140: Technically the logic is a bit different than the go impls: if an error is hit by this but there was a consensus fault created and returned before, we would have returned the consensus fault where they would return this error. This verify signature needs to be checked before returning the consensus fault I believe, or else it would be an invalid argument in the specs-actors (I hate how it's setup but we need to match) Consider keeping a mutable optional consensus fault variable that gets updated in the places it's returned and then this verify signature occurs before it is returned

vm/runtime/src/lib.rs
```diff
@@ -146,6 +142,8 @@ pub trait Syscalls {
             .verify(plaintext, signer)
             .map_err(|e| ActorError::new(ExitCode::ErrPlaceholder, e))
     }
+    /// Verifies signatures of a block header
+    fn verify_block_signature(&self, bh: &BlockHeader) -> Result<(), ActorError>;
```
austinabell, line 146: This should not be added to the syscalls trait (Should be an implementation of just the specific implementation that uses it

vm/interpreter/src/gas_syscalls.rs
```diff
@@ -77,13 +78,17 @@ where
...
             .unwrap();
-        self.syscalls
-            .verify_consensus_fault(h1, h2, extra, earliest)
+        Ok(self
+            .syscalls
+            .verify_consensus_fault(h1, h2, extra, earliest)?)
+    }
+    fn verify_block_signature(&self, _bh: &BlockHeader) -> Result<(), ActorError> {
+        todo!()
     }
```
austinabell, line 92:
```suggestion
```

vm/interpreter/src/gas_syscalls.rs
```diff
@@ -125,12 +130,15 @@ mod tests {
...
+            Ok(Some(ConsensusFault {
                 target: Address::new_id(0),
                 epoch: 0,
                 fault_type: ConsensusFaultType::DoubleForkMining,
-            })
+            }))
+        }
+        fn verify_block_signature(&self, _bh: &BlockHeader) -> Result<(), ActorError> {
+            todo!()
         }
```
austinabell, line 142:
```suggestion
```

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+
+            Ok(None)
+        }
+    }
+    fn verify_block_signature(&self, bh: &BlockHeader) -> Result<(), ActorError> {
```
austinabell, line 145: move this out of the syscalls trait impl (only a helper function for this specific impl

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+        let state = StateTree::new_from_root(self.store, bh.state_root()).map_err(|e| {
+            ActorError::new(
+                ExitCode::ErrPlaceholder,
+                format!("cannot retrieve actor state {}", e),
+            )
+        })?;
```
austinabell, line 151: Why defect from the go impls of referencing a state tree with self and reconstructing here? I think you may run into inconsistencies with snapshotting possibly here (didn't look too in depth)

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+            .map_err(|e| {
+                ActorError::new(
+                    ExitCode::ErrPlaceholder,
+                    format!("cannot retrieve actor state {}", e),
+                )
+            })?
+            .ok_or_else(|| {
+                ActorError::new(
+                    ExitCode::ErrPlaceholder,
+                    "cannot retrieve actor state".to_owned(),
```
austinabell, line 164: error message doesn't need to match, if doing something else can be more specific that it was not found and not just some error

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+            .map_err(|e| {
+                ActorError::new(
+                    ExitCode::ErrPlaceholder,
+                    format!("cannot retrieve miner state {}", e.to_string()),
+                )
+            })?
+            .ok_or_else(|| {
+                ActorError::new(
+                    ExitCode::ErrPlaceholder,
+                    "cannot retrieve miner state".to_owned(),
```
austinabell, line 180: same as above

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+        // for that reason when checking block parent relationships, rather than instantiating a Tipset to do so
+        // (which runs a syntactic check), we do it directly on the CIDs.
+
+        // (0) cheap preliminary checks
+
+        if h1 == h2 {
+            return Err(ActorError::new(
+                ExitCode::ErrPlaceholder,
+                "no consensus fault: submitted blocks are the same".to_owned(),
+            ));
```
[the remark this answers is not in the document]
dutterbutter: Not if we are replicating lotus, they do not return a nil pointer until the initial epoch check.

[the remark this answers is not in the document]
austinabell: Oops forgot to remove this comment, yeah I checked and saw they are returning error here

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// the "parent grinding fault", in which case it must be the sibling of h1 (same parent tipset) and one of the
+    /// blocks in the parent of h2 (i.e. h2's grandparent).
+    /// Returns an error if the headers don't prove a fault.
+    fn verify_consensus_fault(
+        &self,
+        h1: &[u8],
+        h2: &[u8],
+        extra: &[u8],
+        _earliest: ChainEpoch, // unused in lotus
+    ) -> Result<Option<ConsensusFault>, ActorError> {
```
ec2, line 48: Looking at specs-actor, the have `ConsensusFaultNone = 0` commented out. I wonder if they're gunna uncomment that so that we dont need to return an Option.

austinabell: They probably just had that because they didn't want to use pointers/ nil for safety, technically better if they don't and just handle

vm/interpreter/src/default_syscalls.rs
dutterbutter: As per discussion, its mentioned as a TODO for now

vm/actor/src/builtin/paych/mod.rs
```diff
@@ -128,7 +128,13 @@ impl Actor {
...
 
         if !sv.secret_pre_image.is_empty() {
-            let hashed_secret: &[u8] = &rt.syscalls().hash_blake2b(&params.secret)?;
+            let hashed_secret: &[u8] =
+                &rt.syscalls().hash_blake2b(&params.secret).map_err(|e| {
+                    ActorError::new(
+                        ExitCode::ErrPlaceholder,
+                        format!("hashing input error: {}", e),
+                    )
+                })?;
```
austinabell, line 137: This is an example when you'd have to downcast the error returned. Don't assume a placeholder exit code, we will just run into inconsistencies

vm/actor/src/builtin/power/mod.rs
```diff
@@ -482,10 +482,15 @@ impl Actor {
...
                 ActorError::new(
                     ExitCode::ErrIllegalArgument,
-                    format!("fault not verified: {}", e.msg()),
+                    format!("fault not verified: {}", e),
+                )
+            })?
+            .ok_or_else(|| {
+                ActorError::new(
+                    ExitCode::ErrIllegalArgument,
+                    "fault not verified".to_owned(),
```
austinabell, line 491: Just noting here that we should follow up with the go impls, their implementation does not handle the nil pointer and seems like it would just panic unnecessarily

vm/actor/src/builtin/paych/mod.rs
```diff
@@ -128,7 +128,7 @@ impl Actor {
         }
 
         if !sv.secret_pre_image.is_empty() {
-            let hashed_secret: &[u8] = &rt.syscalls().hash_blake2b(&params.secret)?;
+            let hashed_secret: &[u8] = &rt.syscalls().hash_blake2b(&params.secret).unwrap();
```
austinabell, line 131: let me be more specific:
```
            let hashed_secret: &[u8] = &rt
                .syscalls()
                .hash_blake2b(&params.secret)
                .map_err(|e| *e.downcast::<ActorError>().unwrap())?;
```
Downcasting the error to an actor error, if it is not that, then panic (unwrap). This is fine because we expect if there is an error, it is an ActorError from running out of gas (or future proofing for anything else)

vm/interpreter/src/default_syscalls.rs
```diff
@@ -0,0 +1,153 @@
...
+                && bh_2.parents().cids().contains(bh_3.cid())
+                && !bh_2.parents().cids().contains(bh_1.cid())
+            {
+                cf = Some(ConsensusFault {
+                    target: *bh_1.miner_address(),
+                    epoch: bh_2.epoch(),
+                    fault_type: ConsensusFaultType::ParentGrinding,
+                })
+            }
+        };
```
ec2, line 119: In Lotus, they have a step (3) where they check at this point of consensusFault == nil. Do we not need that?

dutterbutter: That is achieved by cf being an Option

dutterbutter:
> upon further investigation, looks like you need to implement step 3

Yeah, will make changes

PR #375, opened 2020-04-24
Implement interfacing with Drand over GRPC

ec2:
**Summary of changes**
Changes introduced in this pull request:

- Allow fetching of beacon entries from a remote Drand node over GRPC
- Allow verifying Drand beacon entries 

**Reference issue to close (if applicable)**

Closes 
#367, #366 

**Other information and links**

There are 3 auto generated files here: common.rs, api.rs, and api_grpc.rs. No need to review those.

PR #387, opened 2020-04-30
Verifying Drand Entries from Blocks

ec2:
**Summary of changes**
- Creates the verification functions for verifying the Beacon Entries within Blocks
- TODO: Use those functions in the block verification process 

Closes 
#375 


blockchain/blocks/src/header.rs
```diff
@@ -339,6 +340,48 @@ impl BlockHeader {
...
+                    .map_err(|e| Error::Validation(e.to_string()))?;
+                if !valid {
+                    return Err(Error::Validation(format!(
+                        "beacon entry was invalid: curr:{:?}, prev: {:?}",
+                        curr, prev.1
+                    )));
+                }
+                Ok((valid, curr))
+            },
+        )?;
```
timvermeulen, line 382: It seems like you're not using the boolean flag here anywhere (and I think it's always `true` anyway), so you should be able to just get rid of it

blockchain/beacon/src/mock_beacon.rs
```diff
@@ -0,0 +1,44 @@
...
+    interval: Duration,
+}
+
+impl MockBeacon {
+    pub fn new(interval: Duration) -> Self {
+        MockBeacon { interval }
+    }
+    fn entry_for_index(index: u64) -> BeaconEntry {
+        let mut buf: Vec<u8> = Vec::with_capacity(8);
+        buf.write_u64::<BigEndian>(index).unwrap();
```
timvermeulen, line 21:
```suggestion
        let mut buf = [0; 8];
        BigEndian::write_u64(&mut buf, index);
```
Saves an allocation. You will need to `use byteorder::ByteOrder` as well

blockchain/beacon/src/drand.rs
```diff
@@ -31,6 +33,25 @@ impl DistPublic {
...
+        &self,
+        curr: &BeaconEntry,
+        prev: &BeaconEntry,
+    ) -> Result<bool, Box<dyn error::Error>>;
+
+    /// Returns a BeaconEntry given a round. It fetches the BeaconEntry from a Drand node over GRPC
+    /// In the future, we will cache values, and support streaming.
+    async fn entry(&self, round: u64) -> Result<BeaconEntry, Box<dyn error::Error>>;
+
+    fn max_beacon_round_for_epoch(&self, fil_epoch: ChainEpoch, _prev_entry: &BeaconEntry) -> u64;
```
timvermeulen, line 52:
```suggestion
    fn max_beacon_round_for_epoch(&self, fil_epoch: ChainEpoch, prev_entry: &BeaconEntry) -> u64;
```

blockchain/beacon/src/drand.rs
```diff
@@ -119,46 +126,24 @@ impl DrandBeacon {
...
-        let beacon = new_beacon().await;
-
-        let e2 = beacon.entry(2).await.unwrap();
-        let e3 = beacon.entry(3).await.unwrap();
-        assert!(!beacon.verify_entry(e2, e3).unwrap());
+    fn max_beacon_round_for_epoch(&self, fil_epoch: ChainEpoch, _prev_entry: &BeaconEntry) -> u64 {
+        // TODO: sometimes the genesis time for filecoin is zero and this goes negative
+        let latest_ts = fil_epoch * self.fil_round_time + self.fil_gen_time - self.fil_round_time;
+        // TODO: self.interval has to be converted to seconds. Dont know what it is right now
+        (latest_ts - self.drand_gen_time) / self.interval
```
austinabell, line 147: I feel like this should probably be checked (a larger genesis time would just crash the node)

blockchain/beacon/src/drand.rs
```diff
@@ -119,46 +126,24 @@ impl DrandBeacon {
...
 
-    #[async_std::test]
-    async fn ask_and_verify_beacon_entry_fail() {
-        let beacon = new_beacon().await;
-
-        let e2 = beacon.entry(2).await.unwrap();
-        let e3 = beacon.entry(3).await.unwrap();
-        assert!(!beacon.verify_entry(e2, e3).unwrap());
+    fn max_beacon_round_for_epoch(&self, fil_epoch: ChainEpoch, _prev_entry: &BeaconEntry) -> u64 {
+        // TODO: sometimes the genesis time for filecoin is zero and this goes negative
```
austinabell, line 144: can it though? Because shouldn't genesis time be in unix time? or is it not?

blockchain/beacon/src/lib.rs
```diff
@@ -6,6 +6,38 @@ mod drand;
...
+    round: ChainEpoch,
+    prev: &BeaconEntry,
+) -> Result<Vec<BeaconEntry>, Box<dyn Error>> {
+    let max_round = beacon.max_beacon_round_for_epoch(round, prev);
+    if max_round == prev.round() {
+        return Ok(vec![]);
+    }
+    // TODO: this is a sketchy way to handle the genesis block not having a beacon entry
+    let prev_round = if prev.round() == 0 {
+        max_round - 1
```
austinabell, line 29: what are you planning on doing with this? seems like you can't rely on max_round not being 0 and causing a panic

blockchain/chain/src/store/chain_store.rs
```diff
@@ -136,6 +137,26 @@ where
...
     }
 
+    /// Finds the latest beacon entry given a tipset up to 20 blocks behind
+    pub fn latest_beacon_entry(&self, ts: &Tipset) -> Result<BeaconEntry, Error> {
+        let mut cur = ts.clone();
+        for _ in 1..20 {
+            let cbe = ts.blocks()[0].beacon_entries();
+            if !cbe.is_empty() {
+                return Ok(cbe[cbe.len() - 1].clone());
+            }
```
austinabell, line 147:
```suggestion
            if let Some(entry) = cbe.last() {
                return Ok(entry.clone());
            }
```

blockchain/blocks/src/header.rs
ec2: Oh yeah, my brain got folded up trying to reason with `fold` haha

blockchain/beacon/src/drand.rs
ec2: True, it should be

blockchain/beacon/src/lib.rs
ec2: max_round cant be 0 is prev.round() == 0. That necessarily means that were at least on round >= 1.

blockchain/beacon/src/drand.rs
ec2: Shouldn't happen in our life time. Current unix time is around: 1589569053 ~= 2^30.

austinabell: sorry, I mean drand genesis time being larger than the fil genesis time, if epoch is 0 then the line I commented on can overflow. Is there a guarantee that epoch cannot be 0? Or is drand genesis time not in unix time?

blockchain/beacon/src/lib.rs
austinabell: Ohh I see, this is handling the logic that was done on construction in go impl

blockchain/beacon/src/drand.rs
ec2: Ah. I'm not sure actually... This is the exact math the have in Lotus though...
```
func (db *DrandBeacon) MaxBeaconRoundForEpoch(filEpoch abi.ChainEpoch, prevEntry types.BeaconEntry) uint64 {
	// TODO: sometimes the genesis time for filecoin is zero and this goes negative
	latestTs := ((uint64(filEpoch) * db.filRoundTime) + db.filGenTime) - db.filRoundTime
	dround := (latestTs - db.drandGenTime) / uint64(db.interval.Seconds())
	return dround
}

```

austinabell: Yeah, just the thing to note is go doesn't panic on overflow, we would, which is a reason against matching this (esp with their TODO comment)

ec2: OH wow, that's wild. I never knew they dont panic on under/overflows. Let me ask Why if fil_gen_time > drand_gen_time is guaranteed. I have a feeling it is...

ec2: Yup, confirmed!

blockchain/blocks/src/header.rs
```diff
@@ -355,6 +356,48 @@ impl BlockHeader {
         // h(vrfout) * totalPower < e * sectorSize * 2^256
         lhs < rhs
     }
+
```
dutterbutter, line 359: Nit: please add comment

blockchain/chain_sync/src/sync.rs
```diff
@@ -987,8 +1013,12 @@ mod tests {
             Cid::from_raw_cid("bafy2bzacecujyfvb74s7xxnlajidxpgcpk6abyatk62dlhgq6gcob3iixhgom")
                 .unwrap();
 
-        let root =
-            ChainSyncer::compute_msg_data(cs.chain_store.blockstore(), &[bls], &[secp]).unwrap();
+        let root = ChainSyncer::<MemoryDB, MockBeacon>::compute_msg_data(
```
austinabell, line 1016: Seems weird for this to be associated, you shouldn't need to infer the beacon type for this. At very least, you shouldn't have to specify DB generic here. Probably just makes sense to move this fn out of the ChainSyncer impl
```suggestion
        let root = ChainSyncer::<_, MockBeacon>::compute_msg_data(
```

blockchain/beacon/Cargo.toml
```diff
@@ -5,22 +5,23 @@ authors = ["ChainSafe Systems <info@chainsafe.io>"]
...
-grpc-protobuf = "0.7.1"
+grpc = "0.8"
+grpc-protobuf = "0.8"
 protobuf = "2.14.0"
+clock = { path = "../../node/clock" }
 
-httpbis =  "0.8.0"
-tls-api = "0.3.2"
-tls-api-openssl = "0.3.2"
+
```
austinabell, line 13: nit: can you remove these line breaks?

blockchain/chain_sync/src/sync.rs
ec2: Agree. I'm taking it out of the impl block

Issue #384, opened 2020-04-27
Investigate Syncing Interop 


amerameen:
- [ ] Investigate other implementation's codebases
- [ ] Try and sync (pull in blocks) with the testnet
- [ ] Create any issues that need to be resolved to sync

PR #388, opened 2020-04-30
Interop updates and refactoring

austinabell:
**Summary of changes**
- Interop updates and refactoring chain store logic to be more consistent
- Updated hello protocol handling to route requests through ChainSyncer
  - This isn't ideal in our current setup, because that thread could be busy syncing, but is fine for now
  - Could potentially access the heaviest tipset with some Arc shared memory between the libp2p thread and chainsyncer, which could be a better solution, but I won't make any big changes yet because we may want those messages being passed through the chainsyncer for metrics on the peers

Still a few things to be ironed out and debugged, but gives a much cleaner setup to work off of and split up remaining tasks.

#384 (Will try to detail issues to do now before closing that issue though)

Closes 


README.md
```diff
@@ -39,9 +39,42 @@ make install
 forest
 ```
 
-### Testing
+### Config
+
+To run the node with custom config and bootnodes, the `--config` or `-c` flag can be used to specify a toml file with the options. The main configurations options include but are not limited to:
```
ec2, line 44:
```suggestion
To run the node with custom config and bootnodes, the `--config` or `-c` flag can be used to specify a TOML file with the options. The main configurations options include but are not limited to:
```

PR #390, opened 2020-05-02
DAGJson support for Ipld

austinabell:
**Summary of changes**
- Useful going forward for selectors and testing, implementation is not efficient since the serialization and deserialization requires some theoretically unnecessary clones/copies, but it doesn't matter because this is just JSON and I don't want to fork serde_json or reimplement
  - Added json support for Cids with this (for obvious reasons) but did not for anything outside of IPLD because I don't think it'll be necessary yet
- Updates serialization tests to use Cid json and cleaned up a few other small things

Also added the ipld! macro for fun (makes things a lot easier)

Closes 

### DIFF
--- a/ipld/src/lib.rs
+++ b/ipld/src/lib.rs
@@ -3,6 +3,7 @@
 
 mod de;
 mod error;
+pub mod json;
 mod ser;
 
 pub use self::error::Error;
--- a/ipld/src/lib.rs
+++ b/ipld/src/lib.rs
@@ -8,6 +8,9 @@ mod ser;
 #[cfg(feature = "json")]
 pub mod json;
 
+#[macro_use]
+mod macros;
+
 pub use self::error::Error;
 
 use cid::Cid;
--- a/ipld/src/lib.rs
+++ b/ipld/src/lib.rs
@@ -113,7 +113,10 @@ pub fn from_ipld<T>(value: &Ipld) -> Result<T, String>
 where
     T: DeserializeOwned,
 {
-    // TODO find a way to convert without going through byte buffer
+    // TODO update to not go through byte buffer to convert
+    // There is a good amount of overhead for this (having to implement serde::Deserializer)
+    // for Ipld, but possible. The benefit isn't worth changing yet since if the value is not
+    // passed by reference as needed by HAMT, then the values will have to be cloned.
     let buf = to_vec(value).map_err(|e| e.to_string())?;
     from_slice(buf.as_slice()).map_err(|e| e.to_string())
 }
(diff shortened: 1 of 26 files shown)


austinabell:
So pr is blocked by breaking changes in rust-fil-proofs being updated, but also could be some functional changes depending on what happens with https://github.com/ipld/specs/pull/259 

Isn't a big rush to get this in so can probably wait for that to resolve

vm/interpreter/src/default_runtime.rs
```diff
@@ -296,7 +296,7 @@ where
         value: &TokenAmount,
     ) -> Result<Serialized, ActorError> {
         let msg = UnsignedMessage::builder()
-            .to(to.clone())
+            .to(*to)
```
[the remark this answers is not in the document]
austinabell: doesn't matter at all really, going to avoid changing in this PR because that will make other changes necessary unrelated to this, but yeah that can be done

ipld/src/lib.rs
```diff
@@ -14,17 +20,84 @@ use serde::de::DeserializeOwned;
...
 
-/// Represents IPLD data structure used when serializing and deserializing data
+/// Represents IPLD data structure used when serializing and deserializing data.
 #[derive(Debug, Clone, PartialEq)]
 pub enum Ipld {
+    /// Represents a null value.
+    ///
+    /// ```
+    /// # use forest_ipld::ipld;
+    /// let v = ipld!(null);
```
ec2, line 30: Feels weird having a doc test for how to use an Enum... Just adds to test compilation and running without much value. Your documentation of the macro is pretty good already

ipld/src/lib.rs
```diff
@@ -44,7 +117,10 @@ pub fn from_ipld<T>(value: &Ipld) -> Result<T, String>
 where
     T: DeserializeOwned,
 {
-    // TODO find a way to convert without going through byte buffer
+    // TODO update to not go through byte buffer to convert
```
ec2, line 120: This is so janky its funny. But agree we dont need to fix this yet haha

ipld/src/lib.rs
austinabell: Made more sense to me to document how to construct each variant instead of doing all on the Ipld enum. These are pretty lightweight but I can add `no_run` to them or remove, just nice to have examples for each variant for practical use imo.

ipld/src/lib.rs
austinabell: I mean, it's not THAT janky, not like it's a common concept to be able to go from a generalized type to a specific one without a buffer usually. This is correct, just not efficient

ec2: Definitely correct, not questioning. I just think its funny how we have to go to CBOR first then deserialize. Anyways, nothing needs to be fixed here.

PR #393, opened 2020-05-04
implement verify seal syscall

### LINKED ISSUE #304: verify_seal syscall

dutterbutter:
**Summary of changes**
- implement verify seal syscall

Closes #304 


vm/runtime/src/lib.rs
```diff
@@ -182,9 +183,25 @@ pub trait Syscalls {
...
     }
     /// Verifies a sector seal proof.
-    fn verify_seal(&self, _vi: &SealVerifyInfo) -> Result<(), Box<dyn StdError>> {
-        // TODO
-        todo!()
+    fn verify_seal(&self, vi: &SealVerifyInfo) -> Result<(), Box<dyn StdError>> {
+        let commd = cid_to_data_commitment_v1(&vi.unsealed_cid)?;
+        let commr = cid_to_replica_commitment_v1(&vi.on_chain.sealed_cid)?;
+        let miner_addr = Address::new_id(vi.sector_id.miner);
+        let prover_id = <[u8; 32]>::try_from(miner_addr.payload_bytes().as_slice())?;
```
austinabell, line 190: This will error always (unless the payload slice is length 32 lol) should initialize an empty `Commitment` (would be nice to use the type from the api for readability) and then copy from the payload bytes

vm/runtime/src/lib.rs
```diff
@@ -182,9 +183,25 @@ pub trait Syscalls {
...
+            vi.on_chain.registered_proof.into(),
+            commr,
+            commd,
+            prover_id,
+            SectorId::from(vi.sector_id.number),
+            vi.randomness,
+            vi.interactive_randomness,
+            &vi.on_chain.proof,
+        )? {
+            return Err(format!("Invalid proof detected: {:?}", vi.on_chain.proof).into());
```
austinabell, line 201: Debug printing bytes as the error probably won't be readable, thoughts on encoding to some string base?

vm/runtime/src/lib.rs
```diff
@@ -182,9 +183,25 @@ pub trait Syscalls {
...
     /// Verifies a sector seal proof.
-    fn verify_seal(&self, _vi: &SealVerifyInfo) -> Result<(), Box<dyn StdError>> {
-        // TODO
-        todo!()
+    fn verify_seal(&self, vi: &SealVerifyInfo) -> Result<(), Box<dyn StdError>> {
+        let commd = cid_to_data_commitment_v1(&vi.unsealed_cid)?;
+        let commr = cid_to_replica_commitment_v1(&vi.on_chain.sealed_cid)?;
+        let miner_addr = Address::new_id(vi.sector_id.miner);
+        let prover_id = <[u8; 32]>::try_from(miner_addr.payload_bytes().as_slice())?;
+        if !verify_seal(
```
ec2, line 191: I got so confused... I was like how is this recursive call happening with different params lmao. Could you maybe prefix the import with `proofs_`, so `proofs_verify_seal`

vm/runtime/src/lib.rs
austinabell: We are sure it isn't always 32 lol. It's an ID address payload, its dynamic size and generally very small

ec2: Oops, this got posted after Dustin made his changes. copy_from_slice will fail still since its not 32 bytes.

vm/runtime/src/lib.rs
```diff
@@ -182,9 +183,31 @@ pub trait Syscalls {
...
     /// Verifies a sector seal proof.
-    fn verify_seal(&self, _vi: &SealVerifyInfo) -> Result<(), Box<dyn StdError>> {
-        // TODO
-        todo!()
+    fn verify_seal(&self, vi: &SealVerifyInfo) -> Result<(), Box<dyn StdError>> {
+        let commd = cid_to_data_commitment_v1(&vi.unsealed_cid)?;
+        let commr = cid_to_replica_commitment_v1(&vi.on_chain.sealed_cid)?;
+        let miner_addr = Address::new_id(vi.sector_id.miner);
+        let mut prover_id = ProverId::default();
+        prover_id.copy_from_slice(&miner_addr.payload_bytes());
```
ec2, line 191: This will fail. The payload bytes is guaranteed to be under 32 bytes.

austinabell: Oh, you're right, didn't realize that function was panic, I was under the assumption that only panics if slice is longer

dutterbutter: Nice catch @ec2 ....hmmm so how should I go about this then? The go impl uses this [to32ByteArray](https://github.com/filecoin-project/filecoin-ffi/blob/870251cd04c54e7a3a08b714f3e71a9edec28445/proofs.go#L759) to make it 32 bytes.

austinabell, line 191:
```suggestion
        let miner_payload = miner_addr.payload_bytes();
        prover_id[..miner_payload.len()].copy_from_slice(&miner_payload);
```
Is probably what we want then? Probably good to write a test for this function, even just to pass fake values in and assert the verification errors and not anything else.

vm/runtime/src/lib.rs
```diff
@@ -227,3 +251,99 @@ pub enum ConsensusFaultType {
...
+            _extra: &[u8],
+            _earliest: ChainEpoch,
+        ) -> Result<Option<ConsensusFault>, Box<dyn StdError>> {
+            Ok(Some(ConsensusFault {
+                target: Address::new_id(0),
+                epoch: 0,
+                fault_type: ConsensusFaultType::DoubleForkMining,
+            }))
+        }
+    }
```
austinabell, line 325: remove all of this, the point is to test the default impl? Can just use the DefaultSyscalls instead of TestSyscalls

dutterbutter: Moved tests to test dir and switched to use DefaultSyscalls

PR #394, opened 2020-05-05
Init test porting

nannick:
**Summary of changes**
-  Ported Tests from https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/init/init_test.go
- Some minor bugs were fixed along the way

Closes 
310


vm/actor/src/builtin/cron/mod.rs
```diff
@@ -74,12 +74,12 @@ impl Actor {
...
         let st: State = rt.state()?;
         for entry in st.entries {
-            rt.send(
+            let _v = rt.send(
                 &entry.receiver,
                 entry.method_num,
                 &Serialized::default(),
                 &TokenAmount::from(0u8),
-            )?;
+            );
```
austinabell, line 82: revert back to what it was, this is ignoring that error

vm/actor/src/builtin/init/mod.rs
```diff
@@ -71,10 +71,6 @@ impl Actor {
             ));
         }
 
-        // Compute a re-org-stable address.
-        // This address exists for use by messages coming from outside the system, in order to
-        // stably address the newly created actor even if a chain re-org causes it to end up with
-        // a different ID.
```
austinabell, line 77: why remove this comment?

vm/actor/src/builtin/init/mod.rs
```diff
@@ -90,18 +86,23 @@ impl Actor {
...
+            )
+            .map_err(|err| rt.abort(err.exit_code(), "constructor failed"));
+
+        match v {
+            Ok(_) => Ok(ExecReturn {
+                id_address,
+                robust_address,
+            }),
+            Err(e) => Err(e),
+        }
```
austinabell, line 105: can you revert this back to what it was? This is just a more unreadable version

vm/actor/tests/common/mod.rs
```diff
@@ -28,6 +28,7 @@ pub struct MockRuntime<'a, BS: BlockStore> {
     pub id_addresses: HashMap<Address, Address>,
     pub actor_code_cids: HashMap<Address, Cid>,
     pub new_actor_addr: Option<Address>,
+    pub message: UnsignedMessage,
```
austinabell, line 31: So about this, since all you need from this is caller, receiver, value_received and is currently storing that data twice (see above) you either need to remove the three fields above and handle those through the message type or update the runtime interface to not be hard coded `UnsignedMessage` and instead be an interface which returns those fields.

I think it makes sense to just do the former and add a TODO since that change is a bit more involved (cc: @ec2)

vm/actor/tests/common/mod.rs
```diff
@@ -229,14 +242,28 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
...
+    #[allow(dead_code)]
+    pub fn set_caller(&mut self, code_id: Cid, address: Address) {
+        self.caller = address.clone();
+        self.caller_type = code_id.clone();
+        self.actor_code_cids.insert(address, code_id);
     }
 }
 
 impl<BS: BlockStore> Runtime<BS> for MockRuntime<'_, BS> {
+    // Cuasing a test to fail have to implement to pass. Ask about the
```
austinabell, line 264: commenting to reference above as it's related, this impl is fine, but probably want to add back the `self.require_in_call()`

vm/actor/tests/common/mod.rs
```diff
@@ -431,7 +459,16 @@ impl<BS: BlockStore> Runtime<BS> for MockRuntime<'_, BS> {
...
+        match expected_msg.exit_code {
+            ExitCode::Ok => return Ok(expected_msg.send_return),
+            x => {
+                return Err(ActorError {
+                    fatal: false,
+                    exit_code: x,
+                    msg: "Expected message Fail".to_string(),
+                });
+            }
+        }
```
austinabell, line 471: This seems wrong, why are you doing this?

vm/actor/tests/common/mod.rs
```diff
@@ -229,14 +242,28 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
             value,
             send_return,
             exit_code,
-        })
+        });
+        ()
```
austinabell, line 246:
```suggestion
```

vm/actor/tests/common/mod.rs
```diff
@@ -80,6 +81,18 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
...
+                .to(Address::new_id(100).unwrap())
+                .from(Address::new_id(100).unwrap())
+                .sequence(0) // optional
+                .value(TokenAmount::from(0u8)) // optional
+                .method_num(MethodNum::default()) // optional
+                .params(Serialized::default()) // optional
+                .gas_limit(0) // optional
+                .gas_price(TokenAmount::from(0u8)) // optional
+                .build()
+                .unwrap(),
```
austinabell, line 94: Where are you getting those addresses? seems like a bad default. (You only need to specify to, from, and value and value will default to 0 if not used in builder)
```suggestion
                .to(Address::new_id(100).unwrap())
                .from(Address::new_id(100).unwrap())
                .build()
                .unwrap(),
```

vm/actor/tests/common/mod.rs
ec2: Yeah it makes sense. I was thinking we need to make a Message trait. In this case, I think this is the only way to make it work. The message() function returns a reference to a UnsignedMessage, so its not like we can construct the message using the caller, receiever, value_received inside of that function

vm/actor/tests/common/mod.rs
ec2: Yeah, please add that back

vm/actor/tests/common/mod.rs
austinabell: Just to be clear about that, it is a bit awkward to use, but only for the mock runtime. I'm not convinced that making it generic is objectively better, can just handle those things through an unsigned message in the mock rt since it's only used for testing, but maybe yeah a bit clunky still. 

Having the message trait makes the mockrt impl more straightforward, at the cost of the actual runtime being unnecessarily complicated

ec2: Yeah. I was thinking about it from the perspective of the MockRt. I dont expect us to be building any more runtimes anyways.

vm/actor/tests/common/mod.rs
nannick: In the go version if the command was expected to fail they return the failed exit code. I assumed the best way to mimic that was to return a error. Before we would always return Ok(expected_msg.send_return) even if the expected message was supposed to fail

vm/actor/src/builtin/cron/mod.rs
nannick: Yeah I was confused by this too. In the spec-actors repo they ignore any error or return values. I wasn't sure about why they did it but I just mimicked what they had. Are they doing something wrong?

austinabell: ohh, you're right. Interesting why that error is ignored, but I guess it makes sense. Possibly add a comment here to indicate that this is intentional?

vm/actor/src/builtin/init/mod.rs
```diff
@@ -90,15 +90,15 @@ impl Actor {
...
-            METHOD_CONSTRUCTOR,
+            1,
             &params.constructor_params,
             &rt.message().value().clone(),
         )
-        .map_err(|err| rt.abort(err.exit_code(), "constructor failed"))?;
-
-        Ok(ExecReturn {
+        .map_err(|err| rt.abort(err.exit_code(), "constructor failed"))
+        .map(|_| ExecReturn {
```
austinabell, line 101: I mean, this works exactly the same but is a bit confusing because you are using map but not mapping any values. This is the same as doing a try return `?` and just returning the Ok `ExecReturn`

vm/actor/tests/common/mod.rs
```diff
@@ -80,6 +81,12 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
             actor_code_cids: HashMap::new(),
             new_actor_addr: None,
 
+            message: UnsignedMessage::builder()
+                .to(Address::new_id(100))
+                .from(Address::new_id(100))
```
austinabell, line 86: So the issue still remains that there is duplicate data (from and to address) which will cause issues with certain tests. If you intend to keep that `message()` function implementation to returning this message, if somewhere in the code they retrieved this to and from address it would be wrong. The equivalents of `to` and `from` here are `receiver` and `caller` in the mockrt respectively

vm/actor/tests/common/mod.rs
austinabell: Oh okay I didn't realize that error was not handled before. @ec2 can you verify this, since you set up this function? Seems weird to attach an arbitrary message to the error, but I guess maybe that's necessary to port tests over from their framework

ec2: Yeah, I think i messed that one up there, or their mock rt changed a bit. Not sure which one, but @nannick 's changes look correct

vm/actor/src/builtin/cron/mod.rs
ec2: I would do `let _ =` instead of `let _v =` but yeah, so weird they dont use the error...

vm/src/error.rs
```diff
@@ -11,11 +11,11 @@ use crate::ExitCode;
...
     /// Is this a fatal error.
-    fatal: bool,
+    pub fatal: bool,
     /// The exit code for this invocation, must not be `0`.
-    exit_code: ExitCode,
+    pub exit_code: ExitCode,
     /// Message for debugging purposes,
-    msg: String,
+    pub msg: String,
 }
```
austinabell, line 19: Side note on this, shouldn't need to expose those fields, instead just use the constructor (`ActorError::new`) instead of just creating this struct as you do in the changes you made in the mock runtime

vm/actor/tests/common/mod.rs
```diff
@@ -231,12 +234,29 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
...
+        self.expect_create_actor = Some(a);
+    }
+
+    #[allow(dead_code)]
+    pub fn set_caller(&mut self, code_id: Cid, address: Address) {
+        self.message = UnsignedMessage::builder()
+            .to(self.message.to().clone())
+            .from(address.clone())
+            .build()
+            .unwrap();
```
austinabell, line 250: So one last thing, sorry this is for sure not ideal, but you also have to make sure the value (value_received in the mockrt) is retained, so include building the `value` as well. Sorry I wasn't explicit on my last review, that variable should also not be managed in two different places

vm/actor/tests/common/mod.rs
```diff
@@ -69,8 +68,6 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
     pub fn new(bs: &'a BS, receiver: Address) -> Self {
```
austinabell, line 68: Probably makes sense to just pass in the message as a constructor parameter now

vm/actor/src/builtin/init/mod.rs
```diff
@@ -90,6 +90,7 @@ impl Actor {
         rt.create_actor(&params.code_cid, &id_address)?;
 
         // Invoke constructor
+
```
austinabell, line 93: nit
```suggestion
```

vm/actor/tests/common/mod.rs
```diff
@@ -66,20 +65,24 @@ pub struct ExpectedMessage {
...
             actor_code_cids: HashMap::new(),
             new_actor_addr: None,
 
+            message: message,
+            // UnsignedMessage::builder()
+            //     .to(receiver.clone())
+            //     .from(Address::default())
+            //     .value(0u8.into())
+            //     .build()
+            //     .unwrap(),
```
austinabell, line 85:
```suggestion
```

vm/actor/tests/init_actor_test.rs
```diff
@@ -0,0 +1,332 @@
...
+        code_cid: code_id,
+        constructor_params: Serialized::serialize(params).unwrap(),
+    };
+
+    rt.message = UnsignedMessage::builder()
+        .to(rt.message.to().clone())
+        .from(rt.message.from().clone())
+        .value(rt.message.value().clone())
+        .build()
+        .unwrap();
```
austinabell, line 310:
```suggestion
```

vm/actor/tests/init_actor_test.rs
```diff
@@ -0,0 +1,332 @@
...
+    );
+
+    // Revert state if call fails
+    let ret = match ret {
+        Ok(v) => Ok(v),
+        Err(e) => {
+            rt.state = prev_state;
+            Err(e)
+        }
+    };
```
austinabell, line 328:
```suggestion
    if let Err(_) = &ret {
        rt.state = prev_state;
    }
```

vm/actor/tests/init_actor_test.rs
```diff
@@ -0,0 +1,332 @@
...
+    let bs = MemoryDB::default();
+    let mut rt = construct_runtime(&bs);
+    construct_and_verify(&mut rt);
+    let anne = Address::new_id(1001);
+
+    rt.set_caller(ACCOUNT_ACTOR_CODE_ID.clone(), anne);
+
+    let err = exec_and_verify(&mut rt, POWER_ACTOR_CODE_ID.clone(), &"")
+        .expect_err("Exec should have failed");
+    assert_eq!(err.exit_code(), ExitCode::ErrForbidden);
```
ec2, line 46: The specs actor test execs and verifies on a cid.Undef as well. I think youre missing that

austinabell: can you give him the more specific details? Is it an option in the mockrt? which field specifically should be checked?

vm/actor/tests/init_actor_test.rs
```diff
@@ -0,0 +1,332 @@
...
+    let returned_address = state
+        .resolve_address(rt.store, &unique_address)
+        .expect("Address should have been found");
+    assert_eq!(expected_id_addr, returned_address);
+
+    // Should return error since the address of flurbo is unknown
+    let unknown_addr = Address::new_actor(b"flurbo");
+    state
+        .resolve_address(rt.store, &unknown_addr)
+        .expect_err("Address should have not been found");
```
ec2, line 161: Huh, never seen this expect_err... Cool find

vm/actor/tests/init_actor_test.rs
nannick: Wouldn't a cid.Undef just be a invalid cid? Wouldn't we just return a error in that case?

ec2: Youre right actually. I dont think we need to test for it. Sorry

nannick: I think @dutterbutter  and I talked about it a while ago. He believed it was a non-issue

PR #397, opened 2020-05-06
Clean up tipsets

timvermeulen:
**Summary of changes**
- Normalization of the `Tipset` types to all use a lowercase `s`
- `Tipset::new` now doesn't clone block headers, and only clones `Cid`s after the verification is done
- `Tipset::is_empty` is removed because tipsets are guaranteed to never be empty
- `FullTipset::new` now asserts that the vector of blocks isn't empty which we were already assuming in some places
- Other minor cleanup

### DIFF
--- a/blockchain/blocks/src/tipset.rs
+++ b/blockchain/blocks/src/tipset.rs
@@ -75,6 +75,7 @@ pub struct Tipset {
     key: TipsetKeys,
 }
 
+#[allow(clippy::len_without_is_empty)]
 impl Tipset {
     /// Builds a new Tipset from a collection of blocks.
     /// A valid tipset contains a non-empty collection of blocks that have distinct miners and all
@@ -184,10 +185,6 @@ impl Tipset {
     /// Returns the number of blocks in the tipset
     pub fn len(&self) -> usize {
         self.blocks.len()
-    }
-    /// Returns true if no blocks present in tipset
-    pub fn is_empty(&self) -> bool {
-        self.blocks.is_empty()
     }
     /// Returns a key for the tipset.
     pub fn key(&self) -> &TipsetKeys {
--- a/blockchain/blocks/src/tipset.rs
+++ b/blockchain/blocks/src/tipset.rs
@@ -14,21 +14,21 @@ use encoding::{
 use num_bigint::BigUint;
 use serde::Deserialize;
 
-/// A set of CIDs forming a unique key for a TipSet.
+/// A set of CIDs forming a unique key for a Tipset.
 /// Equal keys will have equivalent iteration order, but note that the CIDs are *not* maintained in
 /// the same order as the canonical iteration order of blocks in a tipset (which is by ticket)
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Default, Ord, PartialOrd)]
-pub struct TipSetKeys {
+pub struct TipsetKeys {
     pub cids: Vec<Cid>,
 }
 
-impl TipSetKeys {
+impl TipsetKeys {
     pub fn new(cids: Vec<Cid>) -> Self {
         Self { cids }
     }
 
     /// checks whether the set contains exactly the same CIDs as another.
-    pub fn equals(&self, key: &TipSetKeys) -> bool {
+    pub fn equals(&self, key: &TipsetKeys) -> bool {
         if self.cids.len() != key.cids.len() {
             return false;
         }
@@ -46,7 +46,7 @@ impl TipSetKeys {
     }
 }
 
-impl ser::Serialize for TipSetKeys {
+impl ser::Serialize for TipsetKeys {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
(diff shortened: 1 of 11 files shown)


blockchain/blocks/src/tipset.rs
```diff
@@ -55,103 +42,104 @@ impl ser::Serialize for TipSetKeys {
...
-        if headers.is_empty() {
-            return Err(Error::NoBlocks);
-        }
-
-        let mut sorted_headers = Vec::new();
-        let mut cids = Vec::new();
+        let (first_header, other_headers) = match headers.split_first() {
+            Some(x) => x,
+            None => return Err(Error::NoBlocks),
+        };
```
austinabell, line 76:
```suggestion
        let (first_header, other_headers) = headers.split_first().ok_or(Error::NoBlocks)?;
```

blockchain/blocks/src/tipset.rs
```diff
@@ -55,103 +42,104 @@ impl ser::Serialize for TipSetKeys {
...
-                        "miner_addresses are not distinct".to_string(),
-                    ));
-                }
-            }
-            // push headers into vec for sorting
-            sorted_headers.push(headers[i].clone());
-            // push header cid into vec for unique check (can be changed to hashset later)
-            cids.push(headers[i].cid().clone());
+        for header in other_headers {
+            // Skip redundant check
```
austinabell, line 88: Can remove this comment now, was just included before when it was skipping the 0 index

blockchain/blocks/src/tipset.rs
```diff
@@ -162,65 +150,59 @@ impl Tipset {
...
     blocks: Vec<Block>,
 }
 
 impl FullTipset {
-    /// constructor
-    pub fn new(blks: Vec<Block>) -> Self {
-        Self { blocks: blks }
+    /// constructor, panics when the given vector is empty
+    pub fn new(blocks: Vec<Block>) -> Self {
+        assert!(!blocks.is_empty());
```
austinabell, line 199: This isn't safe, can potentially receive a full tipset over the network and converted from a `TipsetBundle` that doesn't get validated before this constructor is called. I would just add a TODO to run the same validation on this as the `Tipset` so we can't panic on runtime here

ec2:
I had not noticed the casing issue... I'm surprised no ones OCD flared up haha. Good catch!

blockchain/chain/src/store/tip_index.rs
```diff
@@ -2,18 +2,18 @@
...
 use std::hash::{Hash, Hasher};
 
-/// TipSetMetadata is the type stored as the value in the TipIndex hashmap.  It contains
+/// TipsetMetadata is the type stored as the value in the TipIndex hashmap.  It contains
 /// a tipset pointing to blocks, the root cid of the chain's state after
 /// applying the messages in this tipset to it's parent state, and the cid of the receipts
 /// for these messages.
 #[derive(Clone, PartialEq, Debug)]
-pub struct TipSetMetadata {
+pub struct TipsetMetadata {
```
austinabell, line 16:
```suggestion
pub struct TipsetMetaData {
```
nit: If we are already changing the name, should probably be fixed

timvermeulen: I think metadata is one word nowadays, so that part seems to have been correct to begin with

blockchain/blocks/src/tipset.rs
timvermeulen: Of course, thanks

blockchain/blocks/src/tipset.rs
timvermeulen: Whoops, fell through the cracks

blockchain/blocks/src/tipset.rs
timvermeulen: Ah, okay, we should probably abstract over `Tipset` and `FullTipset` then

austinabell: Exactly, this is what I was hinting at earlier. Makes sense to do in another issue since there are other changes to make along with it, unless you'd like to do now

timvermeulen: I'm not sure which other changes would be made alongside, there is very marginal duplication between the two types and I'm hesitant about leaking this abstraction outside of this file since you would need to `use` the trait every time

node/forest_libp2p/tests/decode_test.rs
```diff
@@ -38,17 +43,19 @@ fn convert_single_tipset_bundle() {
...
 fn tipset_bundle_to_full_tipset() {
     let h0 = BlockHeader::builder()
         .weight(BigUint::from(1u32))
+        .miner_address(Address::new_id(0))
         .build()
         .unwrap();
     let h1 = BlockHeader::builder()
-        .weight(BigUint::from(2u32))
+        .weight(BigUint::from(1u32))
+        .miner_address(Address::new_id(1))
```
timvermeulen, line 58: @austinabell Let me know if this is the right way to handle this, I had to make the weights equal and the miners distinct in order to pass `FullTipset::new`'s validation

austinabell: Yup! that's intended, I'll quote the spec:

```
A valid tipset contains a non-empty collection of blocks that have distinct miners and all specify identical:

Epoch
Parents
ParentWeight
StateRoot
ReceiptsRoot
```

blockchain/blocks/src/tipset.rs
austinabell: You can do something like make it generic on an iterator instead of generic on this trait, and just map the full tipset before. I made the changes here as an example:
https://github.com/ChainSafe/forest/commit/1642d571f85825d4bcca585024b20dfb2d39dd15

Do with it how you wish, this is just one option, but I agree requiring that new trait probably isn't ergonomic

timvermeulen: Ah yeah that is even better

PR #401, opened 2020-05-08
refactor with structops macro

AshantiMutinta:
**Summary of changes**
- Refactors CLI to use structops rather than Clap cuts off a lot of code
-A few syntax changes

https://crates.io/crates/structopt

### DIFF
--- a/forest/src/cli/mod.rs
+++ b/forest/src/cli/mod.rs
@@ -8,53 +8,45 @@ pub use self::config::Config;
 pub(super) use self::genesis::initialize_genesis;
 
 use async_std::task;
-use clap::{App, Arg};
 use std::cell::RefCell;
 use std::io;
 use std::process;
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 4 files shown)


forest/src/cli/mod.rs
```diff
@@ -8,53 +8,48 @@ pub use self::config::Config;
...
-        None => Config::default(),
-    };
-    if let Some(genesis_file) = app.value_of("genesis") {
-        cfg.genesis_file = Some(genesis_file.to_owned());
-    }
-    // TODO in future parse all flags and append to a configuraiton object
-    // Retrun defaults
+        // TODO in future parse all flags and append to a configuraiton object
+        //(where to find these flags, should be easy to do with structops)
+        // Retrun defaults
```
austinabell, line 49:
```suggestion
        // (where to find these flags, should be easy to do with structops)
```
Can do that if you want, doesn't matter though (comment isn't very meaningful anyway, I don't really know why specifically it was there)

Issue #241, opened 2020-02-24
IPLD: Implement Selectors and Associated Logic


ansermino:

austinabell:
go impl can be found ~ https://github.com/ipld/go-ipld-prime/tree/master/traversal

spec: https://github.com/ipld/specs/blob/master/selectors/selectors.md

PR #402, opened 2020-05-08
Selector explore implementation

austinabell:
**Summary of changes**
- Logic for exploring selectors for progressing with traversing the ipld. I'm closely matching go-prime-ipld for now, but will likely change it in the next PR when I implement the last part of this (traversal) and have full functionality tested

I wrote the tests in json with a runner to be able to test against go impl and provide a language agnostic test suite (I believe JS selectors are in progress rn too)

#241 for reference to spec and go impl

Closes 

### DIFF
--- a/ipld/src/selector/mod.rs
+++ b/ipld/src/selector/mod.rs
@@ -8,6 +8,8 @@ use super::Ipld;
 pub use path_segment::PathSegment;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
+use std::ops::SubAssign;
+use Selector::*;
 
 /// Selectors are expressions that identify and select a subset of data from an IPLD DAG.
 /// Selectors are themselves IPLD and can be serialized and deserialized as such.
@@ -151,6 +153,14 @@ pub enum RecursionLimit {
     Depth(u64),
 }
 
+impl SubAssign<u64> for RecursionLimit {
+    fn sub_assign(&mut self, other: u64) {
+        if let RecursionLimit::Depth(v) = self {
+            *v -= other;
+        }
+    }
+}
+
 /// Condition is expresses a predicate with a boolean result.
 ///
 /// Condition clauses are used several places:
@@ -185,7 +195,6 @@ pub enum Condition {
 impl Selector {
     /// Returns a vector of all sectors of interest, `None` variant is synonymous with all.
     pub fn interests(&self) -> Option<Vec<PathSegment>> {
-        use Selector::*;
         match self {
             ExploreAll { .. } => None,
             ExploreFields { fields } => {
@@ -236,7 +245,6 @@ impl Selector {
 
     /// Processes and returns resultant selector node
     pub fn explore(self, ipld: &Ipld, p: &PathSegment) -> Option<Selector> {
-        use Selector::*;
         match self {
             ExploreAll { next } => Some(*next),
             ExploreFields { mut fields } => match ipld {
@@ -247,7 +255,7 @@ impl Selector {
                         fields.remove(s)
                     }
                     PathSegment::Int(i) => {
-                        let key = format!("{}", i);
+                        let key = i.to_string();
                         m.get(&key)?;
                         fields.remove(&key)
                     }
@@ -260,7 +268,7 @@ impl Selector {
                     }
                     match p {
                         PathSegment::String(s) => fields.remove(s),
-                        PathSegment::Int(i) => fields.remove(&format!("{}", i)),
+                        PathSegment::Int(i) => fields.remove(&i.to_string()),
                     }
                 }
                 _ => None,
@@ -295,13 +303,12 @@ impl Selector {
             ExploreRecursive {
                 current,
                 sequence,
-                limit,
+                mut limit,
                 stop_at,
             } => {
-                let next = match current {
-                    Some(s) => s.explore(ipld, p)?,
-                    None => sequence.clone().explore(ipld, p)?,
-                };
+                let next = current
+                    .unwrap_or_else(|| sequence.clone())
+                    .explore(ipld, p)?;
 
                 if !has_recursive_edge(&next) {
                     return Some(ExploreRecursive {
@@ -311,44 +318,32 @@ impl Selector {
                         stop_at,
                     });
                 }
-                match limit {
-                    RecursionLimit::Depth(depth) => {
-                        if depth < 2 {
-                            // Replaces recursive edge with None on last iteration
-                            // TODO revisit, shouldn't need to replace, would be better to just
-                            // return none when edge is hit on final depth
-                            return replace_recursive_edge(next, None).map(|s| *s);
-                        }
-                        Some(ExploreRecursive {
-                            current: replace_recursive_edge(next, Some(sequence.clone())),
-                            sequence,
-                            limit: RecursionLimit::Depth(depth - 1),
-                            stop_at,
-                        })
+
+                if let RecursionLimit::Depth(depth) = limit {
+                    if depth < 2 {
+                        // Replaces recursive edge with None on last iteration
+                        // TODO revisit, shouldn't need to replace, would be better to just
+                        // return none when edge is hit on final depth
+                        return replace_recursive_edge(next, None);
                     }
-                    RecursionLimit::None => Some(ExploreRecursive {
-                        current: replace_recursive_edge(next, Some(sequence.clone())),
-                        sequence,
-                        limit,
-                        stop_at,
-                    }),
+                    limit -= 1;
                 }
+
+                Some(ExploreRecursive {
+                    current: replace_recursive_edge(next, Some(*sequence.clone())).map(Box::new),
+                    sequence,
+                    limit,
+                    stop_at,
+                })
             }
             ExploreUnion(selectors) => {
-                let mut replace_selectors = Vec::with_capacity(selectors.len());
-                for s in selectors {
-                    if let Some(new) = s.explore(ipld, p) {
-                        // Push all explorable selectors in union
-                        replace_selectors.push(new)
-                    }
-                }
-                if replace_selectors.is_empty() {
-                    return None;
-                }
-                if replace_selectors.len() == 1 {
-                    return Some(replace_selectors.pop().unwrap());
-                }
-                Some(ExploreUnion(replace_selectors))
+                // Push all valid explored selectors to new vector
+                let replace_selectors: Vec<_> = selectors
+                    .into_iter()
+                    .filter_map(|s| s.explore(ipld, p))
+                    .collect();
+
+                Selector::from_selectors(replace_selectors)
             }
             // Go impl panics here, but panic on exploring malformed selector seems bad
             ExploreRecursiveEdge => None,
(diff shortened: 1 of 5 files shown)


ipld/src/selector/mod.rs
```diff
@@ -236,9 +235,126 @@ impl Selector {
...
+            } => {
+                let next = match current {
+                    Some(s) => s.explore(ipld, p)?,
+                    None => sequence.clone().explore(ipld, p)?,
+                };
+
+                if !has_recursive_edge(&next) {
+                    return Some(ExploreRecursive {
+                        sequence,
+                        current: Some(next.into()),
```
timvermeulen, line 309:
```suggestion
                        current: Some(Box::new(next)),
```
Somewhat subjective, but `From` and `Into` are super generic and it's not obvious which conversion is taking place here without carefully reading the code

ipld/src/selector/mod.rs
```diff
@@ -236,9 +235,126 @@ impl Selector {
...
+                            stop_at,
+                        })
+                    }
+                    RecursionLimit::None => Some(ExploreRecursive {
+                        current: replace_recursive_edge(next, Some(sequence.clone())),
+                        sequence,
+                        limit,
+                        stop_at,
+                    }),
+                }
```
timvermeulen, line 335: Some duplication going on here with only `limit` being different, I suggest
```rust
RecursionLimit::Depth(depth) if depth < 2 => ...,
_ => ...,
```
That would probably work best by giving the `RecursionLimit` type something like a `decr()` method

ipld/src/selector/mod.rs
```diff
@@ -236,9 +235,126 @@ impl Selector {
...
+            }
+            ExploreUnion(selectors) => {
+                let mut replace_selectors = Vec::with_capacity(selectors.len());
+                for s in selectors {
+                    if let Some(new) = s.explore(ipld, p) {
+                        // Push all explorable selectors in union
+                        replace_selectors.push(new)
+                    }
+                }
```
timvermeulen, line 344:
```suggestion
                // Push all explorable selectors in union
                let mut replace_selectors: Vec<_> = selectors
                    .into_iter()
                    .filter_map(|s| s.explore(ipld, p))
                    .collect();
```

ipld/src/selector/mod.rs
```diff
@@ -236,9 +235,126 @@ impl Selector {
...
+                if replace_selectors.is_empty() {
+                    return None;
+                }
+                if replace_selectors.len() == 1 {
+                    return Some(replace_selectors.pop().unwrap());
+                }
+                Some(ExploreUnion(replace_selectors))
```
timvermeulen, line 351: You have this twice, you could clean this up with an associated function that does it for you:
```rust
fn from_vec(mut vec: Vec<Self>) -> Option<Self> {
    match vec.len() {
        0 | 1 => vec.pop(),
        _ => Some(ExploreUnion(vec)),
    }
}
```

ipld/src/selector/mod.rs
```diff
@@ -258,3 +374,49 @@ impl Selector {
...
+
+    match next_sel {
+        ExploreRecursiveEdge => replace,
+        ExploreUnion(selectors) => {
+            let mut replace_selectors = Vec::with_capacity(selectors.len());
+            for s in selectors {
+                if let Some(new) = replace_recursive_edge(s, replace.clone()) {
+                    replace_selectors.push(*new)
+                }
+            }
```
timvermeulen, line 392: This is also essentially a `filter_map`, like above

ipld/src/selector/mod.rs
```diff
@@ -258,3 +374,49 @@ impl Selector {
         }
     }
 }
+
+fn replace_recursive_edge(
+    next_sel: Selector,
+    replace: Option<Box<Selector>>,
+) -> Option<Box<Selector>> {
```
timvermeulen, line 381: This function does not need to know about the `Box`es and actually becomes simpler if you leave them out here

ipld/src/selector/mod.rs
```diff
@@ -258,3 +374,49 @@ impl Selector {
...
+        ExploreRecursiveEdge { .. } => true,
+        ExploreUnion(selectors) => {
+            for s in selectors {
+                // Check if union includes recursive edge
+                if has_recursive_edge(s) {
+                    return true;
+                }
+            }
+            false
+        }
```
timvermeulen, line 419:
```suggestion
        ExploreUnion(selectors) => selectors.iter().any(has_recursive_edge),
```

ipld/src/selector/mod.rs
```diff
@@ -236,9 +235,126 @@ impl Selector {
     }
 
     /// Processes and returns resultant selector node
-    pub fn explore(self, _ipld: &Ipld, _p: &PathSegment) -> Option<Selector> {
-        // TODO
-        todo!()
+    pub fn explore(self, ipld: &Ipld, p: &PathSegment) -> Option<Selector> {
+        use Selector::*;
```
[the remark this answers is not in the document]
austinabell: mm yeah, the idea was to future proof and not expose all the variants outside of what needs them. Just seemed weird to expose all the variants within the same file it's defined. Guess it's fine in this case though

ipld/src/selector/mod.rs
austinabell: yeah I guess the counter point to that is that it's harder to read the logic of what's happening with the `Box::new(..)` clutter when that detail (the fact that it's boxed) doesn't matter. I'm happy to change it if you guys think this is more readable

ipld/src/selector/mod.rs
austinabell: what you suggest would break the logic, since Depth and None variant's still need to be differentiated, I will add the decrement method and just return outside of the match to avoid the dedup though, that's a good idea.

ipld/src/selector/mod.rs
austinabell: that's a good idea, I will modify because that will panic on 0 and I think it might make sense to have a `From<Vec<Selector>>` impl instead (I'll think about it a bit more before doing)

ipld/src/selector/mod.rs
austinabell: ah yeah, that was just an artifact of me refactoring to use Rc to avoid all the heap clones then reverting to save for after the full functionality. 

In a practical sense it's functionally is ~ the same, but yeah probably cleaner

austinabell:
> Looks good Austin! Understandable that you closely follow the Go implementation, though the `interests`/`explore` setup seems a bit convoluted on first readthrough. I think we'll probably end up with more clones (of `Selector`) than necessary because of this. Might be interesting to experiment with this after implementation/testing is done.

Yeah that was what I mentioned I was going to do, I don't like the current setup but going to wait until after full functionality tests before refactoring. I have some ideas but counterproductive to make now since we need to match the specific implementation in go

ipld/src/selector/mod.rs
austinabell: decided to go with a SubAssign<u64> impl, seems cleanest way

ipld/src/selector/mod.rs
timvermeulen: Yeah that's a very valid point, `into` somewhat hides the stuff that doesn't matter here. But it unfortunately can do all kinds of conversions, which makes this (though easier to read) slightly harder to understand :)

A third option is always to add a `boxed()` method to every type, might slightly get rid of the `Box::new` clutter without hiding the kind of conversion. But then the reader does need to know that it does the same as `Box::new`, so it's not perfect either.

ipld/src/selector/mod.rs
timvermeulen: The `decr` I suggested would just do nothing in case of `RecursionLimit::None`, so it would have worked in pretty much the same way 🙂

austinabell: Yeah the decr part of the suggestion was valid for sure, switching to subassign was just to be able to do `limit -=1;` instead of `limit.decr()` (only an opinionated change), I was just mentioning that the code block you put wrt breaking the logic

timvermeulen: Actually, in your current code structure you could just do `*depth -= 1` directly (with a `ref mut` binding), no need to implement it on `RecursionLimit`

austinabell: very true! Idea was for future use, but I guess we can assume this decrement will never be needed outside of a match

austinabell: Idk I kinda think this is a bit less readable/usable actually though, thoughts?
```rust
                if let RecursionLimit::Depth(depth) = &mut limit {
                    if *depth < 2 {
                        // Replaces recursive edge with None on last iteration
                        // TODO revisit, shouldn't need to replace, would be better to just
                        // return none when edge is hit on final depth
                        return replace_recursive_edge(next, None);
                    }
                    *depth -= 1;
                }
```

timvermeulen: Yeah, I have to agree, the mutation makes this less readable... Here's a tiny adaptation to consider:
```rust
let new_limit = match limit {
    RecursionLimit::Depth(depth) => {
        if depth < 2 {
            // Replaces recursive edge with None on last iteration
            // TODO revisit, shouldn't need to replace, would be better to just
            // return none when edge is hit on final depth
            return replace_recursive_edge(next, None);
        }
        RecursionLimit::Depth(depth - 1)
    }
    RecursionLimit::None => RecursionLimit::None,
};
```
Takes slightly less mental juggling to read this, I think.

austinabell: This just adds complexity to the way it is rn. I think, although useful, have gone too deep into this minor detail which has a good change of being refactored later anyway

timvermeulen: It's indeed pretty inconsequential 🙂 Though it definitely does not add complexity compared to mutability + the `SubAssign` conformance

timvermeulen: Your original version was very simple though, that might be the best one for now

ipld/src/selector/mod.rs
```diff
@@ -236,14 +244,116 @@ impl Selector {
     }
 
     /// Processes and returns resultant selector node
-    pub fn explore(self, _ipld: &Ipld, _p: &PathSegment) -> Option<Selector> {
-        // TODO
-        todo!()
+    pub fn explore(self, ipld: &Ipld, p: &PathSegment) -> Option<Selector> {
```
AshantiMutinta, line 247: This looks great, I only had a question on is selector a heavy enum? If the answer is yes what do you think about a reference?

It would also be interested(although abit premature right now) to see selector implement an iterator so that we could use it more lazily

austinabell: Not quite sure what you mean by `heavy enum`, care to elaborate? Yeah it would definitely be a more idiomatic interface to have this as an iterator, but I will hold off until the functionality is fully tested (in next PR). Great suggestion!

AshantiMutinta: Should have been more clear sorry! Something that occupies a lot of data, I guess what I am getting at is asking if the clones are going to be expensive but then again that is also a premature optimization

austinabell: Yeah, in practice it'll be very small (depends on the construction of the selector) but since it's recursive, there is no cap on the size. 

Unfortunately the clones are intended (as far as I can see) because of the certain branches which would modify selectors with requiring the previous to not be mutated. This is something I will explore very in depth to make sure there is not a more efficient way around this once I have a testsuite verifying all functionality.

PR #403, opened 2020-05-09
Rupan/reward actor tests

### LINKED ISSUE #315: Port over tests for Reward Actor
Tests for the actors are in the specs-actors repo. They needed to be ported (1:1) to our actor implementations to ensure that our VM implementations match.

specs-actors repo: https://github.com/filecoin-project/specs-actors/tree/master/actors/builtin

nannick:
**Summary of changes**
- Ported reward actor test and fixed some minor bugs found

Closes #315 

Didn't get to do this but we should probably create a func as a wrapper around call to revert state if call is being used instead of manually fixing state every time.


nannick:
Not sure why test is failing because cargo test passes on my laptop

austinabell:
> Not sure why test is failing because cargo test passes on my laptop

Yeah don't worry that's not anything to do with this PR, @ec2 had that on his branch. Just the linker is not working consistently on the CI rn

vm/actor/tests/common/mod.rs
```diff
@@ -161,10 +161,10 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
                 actor::paych::Actor.invoke_method(self, method_num, params)
             }
             x if x == &*MULTISIG_ACTOR_CODE_ID => {
-                actor::cron::Actor.invoke_method(self, method_num, params)
+                actor::multisig::Actor.invoke_method(self, method_num, params)
             }
             x if x == &*REWARD_ACTOR_CODE_ID => {
-                actor::cron::Actor.invoke_method(self, method_num, params)
+                actor::reward::Actor.invoke_method(self, method_num, params)
```
austinabell, line 167: what a lapse this was missed, good catch! I assume this code block was copied from the default runtime because this also exists there, do you mind fixing there was well?

vm/actor/tests/reward_actor_test.rs
```diff
@@ -0,0 +1,76 @@
...
+use common::*;
+use db::MemoryDB;
+use ipld_blockstore::BlockStore;
+use message::UnsignedMessage;
+use vm::{ExitCode, Serialized, TokenAmount, METHOD_CONSTRUCTOR};
+
+fn construct_runtime<BS: BlockStore>(bs: &BS) -> MockRuntime<'_, BS> {
+    let message = UnsignedMessage::builder()
+        .to(REWARD_ACTOR_ADDR.clone())
+        .from(SYSTEM_ACTOR_ADDR.clone())
```
austinabell, line 20:
```suggestion
        .to(*REWARD_ACTOR_ADDR)
        .from(*SYSTEM_ACTOR_ADDR)
```

You can get at the lazy static items from a deref, and Address is `Copy` now

vm/actor/tests/reward_actor_test.rs
```diff
@@ -0,0 +1,76 @@
...
+#[test]
+fn test_award_block_reward() {
+    let bs = MemoryDB::default();
+    let mut rt = construct_runtime(&bs);
+    construct_and_verify(&mut rt);
+
+    let miner = Address::new_id(1000);
+    let gas_reward = TokenAmount::from(10u8);
+
+    rt.expect_validate_caller_addr(&[SYSTEM_ACTOR_ADDR.clone()]);
```
austinabell, line 37:
```suggestion
    rt.expect_validate_caller_addr(&[*SYSTEM_ACTOR_ADDR]);
```

vm/actor/tests/reward_actor_test.rs
```diff
@@ -0,0 +1,76 @@
...
+        .from(SYSTEM_ACTOR_ADDR.clone())
+        .build()
+        .unwrap();
+    let mut rt = MockRuntime::new(bs, message);
+    rt.caller_type = SYSTEM_ACTOR_CODE_ID.clone();
+    return rt;
+}
+
+#[test]
+fn test_award_block_reward() {
```
austinabell, line 29: can you rename this to something around what it's doing? (failure when balance is less than reward)

vm/actor/tests/reward_actor_test.rs
```diff
@@ -0,0 +1,76 @@
...
+        Method::AwardBlockReward as u64,
+        &Serialized::serialize(&params).unwrap(),
+    );
+
+    // Expecting insufficdent funds, revert state
+    assert_eq!(
+        ExitCode::ErrInsufficientFunds,
+        call_result.unwrap_err().exit_code()
+    );
+    rt.state = prev_state;
```
austinabell, line 59: Is there a reason for saving the mock rt state and then resetting it at the end? Tests still pass without those two lines and I can't understand why they'd be there

nannick: I think you don't need it for this case but there are tests where not reverting the state would fail. I remember writing a test for the init actor where not reverting the state cause it to resolve a address it shouldn't have been able to. In any case I'm just following what they got in the go tests

vm/actor/tests/common/mod.rs
```diff
@@ -128,6 +130,26 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
...
+    #[allow(dead_code)]
+    pub fn expect_call_fail(
+        &mut self,
+        to_code: &Cid,
+        method_num: MethodNum,
+        params: &Serialized,
+        expected_code: ExitCode,
+    ) {
+        // will record previous state, execute call. Will expect a certain exit code. Will throw assertion fail if exit code does not match
+        //  if exit code matches will revert state
```
austinabell, line 143:
```suggestion
```

vm/actor/tests/reward_actor_test.rs
```diff
@@ -0,0 +1,69 @@
...
+    rt.expect_validate_caller_addr(&[*SYSTEM_ACTOR_ADDR]);
+
+    let params = AwardBlockRewardParams {
+        miner: miner,
+        penalty: TokenAmount::from(0u8),
+        gas_reward: gas_reward,
+        ticket_count: 0,
+    };
+
+    //Expect call to fail because actor doesnt have enough tokens to reward
```
austinabell, line 46:
```suggestion
    // Expect call to fail because actor doesnt have enough tokens to reward
```

vm/actor/tests/common/mod.rs
```diff
@@ -128,6 +130,26 @@ impl<'a, BS: BlockStore> MockRuntime<'a, BS> {
...
+        let prev_state = self.state.clone();
+
+        let call_result = self.call(to_code, method_num, params);
+
+        assert_eq!(expected_code, call_result.unwrap_err().exit_code());
+
+        self.state = prev_state;
+    }
```
austinabell, line 151: Okay so my question for this is, why not just add a revert if error at the end of our mock rt call function? Seems weird it would be setup to have to manually revert for expected failures and not for a call. @ec2 maybe you have an opinion on this?

ec2: I think whichever is fine. Its just doing this could make it easier to port tests since theres a lot less thinking since this is how they do it specs actor

austinabell: In a way, yes, but if you are referring to this being easier, an example of how it isn't is in the case of the init actor: it can't use this function because there is other logic that is needed before the call. 

This function is setup for a very specific use case so probably won't be applicable in most other tests.

austinabell: The way I suggested, you get that for free and don't have to do that logic in different places. Just sharing thought process before I forget, I don't care about this, it's just tests but may make life easier

nannick: So do you think its better to just change the function signature for call? Have a parameter called expected_code to help decide whether or not to revert states?

PR #404, opened 2020-05-11
Jaden/tipsetconversions

### LINKED ISSUE #398: Improve conversions between Tipset types
**Issue summary**

So after #397 comes in, the FullTipset is now validated on construction. Because of this, there is no need to do the validation on conversion from `FullTipset` -> `Tipset`. 

Currently, there is the two conversions from FullTipset to Tipset, `into_tipset` and `to_tipset`. After this PR comes in, the conversion should not return a result and just be able to map full blocks to headers in that conversion (opinionated and optional is to instead or also have a `From<FullTipset>` trait impl for Tipset).

Also, storing the TipsetKeys with the `Tipset` is a little redundant (but can be nice for less cloning and copying if used multiple times). Would be more beneficial if the tipset keys were some cached type and only generated when needed (mapping blockheaders to their cached cids). This part is more opinionated and optional and can be ignored.

As for the conversions between `Tipset` to `TipsetKeys`, there could be a way to more cleanly convert without cloning. This can be done by exposing some `into_cid(self) -> Cid` function on the blockheader and then there can be a very clean conversion instead of having to clone when we will need that. Again, here a `From<Tipset> for TipsetKeys` impl is encouraged, but not necessary.

To be clear `FullTipset` -> `Tipset` -> `TipsetKeys` can be done without error and very cheap, but the inverse order will always require a blockstore and/or network and (probably) not done in this crate

flodesi:
**Summary of changes**
- Remove result when converting from `FullTipset` to `Tipset`
- Add `From` traits to convert from `FullTipset` to `Tipset`

Closes #398


blockchain/blocks/src/tipset.rs
```diff
@@ -173,14 +186,9 @@ impl FullTipset {
...
-    pub fn into_tipset(self) -> Result<Tipset, Error> {
-        let headers = self.blocks.into_iter().map(|block| block.header).collect();
-        Tipset::new(headers)
-    }
-    /// Returns a Tipset
-    pub fn to_tipset(&self) -> Result<Tipset, Error> {
+    pub fn to_tipset(&self) -> Tipset {
         let headers = self.blocks.iter().map(Block::header).cloned().collect();
-        Tipset::new(headers)
+        Tipset::new(headers).unwrap()
```
ec2, line 191: Why are you unwrapping here instead of returning a Result?

blockchain/blocks/src/tipset.rs
```diff
@@ -62,6 +62,19 @@ pub struct Tipset {
...
+impl From<FullTipset> for Tipset {
+    fn from(full_tipset: FullTipset) -> Self {
+        Tipset::new(
+            full_tipset
+                .blocks
+                .into_iter()
+                .map(|block| block.header)
+                .collect(),
+        )
+        .unwrap()
```
ec2, line 74: Why are you unwrapping here instead of returning a Result?

blockchain/blocks/src/tipset.rs
austinabell, line 191: using unwrap here opens us up to a panic, the validation should be the same at this point so you should be able to construct the struct without going through the `::new(..)` constructor. You may need to abstract some logic around sorting (or ensuring the blocks are sorted) in the `FullTipset` constructor to make sure this is never inconsistent.

blockchain/blocks/src/tipset.rs
AshantiMutinta: Yeah this looks like it should implement a Tryfrom instead

austinabell: the point of this is that the same validation is done in FullTipset as Tipset, so you can just construct a Tipset without having to go through the Tipset constructor (which does all the validation again)

blockchain/blocks/src/tipset.rs
flodesi: Just pushed a new commit that sorts the blocks in the `FullTipset` constructor and removed calls to the `Tipset` constructor when converting from `FullTipset` to `Tipset`.

blockchain/blocks/src/tipset.rs
```diff
@@ -155,7 +175,19 @@ impl FullTipset {
...
+        let mut sorted_blocks = blocks;
+        sorted_blocks.sort_by_key(|block| {
+            (
+                block.header.ticket().vrfproof.clone(),
+                block.header.cid().to_bytes(),
+            )
+        });
+        Ok(Self {
+            blocks: sorted_blocks,
+        })
```
austinabell, line 190: I can't add a suggestion, because a few lines are outside of the changes, but can simplify to:

```rust
    pub fn new(mut blocks: Vec<Block>) -> Result<Self, Error> {
        verify_blocks(blocks.iter().map(Block::header))?;

        // sort blocks on creation to allow for more seamless conversions between FullTipset
        // and Tipset
        blocks.sort_by_key(|block| {
            (
                block.header.ticket().vrfproof.clone(),
                block.header.cid().to_bytes(),
            )
        });
        Ok(Self { blocks })
    }
```

timvermeulen: Why aren't we only sorting by the ticket anyway? Aren't those already unique? (I saw that `Tipset::new` sorts by a tuple as well)

flodesi: Ah didn't catch that, will add that change and push commit.

austinabell: It is (or was) written into the spec to handle the secondary sorting. I'll try to find again, because quickly looking it is just referring to only sorting by ticket.

timvermeulen: If tickets are guaranteed to be unique then it will never hit the second tuple element — if not, then it's indeed needed for a canonical ordering

austinabell: yeah so lotus does this check: https://github.com/filecoin-project/lotus/blob/e25edb0187db62d342d39eebfaa3d70d5b92b77f/chain/types/tipset.go#L91

I can't find in the spec rn cc: @dutterbutter maybe you remember specifics of why it's like this.

In any case the sorting can be improved to not have to convert all cids to bytes for comparison, I think that was just done out of convenience

dutterbutter: The spec reads as follows:

```
The blocks in a tipset are canonically ordered by the lexicographic 
ordering of the bytes in each block’s ticket, breaking ties with the bytes
of the CID of the block itself.
```
My interpretation of that is that we order based on ticket size and if ticket sizes are equal we check the cid of the block itself.

flodesi: @timvermeulen So to clarify, you're proposing just using
```
block.header.ticket(),
```
as opposed to 
```
block.header.ticket().vrfproof.clone(),
```

austinabell: Thanks @dutterbutter for finding :) Yeah the comparison can still be improved by not cloning the ticket and only comparing cid bytes when other matches (having some sort function) if you want to take that on now @flodesi

timvermeulen: My wording was a bit vague, I was questioning whether `block.header.cid().to_bytes()` was needed there as well. But you raise a good point, `Ticket`'s ordering is defined by the ordering of its `vrfproof` so leaving out `.vrfproof` there will have no impact on the ordering whatsoever.

We unfortunately can't get rid of the `.clone()` though because of a language limitation ([StackOverflow](https://stackoverflow.com/questions/47121985/why-cant-i-use-a-key-function-that-returns-a-reference-when-sorting-a-vector-wi))

austinabell: what I'm suggesting is not sorting by keys and just having a sort function, there should be no need to clone the ticket, only cids in the extremely rare case of a ticket match

timvermeulen: @flodesi If you call `max_by` instead of `max_by_key`, you get in two blocks which lets you compare the tickets without cloning them, and only compare the CIDs if necessary like Austin suggested.

PR #407, opened 2020-05-11
SyncBucket cleanup

timvermeulen:
**Summary of changes**

Several loosely related changes:
- A fix for a bug where `SyncBucketSet::heaviest` wouldn't necessarily return the heaviest tipset
- Rename `same_chain_as` -> `is_same_chain_as` according to Rust's naming conventions
- Some other minor cleanup of the `SyncBucket` and `SyncBucketSet` types
- The removal of a couple `PartialOrd`/`Ord` implementations that weren't useful
- The removal of `Cid`'s manual `PartialEq` and `Hash` implementations in favor of derives that (should) have the same behavior but without allocating


blockchain/blocks/src/header.rs
```diff
@@ -202,18 +201,6 @@ impl<'de> Deserialize<'de> for BlockHeader {
...
-        self.cached_bytes.partial_cmp(&other.cached_bytes)
-    }
-}
-
-impl Ord for BlockHeader {
-    fn cmp(&self, other: &Self) -> Ordering {
-        self.cached_bytes.cmp(&other.cached_bytes)
-    }
-}
-
```
austinabell, line 216: @dutterbutter Wasn't there a specific need for this? I remember vaguely it being written into the spec

ipld/cid/src/lib.rs
```diff
@@ -43,7 +42,7 @@ pub struct Prefix {
 }
 
 /// Representation of a IPLD CID.
-#[derive(Eq, Clone)]
+#[derive(PartialEq, Eq, Hash, Clone)]
```
austinabell, line 45: It was intentional to keep those separate before (this was a fork of existing implementations initially) and the specific details of comparing and hashing bytes was necessary. I can't see at this point for that specific comparison (cids aren't used as Hamt keys and that was going to be the only potential need to be specific).

blockchain/blocks/src/header.rs
ec2: I vaguely remember us needing this at some point as well. Clearly not anymore?

ipld/cid/src/lib.rs
timvermeulen: Hmm, could could two CIDs ever be considered equal in one way but not in the other?

austinabell: Equality, no, but ordering potentially yes because of the variant encoding. Just something to keep in mind in the future, that would be a very nuanced bug to find if there was an inconsistency

timvermeulen: So are you talking about ordering and not about `Eq` and `Hash`? I agree that changing the ordering can change behavior but here I'm getting rid of `Cid`'s `Ord` implementation altogether, rather than changing it. Deriving `Eq` and `Hash` should not result in observable changes as long as the implementation is correct

austinabell: hash will be different too, only PartialEq/Eq would be the same (sorry for bringing up order, was just looking on my phone)

austinabell: But I don't think we use the cid hash for any consensus related things so should be fine

timvermeulen: Ah yeah, the hash will be different, but as far as I understand you would never look at the resulting hash value anyway, or somehow rely on what it is – i.e. previously the hash of a cid was the hash of its bytes which as far as we're concerned could be anything

austinabell: Yeah, I mean it could matter if you use a cid as the key to a hamt, but that shouldn't be a need in the near future because the go implementation statically uses bytes for their keys

timvermeulen: I will definitely keep this in mind for if we're ever going to use the CID as the key to a hamt

blockchain/blocks/src/header.rs
dutterbutter: I don't recall the specific need for this, I think its fine to be removed.

PR #408, opened 2020-05-11
Ipld selector traversals implementation

### LINKED ISSUE #241: IPLD: Implement Selectors and Associated Logic

austinabell:
**Summary of changes**
- Implements walk functions for selectors, which is the last missing part needed for GraphSync
  - The part that still can be done is `Focus` impls, which is just being able to traverse ipld by path, which isn't needed for any use case we have (afaik)

- I removed exposing the selector operation since there should be no external use of those (if there is it can just be easily updated when needed)

Test suite are json files that will probably be moved out to some shared repo, but for now I will leave as is

Closes #241 


ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,203 @@
...
+        match selector.interests() {
+            Some(interests) => {
+                for ps in interests {
+                    let v = match ipld.lookup_segment(&ps) {
+                        Some(ipld) => ipld,
+                        None => continue,
+                    };
+                    self.clone()
+                        .traverse_node(ipld, selector.clone(), callback, ps, v)
+                        .await?
```
AshantiMutinta, line 123: can we use a filter_map + map here? interests.filter_map.(...).map(..)

austinabell:
There is a bit of a bug in the go impl and by extension this right now. I'm writing a test for the edge case and will restructure to fix so hold off on reviewing

ipld/src/selector/walk.rs
austinabell: I'm a bit confused by this, can you explain? Do you mean filtering based on the available paths of the ipld?

AshantiMutinta: filtering based on the ipld.look_up_segment with a filtermap essentially. would save us from having the continue clause there since filter_map is based on options. I guess the issue though is that you might have to unwrap v in the map clause but it should be already filtered out

austinabell: yeah I had considered, I think the reason I hadn't was because integer path segments can be used ambiguously as fields of a map, but let me revisit to remember why I didn't go through with it

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+    ) -> Result<(), Error>
+    where
+        F: Fn(&Progress<L>, &Ipld) -> Result<(), String> + Sync,
+        L: LinkResolver + Sync + Send + Clone,
+    {
+        self.walk_all(ipld, resolver, |prog, ipld, reason| -> Result<(), String> {
+            if let VisitReason::SelectionMatch = reason {
+                return callback(prog, ipld);
+            }
+            Ok(())
```
timvermeulen, line 47:
```suggestion
            match reason {
                VisitReason::SelectionMatch => callback(prog, ipld),
                VisitReason::SelectionCandidate => Ok(()),
            }
```
Let me know what you think about this, I'm a sucker for exhaustiveness and symmetry myself – and Rust kinda discourages early returns with how you can omit the `return` keyword

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+    SelectionCandidate,
+}
+
+#[async_trait]
+pub trait LinkResolver {
+    #[allow(unused_variables)]
+    /// Resolves a Cid link into it's respective Ipld node, if it exists.
+    async fn load_link(&self, link: &Cid) -> Result<Option<Ipld>, String> {
+        Err("load_link not implemented on the LinkResolver".into())
+    }
```
timvermeulen, line 68:
```suggestion
    async fn load_link(&self, link: &Cid) -> Result<Option<Ipld>, String>;
```
Seems to me like you'd want to require conforming types to implement this? They can always return an error themselves if that's appropriate behavior

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl LinkResolver for () {}
+
+#[derive(Clone, Debug)]
+pub struct Progress<L = ()>
+where
+    L: Clone,
+{
```
timvermeulen, line 77:
```suggestion
pub struct Progress<L = ()> {
```
No need for the `Clone` bound here

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+    where
+        F: Fn(&Progress<L>, &Ipld, VisitReason) -> Result<(), String> + Sync,
+    {
+        // Resolve any links transparently before traversing
+        if let Ipld::Link(cid) = ipld {
+            if let Some(resolver) = &self.resolver {
+                let mut node = resolver.load_link(cid).await.map_err(Error::Link)?;
+                while let Some(Ipld::Link(c)) = node {
+                    node = resolver.load_link(&c).await.map_err(Error::Link)?;
+                }
```
timvermeulen, line 107: I would probably just leave this out, assuming it's not a super common occurrence — if it links to another link, the recursive call will automatically do the right thing

ipld/src/selector/walk.rs
timvermeulen: The trouble here is that calling `filter_map` with `ipld.lookup_segment` will give you an iterator of `Ipld`s without the paths (which you still need) so it does indeed become pretty verbose:

```rust
for (ps, ipld) in interests
    .into_iter()
    .filter_map(|ps| ipld.lookup_segment(&ps).map(|ipld| (ps, ipld)))
{
    // ...
}
```
Not really an improvement. Unless there's a way to improve this which I haven't thought of

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+                Ok(())
+            }
+            None => {
+                match ipld {
+                    Ipld::Map(m) => {
+                        for (k, v) in m.iter() {
+                            let ps: PathSegment = PathSegment::from(k.as_ref());
+                            self.clone()
+                                .traverse_node(ipld, selector.clone(), callback, ps, v)
+                                .await?
```
timvermeulen, line 151:
```suggestion
                                .await?;
```

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+                        }
+                    }
+                    Ipld::List(list) => {
+                        for (i, v) in list.iter().enumerate() {
+                            let ps: PathSegment = i.into();
+                            self.clone()
+                                .traverse_node(ipld, selector.clone(), callback, ps, v)
+                                .await?
```
timvermeulen, line 159:
```suggestion
                                .await?;
```

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+        callback: &F,
+        ps: PathSegment,
+        v: &Ipld,
+    ) -> Result<(), Error>
+    where
+        F: Fn(&Progress<L>, &Ipld, VisitReason) -> Result<(), String> + Sync,
+    {
+        if let Some(next_selector) = selector.explore(ipld, &ps) {
+            self.path.append(ps);
+            self.walk_all(v, next_selector, callback).await?
```
timvermeulen, line 185:
```suggestion
            self.walk_all(v, next_selector, callback).await?;
```

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+                        for (k, v) in m.iter() {
+                            let ps: PathSegment = PathSegment::from(k.as_ref());
+                            self.clone()
+                                .traverse_node(ipld, selector.clone(), callback, ps, v)
+                                .await?
+                        }
+                    }
+                    Ipld::List(list) => {
+                        for (i, v) in list.iter().enumerate() {
+                            let ps: PathSegment = i.into();
```
timvermeulen, line 156:
```suggestion
                            let ps = PathSegment::from(i);
```

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+                        .traverse_node(ipld, selector.clone(), callback, ps, v)
+                        .await?
+                }
+                Ok(())
+            }
+            None => {
+                match ipld {
+                    Ipld::Map(m) => {
+                        for (k, v) in m.iter() {
+                            let ps: PathSegment = PathSegment::from(k.as_ref());
```
timvermeulen, line 148:
```suggestion
                            let ps = PathSegment::from(k.as_ref());
```

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        match selector.interests() {
+            Some(interests) => {
+                for ps in interests {
+                    let v = match ipld.lookup_segment(&ps) {
+                        Some(ipld) => ipld,
+                        None => continue,
+                    };
+                    self.clone()
```
timvermeulen, line 138: I'm not really stoked about having to clone `self` here, I'm assuming it's to not mess up the path? My main gripe with this is that it puts a `Clone` bound on the link resolver which shouldn't really be necessary. What if we just call `traverse_node` on `self` here, and inside the `traverse_node` body we pop the path segment from the path again after the recursive call in order to restore it?

ipld/src/path.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// Returns slice of `PathSegment`s of the `Path`.
+    pub fn segments(&self) -> &[PathSegment] {
+        &self.segments
+    }
+
+    /// Appends a `PathSegment` to the end of the `Path`.
+    pub fn append(&mut self, seg: PathSegment) {
+        self.segments.push(seg)
+    }
```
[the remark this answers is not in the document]
austinabell: yeah I was matching go wording, I'll switch tho

ipld/src/selector/walk.rs
```diff
@@ -0,0 +1,208 @@
...
+use async_trait::async_trait;
+use cid::Cid;
+
+impl Selector {
+    /// Walks all nodes visited (not just matched nodes) and executes callback with progress and
+    /// Ipld node. An optional link loader/ resolver is passed in to be able to traverse links.
+    pub async fn walk_all<L, F>(
+        self,
+        ipld: &Ipld,
+        resolver: Option<L>,
```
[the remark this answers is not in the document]
austinabell: for what benefit? That just seems less readable

ipld/src/selector/walk.rs
austinabell: idk thoughts were it's more readable, I couldn't care less though so I'll just change. It's obviously functionally the same

ipld/src/selector/walk.rs
austinabell: Nah there will be other methods added to this if/when they are needed, makes sense to just default to not implemented so someone implementing their own resolver doesn't have to specify the functions if they don't want to use

ipld/src/selector/walk.rs
austinabell: no, because if it tries to explore a field on it, it will stop traversing and return Ok silently, when it should have resolved the cid

ipld/src/selector/walk.rs
austinabell: lol

ipld/src/selector/walk.rs
austinabell: yeah exactly, it's necessary, if you don't then the path will be modified and appended to the path. The reason this can't just be iteratively done is that the paths can branch off and explore recursively. In theory we could probably switch to not having any of this recursively done, but until it's benchmarked and we actually need the performance, it's not worth it

ipld/src/selector/walk.rs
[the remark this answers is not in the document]
austinabell: Like, what I mean by this is when you don't specify a link resolver the type inference is either more verbose or you pass in something other than `None` into the function and that be less readable. I think usability is more important than possibly some checks removed in the code (and less prone to unexpected behaviour if expanded on)

ipld/src/selector/walk.rs
timvermeulen: Should it return an error if you don't implement it though, rather than returning `Ok(None)`?

ipld/src/selector/walk.rs
timvermeulen: I think doing it recursively is fine. calling something like `self.path.pop();` at the end of `traverse_node` always restores the path to its previous state and it would get rid of a bunch of `Clone` bounds, so the suggestion is not just for performance reasons but mainly from an API perspective

ipld/src/selector/walk.rs
austinabell: Yeah so I'm not a huge fan of the setup, but the intention is either you don't specify a resolver and those will just be skipped, or you do specify a resolver and if it requires that function for resolving, then it would error. 

Thinking about it more maybe it is just best to return None for loading links, I was just thinking ahead to the other functions like storing links.

ipld/src/selector/walk.rs
austinabell: Yeah actually great suggestion, I got too influenced from the go impl there and had assumed it was necessary.

ipld/src/selector/walk.rs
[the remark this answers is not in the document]
timvermeulen: I guess what I'm saying is that it's double that the link resolver is optional, and the ipld it returns is also optional. if you require an `L` and use `()` or some other type instead of `None` then the code becomes simpler, I've tried it — I'm also happy to make this change later if my description isn't really clear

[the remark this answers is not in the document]
austinabell: Yeah but then it's not obvious that the resolver is optional and a little janky to read when you are passing () or otherwise in as a parameter. That is specifically what `Option` is for, plus also you are auto deriving the trait impl for every linkresolver impl.

Maybe I'm misunderstanding a detail of what you are saying. I am open to considering, it just seems like an opinionated and less idiomatic API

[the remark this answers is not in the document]
timvermeulen: >That is specifically what `Option` is for

I would definitely agree in the case of a concrete type; the annoying thing about `Option` in combination with generics though is that the `None` is still generic over some `T` that you don't really care about when you don't want to pass a link resolver at all.

Anyway, this whole API is inherently not very idiomatic Rust because of how directly it was ported from Go, so it's not a big deal whether this part is idiomatic or not. This is definitely not a blocker, but rather something to think about when we decide to refactor it.

ipld/src/selector/walk.rs
timvermeulen: If quietly skipping links is not deemed acceptable then I would generally speaking rather have the compiler tell me about it than seeing some error at runtime. But I'm having a hard time seeing how this API will evolve exactly so I can't say for sure which approach is the most logical

austinabell: Ah ya mb I forgot about this and had merged in, definitely open to switching it to not default, just for now doesn't matter because this isn't an API used outside of our codebase yet

PR #412, opened 2020-05-12
Update tipset sorting

austinabell:
**Summary of changes**
No need to clone or generate cid bytes for every sort 

Closes 

### DIFF
--- a/blockchain/blocks/src/header.rs
+++ b/blockchain/blocks/src/header.rs
@@ -18,6 +18,7 @@ use num_bigint::{
     BigUint,
 };
 use sha2::Digest;
+use std::cmp::Ordering;
 use std::fmt;
 use std::time::{SystemTime, UNIX_EPOCH};
 use vm::PoStProof;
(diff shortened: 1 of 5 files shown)


blockchain/blocks/src/header.rs
```diff
@@ -201,6 +202,22 @@ impl<'de> Deserialize<'de> for BlockHeader {
     }
 }
 
+impl Ord for BlockHeader {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.ticket().cmp(other.ticket()) {
+            // Only compare cid bytes when tickets are equal
+            Ordering::Equal => self.cid().to_bytes().cmp(&other.cid().to_bytes()),
+            order => order,
+        }
```
timvermeulen, line 211:
```suggestion
        // Only compare cid bytes when tickets are equal
        self.ticket()
            .cmp(other.ticket())
            .then_with(|| self.cid().to_bytes().cmp(&other.cid().to_bytes()))
```
Good stuff, just wanted to bring to your attention the `Ordering::{then, then_with}` methods that serve exactly the purpose of composing orderings

Issue #392, opened 2020-05-04
Verified registry builtin actor


austinabell:
**Task summary**

New builtin actor: https://github.com/filecoin-project/specs-actors/tree/master/actors/builtin/verifreg

Should be pretty straightforward, and have examples from other builtin actor impls for everything in the impl from skimming over the go code

**Specification reference**

- 

**Other information and links**

PR #413, opened 2020-05-13
Implement Verify Registry Actor

AshantiMutinta:
**Summary of changes**
- Implemented Verifyreg actor according to go standard

Closes 
#392 

https://github.com/filecoin-project/specs-actors/tree/master/actors/builtin/verifreg


vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+pub struct Actor;
+impl Actor {
+    /// Constructor for Registry Actor
+    pub fn constructor<BS, RT>(rt: &mut RT, root_key: Address) -> Result<(), ActorError>
+    where
+        BS: BlockStore,
+        RT: Runtime<BS>,
+    {
+        rt.validate_immediate_caller_is(std::iter::once(&*SYSTEM_ACTOR_ADDR))?;
+        let empty_root = Hamt::<String, _>::new_with_bit_width(rt.store(), HAMT_BIT_WIDTH)
```
austinabell, line 40: Although it doesn't functionally matter, having string keys for this was just something outdated (Hamt used to have string keys), would be better to use the actual key type used in state

```suggestion
        let empty_root = Hamt::<BytesKey, _>::new_with_bit_width(rt.store(), HAMT_BIT_WIDTH)
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+                        format!("failed to add verifier: {:}", e),
+                    )
+                })?;
+            Ok(())
+        })??;
+
+        Ok(())
+    }
+
+    pub fn delete_verifier<BS, RT>(rt: &mut RT, params: AddVerifierParams) -> Result<(), ActorError>
```
austinabell, line 75: Best to just keep naming consistent (confusing to have a method name that differs 
```suggestion
    pub fn remove_verifier<BS, RT>(rt: &mut RT, params: AddVerifierParams) -> Result<(), ActorError>
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+    ) -> Result<(), ActorError>
+    where
+        BS: BlockStore,
+        RT: Runtime<BS>,
+    {
+        if params.allowance
+            <= Datacap::new(
+                MINIMUM_VERIFIED_SIZE
+                    .to_biguint()
+                    .expect("Could not convert size to big int"),
```
austinabell, line 109: Shouldn't need to convert just to do the comparison (I will comment next on the Datacap type)

vm/actor/src/builtin/verifreg/types.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub allowance: Datacap,
+}
+
+pub type AddVerifierParams = VerifierParams;
+pub type AddVerifierClientParams = VerifierParams;
+
+pub const MINIMUM_VERIFIED_SIZE: u32 = 1 << 20;
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Datacap(StoragePower);
```
austinabell, line 26: Why is this not just an alias? There is no functionality added on top of this type that you don't get already, and you miss out on the BigUint API

vm/actor/src/builtin/verifreg/types.rs
```diff
@@ -0,0 +1,82 @@
...
+
+impl<'de> Deserialize<'de> for Datacap {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let BigUintDe(datacap) = Deserialize::deserialize(deserializer)?;
+        Ok(Self(datacap))
+    }
+}
```
austinabell, line 82: I assume this is why no aliasing, but I think it's much cleaner if you just wrap the puts and gets with this instead (only two places used, which is a lot less overhead than all of this)

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+            return Err(ActorError::new(
+                ExitCode::ErrIllegalState,
+                format!(
+                    "Allowance {:} below MinVerifiedDealSize for add verified client {:}",
+                    params.allowance, params.address
+                ),
+            ));
+        }
+
+        rt.transaction::<_, Result<_, ActorError>, _>(|st: &mut State, rt| {
```
austinabell, line 121:
```suggestion
        rt.transaction(|st: &mut State, rt| {
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+            return Err(ActorError::new(
+                ExitCode::ErrIllegalState,
+                format!(
+                    "Verified Dealsize {:} is below minimum in usedbytes",
+                    params.deal_size
+                ),
+            ));
+        }
+
+        rt.transaction::<_, Result<_, ActorError>, _>(|st: &mut State, rt| {
```
austinabell, line 207:
```suggestion
        rt.transaction(|st: &mut State, rt| {
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                .get_verifier(rt.store(), &params.address)
+                .map_err(|_| {
+                    ActorError::new(
+                        ExitCode::ErrIllegalState,
+                        format!("Failed to get Verifier {:?}", &params.address),
+                    )
+                })?
+                .ok_or_else(|| {
+                    ActorError::new(
+                        ExitCode::ErrNotFound,
```
austinabell, line 218:
```suggestion
                        ExitCode::ErrIllegalArgument,
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                        format!("Invalid Verifier {:}", params.address),
+                    )
+                })?;
+
+            if verifier_cap >= Zero::zero() {
+                panic!("new verifier cap should be greater than or equal to 0");
+            }
```
austinabell, line 225: This will always be true, the type is a biguint. Aside from this, should avoid panic in the actors, would be best to avoid catching panics whenever possible

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+            let new_verifier_cap = verifier_cap - params.deal_size.clone();
```
austinabell, line 226:
```suggestion
            let new_verifier_cap = verifier_cap - &params.deal_size;
```

Also, is there a guarantee that deal size is less than verifier_cap? if not this will panic

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                    })
+            }
+        })??;
+
+        Ok(())
+    }
+
+    // Called by HandleInitTimeoutDeals from StorageMarketActor when a VerifiedDeal fails to init.
+    // Restore allowable cap for the client, creating new entry if the client has been deleted.
+    pub fn restore_bytes<BS, RT>(rt: &mut RT, params: UseBytesParams) -> Result<(), ActorError>
```
austinabell, line 260:
```suggestion
    pub fn restore_bytes<BS, RT>(rt: &mut RT, params: RestoreBytesParams) -> Result<(), ActorError>
```

Looks like that type is already setup

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+    where
+        BS: BlockStore,
+        RT: Runtime<BS>,
+    {
+        rt.validate_immediate_caller_is(std::iter::once(&*STORAGE_MARKET_ACTOR_ADDR))?;
+        if params.deal_size < MINIMUM_VERIFIED_SIZE.into() {
+            return Err(ActorError::new(
+                ExitCode::ErrIllegalState,
```
austinabell, line 268:
```suggestion
                ExitCode::ErrIllegalArgument,
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                format!(
+                    "Verified Dealsize {:} is below minimum in usedbytes",
+                    params.deal_size
+                ),
+            ));
+        }
+
+        rt.transaction::<_, Result<_, ActorError>, _>(|st: &mut State, rt| {
```
austinabell, line 276:
```suggestion
        rt.transaction(|st: &mut State, rt| {
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    pub fn add_verified_client<BS, RT>(
+        rt: &mut RT,
+        params: AddVerifierClientParams,
+    ) -> Result<(), ActorError>
+    where
+        BS: BlockStore,
+        RT: Runtime<BS>,
+    {
+        if params.allowance <= MINIMUM_VERIFIED_SIZE.into() {
```
austinabell, line 105: I can't remember rn if there was a comparison for this, but if not, maybe worth creating a static BigUint so you aren't converting to biguint so many times in this file just for comparison

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                .get_verifier(rt.store(), &params.address)
+                .map_err(|_| {
+                    ActorError::new(
+                        ExitCode::ErrIllegalState,
+                        format!("Failed to get Verifier {:?}", params.address.clone()),
+                    )
+                })?
+                .unwrap_or_else(|| MINIMUM_VERIFIED_SIZE.into());
+
+            let new_verifier_cap = verifier_cap + params.deal_size.clone();
```
austinabell, line 287:
```suggestion
            let new_verifier_cap = verifier_cap + &params.deal_size;
```

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                format!(
+                    "Verified Dealsize {:} is below minimum in usedbytes",
+                    params.deal_size
+                ),
+            ));
+        }
+
+        rt.transaction::<_, Result<_, ActorError>, _>(|st: &mut State, rt| {
+            let verifier_cap = st
+                .get_verifier(rt.store(), &params.address)
```
austinabell, line 278: Shouldn't this be `get_verified_client`?

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                .map_err(|_| {
+                    ActorError::new(
+                        ExitCode::ErrIllegalState,
+                        format!("Failed to get Verifier {:?}", params.address.clone()),
+                    )
+                })?
+                .unwrap_or_else(|| MINIMUM_VERIFIED_SIZE.into());
```
austinabell, line 285: This differs from the specs-actors, reason? Seems like it should be 0

vm/actor/src/builtin/verifreg/state.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn put<BS: BlockStore>(
+        storage: &mut Cid,
+        store: &BS,
+        verified_addr: &Address,
+        verifier_cap: &Datacap,
+    ) -> StateResult<()> {
+        let mut map: Hamt<BytesKey, _> =
+            Hamt::load_with_bit_width(&storage, store, HAMT_BIT_WIDTH)?;
+        map.set(verified_addr.to_bytes().into(), BigUintSer(&verifier_cap))?;
+        map.flush()?;
```
austinabell, line 98: you need to use the Cid from this flush and update the storage cid

vm/actor/src/builtin/verifreg/state.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    fn delete<BS: BlockStore>(
+        storage: &mut Cid,
+        store: &BS,
+        verified_addr: &Address,
+    ) -> StateResult<()> {
+        let mut map: Hamt<BytesKey, _> =
+            Hamt::load_with_bit_width(&storage, store, HAMT_BIT_WIDTH)?;
+        map.delete(&verified_addr.to_bytes())?;
+        map.flush()?;
```
austinabell, line 121: same with this one, the root cid is not updated

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+pub enum Method {
+    Constructor = METHOD_CONSTRUCTOR,
+    AddVerifier = 2,
+    RemoveVerifier = 3,
+    AddVerifiedClient = 4,
+    UseBytes = 5,
+    RestoreBytes = 6,
+}
+
+pub struct Actor;
```
austinabell, line 31: need to impl `ActorCode` for this at the bottom of the file, once this is done, you can hook it up to be dispatched from the vm as the other actors are (ping me if you want clarification)

vm/actor/src/builtin/verifreg/state.rs
```diff
@@ -0,0 +1,149 @@
...
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use encoding::Cbor;
+use ipld_blockstore::BlockStore;
+use ipld_hamt::{BytesKey, Hamt};
+use num_bigint::biguint_ser::{BigUintDe, BigUintSer};
+
+use crate::builtin::verifreg::types::Datacap;
+
+#[derive(Default)]
```
austinabell, line 16:
```suggestion
```

If you pulled this from another actor, it's outdated, this is now not ideal

vm/actor/src/builtin/verifreg/types.rs
```diff
@@ -0,0 +1,27 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct VerifierParams {
+    pub address: Address,
+    pub allowance: Datacap,
+}
+
+pub type AddVerifierParams = VerifierParams;
+pub type AddVerifierClientParams = VerifierParams;
+
+pub const MINIMUM_VERIFIED_SIZE: u32 = 1 << 20;
```
austinabell, line 16: nit: can you move this to the top (nice to have constants always defined at top of a file)

vm/actor/src/builtin/verifreg/mod.rs
AshantiMutinta: Copied it directly from go but will remove this

vm/actor/src/builtin/verifreg/mod.rs
AshantiMutinta: Does bigunit panic on subs? overflows I guess?

vm/actor/src/builtin/verifreg/mod.rs
AshantiMutinta: Was a question of inline vs non-inline.

Will change to static

austinabell: yeah this was a more opinionated thing, I'm actually not sure which would be better after it goes through the compiler. In theory you should be able to compare a bigint with a u32 without doing any conversion, but maybe that interface doesn't exist

vm/actor/src/builtin/verifreg/mod.rs
austinabell: exactly! a sub that underflows will panic in rust (in this case the unsigned being subtracted below 0)

vm/actor/src/builtin/verifreg/state.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn delete<BS: BlockStore>(
+        storage: &mut Cid,
+        store: &BS,
+        verified_addr: &Address,
+    ) -> StateResult<Cid> {
+        let mut map: Hamt<BytesKey, _> =
+            Hamt::load_with_bit_width(&storage, store, HAMT_BIT_WIDTH)?;
+        map.delete(&verified_addr.to_bytes())?;
+        let root = map.flush()?;
+        Ok(root)
```
austinabell, line 126: Shouldn't need to pass the root back if you are exposing a mutable reference to the storage cid in this function, could just do something like:
```suggestion
        *storage = map.flush()?;
        Ok(())
```

Or am I missing something? if not then you can at least change storage to be immutable ref

vm/actor/src/builtin/verifreg/state.rs
```diff
@@ -0,0 +1,153 @@
...
+        storage: &mut Cid,
+        store: &BS,
+        verified_addr: &Address,
+        verifier_cap: &Datacap,
+    ) -> StateResult<Cid> {
+        let mut map: Hamt<BytesKey, _> =
+            Hamt::load_with_bit_width(&storage, store, HAMT_BIT_WIDTH)?;
+        map.set(verified_addr.to_bytes().into(), BigUintSer(&verifier_cap))?;
+        let root = map.flush()?;
+        Ok(root)
```
austinabell, line 103: same as other comment

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+use address::Address;
+use ipld_blockstore::BlockStore;
+use ipld_hamt::BytesKey;
+use ipld_hamt::Hamt;
+use message::Message;
+use num_derive::FromPrimitive;
+use num_traits::{FromPrimitive, Zero};
+use runtime::{ActorCode, Runtime};
+use vm::{ActorError, ExitCode, MethodNum, Serialized, METHOD_CONSTRUCTOR};
+/// Account actor methods available
```
austinabell, line 19: Nit: if possible just add a line break between this

vm/actor/src/builtin/verifreg/state.rs
AshantiMutinta: This is essentially the part with the helper function. The helper function for the verification and verification client. Returning the root just adds a contract with a caller

AshantiMutinta: The flushing is happening in the helper function.  for the put and delete. 

Forgot to remove the mut references

AshantiMutinta: I guess the HAMT functions already put some side effects in so it is a debate of whether just because we have some side effects, should we put some more basically eitherway fine with any method

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+            }
+            Some(Method::UseBytes) => {
+                Self::use_bytes(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
+            }
+            Some(Method::RestoreBytes) => {
+                Self::restore_bytes(rt, params.deserialize()?)?;
+                Ok(Serialized::default())
+            }
+            _ => Err(rt.abort(ExitCode::SysErrInvalidMethod, "Invalid method".to_owned())),
```
timvermeulen, line 346: This only covers the `None` case, right? If so, I'd suggest changing this to `None => ...` to make this more obvious, and to not introduce bugs when e.g. a new enum variant is added

vm/actor/src/builtin/verifreg/state.rs
austinabell: side effects you are referring to writing to blockstore on flush? Doing this only modifies the cid, and it's being changed still just after the cid is returned. And to clarify this change does not invalidate the previous cid in any way.

I'm good with this now, just curious the thought process behind this (interesting to get the functional perspective to this)

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+                return Err(ActorError::new(
+                    ExitCode::ErrIllegalArgument,
+                    format!(
+                        "Add more DataCap {:} for VerifiedClient than allocated {:}",
+                        params.allowance, verifier_cap
+                    ),
+                ));
+            }
+            let new_verifier_cap = verifier_cap - &params.allowance;
+            st.put_verifier(rt.store(), &*verify_addr, &new_verifier_cap)
```
austinabell, line 147:
```suggestion
            st.put_verifier(rt.store(), &verify_addr, &new_verifier_cap)
```

vm/actor/src/builtin/verifreg/state.rs
AshantiMutinta: Well from a pure functional perspective we wouldn't even be making changes to the HAMT,we would essentially be building a new HAMT strucutre with every operation that we do. A bit overhead here but if we had to take not so much of a deep step functionality, I would rather reduce the amount of mut parameters that these function takes and isolate them a bit at a higher level. The general idea is that if you give these functions certain number of parameters, you know what you are going to get out i mean if the HAMT was a functional data structure this would be more true but the HAMT's version of mutability is a bit more isolated which I think it is fine

austinabell: That is what's happening, the Hamt is being built for every operation (it's very cheap to load because you are just pulling the root node into memory).

That makes sense though to have the mutability further up

vm/actor/src/builtin/verifreg/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        if params.allowance <= *MINIMUM_VERIFIED_SIZE {
+            return Err(ActorError::new(
+                ExitCode::ErrIllegalArgument,
+                format!(
+                    "Allowance {:} below MinVerifiedDealSize for add verified client {:}",
+                    params.allowance, params.address
+                ),
+            ));
+        }
+
```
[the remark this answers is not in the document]
AshantiMutinta: Just changed! Thanks!

Issue #344, opened 2020-04-13
Implement Randomness 


amerameen:
In the Lotus implementation, they use a RAND library within the VM context, we need to implement the equivalent.

amerameen:
@dutterbutter @austinabell can you add some context here when you get a chance?

Issue #162, opened 2020-01-14
Implement Ticket Randomness


dutterbutter:
Although the verifier node is not responsible for generating a ticket, based on the spec it seems that the ability to generate randomness from a ticket is still required. 

Spec definition:
- https://filecoin-project.github.io/specs/#the-ticket-chain-and-drawing-randomness

**Methods required defined as per spec:**
- `randomness_at_epoch(chainEpoch)`
- `get_ticket_production_rand_seed(ChainEpoch)`
- `get_seal_rand_seed(chainEpoch)`
- `get_PoSt_challenge_rand_seed(chainEpoch)`

The above means that ticket randomness is reseeded with every new block, but can indeed be derived by any miner for an arbitrary epoch number using a past epoch.

Miner runs the prior ticket through a VRF to get a new unique ticket which can later be derived for randomness (chain methods described above). The prior ticket is prepended with the ticket domain separation tag and concatenated with the miner actor address (to ensure miners using the same worker keys get different randomness).

**To generate a ticket for a given epoch n:**

```
LastTicket = MinTicketValueAtEpoch(n-1)
newRandomness = VRF_miner(H(TicketProdDST || index || Serialization(pastTicket, minerActorAddress)))
```

See [Randomness.md](https://github.com/filecoin-project/specs/blob/master/src/algorithms/crypto/randomness.md)

PR #415, opened 2020-05-14
Runtime randomness and ChainStore randomness

ec2:
**Summary of changes**
- Introduced deriving (pseudo)randomness in the Runtime and ChainStore.

Closes 
#344, #162 


ec2:
Worth noting that the methods that are descibed in #162 might be outdated. I couldn't find them in Lotus... So I've only implemented the randomness method i could find in the chainstore and the runtime

blockchain/chain/src/store/chain_store.rs
```diff
@@ -273,6 +278,52 @@ where
...
+    }
+}
+
+/// Computes a pseudorandom 32 byte Vec
+fn draw_randomness(
+    rbase: &[u8],
+    pers: DomainSeparationTag,
+    round: ChainEpoch,
+    entropy: &[u8],
+) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
```
austinabell, line 319: Can you just make the return a 32 byte array to remove the need for allocations?

vm/interpreter/src/default_runtime.rs
```diff
@@ -209,11 +213,21 @@ where
...
+        entropy: &[u8],
+    ) -> Result<Randomness, ActorError> {
+        let r = self
+            .rand
+            .get_randomness(&self.store, personalization, rand_epoch, entropy)
+            .map_err(|e| {
+                ActorError::new_fatal(format!("could not get randomness: {}", e.to_string()))
+            })?;
+        let mut array = [0; 32];
+        let bytes = &r[..r.len()];
```
austinabell, line 228: did you mean to do array.len()? Irrelevant if you change based on previous comment

ec2: Oh real nice catch....

blockchain/chain/src/store/chain_store.rs
```diff
@@ -273,6 +278,54 @@ where
         out += &value;
         Ok(out)
     }
+
+    pub fn get_randomness_block(
```
dutterbutter, line 282: Nit: please add comment

blockchain/chain/src/store/chain_store.rs
```diff
@@ -210,6 +215,18 @@ where
...
+    pub fn get_randomness(
+        &self,
+        blocks: &TipsetKeys,
+        pers: DomainSeparationTag,
+        round: ChainEpoch,
+        entropy: &[u8],
+    ) -> Result<[u8; 32], Error> {
+        get_randomness(self.db.as_ref(), blocks, pers, round, entropy)
+            .map_err(|e| Error::Other(e.to_string()))
+    }
```
austinabell, line 229: Things are going to get pretty messy and confusing if we duplicate all of these, any specific reason to have it like this? Another option is to make the get_randomness function associated but use a ref to blockstore instead of `&self`

```suggestion
```

blockchain/chain/src/store/chain_store.rs
```diff
@@ -248,6 +265,43 @@ where
...
+
+/// Computes a pseudorandom 32 byte Vec
+fn draw_randomness(
+    rbase: &[u8],
+    pers: DomainSeparationTag,
+    round: ChainEpoch,
+    entropy: &[u8],
+) -> Result<[u8; 32], Box<dyn std::error::Error>> {
+    let mut state = Params::new().hash_length(32).to_state();
+    state.write_i64::<BigEndian>(pers as i64)?;
```
austinabell, line 295: can you make the DomainSeparationTag repr a i64 just to ensure this is consistent?

vm/interpreter/src/default_runtime.rs
```diff
@@ -323,6 +335,7 @@ where
                 self.origin,
                 self.origin_nonce,
                 self.num_actors_created,
+                self.rand.clone(),
```
austinabell, line 338: I know it's painful to add another lifetime, but there is no need to clone this, and considering how many times this will be called and it's a non-trivial amount of data, should probably switch.

I made changes needed here (sorry I also included other comments stuff) https://github.com/ChainSafe/forest/compare/austin/alt/randomness

vm/interpreter/src/vm.rs
```diff
@@ -361,6 +363,7 @@ where
             *msg.from(),
             msg.sequence(),
             0,
+            self.rand.clone(),
```
austinabell, line 366: just to note other unnecessary clone of this

blockchain/chain/src/store/chain_store.rs
ec2: Ill remove, I agree

vm/interpreter/src/default_runtime.rs
ec2: ill cherrypick that commit onto here

blockchain/chain/src/store/chain_store.rs
```diff
@@ -248,6 +253,43 @@ where
...
+    pers: DomainSeparationTag,
+    round: ChainEpoch,
+    entropy: &[u8],
+) -> Result<[u8; 32], Box<dyn std::error::Error>> {
+    let mut blks = blocks.clone();
+    loop {
+        let nts = tipset_from_keys(db, &blks)?;
+        let mtb = nts.min_ticket_block();
+        if nts.epoch() <= round || mtb.epoch() == 0 {
+            return draw_randomness(mtb.ticket().vrfproof.bytes(), pers, round, entropy);
```
austinabell, line 269:
```suggestion
            return draw_randomness(mtb.ticket().vrfproof.as_bytes(), pers, round, entropy);
```

blockchain/blocks/src/tipset.rs
```diff
@@ -107,6 +107,10 @@ impl Tipset {
     pub fn min_ticket(&self) -> Ticket {
         self.first_block().ticket().clone()
     }
+    /// Returns the block with the smallest ticket of all blocks in the tipset
+    pub fn min_ticket_block(&self) -> &BlockHeader {
+        &self.first_block()
+    }
```
timvermeulen, line 113: If all this does is expose `first_block`, maybe get rid of `first_block` altogether? This is the better name anyway

austinabell: oops yeah I missed that, no need for both

ec2: Yeah, I kinda just left it there before i wasn't sure if they person who put it in intents to use it for something. Ill Remove first_block

timvermeulen: I put it there during a refactor because it was used several times within the file

ec2: ok i removed first_block in favour of min_ticket_block

PR #418, opened 2020-05-15
Last block info for selectors

### LINKED ISSUE #414: Include LastBlock info for selector traversals
**Issue summary**

I avoided this before because I thought it was unnecessary (and there were cases where it would lead to missed data) but is the exact format for graphsync responses.

This issue is to update the lastblock information which gets included in the callback which will be returned as graphsync responses.

The test vectors should be updated as well so that when go impl has a runner for them we can verify that it matches for all cases so there are no differences in the future.

austinabell:
**Summary of changes**
- Needed for graphsync (last block info is sent in responses)
  - Moved test vectors into submodule to be more easily shared

Closes #414 


ipld/src/selector/walk.rs
```diff
@@ -98,6 +116,10 @@ where
         // Resolve any links transparently before traversing
         if let Ipld::Link(cid) = ipld {
             if let Some(resolver) = &self.resolver {
+                self.last_block = Some(LastBlockInfo {
+                    path: self.path.clone(),
+                    link: cid.clone(),
+                });
                 let mut node = resolver.load_link(cid).await.map_err(Error::Link)?;
                 while let Some(Ipld::Link(c)) = node {
                     node = resolver.load_link(&c).await.map_err(Error::Link)?;
```
timvermeulen, line 125: Would `self.last_block`'s CID not need to be updated here as well?

ipld/src/selector/mod.rs
```diff
@@ -3,7 +3,7 @@
 
 mod empty_map;
 mod walk;
-pub use walk::{LinkResolver, Progress, VisitReason};
+pub use self::walk::*;
```
timvermeulen, line 6:
```suggestion
pub use walk::*;
```

ipld/src/selector/walk.rs
austinabell: no, so the intention is that the first link traversed is recorded, if you provide the most recent when there are multiple traversals, you could lose the connection from the initial node.

This case is never really hit in practice, because we never link to an ipld node that is a link, but this just handles that edge case so traversal doesn't stop if this is ever found.

ipld/src/selector/mod.rs
austinabell: why? This is more clear about what is being exported

timvermeulen: Ah, like, to indicate it's not from another crate or something? I thought you put it there by accident, didn't think about it too hard

ipld/src/selector/walk.rs
timvermeulen: >the intention is that the first link traversed is recorded

It seems like it's still possible here to traverse several links with maybe some other layers in between, and it will only save the last one. Or do you in that case actually want to save the last one only?

austinabell: But what I mean, is that if you save the last link traversed instead of the first, you can never know how many links were traversed and the last block would just be an Ipld Link, which isn't useful because you'd have to do some custom logic to check the previously visited block to get the original block, and that may not be available if you are only walking matched nodes.

This is all very in depth unnecessarily because this case is never hit in practice for our uses, it's just a benefit to allow this over the go implementations stopping traversal.

Issue #410, opened 2020-05-12
Asyncify some verifications of Blocks in the ChainSyncer


ec2:
**Issue summary**

There are some checks in the ChainSyncer's `validate(&self, block: &Block) -> Result<(), Error>` method that could use some async. 
This is a nice optimization to have, since checks don't have to happen sequentially. Furthermore, currently when one check fails, the whole method fails. This implies that we can only recover one error per block. When done asynchronously, we can pass back multiple errors, giving us a more rich error. 

Some checks that can be async are:
Miner Check, Signature Check, Message Check, Winner Check

**Other information and links**

PR #419, opened 2020-05-16
Jaden/chainsync/asyncverification

flodesi:
**Summary of changes**
- Added async to compute heavy checks in ChainSyncer's `validate` method
- Created a `FuturesUnordered`, as well as implemented `spawn_blocking` for the async checks
- Modified error handling in ChainSyncer's `validate` function to create a more rich error

#410


blockchain/chain_sync/src/sync.rs
```diff
@@ -617,48 +612,106 @@ where
...
         let parent_tipset = self.chain_store.tipset_from_keys(header.parents())?;
 
         // time stamp checks
-        header.validate_timestamps(&parent_tipset)?;
+        let time_stamp_check = header.validate_timestamps(&parent_tipset);
+        if time_stamp_check.is_err() {
+            error_vec.push(time_stamp_check.err().unwrap().to_string());
+        }
 
         // check messages to ensure valid state transitions
```
timvermeulen, line 633: This comment is now in the wrong spot, I think

blockchain/chain_sync/src/sync.rs
```diff
@@ -617,48 +612,106 @@ where
...
+                Some(result) => {
+                    if result.is_err() {
+                        error_vec.push(result.err().unwrap().to_string());
+                    }
+                }
+                None => {
+                    break;
+                }
+            }
+        }
```
timvermeulen, line 708:
```suggestion
        while let Some(result) = validations.next().await {
            if let Err(err) = result {
                error_vec.push(err.to_string());
            }
        }
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -448,26 +449,24 @@ where
...
             secp_message_root: secp_root,
         };
         // TODO this should be memoryDB for temp storage
         // store message roots and receive meta_root
-        let meta_root = self
-            .chain_store
-            .blockstore()
+        let meta_root = temp_store
             .put(&meta, Blake2b256)
             .map_err(|e| Error::Other(e.to_string()))?;
```
austinabell, line 471: you're going to need to put this back to what it was (or pass in the blockstore in place of where self was) this will not work. You're querying the messages from an empty temporary store then not putting the metadata in the store (temp storage is dropped at the end of this)

blockchain/state_manager/src/lib.rs
```diff
@@ -129,4 +129,8 @@ where
         }
         Ok(kaddr.payload_bytes())
     }
+
+    pub fn get_db(&self) -> Arc<DB> {
+        Arc::clone(&self.bs)
+    }
```
austinabell, line 135: unused

blockchain/chain_sync/src/sync.rs
```diff
@@ -947,7 +992,7 @@ mod tests {
             Cid::from_raw_cid("bafy2bzacecujyfvb74s7xxnlajidxpgcpk6abyatk62dlhgq6gcob3iixhgom")
                 .unwrap();
 
-        let root = cs.compute_msg_data(&[bls], &[secp]).unwrap();
+        let root = ChainSyncer::<MemoryDB>::compute_msg_data(&[bls], &[secp]).unwrap();
```
austinabell, line 995: Reason for it being associated?

flodesi: For the `compute_msg_data` function, `self` had to be removed as one of the parameters in order to get `spawn_blocking` to work. As a result of this, `root` was refactored in that way.

blockchain/chain_sync/src/sync.rs
```diff
@@ -510,16 +513,13 @@ where
...
-        for m in block.bls_msgs() {
-            let pk = self
-                .state_manager
-                .get_bls_public_key(m.from(), tip.parent_state())?;
-            pub_keys.push(pk);
-            cids.push(m.cid()?.to_bytes());
-        }
+    fn check_block_msgs(
+        blockstore: &DB,
+        db: Arc<DB>,
```
austinabell, line 518: You don't need both, these are the same (a note about this is references are needed for things that are done syncronously and you only need to clone the `Arc<DB>` when the action needs to be done in another thread)

blockchain/chain_sync/src/sync.rs
```diff
@@ -947,7 +999,9 @@ mod tests {
             Cid::from_raw_cid("bafy2bzacecujyfvb74s7xxnlajidxpgcpk6abyatk62dlhgq6gcob3iixhgom")
                 .unwrap();
 
-        let root = cs.compute_msg_data(&[bls], &[secp]).unwrap();
+        let root =
+            ChainSyncer::<MemoryDB>::compute_msg_data(cs.chain_store.blockstore(), &[bls], &[secp])
```
austinabell, line 1003: I'm a little lost on how this even compiles tbh
```suggestion
            ChainSyncer::compute_msg_data(cs.chain_store.blockstore(), &[bls], &[secp])
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -133,7 +133,7 @@ where
 
 impl<DB> ChainSyncer<DB>
 where
-    DB: BlockStore,
+    DB: BlockStore + Sync + Send + 'static,
 {
```
austinabell, line 137: Sorry to comment on changes unreleated to yours, but this duplicate declaration is unneeded. There's no reason this is necessary, if you could remove the duplicate declaration and combine that would be awesome!

blockchain/chain_sync/src/sync.rs
```diff
@@ -617,48 +616,98 @@ where
...
+        // Check Block Message and Signatures in them
+        let mut pub_keys = Vec::new();
+        let mut cids = Vec::new();
+        for m in block.bls_msgs() {
+            let pk = self
+                .state_manager
+                .get_bls_public_key(m.from(), parent_tipset.parent_state())?;
+            pub_keys.push(pk);
+            cids.push(m.cid()?.to_bytes());
+        }
```
austinabell, line 647: what was the reason for this being done outside of the `check_block_msgs` call? Only reason I mention is that this isn't done async, and would be beneficial if it was also done in the seperate thread

flodesi: this part of code was put outside the `check_block_msgs` call because in order to get it inside, `self.state_manager` would have to be borrowed, and there is no `clone` method that can be used for this, so the setup of `pub_keys` and `cids` wasn't made async.

austinabell: but the function attached to the state_manager only needs the blockstore, so in similar fashion to what you did here you could also do for that? makes sense for this logic to be done inside of check_block_msgs anyway outside of the performance changes

PR #431, opened 2020-05-19
Update prefix bytes encoding to include mh len

### LINKED ISSUE #428: Cid Prefix to bytes
**Task summary**

Prefix to bytes generation is needed for Graphsync

Use go as a reference:
https://github.com/ipfs/go-cid/blob/master/cid.go#L569

Currently the Prefix type in cid crate does not include (mh_len) because it is very unhelpful and requires some data to be arbitrarily hashed to be determined, so I think the easiest solution would be to just add a function on top of the Cid type to generate the prefix bytes, but if the prefix needs to be reused and the mh_len is needed, that may need to be added back.

**Specification reference**

- 

austinabell:
**Summary of changes**
- I actually have the worst memory, #428 was misleading because it did exist and I had just removed the mh_len during the multihash refactor because it was redundant (at the time) so I added it back
  - Also figured I'd move prefix to own file because that file isn't setup the best (not going to do an actual refactor because hopefully we move to shared rust cid crate in future)

I assume the length is needed for variable length identity hashes, which is why it's included in the prefix bytes.

Closes #428

### DIFF
--- a/ipld/cid/src/prefix.rs
+++ b/ipld/cid/src/prefix.rs
@@ -1,0 +1,53 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use super::{Codec, Error, Version};
+use integer_encoding::{VarIntReader, VarIntWriter};
+use multihash::Code;
+use std::io::Cursor;
+
+/// Prefix represents all metadata of a CID, without the actual content.
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub struct Prefix {
+    pub version: Version,
+    pub codec: Codec,
+    pub mh_type: Code,
+    pub mh_len: usize,
+}
+
+impl Prefix {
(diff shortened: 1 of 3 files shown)


ipld/cid/src/prefix.rs
```diff
@@ -0,0 +1,53 @@
...
+
+        let raw_version = cur.read_varint()?;
+        let raw_codec = cur.read_varint()?;
+        let raw_mh_type: u64 = cur.read_varint()?;
+        let mh_len: usize = cur.read_varint()?;
+
+        let version = Version::from(raw_version)?;
+        let codec = Codec::from(raw_codec)?;
+
+        let mh_type = Code::from_u64(raw_mh_type as u64);
```
timvermeulen, line 31:
```suggestion
        let mh_type = Code::from_u64(raw_mh_type);
```
Seems to be as good a time as any to fix this

PR #432, opened 2020-05-20
Jaden/chainstore refactor

### LINKED ISSUE #430: Convert relevant methods in the ChainStore to functions
**Issue summary**

This is related to building out the RPC. 
We don't want to pass around the ChainStore. We want to pass around the database intstead. But the methods on the ChainStore are very useful as a lot of them match 1 to 1 with the RPC API. The nice thing is, most of those methods do not operate on data from the ChainStore itself, only the DB. So we can create functions that have almost the same logic but takes in DB instead of ChainStore. Then we also convert the Chainstore methods to call these functions instead. 

flodesi:
**Summary of changes**
- Convert methods in ChainStore that depend only on DB to functions that take DB as a parameter and not self.
- General refactoring to functions that relied on the methods described above.

Closes 
#430

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -76,28 +76,12 @@ where
 
     /// Writes genesis to blockstore
     pub fn set_genesis(&mut self, header: BlockHeader) -> Result<(), Error> {
-        self.db.write(GENESIS_KEY, header.marshal_cbor()?)?;
-        Ok(self.persist_headers(&[header])?)
-    }
-
-    /// Writes encoded blockheader data to blockstore
-    fn persist_headers(&mut self, bh: &[BlockHeader]) -> Result<(), Error> {
-        let mut raw_header_data = Vec::new();
-        let mut keys = Vec::new();
-        // loop through block to push blockheader raw data and cid into vector to be stored
-        for header in bh {
-            if !self.db.exists(header.cid().key())? {
-                raw_header_data.push(header.marshal_cbor()?);
-                keys.push(header.cid().key());
-            }
-        }
-
-        Ok(self.db.bulk_write(&keys, &raw_header_data)?)
+        set_genesis(self.blockstore(), header)
     }
 
     /// Writes tipset block headers to data store and updates heaviest tipset
     pub fn put_tipsets(&mut self, ts: &Tipset) -> Result<(), Error> {
-        self.persist_headers(ts.blocks())?;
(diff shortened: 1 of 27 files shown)


blockchain/chain/src/store/chain_store.rs
```diff
@@ -76,49 +76,25 @@ where
 
     /// Writes genesis to blockstore
     pub fn set_genesis(&mut self, header: BlockHeader) -> Result<(), Error> {
```
austinabell, line 78:
```suggestion
    pub fn set_genesis(&self, header: BlockHeader) -> Result<(), Error> {
```
This doesn't really matter but didn't need mutable ref (maybe we were storing the genesis tipset on the chain store before?)

blockchain/chain/src/store/chain_store.rs
```diff
@@ -246,33 +188,46 @@ where
...
+where
+    DB: BlockStore,
+{
+    for m in msgs {
+        let key = m.cid()?.key();
+        let value = &m.marshal_cbor()?;
+        if db.exists(&key)? {
+            return Ok(());
+        }
+        db.write(&key, value)?
```
austinabell, line 228: Idk how this was missed on initial review, if you could pls fix :)

```suggestion
        if !db.exists(&key)? {
            db.write(&key, value)?
        }
```

blockchain/chain/src/store/chain_store.rs
```diff
@@ -246,33 +188,45 @@ where
...
+fn put_messages<DB, T: Cbor>(db: &DB, msgs: &[T]) -> Result<(), Error>
+where
+    DB: BlockStore,
+{
+    for m in msgs {
+        let key = m.cid()?.key();
+        let value = &m.marshal_cbor()?;
+        if db.exists(&key)? {
+            db.write(&key, value)?
+        }
```
austinabell, line 227: Actually, this could all just probably use the blockstore interface instead of using `write`, I believe this was done before the blockstore was setup @dutterbutter can you confirm? I think some of this can probably be cleaned up a bit later with available interfaces now

dutterbutter: Yes, that is correct!

austinabell:
```suggestion
        db.put(m, Blake2b256)
            .map_err(|e| Error::Other(e.to_string()))?;
```
and you'll just have to import Blake2b256 from cid::multihash

Issue #185, opened 2020-01-21
Tuple serialization macro


austinabell:
WIP branch: https://github.com/ChainSafe/ferret/compare/austin/serialize/tuplemacro

Default serialization for structs is as a Map, but it would be more preferred to be able to just derive packing the struct into a tuple to have it encoded as an a CBOR array in our case. The manual implementation looks something like this (UnsignedMessage implementation from #174):
```
impl ser::Serialize for UnsignedMessage {
    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
    where
        S: ser::Serializer,
    {
        (
            self.to.clone(),
            self.from.clone(),
            self.sequence,
            self.value.clone(),
            self.gas_price.clone(),
            self.gas_limit.clone(),
            self.method_num,
            self.params.clone(),
        )
            .serialize(s)
    }
}

impl<'de> de::Deserialize<'de> for UnsignedMessage {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: de::Deserializer<'de>,
    {
        let (to, from, sequence, value, gas_price, gas_limit, method_num, params) =
            Deserialize::deserialize(deserializer)?;
        Ok(Self {
            to,
            from,
            sequence,
            value,
            gas_price,
            gas_limit,
            method_num,
            params,
        })
    }
}
```
where it would be more beneficial to just annotate the structure itself to more easily configure the format and allow it to be applied to every structure that uses this packed format in Filecoin:

```
#[derive(TupleSerialize)]
pub struct UnsignedMessage {
    from: Address,
    to: Address,
    sequence: u64,
    value: TokenAmount,
    method_num: MethodNum,
    params: Serialized,
    gas_price: BigUint,
    gas_limit: BigUint,
}
```
and probably the option to add attributes on the parameters to override the order

austinabell:
Actually, would be more beneficial to implement the macro to serialize each fields (similar to how the serde tuple derive works) instead of copying and cloning into a tuple to be serialized. 

Doing this would increase the efficiency of our serialization and deserialization instead of just being syntactic sugar to reduce code size as insinuated in this issue.

PR #433, opened 2020-05-20
Remove a bajillion manual serde implementations

timvermeulen:
This PR adds [`serde_tuple`](https://docs.rs/serde_tuple) and [`serde_repr`](https://docs.rs/serde_repr) dependencies in order to remove by far most of our manual `Serialize`/`Deserialize` implementations.

`BlockHeader` is the main exception to this because for whatever reason `serde_tuple` and `derive_builder` don't seem to work well together.

Other minor related changes:
- Rename `types/src/sector/serde.rs` to `serde_tests.rs` because after removing the `Serialize`/`Deserialize` implementations, the tests were all that was left
- Add `bitvec_serde::{serialize, deserialize}` method in order to be used as `#[serde(with = "bitvec_serde")]` elsewhere
- Turn the `Randomness` type alias into a proper wrapper struct that correctly handles serialization
- ~Remove the `Byte32De` and `BytesDe` types as they're no longer needed~
- Add `extern crate serde;` in a couple places as workaround for a compiler bug


vm/actor/src/builtin/cron/state.rs
```diff
@@ -2,62 +2,21 @@
...
 use address::Address;
-use encoding::Cbor;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use encoding::{tuple::*, Cbor};
+use serde::{Deserialize, Serialize};
 use vm::MethodNum;
 
 /// Cron actor state which holds entries to call during epoch tick
-#[derive(Default)]
+#[derive(Default, Serialize, Deserialize)]
```
austinabell, line 10: same here, not correct impl (use `#[serde(transparent)]` and you get what you want

vm/src/randomness.rs
```diff
@@ -1,5 +1,57 @@
...
 /// String of random bytes
-pub type Randomness = [u8; 32];
+#[derive(PartialEq, Eq, Default, Copy, Clone)]
+pub struct Randomness(pub [u8; 32]);
+
+impl Debug for Randomness {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
```
austinabell, line 16: if you use this wrapper, we will just have to implement every trait we want to use for it, which gets clunky, probably best to just use a serialize tag to do this logic (as mentioned with the Byte32 removal) instead of unnecessarily having to use this type

austinabell:
So changed my mind on this (especially with cargo expand issue being resolved). If someone could also make a thorough pass to make sure of the field ordering and no bytes are serialized without tagging them to be serialized as byte string, that would be great!

Definitely with this now we have to be very careful of looking at field ordering to match go impl's but in the long run it should be easier. I also didn't know about the serde_repr derives, definitely much cleaner to use this and the tuple serialize macro (even if a bit more unreadable expanded code, that doesn't matter much)

Also closes #185 (very outdated and was created at very start of serde usage)

utils/bigint/src/bigint_ser.rs
```diff
@@ -2,18 +2,13 @@
...
 
 /// Wrapper for serializing big ints to match filecoin spec. Serializes as bytes.
 #[derive(Serialize)]
 #[serde(transparent)]
 pub struct BigIntSer<'a>(#[serde(with = "self")] pub &'a BigInt);
 
-/// Wrapper for deserializing as BigInt from bytes.
-#[derive(Deserialize, Serialize)]
-#[serde(transparent)]
-pub struct BigIntDe(#[serde(with = "self")] pub BigInt);
```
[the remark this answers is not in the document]
timvermeulen: The idea was that all these types will probably go eventually, but I'll keep them for now and then we'll decide later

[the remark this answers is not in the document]
austinabell: totally, that's cool. But I will mention having this type does come in handy for certain cases when you don't have a struct to annotate on a serde derive. Can be very useful for debugging certain things and in tests (not this one specifically yet tho, but could in future). 

I just don't think there is a disadvantage to having available, especially since we are publishing these crates so having this asymetric API is a bit weird.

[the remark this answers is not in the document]
timvermeulen: Ah, didn't think about that!

vm/actor/src/builtin/cron/mod.rs
```diff
@@ -22,31 +22,12 @@ pub enum Method {
 
 /// Constructor parameters for Cron actor, contains entries
 /// of actors and methods to call on each epoch
-#[derive(Default, Clone, PartialEq, Debug)]
+#[derive(Default, Clone, PartialEq, Debug, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
timvermeulen: Thanks, forgot to add `#[serde(transparent)]`

blockchain/blocks/src/block.rs
```diff
@@ -3,9 +3,8 @@
 
 use super::BlockHeader;
 use cid::Cid;
-use encoding::{de::Deserializer, ser::Serializer};
+use encoding::tuple::*;
```
ec2, line 6: Can you import the specific stuff instead of everything? Just to keep consistancy

timvermeulen: Yeah I don't mind either way, the reason I did it like this is that `serde_tuple` for whatever reason requires `serde_tuple` itself to be in scope as well

austinabell: can do `::{self, ..}` in that case, I don't mind the wildcard myself, just mentioning

encoding/src/bytes.rs
```diff
@@ -10,7 +10,7 @@ use serde_bytes::ByteBuf;
 pub struct BytesSer<'a>(#[serde(with = "serde_bytes")] pub &'a [u8]);
 
 /// Wrapper for deserializing dynamic sized Bytes.
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize)]
```
austinabell, line 13: optional: to include Serialize on these deserialize types

It's nice to have it symmetric if you need to use one type for serializing and deserializing (no actual examples in code rn but can be useful to move values instead of serializing a reference to them).

blockchain/blocks/src/block.rs
ec2: Oh, I see. Yeah, Wildcarding should be ok in this case then.

PR #439, opened 2020-05-22
Update libp2p to 0.19

austinabell:
**Summary of changes**
- Because of changes with graphsync, bumping versions before that
  - Also bumped other outdated versions (except for multihash, because there is no reason to and the API breaking changes were a disadvantage)

Closes 


node/forest_libp2p/src/rpc/behaviour.rs
```diff
@@ -25,8 +26,11 @@ impl RPC {
 
     /// Send an RPCEvent to a peer specified by peer_id.
     pub fn send_rpc(&mut self, peer_id: PeerId, event: RPCEvent) {
-        self.events
-            .push(NetworkBehaviourAction::SendEvent { peer_id, event });
+        self.events.push(NetworkBehaviourAction::NotifyHandler {
+            peer_id,
+            event,
+            handler: NotifyHandler::Any,
```
ec2, line 32: I got a little confused as to what this handler is when I attempted to upgrade last time... I was tempted to do th same: `Any`. Glad you figured it out.

PR #441, opened 2020-05-22
Add CircleCI

### LINKED ISSUE #440: Investigate/switch CI
**Issue summary**

With countless issues with Github actions and now builds running out of storage space, either a workaround needs to be added so this doesn't fail or switch over to another CI. 

Discussion indicated CircleCI would be a good switch, but open to discussion.

GregTheGreek:
**Summary of changes**
- Adds circle ci
- Uses a circleci machine instead of docker as intended originally.

Closes #440 

I suggest rolling this out slowly, not making circle builds mandatory *yet*, let a few prs go through to iron out any potential bugs before migrating.


GregTheGreek:
@priom I agree

One thing I've done is actually use a cache on the cargo builds so that we don't need to re-build everything

.circleci/config.yml
```diff
@@ -0,0 +1,146 @@
+version: 2.1
+
+executors:
+  test-executor:
+    machine:
+      image: ubuntu-1604:202004-01
```
austinabell, line 6: Are you sure ubuntu 16 is required?

.circleci/config.yml
```diff
@@ -0,0 +1,146 @@
...
+    description: Save cargo package cache for subsequent jobs
+    steps:
+      - save_cache:
+          key: cargo-package-cache-{{ .Environment.CIRCLE_SHA1 }}
+          # paths are relative to /home/circleci/project/
+          paths:
+            - ../.cargo/git
+            - ../.cargo/registry
+            - ../.cargo/.package-cache
+            # - ../../../usr/local/include
```
austinabell, line 77:
```suggestion
```
? Or is there a convenience need to keep

GregTheGreek: Proroc is stored in there

.circleci/config.yml
ec2: Nope. Its just the lastest VM image CircleCI has

.circleci/config.yml
ec2: im gunna remove it. Dont think its wise to store all the headers lol

austinabell: the headers are in the submodule? What's wrong with caching protoc?

PR #449, opened 2020-05-26
implement msg pool

### LINKED ISSUE #121: Implement Message Pool
The `Message Pool` is a subsystem in the Filecoin blockchain system. The message pool acts as the interface between Filecoin nodes and a peer-to-peer network used for off-chain message transmission. It is used by nodes to maintain a set of messages to transmit to the Filecoin VM (for “on-chain” execution).

**Spec reference:**
- [Message Pool](https://filecoin-project.github.io/specs/#systems__filecoin_blockchain__message_pool) 

**Existing implementations for reference:**
- [Lotus Message Pool](https://github.com/filecoin-project/lotus/blob/master/chain/messagepool/messagepool.go)
- [Go-Filecoin Message Pool](https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/message/pool.go)

**Needs access to the following subsystems:**
- `StateTree`
- `Messages` mined into `blocks`
- `MessagePubSub`
- `SignedMessages`

**Although the spec is very incomplete you can expect the following data structures/methods:**
- [ ] `struct MessagePool` and constructor
- [ ] `add_new_message`
- [ ] `find_messages` //returns set of messages in mempool
- [ ] `profitable_msgs` // returns messages most profitable to mine for the miner
- [ ] `MessageQuery struct` and constructor

From Lotus implementation:

**Pool struct**
```
type MessagePool struct {
	lk sync.Mutex

	closer  chan struct{}
	repubTk *time.Ticker

	localAddrs map[address.Address]struct{}

	pending map[address.Address]*msgSet

	curTsLk sync.Mutex // DO NOT LOCK INSIDE lk
	curTs   *types.TipSet

	api Provider

	minGasPrice types.BigInt

	maxTxPoolSize int

	blsSigCache *lru.TwoQueueCache

	changes *lps.PubSub

	localMsgs datastore.Datastore
}
```

**add_new_message:**

```
type msgSet struct {
	msgs      map[uint64]*types.SignedMessage
	nextNonce uint64
}

func newMsgSet() *msgSet {
	return &msgSet{
		msgs: make(map[uint64]*types.SignedMessage),
	}
}

func (ms *msgSet) add(m *types.SignedMessage) error {
	if len(ms.msgs) == 0 || m.Message.Nonce >= ms.nextNonce {
		ms.nextNonce = m.Message.Nonce + 1
	}
	if _, has := ms.msgs[m.Message.Nonce]; has {
		if m.Cid() != ms.msgs[m.Message.Nonce].Cid() {
			log.Error("Add with duplicate nonce")
			return xerrors.Errorf("message to %s with nonce %d already in mpool")
		}
	}
	ms.msgs[m.Message.Nonce] = m

	return nil
}
```

flodesi:
**Summary of changes**
- Implementing message pool closely following Lotus' implementation
- Refactoring already existing code to accommodate message pool

Closes #121


blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,865 @@
...
+        self.local_addrs.push(*m.from());
+        self.local_msgs.insert(msgb, m.clone());
+        Ok(())
+    }
+
+    /// Push a signed message to the MessagePool
+    pub fn push(&mut self, msg: &SignedMessage) -> Result<Cid, Error> {
+        let msg_serial = msg
+            .marshal_cbor()
+            .map_err(|err| Error::Other(err.to_string()))?;
```
dutterbutter, line 238: Reading Lotus, it seems they return an Undefined CID here and use it as such. This might need to return an Option to check that CID. Will defer to @austinabell though, I didn't realize `cid.Undef` was acceptable.

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,865 @@
...
+        if self.msgs.contains_key(&m.sequence()) {
+            let exms = self.msgs.get(&m.sequence()).unwrap();
+            if m.cid().map_err(|err| Error::Other(err.to_string()))?
+                != exms.cid().map_err(|err| Error::Other(err.to_string()))?
+            {
+                let gas_price = exms.message().gas_price();
+                let replace_by_fee_ratio: f32 = 1.25;
+                let rbf_num = BigUint::from(((replace_by_fee_ratio - 1_f32) * 256_f32) as u64);
+                let rbf_denom = BigUint::from(256 as u64);
+                let min_price = gas_price.clone() + (gas_price / &rbf_num) + rbf_denom;
```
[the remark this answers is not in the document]
flodesi: Oh yeah, completely didn't see that there will change.

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,865 @@
...
+    msgs: HashMap<u64, SignedMessage>,
+    next_nonce: u64,
+}
+
+impl MsgSet {
+    /// Generate a new MsgSet with an empty hashmap and a default next_nonce of 0
+    pub fn new() -> Self {
+        MsgSet {
+            msgs: HashMap::new(),
+            next_nonce: 0,
```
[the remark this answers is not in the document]
flodesi: Ah true, didn't realize that. Will update for clarity's sake.

blockchain/message_pool/src/msgpool.rs
austinabell: No don't think they do, should be fine as is. Can you show where they use the undef cid? They don't accept or allow Cids to be used in anything within the protocol, they just use for returns like this as it's empty data to return with the error (go sucks)

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+        self.add_local(msg, msg_serial)?;
+        msg.cid().map_err(|err| Error::Other(err.to_string()))
+    }
+
+    /// This is a helper to push that will help to make sure that the message fits the parameters
+    /// to be pushed to the MessagePool
+    pub fn add(&mut self, msg: &SignedMessage) -> Result<(), Error> {
+        let size = msg
+            .marshal_cbor()
+            .map_err(|err| Error::Other(err.to_string()))?
```
austinabell, line 250: again with these error conversions, the error type obviously never matters, so seems weird it's being done explicitly

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+            .len();
+        if size > 32 * 1024 {
+            return Err(Error::MessageTooBig);
+        }
+        if msg
+            .value()
+            .gt(&ToBigUint::to_biguint(&2_000_000_000).unwrap())
+        {
```
austinabell, line 258:
```suggestion
        if msg.value() > &BigUint::from(2_000_000_000u64) {
```
something like this?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+
+    /// Add a SignedMessage without doing any of the checks
+    pub fn add_skip_checks(&mut self, m: SignedMessage) -> Result<(), Error> {
+        self.add_helper(m)
+    }
+
+    /// Return the string representation of the message signature
+    pub(crate) fn sig_cache_key(&mut self, msg: &SignedMessage) -> Result<String, Error> {
+        match msg.signature().signature_type() {
+            SignatureType::Secp256k1 => Ok(msg.cid().unwrap().to_string()),
```
austinabell, line 276: remove the unwrap

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+        self.add_tipset(tmp, &self.cur_tipset.clone())
+    }
+
+    /// Add a SignedMessage without doing any of the checks
+    pub fn add_skip_checks(&mut self, m: SignedMessage) -> Result<(), Error> {
+        self.add_helper(m)
+    }
+
+    /// Return the string representation of the message signature
+    pub(crate) fn sig_cache_key(&mut self, msg: &SignedMessage) -> Result<String, Error> {
```
austinabell, line 274: This shouldn't be returning a string, the only reason go impl uses string is because it's an immutable version of a byte slice

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+                    .into_iter()
+                    .map(char::from)
+                    .collect::<String>();
+                beginning += &msg
+                    .signature()
+                    .bytes()
+                    .iter()
+                    .map(|b| char::from(*b))
+                    .collect::<String>();
+                Ok(beginning)
```
austinabell, line 295: once previous changes, you should just be appending the signature bytes to the cid bytes

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+                let verif = msg.signature().verify(umsg.as_slice(), msg.from());
+                match verif {
+                    Ok(()) => {
+                        self.sig_val_cache.put(sck, ());
+                        Ok(())
+                    }
+                    Err(value) => Err(Error::Other(value)),
+                }
+            }
+        }
```
austinabell, line 319:
```suggestion
        let sck = self.sig_cache_key(msg)?;
        let is_verif = self.sig_val_cache.get(&sck);
        if let Some(()) = self.sig_val_cache.get(&sck) {
            return Ok(());
        }
        let umsg = msg.message().marshal_cbor()?;
        let verif = msg.signature().verify(umsg.as_slice(), msg.from())?;
        self.sig_val_cache.put(sck, ());
        Ok(())
```
Easier to just do this instead of commenting on individual things. I didn't add the error mappings because I'm hoping the conversion gets added

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+            0 => Ok(BigInt::from(min_gas_price + 2)),
+            1 => Ok(BigInt::from(min_gas_price + 1)),
+            _ => Ok(BigInt::from(min_gas_price)),
+        }
+    }
+
+    /// Load local messages into pending. As of  right now messages are not deleted from self's
+    /// local_message field, possibly implement this in the future?
+    pub fn load_local(&mut self) -> Result<(), Error> {
+        for (key, value) in self.local_msgs.clone() {
```
austinabell, line 511: shouldn't need to clone if you are only using references, just turn into an iterator

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+            match self.add(&value) {
+                Ok(()) => (),
+                Err(err) => {
+                    if err == Error::SequenceTooLow {
+                        self.local_msgs.remove(&key);
+                    }
+                }
```
austinabell, line 518: maybe worth to at least log this error?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+            }
+        }
+        Ok(())
+    }
+
+    /// This is a helper method for head_change. This method will remove a sequence for a from address
+    /// from rmsgs
+    fn rm(
```
austinabell, line 526: can this at least be more descriptive? Did it not work out for this to be a closure?

vm/message/src/signed_message.rs
```diff
@@ -25,6 +25,9 @@ impl SignedMessage {
         Ok(SignedMessage { message, signature })
     }
 
+    pub fn new_from_fields(message: UnsignedMessage, signature: Signature) -> Self {
```
austinabell, line 28: idk optional but maybe `new_from_parts` is more in line with naming standards, from fields is a bit confusing in this context

blockchain/message_pool/src/msgpool.rs
flodesi: the clone was used because in the case where calling the add function with value fails, local_msgs needs to remove that kv pair.

austinabell: yeah, but you're taking the reference of the key? Why would you need the owned value? Even if you did, should only clone when and what you need to, and not for every key and value

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+        let mut msg: Vec<SignedMessage> = Vec::new();
+        for bh in h.blocks().iter() {
+            let (mut bh_umsg_tmp, mut bh_msg_tmp) = self.messages_for_block(bh)?;
+            let bh_umsg = bh_umsg_tmp.as_mut();
+            let bh_msg = bh_msg_tmp.as_mut();
+            umsg.append(bh_umsg);
+            msg.append(bh_msg);
+        }
+        for msg in &msg {
+            umsg.push(msg.message().clone());
```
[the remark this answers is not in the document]
flodesi: The clone is required because message() returns a reference to the unsigned message and the message field is private so there was no direct way to get the unsigned message without cloning.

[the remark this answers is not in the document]
austinabell: ah right, well optionally can add a function like: 
```
    pub fn into_message(self) -> UnsignedMessage {
        self.message
    }
```
to avoid this

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,866 @@
...
+}
+
+impl<DB> Provider for MpoolProvider<DB>
+where
+    DB: BlockStore,
+{
+    /// TODO find a way to get this function to periodically check for head changes in cs and update
+    /// msgpool accordingly or find a way to have an async function always running in the background
+    /// checking for newly published tipset and add it to msgpool
+    fn subscribe_head_changes(&mut self) -> Tipset {
```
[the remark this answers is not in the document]
ec2: I think this should be an async function and should probably spawn some task inside to periodically pull the chainstore subscribe, right?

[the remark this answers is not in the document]
flodesi: Yeah, I'll look into doing that after I fix the other issues that were brought up

blockchain/message_pool/src/msgpool.rs
flodesi: Initially tried using a closer, but it was getting a little bit wild with the borrowing, so I decided to just use a private method instead.

blockchain/message_pool/src/msgpool.rs
flodesi: But by removing the clone and just using references, issues arose where multiple borrows due to the deletion of an element of the hashmap that was being iterated over.

blockchain/message_pool/src/msgpool.rs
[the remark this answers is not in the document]
ec2: Ok cool, unresolving for now. We can have a chat tomorrow, if you wanna talk about this tomorrow.

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,865 @@
...
+            self.next_nonce = m.sequence() + 1;
+        }
+        if self.msgs.contains_key(&m.sequence()) {
+            let exms = self.msgs.get(&m.sequence()).unwrap();
+            if m.cid().map_err(|err| Error::Other(err.to_string()))?
+                != exms.cid().map_err(|err| Error::Other(err.to_string()))?
+            {
+                let gas_price = exms.message().gas_price();
+                let replace_by_fee_ratio: f32 = 1.25;
+                let rbf_num = BigUint::from(((replace_by_fee_ratio - 1_f32) * 256_f32) as u64);
```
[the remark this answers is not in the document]
flodesi: I'll make the replace_by_fee_ratio a constant, but the others I can't because you can't call BigUint::from() when creating a constant.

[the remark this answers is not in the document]
austinabell: you can make everything within the BigUint::from() a constant

blockchain/chain/src/store/chain_store.rs
```diff
@@ -199,6 +200,23 @@ where
         }
         Ok(())
     }
+
+    /// Return the messages in the chainstore for a given tipset
+    pub fn messages_for_tipset(&self, h: &Tipset) -> Result<Vec<UnsignedMessage>, Error> {
```
ec2, line 205: no need for `self`. Have it take a DB instead

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,951 @@
...
+    {
+        // LruCache sizes have been taken from the lotus implementation
+        let tipset = Mutex::new(
+            api.get_heaviest_tipset()
+                .ok_or_else(|| Error::Other("No ts in api to set as cur_tipset".to_owned()))?,
+        );
+        let bls_sig_cache = LruCache::new(40000);
+        let sig_val_cache = LruCache::new(32000);
+        let api_mutex = Arc::new(Mutex::new(api));
+        let mp = Arc::new(Mutex::new(MessagePool {
```
ec2, line 190: Is it necessary to have the whole MP protected by a mutex? Looks like it should be sufficient that only some fields need to be. Also, If you have mutual exclusion for the whole struct, you automatically have mutual exclusion for its individual fields (i think).

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,951 @@
...
+            api: api_mutex,
+            min_gas_price: Default::default(),
+            max_tx_pool_size: 5000,
+            network_name,
+            bls_sig_cache,
+            sig_val_cache,
+            local_msgs: HashMap::new(),
+        }));
+
+        let mut mp_lock = mp.lock().map_err(|_| Error::MutexPoisonError)?;
```
ec2, line 203: You can just unwrap here I think if youre locking a mutex... Not sure how you would recover from a Poisoned Mutex

flodesi: The issue that I was having with simply calling unwrap was if there was a MutexPoisonError, It wasn't always due to a poisioned lock. For example, in the last test in messagepool, if you remove the sleep call, there will be a MutexPoisonError due to the subscribe new heads async block using the messagepool. Adding that map_err accounted for that so there wouldn't be any crashes due to mutexes that weren't really poisoned.

blockchain/message_pool/src/msgpool.rs
flodesi: Ok, I'll try removing the mutex around messagepool.

blockchain/message_pool/src/msgpool.rs
austinabell: would be nice to use the async locks (from async-std) and you wouldn't have to worry about this at all, but I do see the increased dev in making functions async

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,951 @@
...
+        drop(updater);
+
+        let ts = tipset.clone();
+        task::block_on(async move {
+            // updater.api.lock().unwrap().set_block_messages(&header, vec![message]);
+            api.set_heaviest_tipset(Arc::new(ts)).await;
+        });
+
+        // sleep allows for async block to update mpool's cur_tipset
+        sleep(Duration::new(2, 0));
```
ec2, line 945: since youre calling block_on, why do you need to sleep? set_heaviest_tipset should be guaranteed to run first no?

blockchain/message_pool/src/msgpool.rs
ec2: That's weird... I don't see why the mutex would error if it wasnt poisoned...

blockchain/message_pool/src/msgpool.rs
flodesi: Actually no, so what happens is that set_heaviest_tipset will send a subscriber notification that the head has changed, so then the async block in the MP initializer will get notified so it will run the head_change method.

flodesi: So technically set_heaviest_tipset will run first, but there isn't a guarantee that the head change will run first before the check below

blockchain/message_pool/src/msgpool.rs
flodesi: I know, I was trying to understand why that was going on, but after adding the sleep call in the test, it went away.

ec2: I would also consider using the mutexes provided by async_std instead of the std lib one. The issue could be coming from the fact we're mixing them

AshantiMutinta: The mutexes in the std might also block the future which is problematic

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1065 @@
...
+                if err == Error::SequenceTooLow {
+                    warn!("error adding message: {:?}", err);
+                    self.local_msgs.remove(&key);
+                }
+            });
+        }
+        Ok(())
+    }
+}
+
```
dutterbutter, line 481: Nit: add comment please

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+        }
+
+        Ok(base_sequence)
+    }
+
+    /// Get the state balance for the actor that corresponds to the supplied address and tipset,
+    /// if this actor does not exist, return an error
+    async fn get_state_balance(&mut self, addr: &Address, ts: &Tipset) -> Result<BigInt, Error> {
+        let api = self.api.read().await;
+        let actor = api.state_get_actor(&addr, &ts)?;
```
austinabell, line 374: can just be oneline, but this one doesn't matter

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    }
+
+    /// Return a tuple that contains a vector of all signed messages and the current tipset for
+    /// self.
+    pub async fn pending(&self) -> Result<(Vec<SignedMessage>, Tipset), Error> {
+        let mut out: Vec<SignedMessage> = Vec::new();
+
+        let pending = self.pending.read().await;
+        let pending_hm = pending.clone();
+        drop(pending);
```
austinabell, line 390: same, but it might make sense to just not drop the locks so they get dropped at end of function (is there an invariant that breaks this?)

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+                self.pending_for(&addr)
+                    .await
+                    .ok_or_else(|| Error::InvalidFromAddr)?
+                    .as_mut(),
+            )
+        }
+
+        let cur_t = self.cur_tipset.read().await;
+        let cur_ts = cur_t.clone();
+        drop(cur_t);
```
austinabell, line 403: same as above

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+                    msg_vec.push(item);
+                }
+
+                msg_vec.sort_by_key(|value| value.message().sequence());
+
+                Some(msg_vec)
+            }
+            None => None,
+        }
+    }
```
austinabell, line 432:
```suggestion
    async fn pending_for(&self, a: &Address) -> Option<Vec<SignedMessage>> {
        let pending = self.pending.read().await;
        let mset = pending.get(a)?;

        if mset.msgs.is_empty() {
            return None;
        }

        let mut msg_vec = Vec::new();

        for (_, item) in mset.msgs.iter() {
            msg_vec.push(item.clone());
        }

        msg_vec.sort_by_key(|value| value.message().sequence());

        Some(msg_vec)
    }
```

Few small changes so I'll just do as one. There was the cosmetic change to avoid match but also the functional one which is to avoid cloning the hashset instead of cloning the elements being pushed

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    pub async fn messages_for_blocks(
+        &mut self,
+        blks: &[BlockHeader],
+    ) -> Result<Vec<SignedMessage>, Error> {
+        let mut msg_vec: Vec<SignedMessage> = Vec::new();
+
+        for block in blks {
+            let api = self.api.read().await;
+            let (umsg, mut smsgs) = api.messages_for_block(&block)?;
+            drop(api);
```
austinabell, line 444:
```suggestion
            let (umsg, mut smsgs) = self.api.read().await.messages_for_block(&block)?;
```

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    }
+    Ok(())
+}
+
+/// Attempt to get a signed message that corresponds to an unsigned message in bls_sig_cache
+async fn recover_sig(
+    bls_sig_cache: Arc<RwLock<LruCache<Cid, Signature>>>,
+    msg: UnsignedMessage,
+) -> Result<SignedMessage, Error> {
+    let mut bls_sig_cache = bls_sig_cache.write().await;
```
austinabell, line 517: you don't need write access to do this? can also change the bls_sig_cache type to just be a ref to the lru cache because this function does not need to be aware of the lock

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+        let mut msg_vec: Vec<SignedMessage> = Vec::new();
+
+        for block in blks {
+            let api = self.api.read().await;
+            let (umsg, mut smsgs) = api.messages_for_block(&block)?;
+            drop(api);
+
+            msg_vec.append(smsgs.as_mut());
+            for msg in umsg {
+                let smsg = recover_sig(self.bls_sig_cache.clone(), msg).await?;
```
austinabell, line 448: if you change other point, you won't need to clone the Arc here, you would just get a read into the rwlock and pass the ref in

blockchain/message_pool/src/msgpool.rs
austinabell: yeah cloning the whole hashmap is really expensive, just create a vector of keys to remove, and after this loop remove all those keys from the hashmap.

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    api.put_message(&msg)?;
+    let msett = pending.get_mut(msg.message().from());
+    match msett {
+        Some(mset) => mset.add(&msg)?,
+        None => {
+            let mut mset = MsgSet::new();
+            mset.add(&msg)?;
+            pending.insert(*msg.message().from(), mset);
+        }
+    }
```
austinabell, line 558:
```suggestion
    if msg.signature().signature_type() == SignatureType::BLS {
        bls_sig_cache
            .write()
            .await
            .put(msg.cid()?, msg.signature().clone());
    }

    if msg.message().gas_limit() > 100_000_000 {
        return Err(Error::Other(
            "given message has too high of a gas limit".to_string(),
        ));
    }

    api.read().await.put_message(&msg)?;
    match pending.write().await.get_mut(msg.message().from()) {
        Some(mset) => mset.add(&msg)?,
        None => {
            let mut mset = MsgSet::new();
            mset.add(&msg)?;
            pending.write().await.insert(*msg.message().from(), mset);
        }
    }
```
Is there a reason you have the locks setup as is and not something like this? should really think carefully about how you are using all of these locks, a lot of changes in this PR seem like it's asking for a deadlock

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    apply: Vec<Tipset>,
+) -> Result<(), Error>
+where
+    T: Provider,
+{
+    let mut rmsgs: HashMap<Address, HashMap<u64, SignedMessage>> = HashMap::new();
+    for ts in revert {
+        let api_locked = api.write().await;
+        let pts = api_locked.load_tipset(ts.parents())?;
+        drop(api_locked);
```
austinabell, line 579: same point as before

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+
+        let mut msgs: Vec<SignedMessage> = Vec::new();
+        for block in ts.blocks() {
+            let api_locked = api.read().await;
+            let (umsg, mut smsgs) = api_locked.messages_for_block(&block)?;
+            drop(api_locked);
```
austinabell, line 585: again same

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+            msgs.append(smsgs.as_mut());
+            for msg in umsg {
+                let bls_sig_cache = bls_sig_cache.clone();
+                let smsg = recover_sig(bls_sig_cache, msg).await?;
```
austinabell, line 589: the clone should be unnecessary if you change comment made earlier

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+                msgs.push(smsg)
+            }
+        }
+        let parent = pts.clone();
+        let mut cur_ts_locked = cur_tipset.write().await;
+        *cur_ts_locked = parent;
```
austinabell, line 595:
```suggestion
        let mut cur_ts_locked = cur_tipset.write().await;
        *cur_ts_locked = pts;
```
that is a very expensive clone btw

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+        drop(cur_ts_locked);
```
austinabell, line 596: also if you could make it oneline instead of dropping explicitly again

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+        for msg in msgs {
+            add(msg, rmsgs.borrow_mut());
+        }
+    }
+
+    for ts in apply {
+        for b in ts.blocks() {
+            let api_locked = api.read().await;
+            let (msgs, smsgs) = api_locked.messages_for_block(b)?;
+            drop(api_locked);
```
austinabell, line 607: same

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+                    pending.clone(),
+                    msg.sequence(),
+                    rmsgs.borrow_mut(),
+                )
+                .await;
+            }
+        }
+        let mut cur_ts_locked = cur_tipset.write().await;
+        *cur_ts_locked = ts;
+        drop(cur_ts_locked);
```
austinabell, line 630: same

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    }
+    for (_, hm) in rmsgs {
+        for (_, msg) in hm {
+            let api_locked = api.clone();
+            let pending = pending.clone();
+            let bls_sig_cache = bls_sig_cache.clone();
+            add_helper(api_locked, bls_sig_cache, pending, msg.clone())
+                .await
+                .ok();
```
austinabell, line 639:
```suggestion
            if let Err(e) =
                add_helper(api.clone(), bls_sig_cache.clone(), pending.clone(), msg).await
            {
                error!("Failed to readd message from reorg to mpool: {}", e);
            }
```
rn you are ignoring that error, should log it (error macro is just log::error)

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+/// from the rmsgs hashmap. Also remove the from address and sequence from the mmessagepool.
+async fn rm(
+    from: &Address,
+    pending: Arc<RwLock<HashMap<Address, MsgSet>>>,
+    sequence: u64,
+    rmsgs: &mut HashMap<Address, HashMap<u64, SignedMessage>>,
+) {
+    let s = rmsgs.get_mut(from);
+    if s.is_none() {
+        remove(from, pending.clone(), sequence).await.ok();
```
austinabell, line 655: can you not just put `.ok` on these? you're just ignoring the error, when it would otherwise be handled

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+        return;
+    }
+    let temp = s.unwrap();
+    if temp.get_mut(&sequence).is_some() {
```
austinabell, line 659: instead of checking `is_some`, `is_none` and unwrapping, just do an `if let Some(..)`. Would be much more readable

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+        temp.remove(&sequence);
+        return;
+    }
+    remove(from, pending, sequence).await.ok();
```
austinabell, line 663: error not handled again

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+fn add(m: SignedMessage, rmsgs: &mut HashMap<Address, HashMap<u64, SignedMessage>>) {
+    let s = rmsgs.get_mut(m.from());
+    if s.is_none() {
+        let mut temp = HashMap::new();
+        temp.insert(m.sequence(), m.clone());
+        rmsgs.insert(*m.from(), temp);
+        return;
+    }
+    let temp = s.unwrap();
+    temp.insert(m.sequence(), m);
```
austinabell, line 676: prob best to just do if let Some here, instead of checking if none then unwrapping

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    remove(from, pending, sequence).await.ok();
+}
+
+/// This function is a helper method for head_change. This method will add a signed message to the given rmsgs HashMap
+fn add(m: SignedMessage, rmsgs: &mut HashMap<Address, HashMap<u64, SignedMessage>>) {
+    let s = rmsgs.get_mut(m.from());
+    if s.is_none() {
+        let mut temp = HashMap::new();
+        temp.insert(m.sequence(), m.clone());
+        rmsgs.insert(*m.from(), temp);
```
austinabell, line 672: if you swap the order of these, you don't need to clone the message

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+        if msg.value() > &BigUint::from(2_000_000_000u64) {
+            return Err(Error::MessageValueTooHigh);
+        }
+
+        self.verify_msg_sig(msg)?;
+        let tmp = msg.clone();
+
+        let tip_locked = self.cur_tipset.read().await;
+        let tip = tip_locked.clone();
+        drop(tip_locked);
```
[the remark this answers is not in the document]
flodesi: Sweet thanks! Didn't know that

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1066 @@
...
+    /// in the hashmap does not yet exist, create a new mset that will correspond to the from message
+    /// and push it to the pending phashmap
+    async fn add_helper(&mut self, msg: SignedMessage) -> Result<(), Error> {
+        add_helper(
+            self.api.clone(),
+            self.bls_sig_cache.clone(),
+            self.pending.clone(),
+            msg,
+        )
+        .await
```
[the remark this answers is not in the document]
flodesi: in head_change there is a call to this function, and since head_change is used in an async block that doesn't have access to mpool, just those specific parts of it, it made sense to just have self.head_change act as a wrapper method.

blockchain/message_pool/src/msgpool.rs
flodesi: I don't see any reason to keep a read lock until the end of this function

blockchain/message_pool/src/msgpool.rs
flodesi: write was required to call the get method. with read you can't get a mut RwLock, but that doesn't really matter if I switch to a ref to the lru cache

blockchain/message_pool/src/msgpool.rs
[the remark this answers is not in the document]
austinabell: okay so why does head_change not use self? seems like it probably should instead of passing in all the fields you need?

blockchain/message_pool/src/msgpool.rs
austinabell: Is there no race condition you can hit with that pending cache and current tipset though? Do they have no relation to eachother that this will never be an issue? Point I'm making is that the pending lock is released before you lock the current tipset mutex, so there may be a potential invariant that breaks the logic of the function because another task could modify in between this?

blockchain/message_pool/src/msgpool.rs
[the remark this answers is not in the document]
flodesi: whoops, slight mistake in my last comment that I corrected. head_change can't use self because the async block that calls it doesn't have access to self, if it did, then all of the mpool would have to be wrapped in an RwLock, which I wanted to avoid.

blockchain/message_pool/src/msgpool.rs
flodesi: ah true didn't see that, yeah makes sense to keep that read lock until the end of the function

blockchain/message_pool/src/msgpool.rs
flodesi: I cleaned up the locks a bit, the match statement can't be like you proposed because you get a deadlock because the write lock in 550 is still held if the function is in the none case and needs to run line 555

blockchain/message_pool/src/msgpool.rs
flodesi: after a little bit of train and error, realized that self.local_msgs does not need to be a hashmap, changed it over to a vec in latest commit

blockchain/chain/src/store/chain_store.rs
```diff
@@ -208,16 +208,10 @@ where
...
-        let bh_umsg = &mut bh_umsg_tmp;
-        let bh_msg = &mut bh_msg_tmp;
-        umsg.append(bh_umsg);
-        msgs.append(bh_msg);
-    }
-    for msg in msgs {
-        umsg.push(msg.into_message());
+        let (mut bh_umsg, bh_msg) = block_messages(db, bh)?;
+        umsg.append(&mut bh_umsg);
+        umsg.append(&mut bh_msg.into_iter().map(|msg| msg.into_message()).collect());
```
austinabell, line 214: The collect does an allocation still, can do something like:
```
        umsg.extend(bh_msg.into_iter().map(|msg| msg.into_message()));
```
to just extend with the elements

blockchain/message_pool/src/msgpool.rs
```diff
@@ -57,15 +55,18 @@ impl MsgSet {
                 let gas_price = exms.message().gas_price();
                 let rbf_num = BigUint::from(RBF_NUM);
                 let rbf_denom = BigUint::from(RBF_DENOM);
-                let min_price =
-                    gas_price.clone() + (gas_price / &rbf_num) + rbf_denom + BigUint::from(1_u64);
+                let min_price = gas_price.clone() + ((gas_price * &rbf_num) / rbf_denom) + 1u8;
                 if m.message().gas_price() <= &min_price {
                     // message with duplicate sequence is already in mpool
+                    warn!("try to add message with duplicate sequence");
                     return Err(Error::DuplicateSequence);
```
austinabell: This is the wrong error message, gas price being less has nothing to do with duplicate sequence

blockchain/message_pool/src/msgpool.rs
```diff
@@ -467,13 +445,14 @@ where
...
     pub async fn load_local(&mut self) -> Result<(), Error> {
-        for (key, value) in self.local_msgs.clone() {
-            self.add(&value).await.unwrap_or_else(|err| {
+        let msg_vec = self.local_msgs.clone();
+        for msg in msg_vec {
+            self.add(&msg).await.unwrap_or_else(|err| {
                 if err == Error::SequenceTooLow {
                     warn!("error adding message: {:?}", err);
-                    self.local_msgs.remove(&key);
+                    self.local_msgs.retain(|smsg| *smsg != msg);
```
austinabell, line 453: why did you switch from using a hash map for local_msgs? Removing from the vec and having to do these checks is not very efficient

blockchain/message_pool/src/msgpool.rs
```diff
@@ -467,13 +445,14 @@ where
     /// Load local messages into pending. As of  right now messages are not deleted from self's
     /// local_message field, possibly implement this in the future?
     pub async fn load_local(&mut self) -> Result<(), Error> {
-        for (key, value) in self.local_msgs.clone() {
-            self.add(&value).await.unwrap_or_else(|err| {
+        let msg_vec = self.local_msgs.clone();
```
austinabell, line 448: shouldn't have to clone

blockchain/message_pool/src/msgpool.rs
```diff
@@ -622,21 +595,18 @@ where
...
     for (_, hm) in rmsgs {
         for (_, msg) in hm {
-            let api_locked = api.clone();
-            let pending = pending.clone();
-            let bls_sig_cache = bls_sig_cache.clone();
-            add_helper(api_locked, bls_sig_cache, pending, msg.clone())
-                .await
-                .ok();
+            if let Err(e) =
+                add_helper(api.clone(), bls_sig_cache.clone(), pending.clone(), msg).await
```
austinabell, line 606: so actually you can just change the function signature of this function to take in references so you don't have to clone the `Arc`s, they don't cross any thread bounds

blockchain/message_pool/src/msgpool.rs
flodesi: I was thinking that using a vec and its inefficiencies of this was better than having to clone the hashmap

austinabell: you don't need to clone the hashmap, and yeah this ends up being not very efficient for it to be a vec instead

blockchain/message_pool/src/msgpool.rs
flodesi: clone has to be there, if it isn't the borrow checker complains when calling self.add()

austinabell: there are ways to get around this, such as replacing the loop with something like:
```
        while let Some(msg) = self.local_msgs.pop() {
```

But I noticed also they are persisting these messages, and wonder why they are all kept in memory anyway, any specific reason?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -503,19 +499,16 @@ async fn recover_sig(
 /// in the hashmap does not yet exist, create a new mset that will correspond to the from message
 /// and push it to the pending hashmap
 async fn add_helper<T>(
-    api: Arc<RwLock<T>>,
-    bls_sig_cache: Arc<RwLock<LruCache<Cid, Signature>>>,
-    pending: Arc<RwLock<HashMap<Address, MsgSet>>>,
+    api: RwLockReadGuard<'_, T>,
+    mut bls_sig_cache: RwLockWriteGuard<'_, LruCache<Cid, Signature>>,
+    mut pending: RwLockWriteGuard<'_, HashMap<Address, MsgSet>>,
```
austinabell, line 504: sorry if misleading, I just meant you could swap the `Arc<>` with a reference, not switch the locks to the locked versions

blockchain/message_pool/src/msgpool.rs
flodesi: I don't know either, it would make more sense to just remove the messages from pending when they get loaded.

blockchain/message_pool/src/msgpool.rs
flodesi: I don't see how you don't have to clone the hashmap, Reason being is self.add needs a reference to a signed message, and if you don't clone the hm, borrow checker get mad.

austinabell: Hashmaps have a function, `drain`, which would probably have solved your issue

flodesi: I was thinking about using drain, the only issue with it was that after all the kv pairs are drained from local_msgs, a temp variable would be needed to store the values that need to be reinserted into local_msgs. What I was thinking of doing was using a Box<Vec<SignedMessage>> for local_msgs, and doing something like the following:

```
let mut local = Vec::new();

while let Some(msg) = self.local_msgs.pop() {
    self.add(&msg).await.unwrap_or_else(|err| {
        if err == Error::SequenceTooLow {
            warn!("error adding message: {:?}", err);
        } else {
            local.push(msg);
        }
    })
}
*self.local_msgs = local;
```

blockchain/message_pool/src/msgpool.rs
```diff
@@ -157,17 +157,18 @@ where
...
     pub cur_tipset: Arc<RwLock<Tipset>>,
     api: Arc<RwLock<T>>,
     pub min_gas_price: BigInt,
     pub max_tx_pool_size: i64,
     pub network_name: String,
-    bls_sig_cache: Arc<RwLock<LruCache<Cid, Signature>>>, // mutex this
+    bls_sig_cache: Arc<RwLock<LruCache<Cid, Signature>>>,
     sig_val_cache: LruCache<Cid, ()>,
-    local_msgs: Vec<SignedMessage>,
+    local_msgs: Box<Vec<SignedMessage>>,
```
austinabell, line 171: Why put a vec into a box? Vec is already heap allocated

flodesi: yikes, totally forgot about that, will fix

blockchain/message_pool/src/msgpool.rs
austinabell: This is still not ideal, as you are holding all locks for this whole function when unnecessary, should just be `&RwLock<...>` and you handle the locks when necessary, as it was before you changed this, or do you have a good reason for this to remain as is?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1033 @@
...
+
+    if msg.message().gas_limit() > 100_000_000 {
+        return Err(Error::Other(
+            "given message has too high of a gas limit".to_string(),
+        ));
+    }
+
+    api.put_message(&msg)?;
+
+    // let mut pending = pending.write().await;
```
austinabell:
```suggestion
```

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1033 @@
...
+    Ok(())
+}
+
+/// This function will revert and/or apply tipsets to the message pool. This function should be
+/// called every time that there is a head change in the message pool
+pub async fn head_change<T>(
+    api: Arc<RwLock<T>>,
+    bls_sig_cache: Arc<RwLock<LruCache<Cid, Signature>>>,
+    pending: Arc<RwLock<HashMap<Address, MsgSet>>>,
+    cur_tipset: Arc<RwLock<Tipset>>,
```
austinabell, line 555: pretty sure these don't need to be cloned arcs also, and can just be references to the RwLock<..> to avoid the atomic count (minimal change)

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1033 @@
...
+    local_addrs: Vec<Address>,
+    pending: Arc<RwLock<HashMap<Address, MsgSet>>>,
+    pub cur_tipset: Arc<RwLock<Tipset>>,
+    api: Arc<RwLock<T>>,
+    pub min_gas_price: BigInt,
+    pub max_tx_pool_size: i64,
+    pub network_name: String,
+    bls_sig_cache: Arc<RwLock<LruCache<Cid, Signature>>>,
+    sig_val_cache: LruCache<Cid, ()>,
+    local_msgs: Vec<SignedMessage>,
```
austinabell, line 170: This should really be a hashmap, as it was before. There is no deduplication you are handling in here, and most interactions are less efficient because of how they are used. Can you give me a functional reason why this should be a Vector?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1033 @@
...
+                    warn!("error adding message: {:?}", err);
+                } else {
+                    msg_vec.push(msg);
+                }
+            } else {
+                msg_vec.push(msg);
+            }
+        }
+
+        self.local_msgs = msg_vec;
```
austinabell, line 460: You're removing values and reallocating a new vector without any reason (continuation of previous), if this remains the same, you should at very least set the capacity of the `msg_vec` so the vec doesn't keep getting reallocated as it hits capacity

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1033 @@
...
+                publisher: Publisher::new(1),
+            }
+        }
+
+        pub fn set_state_sequence(&mut self, addr: &Address, sequence: u64) {
+            self.state_sequence.insert(addr.clone(), sequence);
+        }
+
+        pub fn set_block_messages(&mut self, h: &BlockHeader, msgs: Vec<SignedMessage>) {
+            self.bmsgs.insert(h.cid().clone(), msgs.clone());
```
austinabell, line 691:
```suggestion
            self.bmsgs.insert(h.cid().clone(), msgs);
```

blockchain/message_pool/src/msgpool.rs
flodesi: will change back, version I was working off of on old computer reverted it for some reason

blockchain/message_pool/src/msgpool.rs
```diff
@@ -0,0 +1,1033 @@
...
+
+        let mut subscriber = mp.api.write().await.subscribe_head_changes();
+
+        let api = mp.api.clone();
+        let bls_sig_cache = mp.bls_sig_cache.clone();
+        let pending = mp.pending.clone();
+        let cur_tipset = mp.cur_tipset.clone();
+
+        task::spawn(async move {
+            loop {
```
austinabell, line 214: Just to note, this is not ideal to start up an infinite task every time `MessagePool::new` is called without a way to cancel it. Forces it to be used in a very specific way. Can revisit later I guess

blockchain/message_pool/src/msgpool.rs
```diff
@@ -214,10 +215,10 @@ where
             loop {
                 if let Some(ts) = subscriber.next().await {
                     head_change(
-                        api.clone(),
-                        bls_sig_cache.clone(),
-                        pending.clone(),
-                        cur_tipset.clone(),
+                        api.as_ref(),
```
austinabell, line 218: just so you know, you don't have to use `as_ref()` explicitly, can just put a `&` behind the variables if it's easier in future

PR #454, opened 2020-05-27
Ashanti/connect state transition

### LINKED ISSUE #405: Connecting state transition to chain system
**Task summary**
We have implemented the logic for making the state transition via `apply_blocks` however we still need to connect the VM subsystem with the blockchain subsystem. This will likely be in the `state_manager` component. 

- [Apply Blocks - StateManager](https://github.com/ChainSafe/forest/blob/master/blockchain/state_manager/src/lib.rs#L96)

This task may be dependent on the implementation of graph sync #379 as the state transition will be called during the syncing process.

**Specification reference**

- 

The below link shows the logic that is being used in lotus to update state from the blockchain subsystem, this should be used as a reference and not necessarily a 1:1 blueprint. 
- [`TipSetState`: Lotus connecting logic](https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/stmgr.go#L54)

AshantiMutinta:
**Summary of changes**
- Looks like the channel was being used for concurrency, didn't implement due to rust safety guarantees
-a bit of a translation of tipset_state in lotus
-missing a few functions that call tipset_state
-Changes statemanager in sync to an arc readlock due to the need of shared reference in asynchronous context

Closes #405 


blockchain/blocks/src/header.rs
```diff
@@ -294,14 +294,14 @@ impl BlockHeader {
         Ok(())
     }
     /// Check to ensure block signature is valid
-    pub fn check_block_signature(&self, addr: &Address) -> Result<(), Error> {
+    pub fn check_block_signature(&self, addr: Address) -> Result<(), Error> {
```
AshantiMutinta, line 297: Helped in an async context since dealing with references

blockchain/chain_sync/src/sync.rs
```diff
@@ -697,7 +725,7 @@ where
         }
 
         for b in fts.blocks() {
-            if let Err(e) = self.validate(b).await {
+            if let Err(e) = self.validate(b.clone()).await {
```
AshantiMutinta, line 728: that clone might be expensive and would need to be changed

blockchain/blocks/src/header.rs
austinabell: There is no need for it to be like this, I just checked

blockchain/chain_sync/src/sync.rs
```diff
@@ -597,10 +610,9 @@ where
     }
 
     /// Validates block semantically according to https://github.com/filecoin-project/specs/blob/6ab401c0b92efb6420c6e198ec387cf56dc86057/validation.md
-    async fn validate(&self, block: &Block) -> Result<(), Error> {
+    async fn validate(&self, block: Block) -> Result<(), Error> {
```
austinabell, line 613: Reason for the switch from a reference? Cloning a block is not cheap at all, and it works fine to just use a reference

blockchain/chain_sync/src/sync.rs
```diff
@@ -620,30 +632,44 @@ where
...
-                        .map_err(Error::Blockchain)
-                });
-                validations.push(block_sig_task)
+            Ok(_) => {
+  
+                validations.push(
+                    async {
+                        block
+                            .header()
+                            .check_block_signature(work_addr_result.unwrap().clone()) //work_addr_result lives longer than unwrapped_value in scope
```
austinabell, line 660: can you remove the comment? unclear what exactly the meaning is

(Here is where you can add the reference wrt the previous comment and it works fine)

blockchain/chain_sync/src/sync.rs
austinabell, line 728: indicating where clone is unnecessary

blockchain/state_manager/src/lib.rs
```diff
@@ -117,6 +131,98 @@ where
...
+        // generic constants are not implemented yet this is a lowcost method for now
+        let cid_pair = self.compute_tipset_state(&block_headers, &HashMap::new())?;
+        self.cache.insert(tipset.key().clone(), cid_pair.clone());
+        Ok(cid_pair)
+    }
+
+    pub fn compute_tipset_state<'a>(
+        &'a self,
+        blocks: &[BlockHeader],
+        forks_at_heights: &'a HashMap<ChainEpoch, ForkFunctions<'a, DB>>,
```
austinabell, line 169: Can you remove these for now since it's not actually used? Would be nice to come up with a better fork handling mechanism than this (or at least use a fnv hash map since ChainEpoch is just an integer for faster hash lookups at least)

Generally just want to avoid using rust hashmaps when possible because they aren't fast

blockchain/state_manager/src/lib.rs
```diff
@@ -117,6 +131,98 @@ where
         Ok((state_root, rect_root))
     }
 
+    pub fn tipset_state(&mut self, tipset: &Tipset) -> Result<(Cid, Cid), Box<dyn StdError>> {
+        trace!("tipSetState");
```
austinabell, line 135: maybe at least rename to our function casing? would be nice to trace the tipset Cid which state is being generated, this log is generally pretty useless

blockchain/state_manager/src/lib.rs
```diff
@@ -117,6 +131,98 @@ where
...
+        forks_at_heights: &'a HashMap<ChainEpoch, ForkFunctions<'a, DB>>,
+    ) -> Result<(Cid, Cid), Box<dyn StdError>> {
+        trace!("compute tipset state");
+        if blocks.len() > 2
+            && blocks
+                .iter()
+                .zip(blocks.iter().skip(0))
+                .any(|(a, b)| a.miner_address() == b.miner_address())
+        {
+            // Duplicate Minor found
```
austinabell, line 178: miner?

blockchain/state_manager/src/lib.rs
```diff
@@ -117,6 +131,98 @@ where
...
+        self.cache.insert(tipset.key().clone(), cid_pair.clone());
+        Ok(cid_pair)
+    }
+
+    pub fn compute_tipset_state<'a>(
+        &'a self,
+        blocks: &[BlockHeader],
+        forks_at_heights: &'a HashMap<ChainEpoch, ForkFunctions<'a, DB>>,
+    ) -> Result<(Cid, Cid), Box<dyn StdError>> {
+        trace!("compute tipset state");
```
austinabell, line 171: I don't really think this trace log is very useful, go impl uses it for a span to track time in function, this just logs out that message

blockchain/state_manager/src/lib.rs
```diff
@@ -117,6 +131,98 @@ where
...
+                tipset.parent_state().clone(),
+                message_receipts.message_receipts().clone(),
+            );
+            self.cache.insert(tipset.key().clone(), cid_pair.clone());
+            return Ok(cid_pair);
+        }
+
+        let block_headers = tipset.blocks().to_vec();
+        // generic constants are not implemented yet this is a lowcost method for now
+        let cid_pair = self.compute_tipset_state(&block_headers, &HashMap::new())?;
```
austinabell, line 161:
```suggestion
        let cid_pair = self.compute_tipset_state(tipset.blocks(), &HashMap::new())?;
```

the `to_vec()` clone is unnecessary, also if you agree to take out the unnecessary hashmap that can be done here also

blockchain/state_manager/src/lib.rs
```diff
@@ -117,6 +131,98 @@ where
...
+                    header: s.clone(),
+                    bls_messages,
+                    secp_messages,
+                })
+            })
+            .collect::<Result<Vec<Block>, _>>()?;
+        let full_tipset = FullTipset::new(blocks)?;
+        // convert tipset to fulltipset
+        self.apply_blocks(&full_tipset)
+    }
```
austinabell, line 224: Where is this logic coming from? I can't seem to map it to anything

blockchain/chain_sync/src/sync.rs
```diff
@@ -65,7 +66,7 @@ pub struct ChainSyncer<DB, TBeacon> {
     beacon: Arc<TBeacon>,
 
     /// manages retrieving and updates state objects
-    state_manager: StateManager<DB>,
+    state_manager: Arc<RwLock<StateManager<DB>>>,
```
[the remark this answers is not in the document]
AshantiMutinta: There exists a check_block_msgs which doesn't contain the self reference and that method takes in Tbeacon which cannot be shared between threads and if we want the cache to actually work while it is in the scope of the sync object. We could probably use this statemanger without the need for self if we don't have the hashmap.

Might be better just to remove the cache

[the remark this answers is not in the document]
austinabell: I wasn't suggesting removing the Arc, which I think you may be assuming, I'm just saying remove the lock on the whole state manager and just wrap the cache in it

blockchain/chain_sync/src/sync.rs
AshantiMutinta: The unwrapped Ok in this instance won't live as long as this work_result_addr and that is why I have to unwrap

blockchain/chain_sync/src/sync.rs
AshantiMutinta: Just remnants of dealing with a shared reference bugs, still cleaning up for sure :)

blockchain/chain_sync/src/sync.rs
austinabell: I don't have a problem with that

blockchain/state_manager/src/lib.rs
AshantiMutinta: computeTipsetState in lotus

blockchain/chain_sync/src/sync.rs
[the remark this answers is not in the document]
AshantiMutinta: Sorry I don't think I explained it right.

I meant you have a tipset_state that takes in a mutable reference in order to update that cache. Through the validations though, that check_block_msgs takes in the statemanger which also updates again using tipset_state. I might be wrong but with arcs we can use get_mut but it panics whenever more than reference is holding to it. 

In this case we have a shared reference that needs a write opretion, the future that is going to use this statemanger to await and the code under the construction of this future that both call tipset_state.

Unless I am understanding you wrong that is

blockchain/state_manager/src/lib.rs
AshantiMutinta: I will just remove it for now, the issue with it being a lazy_static as well was rust doesn't have generic constants, currently being worked on as well though

blockchain/state_manager/src/lib.rs
AshantiMutinta: Yeah in general wasn't sure how detailed we wanted to go with our tracing but I can expand on this for sure

blockchain/chain_sync/src/sync.rs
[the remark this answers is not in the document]
austinabell: I'm saying you don't need a mutable reference for tipset_state because it only updates the cache? If the cache is wrapped in a mutex you don't need mutable ref to update

ec2:
Was there a reason why you removed the concurrent validation of block headers?

austinabell:
> Was there a reason why you removed the concurrent validation of block headers?

I was going to come back to this in a bit but might as well now, before it was spawning a blocking task which is just synconous (I didn't look too much in detail at the last PR just assumed you guys had verified it) Think it just needs to be changed to `task::spawn` from what it was

AshantiMutinta:
> tually don't know how the `FuturesUnordered` handles these boxed async functions being passed in or when they are executed but I can look into it)

Yeah they have to be boxed because each future has a different type even though in essence it looks the same. So adding different types is according to the future trait rather than a concrete future type. 

It looked like it was returning join handles with futures inside of them and using spawn_blocking so in essence didn't see much of a difference between handling them where they are vs handling them at the end where the errors are being collected

blockchain/chain_sync/src/sync.rs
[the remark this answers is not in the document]
AshantiMutinta: Oh I see what you are saying now, I thinking in an async vs sync lock perspective. I will switch it

blockchain/chain_sync/src/sync.rs
```diff
@@ -620,25 +630,28 @@ where
...
+            .tipset_state(&parent_tipset)
+            .await
+            .map_err(|_| Error::Validation("Could not update state".to_owned()))?;
         let work_addr_result = self
             .state_manager
-            .get_miner_work_addr(&parent_tipset.parent_state(), header.miner_address());
+            .get_miner_work_addr(&state_root, header.miner_address());
+
+        // temp header needs to live long enough in static context returned by task::spawn
+        let temp_header = block.header().clone();
```
austinabell, line 648: This clone is expensive, probably worth considering not using that specific function to verify the signature (all you need is the cid bytes, so cloning the whole block header is unnecessary). This wasn't introduced in this PR so feel free to leave, just commenting because I just noticed it

blockchain/state_manager/src/lib.rs
```diff
@@ -122,6 +134,81 @@ where
...
+    pub fn compute_tipset_state<'a>(
+        &'a self,
+        blocks: &[BlockHeader],
+    ) -> Result<(Cid, Cid), Box<dyn StdError>> {
+        span!("compute_tipset_state", {
+            if blocks.len() > 2
+                && blocks
+                    .iter()
+                    .zip(blocks.iter().skip(0))
+                    .any(|(a, b)| a.miner_address() == b.miner_address())
```
austinabell, line 185: this is just comparing each index with the i+1 ? you need to make sure there are no duplicates, not just no adjacent equivalents

blockchain/state_manager/src/lib.rs
```diff
@@ -122,6 +138,83 @@ where
...
 
+    pub async fn tipset_state(&self, tipset: &Tipset) -> Result<(Cid, Cid), Box<dyn StdError>> {
+        span!("tipset_state", {
+            trace!("tipset {:?}", tipset.cids());
+            // if exists in cache return
+            if let Some(cid_pair) = self.cache.read().await.get(&tipset.key()) {
+                return Ok(cid_pair.clone());
+            }
+
+            if tipset.len() == 0 {
```
ec2, line 149: tipset.is_empty()

AshantiMutinta: hmmm I think the is_empty() is not a function on the tipset struct, checking right now

austinabell: yeah, it isn't. dw about that now

blockchain/chain_sync/src/sync.rs
```diff
@@ -617,28 +631,36 @@ where
 
         let b = block.clone();
 
-        let db = Arc::clone(&self.chain_store.db);
         let parent_clone = parent_tipset.clone();
         // check messages to ensure valid state transitions
-        let x = task::spawn_blocking(move || Self::check_block_msgs(db, b, &parent_clone));
+        let sm = self.state_manager.clone();
+        let x = task::spawn_blocking(move || Self::check_block_msgs(sm, b, parent_clone));
```
austinabell, line 637: Did it turn out that spawn_blocking was correct? Seems weird because I would think it would make this syncronous (just confirming this was looked into)

blockchain/state_manager/src/lib.rs
```diff
@@ -50,6 +61,11 @@ where
         let state = StateTree::new_from_root(self.bs.as_ref(), state_cid).map_err(Error::State)?;
         state.get_actor(addr).map_err(Error::State)
     }
+
+    pub fn get_bs(&self) -> Arc<DB> {
```
austinabell, line 65: maybe could be better named? I don't have an opinion on this but maybe for consistency something like `blockstore` or `get_blockstore`? Some other similar functions use the name `store` but might be better to be more explicit of what it actually is

blockchain/chain_sync/src/sync.rs
AshantiMutinta: Yeah I talked to Eric and I think what we determined is that spawn_blocking since it is very CPU intensinve

blockchain/state_manager/src/lib.rs
AshantiMutinta: Oh yeah this was silly of me, will change that asap

blockchain/chain_sync/src/sync.rs
ec2: I talked to Ashanti this morning. It is correct. Spawn blocking isn't to make it sychronous or anything. It to spawn a thread from a dedicated pool of threads in the async std runtime. This dedicated pool of threads is for CPU intensive tasks (which in this case checking signatures can be quite heavy (hence blocking)).

austinabell: oops yeah, I had a lapse and was mixing up spawn_blocking with block_on

blockchain/state_manager/src/lib.rs
austinabell, line 149: This should be checking if the tipset height is ==0, no?

blockchain/state_manager/src/lib.rs
```diff
@@ -122,6 +138,83 @@ where
...
+                    Ok(Block {
+                        header: s.clone(),
+                        bls_messages,
+                        secp_messages,
+                    })
+                })
+                .collect::<Result<Vec<Block>, _>>()?;
+            // convert tipset to fulltipset
+            let full_tipset = FullTipset::new(blocks)?;
+            let chain_rand = ChainRand::new(full_tipset.to_tipset().key().to_owned());
```
austinabell, line 213: This is cloning all tipset headers (indication from `to_tipset`) when you really only need to generate Cids and should be able to do that through references. Could you do that in this PR or is there a reason not to?

blockchain/state_manager/src/lib.rs
```diff
@@ -122,6 +138,83 @@ where
...
+                    .filter(|val| val.miner_address() == s.miner_address())
+                    .take(2)
+                    .count()
+            };
+            if blocks.iter().any(|s| check_for_duplicates(s) > 1) {
+                // Duplicate Miner found
+                return Err(Box::new(Error::Other(
+                    "Could not get message receipts".to_string(),
+                )));
+            }
```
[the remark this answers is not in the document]
AshantiMutinta: as a general rule I usually avoid indexes where I can when using iterators, I think following high order functions is so easier cause you can essentially tell what it is doing without looking at any index arithmetic. I think from a performance perspective it is not as bad because filter is lazy, take is shortcircuiting and count will iter on the take iterator

blockchain/state_manager/src/lib.rs
```diff
@@ -122,6 +138,83 @@ where
...
+                        header: s.clone(),
+                        bls_messages,
+                        secp_messages,
+                    })
+                })
+                .collect::<Result<Vec<Block>, _>>()?;
+            // convert tipset to fulltipset
+            let full_tipset = FullTipset::new(blocks)?;
+            let chain_rand = ChainRand::new(full_tipset.to_tipset().key().to_owned());
+            self.apply_blocks(&full_tipset, &chain_rand)
```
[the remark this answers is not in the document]
AshantiMutinta: Yeah would rather do that later just so that the scope doesn't stray out too much

blockchain/state_manager/src/lib.rs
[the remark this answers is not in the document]
austinabell: So the argument for this is readability or are you saying you don't like indexing wrt performance? That's fine I mean it's such a trivial cost to have duplicate address comparisons, just I guess myself personally the code was confusing to read what it was actually doing.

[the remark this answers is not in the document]
AshantiMutinta: What is funny is actually find the suggestion much harder to read haha I think it is because I avoid indexes most of the time, I find high order functions much easier to follow since they denote functional flow

blockchain/state_manager/src/lib.rs
AshantiMutinta: That makes sense, I think it was unexperience with the codebase and it made sense form a Type perspective.

I used the block_header but had to add a cloned because TipsetKeys needs owned Cids

blockchain/state_manager/src/lib.rs
AshantiMutinta: No height method on Tipset struct. Do you mean weight?

austinabell: Sorry I mean epoch, I use those synonymously

PR #469, opened 2020-06-01
Implement Wallet

### LINKED ISSUE #464: Implement Wallet for key management
**Issue summary**

The node needs to have a key store for mining. This means we have to implement a Wallet to manage keys. 

Here is how Lotus does it: 
https://github.com/filecoin-project/lotus/blob/master/chain/wallet/wallet.go

flodesi:
**Summary of changes**
- implement wallet for our node to manage keys
- some private key signing functions that are required for the implementation of the wallet

Closes #464

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,4 +32,5 @@ members = [
     "utils/test_utils",
     "utils/commcid",
     "types",
+    "wallet",
 ]
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "wallet"
+version = "0.1.0"
+authors = ["ChainSafe Systems <info@chainsafe.io>"]
+edition = "2018"
+
+[dependencies]
+thiserror = "1.0"
+address = { package = "forest_address", path = "../vm/address", version = "0.2" }
+crypto = { package = "forest_crypto", path = "../crypto" }
--- a/wallet/src/errors.rs
+++ b/wallet/src/errors.rs
@@ -1,0 +1,15 @@
+use thiserror::Error;
+
+#[derive(Debug, PartialEq, Error)]
+pub enum Error {
+    /// info that corresponds to key does not exist
+    #[error("Key info not found")]
+    KeyInfo,
+    /// Key already exists in keystore
+    #[error("Key already exists")]
+    KeyExists,
+    #[error("key not found")]
+    NoKey,
+    #[error("{0}")]
+    Other(String),
+}
--- a/wallet/src/keystore.rs
+++ b/wallet/src/keystore.rs
@@ -1,0 +1,63 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use super::errors::Error;
+use std::collections::HashMap;
+
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct KeyInfo {
+    key_type: String,
+    private_key: Vec<u8>,
+}
+
+impl KeyInfo {
+    pub fn key_type(&self) -> String {
+        self.key_type.clone()
+    }
+
+    pub fn private_key(&self) -> Vec<u8> {
+        self.private_key.clone()
+    }
+}
+
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct KeyStore {
+    pub m: HashMap<String, KeyInfo>,
+}
+
+impl KeyStore {
+    pub fn new() -> Self {
+        KeyStore { m: HashMap::new() }
+    }
+
+    /// Return all of the keys that are stored in the KeyStore
+    pub fn list(&self) -> Vec<String> {
+        let mut out_vec = Vec::new();
+
+        for (key, _) in self.m.iter() {
+            out_vec.push(key.clone())
+        }
+        out_vec
+    }
+
+    /// Return Keyinfo that corresponds to a given key
+    pub fn get(&self, k: &String) -> Result<&KeyInfo, Error> {
+        self.m.get(k).map_or_else(|| Err(Error::KeyInfo), |v| Ok(v))
+    }
+
+    /// Save a key key_info pair to the KeyStore
+    pub fn put(&mut self, key: String, key_info: KeyInfo) -> Result<(), Error> {
+        if self.m.contains_key(&key) {
+            return Err(Error::KeyExists);
+        }
+        self.m.insert(key, key_info);
+        Ok(())
+    }
+
+    pub fn remove(&mut self, key: String) -> Result<(), Error> {
+        match self.m.remove(&key) {
+            Some(_t) => Ok(()),
+            None => Err(Error::NoKey),
+        }
+    }
+}
--- a/wallet/src/lib.rs
+++ b/wallet/src/lib.rs
@@ -1,0 +1,10 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+mod errors;
+mod keystore;
+mod wallet;
+
+pub use errors::*;
+pub use keystore::*;
+pub use wallet::*;
--- a/wallet/src/wallet.rs
+++ b/wallet/src/wallet.rs
@@ -1,0 +1,152 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use super::errors::Error;
+use super::{KeyInfo, KeyStore};
+use address::Address;
+use crypto::{Signature, SignatureType};
+use std::collections::HashMap;
+use std::str::FromStr;
+
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct Key {
+    key_info: KeyInfo,
+    public_key: Vec<u8>,
+    address: Address,
+}
+
+impl Key {
+    pub fn new(key_info: &KeyInfo) -> Self {
+        unimplemented!()
+    }
+}
+
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct Wallet {
+    keys: HashMap<Address, Key>,
+    keystore: KeyStore,
+}
+
+impl Wallet {
+    pub fn new(keystore: KeyStore) -> Self {
+        Wallet {
+            keys: HashMap::new(),
+            keystore,
+        }
+    }
+
+    pub fn new_from_keys(key_vec: Vec<Key>) -> Self {
+        let mut keys: HashMap<Address, Key> = HashMap::new();
+        for item in key_vec.clone() {
+            keys.insert(item.address, item);
+        }
+        Wallet {
+            keys,
+            keystore: KeyStore::new(),
+        }
+    }
+
+    pub fn find_key(&mut self, addr: &Address) -> Result<Key, Error> {
+        let key = self.keys.get(&addr);
+        if let Some(k) = key {
+            return Ok(k.clone());
+        }
+        let mut owned_string = "wallet-".to_owned();
+        owned_string.push_str(addr.to_string().as_ref());
+        let key_info = self.keystore.get(&owned_string)?;
+        let new_key = Key::new(key_info);
+        self.keys.insert(addr.clone(), new_key.clone());
+        Ok(new_key)
+    }
+
+    /// TODO will need to implement this after more research about signing messages is done
+    pub fn sign(&mut self, addr: &Address, msg: Vec<u8>) -> Result<Signature, Error> {
+        unimplemented!()
+    }
+
+    pub fn export(&mut self, addr: &Address) -> Result<KeyInfo, Error> {
+        let k = self.find_key(addr)?;
+        Ok(k.key_info)
+    }
+
+    pub fn import(&mut self, key_info: &KeyInfo) -> Result<Address, Error> {
+        let k = Key::new(key_info);
+        let mut owned_string = "wallet-".to_owned();
+        owned_string.push_str(k.address.to_string().as_ref());
+        self.keystore.put(owned_string, k.key_info)?;
+        Ok(k.address)
+    }
+
+    pub fn list_addrs(&self) -> Result<Vec<Address>, Error> {
+        let mut all = self.keystore.list();
+        all.sort();
+        let mut out = Vec::new();
+        for i in all {
+            if i.starts_with("wallet-") {
+                // TODO replace this with strip_prefix after it has been added to stable rust
+                let name = i.trim_start_matches("wallet-");
+                let addr = Address::from_str(name).map_err(|err| Error::Other(err.to_string()))?;
+                out.push(addr);
+            }
+        }
+        Ok(out)
+    }
+
+    pub fn get_default(&self) -> Result<Address, Error> {
+        let key_info = self.keystore.get(&"default".to_string())?;
+        let k = Key::new(key_info);
+        Ok(k.address)
+    }
+
+    pub fn set_default(&mut self, addr: Address) -> Result<(), Error> {
+        let mut owned_string = "wallet-".to_owned();
+        owned_string.push_str(addr.to_string().as_ref());
+        let key_info = self.keystore.get(&owned_string)?.clone();
+        // TODO change this code to not exit if there is no kv pair with default key in keystore
+        self.keystore.remove("wallet-".to_string())?;
+        self.keystore.put("wallet-".to_string(), key_info)?;
+        Ok(())
+    }
+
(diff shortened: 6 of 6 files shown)


key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Return a new Wallet with a given KeyStore
+    pub fn new(keystore: KeyStore) -> Self {
+        Wallet {
+            keys: HashMap::new(),
+            keystore,
+        }
+    }
+
+    /// Return a new Wallet with Keys contructed by a given list of Keys
+    pub fn new_from_keys(key_vec: Vec<Key>) -> Self {
```
austinabell, line 53: Could just have this generic based on an iterator, hard to tell exact use case because there are no usages of this function but no reason to constrain to needing specifically a Vec.

Something like:
```suggestion
    pub fn new_from_keys(key_vec: impl IntoIterator<Item = Key>) -> Self {
```

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+        Wallet {
+            keys,
+            keystore: KeyStore::new(),
+        }
+    }
+
+    /// Return the Key that is resolved by a given Address, return Error otherwise
+    pub fn find_key(&mut self, addr: &Address) -> Result<Key, Error> {
+        let key = self.keys.get(&addr);
+        if let Some(k) = key {
```
austinabell, line 67: Optional:
```suggestion
        if let Some(k) = self.keys.get(&addr) {
```

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+            return Ok(k.clone());
+        }
+        let mut owned_string = "wallet-".to_owned();
+        owned_string.push_str(addr.to_string().as_ref());
+        let key_info = self.keystore.get(&owned_string)?;
```
austinabell, line 72:
```suggestion
        let key_string = format!("wallet-{}", addr.to_string());
        let key_info = self.keystore.get(&key_string)?;
```

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+        let new_key = Key::new(key_info)?;
+        self.keys.insert(addr.clone(), new_key.clone());
```
austinabell, line 74: Doesn't matter tho, can leave this
```suggestion
        self.keys.insert(*addr, new_key.clone());
```

key_management/src/sigs.rs
```diff
@@ -0,0 +1,84 @@
...
+            Ok(addr)
+        }
+    }
+}
+
+/// Sign takes in SignatureType, private key and message. Returns a Signature for that message
+pub fn sign(
+    sig_type: SignatureType,
+    private_key: Vec<u8>,
+    msg: Vec<u8>,
```
austinabell, line 47: This should be a slice, should always only need to read bytes to generate a signature

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+        Ok(new_key)
+    }
+
+    /// Return the resultant Signature after signing a given message
+    pub fn sign(&mut self, addr: &Address, msg: Vec<u8>) -> Result<Signature, Error> {
```
austinabell, line 79: By extension of the previous review, msg should be a slice here

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+        Ok(k.address)
+    }
+
+    /// Set a default KeyInfo to the Wallet
+    pub fn set_default(&mut self, addr: Address) -> Result<(), Error> {
+        let mut owned_string = "wallet-".to_owned();
+        owned_string.push_str(addr.to_string().as_ref());
+        let key_info = self.keystore.get(&owned_string)?.clone();
+        // TODO change this code to not exit if there is no kv pair with default key in keystore
+        self.keystore.remove("default".to_string())?; // This line should unregister current default key then continue
```
austinabell, line 132: Can just change the remove function to just return an option?

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub fn get_default(&self) -> Result<Address, Error> {
+        let key_info = self.keystore.get(&"default".to_string())?;
+        let k = Key::new(key_info)?;
+        Ok(k.address)
+    }
+
+    /// Set a default KeyInfo to the Wallet
+    pub fn set_default(&mut self, addr: Address) -> Result<(), Error> {
+        let mut owned_string = "wallet-".to_owned();
+        owned_string.push_str(addr.to_string().as_ref());
```
austinabell, line 129: Can use `format!` for this again

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+        self.keystore.remove("default".to_string())?; // This line should unregister current default key then continue
+        self.keystore.put("default".to_string(), key_info)?;
+        Ok(())
+    }
+
+    /// Generate a new Key that fits the requirement of the given SignatureType
+    pub fn generate_key(&mut self, typ: SignatureType) -> Result<Address, Error> {
+        let key = generate_key(typ)?;
+        let mut owned_string = "wallet-".to_owned();
+        owned_string.push_str(key.address.to_string().as_ref());
```
austinabell, line 141: format macro

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+                .put("default".to_string(), key.key_info.clone())
+                .map_err(|err| Error::Other(err.to_string()))?;
+        }
+
+        Ok(key.address)
+    }
+
+    /// Return whether or not the Wallet contains a Key that is resolved by the supplied Address
+    pub fn has_key(&mut self, addr: &Address) -> bool {
+        self.find_key(addr).map_or_else(|_| false, |_| true)
```
austinabell, line 156:
```suggestion
        self.find_key(addr).is_ok()
```

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+/// Return the String that corresponds to each Signature type
+pub fn kstore_sig_type(typ: SignatureType) -> String {
+    match typ {
+        SignatureType::Secp256 => "secp256k1".to_string(),
+        _ => "bls".to_string(),
```
austinabell, line 164: I don't think you need this at all, but if you do, just make it explicit (could run into potentially an issue if other types are added in future)
```suggestion
        SignatureType::BLS => "bls".to_string(),
```

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+/// Return the SignatureType that corresponds to the supplied String
+pub fn act_sig_type(typ: String) -> SignatureType {
+    if typ == "secp256k1" {
+        return SignatureType::Secp256;
+    }
+    SignatureType::default()
+}
```
austinabell, line 174: If you keep this function (again with this I don't think you need, should return an option or result and each key type should be handled with a match that looks something like:
```
    match typ {
        "secp256k1" => SignatureType::Secp256,
        "bls" => SignatureType::BLS,
    }
```

SignatureType default impl probably shouldn't even exist (can probably or should be removed) and this defaults unmatched types to BLS always, which is weird

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+
+/// Generate a new Key that satisfies the given SignatureType
+fn generate_key(typ: SignatureType) -> Result<Key, Error> {
+    let private_key = sigs::generate(typ)?;
```
austinabell, line 178: Kinda misleading to have the module as sigs, because it's generating a private key here, but just thinking out loud

flodesi: Yeah, I'll change that, I was just following Lotus' naming scheme there

key_management/src/keystore.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /// Return all of the keys that are stored in the KeyStore
+    pub fn list(&self) -> Vec<String> {
+        let mut out_vec = Vec::new();
+
+        for (key, _) in self.m.iter() {
+            out_vec.push(key.clone())
+        }
+        out_vec
```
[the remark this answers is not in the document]
ec2: So someone using the node can see what keys they have on the node, since you can have multiple (list keys is an rpc method call)

key_management/src/keystore.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Return a clone of the private_key
+    pub fn private_key(&self) -> Vec<u8> {
+        self.private_key.clone()
+    }
+}
+
+/// KeyStore struct, this contains a HashMap that is a set of KeyInfos resolved by their Address
+#[derive(Default, Clone, PartialEq, Debug, Eq)]
+pub struct KeyStore {
+    pub m: HashMap<String, KeyInfo>,
```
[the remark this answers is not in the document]
flodesi: completely forgot about that. Will make that change.

key_management/src/keystore.rs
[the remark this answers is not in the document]
flodesi: It was used as a helper in list_addrs to get a Vec of all the addresses.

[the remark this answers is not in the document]
austinabell: Yeah all I mean is that this could just return an iterator, I don't see the need for it to create a vector when only used in one place as that

key_management/src/sigs.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use super::errors::Error;
+use address::Address;
+use bls_signatures::{PrivateKey as BlsPrivate, Serialize};
+use crypto::{Signature, SignatureType};
+use rand::rngs::OsRng;
+use secp256k1::{Message as SecpMessage, PublicKey as SecpPublic, SecretKey as SecpPrivate};
+
+/// Return the public key for a given private_key and SignatureType
+pub fn to_public(sig_type: SignatureType, private_key: Vec<u8>) -> Result<Vec<u8>, Error> {
```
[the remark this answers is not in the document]
flodesi: Yeah the reason for having the Vec was that the BLS private key struct and and SECP256K1 private key structs couldn't be used interchangeably here so the way I found to get around that was just to convert each of them to a Vec<u8>

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+use address::Address;
+use crypto::{Signature, SignatureType};
+use std::collections::HashMap;
+use std::str::FromStr;
+
+/// A Key, this contains a key_info, address, and public_key which holds the key type and private key
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct Key {
+    key_info: KeyInfo,
+    // Vec<u8> will be used because Eq has not been implemented for PublicKey type
```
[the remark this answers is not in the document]
flodesi: BLS and SECP256K1 each have their own private and public key structs.

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+use super::errors::Error;
+use super::{sigs, KeyInfo, KeyStore};
+use address::Address;
+use crypto::{Signature, SignatureType};
+use std::collections::HashMap;
+use std::str::FromStr;
+
+/// A Key, this contains a key_info, address, and public_key which holds the key type and private key
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct Key {
```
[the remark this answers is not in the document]
flodesi: True, was matching what Lotus was doing, Will remove and refactor.

key_management/src/wallet.rs
```diff
@@ -0,0 +1,181 @@
...
+        })
+    }
+}
+
+/// This is a Wallet, it contains 2 HashMaps:
+/// - keys which is a HashMap of Keys resolved by their Address
+/// - keystore which is a HashMap of KeyInfos resolved by their Address
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct Wallet {
+    keys: HashMap<Address, Key>,
```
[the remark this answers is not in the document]
flodesi: Will remove that field when completing the previous.

key_management/src/keystore.rs
```diff
@@ -0,0 +1,83 @@
...
+/// KeyStore struct, this contains a HashMap that is a set of KeyInfos resolved by their Address
+pub trait KeyStore {
+    fn list(&self) -> Vec<String>;
+    fn get(&self, k: &str) -> Result<&KeyInfo, Error>;
+    fn put(&mut self, key: String, key_info: KeyInfo) -> Result<(), Error>;
+    fn remove(&mut self, key: String) -> Option<KeyInfo>;
+}
+
+#[derive(Default, Clone, PartialEq, Debug, Eq)]
+pub struct MemKeyStore {
```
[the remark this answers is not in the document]
flodesi: what do you mean?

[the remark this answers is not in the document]
austinabell: This is just in memory storage, when the node stops and starts again, these keys are lost. There should be a way to persist the keys to be reused in some way

key_management/src/keystore.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Save a key key_info pair to the KeyStore
+    fn put(&mut self, key: String, key_info: KeyInfo) -> Result<(), Error> {
+        if self.key_info.contains_key(&key) {
+            return Err(Error::KeyExists);
+        }
+        self.key_info.insert(key, key_info);
+        Ok(())
+    }
+
+    /// Remove the Key and corresponding key_info from the KeyStore
```
austinabell, line 79: can you move the doc comments to the trait definition?

key_management/src/wallet.rs
```diff
@@ -0,0 +1,344 @@
...
+    // Vec<u8> is used because The public keys for BLS and SECP256K1 are not of the same type
+    public_key: Vec<u8>,
+    address: Address,
+}
+
+impl From<KeyInfo> for Key {
+    fn from(key_info: KeyInfo) -> Self {
+        let public_key =
+            wallet_helpers::to_public(*key_info.key_type(), key_info.private_key()).unwrap();
+        let address = wallet_helpers::new_address(*key_info.key_type(), &public_key).unwrap();
```
austinabell, line 25: if this is error prone, should probably be a `TryFrom` implementation, not a `From` so that the errors are handled and not just panic

key_management/src/wallet.rs
```diff
@@ -0,0 +1,344 @@
...
+        for item in key_vec.into_iter() {
+            keys.insert(item.address, item);
+        }
+        let key_store = MemKeyStore::new();
+        Wallet {
+            keys,
+            keystore: key_store,
+        }
+    }
+}
```
austinabell, line 57: Would be better if this was generic and didn't depend on the MemKeyStore type

key_management/src/wallet.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    /// Return the KeyInfo for a given Address
+    pub fn export(&mut self, addr: &Address) -> Result<KeyInfo, Error> {
+        let k = self.find_key(addr)?;
+        Ok(k.key_info)
+    }
+
+    /// Add Key_Info to the Wallet, return the Address that resolves to this newly added KeyInfo
+    pub fn import(&mut self, key_info: &KeyInfo) -> Result<Address, Error> {
+        let k = Key::from(key_info.clone());
```
austinabell, line 101:
```suggestion
    pub fn import(&mut self, key_info: KeyInfo) -> Result<Address, Error> {
        let k = Key::from(key_info);
```
This should be moved in to avoid always cloning if possible

key_management/src/keystore.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Return a clone of the private_key
+    pub fn private_key(&self) -> &Vec<u8> {
+        &self.private_key
+    }
+}
+
+/// KeyStore struct, this contains a HashMap that is a set of KeyInfos resolved by their Address
+pub trait KeyStore {
+    fn list(&self) -> Vec<String>;
+    fn get(&self, k: &str) -> Result<&KeyInfo, Error>;
```
austinabell, line 40:
```suggestion
    fn get(&self, k: &str) -> Result<KeyInfo, Error>;
```
This trait only works for mem key store because it has the key info in memory, for a persisted storage this won't be the case

key_management/src/wallet.rs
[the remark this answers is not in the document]
austinabell: For something like this, it isn't consensus critical and we do not need to match 1:1, did you discover that address was required after all?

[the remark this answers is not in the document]
austinabell: For something like this, you do not need to match lotus 1:1 as it's not consensus critical, did you discover that the address was required to keep in this struct?

PR #480, opened 2020-06-05
Update Reward actor to new spec

### LINKED ISSUE #459: Update reward actor
**Task summary**

The reward actor has been updated quite a bit and needs to reflect those changes. 

**Specification reference**

- [reward actor](https://github.com/filecoin-project/specs-actors/tree/master/actors/builtin/reward)

austinabell:
**Summary of changes**
- self explanatory: https://github.com/filecoin-project/specs-actors/tree/52599b21919df07f44d7e61cc028e265ec18f700/actors/builtin/reward

Closes #459 

### DIFF
--- a/vm/actor/src/builtin/network.rs
+++ b/vm/actor/src/builtin/network.rs
@@ -1,0 +1,5 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+/// The expected number of block producers in each epoch.
+pub(crate) const EXPECTED_LEADERS_PER_EPOCH: u64 = 5;
--- a/vm/actor/src/builtin/network.rs
+++ b/vm/actor/src/builtin/network.rs
@@ -1,5 +1,13 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+pub const EPOCH_DURATION_SECONDS: u64 = 25;
+pub const SECONDS_IN_HOUR: u64 = 3600;
+pub const SECONDS_IN_DAY: u64 = 86400;
+pub const SECONDS_IN_YEAR: u64 = 31556925;
+pub const EPOCHS_IN_HOUR: u64 = SECONDS_IN_HOUR / EPOCH_DURATION_SECONDS;
+pub const EPOCHS_IN_DAY: u64 = SECONDS_IN_DAY / EPOCH_DURATION_SECONDS;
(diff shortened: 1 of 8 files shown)


vm/actor/src/builtin/network.rs
```diff
@@ -0,0 +1,13 @@
...
+pub const EPOCH_DURATION_SECONDS: u64 = 25;
+pub const SECONDS_IN_HOUR: u64 = 3600;
+pub const SECONDS_IN_DAY: u64 = 86400;
+pub const SECONDS_IN_YEAR: u64 = 31556925;
+pub const EPOCHS_IN_HOUR: u64 = SECONDS_IN_HOUR / EPOCH_DURATION_SECONDS;
+pub const EPOCHS_IN_DAY: u64 = SECONDS_IN_DAY / EPOCH_DURATION_SECONDS;
+pub const EPOCHS_IN_YEAR: u64 = SECONDS_IN_YEAR / EPOCH_DURATION_SECONDS;
+
+/// The expected number of block producers in each epoch.
+pub const EXPECTED_LEADERS_PER_EPOCH: u64 = 5;
```
ec2, line 13: I wonder why specs-actors doesn't declare this as a const like the rest... Must be some Go-ism

austinabell: yeah my guess is that they thought it was needed to be that type to convert into a big int or it's just something not switched after they changed something else. It can be changed to a const in their code

austinabell: https://github.com/filecoin-project/specs-actors/pull/449

vm/actor/src/builtin/reward/state.rs
```diff
@@ -1,110 +1,59 @@
...
-        rewards.remove_all(&key)?;
-        for rew in remaining_rewards {
-            rewards.add(key.clone().into(), rew)?;
-        }
-
-        // Update rewards multimap root and total
-        self.reward_map = rewards.root()?;
-        self.reward_total -= &withdrawable_sum;
-        Ok(withdrawable_sum)
+    pub(super) fn get_effective_network_time(
```
dutterbutter, line 47: I don't think it matters but `get_effective_network_time` looks to exist in the actor implementation rather than the state.

austinabell: it does, but I chose not to follow here, as it's much cleaner this way

austinabell: made 0 sense to me why they had it the way they did

PR #484, opened 2020-06-08
Adds Json serialization for Message Receipts and ActorState

ec2:
**Summary of changes**
- Adds Json serialization for Message Receipts and ActorState
- This is only some of the structures that need JSON serialization for State API. There are some more that we do not have yet pertaining to the Miner. Will need to coordinate with Dustin who's doing the miner actor. 

Closes 

### DIFF
--- a/vm/message/src/message_receipt.rs
+++ b/vm/message/src/message_receipt.rs
@@ -58,7 +58,7 @@ pub mod json {
     where
         D: Deserializer<'de>,
     {
-        #[derive(Serialize, Deserialize)]
+        #[derive(Deserialize)]
         #[serde(rename_all = "PascalCase")]
         struct MessageReceiptDe {
             exit_code: u64,
--- a/vm/message/src/message_receipt.rs
+++ b/vm/message/src/message_receipt.rs
@@ -72,9 +72,9 @@ pub mod json {
             gas_used,
         } = Deserialize::deserialize(deserializer)?;
         Ok(MessageReceipt {
-            exit_code: ExitCode::from_u64(exit_code).ok_or(de::Error::custom(
-                "MessageReceipt deserialization: Could not turn u64 to ExitCode",
-            ))?,
+            exit_code: ExitCode::from_u64(exit_code).ok_or_else(|| {
+                de::Error::custom("MessageReceipt deserialization: Could not turn u64 to ExitCode")
+            })?,
             return_data: Serialized::new(return_data),
             gas_used,
         })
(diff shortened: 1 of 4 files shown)


vm/message/src/message_receipt.rs
```diff
@@ -11,3 +11,72 @@ pub struct MessageReceipt {
...
+            gas_used: m.gas_used,
+        }
+        .serialize(serializer)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<MessageReceipt, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Serialize, Deserialize)]
```
austinabell, line 61:
```suggestion
        #[derive(Deserialize)]
```
No need at all, doesn't actually matter though

vm/src/actor_state.rs
```diff
@@ -43,3 +43,79 @@ impl ActorState {
...
+        #[derive(Serialize)]
+        #[serde(rename_all = "PascalCase")]
+        struct ActorStateSer<'a> {
+            #[serde(with = "cid::json")]
+            pub code: &'a Cid,
+            #[serde(rename = "Head", with = "cid::json")]
+            pub state: &'a Cid,
+            #[serde(rename = "Nonce")]
+            pub sequence: u64,
+            pub balance: &'a String,
```
austinabell, line 82:
```suggestion
            pub balance: String,
```
Don't need to take a reference to it if you're already converting to string

vm/src/actor_state.rs
```diff
@@ -43,3 +43,79 @@ impl ActorState {
...
+        }
+        ActorStateSer {
+            code: &m.code,
+            state: &m.state,
+            sequence: m.sequence,
+            balance: &m.balance.to_str_radix(10),
```
austinabell, line 88:
```suggestion
            balance: m.balance.to_str_radix(10),
```
Continuation of previous comment.

Also just so you know, the default `to_string` implementation is base 10, I just havent commented before because this is more clear I guess

vm/src/actor_state.rs
```diff
@@ -43,3 +43,79 @@ impl ActorState {
...
+        }
+        .serialize(serializer)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<ActorState, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Serialize, Deserialize)]
```
austinabell, line 97:
```suggestion
        #[derive(Deserialize)]
```
Again doesn't really matter but no need since it's private and only used as a helper struct

PR #493, opened 2020-06-11
Ashanti/winning posts

### LINKED ISSUE #471: [state manager/sync] - Implement get for winning PoSts
**Issue summary**

Part of the syncing process is to verify winning PoSts. 

- [Lotus - GetSectorForWinningPoSts](https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/utils.go#L151)
- [Lotus - VerifyWinningPoStProof](https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L754)

AshantiMutinta:
**Summary of changes**
- Implements Winning Posts for StateManager Sync
- adds a util fie for StateManager

Closes #471 


blockchain/state_manager/src/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+    state_manager: &StateManager<DB>,
+    st: &Cid,
+    address: &Address,
+    rand: &[u8; 32],
+) -> Result<Vec<SectorInfo>, Error>
+where
+    DB: BlockStore,
+{
+    let mut miner_actor_state: State = state_manager.load_actor_state(&address, &st)?;
+    let sector_set = miner_actor_state.compute_proving_set(&*state_manager.get_block_store())?;
```
AshantiMutinta, line 24: Need to change this to use some bitfield methods

blockchain/chain_sync/src/sync.rs
```diff
@@ -729,6 +736,99 @@ where
         Ok(())
     }
 
+    pub async fn verify_winning_posts_proof(
```
austinabell, line 739:
```suggestion
    pub async fn verify_winning_post_proof(
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -729,6 +736,99 @@ where
...
+        }
+
+        if let Some(first_proof) = block.win_post_proof().first() {
+            if from_utf8(&first_proof.proof_bytes)
+                .map(|buf| buf == "valid proof")
+                .unwrap_or_default()
+            {
+                return Ok(());
+            }
+        }
```
austinabell, line 758: They only use this when an insecure validation flag is set for testing. We don't have this option so should just leave out for now

blockchain/chain_sync/src/sync.rs
```diff
@@ -729,6 +736,99 @@ where
...
+        })?;
+        let miner_id = match block.miner_address().payload() {
+            Payload::ID(new_id) => Address::new_id(*new_id),
+            _ => {
+                return Err(Error::Validation(format!(
+                    "failed to get ID from miner address {:}",
+                    block.miner_address()
+                )))
+            }
+        };
```
austinabell, line 782: This is just reconstructuring the same address, can just check to make sure the protocol is correct, something like:
```
        if block.miner_address().protocol() != Protocol::ID {
            return Err(Error::Validation(format!(
                "failed to get ID from miner address {:}",
                block.miner_address()
            )));
        };
```
And can change the error message to be more fitting if you'd like

blockchain/state_manager/src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::StateManager;
+use actor::miner::State;
+use address::{Address, Payload};
+use blockstore::BlockStore;
+use cid::Cid;
+use fil_types::{RegisteredProof, SectorInfo, SectorSize};
+use filecoin_proofs_api::{post::generate_winning_post_sector_challenge, ProverId};
+use std::convert::TryInto;
+
+pub fn get_sectors_winning_for_winning_post<DB>(
```
austinabell, line 14:
```suggestion
pub fn get_sectors_for_winning_post<DB>(
```

blockchain/state_manager/src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+    state_manager: &StateManager<DB>,
+    st: &Cid,
+    address: &Address,
+    rand: &[u8; 32],
+) -> Result<Vec<SectorInfo>, Error>
+where
+    DB: BlockStore,
+{
+    let miner_actor_state: State = state_manager.load_actor_state(&address, &st)?;
```
austinabell, line 23: can you change the type to `miner::State` to be more clear?

blockchain/state_manager/src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+    let mut not_proving = miner_actor_state
+        .faults
+        .clone()
+        .merge(&miner_actor_state.recoveries)
+        .map_err(|_| Error::Other("Could not merge bitfield".to_string()))?;
+    let sector_set =
+        miner_actor_state.load_sector_infos(&*state_manager.get_block_store(), &mut not_proving)?;
```
austinabell, line 30: If matching lotus with this, might as well put it into it's own function so it can be reused later

blockchain/state_manager/src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+        SectorSize::_512MiB => RegisteredProof::StackedDRG512MiBSeal,
+        SectorSize::_32GiB => RegisteredProof::StackedDRG32GiBSeal,
+        SectorSize::_64GiB => RegisteredProof::StackedDRG64GiBSeal,
+    };
+    let wpt = seal_proof_type.registered_winning_post_proof()?;
+
+    let mid = match address.payload() {
+        Payload::ID(new_id) => Address::new_id(*new_id),
+        _ => return Err(Error::Other(format!("getting miner id {:}", address))),
+    };
```
austinabell, line 46: same with this as before, you don't need to reconstruct the address, they only get the uint from the id protocol for ffi interactions

blockchain/state_manager/src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+    let mut prover_id = ProverId::default();
+    let prover_bytes = mid.to_bytes();
+    prover_id[..prover_bytes.len()].copy_from_slice(&prover_bytes);
+    let ids = generate_winning_post_sector_challenge(
+        wpt.into(),
+        &rand,
+        sector_set.len().try_into().map_err(|_|Error::Other("Could not convert sector set length to u64 for generate_winning_post_sector_Challenge".to_string()))?,
```
austinabell, line 53:
```suggestion
        sector_set.len() as u64,
```
This can't fail?

vm/runtime/src/lib.rs
```diff
@@ -237,6 +237,46 @@ pub trait Syscalls {
...
+        let prover_bytes = verify_info.prover.to_be_bytes();
+        prover_id[..prover_bytes.len()].copy_from_slice(&prover_bytes);
+
+        // verify
+        if !verify_winning_post(&verify_info.randomness.0, &proofs, &replicas, prover_id)? {
+            return Err("Proof was invalid".to_string().into());
+        }
+
+        Ok(())
+    }
```
austinabell, line 279: This isn't a syscall, why is this being added here? What is this referencing?

AshantiMutinta: In lotus this methods seems to live in the sync struct

https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L754

blockchain/state_manager/src/utils.rs
AshantiMutinta: I was just being angsty about it basically thought it would not hurt

vm/runtime/src/lib.rs
austinabell: yeah, and you added that one in the PR, but this block here seems unnecessary?

blockchain/state_manager/src/utils.rs
austinabell: nah usize -> u64 can't fail, the other way around can fail on 32 bit targets (and there is that conversion below) but we don't support 32 bit so it doesn't matter

vm/runtime/src/lib.rs
AshantiMutinta: Oh I saw it in the otherlinks on the issues part and I assumed both needed to be implemented.

 I can remove it

https://github.com/ChainSafe/forest/issues/471

AshantiMutinta: Oh wait no this is a mistake on my part, was in the first stages of implementation and forgot to remove this.

austinabell: yeah, you added those two, but then this third function is unnecessary (there shouldn't be changes to the syscalls/runtime for these changes)

austinabell:
you accidentally committed a file in https://github.com/ChainSafe/forest/pull/493/commits/821f33b1ac518df145ae3bea0e809fcfab1c1ee2

PR #496, opened 2020-06-11
Updates market actor

### LINKED ISSUE #461: Update market actor
**Task summary**

Update the market actor to include `CronTick` and remove `HandleExpiredDeals`

**Specification reference**

- [Market actor](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go)

dutterbutter:
**Summary of changes**
- updates market actor to the following [commit](https://github.com/filecoin-project/specs-actors/commit/af71390e84cdc98cc588df81e3868703f889f7e3) 
- Updated set_multimap
- Added verified actor registry actor code 

Please advise on the best way to handle some of the errors in market_state, as you will see its not quite 1:1 at the moment since they are mixing err types.

Closes #461 


vm/actor/src/builtin/market/mod.rs
```diff
@@ -233,7 +228,7 @@ impl Actor {
         rt.transaction(|st: &mut State, rt| {
             let mut prop = Amt::load(&st.proposals, rt.store())
                 .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e.into()))?;
-            let mut dbp = SetMultimap::from_root(rt.store(), &st.deal_ids_by_party)
+            let mut dbp = SetMultimap::from_root(rt.store(), &st.deal_ops_by_epoch)
```
austinabell, line 231: what does dbp stand for?

vm/actor/src/builtin/market/mod.rs
```diff
@@ -476,6 +469,176 @@ impl Actor {
...
+
+            let mut i = st.last_cron + 1;
+            while i <= rt.curr_epoch() {
+                dbe.for_each(i, |id| {
+                    let mut state: DealState = states
+                        .get(id)
+                        .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e.into()))
+                        .unwrap()
+                        .ok_or(format!("could not find deal state: {}", id))
+                        .unwrap();
```
austinabell, line 510: shouldn't be unwrapping here, should just skip if not found (return `Ok(())` probably, but didn't check how fn is setup)

vm/actor/src/builtin/market/mod.rs
```diff
@@ -476,6 +469,176 @@ impl Actor {
...
+                        let slashed = st
+                            .process_deal_init_timed_out(
+                                rt.store(),
+                                &mut et,
+                                &mut lt,
+                                id,
+                                &deal,
+                                state,
+                            )
+                            .unwrap();
```
austinabell, line 529: again, don't unwrap, they return with a valid exit code if this fails (fn should return an ActorError and this should just try return

vm/actor/src/builtin/market/mod.rs
```diff
@@ -476,6 +469,176 @@ impl Actor {
...
+
+                        states
+                            .set(id, state)
+                            .map_err(|e| {
+                                ActorError::new(
+                                    ExitCode::ErrPlaceholder,
+                                    format!("failed to get deal: {}", e),
+                                )
+                            })
+                            .unwrap();
```
austinabell, line 564: You're mapping the error here, why unwrap lol? Should be returned

vm/actor/src/builtin/market/mod.rs
```diff
@@ -476,6 +469,176 @@ impl Actor {
...
+                        let idx = next_epoch.unwrap();
```
austinabell, line 565: can this just be moved above to not do the unwrap twice? (also optionally you could do an `if let` here no not have to do this

vm/actor/src/builtin/market/state.rs
```diff
@@ -42,8 +52,239 @@ impl State {
...
+        if epoch >= deal.end_epoch {
+            self.process_deal_expired(store, deal_id, &deal, state, lt)?;
+            return Ok((TokenAmount::zero(), OptionalEpoch(None)));
+        }
+
+        let next: ChainEpoch = if epoch + DEAL_UPDATED_INTERVAL > deal.end_epoch {
+            deal.end_epoch
+        } else {
+            epoch + DEAL_UPDATED_INTERVAL
+        };
```
austinabell, line 151: probably just cleaner to to a `std::cmp::min` here?

vm/actor/src/builtin/market/state.rs
```diff
@@ -42,8 +52,239 @@ impl State {
...
+
+        let elapsed_start = if ever_updated && state.last_updated_epoch.unwrap() > deal.start_epoch
+        {
+            state.last_updated_epoch.unwrap()
+        } else {
+            deal.start_epoch
+        };
+
+        let elapsed_end = if epoch < deal_end { epoch } else { deal_end };
```
austinabell, line 106: same here, could be a `std::cmp::min` to be more clear

vm/actor/src/builtin/market/state.rs
```diff
@@ -42,8 +52,239 @@ impl State {
...
+        let r_cid = states.flush().map_err(|e| {
+            ActorError::new(
+                ExitCode::ErrIllegalState,
+                format!("failed to load deal states: {}", e),
+            )
+        })?;
+
+        self.states = r_cid;
+        Ok(())
+    }
```
austinabell, line 175: Just take out (I think I saw they were removing soon anyway)

vm/actor/src/builtin/market/state.rs
```diff
@@ -534,6 +528,47 @@ impl State {
...
+    let buf = proposal
+        .proposal
+        .marshal_cbor()
+        .map_err(|e| e.to_string())?;
+    rt.syscalls()
+        .verify_signature(&proposal.client_signature, &proposal.proposal.client, &buf)
+        .map_err(|e| e.to_string())?;
+
+    Ok(())
+}
```
austinabell, line 570: just remove, no need for this (can always just look back to this to apply if logic is the same)

vm/actor/src/builtin/market/state.rs
```diff
@@ -493,8 +487,8 @@ impl State {
                 )
             })?)
     }
-
-    pub(super) fn must_get_deal_state<BS: BlockStore>(
+    // Unused but exists in spec
+    fn _must_get_deal_state<BS: BlockStore>(
```
austinabell, line 491: prob just best to delete this as well

vm/actor/src/builtin/market/state.rs
```diff
@@ -111,60 +362,51 @@ impl State {
         })
     }
 
-    pub(super) fn maybe_lock_balance<BS: BlockStore>(
+    fn maybe_lock_balance<BS: BlockStore>(
         &mut self,
         store: &BS,
         addr: &Address,
         amount: &TokenAmount,
     ) -> Result<(), ActorError> {
```
austinabell, line 370: This fn should now not return an actor error, since it might cause issues with the caller assuming the exit code is handled correctly (which it isn't here)

vm/actor/src/builtin/market/state.rs
```diff
@@ -111,60 +362,51 @@ impl State {
...
+                Ok(())
+            },
+        )
+        .map_err(|e| ActorError::new(ExitCode::ErrPlaceholder, e))?;
+
         Ok(())
     }
-
-    pub(super) fn unlock_balance<BS: BlockStore>(
+    /// TODO: all these balance table mutations need to happen at the top level and be batched (no flushing after each!)
```
austinabell, line 397: I think in general, we don't need to copy their TODOs, we just need to match their functionality so having the TODOs is a bit confusing because it indicates we have something more to do here

vm/actor/src/util/set_multimap.rs
```diff
@@ -33,33 +33,35 @@ where
...
         // Set hamt node to set new root
-        Ok(self.0.set(key.to_bytes().into(), &new_root)?)
+        Ok(self.0.set(u64_key(key), &new_root)?)
     }
 
     /// Gets the set at the given index of the `SetMultimap`
     #[inline]
-    pub fn get(&self, key: &Address) -> Result<Option<Set<'a, BS>>, String> {
-        match self.0.get(&key.to_bytes())? {
+    pub fn get(&self, key: BytesKey) -> Result<Option<Set<'a, BS>>, String> {
```
austinabell, line 53: This should probably be &ChainEpoch or ChainEpoch, you shouldn't have to expose the underlying type for the public interface for this

vm/actor/tests/set_multimap_test.rs
```diff
@@ -2,43 +2,43 @@
...
 
-    smm.put(&addr, 8).unwrap();
-    smm.put(&addr, 2).unwrap();
-    smm.remove(&addr, 2).unwrap();
+    smm.put(epoch, 8).unwrap();
+    smm.put(epoch, 2).unwrap();
+    smm.remove(epoch, 2).unwrap();
 
-    let set = smm.get(&addr).unwrap().unwrap();
+    let set = smm.get(u64_key(epoch)).unwrap().unwrap();
```
austinabell, line 19: indicating this is the reason why it shouldn't (should just be able to pass through the epoch)

vm/actor/src/builtin/market/state.rs
```diff
@@ -14,6 +17,12 @@ use num_traits::Zero;
 use runtime::Runtime;
 use vm::{ActorError, ExitCode, TokenAmount};
 
+// Market mutations
+// add / rm balance
+// pub deal (always provider)
+// activate deal (miner)
+// end deal (miner terminate, expire(no activation))
```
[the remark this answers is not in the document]
dutterbutter: > what's this doing?

I just copied it over from specs, will remove.

vm/actor/src/builtin/market/mod.rs
```diff
@@ -476,6 +469,176 @@ impl Actor {
...
+            while i <= rt.curr_epoch() {
+                dbe.for_each(i, |id| {
+                    let mut state: DealState = states
+                        .get(id)
+                        .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e.into()))
+                        .unwrap()
+                        .ok_or(format!("could not find deal state: {}", id))
+                        .unwrap();
+
+                    let deal = st.must_get_deal(rt.store(), id).unwrap();
```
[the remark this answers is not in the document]
dutterbutter: Aren't all those errors within the for each closure mapped to the actor error?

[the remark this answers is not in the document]
austinabell: not when you unwrap it for sure

vm/actor/src/builtin/market/state.rs
dutterbutter: As discussed, this will return an Actor Error due to the usage of `get_locked_balance` and `get_escrow_balance` returning an ActorError.

vm/actor/src/builtin/market/mod.rs
```diff
@@ -329,15 +310,18 @@ impl Actor {
...
-                    .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e.into()))?
-                    .ok_or_else(|| {
-                        ActorError::new(
-                            ExitCode::ErrIllegalState,
-                            "Failed to retrieve the DealState".to_owned(),
-                        )
-                    })?;
+                    .map_err(|e| ActorError::new(ExitCode::ErrIllegalState, e.into()))?;
+
+                if deal.is_some() {
```
AshantiMutinta, line 317: You can also do

deal.ok_or_else(|_|ActorError::new(ExitCode::ErrIllegalArgument,format!("given deal already included in another sector: {}", id)))?

vm/actor/src/util/set_multimap.rs
```diff
@@ -97,7 +101,8 @@ where
                 .map_err(|e| format!("Could not parse key: {:?}, ({})", &k.0, e))?;
 
             // Run function on all parsed keys
-            f(v)
+            f(v).map_err(|e| format!("Could parse all keys: ({})", e))
```
austinabell, line 104: This message doesn't make much sense to me, reason?

vm/actor/src/util/set_multimap.rs
```diff
@@ -97,7 +101,8 @@ where
                 .map_err(|e| format!("Could not parse key: {:?}, ({})", &k.0, e))?;
 
             // Run function on all parsed keys
-            f(v)
+            f(v).map_err(|e| e.to_string())
```
austinabell, line 104: This might be misleading, you are requiring the generic function error is ActorError, then converting to string. If that exit code doesn't matter/isn't used, then the generic function error should remain a string

dutterbutter: The error return type here is `String` as its within [set.rs for_each](https://github.com/ChainSafe/forest/blob/dustin/market-updates/vm/actor/src/util/set.rs#L61) function. I still want to return an ActorError for set_multimap for_each.

austinabell: So yeah there is those exit codes that are lost and converted to `ErrIllegalState`, this error cannot be lost in the conversion to string (I'll try to think of a clean way around this)

austinabell: I made some changes here https://github.com/ChainSafe/forest/commit/f5dd90c3d646d570ab7948dc6ae02286ffd3a552 To match go impl, I think the cleanest way is to downcast the error like this. I think it matches the exit codes 1:1 but please verify. 
Keep in mind when they call rt.Abort with an exit code, we need that code to be kept and eventually returned. Please ask any questions about this or offer alternatives, I don't want this to silently go in, but feel free to cherry pick the commit if you agree

dutterbutter: Made requested changes but as discussed removed the actor error from set_multimap.

PR #500, opened 2020-06-15
Fix ecrecover and verify methods

### LINKED ISSUE #498: Fix ecrecover method in signature.rs
**Task summary**
Currently, the ecrecover method in signature.rs is panicking due to incorrect code whenever this function is called. This issue will fix this problem as well as update the logic to correctly get an Address given a message hash and Signature.

**Specification reference**
Refer to go-secp256k1: https://github.com/ipsn/go-secp256k1/blob/9d62b9f0bc52d16160f79bfb84b2bbf0f6276b03/secp256.go#L93
as well as Zondax: https://github.com/Zondax/filecoin-rs/blob/842f61de3c6c3d5d90c5b2ada9d1e342331fe23f/signer/src/lib.rs#L315

flodesi:
**Summary of changes**
- Fix ecrecover method to properly return the required address
- Update wallet signing for secp256 to return the proper signature including required recovery byte
- Move bls_verify method to wallet to take advantage of wallet signing
- Add secp_verify test to test secp verification as well

Closes 
#498

### DIFF
--- a/crypto/src/signature.rs
+++ b/crypto/src/signature.rs
@@ -48,8 +48,8 @@ pub struct Signature {
 
 impl ser::Serialize for Signature {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-        where
-            S: ser::Serializer,
+    where
+        S: ser::Serializer,
     {
         let mut bytes = self.bytes.clone();
         // Insert signature type byte
@@ -61,8 +61,8 @@ impl ser::Serialize for Signature {
 
 impl<'de> de::Deserialize<'de> for Signature {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-        where
-            D: de::Deserializer<'de>,
+    where
+        D: de::Deserializer<'de>,
     {
         let mut bytes: Vec<u8> = serde_bytes::Deserialize::deserialize(deserializer)?;
         if bytes.is_empty() {
@@ -148,8 +148,6 @@ impl Signature {
         if &rec_addr == addr {
             Ok(())
         } else {
-            println!("{:?}", rec_addr);
-            println!("{:?}", addr);
             Err("Secp signature verification failed".to_owned())
         }
     }
@@ -180,11 +178,10 @@ pub fn verify_bls_aggregate(data: &[&[u8]], pub_keys: &[&[u8]], aggregate_sig: &
     verify(&sig, &hashed_data[..], &pks[..])
 }
 
-// TODO: verify signature data format after signing implemented
+/// Return Address for a message given it's hash and signature
 pub fn ecrecover(hash: &[u8; 32], signature: &[u8; 65]) -> Result<Address, Error> {
     // generate types to recover key from
     let rec_id = RecoveryId::parse(signature[64])?;
-    // let rec_id = RecoveryId::parse(2)?;
     let message = Message::parse(&hash);
 
     // Signature value without recovery byte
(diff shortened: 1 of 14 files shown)


crypto/src/signature.rs
```diff
@@ -204,36 +199,10 @@ fn ecrecover(hash: &[u8; 32], signature: &[u8; 65]) -> Result<Address, Error> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use address::Address;
     use bls_signatures::{PrivateKey, Serialize, Signature as BlsSignature};
     use rand::rngs::mock::StepRng;
     use rand::Rng;
 
-    #[test]
```
ec2, line 212: Why was this test removed?

flodesi: it was moved to `wallet.rs` to make use of using the wallet to sign

key_management/src/wallet.rs
```diff
@@ -336,4 +338,28 @@ mod tests {
...
+        let msg = [0u8; 64];
+
+        let sig = wallet.sign(&addr, &msg).unwrap();
+        sig.verify(&msg, &addr).unwrap();
+    }
+
+    #[test]
+    fn bls_verify_test() {
+        let key_vec = construct_priv_keys();
+        let addr = key_vec[6].address.clone();
```
austinabell, line 357: instead of just indexing that vector, can you just generate each type explicitly, so that if construct_priv_keys changes for any reason, this doesn't break or just unknowingly lose coverage? Other option is just to assert the address protocol is Secp and BLS respectively?

Would be nice to also test the case of invalid address verify fails, as in try to verify with some other random address

key_management/src/wallet_helpers.rs
```diff
@@ -53,11 +54,13 @@ pub fn sign(sig_type: SignatureType, private_key: &[u8], msg: &[u8]) -> Result<S
         SignatureType::Secp256 => {
```
austinabell, line 54: nit but I didn't notice this in previous review, can you change this:
```suggestion
        SignatureType::Secp256k1 => {
```

key_management/src/wallet.rs
```diff
@@ -353,13 +355,52 @@ mod tests {
...
+        let invalid_bls_priv = generate(SignatureType::BLS).unwrap();
+        let invalid_bls_key_info = KeyInfo::new(SignatureType::BLS, invalid_bls_priv);
+        let invalid_bls_key = Key::try_from(invalid_bls_key_info).unwrap();
+        let invalid_bls_addr = invalid_bls_key.address;
+
+        // construct wallet and sign 2 messages one with secp addr and one iwth bls addr
+        let mut wallet = Wallet::new_from_keys(MemKeyStore::new(), vec![secp_key, bls_key]);
+        let msg = [0u8; 64];
+        let secp_sig = wallet.sign(&secp_addr, &msg).unwrap();
+        let bls_sig = wallet.sign(&bls_addr, &msg).unwrap();
```
austinabell, line 400: Shouldn't you just use the `generate_key` function that exists to avoid the duplication? (seems like the generate function is a bit ambiguous, could maybe be something like `generate_raw_key` or something?)

(side note you shouldn't have to do this setup all for invalid signing, could have just added a verify against an invalid address at the end of the bls and secp tests)

key_management/src/wallet.rs
```diff
@@ -353,13 +355,52 @@ mod tests {
...
+
+        // assert verification failure with invalid address
+        assert!(secp_sig.verify(&msg, &invalid_secp_addr).is_err());
+        assert!(bls_sig.verify(&msg, &invalid_bls_addr).is_err());
```
austinabell, line 404: You didn't need to create invalid keys just to verify the address is invalid, should just be able to verify against the address, or is that somehow simpler to do this way?

PR #501, opened 2020-06-16
Implement Kademlia discovery

austinabell:
**Summary of changes**
- Implements Kademlia bootstrapping and discovery
  - May be some tweaks in future to make sure it matches as closely as possible with go impls but there isn't a strict need to
- Refactors libp2p crate in various places to allow for cleaner interaction
  - Updates logging levels to be more readable by default

Closes 


ec2:
I don't see where you call `get_closest_peers` periodically. Isn't that how you discover new peers through kad?

austinabell:
> I don't see where you call `get_closest_peers` periodically. Isn't that how you discover new peers through kad?

Yeah, I mean, the way it is already bootstraps finding new peers (gets up to ~50 in 10 seconds on the testnet, should probably also look into limiting this) but this was what I was alluding to making tweaks to match behaviour. I can be more specific and detail an issue to come back to after this PR comes in, or if you have an idea of how this can be periodically polled cleanly I'm open to hearing it.

The reason I went without the periodic polling of closest peers is because it didn't seem clear that was necessary or even how go impls have it (from what I could see) also because that logic was being refactored I didn't want to cause unnecessary conflicts. Also the other thing was it wasn't clear in that periodic poll what peer id would be provided to get closest peers to, so didn't want to introduce technical debt by guessing.

blockchain/chain_sync/src/sync.rs
```diff
@@ -982,6 +971,18 @@ where
     pub fn set_state(&mut self, new_state: SyncState) {
         self.state = new_state
     }
+
+    async fn get_peer(&self) -> PeerId {
+        while self.peer_manager.is_empty().await {
```
AshantiMutinta, line 976: If this can't find a peer, it will loop indefinitely, is it worth it to have a time out mechanism?

AshantiMutinta: Or maybe implement a try_get_peer() that takes in a timeout variable as well?

austinabell: Somewhat intentional, because it can't do anything else until it has a peer to sync with. The logic hasn't changed, I just put in a function because there was another place it should be used. 

The logic is definitely a bit broken now after the logic had been refactored since this was setup, but I don't want to cause conflicts with the work Erlich is doing in a bit. This shouldn't be an issue once the syncing setup is improved, but maybe something to keep in mind @ec2

austinabell: But yeah, definitely agree a timeout is worthwhile to have, just probably after the setup is refactored (unless you guys think otherwise)

node/forest_libp2p/src/service.rs
```diff
@@ -64,37 +67,26 @@ pub struct Libp2pService {
...
-                    Ok(_) => debug!("Dialed {:?}", node),
-                    Err(e) => warn!("Dial {:?} failed: {:?}", node, e),
-                },
-                Err(err) => warn!("Failed to parse address to dial: {:?}", err),
-            }
+        Swarm::listen_on(&mut swarm, config.listening_multiaddr).unwrap();
+
+        // Subscribe to gossipsub topics with the network name suffix
+        for topic in PUBSUB_TOPICS.iter() {
+            swarm.subscribe(Topic::new(format!("{}/{}", topic, network_name)));
```
AshantiMutinta, line 84: Not needed but could also be a one liner by changing to 

PUBSUB_TOPICS.iter().for_each(|topic|swarm.subscribe(Topic::new(format!("{}/{}", topic, network_name))));

blockchain/chain_sync/src/sync.rs
AshantiMutinta: That is totally fine to me!

node/forest_libp2p/src/service.rs
austinabell: ends up being the same because the subscribe returns a boolean for if it was successful (ignoring because it should always be)

PR #503, opened 2020-06-16
Chain syncing verification fixes

ec2:
**Summary of changes**
- Fixes block verification and response handling

Closes 


blockchain/chain_sync/src/sync.rs
```diff
@@ -168,27 +170,40 @@ where
...
                                 heaviest_tipset_weight: heaviest.weight().clone(),
-                                genesis_hash: self.genesis.blocks()[0].cid().clone(),
+                                genesis_hash: chain_syncer.genesis.blocks()[0].cid().clone(),
                             },
                         )
                         .await
                 }
+                NetworkEvent::PubsubMessage { source, topics, message } => {
+                    match topics[0].as_str() {
+                       "/fil/blocks/interop"  => {
```
austinabell, line 188: the topic depends on the network name from genesis (in case that helps debugging)

blockchain/chain_sync/src/sync.rs
```diff
@@ -168,27 +170,40 @@ where
...
+                           let new_blk: BlockHeader = BlockHeader::unmarshal_cbor(&message).unwrap();
+                           let new_blk = Tipset::new(vec![new_blk]).unwrap();
+                           let chain_syncer = chain_syncer.clone();
+                           task::spawn(async move {
+                               chain_syncer.sync(&new_blk).await.unwrap();
+                           });
```
austinabell, line 194: I really don't think this is what we want to do, we shouldn't be manually constructing the tipset and doing a full sync from it, can you explain reasoning?

ec2: Hmm, I think we actually are supposed to call inform_new_head here instead. Does that answer your question?

blockchain/chain_sync/src/sync.rs
ec2: Oh, yeah. I will polish this more later. Just wanted to make it clear that we call sync when we receive something over gossip.

blockchain/chain_sync/src/sync.rs
austinabell: yeah, that makes a lot more sense, calling sync I'm pretty sure isn't what we want

blockchain/chain_sync/src/sync.rs
```diff
@@ -839,8 +869,9 @@ where
             }
 
             // TODO tweak request window when socket frame is tested
-            const REQUEST_WINDOW: i64 = 5;
+            const REQUEST_WINDOW: i64 = 30;
             let epoch_diff = cur_ts.epoch() - to_epoch;
+            info!("BlockSync from: {} to {}", cur_ts.epoch(), to_epoch);
```
austinabell, line 874: can you change this from an info to debug? Don't think this should be logged by default

austinabell (commented):
LGTM pending functional testing

blockchain/chain_sync/src/sync.rs
```diff
@@ -229,7 +229,7 @@ where
             self.state.write().await.error(e.to_string());
             return Err(e.into());
         }
-
+        ChainStore::put_tipset(&mut self.chain_store, &*head.clone()).await?;
```
austinabell, line 232: isn't this stored otherwise? even if not can you remove the clone? shouldn't need to serialize. maybe `head.as_ref()` is what you want

ec2: Its not stored otherwise. The bundle of headers above does not include this tipset. put_tipsets also sets the head for the tracker too.

austinabell: it's stored on line 242? Even if you keep, you should not clone the head just to serialize (it is an Arc, but unnecessary clone when you can just take a ref)

ec2: ah youre right actually, this should be removed

ec2: fixed

austinabell (approved):

blockchain/blocks/src/header/mod.rs
```diff
@@ -418,6 +422,11 @@ impl BlockHeader {
         }
         Ok(())
     }
+    pub fn to_signing_bytes(&self) -> Result<Vec<u8>, String> {
```
dutterbutter, line 425: Nit: Add comment

blockchain/chain_sync/src/sync.rs
```diff
@@ -247,7 +246,7 @@ where
     /// Syncs messages by first checking state for message existence otherwise fetches messages from blocksync
     async fn sync_messages_check_state(&mut self, ts: &[Tipset]) -> Result<(), Error> {
         // see https://github.com/filecoin-project/lotus/blob/master/build/params_shared.go#L109 for request window size
-        const REQUEST_WINDOW: i64 = 200;
+        const REQUEST_WINDOW: i64 = 1;
```
dutterbutter, line 249: Nit: Add TODO as a reminder to change this back to 200 as its the `defaultMessageFetchWindowSize` in Lotus.

dutterbutter (approved):
Just those two comments

blockchain/chain_sync/src/sync.rs
austinabell: intentionally not followed for now, doesn't make sense why this request would be bulked like this since the transition has to be done sequentially. Can benchmark on our own client if batching the requests helps throughput, but this is not a protocol value

PR #506, opened 2020-06-17
Bitfield improvements

### LINKED ISSUE #468: BitField optimization and benchmarking
**Issue summary**

So a bitfield implementation is coming in with https://github.com/ChainSafe/forest/pull/466 to setup the interface needed but can and should probably be optimized.

To summarize the go implementation, their structure is decoded RLE+ bytes with a get and set hashset (it's a map but performs as a set) cache. What they do is iterate over the RLE+ bits as runs of set or unset bits (look into RLE encoding about this) for gets when not found in the cache. This does get a bit more complex when using any bit operator because there needs to be iterators for each operation applied to the RLE+ encoded bits.

How the current implementation differs is that for read and write operations, the bitfield is decoded which has the upfront cost of decoding, but then all operations are fast but also requires more memory to keep the decoded bitvec in memory until encoded again. 

It isn't clear the tradeoffs between the two (especially in a realistic usage) which is why benchmarks should be setup.

To point out the functions to focus on:

`new_from_set` - If the BitField type is changed to not be an enum, the RLE encoded bitvec should be optional (not really a point as is) but this functionality is inefficient currently because it is generating and allocating a decoded bitvector with the set indices.

`get` -  The inefficient part of this function is that on encoded and when not found in cache, the BitField is flushed and decoded, when the RLE bits can just be traversed to check if exists

`first` - This function is an exaggeration of get, as it is just flushing to get the first set bit (multiple ways of going about this being traversing the rle bits and ordering the two cache sets or doing similar to go impl with combining iterators and generating bitfield with cache and using OR operator)

Other functions - Most other functions aren't specifically that bad, but do fall into the tradeoff of using more memory and using CPU in different places (but less computation in repeated operations) and should be benchmarked. Hard to guess the scale of the issue this early and probably still will be hard to setup accurate benchmarks

Ideally would be nice to take into account the usage of BitField in specs-actors for setting up the benchmarks. As I see it, seems to be two main options:
1. Setup an RLE+ bits iterator, which will allow it to be traversed for at least `get` and `first` optimizations
2. Setup similar to go impl (or open to other options) with using iterators for each bit operation and refactor all usages

https://github.com/filecoin-project/go-bitfield
https://github.com/filecoin-project/specs-actors/blob/master/actors/abi/bitfield.go

timvermeulen:
**Summary of changes**

- `BitField` now always contains an encoded bitvec and never a decoded one, which means that most methods no longer need a `&mut self` reference
- `RLEPlus` is an RLE+ encoded `BitVec` that has been verified to be encoded correctly
- `BitField` and `RLEPlus` work with `usize` rather than `u64` in order to minimize numeric conversions
- `AHashMap` replaces `FnvHashMap` due to speedups of around 30%-50% according to pretty rudimentary benchmarks
- the `RangeIterator` trait was added for working with lazy sequences of disjoint ranges
  - the RLE+ encoding works with runs of 1s and 0s, but working with ranges of 1s instead makes implementing several algorithms a lot less error-prone
  - consequently, a (meaningless) trailing run of 0s is no longer encoded, so a few tests needed minor changes to not use a trailing run of 0s
  - any `RangeIterator` is still a regular `Iterator<Item = Range<usize>>` so you can still use them in `for` loops and stuff like that
- `BitField` borrows some behavior from `HashMap`: `union`, `intersection`, and `difference` return (lazy) `RangeIterator`s, and the `|`, `&`, and `-` operators are their eager counterparts that return `BitField`s
  - the existing `union` for taking the union of any number of `BitField`s has been renamed `multi_union`
  - unlike in the Go implementation, these combinators are completely lazy and are linear over the number of ranges, not on the number of bits
- a `BitField` can be iterated with `ranges()` as well as with `iter()` which returns an iterator over the set bits of the bit field
  - the private `retrieve_set_indices` was therefore renamed `bounded_iter(max: usize)` and was made public, it now also returns an iterator instead of a `B: FromIterator`
  - we no longer need `BitField::for_each` which can be directly replaced with `.iter().try_for_each(...)`
- `BitReader` and `BitWriter` are used to efficiently read and write multiple bits at a time (up to 8) which significantly speeds up RLE+ encoding and decoding

I think I haven't forgotten anything!

**Reference issue to close**

Closes #468


utils/bitfield/src/rleplus/reader.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        Err("Invalid varint")
+    }
+
+    /// Reads a length from the buffer according to RLE+ encoding.
+    pub fn read_len(&mut self) -> Result<Option<usize>, &'static str> {
+        let len = if self.read(1) == 1 {
+            // Block Single (prefix 1)
+            Some(1)
+        } else if self.read(1) == 1 {
```
ec2, line 101: You if and else if conditional is the same

timvermeulen: Yeah it looks a little weird, but the second `read` reads the bit after that.
The `if`/`else if`/`else` here reads the prefix which is either `1`, `01`, or `00`

timvermeulen: I'll put them in separate variables, that will make it clearer

ec2: I figured that the intention based on the comments, but i still dont see how it will EVER reach the else if since the conditional is the same as the if

timvermeulen: `self.read` advances `self`, so calling `self.read(1)` twice in a row won't read the same bit, it will read different bits — and the `self.read(1)` in the `else` will obviously only be called after the first `self.read(1)` has returned `0`

ec2: Yeah, I get that. Ok, maybe I'm totally not understanding something here...

ec2: OOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOK Wow, i see whats happening

vm/actor/src/builtin/miner/state.rs
```diff
@@ -505,33 +500,30 @@ impl State {
...
+        if non_faults.is_empty() {
             return Err(format!(
                 "failed to check if bitfield was empty: {:?}",
                 non_faults
             ));
         }
 
         // Select a non-faulty sector as a substitute for faulty ones.
-        let good_sector_no = non_faults.first()?;
+        let good_sector_no = non_faults.first().unwrap();
```
austinabell, line 519: can you swap this to an `ok_or` so if none it returns an error instead of panicing?

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -382,43 +371,36 @@ impl Actor {
...
+                        sectors_by_number.insert(sec.info.sector_number, sec);
                     }
-                    Err(e) => Err(ActorError::new(
-                        ExitCode::ErrIllegalState,
-                        format!("failed to check if bitfield was empty: {}", e),
-                    )),
+                    declared_recoveries.iter().for_each(|i| {
+                        let key = i as u64;
+                        let s = sectors_by_number.get(&key).cloned().unwrap();
+                        recovered_sectors.push(s);
```
austinabell, line 401: Sorry, unrelated to your changes, but can you add a TODO here to handle deduplication of values? 

I didn't notice before but it seems possible that there could be inconsistencies because the go impl puts in a hashmap before iterating and pushing to recovered_sectors. Feel free to make change here yourself if you want btw

utils/bitfield/src/lib.rs
```diff
@@ -1,462 +1,247 @@
...
         }
-        Ok(())
     }
 
-    /// Subtract other bitfield from self (equivalent of `a & !b`)
-    pub fn subtract(mut self, other: &Self) -> Result<Self> {
-        self.subtract_assign(other)?;
-        Ok(self)
+    /// Returns the number of set bits in the bit field.
+    pub fn len(&self) -> usize {
```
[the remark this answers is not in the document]
timvermeulen: Yeah I feel you, I have struggled with this in the documentation. If you look at a bit field as a sequence of 1s and 0s then `len` here is obviously incorrect and `count_ones` would make more sense. But if you look at it as a collection of indices that represent where the 1s are (like a more efficient and ordered `HashSet<usize>`), then `len` as the number of 1s makes sense again.

I think we should probably go with the "ordered collection of indices" definition because then it makes sense that:
- `first` returns an index and not the value of the first bit,
- `iter` iterates over the indices of the 1s and not over a bunch of `bool`s,
- `ranges` only cares about 1s,
- etc

Also `union`/`intersection`/`difference` treat it as a collection of 1s.

The `len`/`count` distinction is so subtle that I don't really care either way, to me `count` doesn't convey that it only cares about the set bits more than `len` does? And common Rust wisdom seems to dictate that `count` consumes the thing (like `Iterator::count`) while `len` doesn't, so that's what I based it on.

Whatever we go for, docs couldn't definitely improve in this regard.

utils/bitfield/src/lib.rs
```diff
@@ -1,462 +1,247 @@
...
-    pub fn union<'a>(bit_fields: impl IntoIterator<Item = &'a Self>) -> Result<Self> {
-        let mut ret = Self::default();
-        for bf in bit_fields.into_iter() {
-            ret.merge_assign(bf)?;
-        }
-        Ok(ret)
+    /// Returns a new `RangeIterator` over the bits that are in `self`, in `other`, or in both.
+    ///
+    /// The `|` operator is the eager version of this.
+    pub fn union<'a>(&'a self, other: &'a Self) -> impl RangeIterator + 'a {
```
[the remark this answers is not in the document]
timvermeulen: Yeah valid, I guess my concern was that `merge` wasn't specific about _how_ the two sets were merged, but after Googling a bit I don't think there can be any ambiguity about it

vm/actor/src/builtin/miner/state.rs
timvermeulen: ...oops

utils/bitfield/src/lib.rs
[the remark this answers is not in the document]
austinabell: yeah, that's all reasonable. I'm fine with this change, so maybe see if anyone else has a preference as I have nothing else I care to add to my previous comment

vm/actor/src/builtin/miner/state.rs
timvermeulen: Oh hang on, there's an emptiness check right above. I guess I'll combine the two

austinabell: ah right, that wasn't in scope from what I was looking at. Yeah that sounds preferable anyway

utils/bitfield/src/lib.rs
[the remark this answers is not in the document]
ec2: I agree with @timvermeulen here. Thinking about this as an "ordered collection of indices" makes sense for `len` insofar as it is consistent with `first`, `iter`, etc... Since the other publicly exposed functions dont really make reference to the underlying RLE encoded bits, there shouldn't be confusion there.

vm/actor/src/builtin/miner/state.rs
timvermeulen: I'm now using the error message that Dustin gave me on Slack in case you're wondering where I got it from

utils/bitfield/src/lib.rs
```diff
@@ -1,462 +1,247 @@
...
 
-    /// Retrieves number of set bits in the bitfield
-    ///
-    /// This function requires a mutable reference for now to be able to handle the cached
-    /// changes in the case of an RLE encoded bitfield.
-    pub fn count(&mut self) -> Result<usize> {
-        Ok(self.as_mut_flushed()?.count_ones())
+    /// Returns the index of the lowest bit present in the bit field.
+    pub fn first(&self) -> Option<usize> {
+        self.iter().next()
```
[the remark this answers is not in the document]
timvermeulen: `flatten` is lazy, so I think both versions will compile the exact same way. There might still be a better way to implement this (and `bits` in general) though — `ranges` has to fully go through the `set` and `unset` bits up front, but `bits` could get away with only going through `set` up front and check `unset` containment lazily... I have to think about this more

vm/actor/src/builtin/miner/mod.rs
timvermeulen: What would the keys of that hash map be? You already can't have two sectors for the same `key` in the code you commented on because they're pulled from a hash map, but I don't know what the Go code looks like so I might be missing the point

austinabell: oops, my bad. I misread skimming quickly and comparing with specs-actors. Ignore that no need to add the TODO

utils/bitfield/src/lib.rs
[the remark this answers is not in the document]
austinabell: yeah fine to keep this way, I was just more double checking, seemed like there was a bit of overlapping logic because `ranges()` handles the set and unset bits, but iter also merges with the set bits. I would think the compiler would handle ignoring this well, but I'm wondering why `iter` merges with set bits and filters unset bits if handled already

timvermeulen:
For future reference, Austin's concerns have been resolved — this is now consistent with the Go implementation

timvermeulen:
@austinabell @ec2 I pushed some minor changes to deal with integer overflow in the same way the Go implementation does for which I had to drop down from ranges to runs because a trailing 0 run still shouldn't overflow 64 bits, which is why `DecodedRanges` was replaced by `Runs`. This PR should be good to go now.

I'll make a separate PR for removing the `bitvec` dependency.

PR #511, opened 2020-06-22
A bare-bones GraphSync ResponseManager

timvermeulen:
**Summary of changes**

Adds a very bare-bones `ResponseManager` that manages `PeerResponseSender`s and responds to graphsync requests.

Basically a proof of concept because this setup doesn't allow interrupting graphsync requests, which is essential.


austinabell:
> Looks good. Noticed the clippy failure has an issue created [here](https://github.com/filecoin-project/rust-fil-proofs/issues/1183). Will hold off on approving until that gets fixed

Should be good now once branch is updated

PR #516, opened 2020-06-23
Refactor CLI and implement fetch-params

### LINKED ISSUE #508: Proof parameters fetch
**Task summary**

Although we don't need proof generation params, we do need the verification params so this logic needs to be built out.

High level is that the params get pulled from ipfs and stored in local fs at a static location, and the fil proofs pull them from this file (or where the env variable `FIL_PROOFS_PARAMETER_CACHE` points to) to use in verification

A CLI command will need to be added to do this

**Specification reference**

- go: https://github.com/filecoin-project/go-paramfetch/blob/master/paramfetch.go
- rust (can't use as is so use go as primary reference): https://github.com/filecoin-project/rust-fil-proofs/blob/master/filecoin-proofs/src/bin/paramfetch.rs
- proofs usage: https://github.com/filecoin-project/rust-fil-proofs/blob/28ebe446aafddef4334c1465168bb04c7a91b4fe/storage-proofs/core/src/parameter_cache.rs

austinabell:
**Summary of changes**
- Refactors CLI to setup framework for subcommands
  - Now `forest` still runs the daemon, but subcommands can be run as `forest <cmd>` such as here is `forest fetch-params ..`

- `fetch-params` command implemented which will fetch params based on the options of command and put into the params cache to be used in proof verification
  - Options are self-documenting by adding the help flag when running the command
  - Verbose flag will use progress bars for the download progress (doesn't always show up cleanly because async logger and the progress bar std out don't always play nicely

The params fetching is all done in parallel and async

Closes #508 


forest/src/paramfetch/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        tasks.push(task::spawn(async move {
+            if let Err(e) = fetch_verify_params(&name, &info, cmb).await {
+                warn!("Error in validating params {}", e);
+            }
+        }));
+    }
+
+    if let Some(multi_bar) = mb {
+        let cmb = multi_bar.clone();
+        let (mb_send, mut mb_rx) = futures::channel::oneshot::channel();
```
AshantiMutinta, line 92: just curious why we are using this channel vs async-std channel?

forest/src/daemon.rs
```diff
@@ -0,0 +1,93 @@
...
+                &gen_keypair.encode(),
+                &format!("{}{}", &config.data_dir, "/libp2p/"),
+                "keypair",
+            ) {
+                info!("Could not write keystore to disk!");
+                trace!("Error {:?}", e);
+            };
+            Keypair::Ed25519(gen_keypair)
+        }
+    };
```
AshantiMutinta, line 37: Could also use an unwrap_or_else(||{}) would remove the need to match that Some(kp)

forest/src/paramfetch/mod.rs
austinabell: yeah no real reason other than just quicker to use oneshot channel for what I needed, I'll switch though (don't think it really matters though)

forest/src/paramfetch/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+    let params: ParameterMap = serde_json::from_str(param_json)?;
+    let mut tasks = Vec::with_capacity(params.len());
+
+    let mb = if is_verbose {
+        Some(Arc::new(MultiBar::new()))
+    } else {
+        None
+    };
+
+    for (name, info) in params {
```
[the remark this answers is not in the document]
austinabell: I like, going to make a few small changes though

forest/src/cli/mod.rs
```diff
@@ -16,22 +16,51 @@ use std::sync::Arc;
...
+        params_size: Option<String>,
+        #[structopt(short, long, help = "Show verbose logging")]
+        verbose: bool,
+    },
+}
+
+/// Daemon process command line options.
+#[derive(StructOpt, Debug)]
+pub struct DaemonOpts {
     #[structopt(short, long, help = "A toml file containing relevant configurations.")]
```
ec2, line 56: Nit: Can we end/not end in periods consistently

austinabell:
> LGTM. I hear reqwest is pretty heavy. It is recommended to use Surf. Surf has a lot of overlapping libs with Tide since they're built by the same people. Up to you if you think its worth the change though

Yeah thing was is that it seemed they were still on alpha for using stable futures so didn't seem worth the trial and error, I will play around with it now for a bit though

forest/src/paramfetch/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+
+fn check_file(path: &Path, info: &ParameterData) -> Result<(), io::Error> {
+    if std::env::var(TRUST_PARAMS_ENV) == Ok("1".to_owned()) {
+        warn!("Assuming parameter files are okay. DO NOT USE IN PRODUCTION");
+        return Ok(());
+    }
+
+    let mut file = SyncFile::open(path)?;
+    let mut hasher = Blake2b::new();
+    sync_copy(&mut file, &mut hasher)?;
```
austinabell, line 242: If someone has a good way to read from async buffer and write to sync buffer let me know lol (blake2b hasher doesn't have async write)

austinabell:
I made a bunch of changes so just going to assume reviews invalid @AshantiMutinta @ec2 (sorry they were pretty necessary changes)

PR #517, opened 2020-06-23
Remove dead code and update CI

austinabell:
**Summary of changes**
- iirc was just code added when we first started and the blockchain and node crates were setup under the assumption they would be binaries run independently
- Updates CI to use lockfile for caching to reduce duplicate builds

Closes 


blockchain/src/main.rs
```diff
@@ -1,4 +0,0 @@
-// Copyright 2020 ChainSafe Systems
-// SPDX-License-Identifier: Apache-2.0, MIT
-
-fn main() {}
```
ec2, line 4: i cant believe we kept this for so long haha....

blockchain/src/message_pool/pool.rs
```diff
@@ -1,71 +0,0 @@
-// Copyright 2020 ChainSafe Systems
```
ec2, line 1: Might make sense to not touch the message_pool stuff, since @flodesi is working on it. Thoughts?

austinabell: he didn't touch any of this code, which is why I remembered to remove this now. I mean that stuff was loosely ported over from what the spec was when we started (I assume) so probably should be relied on anyway

flodesi: Yeah, I'm not using those files, so it's good to be removed.

PR #518, opened 2020-06-24
Bitswap Integration

ec2:
**Summary of changes**
- Added the Bitswap libp2p behaviour
Bitswap is the protocol used to do IPLD data exchange. In this case (will be handled in next PR) Bitswap is needed for when we recieve a BlockHeader over Gossipsub. Bitswap is used to fetch messages (by Cid) in the event we don't have the message already (which is likely). 
The general flow is: Call the swarm to broadcast a WantList of Cids we want. Every relevant task will have a flo_stream::Subscriber that is awaited on. We keep awaiting and matching until we get a NetworkEvent::BitswapBlock(cid) that we are looking for. **Not all this is implemented yet. Only the setup of the architecture is built here.**

- Added support for concurrent Libp2p RPC message handling
Libp2p RPC was originally written in such a way that we assume only one thread/task will send/recieve a request/response. This will not be the case in the next PR where we sync multiple blocks at the same time. 
The way this is "solved" is by keeping a mapping between request_id and channel senders. When a request is made, a request_id -> Sender is stored in a hashmap, and the Receiver is returned immediately, which the sending method can await on. When the network receives the response corresponding to the request via a request_id, the result will be sent to the channel. 

Closes 


blockchain/chain_sync/src/network_handler.rs
```diff
@@ -3,47 +3,51 @@
...
+        let request_table = self.request_table.clone();
 
         task::spawn(async move {
             loop {
                 match receiver.next().await {
                     // Handle specifically RPC responses and send to that channel
                     Some(NetworkEvent::RPCResponse { req_id, response }) => {
-                        rpc_send.send((req_id, response)).await
+                        // look up the request_table for the id and send through channel
+                        let tx = request_table.lock().await.remove(&req_id).unwrap();
```
austinabell, line 49: this isn't impervious to failing, probably safe to not unwrap here. The hashmap may not be purged now, but probably should in future to handle requests that error or timeout

blockchain/chain_sync/src/network_handler.rs
```diff
@@ -3,47 +3,51 @@
...
+                        tx.send(response).unwrap();
```
austinabell, line 50: kinda same here, not exactly familiar when the oneshot channel can fail, but I'm thinking in this case it is probably better for this to be an async-std channel because the responses are potentially very large and would be better for it to be async maybe?

blockchain/chain_sync/src/sync.rs
```diff
@@ -1103,9 +1103,12 @@ mod tests {
         task::block_on(async move {
             cs.peer_manager.add_peer(source.clone(), None).await;
             assert_eq!(cs.peer_manager.len().await, 1);
-
-            let return_set = cs.sync_headers_reverse(head, &to).await;
-            assert_eq!(return_set.unwrap().len(), 4);
+            // make blocksync request
+            let return_set = task::spawn(async move { cs.sync_headers_reverse(head, &to).await });
+            task::sleep(Duration::from_secs(2)).await;
```
austinabell, line 1108: Is the reason for the thread sleep because the message published does not get queued and can only be read when polling? Just trying to understand the potential lost messages

node/forest_libp2p/src/behaviour.rs
```diff
@@ -82,6 +92,29 @@ impl NetworkBehaviourEventProcess<KademliaEvent> for ForestBehaviour {
     }
 }
 
+impl NetworkBehaviourEventProcess<BitswapEvent> for ForestBehaviour {
+    fn inject_event(&mut self, event: BitswapEvent) {
+        match event {
+            BitswapEvent::ReceivedBlock(peer_id, cid, data) => {
+                let cid = cid.to_string();
+                let cid: Cid = Cid::from_str(&cid).unwrap();
```
austinabell, line 100: bytes conversion is probably sufficient. The conversion goes `Cid -> bytes -> string -> bytes -> Cid` so can just cut out turning the bytes to string (which just does with multibase)

Probably safe to handle error also (probably just warning about the error and returning) because there could be cases with inconsistencies. iirc they dropped support for v0 Cids which could be one potential dos vector for someone to crash this

node/forest_libp2p/src/behaviour.rs
```diff
@@ -82,6 +92,29 @@ impl NetworkBehaviourEventProcess<KademliaEvent> for ForestBehaviour {
...
+                self.events.push(ForestBehaviourEvent::BitswapReceivedBlock(
+                    peer_id, cid, data,
+                ));
+            }
+            BitswapEvent::ReceivedWant(peer_id, cid, _priority) => {
+                let cid = cid.to_string();
+                let cid: Cid = Cid::from_str(&cid).unwrap();
```
austinabell, line 107: same as previous

node/forest_libp2p/src/behaviour.rs
```diff
@@ -55,6 +63,7 @@ impl NetworkBehaviourEventProcess<MdnsEvent> for ForestBehaviour {
             MdnsEvent::Discovered(list) => {
                 for (peer, _) in list {
                     trace!("mdns: Discovered peer {}", peer.to_base58());
+                    self.connect(peer.clone());
```
austinabell, line 66: should this also be called on manual dials? I can't comment anywhere else (and isn't really a thing rn) but we probably will introduce functionality to dial peers manually through cli or otherwise and this would be forgotten by then

node/forest_libp2p/src/behaviour.rs
```diff
@@ -82,6 +92,29 @@ impl NetworkBehaviourEventProcess<KademliaEvent> for ForestBehaviour {
...
+                ));
+            }
+            BitswapEvent::ReceivedWant(peer_id, cid, _priority) => {
+                let cid = cid.to_string();
+                let cid: Cid = Cid::from_str(&cid).unwrap();
+                self.events
+                    .push(ForestBehaviourEvent::BitswapReceivedWant(peer_id, cid));
+            }
+            BitswapEvent::ReceivedCancel(_peer_id, _cid) => {
+                // TODO: Determine how to handle cancel
```
austinabell, line 112: should at least log in some way for now

node/forest_libp2p/src/behaviour.rs
```diff
@@ -253,4 +287,41 @@ impl ForestBehaviour {
...
+
+    /// Send a block to a peer over bitswap
+    pub fn send_block(
+        &mut self,
+        peer_id: &PeerId,
+        cid: Cid,
+        data: Box<[u8]>,
+    ) -> Result<(), Box<dyn Error>> {
+        let cid = cid.to_string();
+        log::debug!("send {}", cid);
```
austinabell, line 304: nit: can you import so we call these consistently? Also might be useful to include in the log message that you sent over bitswap since the module would just be `forest_libp2p::behaviour` so could be anything

node/forest_libp2p/src/behaviour.rs
```diff
@@ -253,4 +287,41 @@ impl ForestBehaviour {
...
+        let cid = Cid2::from_str(&cid)?;
+        self.bitswap.send_block(peer_id, cid, data);
+        Ok(())
+    }
+
+    /// Send a request for data over bitswap
+    pub fn want_block(&mut self, cid: Cid, priority: Priority) -> Result<(), Box<dyn Error>> {
+        let cid = cid.to_string();
+        log::debug!("want {}", cid);
```
austinabell, line 313: same as previous

node/forest_libp2p/src/behaviour.rs
```diff
@@ -253,4 +287,41 @@ impl ForestBehaviour {
...
+    }
+
+    /// Send a block to a peer over bitswap
+    pub fn send_block(
+        &mut self,
+        peer_id: &PeerId,
+        cid: Cid,
+        data: Box<[u8]>,
+    ) -> Result<(), Box<dyn Error>> {
+        let cid = cid.to_string();
```
austinabell, line 303: less of an issue for these, but can avoid decoding from string if conversion just to and from bytes

node/forest_libp2p/src/behaviour.rs
```diff
@@ -253,4 +287,41 @@ impl ForestBehaviour {
...
+        let cid = Cid2::from_str(&cid)?;
+        self.bitswap.want_block(cid, priority);
+        Ok(())
+    }
+
+    /// Cancel a bitswap request
+    pub fn cancel_block(&mut self, cid: &Cid) -> Result<(), Box<dyn Error>> {
+        let cid = cid.to_string();
+        log::debug!("cancel {}", cid);
+        let cid = Cid2::from_str(&cid)?;
```
austinabell, line 323: all of the above lol

blockchain/chain_sync/src/sync.rs
ec2: Yeah its just to ensure a request gets made before a response gets sent

blockchain/chain_sync/src/network_handler.rs
ec2: This fails if the receiver is dropped, which it shouldn't be, but ill change it to log an error is something happens here.

node/forest_libp2p/src/behaviour.rs
ec2: Hm i agree, but then we can log debug messages for... well... debugging haha

node/forest_libp2p/src/behaviour.rs
ec2: good catch

node/forest_libp2p/src/behaviour.rs
```diff
@@ -3,6 +3,8 @@
 
 use super::rpc::{RPCEvent, RPCMessage, RPC};
 use crate::config::Libp2pConfig;
+use forest_cid::Cid;
+use libipld_core::cid::Cid as Cid2;
```
dutterbutter, line 7: Opinionated: Change `Cid2` -> `LibCid` or something other than just the addition of 2. Again opinionated so up to you.

node/forest_libp2p/src/service.rs
```diff
@@ -56,18 +62,28 @@ pub enum NetworkMessage {
     RPC { peer_id: PeerId, event: RPCEvent },
 }
 /// The Libp2pService listens to events from the Libp2p swarm.
-pub struct Libp2pService {
+pub struct Libp2pService<DB: BlockStore> {
     pub swarm: Swarm<ForestBehaviour>,
 
+    db: Arc<DB>,
+
```
dutterbutter, line 69: Nit: Remove spacing between fields for struct

node/forest_libp2p/src/behaviour.rs
ec2: I like Cid2. It sounds weird, but I think it better illustrates that we shouldn't be using this library anywhere else... Lemme know, I'm not too emotional about this

ec2:
NOTE: This PR is based on #534, so this shouldnt come in before that.

blockchain/chain_sync/src/network_handler.rs
```diff
@@ -46,9 +48,16 @@ impl NetworkHandler {
...
+                        let tx = request_table.lock().await.remove(&request_id);
+                        if tx.is_none() {
+                            debug!("RPCResponse receive failed: channel not found");
+                            continue;
+                        }
+                        let tx = tx.unwrap();
+
+                        if let Err(e) = tx.send(RPCResponse::BlockSync(response)) {
+                            debug!("RPCResponse receive failed: {:?}", e)
+                        }
```
austinabell, line 60: can do this with an if let or match, seems weird to check none then unwrap? guess that's opinionated tho

ec2: wow yeah... temporary brain fart here... i hate continues.

blockchain/chain_sync/src/network_context.rs
```diff
@@ -129,28 +134,19 @@ impl SyncNetworkContext {
...
         peer_id: PeerId,
         request: RPCRequest,
-    ) -> Result<RPCResponse, &'static str> {
+    ) -> OneShotReceiver<RPCResponse> {
         let request_id = self.request_id;
         self.request_id.0 += 1;
+
+        let (tx, rx) = oneshot_channel();
+        self.request_table.lock().await.insert(request_id, tx);
         self.network_send
```
austinabell, line 143: Why not just pass in the return channel into the RPC message sent into the networking service so you don't have to handle the response in this circular way? Instead of sending the request, modifying a shared hashmap through a mutex, and then manually getting the inbound in another place to await the response, you just pass in the send channel and await the response with a timeout

node/forest_libp2p/src/service.rs
```diff
@@ -172,6 +187,28 @@ impl Libp2pService {
...
                             }).await;
                         }
+                        ForestBehaviourEvent::BitswapReceivedBlock(peer_id, cid, block) => {
+                            match self.db.put(&block, Blake2b256) {
+                                Ok(actual_cid) => {
+                                    trace!("saved bitswap block with returned cid {:?}, expected cid: {:?}", actual_cid, cid);
+                                }
+                                Err(e) => {
+                                    warn!("failed to save bitswap block: {:?}", e.to_string());
+                                }
```
dutterbutter, line 197: Question(s): 
- Why check if the actual is different from the expected without any action if that is true?
- Why do we need to `put` the bitswap blocks here? Wouldn't they be written to the DB during sync? I understand that we need to store the block to be able to send out `ReceivedWant` blocks but I figured it would be stored during the syncing process after validation?

blockchain/chain_sync/Cargo.toml
```diff
@@ -31,6 +31,8 @@ num-traits = "0.2"
 filecoin-proofs-api = "4.0.1"
 fil_types = { path = "../../types" }
 commcid = { path = "../../utils/commcid" }
+flo_stream = "0.4"
+
```
dutterbutter, line 35: Nit: Remove extra white space

node/forest_libp2p/src/service.rs
ec2: I dont check for it. I just log the 2 Cids for debugging purposes. I am not entirely sure how we want to handle scenarios like this yet... 

I'd say its okay to save messages to the db before checking them. Under the assumtion that the Cid of the data matches what we get for the Cid when we save it, we can delete it after. This design might change though. This PR doesnt involve using bitswap yet, so some of this stuff will probably change.

node/forest_libp2p/src/behaviour.rs
austinabell: > Hm i agree, but then we can log debug messages for... well... debugging haha

Idk what this means, but these still aren't resolved

node/forest_libp2p/Cargo.toml
```diff
@@ -23,6 +23,9 @@ fnv = "1.0.6"
 smallvec = "1.1.0"
 clock = { path = "../clock" }
 num-bigint = { path = "../../utils/bigint", package = "forest_bigint" }
+libp2p-bitswap = { git = "https://github.com/ChainSafe/libp2p-bitswap", rev = "c31e0ae31fa7d0b848bf4fb711106e59aac0fa5e" }
```
austinabell, line 26: Can probably switch to 0.5 now (has needed libp2p version bump)

node/forest_libp2p/src/service.rs
```diff
@@ -180,8 +236,12 @@ impl Libp2pService {
...
                             swarm_stream.get_mut().publish(&topic, message);
                         }
-                        NetworkMessage::RPC { peer_id, request, id } => {
-                            swarm_stream.get_mut().send_rpc_request(&peer_id, request, id);
+                        NetworkMessage::HelloRequest { peer_id, request, id } => {
+                            swarm_stream.get_mut().send_rpc_request(&peer_id, RPCRequest::Hello(request), id);
+                        }
+                        NetworkMessage::BlockSyncRequest { peer_id, request, id, response_channel } => {
+                            self.bs_request_table.lock().await.insert(id, response_channel);
+                            swarm_stream.get_mut().send_rpc_request(&peer_id, RPCRequest::BlockSync(request), id);
```
austinabell, line 244: shouldn't need to pass in the id anymore, `send_rpc_request` should change to not use the forked method and let the RequestResponse service handle the id generation and then just insert into the table with the returned ID

blockchain/chain_sync/src/sync.rs
```diff
@@ -1052,20 +1056,24 @@ mod tests {
...
-                .await;
+            match blocksync_message.recv().await.unwrap() {
+                NetworkMessage::BlockSyncRequest {
+                    peer_id: _,
+                    request: _,
+                    response_channel,
+                } => {
+                    response_channel.send(rpc_response).unwrap();
+                }
+                _ => unreachable!(),
```
AshantiMutinta, line 1075: I think if you use an if let here you might not even need the unreachable macro

node/forest_libp2p/src/behaviour.rs
```diff
@@ -336,4 +372,41 @@ impl ForestBehaviour {
...
+    /// Send a block to a peer over bitswap
+    pub fn send_block(
+        &mut self,
+        peer_id: &PeerId,
+        cid: Cid,
+        data: Box<[u8]>,
+    ) -> Result<(), Box<dyn Error>> {
+        debug!("send {}", cid.to_string());
+        let cid = cid.to_bytes();
+        let cid = Cid2::try_from(cid)?;
```
AshantiMutinta, line 390: Could be worth it implementing a try_from trait from cid into cid2?

node/forest_libp2p/src/behaviour.rs
```diff
@@ -336,4 +372,41 @@ impl ForestBehaviour {
     pub fn peers(&self) -> &HashSet<PeerId> {
         &self.peers
     }
+
+    /// Adds peer to bitswap peer set
+    pub fn connect(&mut self, peer_id: PeerId) {
```
austinabell, line 377: how is this different from `add_peer`? Shouldn't that bitswap connect just happen there?

blockchain/chain_sync/src/sync.rs
austinabell: Yeah if you do this, you'd have to wrap in a loop and break when you send the response you want. Doesn't really matter much because it's a test tho

blockchain/chain_sync/src/sync.rs
```diff
@@ -1104,16 +1110,19 @@ mod tests {
         task::block_on(async move {
             cs.peer_manager.add_peer(source.clone(), None).await;
             assert_eq!(cs.peer_manager.len().await, 1);
-
-            let return_set = cs.sync_headers_reverse(head, &to).await;
-            assert_eq!(return_set.unwrap().len(), 4);
+            // make blocksync request
+            let return_set = task::spawn(async move { cs.sync_headers_reverse(head, &to).await });
+            task::sleep(Duration::from_secs(1)).await;
```
austinabell, line 1115: You shouldn't need the sleep now that you're awaiting the receiver channel and then sending the response?

node/forest_libp2p/src/service.rs
```diff
@@ -103,8 +125,12 @@ impl Libp2pService {
 
         let (network_sender_in, network_receiver_in) = channel(20);
         let (network_sender_out, network_receiver_out) = channel(20);
+
+        let bs_request_table = Mutex::new(HashMap::new());
```
austinabell, line 129: You don't actually need the mutex anymore since the table is just used in the libp2p service. You just need to make the self param in the function `mut` and you're good

node/forest_libp2p/src/service.rs
```diff
@@ -167,11 +193,39 @@ impl Libp2pService {
...
+                        ForestBehaviourEvent::BitswapReceivedBlock(peer_id, cid, block) => {
+                            match self.db.put(&block, Blake2b256) {
+                                Ok(actual_cid) => {
+                                    trace!("saved bitswap block with returned cid {:?}, expected cid: {:?}", actual_cid, cid);
+                                }
+                                Err(e) => {
+                                    warn!("failed to save bitswap block: {:?}", e.to_string());
+                                }
+                            }
+                            self.network_sender_out.send(NetworkEvent::BitswapBlock{cid}).await;
```
austinabell, line 216: should you only send the event when the db put is successful?

node/forest_libp2p/src/service.rs
```diff
@@ -167,11 +193,39 @@ impl Libp2pService {
...
+                        },
+                        ForestBehaviourEvent::BitswapReceivedWant(peer_id, cid,) =>  match self.db.get(&cid) {
+                            Ok(Some(data)) => {
+                                swarm_stream.get_mut().send_block(&peer_id, cid, data).unwrap();
```
austinabell, line 220: seems like you shouldn't unwrap here?

node/forest_libp2p/src/behaviour.rs
```diff
@@ -110,6 +120,30 @@ impl NetworkBehaviourEventProcess<KademliaEvent> for ForestBehaviour {
     }
 }
 
+impl NetworkBehaviourEventProcess<BitswapEvent> for ForestBehaviour {
+    fn inject_event(&mut self, event: BitswapEvent) {
+        match event {
+            BitswapEvent::ReceivedBlock(peer_id, cid, data) => {
+                let cid = cid.to_bytes();
+                let cid: Cid = Cid::from_raw_cid(cid.as_slice()).unwrap();
```
austinabell, line 128: don't unwrap on these

node/forest_libp2p/src/behaviour.rs
```diff
@@ -110,6 +120,30 @@ impl NetworkBehaviourEventProcess<KademliaEvent> for ForestBehaviour {
...
+                self.events.push(ForestBehaviourEvent::BitswapReceivedBlock(
+                    peer_id, cid, data,
+                ));
+            }
+            BitswapEvent::ReceivedWant(peer_id, cid, _priority) => {
+                let cid = cid.to_bytes();
+                let cid: Cid = Cid::from_raw_cid(cid.as_slice()).unwrap();
```
austinabell, line 135: don't unwrap, again can't guarantee this will succeed

node/forest_libp2p/src/behaviour.rs
austinabell: might be a bit weird to link the two cid crates together (as that's the only place that trait impl can be added)

I don't really care either way, just noting

AshantiMutinta: Yeah a local helper would be a good alternative, missed the part where they were in two different crates

blockchain/chain_sync/src/sync.rs
AshantiMutinta: Ah yeah since it is a test it's not really an issue

node/forest_libp2p/src/behaviour.rs
austinabell: yeah I'm just thinking it would be nice to not link to the ipld core from our cid crate and just keep it in here.

node/forest_libp2p/src/behaviour.rs
ec2: You're right. Ill call this function there!

node/forest_libp2p/src/behaviour.rs
ec2: ah, sorry. left these in by accident

node/forest_libp2p/src/behaviour.rs
ec2: I agree. I do NOT want these conversions to be easy because they should be very rare. I hope these types of conversion should happen anywhere else.

node/forest_libp2p/src/service.rs
ec2: Yeah, I think thats probably what should happen, but something REALLY weird happens when i move that into the match statement... The error is too long to post here. Will show tomorrow

node/forest_libp2p/src/service.rs
```diff
@@ -171,11 +197,47 @@ impl Libp2pService {
...
-                            self.network_sender_out.send(NetworkEvent::BlockSyncResponse {
-                                request_id,
-                                response,
-                            }).await;
+                            let tx = self.bs_request_table.remove(&request_id);
+
+                            if let Some(tx) = tx {
+                                if let Err(e) = tx.send(response) {
+                                    debug!("RPCResponse receive failed: {:?}", e)
+                              }
```
austinabell, line 205: can you just fix the spacing on this? Got screwed up with last commit

ec2: lol damn macros

PR #521, opened 2020-06-25
Added proposal_hash field for TxnIDParams;

rllola:
**Summary of changes**
- Added `proposal_hash` field in TxnIDParams to fit new specs for actors

No issue associated

Required for multisig support and compatibility with Lotus.
See https://github.com/filecoin-project/specs-actors/commit/03a061425e56f5087f143dd08069117bdfd789bf

I haven't added the check in other place for multisig. Just the field so it can be serialize in order to create a correct message.

### DIFF
--- a/vm/actor/src/builtin/multisig/types.rs
+++ b/vm/actor/src/builtin/multisig/types.rs
@@ -22,6 +22,12 @@ impl TxnID {
     }
 }
 
+/// Proposal Hash
+// Blake2b 32 hash 
+#[derive(Clone, Copy, Default, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct ProposalHash(pub [u8; 32]);
+
 /// Transaction type used in multisig actor
 #[derive(Clone, PartialEq, Debug, Serialize_tuple, Deserialize_tuple)]
 pub struct Transaction {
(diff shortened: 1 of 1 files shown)


vm/actor/src/builtin/multisig/types.rs
```diff
@@ -22,6 +22,12 @@ impl TxnID {
     }
 }
 
+/// Proposal Hash
+// Blake2b 32 hash 
+#[derive(Clone, Copy, Default, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct ProposalHash(pub [u8; 32]);
```
austinabell, line 29: This will serialize incorrectly (as an array of integers not as a slice of bytes) should tag with serde_bytes

nannick: I had something similar already in my PR. I dont see the need to create a separate type though

rllola: @nannick Can I use your branch then ? I just needed this change so it is compatible with lotus.

austinabell: we delete branches and they get squashed, so the commit hash will be invalid once merged, though

rllola:
Not needed anymore. Waiting for the https://github.com/ChainSafe/forest/pull/522 to be merged.

PR #524, opened 2020-06-27
Implementing Market Balance

### LINKED ISSUE #488: implement market_balance
**Task summary**

Functionality to retrieve market balances from the balance table. Requires implementing both `lookup_id` and `market_balance` within the state manager.

**Specification reference**

- [MarketBalance - Lotus](https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/stmgr.go#L708)
- [LookUpID - Lotus](https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/stmgr.go#L451)

nannick:
**Summary of changes**
-  Implementing 2 functions listed in #488 
-  Introduced a new struct MarketBalance
           Currently placed in blockchain/state_manager/src/lib.rs, but not sure if this is  the best place to put it

Closes #488

### DIFF
--- a/blockchain/state_manager/src/lib.rs
+++ b/blockchain/state_manager/src/lib.rs
@@ -264,7 +264,7 @@ where
             },
             locked: {
                 let lt = BalanceTable::from_root(self.bs.as_ref(), &market_state.locked_table)
-                    .map_err(|_x| Error::State("Failed to build Locked Table".to_owned()))?;
+                    .map_err(|_x| Error::State("Failed to build Locked Table".to_string()))?;
                 lt.get(&new_addr).unwrap_or_default()
             },
         };
--- a/blockchain/state_manager/src/lib.rs
+++ b/blockchain/state_manager/src/lib.rs
@@ -260,7 +260,7 @@ where
             escrow: {
                 let et = BalanceTable::from_root(self.bs.as_ref(), &market_state.escrow_table)
                     .map_err(|_x| Error::State("Failed to build Escrow Table".to_owned()))?;
-                et.get(&new_addr).unwrap_or_else(|_| BigUint::from(0u8))
+                et.get(&new_addr).unwrap_or_default()
             },
             locked: {
                 let lt = BalanceTable::from_root(self.bs.as_ref(), &market_state.locked_table)
--- a/blockchain/state_manager/src/lib.rs
+++ b/blockchain/state_manager/src/lib.rs
@@ -4,7 +4,7 @@
 mod errors;
 pub mod utils;
 pub use self::errors::*;
-use actor::{init, miner, power, ActorState, INIT_ACTOR_ADDR, STORAGE_POWER_ACTOR_ADDR};
+use actor::{init, miner, power, ActorState, INIT_ACTOR_ADDR, STORAGE_POWER_ACTOR_ADDR, STORAGE_MARKET_ACTOR_ADDR, market, BalanceTable};
 use address::{Address, BLSPublicKey, Payload, BLS_PUB_LEN};
 use async_log::span;
 use async_std::sync::RwLock;
@@ -17,7 +17,7 @@ use forest_blocks::{Block, BlockHeader, FullTipset, Tipset, TipsetKeys};
 use interpreter::{resolve_to_key_addr, ChainRand, DefaultSyscalls, VM};
 use ipld_amt::Amt;
 use log::trace;
-use num_bigint::BigUint;
+use num_bigint::{BigUint};
 use state_tree::StateTree;
 use std::collections::HashMap;
 use std::error::Error as StdError;
@@ -25,6 +25,12 @@ use std::sync::Arc;
 
 /// Intermediary for retrieving state objects and updating actor states
 pub type CidPair = (Cid, Cid);
+
+#[derive(Default)]
+pub struct MarketBalance{
+    escrow : BigUint,
+    locked : BigUint
+}
 
 pub struct StateManager<DB> {
     bs: Arc<DB>,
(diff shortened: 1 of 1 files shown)


blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +243,35 @@ where
...
+        let market_state: market::State =
+            self.load_actor_state(&*STORAGE_MARKET_ACTOR_ADDR, ts.parent_state())?;
+
+        let new_addr = self.lookup_id(addr, ts)?;
+
+        let et = BalanceTable::from_root(self.bs.as_ref(), &market_state.escrow_table).unwrap();
+
+        let mut out = MarketBalance::default();
+
+        if et.has(&new_addr).is_ok() {
```
austinabell, line 266: this will return true if the result of the function is `Ok` and not if the boolean in the Ok is true. (will give a false positive when returns `Ok(false)`) so should probably just replace `is_ok` with `?`

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +243,35 @@ where
...
+        &'a mut self,
+        addr: &Address,
+        ts: &Tipset,
+    ) -> Result<MarketBalance, Error> {
+        let market_state: market::State =
+            self.load_actor_state(&*STORAGE_MARKET_ACTOR_ADDR, ts.parent_state())?;
+
+        let new_addr = self.lookup_id(addr, ts)?;
+
+        let et = BalanceTable::from_root(self.bs.as_ref(), &market_state.escrow_table).unwrap();
```
austinabell, line 262: take out the unwrap and handle that error

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +243,35 @@ where
...
+
+        let mut out = MarketBalance::default();
+
+        if et.has(&new_addr).is_ok() {
+            out.escrow = et.get(&new_addr)?;
+        }
+
+        let lt = BalanceTable::from_root(self.bs.as_ref(), &market_state.locked_table).unwrap();
```
austinabell, line 270: handle the error

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +243,35 @@ where
...
+
+        if lt.has(&new_addr).is_ok() {
```
austinabell, line 272: same as other

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +243,35 @@ where
             )),
         }
     }
+
+    pub fn lookup_id<'a>(&'a self, addr: &Address, ts: &Tipset) -> Result<Address, Error> {
```
austinabell, line 247: lifetime not needed here
```suggestion
    pub fn lookup_id(&self, addr: &Address, ts: &Tipset) -> Result<Address, Error> {
```

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +243,35 @@ where
...
+        let state_tree = StateTree::new_from_root(self.bs.as_ref(), ts.parent_state())?;
+        state_tree.lookup_id(addr).map_err(Error::State)
+    }
+
+    pub fn market_balance<'a>(
+        &'a mut self,
```
austinabell, line 253: lifetime not needed

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +244,31 @@ where
...
+    pub fn market_balance(&mut self, addr: &Address, ts: &Tipset) -> Result<MarketBalance, Error> {
+        let market_state: market::State =
+            self.load_actor_state(&*STORAGE_MARKET_ACTOR_ADDR, ts.parent_state())?;
+
+        let new_addr = self.lookup_id(addr, ts)?;
+
+        let out = MarketBalance {
+            escrow: {
+                let et = BalanceTable::from_root(self.bs.as_ref(), &market_state.escrow_table)
+                    .map_err(|_x| Error::State("Failed to build Escrow Table".to_owned()))?;
```
AshantiMutinta, line 262:
```suggestion
                    .map_err(|_x| Error::State("Failed to build Escrow Table".to_string()))?;
```

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +244,31 @@ where
...
+                et.get(&new_addr).unwrap_or_else(|_| BigUint::from(0u8))
+            },
+            locked: {
+                let lt = BalanceTable::from_root(self.bs.as_ref(), &market_state.locked_table)
+                    .map_err(|_x| Error::State("Failed to build Locked Table".to_owned()))?;
+                lt.get(&new_addr).unwrap_or_else(|_| BigUint::from(0u8))
```
AshantiMutinta, line 268:
```suggestion
                lt.get(&new_addr).unwrap_or_default()
```

blockchain/state_manager/src/lib.rs
```diff
@@ -234,4 +244,31 @@ where
...
+        let market_state: market::State =
+            self.load_actor_state(&*STORAGE_MARKET_ACTOR_ADDR, ts.parent_state())?;
+
+        let new_addr = self.lookup_id(addr, ts)?;
+
+        let out = MarketBalance {
+            escrow: {
+                let et = BalanceTable::from_root(self.bs.as_ref(), &market_state.escrow_table)
+                    .map_err(|_x| Error::State("Failed to build Escrow Table".to_owned()))?;
+                et.get(&new_addr).unwrap_or_else(|_| BigUint::from(0u8))
```
AshantiMutinta, line 263:
```suggestion
                et.get(&new_addr).unwrap_or_default()
```

blockchain/state_manager/src/lib.rs
austinabell: any reason why you suggest this? `to_string` just calls `to_owned` and I think it's pretty opinionated which is more readable

AshantiMutinta: I guess it is pretty opinionated!, that is true. For me I just think to_string makes it explicit that I am changing from a type of &Str to a type of String.

timvermeulen: I also very slightly prefer `to_string` but I think this is a lost cause without a linter that enforces it 😛 There are just too many ways to do this conversion

austinabell: yeah sure thing, the default throughout the existing code is `to_owned` just because to me it was more explicit (since the to_string method usually comes with other conversion logic from types that use the `FromStr` trait) but happy to start using `to_string` as a default moving forward as it doesn't matter

PR #525, opened 2020-06-27
Remove bitvec dependency and other bit field changes

timvermeulen:
The main changes:

- remove the `bitvec` dependency as well as the `RlePlus::{encode, decode}` methods because we no longer need them
- add tests for `BitReader` and `BitWriter`, they were already tested through general bit field tests but now it's much more thorough
- add `rleplus!` and `bitfield!` macros to easily create bit fields for testing purposes (which loosely replace the `bitvec!` macro)

Minor stuff:

- improve `RlePlus`'s `Debug` implementation
- fix a bug in `RlePlus`'s `PartialEq` implementation
- add `FromIterator<bool>` and `FromIterator<usize>` implementations to `RlePlus` and `BitField`
- simplify `BitWriter::finish`


utils/bitfield/src/rleplus/mod.rs
```diff
@@ -173,140 +176,110 @@ impl RlePlus {
...
-    // Converts a bit field into a byte vector.
+    /// Converts a bit field into a byte vector.
     pub fn into_bytes(self) -> Vec<u8> {
-        self.0.into()
+        self.0
     }
 }
 
+#[macro_export]
+macro_rules! rleplus {
```
austinabell, line 195: can you maybe add a comment explaining that this takes in non-encoded bits and rle encodes them? Think my assumption at first was that the macro was just taking in encoded bits

timvermeulen: Yeah totally, I just forgot to document them

utils/bitfield/src/lib.rs
```diff
@@ -134,7 +137,7 @@ impl BitField {
     /// Returns an iterator over the indices of the bit field's set bits if the number
     /// of set bits in the bit field does not exceed `max`. Returns an error otherwise.
     pub fn bounded_iter(&self, max: usize) -> Result<impl Iterator<Item = usize> + '_> {
-        if max <= self.len() {
+        if self.len() <= max {
```
dutterbutter, line 140: Thank you!

PR #531, opened 2020-07-01
Adding RPC Configuration

### LINKED ISSUE #474: Add CLI/Config for RPC Port and Address
**Issue summary**

Allow the user to set the RPC port and listening address on startup through the CLI and Config

### LINKED ISSUE #475: Enable/Disable RPC in CLI/Config
**Issue summary**

Allow for the enabling and disabling of the RPC service through the CLI/Config

nannick:
**Summary of changes**
Notable changes introduced in this pull request:
- Adds some basic RPC configuration via a TOML file. 
- Changes the Config struct in forest/src/cli/config.rs to have 3 new parameters
- Updates the start_rpc function in node/rpc/src/lib.rs to take a rpc port and listen address 

Closes  #474
Closes  #475 


forest/src/cli/config.rs
```diff
@@ -12,6 +12,9 @@ pub struct Config {
     pub data_dir: String,
     pub genesis_file: Option<String>,
     pub drand_dist_public: DistPublic,
+    pub enable_rpc: bool,
+    pub rpc_port: Option<String>,
+    pub rpc_listen: Option<String>,
```
austinabell, line 17: These make more sense to be flags on the daemon command (although can still keep this) anyone else have thoughts on this?

austinabell, line 17: These don't need to be wrapped in an option, the default implementation is sufficient (this can never be None, the only reason the genesis file can be is when it uses the builtin genesis)

forest/src/cli/config.rs
```diff
@@ -23,7 +26,10 @@ impl Default for Config {
             drand_dist_public: DistPublic{coefficients: [hex::decode("82c279cce744450e68de98ee08f9698a01dd38f8e3be3c53f2b840fb9d09ad62a0b6b87981e179e1b14bc9a2d284c985").unwrap(),
                 hex::decode("82d51308ad346c686f81b8094551597d7b963295cbf313401a93df9baf52d5ae98a87745bee70839a4d6e65c342bd15b").unwrap(),
                 hex::decode("94eebfd53f4ba6a3b8304236400a12e73885e5a781509a5c8d41d2e8b476923d8ea6052649b3c17282f596217f96c5de").unwrap(),
-                hex::decode("8dc4231e42b4edf39e86ef1579401692480647918275da767d3e558c520d6375ad953530610fd27daf110187877a65d0").unwrap(),]}
+                hex::decode("8dc4231e42b4edf39e86ef1579401692480647918275da767d3e558c520d6375ad953530610fd27daf110187877a65d0").unwrap(),]},
+            enable_rpc : true,
+            rpc_port: Some("/api".to_string()),
+            rpc_listen: Some("127.0.0.1:8080".to_string())
```
austinabell, line 32:
```suggestion
            rpc_listen: Some("127.0.0.1:1234".to_string())
```
8080 is probably a bad default port, maybe we should do 1234 since it is Lotus' default and that might be what services plug into by default ?

forest/src/cli/config.rs
```diff
@@ -23,7 +26,10 @@ impl Default for Config {
             drand_dist_public: DistPublic{coefficients: [hex::decode("82c279cce744450e68de98ee08f9698a01dd38f8e3be3c53f2b840fb9d09ad62a0b6b87981e179e1b14bc9a2d284c985").unwrap(),
                 hex::decode("82d51308ad346c686f81b8094551597d7b963295cbf313401a93df9baf52d5ae98a87745bee70839a4d6e65c342bd15b").unwrap(),
                 hex::decode("94eebfd53f4ba6a3b8304236400a12e73885e5a781509a5c8d41d2e8b476923d8ea6052649b3c17282f596217f96c5de").unwrap(),
-                hex::decode("8dc4231e42b4edf39e86ef1579401692480647918275da767d3e558c520d6375ad953530610fd27daf110187877a65d0").unwrap(),]}
+                hex::decode("8dc4231e42b4edf39e86ef1579401692480647918275da767d3e558c520d6375ad953530610fd27daf110187877a65d0").unwrap(),]},
+            enable_rpc : true,
+            rpc_port: Some("/api".to_string()),
```
austinabell, line 31: this is an endpoint, not a port. the port is just `8080` in this case (api endpoint should not be configurable)

forest/src/daemon.rs
```diff
@@ -75,9 +75,29 @@ pub(super) async fn start(config: Config) {
...
+
+    // Use the provided RPC port and listener or use the default one
+    let rpc_port = config
+        .rpc_port
+        .unwrap_or_else(|| "/api".to_string())
+        .to_owned();
+    let rpc_listen = config
+        .rpc_listen
+        .unwrap_or_else(|| "127.0.0.1:8080".to_string())
+        .to_owned();
```
austinabell, line 87: do this within the if statement (shouldn't have to do this if rpc disabled). When the previous comment comes in this can be removed entirely anyway

node/rpc/src/lib.rs
```diff
@@ -19,7 +19,11 @@ async fn handle_json_rpc(mut req: Request<Server<MapRouter>>) -> tide::Result {
     Ok(Response::new(StatusCode::Ok).body_json(&res)?)
 }
 
-pub async fn start_rpc<DB: BlockStore + Send + Sync + 'static>(store: Arc<DB>) {
+pub async fn start_rpc<DB: BlockStore + Send + Sync + 'static>(
+    store: Arc<DB>,
+    rpc_port: String,
+    rpc_listen: String,
```
austinabell, line 25: these can both be &str and you can remove the references below

forest/src/daemon.rs
```diff
@@ -75,9 +75,29 @@ pub(super) async fn start(config: Config) {
...
+
+    let rpc_thread = if config.enable_rpc {
+        task::spawn(async {
+            info!("RPC enabled");
+            info!("Using {} as RPC port", rpc_port);
+            info!("Using {} as RPC listen", rpc_listen);
```
austinabell, line 93: This shouldn't be necessary, and if you do want to keep, just make it a `debug!` and just make this one log (this is unnecessarily verbose)

forest/src/daemon.rs
```diff
@@ -75,9 +75,29 @@ pub(super) async fn start(config: Config) {
...
+            start_rpc(db_rpc, rpc_port, rpc_listen).await;
+        })
+    } else {
+        task::spawn(async {
+            info!("RPC disabled");
```
austinabell, line 98:
```suggestion
            debug!("RPC disabled");
```

forest/src/daemon.rs
ec2: I would leave as info!, but waaaaaay less verbose.... Something like 
("JSONRPC Endpoint at: {}:{}", rpc_listen, rpc_port)

forest/src/daemon.rs
```diff
@@ -35,7 +35,7 @@ pub(super) async fn start(config: Config) {
         });
 
     // Initialize database
-    let mut db = RocksDb::new(config.data_dir + "/db");
+    let mut db = RocksDb::new(config.data_dir.clone() + "/db");
```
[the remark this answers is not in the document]
nannick: I was doing some dumb stuff earlier,  before this set up i kept running into partial move issues. I thought my only way out of it was to clone stuff, should've changed this back when i started doing it this way.

PR #532, opened 2020-07-02
State API - RPC methods

### LINKED ISSUE #476: Implement RPC methods for State API methods
**Issue summary**

We need to integrate the state manager into the RPC. 
The priority RPC methods which allow for Storage Miner integration are here prefixed by the word State: https://github.com/filecoin-project/lotus/blob/master/storage/miner.go#L48

For full matching of Lotus' API, the other State methods are here: https://github.com/filecoin-project/lotus/blob/master/node/impl/full/state.go

The majority of state methods are called into the utils in StateManager [here](https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/utils.go). 

### LINKED ISSUE #487: state manager implementation
Epic for parts related to the state manager implementation

### LINKED ISSUE #489: implement state call and replay methods
**Task summary**

Call methods are used in both [`paychannel_manager`](https://github.com/filecoin-project/lotus/blob/master/paychmgr/paych.go) and [`state_api`.](https://github.com/filecoin-project/lotus/blob/master/node/impl/full/state.go)

Assists in completing the RPC methods for StateAPI referenced in #476.

**Specification reference**

- [Call - Lotus](https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/call.go#L74)
- [ StateReplay - Lotus](https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/call.go#L88)

AshantiMutinta:
**Summary of changes**
- This PR implements states functions needed for the RPC

Closes #476 
Closes #487 
Closes #489 


blockchain/beacon/src/drand_api/api.rs
```diff
@@ -1,4 +1,4 @@
-// This file is generated by rust-protobuf 2.15.1. Do not edit
+// This file is generated by rust-protobuf 2.15.0. Do not edit
```
dutterbutter, line 1: You will need to do a `cargo update` so that it remains at the correct version.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -35,10 +36,18 @@ const BLOCKS_PER_EPOCH: u64 = 5;
 // A cap on the size of the future_sink
 const SINK_CAP: usize = 1000;
 
+#[derive(Clone, Debug)]
+pub enum ChainMessage {
```
dutterbutter, line 40: Please add comments to public structures/funcs where applicable

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,129 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use crate::errors::*;
+use crate::StateManager;
+
```
dutterbutter, line 6: Nit: Remove spaces between imports

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,129 @@
...
+    state_manager: &'a StateManager<DB>,
+    ts: &'a Tipset,
+    mcid: &'a Cid,
+) -> Result<(UnsignedMessage, ApplyRet), Error>
+where
+    DB: BlockStore,
+{
+    let mut outm: Option<UnsignedMessage> = None;
+    let mut outr: Option<ApplyRet> = None;
+    let _error_message_halt = "halt".to_string();
```
dutterbutter, line 103: Maybe, return a specific error type and check against that rather than just a string?

blockchain/state_manager/src/lib.rs
```diff
@@ -13,14 +14,19 @@ use async_log::span;
...
 use blockstore::BlockStore;
 use blockstore::BufferedBlockStore;
-use chain::{block_messages, ChainStore};
+use chain::{block_messages, ChainMessage, ChainStore};
 use cid::Cid;
 use encoding::de::DeserializeOwned;
+use flo_stream::Subscriber;
 use forest_blocks::{Block, BlockHeader, FullTipset, Tipset, TipsetKeys};
-use interpreter::{resolve_to_key_addr, ChainRand, DefaultSyscalls, VM};
+
```
dutterbutter, line 22: Nit: Remove spaces between imports

blockchain/state_manager/src/call.rs
AshantiMutinta: Oh yeah I think this one is not being used anywhere.

ec2:
You mention this PR closes #476 but I don't see it getting hooked up to JSONRPC as done here https://github.com/ChainSafe/forest/blob/main/node/rpc/src/lib.rs#L30

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +434,80 @@ where
...
+pub fn message_for_tipset<DB>(db: &DB, ts: &Tipset) -> Result<Vec<Box<dyn Message>>, Error>
+where
+    DB: BlockStore,
+{
+    let mut applied: HashMap<Address, u64> = HashMap::new();
+    let mut balances: HashMap<Address, BigUint> = HashMap::new();
+    let state = StateTree::new_from_root(db, ts.parent_state())?;
+
+    //message to get all messages for block_header into a single iterator
+    type BoxMessage = Box<dyn Message>;
```
austinabell, line 461: This type is only used in the line below, consider removing?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +434,80 @@ where
...
+    let mut get_message_for_block_header = |b: &BlockHeader| -> Result<Vec<BoxMessage>, Error> {
+        let (unsigned, signed) = block_messages(db, b)?;
+        let unsigned_box = unsigned.into_iter().map(|s| {
+            let box_msg: Box<dyn Message> = Box::new(s);
+            box_msg
+        });
+        let signed_box = signed.into_iter().map(|s| {
+            let box_msg: Box<dyn Message> = Box::new(s);
+            box_msg
+        });
```
austinabell, line 471:
```suggestion
        let unsigned_box = unsigned
            .into_iter()
            .map(|s| Box::new(s) as Box<dyn Message>);
        let signed_box = signed.into_iter().map(|s| Box::new(s) as Box<dyn Message>);
```
Optional

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +434,80 @@ where
...
+
+        let mut messages = Vec::new();
+        for message in unsigned_box.chain(signed_box) {
+            let from_address = message.from();
+            if let Some(_s) = applied.get(&from_address) {
```
austinabell, line 476:
```suggestion
```

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +434,80 @@ where
...
+                let actor_state = state
+                    .get_actor(from_address)?
+                    .ok_or_else(|| Error::Other("Actor state not found".to_string()))?;
+                applied.insert(*from_address, actor_state.sequence);
+                balances.insert(*from_address, actor_state.balance);
+            }
+            let apply = applied.get(from_address);
+            if apply.map(|s| s != &message.sequence()).unwrap_or_default() {
+                continue;
+            }
```
austinabell, line 486: should increment the sequence after this if you're trying to match Lotus, so replacing the get with a get_mut

```suggestion
            if let Some(seq) = applied.get_mut(from_address) {
                if *seq != message.sequence() {
                    continue;
                }
                *seq += 1;
            } else {
                continue;
            }
```
Can definitely be cleaned up, but that is the functionality you need

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +434,80 @@ where
...
+            let balance = balances.get(from_address).cloned();
+            if balance
+                .clone()
+                .map(|s| s < message.required_funds())
+                .unwrap_or_default()
+            {
+                continue;
+            }
+            balance.map(|s| balances.insert(*message.from(), s - message.required_funds()));
```
austinabell, line 495:
```suggestion
            if let Some(bal) = balances.get_mut(from_address) {
                if *bal < message.required_funds() {
                    continue;
                }
                *bal -= message.required_funds();
            } else {
                continue;
            }
```
Avoids cloning the bigint and having to reinsert into hashmap (might also be able to be cleaner, but this is functionally what is needed)

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +434,80 @@ where
...
+
+        Ok(messages)
+    };
+
+    ts.blocks().iter().fold(Ok(Vec::new()), |vec, b| {
+        let mut message_vec = vec?;
+        let mut messages = get_message_for_block_header(b)?;
+        message_vec.append(&mut messages);
+        Ok(message_vec)
+    })
```
austinabell, line 508: optional: don't know if much better but could pass in a mutable vec into `get_message_for_block_header` and avoid the extra vec allocations and appending from them

blockchain/chain/src/store/chain_store.rs
```diff
@@ -441,6 +526,19 @@ where
         .collect()
 }
 
+pub fn get_parent_reciept<DB>(
```
austinabell, line 529: This kinda goes for others as well, but can you add docs on the public methods/types?

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,125 @@
...
+pub type StateCallResult<T> = Result<InvocResult<T>, Error>;
+
+fn state_call_raw<DB>(
+    state_manager: &StateManager<DB>,
+    msg: &mut UnsignedMessage,
+    bstate: &Cid,
+    rand: &ChainRand,
+    bheight: &ChainEpoch,
+) -> StateCallResult<UnsignedMessage>
+where
```
austinabell, line 35: I'll have a deeper look at this later, but what is the point in having state_call and state_call_raw?

blockchain/state_manager/src/lib.rs
```diff
@@ -121,6 +158,7 @@ where
         &self,
         ts: &FullTipset,
         rand: &ChainRand,
+        call_back: Option<impl FnMut(Cid, UnsignedMessage, ApplyRet) -> Result<(), String>>,
```
austinabell, line 161: idk small nit, but shouldn't callback be one word? is call back also correct? Just noticing because it's shown up a few times in this PR

blockchain/state_manager/src/lib.rs
```diff
@@ -176,7 +214,7 @@ where
 
             let block_headers = tipset.blocks();
             // generic constants are not implemented yet this is a lowcost method for now
-            let cid_pair = self.compute_tipset_state(&block_headers)?;
+            let cid_pair = self.compute_tipset_state(&block_headers, Some(|_, _, _| Ok(())))?;
```
austinabell, line 217: what's the point of that callback? Seems it should be None?

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +262,238 @@ where
                 .collect::<Result<Vec<Block>, _>>()?;
             // convert tipset to fulltipset
             let full_tipset = FullTipset::new(blocks)?;
-            self.apply_blocks(&full_tipset, &chain_rand)
+            self.apply_blocks(&full_tipset, &chain_rand, call_back)
         })
     }
 
+    pub fn look_up_id(&self, addr: &Address, tipset: &Tipset) -> Result<Address, Error> {
+        let tipset = ChainStore::new(self.get_block_store())
```
ec2, line 270: Dont need to do all this. If you look at the tipset_from_keys method, its a wrapper for a function that does not take in the chainstore but just the blockstore

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +262,238 @@ where
...
+        let subscribers = self.subscriber.clone().ok_or_else(|| {
+            Error::Other("State Manager not subscribed to tipset head changes".to_string())
+        })?;
+        let mut back_search_wait = self.back_search_wait.clone().ok_or_else(|| {
+            Error::Other("State manager not subscribed to back search wait".to_string())
+        })?;
+        let message = chain::get_chain_message(self.get_block_store_ref(), cid)
+            .map_err(|err| Error::Other(format!("failed to load message {:}", err)))?;
+
+        let subscribers: Vec<ChainMessage> = subscribers.collect().await;
```
ec2, line 398: Shouldn't you just pull off the first item off the stream if I understand this correctly? If you collect it, you everything off that stream and can lose events.. Thoughts?

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +262,238 @@ where
...
+            _ => {
+                return Err(Error::Other(format!(
+                    "expected current head on SHC stream (got {:?})",
+                    first_subscriber
+                )))
+            }
+        };
+        let maybe_message_reciept = self.tipset_executed_message(&tipset, cid, &*message)?;
+        if let Some(r) = maybe_message_reciept {
+            if let ChainMessage::HcCurrent(current) = first_subscriber {
```
ec2, line 414: Is this check necessary? You do the same check in line 404 above

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +262,238 @@ where
...
+                            }
+                            candidate_tipset = Some(tipset);
+                            candidate_receipt = Some(receipt)
+                        }
+                    }
+                    _ => continue,
+                }
+            }
+
+            match back_search_wait.next().await {
```
ec2, line 478: If i understand correctly, this is waiting for the back search to complete essentially right? I dont understand why you need to have a subscriber as part of the struct. IT seems like when you start the search_back_for_msg you can create a one shot channel there and run the search back in a blocking task and it will send a notification through the one shot channel when it is done.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -34,11 +35,19 @@ const BLOCKS_PER_EPOCH: u64 = 5;
 
 // A cap on the size of the future_sink
 const SINK_CAP: usize = 1000;
+//Enum for pubsub channel that defines message type variant and data contained in message type.
+#[derive(Clone, Debug)]
+pub enum ChainMessage {
+    HcCurrent(Arc<Tipset>),
```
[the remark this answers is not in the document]
austinabell: head change I think, which imo the enum should just be `HeadChange` and the variant be current. I don't see how `ChainMessage` comes in, unless it's planning to be used more generally later

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +434,80 @@ where
     })
 }
 
+pub fn get_chain_message<DB>(db: &DB, key: &Cid) -> Result<Box<dyn Message>, Error>
```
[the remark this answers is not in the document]
AshantiMutinta: Yeah I thought since we already have the trait we could easily just call the methods on on the trait objects since they have getters. this is mapped form get_c_message in the lotus and from what I got the idea is that they are using a message type without caring about what type of concrete type it is a bit like a typeclass. I thought about an enum but it was a question of why use it when the trait already exists, didn't want to be reinventing the wheel but I can change that.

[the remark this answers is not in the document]
austinabell: I think it's fine the way it is now, can revisit that point specifically later. Was more thinking out loud for the second part as I setup the trait in anticipation of how it would be used, but has rarely been needed as such so far

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: It should be moved up, used in other functions as well

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: Thought this would cleaner without the internal mutability in get_message_for_block_headers

blockchain/state_manager/src/call.rs
AshantiMutinta: I don't think it is being called anywhere but I remember someone suggesting a do something similar in the past just to keep it consistent

blockchain/state_manager/src/lib.rs
AshantiMutinta: Yes this should be none, think I copied directly from lotus. They have a nullable callback but implemented this

blockchain/state_manager/src/lib.rs
AshantiMutinta: i was really tripped out by some line and I assumed they needed a vec of events. This is true

blockchain/state_manager/src/lib.rs
AshantiMutinta: It looked like almost the search_back_for_msg was being sent in externally? isMostly because all events are coming from tsub which is subheadchanges? What do you think of this?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -34,11 +35,19 @@ const BLOCKS_PER_EPOCH: u64 = 5;
 
 // A cap on the size of the future_sink
 const SINK_CAP: usize = 1000;
+//Enum for pubsub channel that defines message type variant and data contained in message type.
+#[derive(Clone, Debug)]
+pub enum ChainMessage {
+    HcCurrent(Arc<Tipset>),
+    HcApply(Arc<Tipset>),
+    HcStore,
```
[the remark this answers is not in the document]
AshantiMutinta: Does a few checks in wait_for_message thought it would be better just to include the whole enum

blockchain/state_manager/src/lib.rs
ec2: Ok, i think i understand more whats happening

vm/message/src/lib.rs
```diff
@@ -30,8 +31,16 @@ pub trait Message {
     fn params(&self) -> &Serialized;
     /// gas_price returns gas price for the message
     fn gas_price(&self) -> &TokenAmount;
+    //sets the gas price
+    fn set_gas_price(&mut self, _: TokenAmount);
```
ec2, line 35: why are the params `_`?

node/state_api/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+/// StateMinerInfo returns info about the indicated miner
+pub fn state_miner_info<DB>(
+    state_manager: &StateManager<DB>,
+    actor: &Address,
+    key: &TipsetKeys,
+) -> Result<MinerInfo, BoxError>
+where
+    DB: BlockStore,
+{
+    let tipset = ChainStore::new(state_manager.get_block_store()).tipset_from_keys(key)?;
```
ec2, line 95: dont create a new chainstore here. use `chain::tipset_from_keys`

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +437,78 @@ where
...
+        .ok_or_else(|| Error::UndefinedKey(key.to_string()))?;
+    if let Ok(message) = from_slice::<UnsignedMessage>(&value) {
+        Ok(Box::new(message))
+    } else {
+        let signed_message: SignedMessage = from_slice(&value)?;
+        Ok(Box::new(signed_message))
+    }
+}
+
+//given a tipset this functtion will return all messages as a trait object
```
ec2, line 456: spelling mistakes and missing a /

blockchain/chain/src/store/chain_store.rs
```diff
@@ -35,10 +36,21 @@ const BLOCKS_PER_EPOCH: u64 = 5;
...
 const SINK_CAP: usize = 1000;
 
+type BoxMessage = Box<dyn Message>;
+
+/// Enum for pubsub channel that defines message type variant and data contained in message type.
+#[derive(Clone, Debug)]
+pub enum HeadChange {
+    Current(Arc<Tipset>),
+    Apply(Arc<Tipset>),
+    Store,
```
austinabell, line 46: Idk my comment still remains, I remove the Store enum and everything works fine, what am I missing?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +437,78 @@ where
     })
 }
 
+//this attempts to deserialize to unsigend message or signed message and then returns it at as a message trait object
```
austinabell, line 440:
```suggestion
/// Attempts to deserialize to unsigend message or signed message and then returns it at as a message trait object
```

blockchain/chain/src/store/chain_store.rs
```diff
@@ -423,6 +437,78 @@ where
...
+//given a tipset this functtion will return all messages as a trait object
+pub fn messages_for_tipset<DB>(db: &DB, ts: &Tipset) -> Result<Vec<BoxMessage>, Error>
+where
+    DB: BlockStore,
+{
+    let mut applied: HashMap<Address, u64> = HashMap::new();
+    let mut balances: HashMap<Address, BigUint> = HashMap::new();
+    let state = StateTree::new_from_root(db, ts.parent_state())?;
+
+    //message to get all messages for block_header into a single iterator
```
austinabell, line 465: idk small nit but for comments can you add a space after the `//` for consistency?

blockchain/chain/src/store/chain_store.rs
austinabell: yeah I guess that's fine, just seems weird to allocate when it's just a local closure

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,130 @@
...
+    DB: BlockStore,
+{
+    let ts = if let Some(t_set) = tipset {
+        t_set
+    } else {
+        let ts = chain::get_heaviest_tipset(state_manager.get_block_store_ref())
+            .map_err(|_| Error::Other("Could not get heaviest tipset".to_string()))?;
+        ts.ok_or_else(|| Error::Other("Empty Tipset given".to_string()))?
```
austinabell, line 87:
```suggestion
        chain::get_heaviest_tipset(state_manager.get_block_store_ref())
            .map_err(|_| Error::Other("Could not get heaviest tipset".to_string()))?
            .ok_or_else(|| Error::Other("Empty Tipset given".to_string()))?
```
Optional: Just was a bit awkward to read this way for me personally

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,130 @@
...
+            msg: msg.clone(),
+            msg_rct: apply_ret.msg_receipt().clone(),
+            actor_error: apply_ret.act_error().map(|e| e.to_string()),
+        })
+    })
+}
+
+/// runs the given message and returns its result without any persisted changes.
+pub fn state_call<DB>(
+    state_manager: &StateManager<DB>,
```
austinabell, line 75: why aren't these functions implementations of `StateManager`? These referring to this and .._raw

node/state_api/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+actor = { path = "../../vm/actor/" }
+clock = { path = "../../node/clock" }
+message = { package = "forest_message", path = "../../vm/message" }
+num-bigint = "0.2.6"
+num-traits = "0.2"
+cid = { package = "forest_cid", path = "../../ipld/cid" }
+async-std = { version = "1.6.0", features = ["unstable"] }
+state_tree = { path = "../../vm/state_tree/" }
+interpreter = { path = "../../vm/interpreter/" }
+fil_types = { path = "../../types" }
```
austinabell, line 23: Why have all these functions separated into it's own crate? I'm missing the rationale for this

node/state_api/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+use state_manager::{call, call::InvocResult, MarketBalance, StateManager};
+use state_tree::StateTree;
+use std::error::Error;
+
+type BoxError = Box<dyn Error + 'static>;
+pub struct MessageLookup {
+    pub receipt: MessageReceipt,
+    pub tipset: Arc<Tipset>,
+}
+pub fn get_network_name<DB>(state_manager: &StateManager<DB>) -> Result<String, BoxError>
```
austinabell, line 33: Why do you prefix others with `state_` and this one with `get_`? Seems you could get confused with the `get_network_name` from state manager

node/state_api/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+    key: &TipsetKeys,
+) -> Result<(Option<Claim>, Claim), BoxError>
+where
+    DB: BlockStore,
+{
+    let tipset = ChainStore::new(state_manager.get_block_store()).tipset_from_keys(key)?;
+    state_manager::utils::get_power(&state_manager, &tipset, Some(actor)).map_err(|e| e.into())
+}
+
+pub fn state_page_collateral<DB>(
```
austinabell, line 236:
```suggestion
pub fn state_pledge_collateral<DB>(
```
Assume this is what you mean, can you add a TODO on this function also, so this doesn't stay unimplemented later please?

node/state_api/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+    DB: BlockStore,
+{
+    let block_store = state_manager.get_block_store_ref();
+    let tipset = if maybe_tipset.is_none() {
+        chain::get_heaviest_tipset(block_store)?
+    } else {
+        maybe_tipset
+    };
+
+    let (st, _) = task::block_on(state_manager.tipset_state(&tipset.unwrap()))?;
```
austinabell, line 293: There doesn't seem to be any guarantee `get_heaviest_tipset` will be Some, maybe handle the error unless I'm missing something?

node/state_api/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+/// returns the addresses of every actor in the state
+pub fn state_list_actors<DB>(
+    state_manager: &StateManager<DB>,
+    key: &TipsetKeys,
+) -> Result<Vec<Address>, BoxError>
+where
+    DB: BlockStore,
+{
+    let tipset = ChainStore::new(state_manager.get_block_store()).tipset_from_keys(key)?;
+    state_manager::utils::list_miner_actors(&state_manager, &tipset).map_err(|e| e.into())
```
austinabell, line 351: shouldn't this list all actors?

vm/interpreter/src/vm.rs
```diff
@@ -142,6 +143,9 @@ where
             }
 
             // Add callback here for reward message if needed
```
austinabell, line 145: remove this comment since this is being added in

vm/interpreter/src/vm.rs
```diff
@@ -167,10 +171,13 @@ where
         }
 
         // Add callback here for cron message if needed
```
austinabell, line 173: same

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: There is no harm in including it since it copies from the original enum like structure and in terms of allocation the other oneswill probably contribute more to the the size of the allocation than this one but I can remove it

AshantiMutinta:
> I don't get why you're creating this API in between the RPC and State manager. This seems like it should just be on the RPC level. I can't think of a reason why this would need to live in it's own crate, all of the `state_` functions should just be at the RPC level and all functions added with the `StateManager` should just be implementations of that struct it seems to me.

I was following how lotus has it with state_api being it's own self contained workspace while state_api looks like it has no ties to direct rpc requests. If a restructure needs to be done I would rather do it in my RPC_state_methods PR since that is what was meant to connect that part

blockchain/chain/src/store/chain_store.rs
austinabell: can you link to where this variant is? I just see:
```
const (
	HCRevert  = "revert"
	HCApply   = "apply"
	HCCurrent = "current"
)
```
in Lotus

node/state_api/src/lib.rs
AshantiMutinta: This function shouldn't be here, been removed.

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: Been starting at this for too long, I was confused sorry about that haha. Store does not need to be there

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +267,225 @@ where
...
+        tipset: &Tipset,
+        message: &dyn Message,
+    ) -> Result<Option<(Tipset, MessageReceipt)>, Error> {
+        let current = tipset.to_owned();
+
+        if current.epoch() == 0 {
+            return Ok(None);
+        }
+
+        if let Some(actor_state) = self.get_actor(message.from(), tipset.parent_state())? {
```
austinabell, line 285:
```suggestion
        current: &Tipset,
        message: &dyn Message,
    ) -> Result<Option<(Tipset, MessageReceipt)>, Error> {
        if current.epoch() == 0 {
            return Ok(None);
        }

        if let Some(actor_state) = self.get_actor(message.from(), current.parent_state())? {
```
Swap the names if you want, but the clone should be avoided as tipsets are not trivial to clone

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +267,225 @@ where
...
+                    if let Ok(m_cid) = s.to_cid()
+                    {
+                        let error_msg = format!("found message with equal nonce as the one we are looking for (F:{:} n {:}, TS: {:} n{:})",cid,message.sequence(),m_cid,s.sequence());
+                        return Some(Err(Error::Other(error_msg)))
+                    }
+                    let error_msg =format!("found message with equal nonce as the one we are looking for (F:{:} n {:}, TS: `Error Converting message to Cid` n{:})",cid,message.sequence(),s.sequence());
+                    return Some(Err(Error::Other(error_msg)))
+                }
+                if s.sequence() < message.sequence() {
+                    return Some(Ok(None));
```
austinabell, line 366: Are you sure this isn't:
```suggestion
                    return None;
```
Seems like that is what is intended here, otherwise you will keep checking messages even though the sequence cannot be equal to anymore

vm/message/src/lib.rs
```diff
@@ -30,8 +31,16 @@ pub trait Message {
...
+    //sets the gas limit
+    fn set_gas_limit(&mut self, _: u64);
+    /// sets a new sequence to the message
+    fn set_sequence(&mut self, _: u64);
     /// Returns the gas limit for the message
     fn gas_limit(&self) -> u64;
     /// Returns the required funds for the message
     fn required_funds(&self) -> TokenAmount;
+    //turns message into cid
+    fn to_cid(&self) -> Result<Cid, String>;
```
austinabell, line 45: You don't need this, it's in the `Cbor` trait and has nothing to do with a message type

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +267,225 @@ where
...
+                            .map_err(|err| {
+                                Error::Other(err.to_string())
+                            }),
+                        );
+                    }
+                    if let Ok(m_cid) = s.to_cid()
+                    {
+                        let error_msg = format!("found message with equal nonce as the one we are looking for (F:{:} n {:}, TS: {:} n{:})",cid,message.sequence(),m_cid,s.sequence());
+                        return Some(Err(Error::Other(error_msg)))
+                    }
```
austinabell, line 361: why do you need this?
```suggestion
```

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +267,225 @@ where
...
+                    let error_msg =format!("found message with equal nonce as the one we are looking for (F:{:} n {:}, TS: `Error Converting message to Cid` n{:})",cid,message.sequence(),s.sequence());
```
austinabell, line 362:
```suggestion
                    let error_msg = format!("found message with equal nonce as the one we are looking for (F:{:} n {:}, TS: `Error Converting message to Cid` n{:})", cid, message.sequence(), s.sequence());
```
Wonder why rustfmt doesn't work in this fn

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +267,225 @@ where
...
+            .search_back_for_message(&tipset, &*message)?
+            .ok_or_else(|| {
+                Error::Other("State manager not subscribed to back search wait".to_string())
+            })?;
+
+        let mut candidate_tipset: Option<Arc<Tipset>> = None;
+        let mut candidate_receipt: Option<MessageReceipt> = None;
+        let height_of_head = tipset.epoch();
+        let mut reverts: HashMap<TipsetKeys, bool> = HashMap::new();
+        loop {
```
austinabell, line 421: Isn't this an infinite loop? how does this get broken out of?

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +267,225 @@ where
...
+                        }
+                    }
+                    _ => continue,
+                }
+            }
+
+            match back_search_wait.next().await {
+                Some(_) => {
+                    if !reverts.get(back_tipset.key()).unwrap_or(&false)
+                        && height_of_head > back_tipset.epoch() + confidence
```
austinabell, line 473:
```suggestion
                        && height_of_head >= back_tipset.epoch() + confidence
```
Right?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -437,6 +450,78 @@ where
...
+    let mut balances: HashMap<Address, BigInt> = HashMap::new();
+    let state = StateTree::new_from_root(db, ts.parent_state())?;
+
+    // message to get all messages for block_header into a single iterator
+    let mut get_message_for_block_header = |b: &BlockHeader| -> Result<Vec<BoxMessage>, Error> {
+        let (unsigned, signed) = block_messages(db, b)?;
+        let unsigned_box = unsigned.into_iter().map(|s| Box::new(s) as BoxMessage);
+        let signed_box = signed.into_iter().map(|s| Box::new(s) as BoxMessage);
+
+        let mut messages = Vec::new();
```
austinabell, line 484:
```suggestion
        let mut messages = Vec::with_capacity(unsigned_box.len() + signed_box.len());
```
Avoid reallocations of this?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -437,6 +450,78 @@ where
...
+        for message in unsigned_box.chain(signed_box) {
+            let from_address = message.from();
```
austinabell, line 486: Seems we are skipping the step of preloading the address here, is there a specific reason for this? Should at least add a TODO maybe?

blockchain/state_manager/src/lib.rs
AshantiMutinta: They are return closes under Apply ,and back_search_wait variants

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        let actor = state_manager
+            .get_actor(msg.from(), bstate)?
+            .ok_or_else(|| Error::Other("Could not get actor".to_string()))?;
+        msg.set_sequence(actor.sequence);
+        let apply_ret = vm.apply_implicit_message(msg);
+        trace!("gas limit {:}", msg.gas_limit());
+        trace!("gas price {:?}", msg.gas_price());
+        trace!("value {:?}", msg.value());
```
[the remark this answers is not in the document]
AshantiMutinta: I think should be fine since it is under a span to add context but will change that

blockchain/state_manager/src/lib.rs
AshantiMutinta: Since it is a filter_map if it hits this comparison it should propagate this further. There needs to be a condition that calls for https://github.com/filecoin-project/lotus/blob/c4de2373fd13c909876cf459e59ead7749b9939d/chain/stmgr/stmgr.go#L698 if a None is returned the filter_map will go to the next one

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: If we are speaking about this step :  https://github.com/filecoin-project/lotus/blob/c4de2373fd13c909876cf459e59ead7749b9939d/chain/store/store.go#L718 that is moved further down in the loop as I saw no reason as to why they had it in a closure the preload essentially adds it to the hashmap which is done further in the loop

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,130 @@
...
+{
+    pub msg: Msg,
+    pub msg_rct: MessageReceipt,
+    pub actor_error: Option<String>,
+}
+
+pub type StateCallResult<T> = Result<InvocResult<T>, Error>;
+
+fn state_call_raw<DB>(
+    state_manager: &StateManager<DB>,
```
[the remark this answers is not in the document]
AshantiMutinta: Moving these asap!

blockchain/state_manager/src/call.rs
```diff
@@ -0,0 +1,130 @@
...
+                "unexpected error during execution : {:}",
+                error_message
+            )));
+        }
+    }
+
+    let out_ret =
+        outr.ok_or_else(|| Error::Other("given message not found in tipset".to_string()))?;
+    let out_mes =
+        outm.ok_or_else(|| Error::Other("given message not found in tipset".to_string()))?;
```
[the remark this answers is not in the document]
AshantiMutinta: Looks like outr is nullable changed invoc to that

blockchain/state_manager/src/lib.rs
AshantiMutinta: Looks like there is an outer for in the lotus implementation as well 

https://github.com/filecoin-project/lotus/blob/c4de2373fd13c909876cf459e59ead7749b9939d/chain/stmgr/stmgr.go#L538

blockchain/state_manager/src/lib.rs
austinabell: oh gotcha okay, yeah that logic block was confusing to me on first read

blockchain/state_manager/src/lib.rs
austinabell: nah that's just one loop, you have a loop within a loop, if the while let ever comes up with None, it will be an infinite loop? Seems sketchy to rely on breaks for this

AshantiMutinta: Oh yeahh was confused cause I saw a for in a case from the go side and read it wrong to think it would always check the case until it finds something. I should remove it

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +385,219 @@ where
...
+                        candidate_tipset = Some(tipset);
+                        candidate_receipt = Some(receipt)
+                    }
+                }
+                _ => (),
+            }
+        }
+
+        while back_search_wait.next().await.is_some() {
+            if !reverts.get(back_tipset.key()).unwrap_or(&false)
```
austinabell, line 585: This gets called for all `()` put through this pubsub channel, but this `back_tipset` never changes, would this not just loop over and call the same code on the tipset when the `back_tipset` was initiated?

AshantiMutinta: I thought since it is in a loop structure the same thing would be happening at the lotus side? If multiple backsearchWait events are being called the same code will potentially be called as well?

austinabell: Well, they are just calling that function asyncronously and only handling it once (and not recurring, they kinda use that channel as a oneshot. Seems like if you have this subscriber you could potentially be receiving events from another function call, unrelated to the logic of this. 

I don't see why these events would need to be published in a pubsub way, seems pretty local (maybe clarification around why this needs to be emitted like this would be great, as I'm confused)

AshantiMutinta: Definitely! Wanted to use a oneshot but it looks like subheadchanges is the one that is sending these events externally? That is where the tsub starts to match in the case? I was also confused as to why they need to be externally called but following lotus seems like that might be the way they want it? I could also be very confused it is not as easy to parse haha

https://github.com/filecoin-project/lotus/blob/c4de2373fd13c909876cf459e59ead7749b9939d/chain/stmgr/stmgr.go#L494

austinabell: Yes, I see they are using that and that is matched above, but this back search is completely distinct and you are using the same subscriber channel, this will just keep iterating over and never close, so this function will never end. Am I missing something here?

blockchain/state_manager/src/utils.rs
```diff
@@ -100,3 +109,270 @@ where
...
+    tipset: &Tipset,
+) -> Result<Option<SectorOnChainInfo>, Error>
+where
+    DB: BlockStore,
+{
+    let miner_actor_state: miner::State = state_manager
+        .load_actor_state(&address, &tipset.parent_state())
+        .map_err(|err| {
+            Error::State(format!(
+                "(get sset) failed to load miner actor state: {:}",
```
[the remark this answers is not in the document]
AshantiMutinta: That is the accurate error message 

https://github.com/filecoin-project/lotus/blob/9b1d2f4c137433c1fe1f5ed44ba2e70105a409f0/chain/stmgr/utils.go#L149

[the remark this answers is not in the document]
austinabell: So these error messages don't need to match Lotus for any reason, and I assume they just made this inconsistency by accident. All I'm saying is that it would be easier to debug an issue if one of those errors is hit.

Fine to leave as is though, this is just opinionated

vm/message/src/lib.rs
AshantiMutinta: If you look at ChainMessage in lotus, they have a function from switching method of converting to cid. 

https://github.com/filecoin-project/lotus/blob/9b1d2f4c137433c1fe1f5ed44ba2e70105a409f0/chain/types/message.go

Or I could introduce a new type that holds a BoxMessage and also is an implementer of cbor? It just felt like this mapped to what our Message Trait was doing

blockchain/state_manager/src/lib.rs
AshantiMutinta: For diagnostic purposes. if possible I want to be able to show the sequence on the cid but if not it should default. I could probably just return an error string based if the cid was found but the scope was so small that it didnt hurt

vm/message/src/lib.rs
austinabell: They only do that because it's idiomatic go and there is no way to combine interfaces, for us we would never need this. Better just to add the constraint that the type is also `Cbor` and you don't need this duplicate function attached to a trait that logically doesn't fit

blockchain/state_manager/src/lib.rs
austinabell: I don't think it's worth it to have two seperate error returns just to have a slightly different error message, why don't you just debug print the cid? (also why are you putting `{:}` in the format? I can't find what that does, it's not in the std::fmt docs)

blockchain/state_manager/src/lib.rs
```diff
@@ -50,10 +71,39 @@ where
         Self {
             bs,
             cache: RwLock::new(HashMap::new()),
+            subscriber: None,
+            back_search_wait: None,
+        }
+    }
+
+    //Creates a constructor that passes in a HeadChange subscriber
```
austinabell, line 79: Also, can you change all these to doc comments (`///`)? And very specific nit but even when doing regular comments if you could please add a space after the `//` for consistency?

blockchain/state_manager/src/lib.rs
AshantiMutinta: I thought about it and debug_printing the cid will have less context just alone as compared to if it is printed with everything else. I will remove it but just felt really important. 

This is just calling on display {:} instead of fmt

austinabell: well I mean debug printing the option, (wouldn't that still display print the cid inside it?)

Also display is generally `{}`, is this for sure functional? (I have just never seen before)

AshantiMutinta: Yeah it is functional essentially just the same as { } https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=26ab59f94e2f4506c80a090e235685c7

blockchain/state_manager/src/lib.rs
```diff
@@ -185,9 +241,115 @@ where
...
+                outr = Some(apply_ret);
+                return Err("halt".to_string());
+            }
+
+            Ok(())
+        };
+        let result = self.compute_tipset_state(ts.blocks(), Some(callback));
+
+        if let Err(error_message) = result {
+            if error_message.to_string() == "halt" {
```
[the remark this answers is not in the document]
AshantiMutinta: I would prefer not to downcast if possible. It looks like the callback is just using this for halting purposes during execution. I think it is weird as it is because lotus is essentially using an error type as it's halting mechanic. Would have been better imo if a partial application of the sorts was returned. I think it works for now because the checking is only happening when the halt mechanic is applied

blockchain/state_manager/src/lib.rs
AshantiMutinta: this is my lack of go showing, I just recently knew that go func() is supposed to be an anoymous routine. Taking a look at this and fixing this right now, was trying to visualize as to what point the back_search_wait gets sent to and this makes more sense now

node/state_api/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+    state_manager: &StateManager<DB>,
+    cid: &Cid,
+    confidence: i64,
+) -> Result<MessageLookup, BoxError>
+where
+    DB: BlockStore,
+{
+    let (tipset, receipt) = task::block_on(state_manager.wait_for_message(cid, confidence))?;
+    let tipset = tipset.ok_or_else(|| "wait for msg returned empty tipset")?;
+    let receipt = receipt.ok_or_else(|| "wait for msg returned empty tipset")?;
```
austinabell, line 393: can you fix this error message?

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +365,229 @@ where
...
+                    Error::Other("State manager not subscribed to back search wait".to_string())
+                })?;
+            back_tipset = Some(back_t);
+            back_receipt = Some(back_r);
+            sender
+                .send(())
+                .map_err(|e| Error::Other(format!("Could not send to channel {:?}", e)))?;
+            Ok::<(), Error>(())
+        }
+        .await?;
```
austinabell, line 519: this will just compute this sequentially and await on it, would have to start a task to do this, and make sure it's cancelled if returned before it ends

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +365,229 @@ where
...
+
+        while let Some(subscriber) = self
+            .subscriber
+            .clone()
+            .ok_or_else(|| {
+                Error::Other("State Manager not subscribed to tipset head changes".to_string())
+            })?
+            .next()
+            .await
+        {
```
austinabell, line 541: This should be polled at the same time as the subscriber (both the receiver channel and pubsub receiver are futures) as right now you are waiting for the function to complete, then awaiting on the subscriber events. These changes don't change anything I don't think (unless I'm reading wrong)

blockchain/state_manager/src/lib.rs
AshantiMutinta: the issue was that the DB Associated Type is not a send it cannot be shared safely between threads. That was my initial struggle but it could benefit from the way futures are executed form the pool but I guess not. I don't really see the need for this unless search_back_message can be executed async since it is the one that is going to take a lot of execution time

austinabell: the only reason the db is a ref here is because it wasn't needed to remain and have cloned the `Arc`. If this does require concurrency (which is does if you want to match functionality) you should just change the parameter to an `Arc<DB>` and clone it for the async task that is doing this in parallel.

Calling await directly after the async block defeats the purpose of this

blockchain/state_manager/src/utils.rs
```diff
@@ -100,3 +109,270 @@ where
...
+pub fn get_power<DB>(
+    state_manager: &StateManager<DB>,
+    tipset: &Tipset,
+    address: Option<&Address>,
+) -> Result<(Option<Claim>, Claim), Error>
+where
+    DB: BlockStore,
+{
+    get_power_raw(state_manager, tipset.parent_state(), address)
+}
```
austinabell, line 342: I'm confused about why this is different from the one attached to the state manager?

blockchain/state_manager/src/utils.rs
```diff
@@ -100,3 +109,270 @@ where
...
+        .map_err(|err| {
+            Error::State(format!(
+                "(get miner info) failed to load miner actor state: {:}",
+                err
+            ))
+        })?;
+    Ok(miner_actor_state.info)
+}
+
+pub fn get_miner_deadlines<DB>(
```
austinabell, line 241: where is this function coming from? I'm having a hard time mapping some of these functions in this file

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +369,256 @@ where
...
+                            return Ok((Some(tipset), Some(receipt)));
+                        }
+                        candidate_tipset = Some(tipset);
+                        candidate_receipt = Some(receipt)
+                    }
+                }
+                _ => (),
+            }
+        }
+        let (back_tipset, back_receipt) = task.await?;
```
austinabell, line 607: the only thing is this should be awaiting on along with the subscriber channel. The point of this is that whichever comes first should be polled, so they should be polled at the same time, instead of waiting for one to complete then awaiting on this

blockchain/state_manager/src/utils.rs
AshantiMutinta: Yeah this was a mistake, shouldn't even be here.

blockchain/state_manager/src/utils.rs
AshantiMutinta: https://github.com/filecoin-project/lotus/blob/c459a13dcac66941c61b0c67105468e83e1bf873/chain/stmgr/utils.go#L305

blockchain/state_manager/src/lib.rs
```diff
@@ -223,10 +369,276 @@ where
...
+            if !should_revert && larger_height_of_head {
+                return Ok((Some(Arc::new(back_tipset)), Some(back_receipt)));
+            }
+
+            Ok((None, None))
+        });
+        futures.push(search_back_poll);
+
+        futures.next().await.ok_or_else(|| Error::Other("wait_for_message could not be completed due to failure of subscriber poll or search_back functionality".to_string()))?
+        // Ok((None, None));
```
austinabell, line 639: You're losing the error message here for this static string, is there a reason it's dropped here?

Can you also remove the commented out code too?

Cargo.lock
```diff
@@ -284,7 +284,7 @@ dependencies = [
  "ipld_blockstore",
  "ipld_hamt",
  "lazy_static",
- "libp2p",
+ "libp2p 0.21.1",
```
austinabell, line 287: What was the reason for bumping versions? was there something that broke? (wondering if there is something to fix) I tried without the updates and things seemed fine

blockchain/chain/src/store/chain_store.rs
austinabell: bump using capacity to avoid reallocations

Cargo.lock
AshantiMutinta: Not so sure, I just did a cargo update and found this

blockchain/state_manager/src/lib.rs
AshantiMutinta: Yes I can. Oh no it is an Option on ok_or_else so there isn't an error. The error is returned as part of the futures.next() since futures.next().await returrns an option

austinabell: ohhhh my bad, I misread :)

Cargo.lock
austinabell: ah yeah, that's what I meant to ask, just wasn't sure why the cargo update was necessary

PR #534, opened 2020-07-06
Move libp2p from fork and bump versions

austinabell:
**Summary of changes**
- Was using a fork of libp2p as a temporary measure, but they did not want to accept all the changes needed so instead just going to fork just the rpc crate and return it back to the released version for compatibility
  - Will revisit trying to make changes into libp2p for the functionality we need, but not going to allocate much time
- Bumps other versions (was needed to be able to resolve a common version, so I just bumped other non-opinionated versions

@ec2 pls verify this works with your changes needed :)

Closes 


ec2:
NOTE: This PR is based on #534, so this shouldnt come in before that.

PR #539, opened 2020-07-10
Implement Sync API and improve syncing

### LINKED ISSUE #535: Chain sync RPC methods
**Task summary**

Sync methods should be exposed from the jsonrpc server.

Methods include:

- [x] SyncCheckBad - checks if a block was marked as bad, and returns reason if so

- [ ] SyncIncomingBlocks (probably blocked rn without websockets) - incoming, not yet synced block headers

- [x] SyncMarkBad (pretty optional) - Mark blocks as bad manually so they won't be synced

- [x] SyncState - Current status of the chain sync (~opinionated return, not very clearly defined)

- [x] SyncSubmitBlock - Submit newly created block to network through node

**Specification reference**

- 

austinabell:
**Summary of changes**
- Implements sync API #535
  - Additional context is needed for API so I just included what was needed to be least burdensome, then can refactor later
- Implements Block gossipsub type `GossipBlock` to be able to serialize/deserialize them (Equivalent of `BlockMsg` in Lotus)
- Moves bad block cache logic to own type to be made threadsafe
- Creates type for sync state and updates the checkpoints for each stage (was assuming switching to graphsync previously)
- Fix bug in zero case of AMT (was serializing as [0] instead of [] for the AMT node bitfield)
- Update and fix blocksync conversion into FullTipset/Tipset
  - Previously was duplicate logic and bug with conversion to tipset

Also did a few other couple small things so feel free to ask about motivation or explanation for anything

Edit: oops forgot to add example commands to test manually:
```
curl -X POST --data '{ "jsonrpc": "2.0", "method": "Filecoin.SyncCheckBad", "params": [{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"}], "id": 1 }' -H "Content-Type: application/json" http://localhost:1234/rpc/v0
```

```
curl -X POST --data '{ "jsonrpc": "2.0", "method": "Filecoin.SyncMarkBad", "params": [{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"}], "id": 1 }' -H "Content-Type: application/json" http://localhost:1234/rpc/v0
```

```
curl -X POST --data '{ "jsonrpc": "2.0", "method": "Filecoin.SyncState", "id": 3 }' -H "Content-Type: application/json" http://localhost:1234/rpc/v0
```

```
curl -X POST --data '{ "jsonrpc": "2.0", "method": "Filecoin.SyncSubmitBlock", "params": [{"Header":{"Miner":"t01234","Ticket":{"VRFProof":"Ynl0ZSBhcnJheQ=="},"ElectionProof":{"VRFProof":"Ynl0ZSBhcnJheQ=="},"BeaconEntries":null,"WinPoStProof":null,"Parents":null,"ParentWeight":"0","Height":10101,"ParentStateRoot":{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"},"ParentMessageReceipts":{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"},"Messages":{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"},"BLSAggregate":{"Type":2,"Data":"Ynl0ZSBhcnJheQ=="},"Timestamp":42,"BlockSig":{"Type":2,"Data":"Ynl0ZSBhcnJheQ=="},"ForkSignaling":42},"BlsMessages":null,"SecpkMessages":null}], "id": 3 }' -H "Content-Type: application/json" http://localhost:1234/rpc/v0
```

Closes #535


blockchain/chain_sync/src/bad_block_cache.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl BadBlockCache {
+    pub fn new(cap: usize) -> Self {
+        Self {
+            cache: RwLock::new(LruCache::new(cap)),
+        }
+    }
+
+    /// Puts a bad block Cid in the cache with a given reason.
+    pub async fn put(&self, c: Cid, reason: String) -> Option<String> {
```
nannick, line 28: Is there a reason why you wanted to make the reason a String? I would've done a enum similar to the Error Types we have for the Actors

austinabell: for what benefit? The Error in actors isn't itself an enum btw, but I get the point

austinabell: there just isn't a reason to distinguish specifically yet, and the flexibility is nice. Is a good idea though and I may look into it before this PR comes in

austinabell: I made the change, but decided to force push back because using an enum for this uses 56 bytes rather than the 24 needed for a string. There is no functional gain for this, so the extra bytes on such a large cache is counterproductive

blockchain/chain_sync/src/sync.rs
```diff
@@ -267,17 +282,14 @@ where
                             .await?;
 
                         for b in ts_bundle.chain {
-                            let ts = ts[i as usize].clone();
                             // construct full tipsets from fetched messages
-                            let fts = self.tipset_msgs(
```
dutterbutter, line 272: Happy to see this method removed, however, I think we still need to include the BLOCK_MESSAGE_LIMIT check somewhere?

blockchain/chain_sync/src/sync.rs
```diff
@@ -416,9 +383,10 @@ where
         info!("Scheduling incoming tipset to sync: {:?}", tipset.cids());
 
         // check sync status if indicates tipsets are ready to be synced
-        if self.get_state() == &SyncState::Catchup {
+        // TODO revisit this, seems wrong
```
dutterbutter, line 386: Yes, agreed.

blockchain/chain_sync/src/sync_state.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Syncing headers from the heaviest tipset to genesis.
+    Headers,
+    /// Persisting headers on chain from heaviest to genesis.
+    PersistHeaders,
+    /// Syncing messages and performing state transitions.
+    Messages,
+    /// ChainSync completed and is following chain.
+    Complete,
+    /// Error has occured while syncing.
+    Errored,
```
dutterbutter, line 23: I think Error is sufficient, i dont think errored is a word BUT language is what we make it hahaha

blockchain/chain_sync/src/sync.rs
austinabell: I thought that was already validated for some reason, will add it in

node/forest_libp2p/src/blocksync/message.rs
```diff
@@ -65,46 +68,88 @@ pub struct TipsetBundle {
...
+        let secp_messages = values_from_indexes(&secp_msg_includes[i], &secp_msgs)?;
+
+        blocks.push(Block {
+            header,
+            secp_messages,
+            bls_messages,
+        });
     }
+
+    Ok(FullTipset::new(blocks).map_err(|e| e.to_string())?)
```
AshantiMutinta, line 154: Optional : another approach would be to move this into a try_fold but the downside is that you would be allocating

node/forest_libp2p/src/blocksync/message.rs
```diff
@@ -65,46 +68,88 @@ pub struct TipsetBundle {
...
+        let mut msgs = Vec::with_capacity(indexes.len());
+        for idx in indexes.iter() {
+            msgs.push(
+                values
+                    .get(*idx as usize)
+                    .cloned()
+                    .ok_or_else(|| "Invalid message index".to_string())?,
+            );
         }
+        Ok(msgs)
```
AshantiMutinta, line 131: Couldn't we just collect indexes.iter().map() into a Result<Vec<_>,_> here without allocating the msg?

node/forest_libp2p/src/blocksync/message.rs
austinabell: for what benefit?

node/forest_libp2p/src/blocksync/message.rs
austinabell: It gets allocated either way and is functionally the same, but I switched it

node/forest_libp2p/src/blocksync/message.rs
AshantiMutinta: You would be potentially getting rid of a mutable block vec and switching control to the try_fold but the scope if fairly small so not too worried about that

austinabell: why would you try fold instead of just doing an iter() map()? I'm not really seeing any optimization here, but I can commit this?

AshantiMutinta: That's fair, I can see that. 

Will retract my suggestion

blockchain/chain_sync/src/sync.rs
```diff
@@ -45,29 +46,13 @@ use std::convert::{TryFrom, TryInto};
...
-}
-
+/// Struct that handles the ChainSync logic. This handles incoming network events such as
+/// gossipsub messages, Hello protocol requests, as well as sending and receiving BlockSync
+/// messages to be able to do the initial sync.
 pub struct ChainSyncer<DB, TBeacon> {
     /// Syncing state of chain sync
-    state: SyncState,
+    // TODO should be a vector once syncing done async and ideally not wrap each state in mutex.
+    state: Arc<RwLock<SyncState>>,
```
ec2, line 55: I'd rename this to stage since the getter/setter are (get/set)_stage. Or change the method name

austinabell: SyncStage is a different meaning, and there is no methods you are referring to, can you reference what you mean?

ec2: Oh youre right. I misread, the stage is inside the state. M'bad, carry on :)

PR #542, opened 2020-07-14
Implement Mock Runtime Syscalls

### LINKED ISSUE #482: Implement MockRuntime syscalls
**Task summary**

Not currently needed, but will be soon to finish off porting over their tests to verify against.

This will require to create a struct and implement the Syscalls trait on it, then attach it to the mockruntime and add any setup functions needed for testing

**Specification reference**

- https://github.com/filecoin-project/specs-actors/blob/master/support/mock/mockrt.go#L419
(roughly around there, not going to pin to a commit)

nannick:
**Summary of changes**
-  Implementing the five syscalls for MockRunTime
- Added More ExpectVerify.* structs 
- Added PartialEq trait for PieceInfo 
- Added Clone Trait for ConsensusFault

Closes  #482 


vm/actor/tests/common/mod.rs
```diff
@@ -76,6 +80,36 @@ pub struct ExpectedVerifySig {
...
+    block_header_2: Vec<u8>,
+    block_header_extra: Vec<u8>,
+    fault: Option<ConsensusFault>,
+    exit_code: ExitCode,
+}
+
+#[derive(Clone)]
+pub struct ExpectComputeUnsealedSectorCid<'a> {
+    reg: RegisteredSealProof,
+    pieces: &'a Vec<PieceInfo>,
```
austinabell, line 108: you should never need to have a reference to a vec instead of using a slice, is there a reason this is taking a reference to this where other types take owned values?

nannick: No lol I was trying to match the argument in the function that used pieces. Completely forgot slices were a thing [herp derp]

PR #548, opened 2020-07-16
Test Runner for Message Signing Serialization Vectors

### LINKED ISSUE #339: Write test runner for Message signing serialization vectors
**Issue summary**

Should be done after #337 comes in as I added some of the conversions and types needed for this in that PR. 

Should be pretty straightforward to just create the message with the vector (done already and can move the types to be shared) and then just sign and check the signature against the vector. 

Both signatures are BLS signatures, so that can be handled specifically (Not worth to read signature type on vector to see what to sign with yet I don't think)

Test runners are in `tests/serialization_tests` and specific vector is `tests/serialization_vectors/message_signing.json`

flodesi:
**Summary of changes**
- Update Unsigned message serialization to account for null param values
- Write unsigned message test for serialization vectors

Closes #339


tests/serialization_tests/tests/signing_ser.rs
```diff
@@ -0,0 +1,42 @@
...
+fn signing_test() {
+    let mut file = File::open("../serialization-vectors/message_signing.json").unwrap();
+    let mut string = String::new();
+    file.read_to_string(&mut string).unwrap();
+
+    let vectors: Vec<TestVec> =
+        serde_json::from_str(&string).expect("Test vector deserialization failed");
+    for test_vec in vectors {
+        let test = base64::decode(test_vec.PrivateKey).unwrap();
+        let priv_key = PrivateKey::from_bytes(&test).unwrap();
```
austinabell, line 37: can you add a todo for setting up a private key based on the signature type? this is assuming this is always going to be bls signatures

vm/message/src/unsigned_message.rs
```diff
@@ -149,6 +149,29 @@ impl Message for UnsignedMessage {
...
+        D: Deserializer<'de>,
+    {
+        let s: Option<String> = Option::deserialize(deserializer)?;
+        if let Some(s) = s {
+            return Ok(s);
+        }
+        Ok("".to_string())
+    }
+}
+
```
austinabell, line 174: You don't need all this, can just wrap the string in an option

tests/serialization_tests/tests/signing_ser.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Deserialize)]
+struct TestVec {
+    #[serde(with = "unsigned_message::json")]
+    Unsigned: UnsignedMessage,
+    Cid: String,
+    CidHexBytes: String,
+    PrivateKey: String,
+    #[serde(with = "signature::json")]
+    Signature: Signature,
+}
```
austinabell, line 25:
```suggestion
#[derive(Deserialize)]
#[serde(rename_all = "PascalCase")]
struct TestVec {
    #[serde(with = "unsigned_message::json")]
    unsigned: UnsignedMessage,
    cid: String,
    private_key: String,
    #[serde(with = "signature::json")]
    signature: Signature,
}
```
idc because it's a test but you don't need to allow non snake case to do this. (Also you don't need to include the cid hex bytes and idk why they have this)

tests/serialization_tests/tests/signing_ser.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    let vectors: Vec<TestVec> =
+        serde_json::from_str(&string).expect("Test vector deserialization failed");
+    for test_vec in vectors {
+        let test = base64::decode(test_vec.PrivateKey).unwrap();
+        // TODO set up a private key based on sig type
+        let priv_key = PrivateKey::from_bytes(&test).unwrap();
+        let cid = test_vec.Unsigned.cid().unwrap();
+        let sig = priv_key.sign(cid.to_bytes().as_slice());
+        assert_eq!(sig.as_bytes().as_slice(), test_vec.Signature.bytes());
```
austinabell, line 41: One thing that is missing is the test for the signed message Cid. The intention of the test is that the signed message cid matches this. The test is setup pretty bad because there are only bls signed messages and the signed message cid is just the cid of the unsigned message, but there should still be the check for when secp signed message vectors are added

PR #572, opened 2020-07-28
JSON client setup and chain CLI commands

### LINKED ISSUE #571: Add Chain CLI commands
**Task summary**

Add chain cli commands for interacting with blockchain. 

**Specification reference**

- 

dutterbutter:
**Summary of changes**
- Adds json client setup 
- Refactored CLI and subcommand setup
- Adds basic chain CLI commands 

Closes #571 


forest/src/cli/chain_cmd.rs
```diff
@@ -0,0 +1,81 @@
...
+impl ChainCommands {
+    pub async fn run(&self) {
+        // TODO handle cli config
+        match self {
+            Self::Block { cid } => {
+                let cid: Cid = cid.parse().unwrap();
+                let mut client = new_client();
+
+                let blk = block(client, cid).await;
+                println!("{}", serde_json::to_string_pretty(&blk).unwrap());
```
austinabell, line 51: You're putting the result in json, make sure to change this so everything isn't wrapped in "Ok"

forest/src/cli/chain_cmd.rs
```diff
@@ -0,0 +1,81 @@
...
+                let mut client = new_client();
+
+                let gen = genesis(client).await;
+                println!("{}", serde_json::to_string_pretty(&gen).unwrap());
+            }
+            Self::Head => {
+                let mut client = new_client();
+
+                let head = head(client).await;
+                println!("{}", serde_json::to_string_pretty(&head).unwrap());
```
austinabell, line 63: Lotus prints out just the cid of the head, I think we should at least include the cid in the response so it can be used more easily, the full tipset is unlikely to be helpful

forest/src/cli/wallet_cmd.rs
```diff
@@ -0,0 +1,81 @@
...
+            Self::ReadObj { cid } => {
+                let cid: Cid = cid.parse().unwrap();
+                let mut client = new_client();
+
+                let obj = read_obj(client, cid).await;
+                println!("{}", serde_json::to_string_pretty(&obj).unwrap());
+            }
+        }
+    }
+}
```
austinabell, line 81: was this accidentally committed?

node/rpc-client/src/wallet_ops.rs
```diff
@@ -0,0 +1,64 @@
...
+pub async fn import(mut client: RawClient<HTC>) -> Result<String, JsonRpcError> {
+        Ok(Filecoin::wallet_import(&mut client)
+        .await?)
+}
+
+/// Returns the private key of an address in the wallet
+pub async fn export(mut client: RawClient<HTC>) -> Result<KeyInfoJson, JsonRpcError> {
+        Ok(Filecoin::wallet_export(&mut client)
+        .await?)
+}
```
austinabell, line 64: I mean, this part of the wallet stuff looks alright, but isn't connected and doesn't seem to be tested?

forest/src/cli/wallet_cmd.rs
dutterbutter: Yes, and I have no idea how?

node/rpc-client/src/wallet_ops.rs
dutterbutter: Yes, shouldn't have been included.

forest/src/cli/chain_cmd.rs
```diff
@@ -0,0 +1,126 @@
...
+use rpc_client::{block, genesis, head, messages, new_client, read_obj};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub enum ChainCommands {
+    /// Retrieves and prints out the block specified by the given CID
+    #[structopt(about = "<Cid> Retrieve a block and print its details")]
+    Block {
+        #[structopt(help = "Input a valid CID")]
+        cid: String,
```
nannick, line 16: I would just add the short option to all the arguments for subcommands on principle, wouldnt make much of a difference in this case. Its a nice to have

forest/src/cli/chain_cmd.rs
```diff
@@ -0,0 +1,126 @@
...
+            message,
+            data: _,
+        } => {
+            return warn!("JSON RPC Error: Code: {} Message: {}", code, message);
+        }
+        JsonRpcError::Provided { code, message } => {
+            return warn!("JSON RPC Error: Code: {} Message: {}", code, message);
+        }
+    }
+}
```
nannick, line 126: Feel like this should a general function, maybe move to  forest/src/cli/mod.rs

austinabell: this function also doesn't do what it says it does, should either return a string and swap the `warn!` for a `format!` or change this name

forest/src/cli/chain_cmd.rs
```diff
@@ -0,0 +1,126 @@
...
+            Self::Block { cid } => {
+                let cid: Cid = cid.parse().unwrap();
+                let client = new_client();
+
+                let blk = block(client, cid)
+                    .await
+                    .map_err(|e| {
+                        stringify_rpc_err(e);
+                    })
+                    .unwrap();
```
nannick, line 57: For the sync commands I'm handling the unwraps and outputting messages when certain commands fail. I get that this is just a CLI, and it doesn't persist, so error handling isnt a big issue. Just wanted other peoples thoughts, should we deal with the unwraps ?

node/rpc-client/src/chain_ops.rs
```diff
@@ -0,0 +1,38 @@
...
+use super::client::Filecoin;
+use blocks::{header::json::BlockHeaderJson, tipset_json::TipsetJson};
+use cid::{json::CidJson, Cid};
+use jsonrpc_v2::Error as JsonRpcError;
+use jsonrpsee::raw::RawClient;
+use jsonrpsee::transport::http::HttpTransportClient as HTC;
+use message::unsigned_message::json::UnsignedMessageJson;
+
+/// Returns a block with specified CID fom chain via RPC
+pub async fn block(mut client: RawClient<HTC>, cid: Cid) -> Result<BlockHeaderJson, JsonRpcError> {
```
nannick, line 13: Some CLI commands like FileCoin.SyncWait make multiple requests, so it might be better to have the client argument be a &mut in all of the *_ops.rs files

forest/src/cli/chain_cmd.rs
austinabell: the benefit is a cleaner message output, but at the cost of having to handle the error and exit gracefully. do it if it's easy to setup, but it isn't necessary for this sort of thing

forest/src/cli/chain_cmd.rs
```diff
@@ -0,0 +1,125 @@
...
+            message,
+            data: _,
+        } => {
+            return format!("JSON RPC Error: Code: {} Message: {}", code, message);
+        }
+        JsonRpcError::Provided { code, message } => {
+            return format!("JSON RPC Error: Code: {} Message: {}", code, message);
+        }
+    }
+}
```
nannick, line 125: Since the other commands also return RPC errors, i think it makes sense to move this to forest/src/cli/mod.rs instead of keeping it in this file

forest/src/cli/chain_cmd.rs
```diff
@@ -0,0 +1,110 @@
...
+        match self {
+            Self::Block { cid } => {
+                let cid: Cid = cid.parse().unwrap();
+                let mut client = new_client();
+
+                let blk = block(&mut client, cid)
+                    .await
+                    .map_err(|e| {
+                        stringify_rpc_err(e);
+                    })
```
austinabell, line 55:
```suggestion
                    .map_err(|e| stringify_rpc_err)
```
These should all be this, right now it's unwrapping a `()`

PR #593, opened 2020-07-31
Market actor update

### LINKED ISSUE #558: Update market actor
**Task summary**

**Specification reference**

- 

austinabell:
**Summary of changes**
- This was painful lol

The changes overlapping with the miner and power actor refactors I tried to leave out of this PR, so those actors will be pretty broken until I get around to them

Also all of the market tests broke with the protocol changes and there is now 3000 lines of code in the tests so I'm not doing in this PR

Closes #558 


vm/actor/src/builtin/market/policy.rs
```diff
@@ -1,49 +1,94 @@
...
+    size: PaddedPieceSize,
+    verified: bool,
+    network_qa_power: &StoragePower,
+    baseline_power: &StoragePower,
+    network_circulating_supply: &TokenAmount,
 ) -> (TokenAmount, TokenAmount) {
-    let v = TokenAmount::from(TOTAL_FILECOIN) * TokenAmount::from(TOKEN_PRECISION);
-    (TokenAmount::zero(), v) // PARAM_FINISH
+    // minimumProviderCollateral = (ProvCollateralPercentSupplyNum / ProvCollateralPercentSupplyDenom) * normalizedCirculatingSupply
+    // normalizedCirculatingSupply = FILCirculatingSupply * dealPowerShare
```
nannick, line 49: Remove commented code

vm/actor/src/builtin/market/mod.rs
```diff
@@ -195,184 +207,273 @@ impl Actor {
...
-                        "Cannot publish deals from different providers at the same time."
-                            .to_owned(),
-                    ));
+                    return Err(actor_error!(ErrIllegalArgument;
+                        "cannot publish deals from different providers at the same time."));
                 }
 
                 let client = rt.resolve_address(&deal.proposal.client)?.ok_or_else(|| {
                     actor_error!(ErrNotFound;
                         "failed to resolve provider address {}", provider_raw)
```
ec2, line 249: I think the log is wrong here. Should be the client address, not the provider address

vm/actor/src/builtin/market/mod.rs
```diff
@@ -390,46 +491,57 @@ impl Actor {
...
+                    |e| actor_error!(ErrIllegalState; "failed to get deal proposal: {}", e),
+                )?;
+                // deal could have terminated and hence deleted before the sector is terminated.
+                // we should simply continue instead of aborting execution here if a deal is not found.
+                if deal.is_none() {
+                    continue;
+                }
+                let deal = deal.unwrap();
+
+                assert_eq!(
```
ec2, line 511: I see they do an assert, in specs-actor, but do we actually wanna kill the program?

vm/actor/src/builtin/market/mod.rs
```diff
@@ -478,247 +596,361 @@ impl Actor {
...
+                            .as_mut()
+                            .unwrap()
+                            .delete(deal_id)
+                            .map_err(
+                                |e| actor_error!(ErrIllegalState; "failed to delete deal: {}", e),
+                            )?;
+                        msm.pending_deals
+                            .as_mut()
+                            .unwrap()
+                            .delete(&dcid.to_bytes())
```
ec2, line 685: For this delete and all the ones before, should we handle the case where the delete returns false indicating that what you're deleting doesnt exist?

vm/actor/src/builtin/market/mod.rs
austinabell: ah yeah, didn't think too much in detail about this one, seems very weird that they are asserting here (I will flag to them)

vm/actor/src/builtin/market/mod.rs
austinabell: good catch, actually didn't realize, and just checked, they just error whenever trying to delete something that doesn't exist (weird). Will update now

vm/actor/src/builtin/market/mod.rs
```diff
@@ -195,190 +207,278 @@ impl Actor {
...
+                let proposal = msm
+                    .deal_proposals
+                    .as_ref()
+                    .unwrap()
+                    .get(deal_id)
+                    .map_err(|e| {
+                        actor_error!(ErrIllegalState;
+                            "failed to get deal_id ({}): {}", deal_id, e)
+                    })?
+                    .ok_or_else(|| actor_error!(ErrNotFound; "no such deal_id: {}", deal_id))?;
```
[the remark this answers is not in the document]
austinabell: yeah this ones a bit harder to trace, it's within the `getDealProposal` function, the `!found` check. It was counterproductive to add that function and check a downcasted `Box<dyn StdError>` outside of it. (if this is still confusing I can explain in more detail)

PR #598, opened 2020-08-04
Add address bugfix tests and bump crate version

### LINKED ISSUE #595: Check for non-utf-8 characters when converting Addresses from string slices
**Issue summary**

A *panic* occurs when trying to convert an Address from a string slice with non-utf-8 characters, instead of returning an error, which would allow users to handle it gracefully. 
In the implementation of the *FromStr* trait for  Address in [vm/address/src/lib.rs](https://github.com/ChainSafe/forest/blob/main/vm/address/src/lib.rs#L158), there are three [places](https://github.com/ChainSafe/forest/blob/main/vm/address/src/lib.rs#L165) where the panics can happen as a result of sub-slicing by indexing a possible non-boundary character.

There is a function that helps with this in the standard [library](https://doc.rust-lang.org/beta/std/primitive.str.html#method.get).

austinabell:
**Summary of changes**
- Indexing certain multi byte characters was causing a potential panic and attack vector

Closes #595 

### DIFF
--- a/vm/address/tests/address_test.rs
+++ b/vm/address/tests/address_test.rs
@@ -559,3 +559,24 @@ fn set_network() {
     assert_eq!(addr.network(), Network::Mainnet);
     assert_eq!(addr.to_string(), "f01");
 }
+
+#[test]
+fn invalid_strings_tests() {
+    let invalid_strings = &[
+        "fömk3zcefvlgpay4f32c5vmruk5gqig6dumc7pz6q",
+        "🗻∈🌏ömk3zcefvlgpay4f32c5vmruk5gqig6dumc7pz6q",
+        "ö",
+    ];
+    for s in invalid_strings {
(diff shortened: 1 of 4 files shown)


vm/address/tests/address_test.rs
```diff
@@ -559,3 +559,24 @@ fn set_network() {
     assert_eq!(addr.network(), Network::Mainnet);
     assert_eq!(addr.to_string(), "f01");
 }
+
+#[test]
+fn invalid_strings_tests() {
+    let invalid_strings = &[
+        "fömk3zcefvlgpay4f32c5vmruk5gqig6dumc7pz6q",
+        "🗻∈🌏ömk3zcefvlgpay4f32c5vmruk5gqig6dumc7",
```
ec2, line 567: 🌏

PR #601, opened 2020-08-04
Update commcid to new codes and validation

austinabell:
**Summary of changes**
- Issue was just that in copying the buffered blockstore the cids with the new codes were not being ignored

cc @ec2 let me know if this solves

Closes 


austinabell:
Put back to draft because it is updated on `next` but not yet on their `master` branch and network we are interoping against

austinabell:
This change will be needed ~Wed Aug 19, so just opening so it can be reviewed and ready to come in when needed

PR #605, opened 2020-08-05
Adding Sync CLI Commands

### LINKED ISSUE #541: State CLI Commands
Implementing the State commands for our CLI client. 

Notes
We have to implement whatever we think is useful from the list below (likely 3 or more)

List of Commands:
power Query network or miner power
sectors Query the sector set of a miner
proving Query the proving set of a miner
pledge-collateral Get minimum miner pledge collateral
list-actors list all actors in the network
list-miners list all miners in the network
get-actor Print actor information
lookup Find corresponding ID address
replay Replay a particular message within a tipset
sector-size Look up miners sector size
read-state View a json representation of an actors state
list-messages list messages on chain matching given criteria
compute-state Perform state computations
call Invoke a method on an actor locally
get-deal View on-chain deal info
wait-msg Wait for a message to appear on chain
miner-info Retrieve miner information

nannick:
**Summary of changes**
- Changes start and end fields in SyncState to NaiveDateTime 
- Adds Deserialization for SyncState, changed Serialization for SyncState slightly
- Adds 5 commands for Sync CLI

One issue I'm running into is that after a while, the Lotus daemon no longer responds to any requests from our CLI.
Not sure if its something on our end, an issue with Lotus,  or something wrong with my local set up. 

**Sample Commands**
```
target/debug/forest sync wait
```
```
target/debug/forest sync status
```
```
target/debug/forest sync mark-bad "bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42a2"
```
```
target/debug/forest sync check-bad "bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42a2"
```
```
target/debug/forest sync submit  '{"Header":{"Miner":"t01234","Ticket":{"VRFProof":"Ynl0ZSBhcnJheQ=="},"ElectionProof":{"VRFProof":"Ynl0ZSBhcnJheQ=="},"BeaconEntries":null,"WinPoStProof":null,"Parents":null,"ParentWeight":"0","Height":10101,"ParentStateRoot":{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"},"ParentMessageReceipts":{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"},"Messages":{"/":"bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4"},"BLSAggregate":{"Type":2,"Data":"Ynl0ZSBhcnJheQ=="},"Timestamp":42,"BlockSig":{"Type":2,"Data":"Ynl0ZSBhcnJheQ=="},"ForkSignaling":42},"BlsMessages":null,"SecpkMessages":null}'
```

Closes 
#541 


forest/src/cli/sync_cmd.rs
```diff
@@ -0,0 +1,194 @@
...
+                        }
+
+                        if let Some(b) = &active_sync.target {
+                            target = Some(b.cids().to_vec());
+                            height_diff = b.epoch() - height_diff;
+                        } else {
+                            height_diff = 0;
+                        }
+
+                        println!("\tBase:\t{:?}\n", base.unwrap_or_default());
```
austinabell, line 87:
```suggestion
                        println!("\tBase:\n{:?}\n", base.unwrap_or_default());
```
I think you wanted a newline there, and are you sure you want the double newlines after each thing? (println adds one by default and you add `\n` at the end

forest/src/cli/sync_cmd.rs
```diff
@@ -0,0 +1,194 @@
...
+        block_cid: String,
+    },
+
+    #[structopt(
+        name = "submit",
+        about = "Submit newly created block to network through node"
+    )]
+    Submit {
+        #[structopt(short, long, help = "Gossip block as String argument")]
+        gossip_block: String,
```
austinabell, line 44: why is this not positional?

forest/src/cli/sync_cmd.rs
```diff
@@ -0,0 +1,194 @@
...
+        block_cid: String,
+    },
+
+    #[structopt(
+        name = "check-bad",
+        about = "Check if the given block was marked bad, and for what reason"
+    )]
+    CheckBad {
+        #[structopt(short, long, help = "Block Cid given as string argument")]
+        block_cid: String,
```
austinabell, line 35: this seems like it should be a positional arg as well (only one and would remove the need for -b or --block-cid flag)

forest/src/cli/sync_cmd.rs
nannick: good call theres only one argument makes more sense to just have a positional argument

blockchain/blocks/src/tipset.rs
```diff
@@ -312,13 +312,14 @@ pub mod tipset_json {
     {
         #[derive(Serialize, Deserialize)]
         #[serde(rename_all = "PascalCase")]
-        struct TipsetDe {
+        pub struct TipsetDe {
```
ec2, line 315: Does this need to be pub?

node/rpc-client/src/client.rs
```diff
@@ -27,6 +29,19 @@ jsonrpsee::rpc_api! {
...
+
+        /// Sync
+        #[rpc(method = "Filecoin.SyncState")]
+        fn status() -> RPCSyncState ;
+
+        #[rpc(method = "Filecoin.SyncMarkBad", positional_params)]
+        fn mark_bad( params : CidJson)  -> ();
+
+        #[rpc(method = "Filecoin.SyncCheckBad", positional_params)]
+        fn check_bad(params : CidJson)  -> String;
```
ec2, line 41: please fix the spacing here

blockchain/blocks/src/tipset.rs
nannick: Nope my bad forgot to turn it back

forest/src/cli/sync_cmd.rs
```diff
@@ -0,0 +1,198 @@
...
+            SyncCommand::Submit { gossip_block } => {
+                let response = submit_block(&mut client, gossip_block).await;
+                if response.is_ok() {
+                    println!("Successfully submitted block");
+                } else {
+                    println!(
+                        "Did not submit block because {:#?}",
+                        stringify_rpc_err(response.unwrap_err())
+                    );
+                }
```
austinabell, line 159: instead of doing `if ok {..} else {}` statements like this, you should match on the result to be able to log the error reason (and without unwrapping the error. (check-bad has no error message attached)

forest/src/cli/sync_cmd.rs
```diff
@@ -0,0 +1,188 @@
...
+                    .map_err(stringify_rpc_err)
+                    .unwrap();
+                println!("Block {} is \"{}\"", block_cid, statement);
+            }
+            SyncCommand::Submit { gossip_block } => {
+                submit_block(&mut client, gossip_block)
+                    .await
+                    .map_err(stringify_rpc_err)
+                    .unwrap();
+                println!("Successfully submitted block");
```
austinabell, line 149: this doesn't fail in any way when the request should fail (running without a node running), can you look into this?

forest/src/cli/sync_cmd.rs
```diff
@@ -0,0 +1,188 @@
...
+fn get_naive_time_zero() -> NaiveDateTime {
+    NaiveDate::from_ymd(1, 1, 1).and_hms(0, 0, 0)
+}
+
+impl SyncCommand {
+    pub async fn run(self) {
+        let mut client = new_client();
+
+        match self {
+            SyncCommand::Status {} => {
```
austinabell, line 63: This command also doesn't fail for any reason when there isn't a node running (nothing is output)

forest/src/cli/sync_cmd.rs
nannick: Yeah for RPC calls that dont really return any values, we shouldnt use () as the return value because those calls will always succeed for some reason. Solution I found was using a Option<()> as the return type

nannick:
Closing for now, will open up later

PR #606, opened 2020-08-05
Multisig actor update

### LINKED ISSUE #560: Update Multisig actor
**Task summary**

**Specification reference**

- 

austinabell:
**Summary of changes**
- Self explanatory. This one shouldn't be that bad to review

Closes #560 

### DIFF
--- a/vm/actor/src/builtin/multisig/mod.rs
+++ b/vm/actor/src/builtin/multisig/mod.rs
@@ -231,9 +231,13 @@ impl Actor {
                 return Err(actor_error!(ErrIllegalState; "hash does not match proposal params"));
             }
 
-            ptx.delete(&params.id.key()).map_err(
+            let deleted = ptx.delete(&params.id.key()).map_err(
                 |e| actor_error!(ErrIllegalState; "failed to delete pending transaction: {}", e),
             )?;
+            if !deleted {
+                return Err(actor_error!(ErrIllegalState;
+                    "failed to delete pending transaction: does not exist"));
+            }
 
             st.pending_txs = ptx.flush().map_err(
                 |e| actor_error!(ErrIllegalState; "failed to flush pending transactions: {}", e),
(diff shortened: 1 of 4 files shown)


vm/actor/src/builtin/multisig/mod.rs
```diff
@@ -146,36 +201,44 @@ impl Actor {
...
-                return Err(ActorError::new(
-                    ExitCode::ErrIllegalState,
-                    format!("Failed to delete transaction for cancel: {}", e),
-                ));
+            let calculated_hash = compute_proposal_hash(&tx, rt.syscalls()).map_err(|e| {
+                actor_error!(ErrIllegalState;
+                    "failed to compute proposal hash for (tx: {:?}): {}", params.id, e)
+            })?;
+
+            if params.proposal_hash != calculated_hash {
```
ec2, line 230: specs-actor also checks if proposal_hash is nil (im guessing that for us is if its 0 len?) and if it is, i think this check passes.

vm/actor/src/builtin/multisig/mod.rs
```diff
@@ -146,36 +201,44 @@ impl Actor {
...
+                return Err(actor_error!(ErrIllegalState; "hash does not match proposal params"));
             }
 
+            ptx.delete(&params.id.key()).map_err(
```
ec2, line 234: same in other PR, the case where key does not exist isnt handled

vm/actor/src/builtin/multisig/mod.rs
austinabell: nah the empty check is unnecessary because calculated_hash can never be nil. There is no case where that check changes anything.

Even in their code I'm quite sure the nil sanity check is irrelevant because their unmarshalling of those params initializes the proposal hash as bytes so it can't be nil.

vm/actor/src/builtin/multisig/mod.rs
```diff
@@ -292,105 +362,208 @@ impl Actor {
...
+            // Go implementation holds reference to state after transaction so this must be cloned
+            // to match to handle possible exit code inconsistency
+            Ok(st.clone())
+        })??;
+
+        execute_transaction_if_approved(rt, &st, tx_id, txn)
+    }
+}
+
+fn execute_transaction_if_approved<BS, RT>(
```
dutterbutter, line 421: This fn seems really weird to me, I know its 1:1 matching. It just seems weird that we would be returning an Ok exit code most of the time and wouldn't `applied` always be `true`?

austinabell: well the exit code is directly used in a serialized data structure, so I don't see the benefit in not returning the exit code (or how else would you pass that info back, because a result within a result doesn't seem like a good option.

And no applied is only true if the multisig threshold is met and that block of code gets executed

dutterbutter: I see, so if the multisig threshold is NOT met we still return an Ok exit code even though applied would be false?

austinabell: yeah, they leave as uninitialized, which is the 0 case (ExitCode::Ok) and return that (I think it's a bit weird that they don't make the exit code and return nullable, but I assume this may be more consistent and less likely to panic if things are changed)

PR #620, opened 2020-08-11
Authorization Setup for Write Access on RPC Calls

### LINKED ISSUE #592: Setup authorization for 'write' access RPC calls
**Task summary**

We should follow the same RPC authorization permissions as Lotus, which is all write related RPC calls should have an authorized token. Presently, we have no concept of authorization on any of our calls. 

**Specification reference**

- 

- [Lotus auth permissions](https://github.com/filecoin-project/lotus/blob/master/api/apistruct/permissioned.go#L27)

flodesi:
**Summary of changes**
- Create JWT write access to required RPC calls
- Implement auth commands to create and verify JWT tokens
    - This includes the implementing RPC and CLI commands for auth

Closes #592


forest/src/cli/auth_cmd.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub async fn run(&self) {
+        // TODO handle cli config
+        match self {
+            Self::CreateToken { perm } => {
+                let perm: String = perm.parse().unwrap();
+                let mut client = new_client();
+
+                let obj = auth_new(&mut client, perm)
+                    .await
+                    .map_err(stringify_rpc_err)
```
nannick, line 34: If you arent going to use the error message, its probably better to remove the mapping

flodesi: The error message is used, the create_token method returns a JWTResult which is then converted to a JsonRpcError in the event of Err. The Error message is used in the unwrap which shows up in the client CLI upon an error

nannick: Ahhh right because when it crashes the user will see the error, I see your point

utils/auth/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// TODO need to figure out how lotus generates secret key for encoding and decoding JWT Tokens
+
+/// Claim struct for JWT Tokens
+#[derive(Debug, Serialize, Deserialize)]
+struct Claims {
+    #[serde(rename = "Allow")]
+    allow: Vec<String>,
+    // TODO currently lotus does not have an exp value for their JWT tokens, need to figure out what they do instead to overcome invalid validations
```
ec2, line 50: Is there a way to make it not expire instead of setting a super high value?

utils/auth/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+        &payload,
+        &EncodingKey::from_secret("secret".as_ref()),
+    )
+}
+
+/// Verify JWT Token and return the allowed permissions from token
+pub fn verify_token(token: &str) -> JWTResult<Vec<String>> {
+    let token = decode::<Claims>(
+        token,
+        &DecodingKey::from_secret("secret".as_ref()),
```
ec2, line 72: ??? Doesnt look like you actually have a secret saved... Just using the string "secret"

forest/src/daemon.rs
```diff
@@ -38,9 +39,12 @@ pub(super) async fn start(config: Config) {
         });
 
     // Initialize keystore
-    let keystore = Arc::new(RwLock::new(
-        PersistentKeyStore::new(config.data_dir.to_string()).unwrap(),
-    ));
+    let mut ks = PersistentKeyStore::new(config.data_dir.to_string()).unwrap();
+    if ks.get("auth-jwt-private").is_err() {
```
nannick, line 43: I think it would be better to make  "auth-jwt-private" a constant somewhere in auth/src/lib.rs instead of reusing it multiple times

forest/src/cli/auth_cmd.rs
```diff
@@ -0,0 +1,40 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
```
austinabell, line 6:
```suggestion
```

node/rpc/src/lib.rs
```diff
@@ -1,20 +1,38 @@
...
+lazy_static! {
+    pub static ref WRITE_ACCESS: Vec<String> = vec![
+        "Filecoin.MpoolPush".to_string(),
+        "Filecoin.WalletNew".to_string(),
+        "Filecoin.WalletHas".to_string(),
+        "Filecoin.WalletList".to_string(),
+        "Filecoin.WalletDefaultAddress".to_string(),
+        "Filecoin.WalletList".to_string(),
+    ];
+}
```
austinabell, line 35: This does not have to be lazy static, can just be an array of static strings (no need for allocations for the vec or strings

utils/auth/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+        "read".to_string(),
+        "write".to_string(),
+        "sign".to_string(),
+        "admin".to_string()
+    ];
+    pub static ref SIGN: Vec<String> =
+        vec!["read".to_string(), "write".to_string(), "sign".to_string()];
+    pub static ref WRITE: Vec<String> = vec!["read".to_string(), "write".to_string()];
+    pub static ref READ: Vec<String> = vec!["read".to_string()];
+}
```
austinabell, line 44: Same here, none of this needs to be allocated, and why are they kept seperate

utils/auth/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+            return Err(JsonRpcError::from(Error::InvalidPermissions));
+        }
+    }
+    Ok(())
+}
+
+pub fn generate_priv_key() -> KeyInfo {
+    let priv_key = rand::thread_rng().gen::<[u8; 32]>();
+    // TODO temp use of bls key as placeholder, need to update keyinfo to use string instead of keyinfo
+    // for key type
```
austinabell, line 88: Why would string be needed? I don't see how that helps this

austinabell:
I definitely want to think about this more in depth soon because I don't see how this helps us or even has support for multiple tokens (did I miss that part in this PR?)

Also, can you please try running our node and testing functionally, I am getting this error when running and stopping the node:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: IO("File exists (os error 17)")', forest/src/daemon.rs:131:9
```

utils/auth/src/lib.rs
flodesi: yeah, they don't need to be separate, was just matching lotus

utils/auth/src/lib.rs
flodesi: In lotus the key_info is not necesarily either BLS or SECP256K1, so switching to a string key_type would allow storing more types of keys (like the JWT secret key) rather than just BLS or SECP256K1.

flodesi:
> I definitely want to think about this more in depth soon because I don't see how this helps us or even has support for multiple tokens (did I miss that part in this PR?)

This PR basically just lays the ground work for adding privileges on certain RPC calls, just like how lotus has set up. as for multiple tokens, the format is the same as with lotus, each token is identical for each permission level, as long as the private JWT secret key is deleted.

utils/auth/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+/// constant string that is used to identify the JWT secret key in KeyStore
+pub const JWT_IDENTIFIER: &str = "auth-jwt-private";
+/// Admin permissions
+pub const ADMIN: [&str; 4] = ["read", "write", "sign", "admin"];
+/// Signing permissions
+pub const SIGN: [&str; 3] = ["read", "write", "sign"];
+/// Writing permissions
+pub const WRITE: [&str; 2] = ["read", "write"];
+/// Reading permissions
+pub const READ: [&str; 1] = ["read"];
```
austinabell, line 22: This could be a lot better (enums with a value of priority or something) but can always change later I guess

austinabell (approved):
utack

utils/auth/src/lib.rs
AshantiMutinta: Yeah I think this was from Jaden's existing implementation. I think can be just changed later for now

PR #622, opened 2020-08-12
Update system actor

### LINKED ISSUE #564: Update system actor
**Task summary**

**Specification reference**

- 

austinabell:
**Summary of changes**
- ..

Closes #564 


vm/actor/src/builtin/system/mod.rs
```diff
@@ -3,30 +3,41 @@
...
 #[repr(u64)]
 pub enum Method {
     Constructor = METHOD_CONSTRUCTOR,
 }
 
-/// Init actor
+/// System actor state.
+#[derive(Default, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct State([(); 0]);
```
nannick, line 26: This is just for me but why us State([(); 0]) instead of something like State{}

austinabell: Because that would serialize as an empty map, when it needs to serialize as an empty array (different bytes in Cbor)

PR #623, opened 2020-08-13
Update to new empty amt serialization

austinabell:
**Summary of changes**
- ref to change: https://github.com/filecoin-project/go-amt-ipld/commit/c74824dd4dfd02aee0588105471b597e794cb761
- Also removed other unnecessary allocations (previously assumed with serde you needed allocations for deserializing bytes and strings)

Closes 


crypto/src/vrf.rs
```diff
@@ -49,7 +50,9 @@ pub mod json {
     where
         D: Deserializer<'de>,
     {
-        let s: String = Deserialize::deserialize(deserializer)?;
-        Ok(VRFProof::new(base64::decode(s).map_err(de::Error::custom)?))
+        let s: Cow<'de, str> = Deserialize::deserialize(deserializer)?;
```
ec2, line 53: Cow!!!!!

utils/bigint/src/bigint_ser.rs
```diff
@@ -51,5 +51,5 @@ where
             ));
         }
     };
-    Ok(BigInt::from_bytes_be(sign, &bz))
+    Ok(BigInt::from_bytes_be(sign, &bz[1..]))
```
ec2, line 54: What happens if there's only a sign byte? This will panic i think, but what is the expected behaviour? I would assume 0 or something, since empty bytes -> bigint deserializes to 0?

utils/bigint/src/biguint_ser.rs
```diff
@@ -35,16 +35,16 @@ pub fn deserialize<'de, D>(deserializer: D) -> Result<BigUint, D::Error>
...
 
-    if bz.remove(0) != 0 {
+    if bz.get(0) != Some(&0) {
         return Err(serde::de::Error::custom(
             "First byte must be 0 to decode as BigUint",
         ));
     }
 
-    Ok(BigUint::from_bytes_be(&bz))
+    Ok(BigUint::from_bytes_be(&bz[1..]))
```
ec2, line 49: Same thing here

utils/bigint/src/bigint_ser.rs
austinabell: on the 0 case, it is serialized as empty bytes, even if it was a sign byte and nothing else, this would just deserialize as zero. Also the functionality is basically the same as it was before.

utils/bigint/src/biguint_ser.rs
austinabell: empty bytes get serialized as 0 and this would still not have an issue (but again Lotus and our client would not ever serialize as this case)

PR #604, opened 2020-08-05
Add Persistent KeyStore

flodesi:
**Summary of changes**
Changes introduced in this pull request:
- Creation of a persistent keystore: PersistentKeyStore
- This file writes keys to the file keystore.json

**Reference issue to close (if applicable)**

Closes #547

PR #625, opened 2020-08-13
Fix bug in reading and persisting keystore data

austinabell:
**Summary of changes**
- After #604 file was created on initializing a keystore, but was never written. When the file was attempted to be read, if the bytes were invalid, the node would error before starting.
  - I was going to make the file read recoverable, but in the case of a file not being fully written and has keys that are wanted, figured overwriting shouldn't be assumed

Closes 


forest/src/daemon.rs
```diff
@@ -123,12 +123,17 @@ pub(super) async fn start(config: Config) {
     // Block until ctrl-c is hit
     block_until_sigint().await;
 
+    let keystore_write = task::spawn(async move {
+        keystore.read().await.flush().unwrap();
+    });
+
```
flodesi, line 129: Is there a reason why keystore_write.await isn't put directly below this chunk of code?
```suggestion
    let keystore_write = task::spawn(async move {
        keystore.read().await.flush().unwrap();
    });
    keystore_write.await;
```

austinabell: because I don't want the main thread to wait to cancel all tasks before awaiting on the keystore to be written, but I also don't want to wait to start the write until after the tasks are cancelled

PR #635, opened 2020-08-17
Make hamt value type generic and add benchmarks

### LINKED ISSUE #202: Add benchmarks to AMT/HAMT
**Issue summary**

AMT implementation is coming in with #197 but I did not include benchmarks. I will most likely do this along with benchmarking HAMT, but anyone could pick this up if they would like.

Most important one to benchmark is the batch set function, or the new from slice as this is the primary use and need for it in the code as of now.

### LINKED ISSUE #391: Optimize HAMT to not deserialize values internally
**Issue summary**

This was setup as it to be more safe (and was just an easy transition going from the typesafe version to allowing any data to be saved). 

Since the values of the HAMT can be anything, it doesn't really add much benefit to deserialize intermediate values into Ipld and will have a cost in the long term. The reason specifically this is costly is because on every node read, all values will be deserialized into IPLD from bytes and then the `Ipld` -> `T` based on the get will also go through a byte buffer because of the current state of Ipld. 

The solution is just to deserialize bytes at the Hamt API level so for every read/ write the data is only serialized or deserialized once. This will require the BlockStore trait to have a raw put function, because there would be no way to know what the type of the bytes are.

This is not objectively better and has some tradeoffs (like having to build the cbor encoded bytes payload manually and technically more heap allocation due to all bytes being put on the heap) so would be worth it to setup benchmarks #202 before looking into this in detail.

austinabell:
**Summary of changes**
- Seems like this is the best solution for #391. Previously, there was multiple value types in one hamt, but that is not the case anymore it seems and we cannot leave bytes serialized when using Serde (go implementation uses a deferred type for this which avoids deserializing until needed but this is almost impossible without replacing serde for serialization for the hamt).
- Adds benchmarks to the hamt

This will be pointed at hashing branch until #624 comes in (didn't want to deal with conflicts) and I will open when it's pointing back at main

benchmarks:

```
HAMT bulk insert (no flush)
                        time:   [39.967 us 40.263 us 40.560 us]
                        change: [-63.784% -63.178% -62.603%] (p = 0.00 < 0.05)
                        Performance has improved.

HAMT bulk insert with flushing and loading
                        time:   [2.2143 ms 2.2285 ms 2.2429 ms]
                        change: [-65.130% -64.540% -64.033%] (p = 0.00 < 0.05)
                        Performance has improved.

HAMT deleting all nodes time:   [75.801 us 76.273 us 76.734 us]
                        change: [-71.930% -71.549% -71.206%] (p = 0.00 < 0.05)
                        Performance has improved.

HAMT for_each function  time:   [58.529 us 59.057 us 59.562 us]
                        change: [-84.056% -83.806% -83.585%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Closes #202 
Closes #391 


ipld/hamt/Cargo.toml
```diff
@@ -29,3 +29,9 @@ murmur = ["murmur3"]
 
 [dev-dependencies]
 hex = "0.4.2"
+criterion = "0.3.1"
```
dutterbutter, line 32: Not that is matters but the latest version `0.3.3` I believe.

austinabell: patch versions get automatically bumped to the most up to date (if you look inside of the `Cargo.lock` you'll see the version is actually 0.3.3, but I will update this for peace of mind

PR #637, opened 2020-08-18
Separate ticket and beacon randomness

### LINKED ISSUE #628: Update Randomness to distinguish between Drand & Ticket Chain
See the Lotus PR: https://github.com/filecoin-project/specs-actors/pull/679

austinabell:
**Summary of changes**
- Not updating specs-actors to use the two distinct yet as I will do that when I update actors, but this sets up and fixes bugs to allow it to be used.

other commit not referenced in issue:
https://github.com/filecoin-project/lotus/commit/ed72aa1307e312a9dc570b3dc438f3d5fae9c064

Closes #628 


blockchain/chain/src/store/chain_store.rs
```diff
@@ -300,23 +280,85 @@ where
...
-pub fn get_randomness<DB: BlockStore>(
+/// Finds the latest beacon entry given a tipset up to 20 blocks behind
+pub fn latest_beacon_entry<DB>(db: &DB, ts: &Tipset) -> Result<BeaconEntry, Error>
+where
+    DB: BlockStore,
+{
+    let check_for_beacon_entry = |ts: &Tipset| {
+        let cbe = ts.blocks()[0].beacon_entries();
+        if let Some(entry) = cbe.last() {
+            return Ok(Some(entry.clone()));
```
timvermeulen, line 291: You could call `cbe.pop()` here instead of `last()` to avoid a clone

blockchain/chain/src/store/chain_store.rs
```diff
@@ -300,23 +280,85 @@ where
...
+    if let Some(entry) = check_for_beacon_entry(ts)? {
+        return Ok(entry);
+    }
+    let mut cur = tipset_from_keys(db, ts.parents())?;
+    for _ in 1..20 {
+        if let Some(entry) = check_for_beacon_entry(&cur)? {
+            return Ok(entry);
+        }
+        cur = tipset_from_keys(db, cur.parents())?;
+    }
```
timvermeulen, line 310: Pretty minor, you could move the `cur = tipset_from_keys(db, cur.parents())?;` line to the top of the loop and initialise `cur` to just `ts`, this would save one `tipset_from_keys` call

austinabell: nah, because cur is owned, and ts is not. It looks a bit weird but cloning a tipset isn't trivial

blockchain/chain/src/store/chain_store.rs
austinabell: it's a slice, not a vec

timvermeulen: 😬

Issue #634, opened 2020-08-17
Set up Test Runner for new Test Vectors


amerameen:
Vectors: https://github.com/filecoin-project/test-vectors

Lotus test runner: https://github.com/filecoin-project/lotus/pull/3081

PR #638, opened 2020-08-19
Conformance test runner

austinabell:
**Summary of changes**
- Runner based on https://github.com/filecoin-project/lotus/pull/3081 and #634 
- Slight adjustments to bulk write db trait
- Rand switched into trait to be able to be swapped out in VM
- Updates car functionality
- Put buffered blockstore behind feature (remove need for commcid and fil proofs to be brought into certain tests)

Currently blocked on:
- dynamic gas being implemented (unsigned message format changes)
- Randomness split changes #637 
- All other major changes coming in next reset, such as: #624 (These will just affect test failures)

Closes #637


ec2:
This shouldnt be blocked anymore, right?

austinabell:
> This shouldnt be blocked anymore, right?

I'm making changes now and testing to see how many I can get passing, but will open up soon (also want to update to most recent test-vectors commit)

austinabell:
> Looks good! One thing that we will want is the ability to run tests selectively in the case of specific test case failures, but we can add that in later.

How do you expect this to happen? The test cases are based on the submodule json files so I don't see how you do this with rust's test framework. Are you suggesting an env variable with some regex for tests to run? The regex skipping system is a bit clunky because the test file has to be recompiled, but I don't see the problem with using it the way it is to do this

PR #639, opened 2020-08-19
Dynamic Gas Implementation

ec2:
**Summary of changes**
- Implemented the math for calculated base fees
- Added new Gas api calls
- Added base fee verification of the block header
- Updated messages fields

TODO: Fix tests

Closes 


blockchain/chain/src/store/base_fee.rs
```diff
@@ -0,0 +1,75 @@
...
+    next_base_fee
+}
+
+pub fn compute_base_fee<DB>(db: &DB, ts: &Tipset) -> Result<BigInt, crate::Error>
+where
+    DB: BlockStore,
+{
+    let mut total_limit = 0;
+    for b in ts.blocks() {
+        let (msg1, msg2) = crate::block_messages(db, &b)?;
```
AshantiMutinta, line 34: it would be nice to have a function like block_messages which returns ChainMessage because this is one of the cases where we don't care if it is a signed or unsigned message that is returned. We are just worrying about the gas_limit on that particular message.

blockchain/chain/src/store/base_fee.rs
```diff
@@ -0,0 +1,75 @@
...
+        for m in msg1 {
+            total_limit += m.gas_limit();
+        }
+        for m in msg2 {
+            total_limit += m.gas_limit();
+        }
+    }
+    let parent_base_fee = ts.blocks()[0].parent_base_fee();
```
AshantiMutinta, line 42: this might panic if blocks() is empty, do we need to safeguard against that?

node/rpc/src/gas_api.rs
```diff
@@ -0,0 +1,189 @@
...
+    let from_a = data
+        .state_manager
+        .resolve_to_key_addr(msg.from(), &curr_ts)
+        .await?;
+
+    let pending = data.mpool.pending_for(&from_a).await;
+    let prior_messages: Vec<ChainMessage> = match pending {
+        Some(messages) => messages.into_iter().map(ChainMessage::Signed).collect(),
+        None => vec![],
+    };
```
AshantiMutinta, line 171: I don't know how you feel about this but it could also be written as 

pending.map(|s| s.into_iter().map(ChainMessage::Signed).collect()).unwrap_or_default()

blockchain/state_manager/src/lib.rs
```diff
@@ -162,6 +162,7 @@ where
         &self,
         ts: &FullTipset,
         rand: &ChainRand,
+        base_fee: &BigInt,
```
austinabell, line 165: Why ref if it's just going to be cloned?

node/rpc/src/gas_api.rs
```diff
@@ -0,0 +1,189 @@
...
+    let res = data
+        .state_manager
+        .call_with_gas(
+            &mut msg,
+            &prior_messages,
+            Some(data.mpool.cur_tipset.as_ref().read().await.clone()),
+        )
+        .await?;
+    match res.msg_rct.clone() {
```
austinabell, line 180: shouldn't need to clone, can prob just take a ref

node/rpc/src/gas_api.rs
```diff
@@ -0,0 +1,189 @@
...
+        Some(rct) => {
+            if rct.exit_code as u64 != 0 {
+                return Ok(-1);
+            }
+        }
+        None => return Ok(-1),
+    }
+    Ok(res.msg_rct.unwrap().gas_used)
```
austinabell, line 188: why not just have this line in the match to avoid calling return and unwrapping?

vm/interpreter/src/vm.rs
```diff
@@ -397,12 +434,88 @@ where
...
+    let mut miner_tip = gas_premium;
+    if &(base_fee_to_pay + &miner_tip) > fee_cap {
+        miner_tip = fee_cap - base_fee_to_pay;
+    }
+    out.miner_tip = &miner_tip * gas_limit;
+    let (out_gas_refund, out_gas_burned) = compute_gas_overestimation_burn(gas_used, gas_limit);
+    out.gas_refund = out_gas_refund;
+    out.gas_burned = out_gas_burned;
+
+    if out.gas_burned == 0 {
```
austinabell, line 473:
```suggestion
    if out.gas_burned != 0 {
```

vm/interpreter/src/vm.rs
```diff
@@ -397,12 +434,88 @@ where
...
+fn compute_gas_outputs(
+    gas_used: i64,
+    gas_limit: i64,
+    base_fee: &TokenAmount,
+    fee_cap: &TokenAmount,
+    gas_premium: TokenAmount,
+) -> GasOutputs {
+    let gas_used_big: BigInt = 0.into();
+    let mut base_fee_to_pay = base_fee;
+    let mut out = GasOutputs::default();
```
austinabell, line 458: Opinionated but seems like it would be a bit cleaner to construct all variables, then just construct the `GasOutputs` type for the return

vm/interpreter/src/vm.rs
```diff
@@ -397,12 +434,88 @@ where
...
+    if over > gas_used {
+        over = gas_used;
+    }
+
+    let mut gas_to_burn: BigInt = (gas_limit - gas_used).into();
+    gas_to_burn *= over;
+    gas_to_burn /= gas_used;
+
+    (
+        i64::try_from(gas_limit - gas_used - gas_to_burn.clone()).unwrap(),
```
austinabell, line 507:
```suggestion
        i64::try_from(gas_limit - gas_used - &gas_to_burn).unwrap(),
```

blockchain/chain/src/store/base_fee.rs
ec2: i think it should be fine. Tipsets have at least one block by construction

node/rpc/src/gas_api.rs
ec2: love this suggestion. thanks!

PR #643, opened 2020-08-21
Builtin actors 0.9.3 update

austinabell:
**Summary of changes**
- Updates all builtin actors to version 0.9.3

Closes 


vm/actor/src/builtin/shared.rs
```diff
@@ -25,11 +26,50 @@ where
...
+    )
+    .map_err(|e| {
+        e.wrap(&format!(
+            "failed to send zero balance to address {}",
+            address
+        ))
+    })?;
+
+    match rt.resolve_address(address)? {
+        Some(addr) => Ok(addr),
```
nannick, line 70: Something like this might be better

match rt.resolve_address(address).ok_or_else(|| "failed to resolve address {} to ID address even after sending zero balance")?

austinabell: ah yeah, that was intentional, but yeah this is probably more idiomatic. (I also realized I forgot to use the `{}` to place in the address lmao)

PR #647, opened 2020-08-24
Adding Mdns and Kad Toggle

### LINKED ISSUE #502: Allow mdns/kad to be toggled off
**Issue summary**

A CLI option to disable mdns would be useful to have for certain testing. Currently is always enabled, but a boolean flag can be added to the `Config` type to not start mdns. 

Optionally the same can be done with kademlia discovery (and maybe there is a reason to do that as well)

To make the behaviour optional, can wrap in a `Toggle` type from rust-libp2p and just have set as None when flag isn't set (default for both should be set though)

nannick:
**Summary of changes**
- Adds Toggle wrapper around mdns and kad fields in ForestBehaviour struct and the necessary minor fixes

 Sample Config File : 
data_dir = "test_dir"

[network]
listening_multiaddr = "/ip4/0.0.0.0/tcp/40000"
bootstrap_peers = ["/ip4/54.186.82.90/tcp/1347"]
mdns = false
kademlia = false

Closes 
#502 


ec2:
I'm confused by the tags haha. It's currently do not merge? Is there a reason you didnt open a draft PR? Or did you mislabel accidentally?

nannick:
> I'm confused by the tags haha. It's currently do not merge? Is there a reason you didnt open a draft PR? Or did you mislabel accidentally?

I was debating whether to hold off on submitting this because our daemon doesnt start up so actual testing might've been difficult, which is why I added those tags. Wasn't sure what the best tag was for it lol. Logically though its pretty simple and makes sense which is why I ended up submitting it anyway

node/forest_libp2p/src/config.rs
```diff
@@ -18,6 +18,8 @@ const DEFAULT_BOOTSTRAP: &[&str] = &[
 pub struct Libp2pConfig {
     pub listening_multiaddr: Multiaddr,
     pub bootstrap_peers: Vec<Multiaddr>,
+    pub mdns: bool,
+    pub kad: bool,
```
austinabell, line 22: For consistency? or can just serde rename
```suggestion
    pub kademlia: bool,
```

PR #650, opened 2020-08-25
Space race genesis and bootnodes updates

### LINKED ISSUE #648: run failed
2020-08-25T01:01:48.458Z INFO  forest::daemon > Starting Forest daemon
 2020-08-25T01:01:48.458Z INFO  forest_libp2p::service > Recovered libp2p keypair from "/home/ubuntu/.forest/libp2p/keypair"
 2020-08-25T01:01:48.459Z WARN  key_management::keystore > keystore.json does not exist, initializing new keystore
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorImpl { code: Message("invalid type: byte array, expected i64"), offset: 0 }', forest/src/daemon.rs:54:68

austinabell:
**Summary of changes**
- Updates genesis and bootnodes
- Fix bug in car file reader (there was a check for if the buffer is empty, which could exit when the buffer is empty but there are still items to read)
- Update drand config to valid beacon
- Update gossipsub config (default config is explicitly disallowed with new libp2p version)

Seems kademlia is not bootstrapping as it should, and sync is not initializing as it was before, but I'm just doing minimum changes to fix #648 and setup a functional node

Closes #648 


forest/src/daemon.rs
```diff
@@ -72,9 +73,14 @@ pub(super) async fn start(config: Config) {
     let coeff = config.drand_public;
 
     // TODO: Interval is supposed to be consistent with fils epoch interval length, but not yet defined
-    let beacon = DrandBeacon::new(coeff, genesis.blocks()[0].timestamp(), 1)
-        .await
-        .unwrap();
+    let beacon = DrandBeacon::new(
+        "https://api.drand.sh",
```
nannick, line 77: This is used in 2 places might be better to make it a constant

austinabell: yeah I'll do that, was just being lazy since the one is just used for tests that are ignored by default

PR #653, opened 2020-08-26
Adding Chaos Actor

### LINKED ISSUE #640: Implement Chaos Actor (for Test Vectors)
This is an actor that is used by the test vectors to trigger violations

See https://github.com/filecoin-project/test-vectors/pull/39 & https://github.com/filecoin-project/test-vectors/issues/37

nannick:
**Summary of changes**
- Adds the Chaos Actor
- Just one little change regarding cid.UNDEF and address.UNDEF needs to be made before its ready for review

Closes 
#640 


vm/actor/src/builtin/chaos/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+        panic!("Constructor should not be called");
+    }
+
+    // CallerValidation violates VM call validation constraints.
+    //
+    //  CALLER_VALIDATION_BRANCH_NONE performs no validation.
+    //  CALLER_VALIDATION_BRANCH_TWICE validates twice.
+    //  CALLER_VALIDATION_BRANCH_ADDR_NIL_SET validates against an empty caller
+    //  address set.
+    //  CALLER_VALIDATION_BRANCH_TYPE_NIL_SET validates against an empty caller type set.
```
austinabell, line 53: make this doc comments `///`

vm/actor/src/builtin/chaos/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+use super::{ACCOUNT_ACTOR_CODE_ID, SYSTEM_ACTOR_ADDR};
+use ipld_blockstore::BlockStore;
+use num_bigint::BigInt;
+use num_derive::FromPrimitive;
+use num_traits::FromPrimitive;
+use runtime::{ActorCode, Runtime};
+pub use state::*;
+pub use types::*;
+use vm::{actor_error, ActorError, ExitCode, MethodNum, Serialized, METHOD_CONSTRUCTOR};
+// * Updated to specs-actors commit: 4784ddb8e54d53c118e63763e4efbcf0a419da28
```
austinabell, line 16: this doesn't come from specs-actors (assume this was copied)

vm/actor/src/builtin/chaos/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+            x if x == *CALLER_VALIDATION_BRANCH_NONE => {}
+            x if x == *CALLER_VALIDATION_BRANCH_TWICE => {
+                rt.validate_immediate_caller_accept_any()?;
+                rt.validate_immediate_caller_accept_any()?;
+            }
+            x if x == *CALLER_VALIDATION_BRANCH_ADDR_NIL_SET => {
+                rt.validate_immediate_caller_is(vec![])?;
+            }
+            x if x == *CALLER_VALIDATION_BRANCH_TYPE_NIL_SET => {
+                rt.validate_immediate_caller_type(vec![])?;
```
austinabell, line 69:
```suggestion
                rt.validate_immediate_caller_is(&[])?;
            }
            x if x == *CALLER_VALIDATION_BRANCH_TYPE_NIL_SET => {
                rt.validate_immediate_caller_type(&[])?;
```
Kinda a nit because an empty vec doesn't get allocated, but this is just simpler

vm/actor/src/builtin/singletons.rs
```diff
@@ -13,8 +13,10 @@ lazy_static! {
     pub static ref STORAGE_MARKET_ACTOR_ADDR: Address = Address::new_id(5);
     pub static ref VERIFIED_REGISTRY_ACTOR_ADDR: Address = Address::new_id(6);
 
-    // Distinguished AccountActor that is the destination of all burnt funds.
+    pub static ref CHAOS_ACTOR_ADDR: Address    = Address::new_id(97);
     pub static ref PUPPET_ACTOR_ADDR: Address    = Address::new_id(98);
+
+    // Distinguished AccountActor that is the destination of all burnt funds.
```
austinabell, line 19:
```suggestion
    /// Distinguished AccountActor that is the destination of all burnt funds.
```

vm/actor/src/builtin/chaos/state.rs
```diff
@@ -0,0 +1,38 @@
...
+        D: Deserializer<'de>,
+    {
+        Err(de::Error::custom(
+            "Automatic fail when deserializing UnmarshallableCBOR",
+        ))
+    }
+}
+
+impl Cbor for UnmarshallableCBOR {}
+
```
austinabell, line 35: Why not just use a shared structure (same as from puppet actor?)

Also, just a slight pref thing, but all of these test actors could be placed in their own folder/module as to not be grouped with the actual builtin actors (right now they are in the builtin directory but puppet and chaos could be in utils or just in a seperate module). This part can be ignored now though

nannick: > Why not just use a shared structure (same as from puppet actor?)
> 
> Also, just a slight pref thing, but all of these test actors could be placed in their own folder/module as to not be grouped with the actual builtin actors (right now they are in the builtin directory but puppet and chaos could be in utils or just in a seperate module). This part can be ignored now though

I think it makes sense to move it now instead of putting it off for later. If I am moving it tho, should the constants like CHAOS_ACTOR_ADDR still be in vm/actor/src/builtin or should they be moved as well?

utils/bigint/Cargo.toml
```diff
@@ -8,6 +8,6 @@ edition = "2018"
 repository = "https://github.com/ChainSafe/forest"
 
 [dependencies]
-num-bigint = "0.3"
+num-bigint = { version = "0.3",  features = ["serde"] }
```
austinabell, line 11: nooo don't do this, this is the wrong serialization format and will trip up someone next if you leave it in

vm/actor/src/builtin/chaos/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // CallerValidation violates VM call validation constraints.
+    //
+    //  CALLER_VALIDATION_BRANCH_NONE performs no validation.
+    //  CALLER_VALIDATION_BRANCH_TWICE validates twice.
+    //  CALLER_VALIDATION_BRANCH_ADDR_NIL_SET validates against an empty caller
+    //  address set.
+    //  CALLER_VALIDATION_BRANCH_TYPE_NIL_SET validates against an empty caller type set.
+    pub fn caller_validation<BS, RT>(rt: &mut RT, branch: BigInt) -> Result<(), ActorError>
```
austinabell, line 54: wrap the bigint in the `BigIntDe` wrapper structure we have (this is incorrect serialization format)

Alternatively you can just wrap the deserialization in the `invoke_method` call (which I'm thinking might be a better solution anyways)

vm/actor/src/builtin/chaos/state.rs
austinabell: I think it makes sense to keep the addr and code cids together, no need to move those out as well, unless you want to put them in the same module as each actor themselves (up to you!)

utils/bigint/Cargo.toml
nannick: lol thought it was weird we didnt use the serde feature for BigInt anywhere my bad

vm/actor/src/util/mod.rs
```diff
@@ -2,11 +2,14 @@
...
 
 mod balance_table;
+pub mod chaos;
 pub mod math;
 mod multimap;
+pub mod puppet;
 mod set;
 mod set_multimap;
 pub mod smooth;
+pub mod unmarshallable;
```
austinabell, line 12: does this need to be public? Seems it's only used within the util module

vm/actor/src/util/unmarshallable.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
```
austinabell, line 6:
```suggestion
```

vm/actor/src/util/mod.rs
```diff
@@ -9,9 +9,10 @@ pub mod puppet;
...
 mod set_multimap;
 pub mod smooth;
-pub mod unmarshallable;
+mod unmarshallable;
 
 pub use self::balance_table::BalanceTable;
 pub use self::multimap::*;
 pub use self::set::Set;
 pub use self::set_multimap::SetMultimap;
+pub use unmarshallable::*;
```
austinabell, line 18: no I didn't mean making the type public, this type should not be exposed at all

nannick: Its used in the puppet_actor_test for the serialization test i believe

austinabell: ah, right. This test probably shouldn't live there. The `tests` directory should just be used to test publicly exposed functions. Something like that could just live as a test module within the code (around where the Unmarshallable struct is defined).

Probably shouldn't expose a type like that just so that it can be tested somewhere else

vm/actor/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 vm = { package = "forest_vm", path = "../../vm" }
 address = { package = "forest_address", path = "../address" }
 runtime = { path = "../runtime" }
-num-bigint = { path = "../../utils/bigint", package = "forest_bigint" }
+num-bigint = { path = "../../utils/bigint", package = "forest_bigint"}
```
austinabell, line 11:
```suggestion
num-bigint = { path = "../../utils/bigint", package = "forest_bigint" }
```

vm/actor/src/util/chaos/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+        RT: Runtime<BS>,
+    {
+        // TODO Temporarily fine to use default as Undefined Cid, but may need to change in the future
+        let actor_cid = if arg.undef_cid {
+            Cid::default()
+        } else {
+            arg.cid
+        };
+
+        //TODO This may need to change to match address.Undef in Spec-actors
```
austinabell, line 91:
```suggestion
        // TODO This may need to change to match address.Undef in Spec-actors
```

vm/actor/src/util/unmarshallable.rs
```diff
@@ -0,0 +1,60 @@
...
+use encoding::Cbor;
+use serde::de::{self, Deserializer};
+use serde::ser::{self, Serializer};
+use serde::{Deserialize, Serialize};
+
+#[derive(Default)]
+pub struct UnmarshallableCBOR {}
```
austinabell, line 10:
```suggestion
pub struct UnmarshallableCBOR;
```

Doesn't need to be empty struct form, but doesn't matter. (also don't really need the default impl, because there is no data)

PR #654, opened 2020-08-26
Allow registering of Actors to the VM

### LINKED ISSUE #646: Allow VM to register custom actors in VM
**Task summary**

Code is setup to match on the cid to dispatch to builtin actors, and this will not change but something should be added to the vm structure such as a HashMap to allow registering other actors to check after the builtin match (not all should be in a hashmap to avoid having to lookup in that hashmap for builtin successful calls.

This is just used for test vectors to allow the vm to register puppet actor (#566) and chaos actor (#640)

**Specification reference**

- 

ec2:
**Summary of changes**
- Updates conformance tests to register puppet and chaos actor for testing
- Adds custom actors registration to the VM

Closes 
#646 

Blocked by 
#640


tests/conformance_tests/tests/conformance_runner.rs
```diff
@@ -26,7 +27,7 @@ use vm::{ExitCode, Serialized, TokenAmount};
 use walkdir::{DirEntry, WalkDir};
 
 lazy_static! {
-    static ref SKIP_TESTS: [Regex; 7] = [
+    static ref SKIP_TESTS: [Regex; 0] = [
```
austinabell, line 30: you need to change the value here or it will fail to compile

tests/conformance_tests/tests/conformance_runner.rs
```diff
@@ -238,6 +248,24 @@ fn execute_message(
...
+        {
+            vm.register_actor(PUPPET_ACTOR_CODE_ID.clone());
+        }
+        if s.chaos_actor
+            .as_ref()
+            .map(|s| s == "true")
+            .unwrap_or_default()
+        {
+            // TODO
+            // vm.register_actor(code_cid)
```
austinabell, line 265: chaos actor is in now

vm/actor/src/builtin/codes.rs
```diff
@@ -22,7 +22,7 @@ lazy_static! {
         [ACCOUNT_ACTOR_CODE_ID.clone(), MULTISIG_ACTOR_CODE_ID.clone()];
 }
 
-fn make_builtin(bz: &[u8]) -> Cid {
+pub fn make_builtin(bz: &[u8]) -> Cid {
```
austinabell, line 25:
```suggestion
fn make_builtin(bz: &[u8]) -> Cid {
```

vm/interpreter/src/default_runtime.rs
```diff
@@ -730,7 +734,19 @@ where
...
+                return match x {
+                    x if x == *PUPPET_ACTOR_CODE_ID => {
+                        puppet::Actor.invoke_method(rt, method_num, params)
+                    }
+                    _ => Err(
+                        actor_error!(SysErrorIllegalActor; "no code for actor at address {}", to),
+                    ),
+                };
+            }
+            Err(actor_error!(SysErrorIllegalActor; "no code for actor at address {}", to))
```
austinabell, line 748: if you just put this in an `else`, you don't need to call return above

vm/interpreter/src/default_runtime.rs
```diff
@@ -730,7 +734,19 @@ where
         x if x == *MULTISIG_ACTOR_CODE_ID => multisig::Actor.invoke_method(rt, method_num, params),
         x if x == *REWARD_ACTOR_CODE_ID => reward::Actor.invoke_method(rt, method_num, params),
         x if x == *VERIFREG_ACTOR_CODE_ID => verifreg::Actor.invoke_method(rt, method_num, params),
-        _ => Err(actor_error!(SysErrorIllegalActor; "no code for actor at address {}", to)),
+        x => {
+            if rt.registered_actors.contains(&x) {
+                return match x {
+                    x if x == *PUPPET_ACTOR_CODE_ID => {
```
austinabell, line 740: chaos actor is in now, so can add the check here :)

vm/interpreter/src/default_runtime.rs
```diff
@@ -733,7 +737,23 @@ where
...
+                    }
+                    x if x == *CHAOS_ACTOR_CODE_ID => {
+                        chaos::Actor.invoke_method(rt, method_num, params)
+                    }
+                    _ => Err(
+                        actor_error!(SysErrorIllegalActor; "no code for actor at address {}", to),
+                    ),
+                }
+            } else {
+                Err(actor_error!(SysErrorIllegalActor; "no code for actor at address {}", to))
```
nannick, line 754: I would change the error message for this part. Maybe say that the actor isn't registered?

austinabell: technically, should be something to say that the registered actor cid doesn't exist (other way around)

ec2: I agree with austin. i will push that change

nannick: > technically, should be something to say that the registered actor cid doesn't exist (other way around)

True just wanted to distinguish the error messages on 754 and 750. Cuz its possible the chaos or puppet actor werent registered

ec2: makes sense. change pushed

PR #655, opened 2020-08-26
Expose message fields

austinabell:
**Summary of changes**
- The only reason to have kept from being public was for if the Cid was to be cached later, but doesn't really seem like much of a reason to keep. I didn't remove the pattern altogether for backwards compatibility and for convenience in certain places (just testing things)
  - Avoids runtime checks for these commonly used funcs so kinda a micro optimization lol

Closes 

### DIFF
--- a/vm/interpreter/src/default_runtime.rs
+++ b/vm/interpreter/src/default_runtime.rs
@@ -238,15 +238,18 @@ where
         value: TokenAmount,
         params: Serialized,
     ) -> Result<Serialized, ActorError> {
-        let msg = UnsignedMessage::builder()
-            .from(from)
-            .to(to)
-            .method_num(method)
-            .value(value)
-            .params(params)
-            .gas_limit(self.gas_available())
-            .build()
-            .expect("Message creation fails");
+        let msg = UnsignedMessage {
+            from,
+            to,
+            method_num: method,
+            value,
+            params,
+            gas_limit: self.gas_available(),
+            version: Default::default(),
+            sequence: Default::default(),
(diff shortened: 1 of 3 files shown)


vm/interpreter/src/default_runtime.rs
```diff
@@ -238,15 +238,18 @@ where
...
+            from,
+            to,
+            method_num: method,
+            value,
+            params,
+            gas_limit: self.gas_available(),
+            version: Default::default(),
+            sequence: Default::default(),
+            gas_fee_cap: Default::default(),
+            gas_premium: Default::default(),
```
nannick, line 251: Could probably get away with  ..Default::default() for all the fields that are automatic. But might be better for readability if everything was explicit.

ec2: @austinabell and i talked about it. It's not possible to do that unless UnsignedMessage is Default

ec2:
I think the folks at Zondax use our Message library. Should probably tag them here. Not sure who to tag though.

vm/interpreter/src/default_runtime.rs
nannick: Completely failed to notice it didn't implement Default

austinabell: it can't because there are addresses in the struct, which cannot implement default

austinabell:
> I think the folks at Zondax use our Message library. Should probably tag them here. Not sure who to tag though.

It's not a breaking change but @jleni. I will bump the patch version and release this

PR #659, opened 2020-08-27
Fix internal send bug, remove message ref from runtime

austinabell:
**Summary of changes**
- Internal send was not updating the from address of the message for subsequent calls, so some actor validations would fail
- Removed unsigned message ref from runtime (is unnecessary and will PR into Lotus removing theirs https://github.com/filecoin-project/lotus/pull/3366)
  - I just wanted to make sure no where referenced the unsigned message data
- Other small changes to be proactive with inconsistencies

This makes 3 more pass, but 2 fail (previously false positive)
34/66

Closes 

### DIFF
--- a/tests/conformance_tests/tests/conformance_runner.rs
+++ b/tests/conformance_tests/tests/conformance_runner.rs
@@ -22,7 +22,7 @@ use runtime::{ConsensusFault, Syscalls};
 use serde::{Deserialize, Deserializer};
 use std::error::Error as StdError;
 use std::fs::File;
-use std::io::{BufReader, Read};
+use std::io::BufReader;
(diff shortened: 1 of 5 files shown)


tests/conformance_tests/tests/conformance_runner.rs
```diff
@@ -174,11 +174,11 @@ fn is_valid_file(entry: &DirEntry) -> bool {
         Some(file) => file,
         None => return false,
     };
-    for rx in SKIP_TESTS.iter() {
-        if rx.is_match(file_name) {
-            return false;
-        }
-    }
+    // for rx in SKIP_TESTS.iter() {
```
ec2, line 177: ??????

PR #660, opened 2020-08-28
Fix caller validation on nested sends

austinabell:
**Summary of changes**
- So still can't duplicate the runtime to match the go implementation completely, but I verified that this is the only other variable that could cause an issue (there is also the allow_internal flag that would be reset, but I believe that to be a bug in their code that they would allow internal calls while inside a transaction if vm send is called, but their code could never do this so I suppose it's fine https://github.com/filecoin-project/lotus/pull/3377)

Closes 

### DIFF
--- a/vm/interpreter/src/default_runtime.rs
+++ b/vm/interpreter/src/default_runtime.rs
@@ -746,42 +746,34 @@ where
     R: Rand,
 {
     let ret = match code {
-        x if x == *SYSTEM_ACTOR_CODE_ID => system::Actor.invoke_method(rt, method_num, params)?,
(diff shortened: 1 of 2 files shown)


vm/interpreter/src/default_runtime.rs
```diff
@@ -261,12 +261,16 @@ where
             gas_premium: Default::default(),
         };
 
+        // Since it is unsafe to share a mutable reference to the state tree by copying
+        // the runtime, all variables must be copied/and reset at the end of the transition.
```
timvermeulen, line 265: >copied/and

Typo?

PR #662, opened 2020-08-28
check value correctly before transfer

ec2:
**Summary of changes**
- tied for worlds smallest PR

Closes 

### DIFF
--- a/vm/interpreter/src/default_runtime.rs
+++ b/vm/interpreter/src/default_runtime.rs
@@ -660,7 +660,7 @@ where
             .on_method_invocation(msg.value(), msg.method_num()),
     )?;
 
-    if msg.value() > &TokenAmount::from(0) {
+    if msg.value() != &TokenAmount::from(0) {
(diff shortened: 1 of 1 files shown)


vm/interpreter/src/default_runtime.rs
```diff
@@ -660,7 +661,7 @@ where
             .on_method_invocation(msg.value(), msg.method_num()),
     )?;
 
-    if msg.value() > &TokenAmount::from(0) {
+    if !msg.value().is_zero() {
```
austinabell, line 664: this value could be negative, might be better to check 
```
msg.sign() != Sign::Minus
```

austinabell, line 664: Wait, why can this be negative?

ec2: That is handled in the transfer function

austinabell: oh yeah, that's super weird, was wondering about that failed test case

PR #664, opened 2020-08-28
Fix runtime implementation to return gas blockstore

austinabell:
**Summary of changes**
- This call is used within actors to access the blockstore, so they were not being charged within execution when using the base blockstore

22/66 failures now, but most are just due to one extra ipld read, but we will find that :)

Closes 

### DIFF
--- a/vm/interpreter/src/default_runtime.rs
+++ b/vm/interpreter/src/default_runtime.rs
@@ -334,7 +334,8 @@ where
     }
 }
 
-impl<BS, SYS, R, P> Runtime<BS> for DefaultRuntime<'_, '_, '_, '_, '_, BS, SYS, R, P>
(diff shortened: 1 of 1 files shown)


austinabell:
@dutterbutter and @ec2 let me know if not okay with https://github.com/ChainSafe/forest/pull/664/commits/3ccce328ffad6796a56371adbac7bef2c47d579a

Now only 9/66 fail, but I will open an issue or ask a question if this is intended (seems weird they would do this)

ec2:
Looks good. I'd merge it anyways and open another PR later if they determine its not what they intended

austinabell:
> Looks good. I'd merge it anyways and open another PR later if they determine its not what they intended

of course, just pinged them to make sure that's not a mistake (easy change for us later if it is incorrect, but yeah we wouldn't update until test-vectors is updated)

PR #667, opened 2020-08-28
Update EPOCH_DURATION_SECONDS

ec2:
**Summary of changes**
- This fixed some paych conformance tests
9/66

Closes 


austinabell:
can you update the const in the clock crate to point to this? it's still using the old 15 variable and this should be shared (doesn't really matter where the const lives if it's in clock or actors)

PR #668, opened 2020-08-30
Update statetree cache

### LINKED ISSUE #373: Update StateTree to not flush on snapshot

dutterbutter:
**Summary of changes**
- Updates state tree caching to follow lotus

Closes #373 


vm/state_tree/src/lib.rs
```diff
@@ -5,18 +5,182 @@ use actor::{init, INIT_ACTOR_ADDR};
...
+                let res_addr = layer
+                    .resolve_cache
+                    .read()
+                    .get(addr)
+                    .cloned()
+                    .ok_or_else(|| format!("No cached actor state found at address {}", addr))?;
+                Ok(Some(res_addr))
+            });
+
+        Ok(None)
```
austinabell, line 116: This always returns none?

vm/state_tree/src/lib.rs
```diff
@@ -5,18 +5,182 @@ use actor::{init, INIT_ACTOR_ADDR};
...
+            .iter()
+            .map::<Result<Option<ActorState>, Box<dyn StdError>>, _>(|layer: &StateSnapLayer| {
+                let ac =
+                    layer.actors.read().get(addr).cloned().ok_or_else(|| {
+                        format!("No cached actor state found at address {}", addr)
+                    })?;
+                Ok(ac)
+            });
+
+        Ok(None)
```
austinabell, line 151: why are you ignoring the results?

vm/interpreter/src/default_runtime.rs
```diff
@@ -265,20 +270,17 @@ where
...
-            .state
-            .snapshot()
-            .map_err(|e| actor_error!(fatal("failed to create snapshot {}", e)))?;
-
         let send_res = vm_send::<BS, SYS, R, P>(self, &msg, None);
 
         // Reset values back to their values before the call
         self.vm_msg = prev_msg;
         self.caller_validated = prev_val;
+        self.state
```
ec2, line 278: I think this is wrong. clear_snapshot should be called after revert_to_snapshot.

ec2: Just tested the conformance tests with what i said and 3 more tests pass.

austinabell: The way they have their snapshots setup is pretty backwards, their snapshot revert clears a snapshot from their slice then adds a blank one, maybe why the confusion (I don't necessarily think this is a good pattern, it's redundant)

node/rpc/src/wallet_api.rs
```diff
@@ -43,10 +44,8 @@ where
             Ok(actor_balance.to_string())
         }
         Err(e) => {
-            if e == "Address not found" {
-                return Ok(BigUint::default().to_string());
-            }
-            Err(e.into())
+            warn!("Error retrieving actor state: {}", e);
+            Ok(BigUint::default().to_string())
```
austinabell, line 48: Can't comment above, but on the none case, a balance of 0 should be returned. (the ok_or above) since this seemed to be checking that way before for some reason

vm/interpreter/src/default_runtime.rs
```diff
@@ -383,14 +388,14 @@ where
...
-            .map_err(ActorError::new_fatal)
+            .map_err(|e| actor_error!(fatal("failed to look up id: {}", e)))
     }
 
     fn get_actor_code_cid(&self, addr: &Address) -> Result<Option<Cid>, ActorError> {
         Ok(self
             .state
             .get_actor(&addr)
-            .map_err(ActorError::new_fatal)?
+            .map_err(|e| actor_error!(fatal("failed:398 to get actor: {}", e)))?
```
austinabell, line 398: I assume this error message change was just for debugging?

vm/state_tree/src/lib.rs
```diff
@@ -47,56 +199,52 @@ where
...
 
     /// Get an ID address from any Address
-    pub fn lookup_id(&self, addr: &Address) -> Result<Option<Address>, String> {
+    pub fn lookup_id(&self, addr: &Address) -> Result<Option<Address>, Box<dyn StdError>> {
         if addr.protocol() == Protocol::ID {
             return Ok(Some(*addr));
         }
 
+        match self.snaps.resolve_address(addr) {
+            None => warn!("No address cached"),
```
austinabell, line 244: why should this be a warn? address not being cached is an extremely common occurance, should probably just have this as an `if let`

vm/state_tree/src/lib.rs
```diff
@@ -173,30 +327,49 @@ where
...
+                Some(ref state) => {
+                    self.hamt
+                        .set(addr.to_bytes().into(), state.clone())
+                        .map_err(|e| e.to_string())?;
+                }
+            }
         }
 
-        self.hamt.flush().map_err(|e| e.to_string())
+        Ok(self.hamt.flush().map_err(|e| e.to_string())?)
```
austinabell, line 373: All of these `to_string()` calls defeat the purpose of the dynamic errors, you should remove all of these

austinabell (changes requested):
oh man, that error handling is pretty painful now. Looking at this I didn't realize basically all state tree access errors are explicitly fatal (defeating purpose of the `Box<dyn std::error::Error>`). I can try to review if there is an edge case where an error would be downcasted and made non-fatal if encoding error on go impl

vm/state_tree/src/lib.rs
```diff
@@ -108,27 +254,33 @@ where
             .map_err(|e| e.to_string())?
```
austinabell, line 254: you missed this one

vm/state_tree/src/lib.rs
```diff
@@ -27,17 +178,17 @@ where
...
-            actor_cache: RwLock::new(FnvHashMap::default()),
+            snaps: StateSnapshots::new(),
         }
     }
 
     /// Constructor for a hamt state tree given an IPLD store
-    pub fn new_from_root(store: &'db S, root: &Cid) -> Result<Self, String> {
+    pub fn new_from_root(store: &'db S, root: &Cid) -> Result<Self, Box<dyn StdError>> {
         let hamt =
             Hamt::load_with_bit_width(root, store, HAMT_BIT_WIDTH).map_err(|e| e.to_string())?;
```
austinabell, line 188: and this one

austinabell (commented):
Just those error handling fixes for future proofing and should be good!

vm/interpreter/src/default_runtime.rs
```diff
@@ -265,25 +270,25 @@ where
...
                 e
             }
-        })
+        });
+        if let Err(e) = self.state.clear_snapshot() {
+            actor_error!(fatal("failed to clear snapshot: {}", e));
+        }
+        let ret = ret?;
+
+        Ok(ret)
```
ec2, line 291: haha, `let ret = ret?` is so funny. 

You can remove that and just do Ok(ret?)

ec2 (approved):

vm/interpreter/src/default_runtime.rs
austinabell: nah, that would be a bug. The reason you need to have it after the clear, is that returning an error here may not be fatal, and the snapshot would not be cleared and run into a bug later

austinabell (approved):

PR #670, opened 2020-08-31
Storage miner integration

### LINKED ISSUE #629: Find alternative WebSocket Library

### LINKED ISSUE #630: Implement Websocket Support

### LINKED ISSUE #631: Implement Websocket API calls

### LINKED ISSUE #632: Use Local Devnet to Test Storage Miner

AshantiMutinta:
**Summary of changes**
- Adds websockets support and implements missing functions

Closes #629 
Closes #630 
Closes #631 
Closes #632 


blockchain/chain/src/store/chain_store.rs
```diff
@@ -192,6 +196,27 @@ where
...
+            while let Some(val) = subscribed_head_change.next().await {
+                if !sender.is_empty() {
+                    warn!(
+                        "head change sub is slow, has {:} buffered entries",
+                        sender.len()
+                    )
+                }
+                sender.send(val).await
+            }
+        });
```
austinabell, line 217: A task will be spawned for every function call of this, that won't ever exit. This should probably be one task where the receiver is cloned and passed back in this function?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -613,6 +658,31 @@ where
...
+        fn from(wrapper: HeadChange) -> Self {
+            match wrapper {
+                HeadChange::Current(tipset) => HeadChangeJson::Current((*tipset).clone().into()),
+
+                HeadChange::Apply(tipset) => HeadChangeJson::Apply((*tipset).clone().into()),
+                HeadChange::Revert(tipset) => HeadChangeJson::Revert((*tipset).clone().into()),
+            }
+        }
+    }
+}
```
austinabell, line 684: This will not match the Lotus implementation, are you sure you verified against? Enum representations are not internally tagged by default, and the keys are lower case, not first character capitalized.

Also for this, cloning a tipset just to serialize JSON is unnecessary, should take a reference to the tipset in the structure (if that means separating into two if both serialize and deserialize are needed, then it's worth it)

blockchain/chain_sync/src/sync.rs
```diff
@@ -68,7 +68,7 @@ pub struct ChainSyncer<DB, TBeacon> {
     next_sync_target: SyncBucket,
 
     /// access and store tipsets / blocks / messages
-    chain_store: ChainStore<DB>,
+    chain_store: Arc<RwLock<ChainStore<DB>>>,
```
austinabell, line 71: is it possible to just wrap what is needed in an Arc<RwLock<..>> instead of wrapping the whole chain store in a mutex? This kinda worries me

blockchain/chain/src/store/chain_store.rs
```diff
@@ -225,6 +250,26 @@ where
...
+        blocks.push(Block {
+            header,
+            bls_messages,
+            secp_messages,
+        });
+    }
+
+    // the given tipset has already been verified, so this cannot fail
+    Ok(FullTipset::new(blocks).unwrap())
+}
```
austinabell, line 271: can you have the function you copied this from call this so we don't have duplication here?

forest/src/daemon.rs
```diff
@@ -83,14 +82,17 @@ pub(super) async fn start(config: Config) {
...
-        chain_store,
+        chain_store.clone(),
         Arc::new(beacon),
         network_send.clone(),
         network_rx,
         genesis,
     )
+    .await
     .unwrap();
+    info!("init chain syncher");
```
austinabell, line 95: Was this a debugging log that wasn't removed? Seems duplicate and unnecessary

blockchain/chain/src/store/chain_store.rs
ec2: I think ashanti is mostly correct here, except for the fact that the loop never stops. You need to stop the loop once the receiver is dropped, i.e. the websocket connection is terminated. I think here, you could use the futures channels to better handle disconnects https://docs.rs/futures/0.3.5/futures/channel/mpsc/index.html. In async-std channels, i think you only detect a disconnection of all the receivers if the channel is full AND all recievers have been dropped.

node/rpc/src/lib.rs
```diff
@@ -158,7 +162,52 @@ where
...
+                    let response = state.handle(call).await;
+                    let response_text = serde_json::to_string(&response)
+                        .map_err(|s| Error::Protocol(Cow::from(s.to_string())))?;
+                    responses.push(response_text);
+                    Ok(responses)
+                })
+                .await;
+
+            if let Err(error) = responses_result {
+                error!("error obtaining request {:?}", error)
```
ec2, line 200: you mean response?

node/rpc/src/lib.rs
```diff
@@ -158,7 +162,52 @@ where
...
+            } else {
+                for response_text in responses_result.unwrap() {
+                    ws_sender
+                        .send(Message::text(response_text))
+                        .await
+                        .unwrap_or_else(|s| error!("Error sending response {:?}", s))
+                }
+            }
+        } else {
+            error!("web socket connection failed at {:}", addr)
```
ec2, line 210: is this necessarily a fail? this happens at disconnection too right?

ec2:
Gunna go much more in depth tomorrow. My cloud lotus node is currently syncing.

blockchain/chain_sync/src/sync.rs
AshantiMutinta: I thought about that but I think it makes sense in the context of both places being able to reflect the changes that are coming in from both sides. Like unless we go through the route of generating a new chain_store per operation which might take a lot of restructuring. Example things like HeadChange is something that would need to be reflected from both sides.

Additionally, I thought it was okay because state is behind one and trying to circumvent by using module level functions, getting the block_store from the state_manager and using sync types of the async library. 

I will look closer and see if we can just get what we need through

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: nice! Shouldn't be hard to switch between the two. Thank you!

blockchain/chain_sync/src/sync.rs
austinabell: I just mean wrapping the variables other than the blockstore in the mutex, so you don't wrap the whole struct (and would restrict blockstore access even if just one of the other variables are locked)

node/rpc/src/lib.rs
AshantiMutinta: That is true, maybe a warn might be better here?

ec2: Sounds good to me

AshantiMutinta:
> Gunna go much more in depth tomorrow. My cloud lotus node is currently syncing.
> Could you also show what commands you run to do some of the testing? What I do to lotus nodes is
> 
>     1. Run `wscat -c 127.0.0.1:1234/rpc/v0`
> 
>     2. Send requests like:` {"jsonrpc":"2.0","method":"Filecoin.ChainNotify","params":[],"id":1}`

I would use the same commands :)

austinabell:
You committed the grpc protobuf submodule by the way (and I assume unintended)

blockchain/chain/src/store/chain_store.rs
```diff
@@ -192,6 +187,34 @@ where
...
+                sender
+                    .clone()
+                    .send(s)
+                    .await
+                    .unwrap_or_else(|e| warn!("failed to send to channel : {:?}", e));
+            });
+            let dur = Duration::from_millis(10);
+            future::timeout(dur, wait_for_each)
+                .await
+                .unwrap_or_else(|_| info!("waited for head_change for 10 miliseconds"));
```
austinabell, line 214: Won't this just immediately time out? what is the purpose of this?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -192,6 +187,34 @@ where
...
+        let subscribed_head_change = self.subscribe();
+        info!("sub head changes");
+        let head = self
+            .heaviest_tipset()
+            .ok_or_else(|| Error::Other("Could not get heaviest tipset".to_string()))?;
+        let (mut sender, reciever) = channel(16);
+        sender
+            .send(HeadChange::Current(head))
+            .await
+            .map_err(|e| Error::Other(format!("Could not send to channel {:?}", e)))?;
```
austinabell, line 201: This just errors when all receivers have been dropped, are you sure you want to error here?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -613,6 +656,32 @@ where
...
 
+#[cfg(feature = "json")]
+pub mod headchange_json {
+    use super::*;
+    use blocks::tipset_json::TipsetJson;
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Serialize, Deserialize)]
+    #[serde(untagged)]
+    #[serde(rename_all = "PascalCase")]
```
austinabell, line 667: Is this untagged or tagged with PascalCase? These are contradictory. How are you checking the format?

austinabell (commented):
I still do believe that only the variables that need to be mutated across threads in the chain store should be wrapped in a mutex, you are adding locks around a lot of logic that doesn't need it and performance and/or deadlocks might be a problem

AshantiMutinta:
Mentioned on the slack chat that this will probably be done on Sunday. Put this up like this in case you needed a temporary merge

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: This is for cases where the pending future doesn't change which is flostreams design. This will timeout when future in the foreach hasn't  been resolved

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: No harm in leaving it aside from if there is any other problem into sending to a channel

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: This was a mistake, it is supposed to be tagged

austinabell: But the enum variants are not pascal cased in Lotus? They are all lower case from what I can see

node/forest_libp2p/src/service.rs
```diff
@@ -259,7 +259,8 @@ where
                     None => { break; }
                 },
                 interval_event = interval.next() => if interval_event.is_some() {
-                    info!("Peers connected: {}", swarm_stream.get_ref().peers().len());
+                    //info!("Peers connected: {}", swarm_stream.get_ref().peers().len());
```
nannick, line 262: Delete comment

austinabell: This just shouldn't be commented out (imo, if anyone disagrees please say why)

ec2: I'd say uncomment it.

AshantiMutinta:
This PR uses a forked version of JSONRPC_V2

https://github.com/AshantiMutinta/jsonrpc-v2

austinabell:
> This PR uses a forked version of JSONRPC_V2
> 
> https://github.com/AshantiMutinta/jsonrpc-v2

If this is planning to stay as such for a non-trivial amount of time, can you fork it to the ChainSafe org and give a few of us privileges in case we need to change something?

Also, did you end up trying to use jsonrpsee for this?

AshantiMutinta:
> 
> 
> > This PR uses a forked version of JSONRPC_V2
> > https://github.com/AshantiMutinta/jsonrpc-v2
> 
> If this is planning to stay as such for a non-trivial amount of time, can you fork it to the ChainSafe org and give a few of us privileges in case we need to change something?
> 
> Also, did you end up trying to use jsonrpsee for this?

I thought the jsonrpsee was still in it's fairly early stages. It shouldn't be too hard to switch if we really want it though, I think for the time being since we want the storage miner in, forking is a good alternative. 

I would switch in a subsequent PR if we really prefer it though

blockchain/chain/src/store/chain_store.rs
```diff
@@ -646,6 +739,62 @@ where
     Ok(out)
 }
 
+#[cfg(feature = "json")]
+pub mod headchange_json {
+    use super::*;
+    use blocks::tipset_json::TipsetJsonRef;
+    use serde::Serialize;
+
+    #[derive(Debug, Serialize, Clone)]
```
austinabell, line 748:
```suggestion
    #[derive(Serialize)]
```

blockchain/blocks/src/tipset.rs
```diff
@@ -271,12 +271,12 @@ pub mod tipset_json {
     use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 
     /// Wrapper for serializing and deserializing a SignedMessage from JSON.
-    #[derive(Deserialize, Serialize, Debug)]
+    #[derive(Deserialize, Serialize, Debug, Clone)]
```
austinabell, line 274:
```suggestion
    #[derive(Deserialize, Serialize)]
```

blockchain/blocks/src/tipset.rs
```diff
@@ -271,12 +271,12 @@ pub mod tipset_json {
...
     #[serde(transparent)]
     pub struct TipsetJson(#[serde(with = "self")] pub Tipset);
 
     /// Wrapper for serializing a SignedMessage reference to JSON.
-    #[derive(Serialize)]
+    #[derive(Serialize, Clone, Debug)]
```
austinabell, line 279:
```suggestion
    #[derive(Serialize)]
```

forest/src/daemon.rs
```diff
@@ -123,6 +127,9 @@ pub(super) async fn start(config: Config) {
                     sync_state,
                     network_send,
                     network_name,
+                    heaviest_tipset,
+                    subscriber,
```
austinabell, line 131: This subscriber is cloned here, but never polled off of, so this will stall eventually (need to be certain no subscriber _doesn't_ get polled

node/rpc/src/lib.rs
```diff
@@ -28,19 +52,16 @@ where
 {
     pub state_manager: Arc<StateManager<DB>>,
     pub keystore: Arc<RwLock<KS>>,
+    pub heaviest_tipset: Arc<RwLock<Option<Arc<Tipset>>>>,
+    pub subscriber: Subscriber<HeadChange>,
```
austinabell, line 56: If anything, this should be an `Arc<Publisher<..>>` where if you need to get something to subscribe to it, you use that to create a new subscriber

node/rpc/src/lib.rs
```diff
@@ -51,114 +72,331 @@ where
...
+    let listener = try_socket.expect("Failed to bind to addr");
+    let state = Arc::new(rpc);
+
+    info!("waiting for web socket connections");
+    while let Ok((stream, addr)) = listener.accept().await {
+        task::spawn(handle_connection_and_log(
+            state.clone(),
+            stream,
+            addr,
+            subscriber.clone(),
```
austinabell, line 264: be careful cloning the subscribers, this would probably change if you change the previous comments, but the subscriber that's being cloned isn't being polled off

austinabell (changes requested):

node/rpc/src/lib.rs
```diff
@@ -51,114 +72,331 @@ where
...
+                                    chain_notify_count,
+                                )
+                                .await;
+                            }
+                            Err(_) => {
+                                let response = ResponseObjects::One(ResponseObject::Error {
+                                    jsonrpc: V2,
+                                    error: Error::Provided {
+                                        code: 1,
+                                        message: "Error serialization request",
```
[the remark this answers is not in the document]
AshantiMutinta: Hmm I could switch that around, annoying because it only accepts static strings but easy to change since we have forks

node/rpc/src/lib.rs
```diff
@@ -51,114 +72,331 @@ where
...
+                while let Some(index_to_head_change) = subscriber.next().await {
+                    if streaming_count == index_to_head_change.0 {
+                        let head_change = (&index_to_head_change.1).into();
+                        let data = StreamingData {
+                            json_rpc: "2.0".to_string(),
+                            method: "xrpc.ch.val".to_string(),
+                            params: (streaming_count, vec![head_change]),
+                        };
+                        let response_text =
+                            serde_json::to_string(&data).expect("Bad Serialization of Type");
```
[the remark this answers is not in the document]
AshantiMutinta: I think this is one of those cases where we have unrecoverable errors. I looked at it like we are the ones doing the serialization and at this point the serialization should always pass from our side since we are in charge of the implementation

blockchain/chain/src/store/chain_store.rs
```diff
@@ -646,6 +739,62 @@ where
...
+            match wrapper {
+                HeadChange::Current(tipset) => HeadChangeJson::Current(TipsetJsonRef(&tipset)),
+                HeadChange::Apply(tipset) => HeadChangeJson::Apply(TipsetJsonRef(&tipset)),
+                HeadChange::Revert(tipset) => HeadChangeJson::Revert(TipsetJsonRef(&tipset)),
+            }
+        }
+    }
+
+    pub async fn sub_head_changes(
+        publisher: Arc<RwLock<Publisher<IndexToHeadChangeJson>>>,
```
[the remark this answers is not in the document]
AshantiMutinta: The publisher is needed here because you need to make sure you are iterating from the first the beginning of the channel in situations where RPCState is being shared amongst processes

blockchain/chain/src/store/chain_store.rs
```diff
@@ -214,6 +211,73 @@ where
...
+    *heaviest = Some(ts.clone());
+    publisher
+        .write()
+        .await
+        .publish(HeadChange::Current(ts))
+        .await;
+    Ok(())
+}
+
+pub async fn update_heaviest_lock<DB>(
```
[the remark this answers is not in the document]
AshantiMutinta: This came in before your changes for the ChainSyncer, they were lock versions of the functions. They should be removed

node/rpc/src/lib.rs
AshantiMutinta: The reason why this is a subscribe is that it is a one time usage just for the purpose of the sub_head_changes function. We are only ever using this to poll from the subscribe channel to iterate our HeadChangeJson.

Having a publisher return subscribers will mean that publisher will have to be behind a Arc<Lock which also carries overhead

forest/src/daemon.rs
AshantiMutinta: What do you mean? It's being polled in the sub_head_changes. Maybe im missing something?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -245,6 +309,35 @@ where
...
+    for header in ts.into_blocks() {
+        let (bls_messages, secp_messages) = block_messages(db, &header)?;
+        debug!(
+            "Fill Tipsets for header {:?} with bls_messages: {:?}",
+            header.cid(),
+            bls_messages
+                .iter()
+                .map(|b| b.cid().unwrap())
+                .collect::<Vec<_>>()
+        );
```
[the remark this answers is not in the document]
AshantiMutinta: remnants of debugging I think, will remove this

node/rpc/src/lib.rs
```diff
@@ -9,17 +9,41 @@ mod sync_api;
...
-use tide::{Request, Response, StatusCode};
 use wallet::KeyStore;
 
+type WsSink = SplitSink<WebSocketStream<TcpStream>, async_tungstenite::tungstenite::Message>;
+
+const CHAIN_NOTIFY_METHOD_NAME: &str = "Filecoin.ChainNotify";
+#[derive(Serialize)]
+struct StreamingData<'a> {
+    json_rpc: String,
+    method: String,
```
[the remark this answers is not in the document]
AshantiMutinta: Changed to &'a str instead

node/rpc/src/lib.rs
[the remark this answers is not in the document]
austinabell: why don't you use the `Full` error variant, and then no need to change?

node/rpc/src/lib.rs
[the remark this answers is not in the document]
austinabell: yeah, that might be correct, but I don't see the negative in handling to avoid crashing or to future proof if something else is changed

node/rpc/src/lib.rs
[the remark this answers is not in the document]
austinabell: it doesn't need a lifetime though?

blockchain/chain/src/store/chain_store.rs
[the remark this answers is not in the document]
austinabell: Maybe I'm confused about how this is being published to multiple threads, I can look into this later.

forest/src/daemon.rs
austinabell: but that's a cloned subscriber?

node/rpc/src/lib.rs
austinabell: but this RpcState is holding this subscriber that never gets polled, as far as I can see (can you point to where it's being handled?).

The only reason you would hold the publisher is so that you'd be able to subscribe without holding a dead subscriber. I do see the overhead concern with this, especially since the subscribe would require a write lock for some reason, but as of now that would be better than the node stalling after this subscriber buffer fills

node/rpc/src/lib.rs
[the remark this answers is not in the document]
AshantiMutinta: static will mean those string will exist the whole time the program is running. &'a in this case will only limit to the lifetime of the enum which is better imo

blockchain/chain/src/store/chain_store.rs
```diff
@@ -646,6 +663,102 @@ where
...
+                _ => None,
+            }
+        }
+
+        pub fn task_cancel(&self) -> Option<(usize, ())> {
+            match self {
+                EventsPayload::TaskCancel(val, _) => Some((*val, ())),
+                _ => None,
+            }
+        }
```
austinabell, line 712: These make no sense to me, why would you ever need this that you wouldn't be able to do with a match. (I see where it's used, but it doesn't seem clear to me why

blockchain/chain/src/store/chain_store.rs
```diff
@@ -646,6 +663,102 @@ where
...
+                            .map(|s| s.0 == current_index)
+                            .unwrap_or_default(),
+                    )
+                })
+                .next()
+                .await
+            {
+                handle.cancel().await;
+            }
+        });
```
austinabell, line 757: why is a cancel event published on a pubsub type channel? When would you ever want to send something to cancel all subscriptions? I'm really confused on this pattern, or how it would be used, if you don't mind elaborating the intention

blockchain/chain_sync/src/sync.rs
```diff
@@ -346,18 +346,18 @@ where
     }
     /// Validates message root from header matches message root generated from the
     /// bls and secp messages contained in the passed in block and stores them in a key-value store
-    fn validate_msg_meta(&self, block: &Block) -> Result<(), Error> {
+    async fn validate_msg_meta(&self, block: &Block) -> Result<(), Error> {
```
austinabell, line 349: there is nothing async in this function?

forest/src/daemon.rs
```diff
@@ -68,7 +69,8 @@ pub(super) async fn start(config: Config) {
     let state_manager = Arc::new(StateManager::new(Arc::clone(&db)));
 
     // Initialize mpool
-    let subscriber = chain_store.subscribe().await;
+    let publisher = chain_store.publisher_arc();
+    let subscriber = (*publisher.write().await).subscribe();
```
austinabell, line 73:
```suggestion
    let subscriber = publisher.write().await.subscribe();
```
This will deref automatically

node/rpc/src/lib.rs
```diff
@@ -9,17 +9,45 @@ mod sync_api;
 mod wallet_api;
 
 use crate::state_api::*;
+use async_log::span;
+use async_std::net::{TcpListener, TcpStream};
+use async_std::sync::Arc;
 use async_std::sync::{RwLock, Sender};
+use async_std::task;
+use async_std::task::JoinHandle;
```
austinabell, line 17:
```suggestion
use async_std::sync::{RwLock, Sender, Arc};
use async_std::task::{self,JoinHandle};
```
Can you group imports like this?

node/rpc/src/lib.rs
```diff
@@ -9,17 +9,45 @@ mod sync_api;
...
+use async_tungstenite::{tungstenite::Message, WebSocketStream};
+use blocks::Tipset;
 use blockstore::BlockStore;
+use chain::headchange_json::{EventsPayload, HeadChangeJson};
+use chain::HeadChange;
 use chain_sync::{BadBlockCache, SyncState};
+use flo_stream::MessagePublisher;
+use flo_stream::{Publisher, Subscriber};
```
austinabell, line 25:
```suggestion
use flo_stream::{MessagePublisher, Publisher, Subscriber};
```

node/rpc/src/lib.rs
```diff
@@ -51,114 +76,421 @@ where
...
-    let mut app = tide::Server::with_state(rpc);
-    app.at("/rpc/v0").post(handle_json_rpc);
-    app.listen(rpc_endpoint).await.unwrap();
+    let try_socket = TcpListener::bind(rpc_endpoint).await;
+    let listener = try_socket.expect("Failed to bind to addr");
+    let rpc_state = Arc::new(rpc);
+
+    info!("waiting for web socket connections");
+    while let Ok((stream, addr)) = listener.accept().await {
+        let subscriber = (*events_pubsub.write().await).subscribe();
```
austinabell, line 264:
```suggestion
        let subscriber = events_pubsub.write().await.subscribe();
```

node/rpc/src/lib.rs
```diff
@@ -51,114 +76,421 @@ where
...
+            }
+        } else {
+            warn!("web socket connection failed at {:}", addr)
+        }
+    })
+}
+
+async fn send_error(
+    code: i64,
+    ws_sender: Arc<RwLock<WsSink>>,
```
austinabell, line 412: just a thought, can this not just be `&mut WsSink` and just await a write lock on the function call so you don't need to clone the arc when you are awaiting the response anyway?

node/rpc/src/lib.rs
```diff
@@ -51,114 +72,331 @@ where
...
+        .await
+        .send(Message::text(response_text))
+        .await
+        .unwrap_or_else(|_| warn!("Could not send to response to socket"));
+    if let ResponseObjects::One(ResponseObject::Result {
+        jsonrpc: _,
+        result: _,
+        id: _,
+        streaming,
+    }) = response_object
```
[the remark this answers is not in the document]
austinabell: ^^ What was the reason why this is resolved? Intentionally doing for what reason?

node/rpc/src/lib.rs
```diff
@@ -51,114 +76,421 @@ where
...
+    {
+        if streaming {
+            let handle = task::spawn(async move {
+                let sender = ws_sender.clone();
```
austinabell, line 454: why does this need to be cloned?

node/rpc/src/state_api.rs
```diff
@@ -304,7 +286,7 @@ pub(crate) async fn state_account_key<
     let state = state_for_ts(&state_manager, Some(tipset))?;
     let address =
         interpreter::resolve_to_key_addr(&state, state_manager.get_block_store_ref(), &actor)?;
-    Ok(address.into())
+    Ok(Some(address))
```
austinabell, line 289: I can't comment on previous, but won't this be incorrect serialization because it's going to serialize as bytes? Default json serialization should be string?

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: The cancel even t is not sent to a publish. The channel that is done with the sub_head_changes is sent to this function with the appropriate id and the sub_head_changes method that matches up with this ID is cancelled. They can be a of sub_head_changes running at the same time. So this doesn't publish but checks if an event with the subscription ID has been passed through

blockchain/chain_sync/src/sync.rs
AshantiMutinta: Probably a remnant from an old refactor

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: This is for the purpose of using it in filter closures. I want to be able to say task_cancel().map(|s|//operation goes here), I find it match cleaner than adding a match in a closure

blockchain/chain/src/store/chain_store.rs
```diff
@@ -51,16 +53,16 @@ pub enum HeadChange {
...
     // key-value datastore
     pub db: Arc<DB>,
 
     // Tipset at the head of the best-known chain.
-    heaviest: RwLock<Option<Arc<Tipset>>>,
+    heaviest: Arc<RwLock<Option<Arc<Tipset>>>>,
 
     // tip_index tracks tipsets by epoch/parentset for use by expected consensus.
-    tip_index: TipIndex,
+    tip_index: Arc<RwLock<TipIndex>>,
```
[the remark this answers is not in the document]
AshantiMutinta: I thought it was suggested the first time that the RPCState should just obtain mutable values that it needs to refrain from potentially blocking the blockstore. 

These both need to be in Sync Channel and RPCState, this is straying away from the block_store since we do not want to block that. It is a "taking what we need approach"

node/rpc/src/lib.rs
AshantiMutinta: The downside of that is the lock will be obtained for the whole lifetime of the function rather than just the single line that needs the send but I can change

austinabell: Ah ya, I guess I meant a ref to the RwLock, to avoid the arc clone

node/rpc/src/lib.rs
AshantiMutinta: That was a mistake, will remove that

blockchain/chain/src/store/chain_store.rs
```diff
@@ -157,6 +159,22 @@ where
...
+
+    pub fn tip_index_rwlock(&self) -> &RwLock<TipIndex> {
+        &self.tip_index
+    }
+
+    pub fn publisher_rwlock(&self) -> &RwLock<Publisher<HeadChange>> {
+        &self.publisher
+    }
+
+    pub fn blockstore_arc(&self) -> Arc<DB> {
```
austinabell, line 174: You really shouldn't be describing these functions with the field type. Reason being is that if that type ever changes, you need to either have to rename the function and make breaking changes, or leave and have the name be misleading.

I suggest taking off the `_rwlock` suffixes, and switching the `_arc` suffix to `_cloned`

blockchain/chain/Cargo.toml
```diff
@@ -26,9 +26,14 @@ byteorder = "1.3.4"
 beacon = { path = "../beacon" }
 flo_stream = "0.4.0"
 address = { package = "forest_address", path = "../../vm/address" }
-lazy_static = "1.4"
+futures ="0.3.5"
```
austinabell, line 29:
```suggestion
futures = "0.3.5"
```

austinabell (approved):
LGTM pending @ec2's checks

ec2 (approved):
forgot the hit approve lgtm

blockchain/chain/src/store/chain_store.rs
```diff
@@ -705,6 +713,102 @@ where
...
+            match wrapper {
+                HeadChange::Current(tipset) => HeadChangeJson::Current(TipsetJsonRef(&tipset)),
+                HeadChange::Apply(tipset) => HeadChangeJson::Apply(TipsetJsonRef(&tipset)),
+                HeadChange::Revert(tipset) => HeadChangeJson::Revert(TipsetJsonRef(&tipset)),
+            }
+        }
+    }
+
+    #[derive(Clone)]
+    pub enum EventsPayload {
```
austinabell, line 744: Also, why is this type in the `headchange_json` mod?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -705,6 +713,102 @@ where
...
+    #[serde(rename_all = "lowercase")]
+    #[serde(tag = "type", content = "val")]
+    pub enum HeadChangeJson<'a> {
+        Current(TipsetJsonRef<'a>),
+        Apply(TipsetJsonRef<'a>),
+        Revert(TipsetJsonRef<'a>),
+    }
+
+    #[derive(Debug, Clone)]
+    pub struct IndexToHeadChangeJson(pub usize, pub HeadChange);
```
austinabell, line 732: Also, did you mean to annotate the headchange type here to use the `HeadChangeJson` serialization?

PR #678, opened 2020-09-01
Update apply_blocks call

austinabell:
**Summary of changes**
- There were some functional changes since last pass, as well as some necessary changes, and this needs to be 1:1 for conformance tests, and the block ones are being finalized.

More of a preemptive pass to make sure everything is in line

Closes 


vm/interpreter/src/vm.rs
```diff
@@ -166,13 +211,20 @@ where
...
             if let Some(err) = ret.act_error {
                 return Err(format!(
                     "failed to apply reward message for miner {}: {}",
-                    block.header().miner_address(),
-                    err
+                    block.miner, err
+                )
+                .into());
+            }
+
```
ec2, line 222: I believe you want to call the callback function after you send the message to the RewardActor.

austinabell: Yeah, it is, look a few lines below

ec2: Ah shoot, derp.

AshantiMutinta (approved):

PR #682, opened 2020-09-02
Tipset vector runner

### LINKED ISSUE #680: Tipset test vector runner
**Task summary**

Now can be done since vectors are updated: https://github.com/filecoin-project/lotus/pull/3449

**Specification reference**

- 

austinabell:
**Summary of changes**
- Updates test runner framework for tipset vectors
  - Also refactors so all code isn't in a monolithic file and can be more easily navigated/built upon and dependencies correctly optionally compiled
- The only tipset vector fails because blocked on miner actor impl as mentioned, but seems like that is only discrepancy, so should be fine to pull in and if any other difference it can be debugged after

Closes #680 


austinabell:
currently blocked by #559. All transactions pass and return the correct exit code, but the reward transaction fails because the miner state has changed (and too breaking to just update for this) and burns the funds instead.

vm/actor/src/util/chaos/mod.rs
```diff
@@ -94,6 +94,22 @@ impl Actor {
 
         rt.create_actor(actor_cid, &actor_address)
     }
+
```
dutterbutter, line 97: Nit: Add comment please

dutterbutter (approved):

vm/actor/src/util/chaos/mod.rs
austinabell: just used for testing, but done

ec2 (approved):

PR #691, opened 2020-09-08
Update Miner actor

### LINKED ISSUE #559: Update Miner actor
**Task summary**

**Specification reference**

- 

timvermeulen:
Closes #559


ec2:
One thing throughout is that when you return Actor Errors, you're not using `actor_error!`. It actually makes the code all around quite cleaner IMO. You can see an example in multi sig actor.

timvermeulen:
@ec2 Oh yeah, I initially did but they caused some formatting issues so I (temporarily) replaced them. I'd be happy to put them back, they're indeed more elegant.

ec2:
I'd prefer it, but no strong feelings 👍

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -1,62 +1,87 @@
...
+use std::{cmp, iter, ops::Neg};
 use vm::{
-    actor_error, ActorError, DealID, ExitCode, MethodNum, Serialized, TokenAmount,
-    METHOD_CONSTRUCTOR, METHOD_SEND,
+    ActorError, DealID, ExitCode, MethodNum, Serialized, TokenAmount, METHOD_CONSTRUCTOR,
+    METHOD_SEND,
 };
 
-// * Updated to specs-actors commit: 9e8c0d1c40d8b41de5dc727b6791c89e14fea4a8
+// * Updated to specs-actors commit: b8a3a6ff7b15ac01f0534c47059e1c81652a61f0
```
austinabell, line 84: Do you mind updating to 0.9.3 commit? There is only a few changes, but the control address logic that is used in the other actors for this version would be more inline.

Here is the diff: https://github.com/filecoin-project/specs-actors/compare/b8a3a6ff7b15ac01f0534c47059e1c81652a61f0...f4024efad09a66e32bfeef10a2845b2b35325297

0.9.3 commit is f4024efad09a66e32bfeef10a2845b2b35325297 (if you could specify the version here also that would be so helpful for tracking)

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -415,16 +613,38 @@ impl Actor {
...
-        if params.expiration <= rt.curr_epoch() {
+        if !check_supported_proof_types(params.seal_proof) {
+            return Err(ActorError::new(
+                ExitCode::ErrIllegalArgument,
+                format!("unsupported seal proof type: {:?}", params.seal_proof),
+            ));
+        }
+
+        #[allow(clippy::absurd_extreme_comparisons)]
+        if params.sector_number > MAX_SECTOR_NUMBER {
```
ec2, line 624: i know they do this in lotus, but can this ACTUALLY ever occur in Rust?

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -415,16 +613,38 @@ impl Actor {
...
-                    "sector expiration {} must be after now {}",
-                    params.expiration,
-                    rt.curr_epoch()
+                    "sector number {} out of range 0..(2^63-1)",
+                    params.sector_number
                 ),
             ));
         }
+
+        if params.sealed_cid == Cid::default() {
```
ec2, line 634: This does not actually match what Lotus is doing. They have a CID.Undef, which is different than our Cid::default. @austinabell can explain better.

ec2: I always forget/confuse their CID Undef lol

vm/actor/src/builtin/miner/mod.rs
timvermeulen: It cannot, hence the clippy thing, but this seems like the safest way to do it in case they lower that constant later?

ec2: Makes sense. Its just bizarre haha

austinabell: The bound is based on a limitation of their Amt implementation. This definitely needs to be checked because otherwise a param could just be passed in to create a discrepancy.

Thanks for flagging this, I noticed that they have changed the max in the Amt implementation. I'll fix that now

vm/actor/src/builtin/miner/mod.rs
austinabell: yeah, you should be able to safely ignore Cid.Undef checks, that is just a sanity check they have. Cid undef cannot be serialized, so it can never be an actual param. This is only added because these functions can be called directly in the go implementation

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -820,17 +1244,25 @@ impl Actor {
...
+                    params.sector_number, e
+                ),
+            )),
+            Ok(None) => Err(ActorError::new(
                 ExitCode::ErrNotFound,
-                format!("sector hasn't been proven {}", params.sector_number),
-            );
+                format!("sector {} not proven", params.sector_number),
+            )),
+            Ok(_sector) => Ok(()),
```
ec2, line 1259: Technically more clear if you say Ok(Some(_)) instead of Ok(_)

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -2020,113 +2820,100 @@ where
...
-    //     TokenAmount::zero(),
-    // )?;
-    Ok(())
+fn have_pending_early_terminations<BS, RT>(_rt: &mut RT, state: &State) -> bool
+where
+    BS: BlockStore,
+    RT: Runtime<BS>,
+{
+    let no_early_terminations = state.early_terminations.is_empty();
+    !no_early_terminations
```
ec2, line 2916:
```suggestion
    !state.early_terminations.is_empty()
```

vm/actor/src/builtin/miner/monies.rs
```diff
@@ -0,0 +1,176 @@
...
+// LockTarget = (LockTargetFactorNum / LockTargetFactorDenom) * FILCirculatingSupply(t)
+// PledgeShare(t) = sectorQAPower / max(BaselinePower(t), NetworkQAPower(t))
+// PARAM_FINISH
+pub const PRE_COMMIT_DEPOSIT_FACTOR: u64 = 20;
+pub const INITIAL_PLEDGE_FACTOR: u64 = 20;
+pub const PRE_COMMIT_DEPOSIT_PROJECTION_PERIOD: i64 =
+    (PRE_COMMIT_DEPOSIT_FACTOR as ChainEpoch) * EPOCHS_IN_DAY;
+pub const INITIAL_PLEDGE_PROJECTION_PERIOD: i64 =
+    (INITIAL_PLEDGE_FACTOR as ChainEpoch) * EPOCHS_IN_DAY;
+pub fn lock_target_factor_num() -> BigInt {
```
ec2, line 28: Should probably make these lazy statics

vm/actor/src/builtin/miner/mod.rs
timvermeulen: I know that this is insignificant either way but here I deliberately mirrored their variable name to make it easier to see that the logic matches (with the `empty` + `!` double negative)

vm/actor/src/builtin/miner/mod.rs
timvermeulen: Absolutely!

vm/actor/src/builtin/miner/monies.rs
timvermeulen: Good call, I'll do that

vm/actor/src/builtin/miner/mod.rs
ec2: true!

vm/actor/src/builtin/miner/partition_state.rs
```diff
@@ -0,0 +1,730 @@
...
+        self.recovering_power = PowerPair::zero();
+
+        Ok((new_fault_power, failed_recovery_power))
+    }
+
+    pub fn pop_early_terminations<BS: BlockStore>(
+        &mut self,
+        store: &BS,
+        max_sectors: u64,
+    ) -> Result<(TerminationResult, /* has more */ bool), Box<dyn StdError>> {
```
ec2, line 526: Instead of labeling the return value, might be better to doc comment this method as best as you can

ec2: This happens many places throughout too. I'd say to have them as part of the doc comments

vm/actor/src/builtin/miner/vesting_state.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    pub fn add_locked_funds(
+        &mut self,
+        current_epoch: ChainEpoch,
+        vesting_sum: &TokenAmount,
+        proving_period_start: ChainEpoch,
+        spec: VestSpec,
+    ) {
+        // maps the epochs in VestingFunds to their indices in the vec
+        let mut epoch_to_index = HashMap::<ChainEpoch, usize>::new();
```
ec2, line 52: You can create the map with capacity self.funds.len

ec2 (approved):
LGTM. Just some minor things w.r.t. style which i dont feel strongly about.

vm/actor/src/builtin/miner/bitfield_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn cut(&mut self, to_cut: &BitField) -> Result<(), String> {
+        let mut epochs_to_remove = Vec::<u64>::new();
+
+        self.amt
+            .for_each_mut(|epoch, bitfield| {
+                *bitfield = bitfield.cut(to_cut);
+
+                if bitfield.is_empty() {
+                    epochs_to_remove.push(epoch);
+                }
```
austinabell, line 77: There seems to me to be an issue which is that if the bitfield is empty, the indexed bitfield would still be updated when in specs-actors they only update if the resultant bitfield is not empty.

Is this the same logically and I'm just missing something about cut?

vm/actor/src/builtin/miner/bitfield_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    pub fn add_many_to_queue_values(
+        &mut self,
+        values: &HashMap<ChainEpoch, Vec<u64>>,
+    ) -> Result<(), Box<dyn StdError>> {
+        // Update each epoch in-order to be deterministic.
+        // Pre-quantize to reduce the number of updates.
+
+        let mut quantized_values = HashMap::<ChainEpoch, Vec<u64>>::new();
+        let mut updated_epochs = Vec::<ChainEpoch>::new();
```
austinabell, line 101: can set capacity on both of these to the length of values

vm/actor/src/builtin/miner/bitfield_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+        for (&raw_epoch, entries) in values {
+            let epoch = self.quant.quantize_up(raw_epoch);
+            updated_epochs.push(epoch);
+            quantized_values.entry(epoch).or_default().extend(entries);
+        }
+
+        updated_epochs.sort_unstable();
+
+        for epoch in updated_epochs {
+            self.add_to_queue_values(epoch, &quantized_values.remove(&epoch).unwrap_or_default())?;
```
austinabell, line 112: Noting that an unwrap to default would just create a discrepancy, but specs-actors would crash so I am fine with this

vm/actor/src/builtin/miner/bitfield_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        if popped_keys.is_empty() {
+            // Nothing expired.
+            return Ok((BitField::new(), false));
+        }
+
+        // TODO: batch delete
+        for key in popped_keys {
+            self.amt.delete(key)?;
+        }
```
austinabell, line 143: maybe worth just adding the bulk delete function on the amt and calling it with this logic. Just to future proof it so if it does get optimized later this doesn't get missed

austinabell (commented):
I'm going to split up my review by file to break it up

vm/actor/src/builtin/miner/deadline_assignment.rs
```diff
@@ -0,0 +1,185 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use super::{Deadline, SectorOnChainInfo};
+use std::{cmp::Ordering, collections::BinaryHeap};
+
+fn div_rounding_up(dividend: u64, divisor: u64) -> u64 {
```
austinabell, line 7: maybe worth tagging this function with `#[inline]`? Definitely optional

vm/actor/src/builtin/miner/deadline_assignment.rs
```diff
@@ -0,0 +1,185 @@
...
+
+// Assigns partitions to deadlines, first filling partial partitions, then
+// adding new partitions to deadlines with the fewest live sectors.
+pub fn assign_deadlines(
+    partition_size: u64,
+    deadlines: &[Option<Deadline>],
+    sectors: Vec<SectorOnChainInfo>,
+) -> Vec<Vec<SectorOnChainInfo>> {
+    struct Entry {
+        partition_size: u64,
```
austinabell, line 134: You're storing the partition size with every entry, is there a reason you do this instead of wrapping the `BinaryHeap` and just storing the partition size once? Is it just because of the trait bounds of the compares needed in the binary heap?

This isn't functionally a problem but depending on how many deadlines there are, it's just unnecessary duplicated data

vm/actor/src/builtin/miner/deadline_assignment.rs
```diff
@@ -0,0 +1,185 @@
...
+            partition_size,
+            info: DeadlineAssignmentInfo {
+                index,
+                live_sectors: deadline.live_sectors,
+                total_sectors: deadline.total_sectors,
+            },
+        })
+        .collect();
+
+    assert!(!heap.is_empty());
```
austinabell, line 173: why is this assertion okay? Can you be sure that deadlines will always not be empty? I don't see how this is guaranteed, but maybe I'll find the answer as I review more

ec2: I dont think the assertion is really necessary, but it is true that the deadlines array is none empty. See miner_state.rs:422

austinabell: Yeah, I know that, but I wasn't sure if the array of `None` deadlines translated into the deadline_array variable that is passed into the function. I'm checking that file now though

timvermeulen: This definitely needs a comment, let me look into why I put this here

timvermeulen: So as far as I can see, the Go impl implicitly asserts that the heap is non-empty by accessing the top element at https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/deadline_assignment.go#L204

timvermeulen: Which is more or less equivalent to our `heap.peek_mut().unwrap()`. So afaics it really doesn't matter whether that assert is there or not, either way it will be a runtime error if the heap is empty.

vm/actor/src/builtin/miner/deadline_assignment.rs
timvermeulen: I think `#[inline]` only matters between crate boundaries, here the function is called in the same crate (in the same file even) so the compiler is already free to inline it. Unless you're talking about force-inlining it?

vm/actor/src/builtin/miner/partition_state.rs
timvermeulen: In many cases I genuinely don't know exactly what the return value means, this seemed like the most direct way to translate their named return values. I'm not disagreeing but I don't think I have the understanding to write more meaningful doc comments than they already have

vm/actor/src/builtin/miner/deadline_assignment.rs
timvermeulen: Yeah I'm not happy about this either, but `BinaryHeap` requires that the element type is `Ord` rather than being able to pass a custom comparator. So afaik there's literally no other way to do this, since `Ord::cmp` isn't allowed to access it otherwise

vm/actor/src/builtin/miner/deadline_state.rs
```diff
@@ -0,0 +1,998 @@
...
+    ) -> Result<PowerPair, Box<dyn StdError>> {
+        if sectors.is_empty() {
+            return Ok(PowerPair::zero());
+        }
+
+        // First update partitions, consuming the sectors
+        let mut partition_deadline_updates = HashMap::<ChainEpoch, Vec<u64>>::new();
+        let mut new_power = PowerPair::zero();
+        self.live_sectors += sectors.len() as u64;
+        self.total_sectors = sectors.len() as u64;
```
austinabell, line 279:
```suggestion
        self.total_sectors += sectors.len() as u64;
```

vm/actor/src/builtin/miner/deadline_state.rs
```diff
@@ -0,0 +1,998 @@
...
+            if !more {
+                partitions_finished.push(partition_idx);
+            }
+
+            // Save partition
+            partitions
+                .set(partition_idx, partition)
+                .map_err(|e| format!("failed to store partition {}: {:?}", partition_idx, e))?;
+
+            if !result.below_limit(max_partitions, max_sectors) {
```
austinabell, line 391:
```suggestion
            if result.below_limit(max_partitions, max_sectors) {
```

vm/actor/src/builtin/miner/deadline_state.rs
```diff
@@ -0,0 +1,998 @@
...
+            self.faulty_power -= &removed.faulty_power;
+
+            // Aggregate power lost from active sectors
+            power_lost += &removed.active_power;
+        }
+
+        // save partitions back
+        self.partitions = partitions
+            .flush()
+            .map_err(|e| format!("failed to persist partitions: {:?}", e))?;
```
austinabell, line 482: Hmm I'm noticing with these that there could be an out of gas error that would be ignored when translated like this, I think we will have to make a pass through and be able to downcast errors from amt and hamt flushes as well, since go uses the %w xerrors here.

I can open up an issue later, just noting one case for reference

vm/actor/src/builtin/miner/bitfield_queue.rs
timvermeulen: Yep, great call

vm/actor/src/builtin/miner/deadline_state.rs
```diff
@@ -0,0 +1,998 @@
...
+        // Update sector counts.
+        let removed_dead_sectors = dead.len() as u64;
+        let removed_live_sectors = live.len() as u64;
+
+        self.live_sectors -= removed_live_sectors;
+        self.total_sectors -= removed_live_sectors + removed_dead_sectors;
+
+        // Update expiration bitfields.
+        let mut expiration_epochs = BitFieldQueue::new(store, &self.expirations_epochs, quant)
+            .map_err(|e| format!("failed to load expiration queue: {:?}", e))?;
```
austinabell, line 582: here is another case of the previously mentioned. Nothing to change here, just noting

austinabell (changes requested):

vm/actor/src/builtin/miner/deadline_assignment.rs
austinabell: ah, honestly didn't know the compiler could do this automatically

vm/actor/src/builtin/miner/bitfield_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+            return Ok(());
+        }
+
+        let epoch = self.quant.quantize_up(raw_epoch);
+
+        let bitfield = self
+            .amt
+            .get(epoch as u64)
+            .map_err(|e| format!("failed to lookup queue epoch {}: {:?}", epoch, e))?
+            .ok_or_else(|| actor_error!(ErrNotFound; ""))?;
```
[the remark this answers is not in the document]
timvermeulen: My bad, I thought `get` could throw an `ErrNotFound`

vm/actor/src/builtin/miner/bitfield_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        Ok(())
+    }
+
+    pub fn add_to_queue_values(
+        &mut self,
+        epoch: ChainEpoch,
+        values: &[u64],
+    ) -> Result<(), Box<dyn StdError>> {
+        self.add_to_queue(epoch, &values.iter().map(|&i| i as usize).collect())
```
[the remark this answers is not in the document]
timvermeulen: Yeah I assume that's what my reasoning was, creating a bitfield from an empty iterator should be completely allocation-free because `Vec` and `HashMap` don't allocate when they're empty. I'll add the check though, might as well match their code when it behaves the same way

vm/actor/src/builtin/miner/bitfield_queue.rs
timvermeulen: I figured it doesn't matter because empty bitfields are removed from the AMT right after, so it's not important whether we make it empty before removing it or not. Does that make sense?

vm/actor/src/builtin/miner/deadline_state.rs
```diff
@@ -0,0 +1,998 @@
...
+use std::{cmp, collections::HashMap, collections::HashSet, error::Error as StdError};
+
+/// Deadlines contains Deadline objects, describing the sectors due at the given
+/// deadline and their state (faulty, terminated, recovering, etc.).
+#[derive(Serialize_tuple, Deserialize_tuple)]
+pub struct Deadlines {
+    // Note: we could inline part of the deadline struct (e.g., active/assigned sectors)
+    // to make new sector assignment cheaper. At the moment, assigning a sector requires
+    // loading all deadlines to figure out where best to assign new sectors.
+    pub due: Vec<Cid>, // []Deadline
```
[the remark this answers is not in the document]
timvermeulen: I think I tried it but ran into problems because `WPOST_PERIOD_DEADLINES` is greater than 32 😛 So some traits weren't implemented for it (I don't remember which one was the issue). I'll add a TODO to change this to an array later, let me know if you think that's sensible

vm/actor/src/builtin/miner/expiration_queue.rs
```diff
@@ -0,0 +1,785 @@
...
+            &rescheduled_sectors,
+            &PowerPair::zero(),
+            &rescheduled_power,
+            &TokenAmount::zero(),
+        )?;
+
+        // Trim the rescheduled epochs from the queue.
+        for epoch in rescheduled_epochs {
+            self.amt.delete(epoch)?;
+        }
```
austinabell, line 308: noting that if you do change to batch delete, to also switch this one (doesn't have a TODO so I'm commenting)

vm/actor/src/builtin/miner/deadline_state.rs
```diff
@@ -0,0 +1,998 @@
...
+            .map_err(|e| format!("error: {:?}", e))?;
+        Ok(())
+    }
+}
+
+/// Deadline holds the state for all sectors due at a specific deadline.
+#[derive(Serialize_tuple, Deserialize_tuple)]
+pub struct Deadline {
+    // Partitions in this deadline, in order.
+    // The keys of this AMT are always sequential integers beginning with zero.
```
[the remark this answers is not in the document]
timvermeulen: Thanks, missed those

vm/actor/src/builtin/miner/expiration_queue.rs
```diff
@@ -0,0 +1,785 @@
...
+            SectorEpochSet {
+                epoch: expiration,
+                sectors: sector_numbers,
+                power: total_power,
+                pledge: total_pledge,
+            }
+        })
+        .collect();
+
+    sector_epoch_sets.sort_by_key(|epoch_set| epoch_set.epoch);
```
austinabell, line 783: Isn't it possible for two sectors to have the same expiry epoch? In which case we may have a discrepancy with the hashmap ordering differences between rust/go. Thoughts?

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
+// Copyright 2020 ChainSafe Systems
```
austinabell, line 1: Question: why was this file renamed from `state.rs`? It doesn't follow the convention used for other actors, which is just that the `miner_` prefix is redundant because of the module it's in

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+use std::{cmp, error::Error as StdError};
+use vm::{actor_error, ActorError, ExitCode, TokenAmount};
+
+// Balance of Miner Actor should be greater than or equal to
+// the sum of PreCommitDeposits and LockedFunds.
+// It is possible for balance to fall below the sum of PCD, LF and
+// InitialPledgeRequirements, and this is a bad state (IP Debt)
+// that limits a miner actor's behavior (i.e. no balance withdrawals)
+// Excess balance as computed by st.GetAvailableBalance will be
+// withdrawable or usable for pre-commit deposit or pledge lock-up.
```
austinabell, line 29: doc comment?

vm/actor/src/builtin/miner/deadlines.rs
```diff
@@ -55,218 +57,120 @@ impl DeadlineInfo {
...
+            partitions
+                .for_each_while(|i, partition| {
+                    if partition.sectors.get(sector_number as usize) {
+                        partition_idx = Some(i);
+                        Ok(false)
+                    } else {
+                        Ok(true)
+                    }
+                })
+                .map_err(|e| e.to_string())?;
```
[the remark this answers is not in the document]
timvermeulen: Oh I think I ported this code before I fully understood the error dance, this indeed makes no sense

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+
+    /// Optional worker key to update at an epoch
+    pub pending_worker_key: Option<WorkerKeyChange>,
+
+    /// Libp2p identity that should be used when connecting to this miner
+    #[serde(with = "serde_bytes")]
+    pub peer_id: Vec<u8>,
+    /// Slice of byte arrays representing Libp2p multi-addresses used for establishing a connection with this miner.
+    #[serde(with = "serde_bytes")]
+    pub multi_address: Vec<u8>,
```
austinabell, line 932: This should be a 2d vector, a MultiAddress is a byte slice, and it's a vector of them

vm/actor/src/builtin/miner/types.rs
```diff
@@ -117,51 +191,85 @@ pub struct ReportConsensusFaultParams {
...
+    /// Expected one day projection of reward for sector computed at activation time
+    #[serde(with = "bigint_ser")]
+    pub expected_day_reward: TokenAmount,
+    /// Expected twenty day projection of reward for sector computed at activation time
+    #[serde(with = "bigint_ser")]
+    pub expected_storage_pledge: TokenAmount,
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize_tuple, Deserialize_tuple)]
 pub struct ChainSectorInfo {
```
austinabell, line 271: I don't see this type anywhere, was it just kept because it's used in the state api rn?

vm/actor/src/builtin/miner/types.rs
```diff
@@ -9,105 +9,179 @@ use clock::ChainEpoch;
...
 pub const CRON_EVENT_WORKER_KEY_CHANGE: CronEvent = 1;
-pub const CRON_EVENT_PRE_COMMIT_EXPIRY: CronEvent = 2;
-pub const CRON_EVENT_PROVING_PERIOD: CronEvent = 3;
+pub const CRON_EVENT_PROVING_DEADLINE: CronEvent = 2;
+pub const CRON_EVENT_PROCESS_EARLY_TERMINATIONS: CronEvent = 3;
 
 /// Storage miner actor constructor params are defined here so the power actor can send them to the init actor
 /// to instantiate miners.
 #[derive(Serialize_tuple, Deserialize_tuple)]
 pub struct MinerConstructorParams {
```
austinabell, line 22: There is a duplicate of this in `power/types.rs` (go puts it there because there was a circular dep for them, and I only kept it there to not make a conflict with these changes). Can you just delete that other one, or have it alias this to future proof?

vm/actor/src/builtin/miner/types.rs
```diff
@@ -9,105 +9,179 @@ use clock::ChainEpoch;
...
+}
+
+#[derive(Serialize_tuple, Deserialize_tuple)]
+pub struct TerminationDeclaration {
+    pub deadline: u64,
+    pub partition: u64,
     pub sectors: BitField,
 }
+
+pub struct TerminateSectorsReturn {
```
austinabell, line 135: This type should be Serialize/deserialize. I noticed you are not using the return for the `terminate_sectors` function. I will comment on that when I get to that file, but mentioning now

vm/actor/src/builtin/miner/deadline_state.rs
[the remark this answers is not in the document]
austinabell: yeah lets just wait, that's fine. Hopefully the change away from that >32 trait gets removed at some point soon

vm/actor/src/builtin/miner/expiration_queue.rs
timvermeulen: I'm not sure I understand but I think you misread this function, the purpose of `sectors_by_expiration` is to group sectors together with the same expiry epoch. So that guarantees that the `SectorEpochSet`s in `sector_epoch_sets` have distinct epochs.

vm/actor/src/builtin/miner/bitfield_queue.rs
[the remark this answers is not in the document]
austinabell: Just also future proofs it in case something else changes internally

vm/actor/src/builtin/miner/miner_state.rs
timvermeulen: Oh my bad, I figured `state.rs` only made sense while it was the only `_state.rs` file for this actor. I'll change it back.

austinabell: ah, that's a good point, opinionated change then, feel free to leave if you want (just seems asymmetric to me)

vm/actor/src/builtin/miner/types.rs
timvermeulen: I'm not sure what you're suggesting? Should this type be somewhere else?

vm/actor/src/builtin/miner/types.rs
timvermeulen: Ahh yep, got it

vm/actor/src/builtin/miner/types.rs
austinabell: I just don't know what it's for, since I don't see an equivalent in specs-actors or Lotus. Fine to just keep as is and can remove once the state API is fixed. Maybe add a TODO to this to make sure it doesn't get ignored?

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+    pub fn add_pre_commit_expiry<BS: BlockStore>(
+        &mut self,
+        store: &BS,
+        expire_epoch: ChainEpoch,
+        sector_number: SectorNumber,
+    ) -> Result<(), Box<dyn StdError>> {
+        // Load BitField Queue for sector expiry
+        let quant = self.quant_spec_every_deadline();
+        let mut queue = super::BitFieldQueue::new(store, &self.pre_committed_sectors_expiry, quant)
+            .map_err(|e| format!("failed to load pre-commit sector queue: {:?}", e))?;
```
austinabell, line 862: might want to downcast here if you feel it's important to add the message, otherwise might just be better to remove the map_err

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -1,62 +1,87 @@
...
+pub use expiration_queue::*;
+pub use miner_state::*;
+pub use monies::*;
+pub use partition_state::*;
+pub use policy::*;
+pub use quantize::*;
+pub use sector_map::*;
+pub use sectors::*;
+pub use termination::*;
+pub use types::SectorOnChainInfo;
```
austinabell, line 33:
```suggestion
```

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +121,128 @@ impl Actor {
...
             ));
         }
 
-        let owner = resolve_owner_address(rt, params.owner)?;
+        let owner = resolve_control_address(rt, params.owner)?;
         let worker = resolve_worker_address(rt, params.worker)?;
+        let control_addresses: Vec<_> = params
+            .control_addresses
+            .into_iter()
+            .map(|address| resolve_worker_address(rt, address))
```
austinabell, line 139:
```suggestion
            .map(|address| resolve_control_address(rt, address))
```

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +121,128 @@ impl Actor {
...
+            actor_error!(ErrIllegalState, "failed to construct initial state: {}", e)
         })?;
 
-        let empty_deadlines_cid = rt.store().put(&Deadlines::new(), Blake2b256).unwrap();
+        let empty_bitfield_cid = rt.store().put(&BitField::new(), Blake2b256).unwrap();
+
+        let empty_deadline_cid = rt
+            .store()
+            .put(&Deadline::new(empty_array.clone()), Blake2b256)
+            .unwrap();
```
austinabell, line 155: These should not be unwrapped, they could hit an out of gas error, which should not crash

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +121,128 @@ impl Actor {
...
+
+        let empty_deadlines_cid = rt
+            .store()
+            .put(&Deadlines::new(empty_deadline_cid), Blake2b256)
+            .unwrap();
```
austinabell, line 160: same

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +121,128 @@ impl Actor {
...
+
+        let empty_vesting_funds_cid = rt.store().put(&VestingFunds::new(), Blake2b256).unwrap();
```
austinabell, line 162: same

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +121,128 @@ impl Actor {
...
         )
-        .map_err(|e| ActorError::new(ExitCode::ErrIllegalArgument, e))?;
+        .map_err(|e| {
+            actor_error!(
+                ErrIllegalArgument,
+                "failed to construct initial miner info: {}",
+                e
+            )
+        })?;
+        let info_cid = rt.store().put(&info, Blake2b256).unwrap();
```
austinabell, line 193: same here

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +121,128 @@ impl Actor {
...
-                event_type: CRON_EVENT_PROVING_PERIOD,
-                sectors: Default::default(),
+                event_type: CRON_EVENT_PROVING_DEADLINE,
             },
         )?;
 
         Ok(())
     }
 
+    fn get_miner_info<BS, RT>(rt: &mut RT, state: &State) -> Result<MinerInfo, ActorError>
```
austinabell, line 218: You only need the store for this function, so doesn't need to be associated, or pass in a runtime. Only reason go implementation does is because they use it to abort to return the exit code

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +121,128 @@ impl Actor {
...
+    where
+        BS: BlockStore,
+        RT: Runtime<BS>,
+    {
+        state
+            .get_info(rt.store())
+            .map_err(|e| actor_error!(ErrIllegalState, "could not read miner info: {}", e))
```
austinabell, line 225: This should be a downcast, not a mapping (The exit code passed into `RequireNoErr` on go impl is only a default

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -186,37 +251,74 @@ impl Actor {
...
+                    new_worker,
+                    effective_at: effective_epoch,
+                });
+
+                Some(effective_epoch)
+            };
+
+            state
+                .save_info(rt.store(), info)
+                .map_err(|e| actor_error!(ErrIllegalState, "could not save miner info: {:?}", e))?;
```
austinabell, line 288: downcast, not map

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -186,37 +251,74 @@ impl Actor {
...
+            rt.validate_immediate_caller_is(
+                info.control_addresses
+                    .iter()
+                    .chain(&[info.worker, info.owner]),
+            )?;
+
+            info.peer_id = params.new_id;
+            state
+                .save_info(rt.store(), info)
+                .map_err(|e| actor_error!(ErrIllegalState, "could not save miner info: {:?}", e))?;
```
austinabell, line 320: downcast, not map

vm/actor/src/builtin/miner/types.rs
```diff
@@ -9,105 +9,179 @@ use clock::ChainEpoch;
...
 #[derive(Serialize_tuple, Deserialize_tuple)]
 pub struct ChangePeerIDParams {
     #[serde(with = "serde_bytes")]
     pub new_id: Vec<u8>,
 }
+
 #[derive(Serialize_tuple, Deserialize_tuple)]
 pub struct ChangeMultiaddrsParams {
     #[serde(with = "serde_bytes")]
     pub new_multi_addrs: Vec<u8>,
```
austinabell, line 66: This is wrong, but you'll come across it when you change the other usage of it, should be 2d vec

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -230,9 +332,20 @@ impl Actor {
...
+            rt.validate_immediate_caller_is(
+                info.control_addresses
+                    .iter()
+                    .chain(&[info.worker, info.owner]),
+            )?;
+
+            info.multi_address = params.new_multi_addrs;
+            state
+                .save_info(rt.store(), info)
+                .map_err(|e| actor_error!(ErrIllegalState, "could not save miner info: {:?}", e))?;
```
austinabell, line 347: downcast, not map

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -248,163 +361,253 @@ impl Actor {
...
-                ActorError::new(
-                    ExitCode::ErrIllegalState,
-                    format!(
-                        "failed to compute partitions sectors at deadline {}, partitions {:?}",
-                        deadline.index, params.partitions
-                    ),
-                )
-            })?;
+            let sectors = Sectors::new(rt.store(), &state.sectors)
+                .map_err(|e| actor_error!(ErrIllegalState, "failed to load sectors: {:?}", e))?;
```
austinabell, line 463: DE (downcast would be ideal here)

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -248,163 +361,253 @@ impl Actor {
...
-                declared_recoveries.iter().for_each(|i| {
-                    let key = i as u64;
-                    let s = sectors_by_number.get(&key).cloned().unwrap();
-                    recovered_sectors.push(s);
-                });
-                Ok(st.info.sector_size)
-            }
+            state
+                .save_deadlines(rt.store(), deadlines)
+                .map_err(|e| actor_error!(ErrIllegalState, "failed to save deadlines: {}", e))?;
```
austinabell, line 596: these two errors should be downcasted as well (requires previously mentioned change)

austinabell (changes requested):
Going to break this one up because I need to take a break and there are some common issues

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+            .get(&self.allocated_sectors)
+            .map_err(
+                |e| actor_error!(ErrIllegalState; "failed to load allocated sectors bitfield {:?}", e),
+            )?
+            .ok_or_else(|| actor_error!(ErrNotFound; ""))?;
+
+        allocated_sectors |= sector_numbers;
+
+        self.allocated_sectors = store.put(&allocated_sectors, Blake2b256).map_err(
+            |e| actor_error!(ErrIllegalState; "failed to mask allocated sectors bitfield: {:?}", e),
```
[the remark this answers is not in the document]
timvermeulen: You're giving me too much credit here, I just misread their code

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+
+    /// Gets and returns the requested pre-committed sectors, skipping missing sectors.
+    pub fn find_precommitted_sectors<BS: BlockStore>(
+        &self,
+        store: &BS,
+        sector_numbers: &[SectorNumber],
+    ) -> Result<Vec<SectorPreCommitOnChainInfo>, String> {
+        let precommitted =
+            Hamt::<_, SectorPreCommitOnChainInfo>::load(&self.pre_committed_sectors, store)
+                .map_err(|e| e.to_string())?;
```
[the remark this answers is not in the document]
timvermeulen: `Hamt::load` doesn't itself return a dynamic error though, and neither does `make_map_with_root`, so where would that exit code come from?

vm/actor/src/builtin/miner/deadline_state.rs
timvermeulen: I think my logic is correct here, the Go code returns `nil` (i.e. continues iteration) when `below_limit` returns false

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+            result += deadline_result;
+
+            if !more {
+                to_unset.push(i);
+            }
+
+            // Save the deadline
+            deadlines.update_deadline(store, deadline_idx, &deadline)?;
+
+            if !result.below_limit(max_partitions, max_sectors) {
```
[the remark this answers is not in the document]
timvermeulen: Same as before, I think this logic is correct

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+
+    /// Loads sector info for a sequence of sectors, substituting info for a stand-in sector for any that are faulty.
+    fn load_sector_infos_with_fault_mask<BS: BlockStore>(
+        &self,
+        store: &BS,
+        sectors_bf: &BitField,
+        faults: &BitField,
+        fault_stand_in: SectorNumber,
+    ) -> Result<Vec<SectorOnChainInfo>, String> {
+        let sectors = Sectors::new(store, &self.sectors)
```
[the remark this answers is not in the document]
timvermeulen: I wanted to call it that but it also has an instance method called `load`, any suggestions?

vm/actor/src/builtin/miner/mod.rs
timvermeulen: I'm fully aware, I don't really know when to stay close to their implementation and when to deviate, I get that only passing in the store keeps it functionally the same but it also doesn't really gain us anything? Happy to change it either way, just explaining my thought process

vm/actor/src/builtin/miner/miner_state.rs
```diff
@@ -0,0 +1,993 @@
...
+        store: &BS,
+        current_epoch: ChainEpoch,
+        vesting_sum: &TokenAmount,
+        spec: VestSpec,
+    ) -> Result<TokenAmount, Box<dyn StdError>> {
+        let mut vesting_funds = self.load_vesting_funds(store)?;
+
+        // unlock vested funds first
+        let amount_unlocked = vesting_funds.unlock_vested_funds(current_epoch);
+        self.locked_funds -= &amount_unlocked;
```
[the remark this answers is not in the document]
timvermeulen: Not intentional

vm/actor/src/builtin/miner/miner_state.rs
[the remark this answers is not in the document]
austinabell: ah true, didn't see that. I think it's fine the way it is then, or maybe something like `load_new` or the other function be `load_bitfield` idrc, just commenting because it seemed a bit weird to me

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -415,161 +616,278 @@ impl Actor {
...
+                        e
+                    )
+                })?
+                .ok_or_else(|| {
+                    actor_error!(
+                        ErrIllegalState,
+                        "sector {} already pre-committed",
+                        params.sector_number
+                    )
+                })?;
```
austinabell, line 763: This should error on `Some`, not `None` (check is that the sector should not be pre-committed)

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -415,161 +616,278 @@ impl Actor {
...
+                    e.wrap(format!(
+                        "failed to allocate sector id {}",
+                        params.sector_number
+                    ))
+                })?;
+
+            // The following two checks shouldn't be necessary, but it can't
+            // hurt to double-check (unless it's really just too
+            // expensive?).
+            let _ = state
```
austinabell, line 747:
```suggestion
            state
```
`let _` just always opens up to future proof issues

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -415,161 +616,278 @@ impl Actor {
...
+                    },
+                )
+                .map_err(|e| {
+                    actor_error!(
+                        ErrIllegalState,
+                        "failed to write pre-committed sector {}: {:?}",
+                        sector_number,
+                        e
+                    )
+                })?;
```
austinabell, line 855: conceptually should be a downcast

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -580,50 +898,48 @@ impl Actor {
...
-                    format!(
-                        "failed to get precommitted sector: {}, {}",
-                        sector_number, e
-                    ),
+                actor_error!(
+                    ErrIllegalState,
+                    "failed to load precommitted sector: {}, {}",
+                    sector_number,
+                    e
                 )
```
austinabell, line 919: Downcast if possible

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -655,157 +971,255 @@ impl Actor {
...
+        // This skips missing pre-commits.
+        let precommitted_sectors = state
+            .find_precommitted_sectors(rt.store(), &params.sectors)
+            .map_err(|e| {
+                actor_error!(
+                    ErrIllegalState,
+                    "failed to load pre-committed sectors: {}",
+                    e
+                )
+            })?;
```
austinabell, line 1009: downcast if pos

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -655,157 +971,255 @@ impl Actor {
...
-                st.delete_precommitted_sector(rt.store(), num)
-                    .map_err(|e| {
-                        ActorError::new(
-                            ExitCode::ErrIllegalState,
-                            format!("failed to delete precommit for sector {}: {}", num, e),
-                        )
-                    })?;
+            state
+                .put_sectors(store, new_sectors.clone())
+                .map_err(|e| actor_error!(ErrIllegalState, "failed to put new sectors: {}", e))?;
```
austinabell, line 1161: downcast instead of map

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -655,157 +971,255 @@ impl Actor {
...
-                    })?;
+            state
+                .delete_precommitted_sectors(store, &new_sector_numbers)
+                .map_err(|e| {
+                    actor_error!(
+                        ErrIllegalState,
+                        "failed to delete precommited sectors: {:?}",
+                        e
+                    )
+                })?;
```
austinabell, line 1171: dc

vm/actor/src/builtin/miner/miner_state.rs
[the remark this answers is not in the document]
timvermeulen: Okay, renaming the existing `load` to `load_sector`, I like that more

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+                    .entry(decl.deadline)
+                    .or_insert_with(|| {
+                        deadlines_to_load.push(decl.deadline);
+                        Vec::new()
+                    })
+                    .push(decl);
+            }
+
+            let mut sectors = Sectors::new(rt.store(), &state.sectors).map_err(|e| {
+                actor_error!(ErrIllegalState, "failed to load sectors array: {:?}", e)
```
austinabell, line 1337: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+            })?;
+
+            let mut power_delta = PowerPair::zero();
+            let mut pledge_delta = TokenAmount::zero();
+
+            for deadline_idx in deadlines_to_load {
+                let mut deadline = deadlines.load_deadline(store, deadline_idx).map_err(|e| {
```
austinabell, line 1344: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+
+                for &decl in &decls_by_deadline[&deadline_idx] {
+                    let key = PartitionKey {
+                        deadline: deadline_idx,
+                        partition: decl.partition,
+                    };
+
+                    let mut partition = partitions
+                        .get(decl.partition)
+                        .map_err(|e| {
```
austinabell, line 1370: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+                            "failed to update sectors {:?}: {}",
+                            decl.sectors,
+                            e
+                        )
+                    })?;
+
+                    // Remove old sectors from partition and assign new sectors.
+                    let (partition_power_delta, partition_pledge_delta) = partition
+                        .replace_sectors(store, &old_sectors, &new_sectors, info.sector_size, quant)
+                        .map_err(|e| {
```
austinabell, line 1422: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+                                "failed to replaces sector expirations at {:?}: {}",
+                                key,
+                                e
+                            )
+                        })?;
+
+                    power_delta += &partition_power_delta;
+                    pledge_delta += partition_pledge_delta; // expected to be zero, see note below.
+
+                    partitions.set(decl.partition, partition).map_err(|e| {
```
austinabell, line 1434: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+                        ActorError::downcast(
+                            e,
+                            ExitCode::ErrIllegalState,
+                            format!("failed to declare faults for deadline {}", deadline_idx),
+                        )
+                    })?;
+
+                deadlines
+                    .update_deadline(store, deadline_idx, &deadline)
+                    .map_err(|e| {
```
austinabell, line 1733: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+                        ErrIllegalArgument,
+                        "failed fault declaration at deadline {}: {}",
+                        deadline_idx,
+                        e
                     )
                 })?;
 
-            Ok(())
-        })
+                let mut deadline = deadlines.load_deadline(store, deadline_idx).map_err(|e| {
```
austinabell, line 1703: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -839,397 +1261,848 @@ impl Actor {
...
+                            e
+                        )
+                    })?;
+
+                new_fault_power_total += &new_faulty_power;
+            }
+
+            state
+                .save_deadlines(store, deadlines)
+                .map_err(|e| actor_error!(ErrIllegalState, "failed to save deadlines: {}", e))?;
```
austinabell, line 1747: dc

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -1347,647 +2230,524 @@ impl Actor {
...
     Ok(())
 }
 
-fn terminate_sectors<BS, RT>(
-    rt: &mut RT,
-    sector_nos: &BitField,
-    termination_type: SectorTermination,
-) -> Result<(), ActorError>
+fn validate_replace_sector<BS, RT>(
+    _rt: &RT,
```
austinabell, line 2680: just remove? They were only using to throw errors

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -2248,65 +3010,151 @@ where
...
+        MarketMethod::VerifyDealsForActivation as u64,
+        Serialized::serialize(VerifyDealsForActivationParamsRef {
+            deal_ids,
+            sector_start,
+            sector_expiry,
+        })?,
+        TokenAmount::zero(),
+    )?;
+
+    Ok(serialized.deserialize().unwrap())
```
austinabell, line 3035: dont need unwrap, can just `?` as it will convert to a serialize error (I see you are trying to match their behaviour, but when they panic with a serialize error, they recover into the ActorError/ExitCode needed

types/src/sector/mod.rs
```diff
@@ -17,6 +17,10 @@ use vm::ActorID;
 
 pub type SectorNumber = u64;
 
+/// The maximum assignable sector number.
+/// Raising this would require modifying our AMT implementation.
+pub const MAX_SECTOR_NUMBER: SectorNumber = SectorNumber::MAX;
```
austinabell, line 22: This should be the max number from the Amt, not u64 max. The max for the Amt is slightly less (Max i64), probably best to just use the value directly from the Amt to future proof though

vm/actor/src/builtin/miner/monies.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// Computes the pledge requirement for committing new quality-adjusted power to the network, given the current
+// total power, total pledge commitment, epoch block reward, and circulating token supply.
+// In plain language, the pledge requirement is a multiple of the block reward expected to be earned by the
+// newly-committed power, holding the per-epoch block reward constant (though in reality it will change over time).
+pub fn initial_pledge_for_power(
+    qa_power: &StoragePower,
+    baseline_power: &StoragePower,
+    _network_total_pledge: &TokenAmount,
```
austinabell, line 159: I think preference would be to remove this

vm/actor/src/builtin/miner/mod.rs
timvermeulen: Ah, very good to know

vm/actor/src/builtin/miner/partition_state.rs
```diff
@@ -0,0 +1,730 @@
...
+
+impl ops::Neg for PowerPair {
+    type Output = PowerPair;
+
+    fn neg(self) -> Self::Output {
+        PowerPair {
+            raw: -self.raw,
+            qa: -self.qa,
+        }
+    }
```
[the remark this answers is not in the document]
timvermeulen: `BigInt` has a `Neg` impl that just flips the sign without cloning anything, so doing this is definitely more efficient than calling the `&PowerPair` one. Having said that, the `Neg` impl of `&PowerPair` should probably call this one instead, so I'll do that

vm/actor/src/builtin/power/mod.rs
```diff
@@ -95,10 +96,10 @@ impl Actor {
                 Serialized::serialize(MinerConstructorParams {
                     owner: params.owner,
                     worker: params.worker,
+                    control_addresses: Default::default(),
                     seal_proof_type: params.seal_proof_type,
-                    peer: params.peer,
-                    multiaddrs: params.multiaddrs,
-                    control_addrs: Default::default(),
+                    peer_id: params.peer.0,
+                    multi_address: params.multiaddrs,
```
austinabell, line 102: This name should be changed (it's plural addresses). The type is still incorrectly just one slice of bytes

vm/actor/src/builtin/power/types.rs
```diff
@@ -18,17 +18,6 @@ pub const SECTOR_TERMINATION_MANUAL: SectorTermination = 1;
...
 
-// TODO revisit on miner actor completion if duplicate type
-#[derive(Serialize_tuple, Deserialize_tuple)]
-pub struct MinerConstructorParams {
-    pub owner: Address,
-    pub worker: Address,
-    pub control_addrs: Vec<Address>,
-    pub seal_proof_type: RegisteredSealProof,
-    pub peer: BytesDe,
-    pub multiaddrs: Vec<BytesDe>,
```
austinabell, line 29: Just to indicate previous comment what the type should look like. (You don't need to use the `BytesDe` type, I just did out of convenience)

austinabell (changes requested):

vm/actor/src/builtin/miner/mod.rs
austinabell: This is still not updated

vm/actor/src/builtin/miner/mod.rs
```diff
@@ -94,90 +120,127 @@ impl Actor {
...
+            .store()
+            .put(&Deadline::new(empty_array.clone()), Blake2b256)
+            .unwrap();
+
+        let empty_deadlines_cid = rt
+            .store()
+            .put(&Deadlines::new(empty_deadline_cid), Blake2b256)
+            .unwrap();
+
+        let empty_vesting_funds_cid = rt.store().put(&VestingFunds::new(), Blake2b256).unwrap();
```
austinabell, line 161: These errors are still not downcasted, and no reason not too, since a direct interaction with the store. It's okay for the error to be incorrect until I make another pass, but I'd really like to avoid a potential crash when using.

Can probably downcast all of these errors at once if you put this logic in a smaller scope

austinabell (approved):

Issue #692, opened 2020-09-09
Implement peer manager stats for each peer


austinabell:
**Task summary**

The peer manager currently just takes the first peer it iterates over in the hashmap (and is currently being changed to be a random sample) but it should be a random sample of the top `x` peers to avoid sending requests to peers that have a lot of failures.

This is blocked by #690 as there are a lot of changes to similar things.

**Specification reference**

- 

**Other information and links**

PR #693, opened 2020-09-09
ChainSync refactor

### LINKED ISSUE #690: Refactor ChainSyncer to handle initial sync asyncronously
**Task summary**

Currently events and syncing is done syncronously, which causes the node to stall at a certain point after running (depending on config and network). This needs to change to allow network events to be polled and handled to update the sync and syncing state.

The main issue is that the network channel is filling up with gossipsub messages that aren't being polled off and this stalls the libp2p and syncing threads.

**Specification reference**

- 

austinabell:
**Summary of changes**
- Handles syncing asynchronously through sync workers on separate tasks
  - All syncing logic now belongs in the `SyncWorker` struct, and the `ChainSyncer` just handles network events and scheduling/ handling sync workers
  - Separating some of the fields so they are threadsafe across tasks meant there was some auxiliary updates
- Updates peer selection for blocksync request to be random (should be made better with #692)
- Fixes some bugs with tipset scheduling and other functions

So the loading/ fetching of tipsets from hello protocol is done in separate threads as to not block the `ChainSyncer` polling or the network event channel will fill and stall (what was happening before). This can be handled better but I want to keep the scope of this PR to a minimum as it's already changing a lot.

Closes #690 


blockchain/chain/src/store/chain_store.rs
```diff
@@ -71,44 +74,50 @@ where
             .map(Arc::new);
         Self {
             db,
-            publisher: Publisher::new(SINK_CAP),
+            publisher: Publisher::new(SINK_CAP).into(),
             tip_index: TipIndex::new(),
-            heaviest,
+            heaviest: heaviest.into(),
```
ec2, line 79: Woah! Didn't know RwLock implemented into!!!

node/forest_libp2p/src/service.rs
```diff
@@ -276,6 +276,13 @@ where
         self.network_receiver_out.clone()
     }
 }
+async fn emit_event(sender: &Sender<NetworkEvent>, event: NetworkEvent) {
+    if !sender.is_full() {
+        sender.send(event).await
+    } else {
+        warn!("network sender channel was full, ignoring event");
```
ec2, line 283: Oof, is this okay? Is there anyway to avoid this? How often do you see us hitting his with the channel size increased to 50? I see this potentially being a problem when we have like a bajillion peers on a production network

ec2 (commented):
It seems like while in bootstrap phase, every hello we get, we'll be doing a sync from that Tipset? So if there are 3 Hello messages, they all get scheduled? Or am I misreading?

austinabell:
> It seems like while in bootstrap phase, every hello we get, we'll be doing a sync from that Tipset? So if there are 3 Hello messages, they all get scheduled? Or am I misreading?

No, only tipsets that aren't connected get scheduled, which is the point of having multiple workers (so one doesn't get sidetracked with an invalid chain and have to start from scratch again)

node/forest_libp2p/src/service.rs
austinabell: yeah, I can add a TODO to revisit (or maybe what can be done is keeping a vector buffer of events to send out once the channel empties). The issue with this is that if this channel fills and this send awaits, then the libp2p thread will be blocked and the whole node stalls. 

For now I think ignoring an event is fine (I haven't hit this warning yet with the current code, but let me know if you do).

ec2: Just afraid of when we hit production systems. For reference, on space race, lotus takes 2-4 seconds to process a block. But should be ok for now

ec2:
> No, only tipsets that aren't connected get scheduled, which is the point of having multiple workers (so one doesn't get sidetracked with an invalid chain and have to start from scratch again)

What do you mean by connected here? If i understand, connected means common ancestor? 

Lets say we receive a Hello from PeerA. We start doing a reverse header sync. Half way through a reverse header sync, we receive another Hello from PeerB. PeerB will probably have a newer head than PeerA. If we start to do a reverse header sync on PeerB as well, in most cases (assuming no attacks or weird things which I think is the most likely and far more common scenario), we will have a large amount of the same fetches. This is because a reverse header sync can take quite a bit of time if the chain is long, and the blocks dont get persisted into the store until ALL the blocks from head -> 0 are fetched, so WorkerB fetching from PeerB will not know that WorkerA fetching from PeerA is already fetching the same blocks.

austinabell:
> > No, only tipsets that aren't connected get scheduled, which is the point of having multiple workers (so one doesn't get sidetracked with an invalid chain and have to start from scratch again)
> 
> What do you mean by connected here? If i understand, connected means common ancestor?
> 
> Lets say we receive a Hello from PeerA. We start doing a reverse header sync. Half way through a reverse header sync, we receive another Hello from PeerB. PeerB will probably have a newer head than PeerA. If we start to do a reverse header sync on PeerB as well, in most cases (assuming no attacks or weird things which I think is the most likely and far more common scenario), we will have a large amount of the same fetches. This is because a reverse header sync can take quite a bit of time if the chain is long, and the blocks dont get persisted into the store until ALL the blocks from head -> 0 are fetched, so WorkerB fetching from PeerB will not know that WorkerA fetching from PeerA is already fetching the same blocks.

It's all based on the hello messages received, they get grouped as they come in, so as long as there isn't a break in between two epochs (probably rare) they will be considered on the same chain. This is determined by checking if either the tipset parent is the other or if they are the same. 

This logic existed, but is used more correctly than before (can still be improved on later). This is the point of having those tipset buckets (to track which tipsets are being synced and what to assign another worker to)

node/forest_libp2p/src/service.rs
austinabell: But the thread that polls this channel is not blocked by anything, that's the point of this PR. All of that logic will be done asyncronously. (This is why even the tipset retrieval is done in a separate task, because otherwise the channel fills and stalls)

blockchain/chain_sync/src/sync.rs
```diff
@@ -136,306 +128,226 @@ where
     }
 
     /// Returns the atomic reference to the syncing state.
-    pub fn sync_state_cloned(&self) -> Arc<RwLock<SyncState>> {
-        self.state.clone()
+    pub fn sync_state_cloned(&self) -> WorkerState {
```
ec2, line 131: Fix the commend for this funcc

ec2 (approved):
it L(Real)GTM. I noticed that currently we dont verify the winning post proof. Cant remember why that is, but that can be addressed in another PR.  Also nice one getting rid of the network_handler. I always got mixed up between that and the context when i was heads deep.

blockchain/chain_sync/src/sync.rs
austinabell: I mean, it is still correct? The type is just aliased because it's verbose

ec2: Hm, yeah youre right. I got a little confused. This is the nested arc and rwlock madness. 
Nit: but i would change it to specify its the worker sync state and not the overall chainsync sync state. Not a real big issue, ignore if you want

blockchain/chain/src/store/chain_store.rs
```diff
@@ -71,44 +74,50 @@ where
...
+    pub async fn subscribe(&self) -> Subscriber<HeadChange> {
+        self.publisher.write().await.subscribe()
     }
 
     /// Sets tip_index tracker
-    pub fn set_tipset_tracker(&mut self, header: &BlockHeader) -> Result<(), Error> {
-        let ts: Tipset = Tipset::new(vec![header.clone()])?;
+    // TODO this is really broken, should not be setting the tipset metadata to a tipset with just
+    // the one header.
+    pub async fn set_tipset_tracker(&self, header: &BlockHeader) -> Result<(), Error> {
```
dutterbutter, line 103: I don't really see the need for the `tip_index.rs` file anymore. I recall implementing it really early thinking it would be good for tipset data retrievals but it doesn't look like we are using it anywhere for that purpose? We set the tracker but never make use of it?

blockchain/chain_sync/src/sync.rs
```diff
@@ -6,54 +6,52 @@ mod peer_test;
...
-use std::collections::{BTreeMap, HashMap};
-use std::convert::{TryFrom, TryInto};
+use log::{debug, warn};
+use message::{SignedMessage, UnsignedMessage};
+use state_manager::StateManager;
 use std::sync::Arc;
-use vm::TokenAmount;
+
+// TODO revisit this type, necessary for two sets of Arc<Mutex<>> because each state is
+// on seperate thread and needs to be mutated independently, but the vec needs to be read
```
dutterbutter, line 32: Nit: seperate -> separate

blockchain/chain_sync/src/sync.rs
```diff
@@ -136,306 +128,226 @@ where
...
-        };
-        let target_weight = fts.weight();
-
-        if target_weight.gt(&best_weight) {
-            self.set_peer_head(peer, Arc::new(fts.to_tipset())).await?;
+        let candidate_ts = self
+            .chain_store
+            .heaviest_tipset()
+            .await
+            .map(|heaviest| ts.weight() >= heaviest.weight())
```
dutterbutter, line 273: Shouldn't this just be `ts.weight() > heaviest.weight()`? Do we want to validate and set peer head when weight is equal?

blockchain/chain_sync/src/sync.rs
```diff
@@ -136,306 +128,226 @@ where
...
+        }
+        .spawn(channel)
+        .await
     }
 
     /// informs the syncer about a new potential tipset
     /// This should be called when connecting to new peers, and additionally
     /// when receiving new blocks from the network
-    pub async fn inform_new_head(&mut self, peer: PeerId, fts: &FullTipset) -> Result<(), Error> {
+    pub async fn inform_new_head(&mut self, peer: PeerId, ts: &FullTipset) -> Result<(), Error> {
```
dutterbutter, line 254: I noticed in the Lotus implementation they are adding the `peerId` to their `ChainExchange` pool, should there be a TODO to add this peer to the available peers for blocksync peers?

dutterbutter (changes requested):
Awesome job! Great separating out the `sync_worker` logic away from `sync.rs`. Just a few nits and more questions than anything.

blockchain/chain/src/store/chain_store.rs
austinabell: Yeah I was planning on leaving it for now because out of scope of this and the changes were already a lot

blockchain/chain_sync/src/sync.rs
austinabell: Nah, intentional that if the weight is the same it will also be considered. Also the logic matches Lotus (Although they setup the function differently)

blockchain/chain_sync/src/sync.rs
austinabell: we're doing it when the hello message is handled by the syncer. I don't see why it would need to be handled after this process

dutterbutter (approved):

PR #618, opened 2020-08-11
Rpc state implementation

AshantiMutinta:
**Summary of changes**
Changes introduced in this pull request:
- Implements RPC Layer functions for the state
-Introduced Bitfield and Address Serialization

**Reference issue to close (if applicable)**

Closes #528 

**Other information and links**

PR #694, opened 2020-09-10
Fix StateManager use in different components

ec2:
**Summary of changes**
- While working on the message pool, it seems like i need to integrate the StateManager into the Provider. This led me to see that we were creating a new  StateManager in ChainSyncer and in the RPC. And that seemed wrong.
- Creates StateManager in the node startup, and shares it with ChainSyncer, RPC, and Mpool. 

Closes 

### DIFF
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -229,7 +229,6 @@ where
             self.state.write().await.error(e.to_string());
             return Err(e.into());
         }
-
         // Sync and validate messages from fetched tipsets
         self.set_stage(SyncStage::Messages).await;
         if let Err(e) = self.sync_messages_check_state(&tipsets).await {
@@ -247,7 +246,7 @@ where
     /// Syncs messages by first checking state for message existence otherwise fetches messages from blocksync
     async fn sync_messages_check_state(&mut self, ts: &[Tipset]) -> Result<(), Error> {
         // see https://github.com/filecoin-project/lotus/blob/master/build/params_shared.go#L109 for request window size
-        const REQUEST_WINDOW: i64 = 200;
+        const REQUEST_WINDOW: i64 = 1;
         // TODO refactor type handling
         // set i to the length of provided tipsets
         let mut i: i64 = i64::try_from(ts.len())? - 1;
@@ -270,9 +269,13 @@ where
                         let idx = i - batch_size;
                         let next = &ts[idx as usize];
                         let req_len = batch_size + 1;
-
+                        debug!(
+                            "BlockSync message sync tipsets: epoch: {}, len: {}",
+                            next.epoch(),
+                            req_len
+                        );
                         // receive tipset bundle from block sync
-                        let ts_bundle = self
+                        let mut ts_bundle = match self
                             .network
                             .blocksync_request(
                                 peer_id,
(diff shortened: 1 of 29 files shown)


austinabell:
came in with #618, I assumed it was intentional to have it separate. @AshantiMutinta want to weigh in on this?

I do think that the state calculation cache should be shared, but the subscriber channel cloned on the RPC side worries me. Since those head change events aren't polled off on the RPC side, the receiver will fill and block the sender. Why is that subscriber field added? It isn't used right now

blockchain/chain_sync/src/sync.rs
```diff
@@ -1075,6 +1075,7 @@ mod tests {
         (
             ChainSyncer::new(
                 chain_store,
+                Arc::new(StateManager::new(db.clone())),
```
austinabell, line 1078:
```suggestion
                Arc::new(StateManager::new(db)),
```

blockchain/chain_sync/src/sync/peer_test.rs
```diff
@@ -31,8 +32,10 @@ fn peer_manager_update() {
 
     let genesis_ts = Tipset::new(vec![dummy_header]).unwrap();
     let beacon = Arc::new(MockBeacon::new(Duration::from_secs(1)));
+    let state_manager = Arc::new(StateManager::new(db.clone()));
```
austinabell, line 35:
```suggestion
    let state_manager = Arc::new(StateManager::new(db));
```

forest/src/daemon.rs
```diff
@@ -102,14 +106,13 @@ pub(super) async fn start(config: Config) {
...
     let rpc_task = if config.enable_rpc {
-        let db_rpc = StateManager::new(Arc::clone(&db));
         let keystore_rpc = Arc::clone(&keystore);
         let rpc_listen = format!("127.0.0.1:{}", &config.rpc_port);
         Some(task::spawn(async move {
             info!("JSON RPC Endpoint at {}", &rpc_listen);
             start_rpc(
                 RpcState {
-                    state_manager: db_rpc,
+                    state_manager: Arc::clone(&state_manager),
```
austinabell, line 115:
```suggestion
                    state_manager,
```

node/rpc/src/sync_api.rs
```diff
@@ -122,14 +124,14 @@ mod tests {
...
-            let provider = MpoolRpcProvider::new(subscriber, cs.db);
+            let provider = MpoolRpcProvider::new(subscriber, state_manager.clone());
             MessagePool::new(provider, "test".to_string())
                 .await
                 .unwrap()
         });
 
         let state = Arc::new(RpcState {
-            state_manager: StateManager::new(Arc::new(MemoryDB::default())),
+            state_manager: state_manager.clone(),
```
austinabell, line 134:
```suggestion
            state_manager,
```

austinabell (commented):
remove clones and I'll be good with this and I'll just resolve conflicts on my PR

austinabell (approved):

nannick (approved):

AshantiMutinta (approved):

AshantiMutinta:
This should be fine, I think having these two separate was an oversight. Most state operations are just communicating with internal operations. In this way, we can even look into implementing caching mechanisms across the board

PR #697, opened 2020-09-10
Drand ignore env variable

### LINKED ISSUE #642: Flag to ignore Drand for testing
Lotus has a flag LOTUS_IGNORE_DRAND that skips verification for Drand. This is useful for testing purposes.

nannick:
**Summary of changes**
- Introduces drand flag in command line and config file
- Prevents Drand block verification when Drand flag set to false

Closes 
#642 

### DIFF
--- a/blockchain/beacon/src/drand.rs
+++ b/blockchain/beacon/src/drand.rs
@@ -17,6 +17,7 @@ use std::error;
 /// Default endpoint for the drand beacon node.
 pub const DEFAULT_DRAND_URL: &str = "https://api.drand.sh";
 
+/// Enviromental Variable to ignore Drand. Lotus parallel is LOTUS_IGNORE_DRAND
 pub const IGNORE_DRAND_VAR: &str = "IGNORE_DRAND";
 
 /// Coeffiencients of the publicly available Drand keys.
--- a/blockchain/beacon/src/drand.rs
+++ b/blockchain/beacon/src/drand.rs
@@ -17,6 +17,8 @@ use std::error;
 /// Default endpoint for the drand beacon node.
 pub const DEFAULT_DRAND_URL: &str = "https://api.drand.sh";
 
+pub const IGNORE_DRAND_VAR: &str = "IGNORE_DRAND";
+
 /// Coeffiencients of the publicly available Drand keys.
 /// This is shared by all participants on the Drand network.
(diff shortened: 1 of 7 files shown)


austinabell:
This should be an env variable, not set through a CLI flag. For something like this, it's not worth passing around and storing the boolean just for these checks

blockchain/chain_sync/src/sync.rs
```diff
@@ -699,9 +699,11 @@ where
             &self.chain_store.tipset_from_keys(header.parents())?,
         )?;
 
-        header
-            .validate_block_drand(Arc::clone(&self.beacon), prev_beacon)
-            .await?;
+        if std::env::var("LOTUS_IGNORE_DRAND") == Ok("1".to_owned()) {
```
austinabell, line 702: This probably doesn't need to be `LOTUS_`.. Can just trim that or have it prefixed with forest

austinabell:
You need the check inside of `latest_beacon_entry` to return a static beacon entry in the case of drand being ignored.

This is actually the more important one, because otherwise the node would just error whenever a block is being validated because the retrieval of a beacon entry would fail

blockchain/chain/src/store/chain_store.rs
```diff
@@ -294,6 +294,10 @@ pub fn latest_beacon_entry<DB>(db: &DB, ts: &Tipset) -> Result<BeaconEntry, Erro
 where
     DB: BlockStore,
 {
+    if std::env::var("IGNORE_DRAND") == Ok("1".to_owned()) {
+        return Ok(BeaconEntry::default());
+    }
```
austinabell, line 299: nah this check should happen at the end of checking for drand entries. and can you have the bytes be the same as Lotus for interop reasons (if we want to interop without drand with both clients)

blockchain/beacon/src/drand.rs
```diff
@@ -17,6 +17,8 @@ use std::error;
 /// Default endpoint for the drand beacon node.
 pub const DEFAULT_DRAND_URL: &str = "https://api.drand.sh";
 
+pub const IGNORE_DRAND_VAR: &str = "IGNORE_DRAND";
```
austinabell, line 20: comment would be nice

austinabell (approved):

ec2 (approved):

PR #703, opened 2020-09-15
Semantic Validation for Messages

ec2:
**Summary of changes**
- It's in the title!

Closes 


vm/message/src/lib.rs
```diff
@@ -46,4 +46,6 @@ pub trait Message {
     fn set_gas_fee_cap(&mut self, cap: TokenAmount);
     /// sets the gas premium
     fn set_gas_premium(&mut self, prem: TokenAmount);
+    /// semantic validation of the message
+    fn valid_for_block_inclusion(&self, min_gas: i64) -> Result<(), String>;
```
austinabell, line 50: Is there a reason this needs to be part of the message trait? Isn't this only needed for unsigned messages?

vm/message/src/unsigned_message.rs
```diff
@@ -164,6 +165,34 @@ impl Message for UnsignedMessage {
         let total: TokenAmount = self.gas_fee_cap() * self.gas_limit();
         total + self.value()
     }
+    fn valid_for_block_inclusion(&self, min_gas: i64) -> Result<(), String> {
+        if self.version != 0 {
+            return Err(format!("Message version: {} not  supported", self.version));
+        }
+        if self.value < 0.into() {
```
austinabell, line 172:
```suggestion
        if self.value.sign() == Sign::Minus {
```

Is better than constructing a bigint to compare with

vm/message/src/unsigned_message.rs
```diff
@@ -164,6 +165,34 @@ impl Message for UnsignedMessage {
...
+            return Err("message value cannot be negative".to_string());
+        }
+        if self.value > TOTAL_FILECOIN.into() {
+            return Err("message value cannot be greater than total FIL supply".to_string());
+        }
+        if self.gas_fee_cap < 0.into() {
```
austinabell, line 178: same with this one

vm/message/src/unsigned_message.rs
```diff
@@ -164,6 +165,34 @@ impl Message for UnsignedMessage {
...
+            return Err("gas_fee_cap cannot be negative".to_string());
+        }
+        if self.gas_premium > self.gas_fee_cap {
```
austinabell, line 181: You're missing a check to make sure the premium is not negative

vm/message/src/lib.rs
ec2: It will be used on signed messages in the Mpool

ec2: Is there a reason against putting it in the trait?

austinabell: alrighty, but the check only needs the unsigned message, I'm fine with how it is, but consider just calling `.message()` on the signed message when you need this instead of expanding this trait (already getting bulky and this doesn't seem like it should be part of the trait).

austinabell: > Is there a reason against putting it in the trait?

not syntactically, but logically it doesn't make sense to me, might be opinionated

austinabell (approved):
LGTM, other change is pretty opinionated and can be altered later if needed

vm/message/src/lib.rs
ec2: hehe, i like it in the trait more. im gunna leave it for now.

vm/message/Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde = { version = "1.0", features = ["derive"] }
 base64 = { version = "0.12.1", optional = true }
 forest_json_utils = { path = "../../utils/json_utils", optional = true, version = "0.1" }
 num-traits = "0.2"
+types = { package = "fil_types", path = "../../types"}
```
dutterbutter, line 25: Nit: `types = { package = "fil_types", path = "../../types" }` add space in the back 🙃

dutterbutter (approved):

vm/message/Cargo.toml
ec2: i hate you. need to run ci again now 😂

PR #710, opened 2020-09-21
Adding CircSupply Calculations

### LINKED ISSUE #677: Update total fil circ supply calculations
**Task summary**

Was changed with https://github.com/filecoin-project/lotus/pull/2961 for the function to live outside of the VM. Since this pattern is memory unsafe, we will have to implement differently or just keep the way it is and just update for functional changes.

There was also at least this fix https://github.com/filecoin-project/lotus/pull/2605, so the logic should be checked against Lotus master to make sure we can interop

Edit: Just to be clear, there could very well not be any difference functionally, but this should be verified as there are no tests currently pinning this functionality to the go implementation's logic

**Specification reference**

- 

nannick:
**Summary of changes**
- Ports over CircSupply related functionality from Lotus 

Closes  #677 


blockchain/state_manager/src/lib.rs
```diff
@@ -69,18 +96,20 @@ where
     /// constructor
     pub fn new(bs: Arc<DB>) -> Self {
         Self {
-            bs,
+            cs: ChainStore::new(bs),
```
austinabell, line 99: You can't just create a ChainStore that is independent from the actual `ChainStore` and expect it to work. If you need to have a chainstore attached here, you should be passing it in as an `Arc` of the actual one, to avoid duplication and bugs with that

blockchain/state_manager/src/lib.rs
```diff
@@ -69,18 +96,20 @@ where
...
             subscriber: None,
+            genesis_info: None,
         }
     }
 
     // Creates a constructor that passes in a HeadChange subscriber and a back_search subscriber
     pub fn new_with_subscribers(bs: Arc<DB>, chain_subs: Subscriber<HeadChange>) -> Self {
         Self {
-            bs,
+            cs: ChainStore::new(bs),
```
austinabell, line 109: same here

austinabell (changes requested):
This logic should be replacing that in `total_fil_circ_supply`, this isn't connected in any way.

Also there is a lot of duplicate logic, try to create abstractions to reduce this, somewhat like how it is done in `total_fil_circ_supply` currently

ec2:
Building on what Austin said, I dont see why you need the ChainStore in the first place? I dont see any indications that the ChainStore is needed. Every time  `cs` is used, you're using `cs.db`, which indicates to me you really only want the database (which is already in the StateManager) and not the ChainStore.

blockchain/state_manager/src/utils.rs
```diff
@@ -330,3 +333,63 @@ where
     .map_err(|e| Error::Other(e.to_string()))?;
     Ok(miners)
 }
+
+pub fn get_fil_mined<DB: BlockStore>(state_tree: &StateTree<DB>) -> Result<TokenAmount, Error> {
+    let reward_actor = state_tree
```
ec2, line 338: You can call load_actor_state instead of all this

ec2 (changes requested):

types/src/lib.rs
```diff
@@ -66,3 +66,5 @@ impl NetworkParams for DevnetParams {
 
 /// Space-race specific fork
 pub const UPGRADE_SMOKE_HEIGHT: i64 = 51000;
+pub const UPGRADE_IGNITION_HEIGHT: i64 = -2;
+pub const UPGRADE_LIFTOFF_HEIGHT: i64 = -3;
```
austinabell, line 70: These are fake params given to new networks, if you're going to use the fork upgrade logic, this should be correct.

vm/actor/src/builtin/power/state.rs
```diff
@@ -81,6 +81,10 @@ impl State {
         }
     }
 
+    pub fn total_locked(&self) -> TokenAmount {
+        self.total_pledge_collateral.clone()
```
austinabell, line 85:
```suggestion
    pub fn total_locked(&self) -> &TokenAmount {
        &self.total_pledge_collateral
```

vm/actor/src/builtin/reward/state.rs
```diff
@@ -123,6 +123,10 @@ impl State {
         self.this_epoch_reward_smoothed =
             filter_reward.next_estimate(&self.this_epoch_reward, delta);
     }
+
+    pub fn total_storage_power_reward(&self) -> TokenAmount {
+        self.total_mined.clone()
```
austinabell, line 128:
```suggestion
    pub fn total_storage_power_reward(&self) -> &TokenAmount {
        &self.total_mined
```

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    return_value += &pre_ignition.genesis_pledge + &pre_ignition.genesis_market_funds;
+    Ok(return_value)
+}
+
+pub fn get_fil_mined<DB: BlockStore>(state_tree: &StateTree<DB>) -> Result<TokenAmount, String> {
+    let reward_actor = get_actor_state(state_tree, &*REWARD_ACTOR_ADDR)?;
+    let reward_state: reward::State = state_tree
+        .store()
+        .get(&reward_actor.code)
```
austinabell, line 75: you're loading the actor code? There is nothing in store for this

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,227 @@
...
+        .map_err(|e| e.to_string())?
```
austinabell, line 76: why map the string if only one type of error?

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,227 @@
...
+
+pub fn get_fil_market_locked<DB: BlockStore>(
+    state_tree: &StateTree<DB>,
+) -> Result<TokenAmount, String> {
+    let market_actor = get_actor_state(state_tree, &*STORAGE_MARKET_ACTOR_ADDR)?;
+
+    let market_state: market::State = state_tree
+        .store()
+        .get(&market_actor.state)
+        .map_err(|e| e.to_string())?
```
austinabell, line 90: Same here, I'll stop commenting on all of these

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,227 @@
...
+            initial_balance: initial_balance * FILECOIN_PRECISION,
+            start_epoch: UPGRADE_LIFTOFF_HEIGHT,
+            unlock_duration,
+            pending_txs: Cid::default(),
+        };
+        post_ignition.genesis_msigs.push(ms);
+    }
+
+    Ok(post_ignition)
+}
```
austinabell, line 227: First, this adds a lot of complexity to handle the upgrade logic, and this isn't done correctly. FIrst, we don't have any use of GenesisInfo, but also you're porting over go specific things that will cause issues for us.

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    for (unlock_duration, initial_balance) in totals_by_epoch {
+        let ms = multisig::State {
+            signers: vec![],
+            num_approvals_threshold: 0,
+            next_tx_id: multisig::TxnID(0),
+            initial_balance: initial_balance * FILECOIN_PRECISION,
+            start_epoch: UPGRADE_LIFTOFF_HEIGHT,
+            unlock_duration,
+            pending_txs: Cid::default(),
```
austinabell, line 221: Never use the Cid default, if you need an optional Cid, use `Option<Cid>`

vm/interpreter/src/vm.rs
```diff
@@ -483,6 +484,8 @@ where
             0,
             self.rand,
             &self.registered_actors,
+            circ_supply::setup_preignition_genesis_actors_testnet(self.store).ok(),
+            circ_supply::setup_postignition_genesis_actors_testnet(self.store).ok(),
```
austinabell, line 488: You're ignoring the error and just acting as if it was none? this will not be intentional behaviour

austinabell (changes requested):
I don't think incomplete network upgrade logic is worth bringing in, especially since untested. Can you please just remove it and just have the calculation based on the original and add TODOs where necessary? It's unclear we will even support this upgrade

vm/interpreter/src/circ_supply.rs
nannick: My bad that should be reward_actor.state

vm/interpreter/src/circ_supply.rs
nannick: Kinda confused by what you mean here. As I have it right now the GenesisInfo is used in get_fil_vested

ec2:
Most of this stuff should probably live in the state manager. This is because there will API endpoint implemented for most/all of these functions.

vm/interpreter/src/circ_supply.rs
nannick: pending_txs is never used so ill resolve this

vm/actor/src/builtin/singletons.rs
```diff
@@ -17,6 +17,8 @@ lazy_static! {
 
     /// Distinguished AccountActor that is the destination of all burnt funds.
     pub static ref BURNT_FUNDS_ACTOR_ADDR: Address    = Address::new_id(99);
+
+    pub static ref RESERVE_ADDRESS : Address = Address::from_bytes(b"t090").unwrap();
```
austinabell, line 21: This won't be correct, this will load the address by the bytes of the utf8 string. Just use `new_id(90)` lol

What you were looking for here is from string

vm/interpreter/src/default_runtime.rs
```diff
@@ -126,6 +131,11 @@ where
             value_received: message.value().clone(),
         };
 
+        let default_preignition =
+            setup_preignition_genesis_actors_testnet(&gas_block_store).unwrap();
+        let default_postignition =
+            setup_postignition_genesis_actors_testnet(&gas_block_store).unwrap();
```
austinabell, line 137: I don't see why you would possibly need the unwrap? just downcast to an `ActorError`

vm/interpreter/src/default_runtime.rs
```diff
@@ -599,42 +611,14 @@ where
...
-            - burnt.balance
-            - st.total_pledge_collateral;
-        Ok(total)
+        let total_circ = get_circulating_supply(
+            &self.pre_ignition,
+            &self.post_ignition,
+            self.epoch,
+            self.state,
+        )
+        .map_err(|e| actor_error!(fatal(e)))?;
```
austinabell, line 620: nah this isn't fatal, it's `ErrIllegalState`

austinabell (changes requested):

vm/actor/src/builtin/singletons.rs
```diff
@@ -17,6 +17,8 @@ lazy_static! {
 
     /// Distinguished AccountActor that is the destination of all burnt funds.
     pub static ref BURNT_FUNDS_ACTOR_ADDR: Address    = Address::new_id(99);
+
+    pub static ref RESERVE_ADDRESS : Address  = Address::new_id(90);
```
austinabell, line 21:
```suggestion
    pub static ref RESERVE_ADDRESS: Address = Address::new_id(90);
```

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+use ipld_blockstore::BlockStore;
+use num_bigint::BigInt;
+use state_tree::StateTree;
+use std::collections::HashMap;
+use std::error::Error as StdError;
+
+#[derive(Default)]
+pub struct GenesisInfo {
+    genesis_msigs: Vec<multisig::State>,
+    // info about the Accounts in the genesis state
```
austinabell, line 23:
```suggestion
    /// info about the Accounts in the genesis state
```

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+fn get_actor_state<DB: BlockStore>(
+    state_tree: &StateTree<DB>,
+    addr: &Address,
+) -> Result<ActorState, ActorError> {
+    state_tree
+        .get_actor(&addr)
+        .map_err(|e| {
+            e.downcast_default(
+                ExitCode::ErrIllegalState,
+                "failed to get reward actor for cumputing total supply",
```
austinabell, line 43: what is this error message lol

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+fn get_actor_state<DB: BlockStore>(
+    state_tree: &StateTree<DB>,
+    addr: &Address,
+) -> Result<ActorState, ActorError> {
+    state_tree
+        .get_actor(&addr)
+        .map_err(|e| {
+            e.downcast_default(
```
austinabell, line 41: this should be a fatal downcast, there is no reason why this shouldn't be a bug

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+                ExitCode::ErrIllegalState,
+                "failed to get reward actor for cumputing total supply",
+            )
+        })?
+        .ok_or_else(|| actor_error!(ErrIllegalState; "Actor address ({}) does not exist", addr))
```
austinabell, line 46: This does not need to be handled by actor errors at all, but if it is, this should not use this exit code

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+pub fn get_fil_vested<DB: BlockStore>(
+    pre_ignition: &GenesisInfo,
+    post_ignition: &GenesisInfo,
+    height: ChainEpoch,
+    state_tree: &StateTree<DB>,
+) -> Result<TokenAmount, ActorError> {
```
austinabell, line 54: This can just be a boxed std Error or even string error, there is no need for ActorError outside of the VM, and when it's called within, everything is overridden to `ErrIllegalState`

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    for actor in &pre_ignition.genesis_actors {
+        let state = get_actor_state(state_tree, &actor.addr)?;
+        let diff = &actor.init_bal - state.balance;
+        if diff > TokenAmount::default() {
+            return_value += diff
+        }
+    }
+
+    if height < UPGRADE_ACTORS_V2_HEIGHT {
```
austinabell, line 76:
```suggestion
    if height <= UPGRADE_ACTORS_V2_HEIGHT {
```

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+    Ok(power_state.total_locked().clone())
+}
+
+pub fn get_fil_reserve_disbursed<DB: BlockStore>(
+    state_tree: &StateTree<DB>,
+) -> Result<TokenAmount, Box<dyn StdError>> {
+    let power_actor = get_actor_state(state_tree, &RESERVE_ADDRESS)?;
+
+    // If money enters the reserve actor, this could lead to a negative term
+    Ok(FIL_RESERVED.clone() - power_actor.balance)
```
austinabell, line 127:
```suggestion
    let reserve_actor = get_actor_state(state_tree, &RESERVE_ADDRESS)?;

    // If money enters the reserve actor, this could lead to a negative term
    Ok(&*FIL_RESERVED - reserve_actor.balance)
```

vm/interpreter/src/default_runtime.rs
```diff
@@ -599,42 +606,28 @@ where
...
-        let total = P::from_fil(P::TOTAL_FILECOIN)
-            - rew.balance
-            - market.balance
-            - burnt.balance
-            - st.total_pledge_collateral;
-        Ok(total)
+                })?,
+            ),
+            self.post_ignition.as_ref().unwrap_or(
+                &setup_postignition_genesis_actors_testnet(&self.store).map_err(|e| {
```
austinabell, line 619: so the difference here is that the go implementation only performs this computation once, where you are doing every single time total_fil_circ_supply is called. This cannot be this way

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+    };
+    let fil_circulating = BigInt::max(
+        &fil_vested + &fil_mined + fil_reserve_distributed - &fil_burnt - &fil_locked,
+        TokenAmount::default(),
+    );
+
+    Ok(fil_circulating)
+}
+
+fn get_totals_by_epoch() -> HashMap<ChainEpoch, TokenAmount> {
```
austinabell, line 169: If you actually need to use a map (I assume this is not necessary), then the return should be an AHashMap or FnvHashMap, it's an integer key which these hash functions are much faster for

Checking usage it seems like this can just be a tuple vector?

austinabell (changes requested):

vm/interpreter/src/circ_supply.rs
ec2: yeah, looks like this can be a static ref vec

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+use num_bigint::BigInt;
+use state_tree::StateTree;
+use std::collections::HashMap;
+use std::error::Error as StdError;
+
+#[derive(Default)]
+pub struct GenesisInfo {
+    genesis_msigs: Vec<multisig::State>,
+    // info about the Accounts in the genesis state
+    genesis_actors: Vec<GenesisActor>,
```
ec2, line 24: weird how this is never actually set to anything other than []...

vm/interpreter/src/circ_supply.rs
```diff
@@ -0,0 +1,254 @@
...
+    let mut pre_ignition = init_genesis_info(bs)?;
+
+    let totals_by_epoch: HashMap<ChainEpoch, TokenAmount> = get_totals_by_epoch();
+
+    for (unlock_duration, initial_balance) in totals_by_epoch {
+        let ms = multisig::State {
+            signers: vec![],
+            num_approvals_threshold: 0,
+            next_tx_id: multisig::TxnID(0),
+            initial_balance,
```
ec2, line 220: hm, i wonder why this doesnt need to be multipled by FILECOIN_PRECISION....

vm/interpreter/src/circ_supply.rs
nannick: making it a static vec

vm/interpreter/src/circ_supply.rs
nannick: Freudian slip my bad

vm/interpreter/src/vm.rs
```diff
@@ -481,6 +481,11 @@ where
         Option<DefaultRuntime<'db, '_, DB, R, V, P>>,
         Option<ActorError>,
     ) {
+        let default_preignition =
+            setup_preignition_genesis_actors_testnet(self.store).unwrap_or_default();
+        let default_postignition =
+            setup_postignition_genesis_actors_testnet(self.store).unwrap_or_default();
```
austinabell, line 487: This logic should not be inside of the VM. Also, this is being called every time there is a send in the VM, terribly inefficient

vm/interpreter/src/default_runtime.rs
```diff
@@ -101,6 +104,8 @@ where
         origin: Address,
         origin_nonce: u64,
         num_actors_created: u64,
+        pre_ignition: GenesisInfo,
+        post_ignition: GenesisInfo,
```
austinabell, line 108: This should not be attached to the runtime, this does not allow for swapable circ supply calculations

tests/conformance_tests/tests/conformance_runner.rs
```diff
@@ -140,7 +148,9 @@ fn execute_message_vector(
     randomness: &Randomness,
     variant: &Variant,
 ) -> Result<(), Box<dyn StdError>> {
+    println!("Starting loaf");
     let bs = load_car(car)?;
+    println!("Passing load");
```
austinabell, line 153: prints?

tests/conformance_tests/tests/conformance_runner.rs
```diff
@@ -68,7 +75,8 @@ fn load_car(gzip_bz: &[u8]) -> Result<db::MemoryDB, Box<dyn StdError>> {
     let d = GzDecoder::new(gzip_bz);
 
     // Load car file with bytes
-    forest_car::load_car(&bs, d)?;
+    let _ = forest_car::load_car(&bs, d)?;
```
austinabell, line 78:
```suggestion
    forest_car::load_car(&bs, d)?;
```

austinabell (changes requested):

vm/interpreter/src/vm.rs
ec2: I think this should be passed into the VM::new() as a function pointer right? I believe that was the whole point of that change.

nannick: Ahhh Okay so it would make more sense to keep it as part of the DefaultRuntime then

ec2: Well, acutally I dont meant that directly. It should be called every time you do a get_vm_circulating_supply. and that is what should be passed as the funciton pointer

austinabell: The circ supply calculation should yes, this is not really related to that, and this function should _not_ be called every time that is called, only when it's not initialized

blockchain/state_manager/src/vm_circ_supply.rs
```diff
@@ -0,0 +1,284 @@
...
+            // Default Cid is ok here because this field is never read
+            pending_txs: Cid::default(),
+        };
+        pre_ignition.genesis_msigs.push(ms);
+    }
+
+    Ok(pre_ignition)
+}
+
+pub fn setup_postignition_genesis_actors_testnet<DB: BlockStore>(
```
ec2, line 264:
```suggestion
pub fn setup_postignition_genesis_actors<DB: BlockStore>(
```

blockchain/state_manager/src/vm_circ_supply.rs
```diff
@@ -0,0 +1,284 @@
...
+    bs: &DB,
+) -> Result<GenesisInfo, Box<dyn StdError>> {
+    let mut post_ignition = init_genesis_info(bs)?;
+
+    for (unlock_duration, initial_balance) in &*TOTALS_BY_EPOCH {
+        let ms = multisig::State {
+            signers: vec![],
+            num_approvals_threshold: 0,
+            next_tx_id: multisig::TxnID(0),
+            initial_balance: initial_balance * FILECOIN_PRECISION,
```
ec2, line 274: i'd add a note here similar to how to lotus has to indicate why you multiply by the precision here but not in preignition

ec2 (approved):

austinabell (approved):
did some cleanup after your review @ec2 if you want to double check

ec2:
looked at the changes, lgtm

PR #711, opened 2020-09-22
Update block validations

### LINKED ISSUE #78: Add Block Validation Check Methods

austinabell:
**Summary of changes**
- Updated and implemented all block validations with all non-trivial checks done async
- Moved `ElectionProof` to blocks crate as logic was not tied to crypto functions and want to keep crypto crate as small as possible
  - Also implementing computing win count, which is the inverse of logic done in reward actor (but with different precision and other slight differences which is why it may seem like some duplication with the params, also to avoid circular dep)
- Makes ticket optional in block header
  - This was always a pointer in go impl, so not sure if they did to heap allocate or if we just haven't hit the case where their is nil yet

What left unfinished (I will create issues for):
- Updating `verify_winning_post_proof`: there is conflicts with miner actor PR, and the changes should be focused on in their own changes as it's very specific integration with proofs
- Allowing for insecure post validation (optional, but will most likely be helpful when debugging)

Closes #78 


ec2:
I noticed in Lotus that the whether the block has been validated or not only effects whether the signature of the block is checked again or not. It looks like everything else gets checked again. Any thoughts on if this is desired?

austinabell:
> I noticed in Lotus that the whether the block has been validated or not only effects whether the signature of the block is checked again or not. It looks like everything else gets checked again. Any thoughts on if this is desired?

You're mixing up the block header signature verification cache with the block validation. We don't cache the validated status of the signature check, but that's why the `TODO` is there, just isn't much of a need now.

ec2 (approved):

dutterbutter (approved):

PR #718, opened 2020-09-24
Switches out flostream for Single Producer Multiconsumer Bus

### LINKED ISSUE #700: Investigate flo-stream usage
** Acceptance Criteria **
- Broadcast channels should only be used for subscriptions
- Whenever events are being published, there is no deadlock 

Notes
tldr: Usage is currently incorrect (given implementation details) which will cause the node to stall eventually, and library has flaws
- Ideally use tokio instead of flo-stream broadcast channel

AshantiMutinta:
**Summary of changes**
- lock free, behaves in the same way flo_stream does. Downside is that it does not implement Stream or Sink traits

Closes #700 

https://crates.io/crates/bus

### DIFF
--- a/blockchain/chain/Cargo.toml
+++ b/blockchain/chain/Cargo.toml
@@ -24,7 +24,7 @@ actor = { path = "../../vm/actor/" }
 blake2b_simd = "0.5.9"
 byteorder = "1.3.4"
 beacon = { path = "../beacon" }
-flo_stream = "0.4.0"
+bus = "2.2.3"
 address = { package = "forest_address", path = "../../vm/address" }
 lazy_static = "1.4"
 async-std = "1.6.3"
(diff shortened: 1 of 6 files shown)


austinabell:
Have you considered https://docs.rs/tokio/0.2.22/tokio/sync/broadcast/index.html ? This change only really makes the send operations trivially more efficient, but still runs into the issue of the sender being blocked on any receivers queue being full which will stall our node completely. This impl will just error on and receiver lagging behind while allowing it to continue to poll off most recent, and does not stall or block any other senders or receivers.

I like the design of the tokio broadcast channel, and it isn't tied to their runtime as far as I can tell. Also another benefit is async reads since receiving with sync primitives might starve our async executor?

austinabell:
Closing for now until this continues to clean up PRs. Also, proposed changes are not based on this and the branch is out of date now

PR #722, opened 2020-09-29
Update Actors error handling

### LINKED ISSUE #720: Allow Amt and Hamt to forward actor errors through blockstore usage
**Issue summary**

Currently, actor errors through the blockstore trait (which can be passed as the dynamic error return) will be overwritten because they are converted to a string to be an error variant of the Amt and Hamt errors.

This needs to allow the `ActorError` (in our case can only be an out of gas error through the gas blockstore currently) to be kept and downcasted in every usage in Actors (yeah lol)

austinabell:
**Summary of changes**
- This all stems from the gas charging blockstore (and some from gas charging syscalls). In the go implementation, they panic exactly where the error is encountered and that error is caught outside of their vm. Since we are (and should) bubble the errors up instead of panicking, that error exit code cannot be lost. This is done through the error downcasting. The reason for the trait implementation is because Amt and Hamt errors, as well as dynamic errors can downcast into a `ActorError` because of `SysErrOutOfGas` from blockstore interactions within those structures.
  - I've checked every mapped error and conversion (there is over 500) and this is what I've come up with

The only time this can ever be a difference is if the go implementation uses `(ExitCode) Wrapf` or `rt.Abort`, an out of gas error was not hit, and there was a different actor error was hit that would be overwritten by this. After checking all usages of this (I could be missing a case and was just skimming, there was ~300 of these as well) I'm semi confident all of them do not fit this criteria, or when they do the go implementation would just fatal error (so it doesn't matter)

Closes #720 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2412,6 +2412,7 @@ dependencies = [
  "ctrlc",
  "db",
  "fil_types",
+ "flo_stream",
  "forest_address",
  "forest_bigint",
  "forest_blocks",
@@ -2421,7 +2422,7 @@ dependencies = [
  "futures 0.3.5",
  "hex",
  "ipld_blockstore",
- "jsonrpc-v2",
+ "jsonrpc-v2 0.5.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "key_management",
  "libp2p",
  "log",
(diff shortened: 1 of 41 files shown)


timvermeulen:
Good stuff! The `ActorDowncast` trait looks like a good approach to me.

timvermeulen (approved):

dutterbutter (approved):

PR #723, opened 2020-09-30
Switch bigint division usage to Euclidean to match Go

austinabell:
**Summary of changes**
- Updates all bigint divisions that can be negative to Euclidean (floored) division from truncated
  - Go uses Euclidean for bigints and truncated for integers, we can match with these changes
  - Did not change some of the bigint division that is guaranteed to be positive (truncated and floored are same for positive values). If anyone is really uncomfortable by this, I can switch others, but it means more allocations because `div_floor` does not allow for dividing by anything but a bigint
- Fixes bug in miner actor `VestSpec` constant
  - With both changes, can uncomment the tipset conformance vector skip

Closes 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2038,6 +2038,7 @@ name = "forest_bigint"
 version = "0.1.1"
 dependencies = [
  "num-bigint 0.3.0",
+ "num-integer",
  "serde",
  "serde_bytes",
 ]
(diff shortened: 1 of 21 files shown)


timvermeulen (approved):
This seems reasonable enough, it does improve the case for using newtypes though – it's quite possible we'll mess this up at some point in the future by incorrectly writing `/`. Something to think about

austinabell:
> This seems reasonable enough, it does improve the case for using newtypes though – it's quite possible we'll mess this up at some point in the future by incorrectly writing `/`. Something to think about

I totally agree, but there is a huge amount of overhead with wrapping the bigint interface, I've done it before and there is way too many traits needed to be forwarded. I'm definitely open to this in the future, but maybe until we hit milestones we just keep in mind and be a bit careful?

nannick (approved):

PR #724, opened 2020-10-01
BlockSync provider

### LINKED ISSUE #673: Implement blocksync provider
**Task summary**

Currently, we are ignoring blocksync incoming requests, but this needs to be done now. It should be very easy, there is an `Arc` reference to db in the `Libp2pService` so this should just spawn a task (so it doesn't stall the libp2p service) and just send the response through the channel that is provided already in the request.

**Specification reference**

- 

snaumov:
**Summary of changes**
- 

BlockSync "provider" - the logic to run for an incoming `BlockSync` libp2p request.

**Test notes**

It was somewhat hard to test this MR. I've added a unit test prepopulating MemoryDB with some chain data (thanks @austinabell  for generating file for me), and compared the response with FilFox, see comments in the test.

Closes #673 


node/forest_libp2p/src/service.rs
```diff
@@ -187,13 +187,14 @@ where
...
                         }
-                        ForestBehaviourEvent::BlockSyncRequest { channel, .. } => {
-                            // TODO implement blocksync provider
-                            let _ = channel.send(BlockSyncResponse {
-                                chain: vec![],
-                                status: 203,
-                                message: "handling requests not implemented".to_owned(),
-                            }).await;
+                        ForestBehaviourEvent::BlockSyncRequest { channel, peer, request } => {
+                            debug!("Received blocksync request (peerId: {:?})", peer);
```
snaumov, line 191: @austinabell you've suggested to `thread::spawn` here? I thought about it, but In order to do that, the `Arc<DB>` will need to be changed to `Arc<RwLock<DB>>` I think, however DB is already being passed around in several services with just `Arc`

node/forest_libp2p/src/blocksync/message.rs
```diff
@@ -2,70 +2,26 @@
...
-    {
-        if self.status != 0 {
-            // TODO implement a better error type than string if needed to be handled differently
-            return Err(format!("Status {}: {}", self.status, self.message));
-        }
-
-        self.chain.into_iter().map(T::try_from).collect()
-    }
-}
-
```
austinabell, line 52: Why did these types need to be moved? Makes more sense for these to be here (also to be symmetric with `hello` protocol message definitions)

Cargo.lock
```diff
@@ -217,7 +217,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d0864d84b8e07b145449be9a8537db86bf9de5ce03b913214694643b4743502"
 dependencies = [
  "quote 1.0.7",
- "syn 1.0.42",
+ "syn 1.0.38",
```
austinabell, line 220: The cargo lock is downgrading a lot of versions, do you mind trying to keep the lockfile the same? Best if not changed unless needed to becuase of cached builds

austinabell (changes requested):

austinabell:
Thanks so much for this contribution! Looks generally good, and happy to help with any of the changes or suggestions if you would like :)

node/forest_libp2p/src/service.rs
snaumov: yeah, I was attempting `RwLock`, because it implements `Sync + Send`, `task::spawn` was complaining that `DB` is not thread-safe. 
Solved it now by simply adding `+Sync+Send` to `DB`.

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        self.chain.into_iter().map(T::try_from).collect()
+    }
+}
+
+/// Builds blocksync response out of chain data.
+pub struct BlockSyncProvider<DB: BlockStore> {
+    db: Arc<DB>,
+}
+
```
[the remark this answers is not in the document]
snaumov: yeah, sounds good. Something like this you have in mind?

```
BlockSyncProvider {
  db,
  channel_in,
  channel_out,
}
```

```
Libp2pProvider {
  ...
  blockSync_channel_in,
  blockSync_channel_out,
}

impl Libp2pProvider {
 pub async fn run(mut self) { 
   let provider = new BlockSyncProvider {
     self.blockSync_channel_in,
     self.blockSync_channel_out,
   };

   provider.run();
 }
}
```

```
ForestBehaviourEvent::BlockSyncRequest { channel, peer, request } => {
       self.blockSync_channel_in.send(request);
}
```

Alternatively, I can get rid of the BlockSyncProvider, and just relocate compact messages logic somewhere. I wasn't sure where is the best place to put so created a new struct.

node/forest_libp2p/src/blocksync/message.rs
snaumov: This was my attempt to justify the creation of a new struct and module (`provider.rs`) by relocating these how I thought relevant structs over there. See my comment above regarding the channel/function idea.

node/forest_libp2p/src/blocksync/provider.rs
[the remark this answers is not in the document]
austinabell: Actually, I'm kinda coming back to the conclusion that just spawning tasks to handle the requests is probably best, the issue with this is that the channel would have to be managed and the provider thread either scaled up when the channel fills and there is a lot of overhead to this.

I'd say just keep this as it is (can remove the provider struct or remove, doesn't matter) and just call the handle function in a spawned task

node/forest_libp2p/src/blocksync/message.rs
austinabell: Can you just keep these types where they were and just have the provider logic in that new file? There should be no issue accessing, they are all in the same crate

node/forest_libp2p/src/service.rs
austinabell: yeah just to clarify on that, `task::spawn_blocking` should be used, as the database access and this logic is synchronous

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,389 @@
...
+    PartialResponse = 101,
+    // Request.Start not found.
+    BlockNotFound = 201,
+    // Requester is making too many requests.
+    GoAway = 202,
+    // Internal error occured.
+    InternalError = 203,
+    // Request was bad
+    BadRequest = 204,
+}
```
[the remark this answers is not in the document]
snaumov: yeah, my bad, tested the serialization and it's indeed wrong. 
So I was wondering if in rust there is a way to represent the integers in enum, and be able to serialize them properly. 

I can think of two ways:

```
struct Status(u64);

impl Struct {
  pub fn set_success() {
   // check me on this
    self.0 = 500;
  }

  pub fn get_status() {
    self.0
  }
  
}
```

or use some crate like https://docs.rs/num_enum/0.5.1/num_enum/

Anyways, switched from `enum` to `const`s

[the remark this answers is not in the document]
austinabell: Yeah so there are a few ways to match using [serde repr](https://github.com/dtolnay/serde-repr) which isn't as ideal in this case because it'll only deserialize if values match, or you can just manually implement deserialize through an integer and just have an enum for Other codes for when it doesn't match.

Here is the change I am suggesting as a commit, for consideration: https://github.com/ChainSafe/forest/commit/61e857033dbef88b600a76bfd5de64034bd34e8d. Using an integer value with constants is completely fine as well.

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,389 @@
...
+        // https://filfox.info/en/tipset/39
+        // https://filfox.info/en/tipset/38
+
+        // Response chain should contain 2 tipsets.
+        assert_eq!(response.chain.len(), 2);
+
+        // Tipset at height 39...
+        // ... has 22 signed messages
+        assert_eq!(
+            response.chain[0].messages.as_ref().unwrap().secp_msgs.len(),
```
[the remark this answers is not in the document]
snaumov: updated.

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,389 @@
...
+                match chain::tipset_from_keys(self.db.as_ref(), &TipsetKeys::new(curr_tipset_cids))
+                {
+                    Ok(tipset) => tipset,
+                    Err(err) => {
+                        debug!("Cannot get tipset from keys: {}", err);
+
+                        return BlockSyncResponse {
+                            chain: vec![],
+                            status: BlockSyncResponseStatus::InternalError,
+                            message: "Can not fullfil the request".to_owned(),
```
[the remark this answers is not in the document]
snaumov: updated.

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,389 @@
...
+
+            curr_tipset_cids = tipset.parents().cids().to_vec();
+        }
+
+        let result_chain_length = response_chain.len() as u64;
+
+        BlockSyncResponse {
+            chain: response_chain,
+            status: if result_chain_length < request.request_len {
+                BlockSyncResponseStatus::PartialResponse
```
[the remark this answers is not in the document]
snaumov: we push `compactedMessages` into `response_chain` in the loop above. The loop will halt either if we reached the requested length, or if the tipset.epoch == 0 ( we reached genesis block). In latter case, the resulting length of the chain will be shorter, than requested, so we return status as `Partial`

node/forest_libp2p/src/blocksync/message.rs
snaumov: sure, moved them back.

Cargo.lock
snaumov: my bad. Checked out the file from the current `main`

node/forest_libp2p/src/blocksync/provider.rs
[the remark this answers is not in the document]
snaumov: nice! Replaced with your suggestion.

node/forest_libp2p/src/blocksync/provider.rs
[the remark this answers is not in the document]
snaumov: attempting `task::spawn_blocking` but getting an error as doesn't seem to be able to use `async` inside
  
```
error[E0277]: expected a `std::ops::FnOnce<()>` closure, found `impl core::future::future::Future`
   --> node/forest_libp2p/src/service.rs:194:33
    |
194 | / ...                   async move {
195 | | ...                       let response = make_blocksync_response(db, &request.clone());
196 | | ...
197 | | ...                       let _ = channel.send(response).await;
198 | | ...                   }
    | |_______________________^ expected an `FnOnce<()>` closure, found `impl core::future::future::Future`
```


Will attempt to fix tomorrow with fresh eyes.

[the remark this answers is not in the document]
austinabell: My apologies, I completely forgot the channel send was async. If you want to just have it `task::spawn` that is fine for now, and I or someone else can optimize later.

The only reason I was being very cautious about this, is that many tasks will be spawned, and I don't want to starve the runtime executor with all the synchronous logic. The solution here would be to wrap just the `make_blocksync_response` in a spawn blocking, while keeping the outer logic in a task (to be in an async block). There is something similar done in our paramfetch logic, but as I said I'm happy to make a pass after, this isn't anything crucial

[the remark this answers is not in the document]
snaumov: made it working, see the latest changes.

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,241 @@
...
+use log::debug;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use super::{
+    BlockSyncRequest, BlockSyncResponse, BlockSyncResponseStatus, CompactedMessages, TipsetBundle,
+};
+
+/// Builds blocksync response out of chain data.
+pub fn make_blocksync_response<DB>(db: Arc<DB>, request: &BlockSyncRequest) -> BlockSyncResponse
```
austinabell, line 17: This is a small detail, but you don't need an `Arc` to be passed in here, only need a reference. I assumed you did because it's done in a new task, and you would need to clone to move the `Arc` into that task, but you don't need to move in the `Arc` into this function. Only reason I mention this is that it makes it easier to test if you just need to pass in a db reference

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,241 @@
...
+where
+    DB: BlockStore,
+{
+    let mut response_chain: Vec<TipsetBundle> = vec![];
```
austinabell, line 21:
```suggestion
    let mut response_chain: Vec<TipsetBundle> = Vec::with_capacity(request.request_len as usize);
```
The bundles will be of non-trivial size, best to avoid reallocations as it's rare for partial responses. (If you want, you could also just set capacity to the max of the request length or epoch)

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    let mut curr_tipset_cids = request.start.clone();
+
+    loop {
+        let mut tipset_bundle: TipsetBundle = Default::default();
```
austinabell, line 26:
```suggestion
        let mut tipset_bundle = TipsetBundle::default();
```
Little more readable imo

node/forest_libp2p/src/blocksync/message.rs
```diff
@@ -22,11 +24,82 @@ pub struct BlockSyncRequest {
...
+impl BlockSyncRequest {
+    /// If a request expects blocks to be included in response.
+    pub fn include_blocks(&self) -> bool {
+        self.options == BLOCKS || self.options == BLOCKS_MESSAGES
+    }
+
+    /// If a request expects messages to be included in response.
+    pub fn include_messages(&self) -> bool {
+        self.options == MESSAGES || self.options == BLOCKS_MESSAGES
+    }
```
austinabell, line 36: These can be bit operations, `BLOCKS` is the least significant bit, and `MESSAGES` is the second least significant bit.

Can be something like:
```
    pub fn include_blocks(&self) -> bool {
        self.options & BLOCKS != 0
    }

    /// If a request expects messages to be included in response.
    pub fn include_messages(&self) -> bool {
        self.options & MESSAGES != 0
    }
```

node/forest_libp2p/src/blocksync/message.rs
```diff
@@ -5,11 +5,13 @@ use forest_blocks::{Block, BlockHeader, FullTipset, Tipset, BLOCK_MESSAGE_LIMIT}
 use forest_cid::Cid;
 use forest_encoding::tuple::*;
 use forest_message::{SignedMessage, UnsignedMessage};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::convert::TryFrom;
 
 /// Blocksync request options
 pub const BLOCKS: u64 = 1;
 pub const MESSAGES: u64 = 2;
+pub const BLOCKS_MESSAGES: u64 = 3;
```
austinabell, line 14: Don't need this

node/forest_libp2p/src/blocksync/message.rs
```diff
@@ -22,11 +24,82 @@ pub struct BlockSyncRequest {
...
+    }
+
+    /// If a request expects messages to be included in response.
+    pub fn include_messages(&self) -> bool {
+        self.options == MESSAGES || self.options == BLOCKS_MESSAGES
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum BlockSyncResponseStatus {
```
austinabell, line 40: doc comment would be nice, nbd tho

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,241 @@
...
+                    chain: vec![],
+                    status: BlockSyncResponseStatus::InternalError,
+                    message: "Tipset was not found in the database".to_owned(),
+                };
+            }
+        };
+
+        if request.include_blocks() {
+            tipset_bundle.blocks = tipset.blocks().to_vec();
+        }
```
austinabell, line 43: Might be worth rearranging this logic so you don't need to clone the headers. I can do this later so no need to worry about it if you don't want

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,241 @@
...
+    let mut bls_cids_combined: Vec<Cid> = vec![];
+    let mut secp_cids_combined: Vec<Cid> = vec![];
+    let mut bls_msg_includes: Vec<Vec<u64>> = vec![];
+    let mut secp_msg_includes: Vec<Vec<u64>> = vec![];
+
+    for block_header in tipset.blocks().iter() {
+        let (bls_cids, secp_cids) = chain::read_msg_cids(db, block_header.messages())?;
+
+        let mut block_include = vec![];
+        let mut secp_include = vec![];
```
austinabell, line 98: can you set the capacity to the length of the cids read above?

node/forest_libp2p/src/service.rs
```diff
@@ -187,13 +187,15 @@ where
...
+                            let db = self.db.clone();
+                            async {
+                                let response = task::spawn_blocking(move || -> BlockSyncResponse {
+                                    make_blocksync_response(db, &request.clone())
```
austinabell, line 195:
```suggestion
                                    make_blocksync_response(db, &request)
```
This request can be moved, no need to clone

austinabell (changes requested):

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,241 @@
...
+
+        for bls_cid in bls_cids.iter() {
+            let order = match bls_messages_order.get(bls_cid) {
+                Some(order) => *order,
+                None => {
+                    let order = block_include.len() as u64;
+                    bls_cids_combined.push(bls_cid.clone());
+                    bls_messages_order.insert(bls_cid.clone(), order);
```
austinabell, line 106: Oh and one other thing, this can be `.into_iter()` to move the vector and require one less clone for each. (Nbd though, I can change after also)

node/forest_libp2p/src/blocksync/provider.rs
snaumov: I will double-check on this tonight, but I can recall that there was an error along the lines of `Sync+Send` is not implemented for `&DB` when attempting to pass a reference only.

austinabell: Yeah so you have to still clone the arc, but you only need to pass a reference to the cloned `Arc<DB>` to the function

snaumov: yes, just tested and it works :+1: Will update today.

node/forest_libp2p/src/blocksync/provider.rs
snaumov: updated

node/forest_libp2p/src/blocksync/provider.rs
snaumov: updated

node/forest_libp2p/src/blocksync/message.rs
snaumov: it's in a spec though?

node/forest_libp2p/src/blocksync/message.rs
snaumov: ya, thought about it, but couldn't squeeze anything less obvious than "Status of the BlockSync response" haha. Let me try once again.

snaumov: updated

node/forest_libp2p/src/blocksync/message.rs
snaumov: feels a little less readable imo :thinking:

node/forest_libp2p/src/blocksync/message.rs
austinabell: yeah it's not necessary though, that only exists because it wasn't bit flags before. `0b01 | 0b10 == 3`. There is no need to use this constant itself.

node/forest_libp2p/src/blocksync/provider.rs
snaumov: moved this logic at the end of the function and used `into_blocks()` instead.

node/forest_libp2p/src/blocksync/provider.rs
snaumov: sure, updated.

node/forest_libp2p/src/service.rs
snaumov: updated.

node/forest_libp2p/src/blocksync/provider.rs
snaumov: Good call :+1: Updated. Though I need the first clone, as I'm pushing to `Vec` and inserting to `HashMap` the same value, and hashmap consumes the ownership.

node/forest_libp2p/src/blocksync/provider.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut secp_messages_order = HashMap::new();
+    let mut bls_cids_combined: Vec<Cid> = vec![];
+    let mut secp_cids_combined: Vec<Cid> = vec![];
+    let mut bls_msg_includes: Vec<Vec<u64>> = vec![];
+    let mut secp_msg_includes: Vec<Vec<u64>> = vec![];
+
+    for block_header in tipset.blocks().iter() {
+        let (bls_cids, secp_cids) = chain::read_msg_cids(db, block_header.messages())?;
+
+        let mut block_include = Vec::with_capacity(bls_cids.len());
```
austinabell, line 96: this should probably be named `bls_include`, but doesn't matter that much

austinabell (approved):
LGTM, thanks so much for the contribution. The only comments left are opinionated and I'm good with this :)

node/forest_libp2p/src/blocksync/message.rs
```diff
@@ -22,11 +24,83 @@ pub struct BlockSyncRequest {
...
+        let code: i32 = Deserialize::deserialize(deserializer)?;
+
+        use BlockSyncResponseStatus::*;
+        let status = match code {
+            x if x == 1 => Success,
+            x if x == 101 => PartialResponse,
+            x if x == 201 => BlockNotFound,
+            x if x == 202 => GoAway,
+            x if x == 203 => InternalError,
+            x if x == 204 => BadRequest,
```
timvermeulen, line 92:
```suggestion
            1 => Success,
            101 => PartialResponse,
            201 => BlockNotFound,
            202 => GoAway,
            203 => InternalError,
            204 => BadRequest,
```

timvermeulen (approved):

node/forest_libp2p/src/blocksync/message.rs
austinabell: oops yeah that was a lapse on my part (I suggested that).

node/forest_libp2p/src/blocksync/message.rs
snaumov: > yeah it's not necessary though, that only exists because it wasn't bit flags before

Sorry, when you say `before` are you referring to your suggestion here: https://github.com/ChainSafe/forest/pull/724#discussion_r500290219, or it's bit flags in a spec now too?

node/forest_libp2p/src/blocksync/provider.rs
snaumov: renamed

node/forest_libp2p/src/blocksync/message.rs
snaumov: updated

austinabell (approved):
Sorry was just waiting for branch to update before pulling in, didn't want to merge into your branch if you were making changes. I just updated the branch and will pull in when CI passes. Thanks!

Issue #713, opened 2020-09-22
Allow insecure post validation


austinabell:
**Issue summary**

Can put this logic behind a feature, so it has to be specifically built to skip checks, but this will be useful to have for debugging unrelated changes. This could also be an env variable check, but I think it might be best for it to be a feature, since the only use for it would be for developers.

Should wait for #711 to come in, as the main functions which should be skipped are added with this.

**Other information and links**

PR #726, opened 2020-10-02
Update proof verification

### LINKED ISSUE #712: Update winning post proof validation
**Issue summary**

There were conflicts with other changes so left out of #711, but found a few bugs from trying to connect so this function should be revisited after this PR and #691 come in.

With this change, proofs api version should be updated and checked for if any logical changes have been made.

austinabell:
**Summary of changes**
- Fixes winning post proof check (and sub functions)
  - Bumps proofs version
- Implements new `ProofVerifier` trait to allow for verification to be swapped out for a mock verifier for testing
  - There was also a bunch of duplicated and incorrect logic, so consolidating was necessary
  - This trait just has associated functions, and is just used with `PhantomData` so it has no runtime cost

I'll get ahead of a few questions since I assume it might not be intuitive from scanning changes:

> Why not have the runtime `Syscalls` trait a `supertrait` ([yes that's the name](https://doc.rust-lang.org/rust-by-example/trait/supertraits.html)) of `ProofVerifier`?

While there is overlap, syscalls does not need the functionality of verifying winning post proofs or generating sector challenge ranges, so that is not a direct subset and would require implementing extra logic for other runtimes/VM.

> Why not have `StateManager` own the `ProofVerifier` generic?

This is a bit more opinionated, but it made more sense to me to only require the generic type when calling a function that uses the VM. Negative to this is that it opens up a possibility of calling a function with a full and mock verifier, and they could theoretically disagree on state because the full verifier can fail where the mock verifier would succeed. This can always change when #713 is implemented (and our strategy of swapping this out is clear)

> Why was the default proof verification removed from `Syscalls` trait?

Didn't want to add the generic bound on the trait, as only the `DefaultSyscalls` type has the option of using the generic. Also because the logic was moved into the `ProofVerifier` trait.

> Why were the `state_manager::utils` functions changed to implementation of `StateManager`?

Idk, why was it setup the way it was in the first place? Got very verbose quick with introducing the extra generic.

> Why do the RPC functions that use the VM default to `FullVerifier` implementation?

The mechanism for swapping verifier implementation at runtime isn't decided, and didn't want to add another generic bound to the RPC state when the node also defaults to `FullVerifier`.

Closes #712 


ec2:
> Why were the state_manager::utils functions changed to implementation of StateManager?
Probably because they are really DB wrappers was the original reason they weren't in the SM impl.

ec2 (approved):

dutterbutter (approved):
LGTM! Although `blockchain/state_manager/src/utils.rs ` could benefit from some documentation on the public fns

austinabell:
> > Why were the state_manager::utils functions changed to implementation of StateManager?
> > Probably because they are really DB wrappers was the original reason they weren't in the SM impl.

I don't know what you mean, the Lotus implementation separates into own functions not associated, which I assume is the reason, but it makes no sense to do in Rust and adds a lot of clunkyness with generics. (https://github.com/filecoin-project/lotus/blob/bc495a5c8cdf3934b4f559f735141bdeaf06c64d/chain/stmgr/utils.go)

PR #727, opened 2020-10-02
Fix signature verification in mempool

### LINKED ISSUE #725: `verify_secp256k1_sig` use an invalid hash for signature
**Describe the bug**
`verify_secp256k1_sig` seems to use an invalid hash of the message in order to verify the signature. The hash is different than the one actually signed so the public key recover doesn't match the one in the `from` field.

**To Reproduce**
1. Get a signature from lotus
2. Verify signature using the function
3. See "Secp signature verification failed" error

You can see the difference between the digest used in our project and the one used in the forest verify function. https://github.com/Zondax/filecoin-signing-tools/pull/293/files#diff-46b5d15cb95e5866c7ed4a201c2fed7fR861

austinabell:
**Summary of changes**
- Fixes signature verification in `new_from_parts`
- Fixes signature generation and verification in message pool
- Released new versions
  - Only message crate was a breaking change, and removed the types dependency from the crate (to avoid fil-proofs from being pulled into this dep tree)
  - Added CHANGELOG to message crate, to allow @rllola and @jleni to easily follow changes in releases for their usages.
- Cleaned up and added convenience functions for verifications to make this harder to be an issue later
  - Didn't change the `new_from_parts` function to not verify, because it's a good sanity check for a bit, but it probably should be removed later as there is no need for it to verify.

The issue was that message Cbor was being signed, not the message cid bytes. The only borked usage was in the mempool, which is why we didn't run into this yet. I could have sworn I commented to change this, but I guess I didn't follow up.

These interfaces can definitely be cleaned up (especially in the mempool), but I'm just going to keep scope small for this PR

Closes #725 


blockchain/message_pool/src/msgpool.rs
```diff
@@ -373,10 +374,7 @@ where
             return Ok(());
         }
 
-        let umsg = msg.message().marshal_cbor()?;
-        msg.signature()
-            .verify(umsg.as_slice(), msg.from())
-            .map_err(Error::Other)?;
+        msg.verify().map_err(Error::Other)?;
```
dutterbutter, line 377: Should this error type be `InvalidFromAddr` rather than `Other`? It seems the `verify` fn would fail if the `from` address is not the correct address which seems more of an `InvalidFromAddr` error type.

ec2:
I don't get why you moved `valid_for_block_inclusion` from being a method on the Message trait into the msgpool

austinabell:
> I don't get why you moved `valid_for_block_inclusion` from being a method on the Message trait into the msgpool

The reason is because filecoin proofs is brought into the dependency tree of messages and it makes it unusable by others using published crate for no benefit to us (because proofs dependency only works on certain targets).

Idea is just to keep the message crate lean, and separating proofs logic by a feature was not something I was eager to do now, since there was no need.

blockchain/message_pool/src/msgpool.rs
austinabell: That's not the only error that can happen with signature validation. You also lose the error message for no benefit. 

Side note, I don't like the error type in this crate, would be nice to improve later.

dutterbutter (approved):

ec2 (approved):

PR #732, opened 2020-10-07
Add Networking and Republishing logic to MsgPool

### LINKED ISSUE #708: Add Networking to Mempool
- [ ] Mempool should be able to push messages to gossipsub
- [ ] Republishing & Revert logic (if a message doesn't get included, send it out after a while)

ec2:
**Summary of changes**
- Adds networking to the message pool to broadcast messages over GossipSub
- Adds logic for republishing of messages over GossipSub

The MsgChainNode and MsgChain structs indeed look kinda of funny. I feel like it can definitely be improved on, but the idea is to mimic a doubly linked list in a way that is close enough for our purposes. There are definitely a few things that seem kinda sketchy. Please reach out to me if the logic is super hard to follow. 
Those unwraps on `.curr()` will definitely raise some eyebrows, but they are safe insofar that you create MsgChain through the constructor. I figured this might not matter, since this is an internal datastructure. If you have a better way of doing around this, please let me know 😄 

Closes 
#708 


blockchain/message_pool/src/msg_chain.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+/// Mimics the doubly linked circular-referenced message chain from Lotus by keeping a current index
+#[derive(Clone, Debug)]
+pub(crate) struct MsgChain {
+    pub index: usize,
+    pub chain: Vec<MsgChainNode>,
+}
+
+impl MsgChain {
```
nannick, line 58: Thats pretty dope didn't realize you could separate the getters and simple commands from  the more complicated logic

blockchain/message_pool/src/msgpool.rs
```diff
@@ -318,6 +352,49 @@ where
...
+                            pending.as_ref(),
+                            cur_tipset.as_ref(),
+                            republished.as_ref(),
+                            local_addrs.as_ref(),
+                        )
+                        .await
+                        {
+                            warn!("Failed to republish pending messages: {}", e.to_string());
+                        }
+                    }
```
austinabell, line 394: You don't need to match if you are doing the same logic with both?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -776,11 +877,303 @@ where
...
+            .send(NetworkMessage::PubsubMessage {
+                topic: PUBSUB_MSG_TOPIC.clone(),
+                message: mb,
+            })
+            .await;
+        count += 1;
+        if count < msgs.len() {
+            // this delay is here to encourage the pubsub subsystem to process the messages serially
+            // and avoid creating nonce gaps because of concurrent validation
+            // TODO: I am not sure if we will run into the same issues
```
austinabell, line 987: uh, I don't think we will, all requests are passed in through the channel and the ordering is sync compatible. They only do this because their pubsub publishing is accessible anywhere and memory unsafe I think. (100 ms sleep on every message on every republish scares me)

blockchain/message_pool/src/msgpool.rs
```diff
@@ -776,11 +877,303 @@ where
...
+                head.last_mut().unwrap().curr_mut().unwrap().gas_reward +=
+                    &tail[0].curr().unwrap().gas_reward;
+                head.last_mut().unwrap().curr_mut().unwrap().gas_limit +=
+                    head.last().unwrap().curr().unwrap().gas_limit;
+                head.last_mut().unwrap().curr_mut().unwrap().gas_perf = get_gas_perf(
+                    &head.last().unwrap().curr().unwrap().gas_reward,
+                    head.last().unwrap().curr().unwrap().gas_limit,
+                );
+                tail[0].curr_mut().unwrap().valid = false;
+                merged += 1;
```
austinabell, line 1147: ouuf

blockchain/message_pool/src/msgpool.rs
```diff
@@ -776,11 +877,303 @@ where
...
+            cur.gas_perf = gas_perf;
+        }
+    }
+    chains.push(cur_chain);
+
+    // merge chains to maintain the invariant
+    loop {
+        let mut merged = 0;
+        for i in (1..chains.len()).rev() {
+            let (head, tail) = chains.split_at_mut(i);
```
austinabell, line 1129: Won't this panic when indexing the 0th element of the tail when splitting a `chains.len()`, what am I missing here?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -776,11 +877,303 @@ where
...
+pub fn get_gas_reward(msg: &SignedMessage, base_fee: &BigInt) -> BigInt {
+    let mut max_prem = msg.gas_fee_cap() - base_fee;
+    if &max_prem < msg.gas_premium() {
+        max_prem = msg.gas_premium().clone();
+    }
+    max_prem * msg.gas_limit()
+}
+pub fn get_gas_perf(gas_reward: &BigInt, gas_limit: i64) -> f64 {
+    let a = gas_reward * types::BLOCK_GAS_LIMIT;
+    a.to_f64().unwrap() / gas_limit as f64
```
austinabell, line 1167: this isn't safe? Why is it assumed that f64 can accommodate the value?

blockchain/message_pool/src/msgpool.rs
```diff
@@ -939,6 +1353,11 @@ pub mod test_provider {
                     }
                 }
             }
+            let balance = match self.balances.get(addr) {
+                Some(b) => b.clone(),
+                None => (10_000_000_000 as u64).into(),
```
austinabell, line 1358: if this is intentional, can you at least document where this is coming from? Seems like it should be coming from a protocol constant

blockchain/state_manager/Cargo.toml
```diff
@@ -23,7 +23,7 @@ async-std = "1.6.3"
 async-log = "2.0.0"
 log = "0.4.8"
 fil_types = { path = "../../types" }
-message = { package = "forest_message", path = "../../vm/message" }
+message = { package = "forest_message", path = "../../vm/message" , features = ["json"]}
```
austinabell, line 26:
```suggestion
message = { package = "forest_message", path = "../../vm/message", features = ["json"]}
```

austinabell (changes requested):
Have you considered using some actual linked list implementation such as the [rust one](https://doc.rust-lang.org/std/collections/struct.LinkedList.html)? Seems very inefficient to use the vector and with all the unwraps and unsafe indexing scares me.

I'll go through more in depth later too because it's hard for me to wrap my head around all the mutex usage here

blockchain/message_pool/src/msgpool.rs
ec2: This would, but we never split at chains.len(), its [1, chains.len() -1]

austinabell: ah, I had wrong idea of the split, didn't realize it was excluding the index

blockchain/message_pool/src/msgpool.rs
ec2: its just for testing

blockchain/message_pool/src/msgpool.rs
ec2: https://doc.rust-lang.org/std/f64/constant.MAX.html
MAX is stupidly high

blockchain/message_pool/src/msg_chain.rs
```diff
@@ -0,0 +1,226 @@
...
+            mc.msgs = Vec::new();
+            mc.valid = false;
+        } else {
+            mc.msgs.drain(0..i as usize);
+        }
+
+        if self.move_forward().is_some() {
+            self.invalidate();
+            self.move_backward();
+            self.chain.remove(self.chain.len() - 1);
```
nannick, line 169: Think this last line actually deletes the node you want to keep.

blockchain/message_pool/src/msgpool.rs
```diff
@@ -16,22 +18,37 @@ use crypto::{Signature, SignatureType};
...
+const BASE_FEE_LOWER_BOUND_FACTOR: i64 = 10;
+const REPUB_MSG_LIMIT: usize = 30;
+const PROPAGATION_DELAY_SECS: u64 = 6;
+const REPUBLISH_INTERVAL: u64 = 10 * types::BLOCK_DELAY_SECS + PROPAGATION_DELAY_SECS;
+
+// this is *temporary* mutilation until we have implemented uncapped miner penalties -- it will go
+// away in the next fork.
+// TODO: Get rid of this?
+fn allow_negative_chains(epoch: i64) -> bool {
+    epoch < 41280 + 5
```
nannick, line 50: Make 41280 a constant, it lotus its UpgradeBreezeHeight

blockchain/message_pool/src/msgpool.rs
```diff
@@ -776,11 +857,296 @@ where
...
+                    .unwrap()
+                    .curr_mut()
+                    .unwrap()
+                    .msgs
+                    .append(&mut chain_a_msgs);
+                head.last_mut().unwrap().curr_mut().unwrap().gas_reward +=
+                    &tail[0].curr().unwrap().gas_reward;
+                head.last_mut().unwrap().curr_mut().unwrap().gas_limit +=
+                    head.last().unwrap().curr().unwrap().gas_limit;
+                head.last_mut().unwrap().curr_mut().unwrap().gas_perf = get_gas_perf(
```
nannick, line 1114: Since you are unwrapping the same stuff for each iteration, you can just unwrap once and make all the changes to that variable, doesn't look as scary lol. Something like this might work : 

let mut head_last_curr = head.last_mut().unwrap().curr_mut().unwrap();
head_last_curr.msgs.append(&mut chain_a_msgs);

blockchain/message_pool/src/msg_chain.rs
nannick: Basically i think the line at 169 shouldnt be there

nannick (changes requested):
Looks good except for those small changes

blockchain/message_pool/src/msgpool.rs
ec2: OK, so i did some research. Im using the num-rational crate now here. It has the ability to convert a ratio of integers to a f64. I unwrap because it seems the only case where it returns `None` is when the denominator is 0. This should never be the case because the `gas_limit` is guaranteed to have a non-zero value.

timvermeulen (approved):

blockchain/message_pool/src/msgpool.rs
```diff
@@ -815,13 +1176,22 @@ where
...
                 rm(msg.from(), pending, msg.sequence(), rmsgs.borrow_mut()).await?;
+                if !repub && republished.write().await.insert(msg.cid()?) {
+                    repub = true;
+                }
             }
         }
         *cur_tipset.write().await = Arc::new(ts);
     }
+    if repub {
+        repub_trigger.send(()).await;
```
austinabell, line 1193: should this not check if the channel is full before sending? The only issue I have with this is that it's a channel of undefined length and if you send multiple republish signals, would it not do duplicate unnecessary work? Wouldn't you just want it to behave like an atomic bool and some waking mechanism, where the other thread is just notified when there was an event, perform the work, then go back to sleep?

ec2: Only place that a repub_trigger.send() happens is in head_change, so there shouldnt be duplicate work. This is because the republish function will return early if theres nothing pending. In head_change the pending messages map will be updated to delete messages that dont need to be propitigated.

austinabell (approved):

PR #738, opened 2020-10-09
State diff on root mismatch option

austinabell:
**Summary of changes**
- Prints a state diff if `FOREST_DIFF=1` env variable is set and there is a state root mismatch
  - Resolves Ipld hash links recursively and the diff is in json

Looks something like:

Edit (looks like this now with the state tree resolve):
Screen Shot 2020-10-09 at 12 23 44 PM

Which can be useful sometimes but other times not extremely useful (reverted state doesn't get resolved and the hamt/ amt in ipld can be cryptic if you don't know what to look for)

Closes 


ec2 (approved):

ec2 (approved):

ipld/blockstore/src/resolve.rs
```diff
@@ -0,0 +1,56 @@
...
+                }
+                bs.get(cid)?
+            } else {
+                unreachable!()
+            };
+
+            if let Some(ipld) = resolved {
+                *link = ipld;
+            }
+        }
```
timvermeulen, line 52:
```suggestion
        Ipld::Link(cid) => {
            if cid.codec == Codec::DagCBOR {
                if let Some(x) = bs.get(cid)? {
                    *ipld = x;
                }
            }
        }
```

timvermeulen (approved):

ipld/blockstore/src/resolve.rs
austinabell: ah ya oops, for some reason I thought ipld would not be able to be overwritten while cid was bound in the match

PR #741, opened 2020-10-11
Fix block validations and add chain export test

austinabell:
**Summary of changes**
- e2e testing sync with chain export of first 40 tipsets (served through blocksync requests/responses)

Fixes:
- Weight calculation was incorrect
- Logic was incorrect on timestamp check
- Validate miner was using current ts to validate, not the base ts
- Ticket vrf was using header bytes to draw randomness, not miner address bytes
- Message validation was very outdated (and broken logic)
- Addresses resolved to key address for sig validation
- miner_has_min_power was using wrong address for loading state
- Blocksync response status code was incorrect

Closes 


timvermeulen (approved):

timvermeulen (approved):

vm/actor/src/builtin/codes.rs
```diff
@@ -52,6 +52,11 @@ pub fn is_singleton_actor(code: &Cid) -> bool {
         || code == &*VERIFREG_ACTOR_CODE_ID
 }
 
+/// Returns true if the code belongs to a singleton actor.
+pub fn is_account_actor(code: &Cid) -> bool {
```
ec2, line 56: Comment is wrong?

ec2 (approved):

PR #744, opened 2020-10-13
Fix Balance Table get

ec2:
**Summary of changes**
- When the Hamt returns Ok(None), we convert it to an Err. We need to return 0 instead. 

500/592

Closes 


austinabell (approved):
Ah, this has changed, it used to throw an error

dutterbutter (approved):

PR #749, opened 2020-10-15
Change Amt::for_each_mut to ignore unchanged values

### LINKED ISSUE #736: Change Amt::for_each_mut to ignore unchanged values
Currently all values are written back, regardless of whether they were mutated or not.

See:
- https://github.com/ChainSafe/forest/blob/2561c51c321f7f1a58c9444b7e3f4988377564fd/ipld/amt/src/node.rs#L508
- https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/expiration_queue.go#L635-L667

timvermeulen:
**Summary of changes**

Adds an `amt::ValueMut` wrapper type that wraps a mutable reference to an element in the Amt and keeps track of whether it was mutated, to avoid flushing nodes later that weren't changed in the first place.

Also updates the `BitFieldQueue` to correctly only write a value back conditionally.

Closes #736


austinabell:
I'm pretty sure that for the exact usage there will still be a difference in blockstore interactions from the go implementation, but I guess this can be iteratively fixed. Would have been nice to add or modify a test to check this use

ec2 (approved):

PR #752, opened 2020-10-16
Build and Api Versoining

AshantiMutinta:
**Summary of changes**
- Implements versioning for forest

Closes 


types/src/build_version/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+                Command::new("git")
+                    .args(&["rev-parse", "--short", "HEAD", "2>/dev/null"])
+                    .output()
+                    .map(|s| s.stdout)
+                    .unwrap_or_default()
+            });
+        String::from(std::str::from_utf8(&output).unwrap_or_default())
+    };
+    pub static ref BUILD_TYPE: RwLock<BuildType> = RwLock::new(BuildType::BuildDefault);
+    pub static ref RUNNING_NODE_TYPE: RwLock<NodeType> = RwLock::new(NodeType::Full);
```
austinabell, line 50: why rwlock? Always same?

types/src/build_version/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/// represents the current version of the api
+
```
austinabell, line 54:
```suggestion
```

types/src/build_version/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+impl PartialEq for Version {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 & MINOR_MASK == other.0 & MINOR_MASK
+    }
+}
+
+impl ToString for Version {
```
austinabell, line 126: why not display?

types/src/build_version/mod.rs
AshantiMutinta: Would be injected depending

types/src/build_version/mod.rs
AshantiMutinta: matching lotus for concatenation but can change!

dutterbutter (approved):

types/src/build_version/mod.rs
austinabell: The only point I was making here was that there is no need to wrap it in a mutex when it is never overwritten. Doesn't really matter but just no need for the overhead right now

austinabell (approved):

PR #781, opened 2020-10-23
Fix get_sectors_for_winning_post and cleanup

austinabell:
**Summary of changes**
- There was logic broken within getting sectors for winning verification
  - There was an issue with prover id using address bytes with the protocol byte included, where they don't (I removed the `pub` tag on this method so no one else goes through the same debugging I had to
- Updates state diff to allow passing a depth to limit recursion (since when using on real network it gets unusable quick

Closes 

### DIFF
--- a/blockchain/blocks/src/header/mod.rs
+++ b/blockchain/blocks/src/header/mod.rs
@@ -43,7 +43,7 @@ const BLOCKS_PER_EPOCH: u64 = 5;
 ///     .state_root(Cid::new_from_cbor(&[], Identity)) // required
 ///     .miner_address(Address::new_id(0)) // optional
 ///     .beacon_entries(Vec::new()) // optional
-///     .win_post_proof(Vec::new()) // optional
+///     .winning_post_proof(Vec::new()) // optional
 ///     .election_proof(None) // optional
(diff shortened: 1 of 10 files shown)


dutterbutter (approved):

blockchain/blocks/src/header/mod.rs
```diff
@@ -236,7 +236,7 @@ impl BlockHeader {
     pub fn beacon_entries(&self) -> &[BeaconEntry] {
         &self.beacon_entries
     }
-    /// Getter for window PoSt proof
+    /// Getter for winning PoSt proof
     pub fn win_post_proof(&self) -> &[PoStProof] {
```
ec2, line 240: lmao i can see why this function name can cause confusion (winning vs window)

ec2 (approved):

blockchain/blocks/src/header/mod.rs
austinabell: I mean, there is nothing to confuse it with since there is no window post on the header, but I might as well change the function name too.

This was probably from before the PoSt proof split (iirc the name of it was window? If not idk why this comment)

PR #785, opened 2020-10-25
Wrap cached state in async mutex to avoid duplicate state calculation

### LINKED ISSUE #776: Avoid duplicate state computations by awaiting a seperate task's computation
**Issue summary**

Currently, the blocks will attempt to get a tipset's state and check the cache before running, but now that it is async, we need a locking or some mechanism to put the tasks to sleep until the computation is done, then retrieve the state/receipt root from the cache.

I will most likely do this, but if someone else wants to pick up, let me know and I can give more context

austinabell:
**Summary of changes**
- So I documented the behaviour, but tldr tasks will wait for ones already computing state to finish instead of doing duplicate work in parallel, without blocking separate tipset's state transitions 
- Changed callback function to references, to avoid unnecessary clones

You can test the functionality by uncommenting the full sync test and setting the RUST_LOG env variable to see logs, no duplicate tipset state calculation should happen

Closes #776 

### DIFF
--- a/blockchain/state_manager/src/lib.rs
+++ b/blockchain/state_manager/src/lib.rs
@@ -192,7 +192,7 @@ where
     where
         R: Rand,
         V: ProofVerifier,
-        CB: FnMut(Cid, &ChainMessage, ApplyRet) -> Result<(), String>,
+        CB: FnMut(&Cid, &ChainMessage, &ApplyRet) -> Result<(), String>,
     {
         let mut buf_store = BufferedBlockStore::new(self.bs.as_ref());
         // TODO change from statically using devnet params when needed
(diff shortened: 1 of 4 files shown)


blockchain/state_manager/src/lib.rs
```diff
@@ -218,18 +222,42 @@ where
...
-                return Ok(cid_pair.clone());
+            // Get entry in cache, if it exists.
+            // Arc is cloned here to avoid holding the entire cache lock until function ends.
+            // (tasks should be able to compute different tipset state's in parallel)
+            //
+            // In the case of task `A` computing the same tipset as task `B`, `A` will hold the
+            // mutex until the value is updated, which task `B` will await.
+            //
+            // If two tasks are computing different tipset states, they will only block computation
+            // when accessing/ initializing the entry in cache, not during the whole tipset calc.
```
ec2, line 238: remove space after "/"

blockchain/state_manager/src/lib.rs
```diff
@@ -60,7 +60,11 @@ pub struct MarketBalance {
 
 pub struct StateManager<DB> {
     bs: Arc<DB>,
-    cache: RwLock<HashMap<TipsetKeys, CidPair>>,
+
+    /// This is a cache which indexes tipsets to their calculated state.
+    /// The calculated state is wrapped in a mutex to avoid duplicate computation
+    /// of the state/ receipt root.
```
ec2, line 66: remove space after "/"

ec2 (approved):

AshantiMutinta (approved):

PR #786, opened 2020-10-26
Introduces paychannel logic

dutterbutter:
**Summary of changes**
- Introduces paychannel logic from lotus

Please provide suggestions to improve and pay careful attention to the possibility of deadlocks. The testing suite associated with paychannel in lotus utilizes v2 paych actor which is why it is missing here. Will need to be added and adequately tested in future.

Background information regarding paychannel usage and functionality can be viewed [here](https://spec.filecoin.io/#section-systems.filecoin_token.payment_channels).

Closes 


q9f:
let's revisit this once we get around it again.

PR #789, opened 2020-10-26
Import snapshots and import chains 

### LINKED ISSUE #761: Add import-snapshot flag and functionality
**Issue summary**

We currently do not have a way to use a chain export when running the node. Lotus uses the flag `--import-snapshot` but for our use case we will need to have it import a snapshot past the actors v2 upgrade at least by default.

All of the logic needed to load and validate the chain should exist, but there may have to be a few changes to allow it to be connected from the forest binary before the daemon is started.

ec2:
**Summary of changes**
- Import a snapshot through `--import-snapshot <filename>`
- Import a chain through `--import-chain <filename>`

Closes 
#761 


forest/src/daemon.rs
```diff
@@ -175,3 +218,31 @@ pub(super) async fn start(config: Config) {
...
+    use super::*;
+    use db::MemoryDB;
+    use std::fs::File;
+    use std::io::BufReader;
+    use std::io::Read;
+
+    #[async_std::test]
+    async fn import_snapshot_from_file() {
+        let db = Arc::new(MemoryDB::default());
+        let file = File::open("chain4.car").expect("Snapshot file path not found!");
```
dutterbutter, line 233: I would move the binary file and test into own dir perhaps?

dutterbutter (changes requested):

forest/src/daemon.rs
austinabell: Also, could use `include_bytes!()` instead to avoid io during test if you wanted (doesn't really matter tho since it's a tradeoff)

I would say for the car file at least move it from the forest base directory into a sub folder at least

forest/src/daemon.rs
```diff
@@ -28,6 +35,32 @@ use wallet::{KeyStore, PersistentKeyStore};
...
+    let cids = load_car(bs.as_ref(), reader)?;
+    let ts = chain::tipset_from_keys(bs.as_ref(), &TipsetKeys::new(cids))?;
+    let gb = chain::tipset_by_height(bs.as_ref(), 0, &ts, true)?.unwrap();
+    let sm = StateManager::new(bs.clone());
+    if !snapshot {
+        info!("Validating imported chain");
+        sm.validate_chain::<V>(ts.clone()).await?;
+    }
+    let gen_cid = chain::set_genesis(bs.as_ref(), &gb.blocks()[0])?;
+    bs.write(chain::HEAD_KEY, ts.key().marshal_cbor()?)?;
```
austinabell, line 55: this is done automatically in `set_genesis`, why duplicate?

austinabell (changes requested):

forest/src/daemon.rs
austinabell: this wasn't addressed

blockchain/state_manager/src/lib.rs
```diff
@@ -855,4 +855,36 @@ where
...
+            if ts_chain[i].parent_state() != &last_state {
+                let e: Box<dyn StdError> = format!(
+                    "Tipset chain has state mismatch at height: {:?}",
+                    ts_chain[i].epoch()
+                )
+                .into();
+                return Err(e);
+            }
+            let (st, _) = self.tipset_state::<V>(&ts_chain[i]).await?;
+            last_state = st;
```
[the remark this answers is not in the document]
austinabell: this also not addressed (only commenting because you re-requested review)

forest/src/daemon.rs
dutterbutter: Would it be preferred to call `set_heaviest_tipset` which is already public than to make HEAD_KEY public?

forest/src/daemon.rs
```diff
@@ -60,6 +93,16 @@ pub(super) async fn start(config: Config) {
     let mut db = RocksDb::new(config.data_dir + "/db");
     db.open().unwrap();
     let db = Arc::new(db);
+
+    // Sync from snapshot
```
dutterbutter, line 97: If I understand correctly we dont actually sync from snapshot rather we set the snapshot? Reason I mention this is because I dont think any sync logic is called in import_chain?

ec2: Well, we populate our blockstore from the snapshot file. And furthermore, we can also recompute all the state if we import with `--import-chain`.

austinabell: Yeah if you're wondering why not full block valation, that is intended, it's just calculating intermediate states and validating those roots are correct (full block validation would be too slow when you want to trust the export)

forest/src/daemon.rs
austinabell: I'm not sure this will load correctly to use as the correct head, or have you tested?

blockchain/state_manager/src/lib.rs
```diff
@@ -887,4 +888,42 @@ where
...
+    ) -> Result<(), Box<dyn StdError>> {
+        let mut ts_chain: Vec<Tipset> = vec![ts.clone()];
+        let mut ts = ts;
+        while ts.epoch() != 0 {
+            let next = chain::tipset_from_keys(self.blockstore(), ts.parents())?;
+            ts_chain.push(next.clone());
+            ts = next;
+        }
+        let mut last_state = ts_chain.last().unwrap().parent_state().clone();
+        let (_, mut last_receipt) = self.tipset_state::<V>(ts_chain.last().unwrap()).await?;
```
austinabell, line 904: why not just use the receipts from the tipset as the above does? It's not exposed right now, but there is no reason it shouldn't be. I'm not confident this will always work and it's unnecessary cycles

blockchain/state_manager/src/lib.rs
```diff
@@ -887,4 +888,42 @@ where
...
+        &self,
+        ts: Tipset,
+    ) -> Result<(), Box<dyn StdError>> {
+        let mut ts_chain: Vec<Tipset> = vec![ts.clone()];
+        let mut ts = ts;
+        while ts.epoch() != 0 {
+            let next = chain::tipset_from_keys(self.blockstore(), ts.parents())?;
+            ts_chain.push(next.clone());
+            ts = next;
+        }
```
austinabell, line 902:
```suggestion
        let mut ts_chain: Vec<Tipset> = vec![ts];
        while ts_chain.last().unwrap().epoch() != 0 {
            let last = ts_chain.last().unwrap();
            let next = chain::tipset_from_keys(self.blockstore(), last.parents())?;
            ts_chain.push(next);
        }
```
Can be cleaned up more, but this avoids a lot of clones (cloning every tipset as a duplicate just to reference is unnecessary)

austinabell (changes requested):

blockchain/state_manager/src/lib.rs
timvermeulen: Another option is to keep the `ts` variable, start out with an empty vector, and only add the last one after the loop. Something like
```rust
let mut ts_chain = Vec::<Tipset>::new();
let mut ts = ts;
while ts.epoch() != 0 {
    let next = chain::tipset_from_keys(self.blockstore(), ts.parents())?;
    ts_chain.push(mem::replace(&mut ts, next));
}
ts_chain.push(ts);
```

forest/src/daemon.rs
ec2: @dutterbutter No, i would have to instantiate a ChainStore to do that because set_heaviest_tipset will update some cache stuff in the ChainStore. 

@austinabell Not sure why you think the head will be wrong? I have ran Forest with a snapshot and got the correct head over RPC.

dutterbutter (approved):

austinabell (approved):

PR #790, opened 2020-10-27
Refactor types out of actors crate to prep for v2 upgrade

### LINKED ISSUE #788: Moving types for actors into a different crate
**Issue summary**

This is to facilitate the upgrade to actors v2, while still supporting v1 in the future. Need to look into actors v2 to determine what types are shared. All types within actors that are shared between v1 and v2 must be moved into it's own crate.

austinabell:
**Summary of changes**
- We had most types that were separated by the go implementation separated already, with the exception of `DeadlineInfo`
  - The JSON format is updated now to include parameters, which is a necessary change to match Lotus' RPC interface
- Puts proofs api under feature to only pull in dependency when needed (it's a huge import so avoiding for most crates that just use the types)

Closes #788 


timvermeulen (approved):

vm/actor/src/lib.rs
```diff
@@ -30,16 +33,16 @@ lazy_static! {
 /// Number of token units in an abstract "FIL" token.
 /// The network works purely in the indivisible token amounts.
 /// This constant converts to a fixed decimal with more human-friendly scale.
-const TOKEN_PRECISION: u64 = 1_000_000_000_000_000_000;
+pub const TOKEN_PRECISION: u64 = 1_000_000_000_000_000_000;
```
ec2, line 36: I know it isnt really part of this PR, but while youre at it, could you make this point to the one in types? (i think reward/types.rs also redefines this

ec2 (approved):

vm/actor/src/lib.rs
austinabell: Yeah it was semi intentional since the go implementation redefines -- maybe they want the actors execution to not change even if the params outside changes? That was my thought, but it's not a good reason so I will share same

dutterbutter (approved):

PR #792, opened 2020-10-28
Update Dockerfile and include make command

dutterbutter:
**Summary of changes**
- Updated dockerfile to include required deps and expose correct ports
- Added a make command `make docker-run` that builds and runs docker container
- Moved from old stable to new stable buster
- Upgraded rust image version

**Question:** Should we include docker instructions in readme?

Closes 


Dockerfile
```diff
@@ -27,7 +27,18 @@ RUN cargo install --path forest
...
+    autoconf automake cmake libtool libcurl4 libcurl4-openssl-dev libssl-dev \
+    libelf-dev libdw-dev binutils-dev zlib1g-dev libiberty-dev wget \
+    xz-utils pkg-config python clang ocl-icd-opencl-dev ca-certificates
+
 # Copy over binaries from the build-env
 COPY --from=build-env /usr/local/cargo/bin/forest /usr/local/bin/forest
 
+# Expose ports for RPC
+EXPOSE 1234
+EXPOSE 64899
```
austinabell, line 42: uh, isn't this the libp2p port that is randomly generated unless specified? Might want to make this deterministic if so

austinabell: wait you removed the port, but won't it be necessary to expose the libp2p port to use the node meaningfully? Why not expose an arbitrary port and just set that port in config with the command in the makefile to run?

dutterbutter: Container runs fine without exposing port, if someone does need to expose a port to call into the container they can do so via command line still with the port of their choosing.

Dockerfile
```diff
@@ -27,6 +27,13 @@ RUN cargo install --path forest
 # Prod image for forest binary
 FROM debian:buster-slim
 
+# Install binary dependencies
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y curl file gcc g++ hwloc libhwloc-dev make openssh-client \
+    autoconf automake cmake libtool libcurl4 libcurl4-openssl-dev libssl-dev \
+    libelf-dev libdw-dev binutils-dev zlib1g-dev libiberty-dev wget \
+    xz-utils pkg-config python clang ocl-icd-opencl-dev ca-certificates
```
austinabell, line 35: Do you need these to run? I just want to know what specifically was wrong with the dockerfile

dutterbutter: Yes the binary still needs the dependencies installed or else the error `error while loading shared libraries: libcurl.so.4: cannot open shared object file: No such file or directory` will be present.

dutterbutter: In terms of what was specifically wrong, the previous docker image was using an outdated version, missing required deps of `hwloc libhwloc-dev ca-certificates` and the multistage build didn't have the required deps installed.

dutterbutter: Can definitely make the argument that a multi stage build in this circumstance is not really advantageous, happy to remove if so.

austinabell (approved):

Dockerfile
austinabell: hmm interesting, I wouldn't expect the second image would need those deps, I assume it needed hwloc? maybe we can try to narrow down exactly what's needed for this second image (doesn't have to be now)

ec2 (approved):

PR #803, opened 2020-11-02
Defer bitfield decoding until its first use

### LINKED ISSUE #735: Validate bitfields when they're first used, not at deserialization
We need to match https://github.com/filecoin-project/go-bitfield/pull/32

timvermeulen:
This PR adds the `UnvalidatedBitField` type that may or may not have been validated for valid RLE+ yet. The existing `BitField` type still represents validated bitfields only.

Also adds the `bitfield::Validate` trait which is useful for when a function needs to be able to work with both validated and unvalidated bitfields.

Closes #735


ec2 (approved):

utils/bitfield/src/unvalidated.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            Self::Validated(bf) => serde_bytes::serialize(&bf.to_bytes(), serializer),
+            Self::Unvalidated(bytes) => serde_bytes::serialize(&bytes, serializer),
+        }
+    }
+}
```
austinabell, line 68: optionally, can remove impl for this with serde `untagged` https://serde.rs/enum-representations.html#untagged (prob best to keep the deserialize either way)

austinabell (commented):
LGTM, but can you update the branch so I can make sure there is no regression? Hard to verify specifics with a few of the cases

austinabell (approved):

Issue #759, opened 2020-10-19
Verify tipset, perform state transitions, & save tipset in ChainSyncer 


amerameen:
**@austinabell will look into this more closely and scope out the specifics of how we will implement this**

For a tipset that is formed with blocks received over pubsub, we must verify the tipset, perform the state transtions, update our chain, and save the tipset. 

**Acceptance Criteria**
The head of our chain is updated with the heaviest verified tipset and the state is updated accordingly. This only happens for valid blocks/tipsets. Invalid blocks are discarded.

**Notes**
- The logic for doing this is not yet clear. We could potentially modify informnewhead (used for the initial sync) so that it can do this
- See [Lotus - HandleIncomingBlocks](https://github.com/filecoin-project/lotus/blob/master/chain/sub/incoming.go#L55)
- See [Lotus - informnewhead](https://github.com/filecoin-project/lotus/blob/0f5aa9ba5c7b0f841e5bb376e8ef95f075068b56/chain/sync.go#L250) modified for syncing blocks received over pubsub

PR #808, opened 2020-11-04
Process blocks coming off of GossipSub

### LINKED ISSUE #757: Fetch messages for block received over pubsub
Blocks that are received over pubsub only have the CIDs for the included messages. We need the messages in order to process the block (i.e. verify it, perform state transitions, update our chain).

**Acceptance Criteria**
We have all the messages for a block received over pubsub

** Steps **
- [ ] Check which messages we already have in the blockstore or cache 
- [ ] Using bitswap, retrieve any messages we don't already have

ec2:
**Summary of changes**
- Process GossipBlocks in ChainSyncer
- Fetch missing messages from GossipBlocks over Bitswap
- Queue block through inform_new_head

To test this on a Lotus devnet, you will need to edit miner and power actor policies to allow for 2K sectors and minimum consensus power to 2048 instead of 10 << 40

This is the fork of bitswap we use now: https://github.com/ChainSafe/libp2p-bitswap/commit/bd7e083cc5d9ad13bb357e0e0055338c33b18173

Closes 
#757
#759 (if I understand the scope of that properly)


ec2:
Oops forgot to fix unit tests that this breaks...

blockchain/chain_sync/src/network_context.rs
```diff
@@ -93,6 +114,38 @@ impl SyncNetworkContext {
         Ok(fts.remove(0))
     }
 
+    /// Requests that some content with a particular Cid get fetched over Bitswap if it doesn't
+    /// exist in the BlockStore.
+    pub async fn bitswap_get<TMessage: DeserializeOwned>(
+        &self,
+        content: Cid,
+    ) -> Result<TMessage, String> {
+        // TODO: check to see if we have it in our database
```
austinabell, line 123: why not just do it in this PR? It's extremely low cost since self has a blockstore

blockchain/chain_sync/src/network_context.rs
```diff
@@ -93,6 +114,38 @@ impl SyncNetworkContext {
...
+        match res {
+            Ok(Ok(())) => {
+                match self.db.get(&content) {
+                    Ok(Some(b)) => Ok(b),
+                    Ok(None) => Err(format!("Bitswap response successful for: {:?}, but can't find it in the database", content)),
+                    Err(e) => Err(format!("Bitswap response successful for: {:?}, but can't retreive it from the database: {}", content, e.to_string())),
+                }
+            }
+            Err(e) => {
+               Err(format!("Bitswap get for {:?} timed out: {}", content, e.to_string()))
```
austinabell, line 141: I think adding the error string is redundant, it will just say timed out

blockchain/chain_sync/src/sync.rs
```diff
@@ -193,6 +194,38 @@ where
...
+                                for m in b.bls_messages.clone() {
+                                    let bmsg: UnsignedMessage = self.network.bitswap_get(m).await.unwrap();
+                                    bmsgs.push(bmsg);
+                                }
+                                // Get secp_messages in the store or over Bitswap
+                                let mut smsgs: Vec<SignedMessage> = Vec::with_capacity(b.secpk_messages.len());
+                                for m in b.secpk_messages.clone() {
+                                    let smsg: SignedMessage = self.network.bitswap_get(m).await.unwrap();
+                                    smsgs.push(smsg);
+                                }
```
austinabell, line 212: two problems I see with these
1. Unwraps when error is very common
2. you're performing the requests in sequence since you're awaiting on all of them one at a time

Either you should spawn tasks for each and join after all the requests have started, or the bitswap_get should return the receiver channel and then all should be awaited after requests sent

blockchain/chain_sync/src/sync.rs
```diff
@@ -193,6 +194,38 @@ where
...
+                                // Form block
+                                let block = Block {
+                                    header: b.header,
+                                    bls_messages: bmsgs,
+                                    secp_messages: smsgs,
+                                };
+                                let ts = FullTipset::new(vec![block]).unwrap();
+                                if let Err(e) = self.inform_new_head(source.clone().unwrap(), &ts).await {
```
austinabell, line 220: the source isn't guaranteed because I don't believe we reject responses with no peer id set, should handle this in some way

utils/genesis/src/lib.rs
```diff
@@ -63,6 +63,7 @@ where
     if genesis_cids.len() != 1 {
         panic!("Invalid Genesis. Genesis Tipset must have only 1 Block.");
     }
+    println!("Do we even get here?");
```
austinabell, line 66: lol

node/forest_libp2p/src/service.rs
```diff
@@ -383,3 +410,6 @@ pub fn get_keypair(path: &str) -> Option<Keypair> {
         },
     }
 }
+
+#[cfg(test)]
+mod tests {}
```
austinabell, line 415:
```suggestion
```

node/forest_libp2p/src/service.rs
```diff
@@ -309,6 +324,18 @@ where
...
+                        NetworkMessage::BitswapRequest { cid, response_channel } => {
+                            if let Err(e) = swarm_stream.get_mut().want_block(cid.clone(), 1000) {
+                                warn!("Failed to send a bitswap want_block: {}", e.to_string());
+                            } else {
+                                if let Some(chans) = self.bitswap_response_channels.get_mut(&cid) {
+                                    chans.push(response_channel);
+                                } else {
+                                    self.bitswap_response_channels.insert(cid, vec![response_channel]);
+                                }
+                                debug!("");
```
austinabell, line 336: ?

blockchain/chain_sync/src/network_context.rs
```diff
@@ -93,6 +114,38 @@ impl SyncNetworkContext {
...
+            }
+            Ok(Err(e)) => {
+                Err(format!("Bitswap get for {:?} failed: {}", content, e.to_string()))
+            }
+        }
```
austinabell, line 146: why does this have to be all matches when it's just returning errors or the result?

austinabell (changes requested):

blockchain/chain_sync/src/sync.rs
ec2: Good catch. Didn't see any performance issues because i was running on devnet, but its pretty clear this will not work in mainnet

blockchain/chain_sync/src/sync.rs
ec2: The source address should be guaranteed because of the signing policies we enforce on gossipsub messages.

blockchain/chain_sync/src/network_context.rs
ec2: Just wanted to return more concise err messages.

dutterbutter (approved):

blockchain/chain_sync/src/sync.rs
austinabell: no, actually we explicitly allow this (because with out it before it was an issue) If you want to ensure this is the case, you need to change the `ValidationMode` on the GossipSub config. In any case, it would be much safer if there wasn't this unwrap when dealing with inputs directly from the network as it's too easy to trigger

ec2: I've changed the validation mode to Strict and ive tested it against a local lotus devnet.

blockchain/chain_sync/src/sync.rs
```diff
@@ -193,6 +195,40 @@ where
...
+                    Some(NetworkEvent::PubsubMessage { source, message }) => {
+                        match message {
+                            forest_libp2p::PubsubMessage::Block(b) => {
+                                info!("Received block over GossipSub: {} from {}", b.header.epoch(), source.clone().unwrap());
+                                // Get bls_messages in the store or over Bitswap
+                                let bmsgs: Vec<_> = b.bls_messages.into_iter().map(|m| self.network.bitswap_get::<UnsignedMessage>(m)).collect();
+                                let bmsgs = try_join_all(bmsgs).await;
+                                if let Err(e) = &bmsgs {
+                                    warn!("Failed to get UnsignedMessage: {}", e);
+                                    ()
```
austinabell, line 207:
```suggestion
```

blockchain/chain_sync/src/sync.rs
```diff
@@ -193,6 +195,40 @@ where
                             )
                             .await
                     }
+                    Some(NetworkEvent::PubsubMessage { source, message }) => {
+                        match message {
+                            forest_libp2p::PubsubMessage::Block(b) => {
+                                info!("Received block over GossipSub: {} from {}", b.header.epoch(), source.clone().unwrap());
```
austinabell, line 201: You're calling the source clone and unwrap three times in this block, consider doing once or handling this better (I get that the unwrap should not fail, but doesn't mean it shouldn't be handled)

node/forest_libp2p/src/behaviour.rs
```diff
@@ -301,7 +301,7 @@ impl ForestBehaviour {
         let local_peer_id = local_key.public().into_peer_id();
         let gossipsub_config = GossipsubConfig {
             // TODO revisit validation (permissive validation allows unsigned messages)
```
austinabell, line 303:
```suggestion
```

node/forest_libp2p/src/service.rs
```diff
@@ -262,13 +268,22 @@ where
                             };
                         }
                         ForestBehaviourEvent::BitswapReceivedBlock(peer_id, cid, block) => {
-                            let res: Result<_, String> = self.db.put(&block, Blake2b256).map_err(|e| e.to_string());
+                            let res: Result<_, String> = self.db.put_raw(block.to_vec(), Blake2b256).map_err(|e| e.to_string());
```
austinabell, line 271:
```suggestion
                            let res: Result<_, String> = self.db.put_raw(block.into(), Blake2b256).map_err(|e| e.to_string());
```
`to_vec` clones

blockchain/chain_sync/src/sync.rs
```diff
@@ -193,6 +195,40 @@ where
...
+                                // Get bls_messages in the store or over Bitswap
+                                let bmsgs: Vec<_> = b.bls_messages.into_iter().map(|m| self.network.bitswap_get::<UnsignedMessage>(m)).collect();
+                                let bmsgs = try_join_all(bmsgs).await;
+                                if let Err(e) = &bmsgs {
+                                    warn!("Failed to get UnsignedMessage: {}", e);
+                                    ()
+                                }
```
austinabell, line 208: join after you send the signed message requests out, no reason to await all of these to get the secp ones

blockchain/chain_sync/src/network_context.rs
austinabell: This is more confusing imo, at least for reading, I don't see why you couldn't at least return the timeout error before this match

austinabell (approved):

PR #809, opened 2020-11-04
Refactor chainstore and related components

austinabell:
**Summary of changes**
- A lot of functions were not using the `StateManager` or `ChainStore`, which was leading to incorrect behaviour in some cases, and also this allows us to build caching for these things to avoid unnecessary work

There should be no functional change from this, aside from fixing the incorrect behaviour that existed

Closes 

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -81,15 +81,16 @@ impl EventsPayload {
 /// This structure is threadsafe, and all caches are wrapped in a mutex to allow a consistent
 /// `ChainStore` to be shared across tasks.
 pub struct ChainStore<DB> {
+    /// Publisher for head change events
     publisher: RwLock<Publisher<HeadChange>>,
 
-    // key-value datastore
+    /// key-value datastore.
     pub db: Arc<DB>,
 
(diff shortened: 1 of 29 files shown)


blockchain/chain/src/store/chain_store.rs
```diff
@@ -261,7 +283,147 @@ where
...
+                "searching for tipset that has a height less than starting point".to_owned(),
+            ));
+        }
+        if height == ts.epoch() {
+            return Ok(None);
+        }
+        // TODO get tipset by height using cache instead of reloading tipsets
+        let mut ts_temp: Option<Tipset> = None;
+        loop {
+            let pts = if let Some(temp) = &ts_temp {
```
dutterbutter, line 312: Could you use the `tip_index.rs` here to act as the cache? If not what ts cache would that be?

dutterbutter (approved):

blockchain/chain/src/store/chain_store.rs
austinabell: tip index is really broken right now and I'd rather not add any more usages until #814 is done. (I didn't change this code, just was moved)

timvermeulen (approved):

PR #811, opened 2020-11-05
Allow importing snapshot from URL + Progress bar (#762)

### LINKED ISSUE #762: Allow snapshot to be loaded from an http request and add progress
**Issue summary**

Once #761 is in, would be very nice to allow for loading a snapshot from a url, as Lotus does it. 

They also added in a progress bar for the snapshot import, and I think this would be a very nice UX to add. For this progress to show up well, it has to be before any of the async tasks are started, as the logger conflicts with the progress bar output (see `fetch-params` progress bar when requests are done async), but should be pretty straightforward

This issue is definitely low priority since it wouldn't add any new functionality.

snaumov:
**Summary of changes**
- Ability to import the snapshot from the remote url
- Download progress indicator

forest

Closes #762 


utils/net_utils/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "net_utils"
+version = "0.1.0"
+authors = ["ChainSafe Systems <info@chainsafe.io>"]
+edition = "2018"
+
+[dependencies]
+indicatif = "0.15.0"
+futures = "0.3.5"
+async-std = { version = "1.6.3" }
+isahc = "0.9.11"
```
snaumov, line 11: `isahc` was selected because

1. `reqwest` depends on `tokio`, not `async-std`
2. I attempted `surf` but got some issues https://github.com/http-rs/surf/issues/218#issuecomment-721774449
3. `surf` uses `isahc` as one of the backends anyways.

utils/net_utils/src/download.rs
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Debug, Error)]
+enum DownloadError {
+    #[error("Cannot read a file header")]
+    HeaderError,
+    #[error("Filename encoding error")]
+    EncodingError,
+}
+
+impl<S: AsyncRead + Unpin> AsyncRead for DownloadProgress<S> {
+    fn poll_read(
```
snaumov, line 34: implementing synchronous `Read` would be another option, I initially implemented `AsyncRead` when attempting to use `reqwest` as `reqwest` only implements `AsyncRead` and stick to it since.

austinabell: Reason for not reusing logic from paramfetch which does the same as this?

utils/net_utils/src/download.rs
```diff
@@ -0,0 +1,103 @@
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<Result<usize, io::Error>> {
+        match Pin::new(&mut self.inner).poll_read(cx, buf) {
+            Poll::Ready(Ok(size)) => {
+                if size == 0 {
+                    self.progress_bar.finish();
```
austinabell, line 42: This case won't be hit (unless I'm missing something here)

utils/net_utils/src/download.rs
```diff
@@ -0,0 +1,103 @@
...
+        inner: request.body_mut(),
+    };
+
+    let mut dest = fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(&file)
+        .await?;
+
+    let _ = copy(&mut source, &mut dest).await?;
```
austinabell, line 96:
```suggestion
    copy(&mut source, &mut dest).await?;
```

forest/src/daemon.rs
```diff
@@ -36,11 +36,23 @@ use wallet::{KeyStore, PersistentKeyStore};
...
 ) -> Result<(), Box<dyn std::error::Error>> {
+    let is_remote_file: bool = path.starts_with("http://") || path.starts_with("https://");
+
+    let mut file_path = path;
+    if is_remote_file {
+        match download_file(file_path).await {
+            Ok(file) => file_path = file,
+            Err(err) => return Err(err),
+        }
+    }
```
austinabell, line 52: There is no need to save the file to file storage before loading it, it's completely unnecessary and will be very inefficient with very large car files. Should just load data received from network as the reader for `load_car`

utils/net_utils/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "net_utils"
+version = "0.1.0"
+authors = ["ChainSafe Systems <info@chainsafe.io>"]
+edition = "2018"
+
+[dependencies]
+indicatif = "0.15.0"
```
austinabell, line 8: Can you share the same progress bar implementation as paramfetch? Doesn't matter much to me which one is used, but it's weird to see multiple types of bars

austinabell (changes requested):
Happy to make a pass after to clean up some of these things, but the one thing that absolutely needs to change is writing to a temporary file just to read and delete it

austinabell:
btw, conformance test vectors are failing because you committed a previous commit of the repo to this PR

utils/net_utils/src/download.rs
snaumov: oh, didn't know it existed. However, `FetchProgress` has `inner` defined as `ProgressBar<prb::Pipe>`, and it's a multibar, in my case I wanted to use single bar and have `inner` as `ProgressBar<Stdout>`. 

Attempted to make `inner` be generic enough to cover both types, ie `ProgressBar<W: Write>`, but didn't really succeed so far.

utils/net_utils/src/download.rs
snaumov: not relevant in a new iteration of code.

forest/src/daemon.rs
snaumov: good point. My intent originally was to have a generic function that will be used for file downloading. However it's better to just return a `Reader`. Updated now.

utils/net_utils/Cargo.toml
snaumov: using `pbr` now

utils/net_utils/src/download.rs
snaumov: actually, nevermind, `ProgressBar<W>` works well enough!

utils/paramfetch/src/lib.rs
```diff
@@ -149,45 +146,6 @@ async fn fetch_verify_params(
     })
 }
 
-pin_project! {
```
snaumov, line 152: These struct and impls have been migrated to `net-tools` for reuse

utils/net_utils/src/download.rs
```diff
@@ -0,0 +1,97 @@
...
+        self.inner.read(buf).map(|n| {
+            self.progress_bar.add(n as u64);
+            n
+        })
+    }
+}
+
+/// Builds Reader for a provided URL.
+pub fn make_reader(
+    raw_url: String,
```
austinabell, line 71: this can be a `&str` right?

utils/genesis/src/lib.rs
```diff
@@ -89,17 +90,27 @@ where
 
 /// Import a chain from a CAR file. If the snapshot boolean is set, it will not verify the chain
 /// state and instead accept the largest height as genesis.
-pub async fn import_chain<V: ProofVerifier, R: Read, DB>(
+pub async fn import_chain<V: ProofVerifier, DB>(
     sm: &Arc<StateManager<DB>>,
-    reader: R,
+    path: String,
```
austinabell, line 95: this can also be a `&str`

utils/genesis/src/lib.rs
```diff
@@ -89,17 +90,27 @@ where
...
     info!("Importing chain from snapshot");
     // start import
-    let cids = load_car(sm.blockstore(), reader)?;
+    let cids = if is_remote_file {
+        let reader = make_reader(path)?;
+        info!("Downloading file...");
+        load_car(sm.blockstore(), reader)?
+    } else {
+        let reader = File::open(&path).expect("Snapshot file path not found!");
+        load_car(sm.blockstore(), reader)?
```
austinabell, line 111: Can you wrap the reader in `BufReader`? It's a lot better for reading files

utils/net_utils/src/download.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn read(&mut self, buf: &mut [u8]) -> IOResult<usize> {
+        self.inner.read(buf).map(|n| {
+            self.progress_bar.add(n as u64);
+            n
+        })
+    }
+}
+
+/// Builds Reader for a provided URL.
+pub fn make_reader(
```
austinabell, line 70: can you rename this function? This won't work in a generalized case and can only be used for car files over the network, perhaps `fetch_car_file` or something?

snaumov: my intent was to make it usable in a generalized case - anything else is missed you think?

austinabell: ah, my apologies, I misread something. Optionally can just indicate that the reader is coming from an http request, since `make_reader` can be ambiguous to any type of reader (possibly confusing with a file reader since it's only a string param)

snaumov: yeah. Renamed it to `make_http_reader` and passing `url: Url` instead of `&str` to minimize the confusion

utils/genesis/src/lib.rs
snaumov: updated

utils/net_utils/src/download.rs
snaumov: replaced with `Url`

utils/genesis/src/lib.rs
snaumov: updated

austinabell (approved):
LGTM great changes, thanks for making those adjustments! One thing about this is that it doesn't add the progress bar to loading a local car file (and those imports can take a long time) but this can be easily added after this PR comes in

snaumov:
@austinabell right. I thought about this too at some point :) I can rename `make_http_reader` to `make_reader` back, and make it accept `&str` and check if the incoming `str` is a Url or local path, and depending on that either return `FetchProgress` that reads from `url` or from local file. Thoughts?

austinabell:
> @austinabell right. I thought about this too at some point :) I can rename `make_http_reader` to `make_reader` back, and make it accept `&str` and check if the incoming `str` is a Url or local path, and depending on that either return `FetchProgress` that reads from `url` or from local file. Thoughts?

Or you can just wrap the file reader in that progress bar (capacity you can get from the file metadata). Either way should work, it's up to you. My intuition would be that the generics of the former would be a little tricky to workaround, but it would be a slightly cleaner solution indeed

snaumov:
> > @austinabell right. I thought about this too at some point :) I can rename `make_http_reader` to `make_reader` back, and make it accept `&str` and check if the incoming `str` is a Url or local path, and depending on that either return `FetchProgress` that reads from `url` or from local file. Thoughts?
> 
> Or you can just wrap the file reader in that progress bar (capacity you can get from the file metadata). Either way should work, it's up to you. My intuition would be that the generics of the solution you suggest would be a little tricky to workaround, but it would be a slightly cleaner solution indeed

Updated MR, using `TryFrom` now instead for both `File` and `Url`

ec2 (approved):

ec2:
Looks great! Thanks as always for the contribution!!!!!!!

austinabell (approved):
oo very nice, thanks!

PR #841, opened 2020-11-10
Rpc fixes and implementations

### LINKED ISSUE #800: MinerGetBaseInfo needed for Storage Miner
**Issue summary**
In order to initialize storage miner, MinerGetBaseInfo should be implemented.

https://github.com/filecoin-project/lotus/blob/master/api/api_full.go#L234

### LINKED ISSUE #830: Add GetNtwkVersion to State Manager
**Issue summary**

Implement a function to get network version based on the epoch in the State Manager. 

[See Lotus implementation](https://github.com/filecoin-project/lotus/blob/efd2dff0cac40934ec5b4472e48887ecbb2efe44/chain/stmgr/stmgr.go#L1140)

[We have this implemented in Forest already](https://github.com/ChainSafe/forest/blob/0b7a49b6ac3879ecea953c487218a994a91c14f7/types/src/version.rs#L66)

AshantiMutinta:
**Summary of changes**
- This implements various fixes for the rpc layer

Closes #830 
Closes #800 


vm/actor/src/builtin/miner/state.rs
```diff
@@ -1024,7 +1138,25 @@ mod tests {
...
+            pending_worker_key: None,
+            peer_id: PeerId::random().into_bytes(),
+            multi_address: vec![BytesDe(PeerId::random().into_bytes())],
+            sector_size: SectorSize::_2KiB,
+            seal_proof_type: RegisteredSealProof::from(1),
+            window_post_partition_sectors: 0,
+        };
+        let val = serde_json::to_string(&info);
+        println!("miner info {:?}", val);
+    }
```
austinabell, line 1161: This test doesn't do anything, consider asserting the format? (You also shouldn't need the debug print for this)

vm/address/src/lib.rs
```diff
@@ -359,4 +359,92 @@ pub mod json {
...
+        use std::borrow::Cow;
+
+        pub fn serialize<S>(v: &Option<Address>, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer,
+        {
+            if let Some(unwrapped_address) = v.as_ref() {
+                serializer.serialize_str(&encode(unwrapped_address))
+            } else {
+                serializer.serialize_str("\u{003c}\u{0065}\u{006d}\u{0070}\u{0074}\u{0079}\u{003e}")
```
austinabell, line 411: wait, what is this?

vm/actor/src/builtin/miner/state.rs
```diff
@@ -1005,6 +1025,100 @@ impl MinerInfo {
...
+    pub struct MinerInfoJson {
+        pub owner: AddressJson,
+
+        pub worker: AddressJson,
+
+        #[serde(with = "address::json::opt")]
+        pub new_worker: Option<Address>,
+
+        #[serde(with = "address::json::opt_vec")]
+        pub control_addresses: Option<Vec<Address>>, // Must all be ID addresses.
```
austinabell, line 1068: This should just be a vec?

vm/actor/src/builtin/miner/state.rs
```diff
@@ -1005,6 +1025,100 @@ impl MinerInfo {
...
+
+        pub worker_change_epoch: ChainEpoch,
+
+        /// Optional worker key to update at an epoch
+        pub pending_worker_key: Option<WorkerKeyChange>,
```
austinabell, line 1073: You shouldn't need to document or expose the fields? Or is there an issue with that?

vm/address/src/lib.rs
```diff
@@ -359,4 +359,92 @@ pub mod json {
...
+                let mut seq = serializer.serialize_seq(Some(unwrapped_addresses.len()))?;
+                for e in unwrapped_addresses {
+                    seq.serialize_element(&AddressJsonRef(e))?;
+                }
+                seq.end()
+            } else {
+                serializer.serialize_none()
+            }
+        }
+    }
```
austinabell, line 449: This should never be necessary, best not to fall into this pattern of wrapping empty vecs in an `Option`

blockchain/beacon/src/drand.rs
```diff
@@ -34,6 +35,40 @@ impl DrandPublic {
     }
 }
 
+pub struct Schedule<T>(pub Vec<BeaconPoint<T>>)
+where
+    T: Beacon;
```
austinabell, line 40: you shouldn't need to constrain T here

blockchain/beacon/src/drand.rs
```diff
@@ -34,6 +35,40 @@ impl DrandPublic {
...
+            .map(|s| &s.beacon)
+        {
+            Ok(&*beacon_point)
+        } else {
+            self.0
+                .first()
+                .map(|s| &*s.beacon)
+                .ok_or_else(|| Box::from("Could not get first_value of beacon in beacon_for_epoch"))
+        }
+    }
```
austinabell, line 61: Should maybe use shared logic as in `types::version` or connect it in some way so we don't have two similar functionalities not connected?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -534,6 +534,91 @@ where
...
+        .map(|(header, (bls_messages, secp_messages))| Block {
+            header,
+            bls_messages,
+            secp_messages,
+        })
+        .collect();
+
+    // the given tipset has already been verified, so this cannot fail
+    Ok(FullTipset::new(blocks).unwrap())
+}
```
austinabell, line 588:
```suggestion
```
No need to reintroduce this

blockchain/chain/src/store/chain_store.rs
```diff
@@ -610,6 +695,87 @@ pub fn draw_randomness(
...
+                return Ok(Some(pts));
+            }
+            return Ok(ts_temp);
+        }
+        if height == pts.epoch() {
+            return Ok(Some(pts));
+        }
+        ts_temp = Some(pts);
+    }
+}
```
austinabell, line 777: please revert this as well, shouldn't have a regression in this (there are the functions attached to `ChainStore` you can use

blockchain/chain/src/store/chain_store.rs
```diff
@@ -534,6 +534,91 @@ where
...
+    address: &Address,
+    tsk: &TipsetKeys,
+) -> Result<miner::State, Error>
+where
+    DB: BlockStore,
+{
+    let state = parent_state_tsk(db, tsk)?;
+    let actor = state
+        .get_actor(address)
+        .map_err(|_| Error::Other("Failure getting actor".to_string()))?
```
austinabell, line 602: why ignore the underlying error?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -534,6 +534,91 @@ where
...
+    Ok(act)
+}
+
+fn parent_state_tsk<'a, DB>(db: &'a DB, key: &TipsetKeys) -> Result<StateTree<'a, DB>, Error>
+where
+    DB: BlockStore,
+{
+    let ts = tipset_from_keys(db, key)?;
+    StateTree::new_from_root(db, ts.parent_state())
+        .map_err(|_| Error::Other("Could not get actor state".to_string()))
```
austinabell, line 619: same here, use `format!` or something

austinabell (changes requested):

vm/address/src/lib.rs
AshantiMutinta: This serializes the empty address. If you look at the NewWorker in lotus it should be expecting "\u003cempty\u003e"

vm/actor/src/builtin/miner/state.rs
AshantiMutinta: Vec will still serialize to [] need to serialize it to nil

blockchain/beacon/src/drand.rs
AshantiMutinta: Just following lotus implementing in which the beacon can be switched around. Since it is all at compile time, don't see the issue but will change

vm/actor/src/builtin/miner/state.rs
austinabell: Why? The go implementation should be able to deserialize the empty array the same as nil

vm/address/src/lib.rs
AshantiMutinta: This is the difference between serialization of an empty vec that returns nil verses an empty vec that serializes an empty set. An empty optional vec will serialize as nil while an empty vec will serialize as [] https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ed7ce53123d1bc5418b3bb69bda0c8b2

this is matching the data packets that lotus is expecting, had it as a vector before but found failure

blockchain/chain/src/store/chain_store.rs
AshantiMutinta: Probably an issue with the merge

vm/address/src/lib.rs
austinabell: I understand that is what Lotus provides to it's miner, my point is that the Lotus miner will parse the empty array the same as nil, and then we don't have to follow this bad pattern of ambiguous formats and handling the `None` case

blockchain/beacon/src/drand.rs
austinabell: Yeah, you just don't need to constrain it _there_. It should be constrained only on the implementation that requires it

vm/actor/src/builtin/miner/state.rs
AshantiMutinta: This is the issue I came across when running it against the miner

vm/address/src/lib.rs
AshantiMutinta: Okay,I had issues running an empty array but I will switch it to a vector

vm/address/src/lib.rs
austinabell: so can you at least switch this for `<empty>` instead of the codes?

vm/address/src/lib.rs
austinabell: I'm not saying that's absolutely the case, but there is other instances that had been tested and assume that the go implementation deserializes it as such. Are you sure, or can you reproduce this with this difference? I don't want to handle this pattern for every single vector just to match the go daemon implementation's json output, if it's functionally the same

key_management/src/keystore.rs
```diff
@@ -84,13 +90,15 @@ pub mod json {
...
         let JsonHelper {
             sig_type,
             private_key,
         } = Deserialize::deserialize(deserializer)?;
+        println!("deserialize key info");
         Ok(KeyInfo {
-            key_type: sig_type,
-            private_key: base64::decode(private_key).map_err(de::Error::custom)?,
+            key_type: sig_type.0,
+            private_key: base64::decode(private_key).unwrap(),
```
austinabell, line 101: why'd you swap handling to unwrap?

key_management/src/keystore.rs
```diff
@@ -84,13 +90,15 @@ pub mod json {
     where
         D: Deserializer<'de>,
     {
+        println!("deserialize json helper");
         let JsonHelper {
             sig_type,
             private_key,
         } = Deserialize::deserialize(deserializer)?;
+        println!("deserialize key info");
```
austinabell, line 98:
```suggestion
        let JsonHelper {
            sig_type,
            private_key,
        } = Deserialize::deserialize(deserializer)?;
```

key_management/src/wallet.rs
```diff
@@ -176,6 +176,21 @@ pub fn find_key<T: KeyStore>(addr: &Address, keystore: &T) -> Result<Key, Error>
     Ok(new_key)
 }
 
+pub fn try_find<T: KeyStore>(addr: &Address, keystore: &mut T) -> Result<KeyInfo, Error> {
+    let key_string = format!("wallet-{}", addr.to_string());
+    match keystore.get(&key_string) {
+        Ok(k) => Ok(k),
+        Err(_) => {
+            let mut new_addr = addr.to_string();
+            new_addr.replace_range(0..1, "t");
```
austinabell, line 185: Can you add some sort of todo to replace the network prefix default in address, or change this with it? Whenever that is changed, this would be a very tricky workaround to debug?

vm/address/src/lib.rs
austinabell: this can still be removed

node/rpc/src/wallet_api.rs
```diff
@@ -182,40 +196,55 @@ where
...
+        }
+    };
 
     let sig = wallet::sign(
         *key.key_info.key_type(),
         key.key_info.private_key(),
         msg.as_slice(),
     )?;
-
+    println!("sig");
```
austinabell, line 235:
```suggestion
```

vm/actor/src/builtin/miner/state.rs
```diff
@@ -1064,6 +1163,7 @@ mod tests {
             seal_proof_type: RegisteredSealProof::from(1),
             window_post_partition_sectors: 0,
         };
+        println!("miner info {:?}", &info);
```
austinabell, line 1166:
```suggestion
```

crypto/src/signature.rs
```diff
@@ -343,4 +343,46 @@ pub mod json {
...
+        }
+
+        pub fn deserialize<'de, D>(deserializer: D) -> Result<SignatureType, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            println!("deserializer");
+            let json_enum: JsonHelperEnum = Deserialize::deserialize(deserializer)?;
+
+            println!("deserializer json {:?}", &json_enum);
```
austinabell, line 380:
```suggestion
            let json_enum: JsonHelperEnum = Deserialize::deserialize(deserializer)?;
```

blockchain/state_manager/src/lib.rs
```diff
@@ -471,6 +486,211 @@ where
...
+        schedule: &Schedule<B>,
+        key: &TipsetKeys,
+        round: ChainEpoch,
+        address: Address,
+    ) -> Result<Option<MinerBaseInfo>, Box<dyn StdError>> {
+        let tipset = self.cs.tipset_from_keys(key).await?;
+        let prev = match self.cs.latest_beacon_entry(&tipset).await {
+            Ok(prev) => prev,
+            Err(err) => {
+                if std::env::var("LOTUS_IGNORE_DRAND")
```
austinabell, line 636: our equivalent is just `IGNORE_DRAND` and is a constant you can reference

blockchain/state_manager/src/lib.rs
```diff
@@ -471,6 +486,211 @@ where
...
+        V: ProofVerifier,
+    {
+        let mut lbr: ChainEpoch = ChainEpoch::from(0);
+        let version = self
+            .network_versions
+            .iter()
+            .find(|s| round == s.at_or_below)
+            .map(|s| s.network_version)
+            .unwrap_or(self.latest_version);
+        let lb = if version as u8 <= 3 {
```
austinabell, line 504: Ideal to compare against version type, not as u8

austinabell (changes requested):

blockchain/beacon/src/drand.rs
ec2: I dont see how this is connected to anything in types::version...

austinabell: ah my bad, was a misread on my part

crypto/src/signature.rs
```diff
@@ -343,4 +343,44 @@ pub mod json {
...
+        {
+            let json_enum: JsonHelperEnum = Deserialize::deserialize(deserializer)?;
+
+            let signature_type = match json_enum {
+                JsonHelperEnum::Bls => SignatureType::BLS,
+                JsonHelperEnum::Secp256k1 => SignatureType::Secp256k1,
+            };
+            Ok(signature_type)
+        }
+    }
```
austinabell, line 385: Is there really an inconsistency in signature json representations? It is serialized as an integer inside of Signature json (is this wrong and we can just make this the default serialization impl of SignatureType?)

types/src/version.rs
```diff
@@ -42,30 +47,42 @@ const MAINNET_SCHEDULE: [Upgrade; 5] = [
...
     },
+    Upgrade {
+        height: UPGRADE_KUMQUAT_HEIGHT,
+        network: NetworkVersion::V6,
+    },
 ];
 
 /// Specifies the network version
-#[derive(Debug, PartialEq, Clone, Copy, PartialOrd)]
+#[derive(Debug, PartialEq, Clone, Copy, PartialOrd, Serialize)]
```
austinabell, line 57: Should this be serialize repr? what is this needed for?

vm/address/src/lib.rs
```diff
@@ -359,4 +359,70 @@ pub mod json {
...
+                serializer.serialize_str("\u{003c}\u{0065}\u{006d}\u{0070}\u{0074}\u{0079}\u{003e}")
+            }
+        }
+
+        pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<Address>, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            let address_as_string: Cow<'de, str> = Deserialize::deserialize(deserializer)?;
+            if address_as_string == "\u{003c}\u{0065}\u{006d}\u{0070}\u{0074}\u{0079}\u{003e}" {
```
austinabell, line 420: putting note to also change this to same as above

austinabell (changes requested):

austinabell:
I just added a bunch of notes to index the things that should be looked into, @ec2 let me know which (if any) you'd like me to update. I don't want to step on your toes with the changes you're doing

blockchain/chain_sync/src/sync_worker.rs
```diff
@@ -445,8 +445,8 @@ where
             .map_err(|e| (*block_cid, e.into()))?;
 
         // Retrieve lookback tipset for validation.
-        let lbts = cs
-            .get_lookback_tipset_for_round(base_ts.clone(), block.header().epoch())
+        let (lbts, _) = sm
```
austinabell, line 448: can get rid of the tipset state retrieval below if you just get it from here (Tipset state isn't calculated twice, but can still avoid the atomic access through mutex)

crypto/src/signature.rs
ec2: Yeah, it is serialized in json as `bls` or `secp256k1`. Ill make it the default.

ec2: Lol nvm, i just double checked, there is an inconsistancy. In the wallet rpc, they'll serialize the signature type as `bls` or as `secp256k1`, but in the blockheader for example, they'll use 1 or 2... So i think it is fine to leave this as is

types/src/version.rs
ec2: looks like this should indeed be serialize_repr

crypto/src/signature.rs
austinabell: so weird hahahah

ec2: Probably one of the lesser weird things xD

types/src/version.rs
```diff
@@ -42,30 +47,42 @@ const MAINNET_SCHEDULE: [Upgrade; 5] = [
...
     /// actors v2 (specs-actors v2.0.x)
-    V4,
+    V4 = 4,
     /// tape (increases max prove commit size by 10x)
-    V5,
-    /// kumquat (specs-actors v2.2.0)
-    V6,
+    V5 = 5,
+    // kumquat   (specs-actors v2.2.0)
+    V6 = 6,
```
austinabell, line 73: none of the integers matter, but if it is to be added, should also add to V7 (these values are already implied though)

austinabell (approved):
take approval with grain of salt because I was last to push to this branch

ec2:
> take approval with grain of salt because I was last to push to this branch

Your changes LGTM

PR #842, opened 2020-11-11
Use concrete implementations

vmx:
**Summary of changes**

Instead of trait bounds, use concrete implementations. I think this improves the code, I prefer having less trait bounds.

It is possible to change this due to the upgrade to `multihash` 0.13.


austinabell (approved):
I mean, this does just limit the API externally to not allow other multihash codes to be used with our tooling, but I doubt that usage exists and it's not a priority.

It is for sure cleaner

dutterbutter (approved):

austinabell:
Thanks! :)

PR #847, opened 2020-11-12
Update amt for_each caching mechanisms and usages

austinabell:
**Summary of changes**
- We were doing too little reads on one message on height 3080, this solves the issue
  - This just marks the mutated node as unchanged, which would lead to unexpected behaviour, if it wasn't removed directly after (which is the case here)
- Changes gas charge names to match go casing, for a cleaner diff when interop

~This PR does not yet solve the issue of gas diff on mutation instead of removals~, because we do the mutation in place (as it should be if the go implementation was caching and iterating correctly) but they set the value while the amt is being iterated over (memory unsafe). I will find a way around this, but it might involve an inefficient Clone, because we can't exactly pull the element from the amt to set it back into the amt. I will open this PR when I have a functional workaround and tested adequately. 

This now solves the issue marked above. I've documented that it isn't a perfect solution, since the reads and writes are in different order, which could cause a consensus fault if an out of gas error is hit in the middle of this. We cannot match exactly without memory unsafe usage or an extensive implementation of this, so I think this is a good enough solution for now. I will try to make changes upstream/open a FIP to make their caching more efficient or fix this specific issue.

Currently resyncing to check to make sure no regressions

Go equivalent tests: https://github.com/austinabell/go-amt-ipld/commit/747191512a743a88e63456bbb644f3d1a77e4f7d

Closes 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3115,6 +3115,7 @@ dependencies = [
 name = "ipld_amt"
 version = "0.1.0"
 dependencies = [
+ "ahash 0.5.8",
  "criterion",
  "db",
  "forest_cid",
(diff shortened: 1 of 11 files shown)


timvermeulen (approved):
This looks reasonable!

dutterbutter (approved):

PR #850, opened 2020-11-13
Use upstream cid

vmx:
Instead of a custom implementation of cid, wrap the upstream version.

**Summary of changes**
- Using upstream rust-cid

I removed the "default CID" case from the CID serializer as the old default CID won't work with rust-cid as it was an invalid CID.

This PR is intentionally using the rust-cid master branch and not a release as I would like to see if things would work with that version of rust-cid or if any changes would be nice/needed before I do a release (as this release is a huge breaking one, so it's the chance to break more things).

There is a `new_from_cbor()` function which I made into a module's function instead of adding it to the `Cid` impl. The reason is that I wanted to keep the `Cid` a pure wrapper around the upstream `Cid` without additional methods. Though I'm happy to change that if requested.

Feedback in welcome.


ipld/cid/src/lib.rs
```diff
@@ -35,28 +29,63 @@ const MULTIBASE_IDENTITY: u8 = 0;
...
 pub mod json;
 
-/// Representation of a IPLD CID.
-#[derive(PartialEq, Eq, Clone, Copy)]
-pub struct Cid {
-    pub version: Version,
-    pub codec: Codec,
-    pub hash: Multihash,
+pub const DAG_CBOR: u64 = 0x71;
+pub const DAG_PB: u64 = 0x70;
```
austinabell, line 33: This doesn't have to be an exposed constant, it's not used for anything meaningful (just a test that could use any value)

ipld/src/lib.rs
```diff
@@ -101,7 +101,7 @@ pub enum Ipld {
     /// let cid: Cid = "QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n".parse().unwrap();
     /// let v1 = ipld!(Link(cid.clone()));
     /// let v2 = ipld!(cid);
-    /// assert_eq!(v1, v2);
+    /// //assert_eq!(v1, v2);
```
austinabell, line 104: does this fail now?

austinabell (commented):
Looks good! I'll just need to run interop tests and make sure no regressions as well as waiting for a few interop PRs to come in before merging this in though since this is a change that touches pretty much everything

ipld/cid/src/lib.rs
vmx: Changed tests to `DAG_CBOR`.

ipld/src/lib.rs
vmx: That was an oversight during development. I doesn't fail.

vmx:
@austinabell what about this plan:

 1. You run the integration tests. On success I'll cut a rust-cid release and update this PR. On failure I'll have a look into the failure (resp. helping you looking into).
 2. Once all PRs that should get merged before that one are in, you let me know and I'll then try to rebase this PR as quickly as possible, so that it doesn't block things.

austinabell:
> @austinabell what about this plan:
> 
> 1. You run the integration tests. On success I'll cut a rust-cid release and update this PR. On failure I'll have a look into the failure (resp. helping you looking into).
> 2. Once all PRs that should get merged before that one are in, you let me know and I'll then try to rebase this PR as quickly as possible, so that it doesn't block things.

Sounds great to me, the PRs I would like to get in before retesting are #847 #849 and #851 and I will let you know when those come in. I do believe your changes should be good as is, as I did some testing of it, I just want to check a couple things first and be completely sure if you don't mind

ipld/cid/src/lib.rs
```diff
@@ -105,121 +127,8 @@ impl<'de> de::Deserialize<'de> for Cid {
...
-            Version::V1 => self.to_string_v1(),
-        };
-        write!(f, "{}", encoded)
-    }
-}
-
-impl fmt::Debug for Cid {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "Cid(\"{}\")", self)
+        write!(f, "{}", self.0)
```
austinabell, line 132: This print is showing up as a debug print and it's really hard to read in certain logs, can you change this back to printing the actual `to_string`?

ipld/blockstore/Cargo.toml
```diff
@@ -6,6 +6,8 @@ edition = "2018"
 
 [dependencies]
 cid = { package = "forest_cid", path = "../cid" }
+#cid = { path = "/home/vmx/src/pl/rust/rust-cid", features = ["serde-codec"] }
+#cid = { path = "/home/vmx/src/pl/rust/rust-cid" }
```
austinabell, line 10:
```suggestion
```

austinabell (commented):
LGTM, should be good for release (these two comments are local changes)

austinabell (approved):

ec2 (approved):

austinabell:
Thanks :)

Issue #814, opened 2020-11-07
Fix Tipindex and implement Tipset caching


amerameen:
**Issue summary**
This issue will be updated by @austinabell 

**Other information and links**

- tipindex is called chainindex on Lotus

PR #851, opened 2020-11-14
Refactor ChainStore, tipset usage, cache loaded tipsets

austinabell:
**Summary of changes**
- Most of the usage affects things outside of the state transition, but this >4x our state transition speed, I can't exactly benchmark or compare other usages feasibly rn
- Removes TipIndex
  - It will be replaced in the future, but it was faulty and for some reason wrapped in two mutexes for access
  - I was originally going to just replace this, but it became apparent these changes were more impactful to do first (I will implement it on top of these changes #814)

Only half of #814, but made sense to split up and get this in asap

Closes 


blockchain/chain_sync/src/sync_worker.rs
```diff
@@ -390,8 +391,8 @@ where
 
         while let Some(result) = validations.next().await {
             match result {
-                Ok(b) => {
-                    self.chain_store().set_tipset_tracker(b.header()).await?;
+                Ok(_) => {
+                    // self.chain_store().set_tipset_tracker(b.header()).await?;
```
dutterbutter, line 395: Remove or add a TODO

dutterbutter (approved):

blockchain/chain/src/store/chain_store.rs
```diff
@@ -91,19 +95,22 @@ pub struct ChainStore<DB> {
     /// Tipset at the head of the best-known chain.
     heaviest: RwLock<Option<Arc<Tipset>>>,
 
+    ts_cache: RwLock<LruCache<TipsetKeys, Arc<Tipset>>>,
+
     /// tip_index tracks tipsets by epoch/parentset for use by expected consensus.
-    tip_index: RwLock<TipIndex>,
+    tip_index: TipIndex,
```
ec2, line 101: The PR description says we're removing TipIndex? Why u no remove tipindex? Is this what you mean by half of #814 ?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -303,12 +307,12 @@ where
             return Ok(None);
         }
         // TODO get tipset by height using cache instead of reloading tipsets
```
ec2, line 309: Is this comment still relevant? Seems like you implement that caching in tipset_from_keys now

blockchain/state_manager/src/chain_rand.rs
```diff
@@ -25,16 +26,18 @@ impl<DB> ChainRand<DB> {
...
 {
     fn get_chain_randomness(
         &self,
         pers: DomainSeparationTag,
         round: ChainEpoch,
         entropy: &[u8],
     ) -> Result<[u8; 32], Box<dyn Error>> {
-        self.cs
-            .get_chain_randomness(&self.blks, pers, round, entropy)
+        task::block_on(
```
ec2, line 37: We dont want to just make this an async_trait? Is it because this is really only used in the VM?

ec2 (approved):

blockchain/chain/src/store/chain_store.rs
austinabell: I just mean I gutted the TipIndex functionality and usages of it, I decided to keep for now because that will be built upon in next PR (why remove when it will just be re-added soon?)

blockchain/state_manager/src/chain_rand.rs
austinabell: yeah, the vm execution is sync, and I don't want to introduce async primitives when there will never be any benefit to it

blockchain/chain/src/store/chain_store.rs
austinabell: ah, nice catch

Issue #801, opened 2020-11-02
Blocksync: Implement dynamic request length 


amerameen:
**Issue summary**

Currently, for Blocksync, we're using a static request length (10) (the number of blocks requested at once). 

It would be better to have dynamic requests (i.e. make longer requests when we can successfully, or make shorter requests when they fail)

For example:  Lotus starts with 500 and if request fails, they shrink request size.

**Acceptance Criteria**
The Blocksync request lengths are dynamic and adjust (i.e. get bigger or smaller) based on what's feasible with current peers and network conditions.

austinabell:
To clarify, Lotus has a default window size of 500, and they send the request to 16 peers before changing window size (used to be 5, so this should also change on our side, assuming there is a good reason for this).

dutterbutter:
After discussing with @austinabell rather than having a dynamic window size as described above, I will be implementing a sync configuration that will sit within the `chainsyncer` where users can update the window size via command line as well as the number worker tasks related to syncing.

PR #852, opened 2020-11-14
Rename BlockSync -> ChainExchange and tweak parameters

austinabell:
**Summary of changes**
- I was tweaking parameters to test syncing and decided to just do this now while I'm testing other things
  - Yamux window is now large enough to handle 500 length request (for #801) and switches default to 200 for now as 10 is too low to test well
- Deduplicates RequestResponse codecs using generics


dutterbutter (approved):

node/forest_libp2p/src/service.rs
```diff
@@ -364,8 +366,8 @@ pub fn build_transport(local_key: Keypair) -> Boxed<(PeerId, StreamMuxerBox), Er
         .into_authentic(&local_key)
         .expect("Noise key generation failed");
     let mut yamux_config = yamux::Config::default();
-    yamux_config.set_max_buffer_size(1 << 20);
-    yamux_config.set_receive_window(1 << 20);
+    yamux_config.set_max_buffer_size(16 * 1024 * 1024);
+    yamux_config.set_receive_window(16 * 1024 * 1024);
```
ec2, line 370: Can you explain a little how these parameters are decided on? You said in the PR description the window size is now large enough, but 1 << 20 is larger than 16 * 1024 * 1024...

ec2 (approved):
Everything LGTM, just wanna understand the Yamux param changes a little more... The generic codecs is pretty nice <3 very cool

node/forest_libp2p/src/service.rs
austinabell: nah it isn't (16_777_216 > 1_048_576) I updated the window size to be equal to the default go one, just to make sure no issues. 1<<20 was an arbitrary number I didn't think we would hit lol

ec2: oh shoot, youre right.... *throws out my math degree*

austinabell: It looks smaller though, right?

PR #853, opened 2020-11-16
Switch param deserialization to the runtime to interop

austinabell:
**Summary of changes**
- Lotus incorrectly threw wrong exit code before network version 7, and this is necessary to match
- Groups syscalls in with runtime (no need to have separate)

gets past diff at 12_962

Closes 


types/src/version.rs
```diff
@@ -60,6 +60,10 @@ pub enum NetworkVersion {
     V4,
     /// tape (increases max prove commit size by 10x)
     V5,
+    /// kumquat (specs-actors v2.2.0)
```
ec2, line 63: lmao what a gr8 name... how in the world do they come up with these

ec2 (approved):

timvermeulen (approved):

PR #855, opened 2020-11-17
Implement chain index

### LINKED ISSUE #814: Fix Tipindex and implement Tipset caching

austinabell:
**Summary of changes**
- sparse cache of tipsets for quicker lookback tipset retrievals #814 
  - I decided to just match Lotus' naming for convenience
- Cleans up mutex usage to avoid holding locks longer than needed
- Removes mutex from state tree (overhead is not needed since it's only used in sync contexts)

Closes #814


blockchain/chain/src/store/index.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        if to == from.epoch() {
+            return Ok(from);
+        }
+
+        let mut ts = from;
+        loop {
+            let pts = self.load_tipset(ts.parents()).await?;
+
+            if to > pts.epoch() {
```
ec2, line 151: I'm guessing this to detect for empty epochs? Had to stare at that for a hot sec to see why this would ever happen lol

ec2 (approved):

blockchain/chain/src/store/chain_store.rs
```diff
@@ -515,6 +496,37 @@ where
...
+
+pub(crate) async fn tipset_from_keys<BS>(
+    cache: &TipsetCache,
+    store: &BS,
+    tsk: &TipsetKeys,
+) -> Result<Arc<Tipset>, Error>
+where
+    BS: BlockStore + Send + Sync + 'static,
+{
+    if let Some(ts) = cache.write().await.get(tsk) {
```
perfectmak, line 509: why not use `cache.read()` here instead of getting a write lock?

austinabell: Because it's a least recently used (LRU) cache, and it needs to update the position of the entry on read to function correctly, which requires mutable access. There is a method called `peek` on this cache, which only requires immutable access, but that defeats the purpose of this cache.

Yes this limits the concurrent reads, but the lock is kept for a very short amount of time and this means that frequently accessed keys don't get evicted from the cache early

blockchain/chain/src/store/index.rs
austinabell: yeah exactly haha

dutterbutter (approved):

PR #856, opened 2020-11-18
MessagePool Greedy Message Selection

### LINKED ISSUE #709: Message Selection
Miner calls mempool to grab a batch of pending messages. Add logic to decide which messages get selected for mining.

ec2:
**Summary of changes**
- Implements greedy message selection for mining 

Closes #709 

- Remaining to do for future PRs is hooking this up to RPC and writing the optimized selection algorithm for more dense blocks.


blockchain/message_pool/src/msgpool/mod.rs
```diff
@@ -1204,6 +1204,60 @@ where
...
+            right_chain.push(right.as_ref().clone());
+            let par = api.read().await.load_tipset(right.parents()).await?;
+            right = par;
+        }
+    }
+    for ts in left_chain {
+        let mut msgs: Vec<SignedMessage> = Vec::new();
+        for block in ts.blocks() {
+            let (_, mut smsgs) = api.read().await.messages_for_block(&block)?;
+            msgs.append(smsgs.as_mut());
```
timvermeulen, line 1239:
```suggestion
            let (_, smsgs) = api.read().await.messages_for_block(&block)?;
            msgs.extend(smsgs);
```

blockchain/message_pool/src/msgpool/mod.rs
```diff
@@ -1204,6 +1204,60 @@ where
...
+            let par = api.read().await.load_tipset(left.parents()).await?;
+            left = par;
+        } else {
+            right_chain.push(right.as_ref().clone());
+            let par = api.read().await.load_tipset(right.parents()).await?;
+            right = par;
+        }
+    }
+    for ts in left_chain {
+        let mut msgs: Vec<SignedMessage> = Vec::new();
```
timvermeulen, line 1236: You could reserve a capacity here

blockchain/message_pool/src/msgpool/mod.rs
```diff
@@ -1204,6 +1204,60 @@ where
...
+        for block in ts.blocks() {
+            let (_, mut smsgs) = api.read().await.messages_for_block(&block)?;
+            msgs.append(smsgs.as_mut());
+        }
+        for msg in msgs {
+            add(msg, rmsgs.borrow_mut());
```
timvermeulen, line 1242:
```suggestion
            add(msg, rmsgs);
```

blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -0,0 +1,592 @@
...
+{
+    /// Selects messages for including in a block.
+    pub async fn select_messages(
+        &mut self,
+        ts: Tipset,
+        _tq: f64,
+    ) -> Result<Vec<SignedMessage>, Error> {
+        let cur_ts = self.cur_tipset.read().await.clone();
+        // TODO: Implement a more optimal message selection to pack more msgs into a block
+        self.select_messages_greedy(cur_ts.as_ref(), ts).await
```
timvermeulen, line 24:
```suggestion
        self.select_messages_greedy(&cur_ts, ts).await
```

blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -0,0 +1,592 @@
...
+                    break 'tail_loop;
+                }
+
+                // does it fit in the block?
+                if chains[i].curr().gas_limit <= gas_limit {
+                    gas_limit -= chains[i].curr().gas_limit;
+                    result.append(&mut chains[i].curr_mut().msgs);
+                    continue;
+                }
+                last += i;
```
timvermeulen, line 118: Should this be `= i`?

blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -0,0 +1,592 @@
...
+        if !allow_negative_chains(cur_ts.epoch())
+            && !chains.is_empty()
+            && chains[0].curr().gas_perf < 0.0
+        {
+            return Ok(result);
+        }
+        // 3. Merge the head chains to produce the list of messages selected for inclusion, subject to
+        //    the block gas limit.
+        let mut last = chains.len();
+        for i in 0..chains.len() {
```
timvermeulen, line 71: `enumerate()` would probably clean this up a little bit

blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -0,0 +1,592 @@
...
+        // 1. Create a list of dependent message chains with maximal gas reward per limit consumed
+        let mut chains = Vec::new();
+        for (actor, mset) in pending.into_iter() {
+            chains.append(
+                &mut create_message_chains(&self.api, &actor, &mset, &base_fee, &ts).await?,
+            );
+        }
+
+        // 2. Sort the chains
+        chains.sort_by(|a, b| b.compare(&a));
```
timvermeulen, line 61: Just making sure that this should indeed be reversed?

blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -0,0 +1,592 @@
...
+            cur_ts.clone(),
+            ts.clone(),
+            &mut result,
+        )
+        .await?;
+
+        Ok(result)
+    }
+
+    async fn select_priority_messages(
```
timvermeulen, line 162: This seems to have a ton of overlap with `select_messages_greedy`, can we do better?

blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -0,0 +1,592 @@
...
+        //    the block gas limit.
+        let mut last = chains.len();
+        for i in 0..chains.len() {
+            if !allow_negative_chains(cur_ts.epoch()) && chains[i].curr().gas_perf < 0.0 {
+                break;
+            }
+
+            if chains[i].curr().gas_limit <= gas_limit {
+                gas_limit -= chains[i].curr().gas_limit;
+                result.append(&mut chains[i].curr().msgs.clone());
```
timvermeulen, line 78: This should probably be an `extend` call as well

blockchain/message_pool/src/msgpool/selection.rs
ec2: No I dont think so.. why do you say that?

blockchain/message_pool/src/msgpool/selection.rs
ec2: Yeah, i double checked.

blockchain/message_pool/src/msgpool/selection.rs
ec2: Thanks Mr. Tim "cargo clippy" Vermeulen

blockchain/message_pool/src/msgpool/selection.rs
ec2: Whats the diff with extend vs append in this context

blockchain/message_pool/src/msgpool/mod.rs
ec2: Could I really? I would need to know how many messages there are in each block... Dont think i can know that without going through each block first.

timvermeulen: Oh my bad, you indeed don't have that info

blockchain/message_pool/src/msgpool/selection.rs
austinabell: Well more importantly you're allocating a new vector just to drain it into the result, when all you need to do is extend by cloning the elements. I don't know what specifically gets optimized by llvm here, but idk if it can avoid the reallocation of the vector this way

blockchain/message_pool/src/msgpool/selection.rs
timvermeulen: It seems that way because `i` is an index into the `chains` vector, why is it added to `last`? If the `.enumerate()` call came after the `.skip(last)` call then `+= i` would make sense to me

blockchain/message_pool/src/msgpool/selection.rs
timvermeulen: `append` is useful when you want to append the elements of another vector that you want to empty and keep around (iirc its capacity is maintained as well), but that's a pretty specific use case where you have a buffer that you keep filling, or something like that. In general `extend` is the method to reach for

ec2: MMMMMMMM, yeah ok that makes a lot of sense. I will do that :D

blockchain/message_pool/src/msgpool/selection.rs
ec2: ohhhhhhhh wow thats a really really good catch dude... enumerate should come after skip

blockchain/message_pool/src/msgpool/selection.rs
ec2: I fine its a little easier (for me) to read like this weirdly, but yeah makes sense, i can remove some of the over lap. 👍

timvermeulen (approved):

blockchain/message_pool/src/msgpool/mod.rs
```diff
@@ -1203,6 +1203,60 @@ where
     Ok(())
 }
 
+/// Like head_change, except it doesnt change the state of the MessagePool.
+/// It simulates a head change call.
+pub async fn run_head_change<T>(
```
austinabell, line 1208: this doesn't need to be public, just used internally, no?

austinabell (approved):

PR #858, opened 2020-11-18
Add sled backend and cleanup RocksDb type

austinabell:
**Summary of changes**
- Alternate backend, https://github.com/spacejam/sled (ignored by default)
  - I just wanted to be able to profile the db better and see if this gave better performance (it hasn't for the specific use cases I've checked so far, main issue is bulk writes is exponentially slower)
- Cleans up RocksDb type
  - been bugging me for a while, I just needed a reason to fix it

Closes 


ec2 (approved):

node/db/src/sled.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+#![cfg(feature = "sled")]
```
timvermeulen, line 4: Isn't this implied by the `#[cfg(feature = "sled")]` on `pub mod sled;`?

timvermeulen (approved):

node/db/src/sled.rs
austinabell: ah, yeah, I had it this way before and forgot to remove (I'll remove this as well as I didn't remove from the rocksdb redundant one as well)

PR #874, opened 2020-11-30
A bunch of fixes for the RPC

ec2:
**Summary of changes**
- Refactor Gas API so that logic in those methods are reusable.
- Adds gas_estimate_message_gas API call that allows for estimation of gas parameters like the premium, fee cap, etc...

- Fixes some of the JSONRPC method names that have changed since initially implemented
- Fixes Websocket Auth handling. Before a new PersistantKeyStore was created upon every connection and saved to `~./forest`. This is not right, as we pass in the PersistantKeyStore via the RpcState.

- Adds message pool message selection to the RPC.
- Fixes some methods in message pool RPC.

- Fixes a bunch of methods in the state manager RPC.

Closes 

Making this PR before more random fixes come in and PR blows up with too many changes in too many different places. 
There is still some weird issue on when a message gets pushed to the message pool. I believe it has something to do with the VM execution though...

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -182,6 +182,7 @@ where
 
     /// Returns the currently tracked heaviest tipset
     pub async fn heaviest_tipset(&self) -> Option<Arc<Tipset>> {
+        // TODO: Figure out how to remove optional and return something everytime.
         self.heaviest.read().await.clone()
     }
 
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -180,7 +180,7 @@ where
         genesis(self.blockstore())
     }
 
-    /// Returns heaviest tipset from blockstore
+    /// Returns the currently tracked heaviest tipset
     pub async fn heaviest_tipset(&self) -> Option<Arc<Tipset>> {
         self.heaviest.read().await.clone()
     }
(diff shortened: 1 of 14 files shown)


types/src/deadlines/mod.rs
```diff
@@ -29,10 +30,15 @@ pub struct DeadlineInfo {
...
     // Protocol parameters (This is intentionally included in the JSON response for deadlines)
+    #[serde(rename = "WPoStPeriodDeadlines")]
     w_post_period_deadlines: u64,
+    #[serde(rename = "WPoStProvingPeriod")]
     w_post_proving_period: ChainEpoch,
+    #[serde(rename = "WPoStChallengeWindow")]
     w_post_challenge_window: ChainEpoch,
+    #[serde(rename = "WPoStChallengeLookback")]
     w_post_challenge_lookback: ChainEpoch,
+    #[serde(rename = "FaultDeclarationCutoff")]
```
austinabell, line 41:
```suggestion
```
Don't need?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -201,6 +201,9 @@ where
 
     /// Returns Tipset from key-value store from provided cids
     pub async fn tipset_from_keys(&self, tsk: &TipsetKeys) -> Result<Arc<Tipset>, Error> {
+        if tsk.cids().is_empty() {
+            return Ok(self.heaviest_tipset().await.unwrap());
```
austinabell, line 205: Would be nice if this was handled, although I get this should never happen (we should maybe looking into changing this from being an Optional value and always initializing with a heaviest, initializing with genesis ts so we don't have to handle these

austinabell (approved):

PR #875, opened 2020-11-30
Form tipsets

### LINKED ISSUE #758: Form a tipset with other blocks received at the same height
We receive blocks over pubsub. Ultimately, we extend the chain by adding a tipset (1+ blocks at the same height). We need to implement logic to group received blocks together into a tipset. 

Acceptance Criteria 
- A tipset is formed with blocks at the same height. Any blocks received at the same height (before the next epoch) are added to the tipset.

Notes
- Blocks will not all arrive at the same time, so that must be taken into account.
- This _task may not be required_ if informnewblock is modified such that it processes received blocks (at the same height) in parallel and groups them together in a tipset after processing them.

timvermeulen:
**Summary of changes**
- Add the `TipsetTracker` type that tracks blocks and can form tipsets from them
- Minor cleanup

Closes #758 

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -104,7 +104,7 @@ pub struct ChainStore<DB> {
     chain_index: ChainIndex<DB>,
 
     /// Tracks blocks for the purpose of forming tipsets.
-    tipset_tracker: RwLock<TipsetTracker<DB>>,
+    tipset_tracker: TipsetTracker<DB>,
 }
 
 impl<DB> ChainStore<DB>
@@ -116,7 +116,7 @@ where
         let cs = Self {
             publisher: RwLock::new(Publisher::new(SINK_CAP)),
             chain_index: ChainIndex::new(ts_cache.clone(), db.clone()),
-            tipset_tracker: RwLock::new(TipsetTracker::new(db.clone())),
+            tipset_tracker: TipsetTracker::new(db.clone()),
             db,
             ts_cache,
             heaviest: Default::default(),
@@ -157,7 +157,7 @@ where
     }
 
     pub async fn add_to_tipset_tracker(&self, header: &BlockHeader) {
-        self.tipset_tracker.write().await.add(header);
+        self.tipset_tracker.add(header).await;
     }
 
     /// Writes tipset block headers to data store and updates heaviest tipset
@@ -172,7 +172,7 @@ where
     }
 
     async fn expand_tipset(&self, header: BlockHeader) -> Result<Tipset, Error> {
-        self.tipset_tracker.read().await.expand(header)
+        self.tipset_tracker.expand(header).await
     }
 
     /// Loads heaviest tipset from datastore and sets as heaviest in chainstore
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -1,7 +1,7 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use super::{ChainIndex, Error};
+use super::{tipset_tracker::TipsetTracker, ChainIndex, Error};
 use actor::{miner, power::State as PowerState, STORAGE_POWER_ACTOR_ADDR};
 use address::Address;
 use async_std::sync::{channel, RwLock};
@@ -102,6 +102,9 @@ pub struct ChainStore<DB> {
 
     /// Used as a cache for tipset lookbacks.
     chain_index: ChainIndex<DB>,
+
+    /// Tracks blocks for the purpose of forming tipsets.
+    tipset_tracker: RwLock<TipsetTracker<DB>>,
 }
 
 impl<DB> ChainStore<DB>
@@ -113,6 +116,7 @@ where
         let cs = Self {
             publisher: RwLock::new(Publisher::new(SINK_CAP)),
             chain_index: ChainIndex::new(ts_cache.clone(), db.clone()),
+            tipset_tracker: RwLock::new(TipsetTracker::new(db.clone())),
             db,
             ts_cache,
             heaviest: Default::default(),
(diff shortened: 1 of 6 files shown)


blockchain/chain/src/store/chain_store.rs
```diff
@@ -102,6 +102,9 @@ pub struct ChainStore<DB> {
 
     /// Used as a cache for tipset lookbacks.
     chain_index: ChainIndex<DB>,
+
+    /// Tracks blocks for the purpose of forming tipsets.
+    tipset_tracker: RwLock<TipsetTracker<DB>>,
```
austinabell, line 107: I think it would be best to only wrap the type that interior mutability with the mutex, instead of wrapping the whole tracker type. Also, what are your thoughts on using something like https://github.com/xacrimon/dashmap for a threadsafe map, instead of using mutexes here?

blockchain/chain/src/store/tipset_tracker.rs
```diff
@@ -0,0 +1,83 @@
...
+                    log::warn!(
+                        "Have multiple blocks from miner {} at height {} in our tipset cache {}-{}",
+                        header.miner_address(),
+                        header.epoch(),
+                        header.cid(),
+                        cid
+                    );
+                }
+            }
+        }
```
austinabell, line 46: This seems like it might be worth to split up these checks into two loops, I would think it might be nice to short circuit the cids checks, so if any are the same, you don't have to load and deserialize a whole block header from storage. 

The second check also seems like a good candidate for a cache in the future (no need to do now) but may be worth it to cache miner addresses -> block cids, since entries are never removed? What are your thoughts?

blockchain/chain/src/store/tipset_tracker.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        entries.push(header.cid().clone());
+    }
+
+    /// Expands the given block header into the largest possible tipset by
+    /// combining it with known blocks at the same height with the same parents.
+    pub fn expand(&self, header: &BlockHeader) -> Result<Tipset, Error> {
+        let mut headers = vec![header.clone()];
```
austinabell, line 54: why not just have the function signature take in an owned header, to potentially avoid a clone if the header can be moved in the future?

blockchain/chain_sync/src/sync.rs
```diff
@@ -137,6 +137,137 @@ where
         self.worker_state.clone()
     }
 
+    async fn handle_network_event(
```
austinabell, line 140: out of curiosity, why was this moved into a function, just to clean up or was there a functional need?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -385,7 +397,7 @@ where
     }
 
     /// Constructs and returns a full tipset if messages from storage exists - non self version
-    pub fn fill_tipset(&self, ts: Arc<Tipset>) -> Result<FullTipset, Arc<Tipset>>
+    pub fn fill_tipset(&self, ts: &Tipset) -> Option<FullTipset>
```
austinabell, line 400: ah oops good call this is no longer needed, was an artifact from when Tipset was passed in and was not in wrapped in an Arc and this was a benefit

blockchain/chain_sync/src/sync.rs
timvermeulen: The direct reason I did this was that rust-analyzer isn't functioning properly inside the `select!` macro. But I think this probably makes the code more readable regardless

blockchain/chain/src/store/tipset_tracker.rs
timvermeulen: Oh yeah, I thought we would need to clone here regardless, but I can just extract the epoch first. Good catch

blockchain/chain/src/store/tipset_tracker.rs
timvermeulen: Heh, I started out caching miner addresses to avoid having to fetch the entire block header from the blockstore. In the end I reverted to this because I don't know how crucial it is for this tracker to minimize its memory footprint.

The `expand` method could potentially benefit from caching the parent tipset keys as well in order to avoid fetching block headers with different parents, but again, I'm unsure about the memory/speedup trade-offs.

austinabell: Yeah, let's leave it as is for now, but something for us to keep in mind

blockchain/chain/src/store/chain_store.rs
timvermeulen: To me this seemed like the more natural place for the lock because it's only in `ChainStore` that the need for concurrent writing arises, `TipsetTracker` in isolation doesn't really care about any of this. e.g. if at some point we hypothetically need to use this tracker without the need for concurrency, locks would be unnecessary. Is that fair?

dashmap is probably faster here, though I don't know if it also optimizes for many concurrent readers like `RwLock` does? I'll leave a note about possibly improving this, at the moment I'm unsure about pulling in a dependency without knowing how much of a speedup it gives us.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -152,14 +156,22 @@ where
...
+    pub async fn add_to_tipset_tracker(&self, header: &BlockHeader) {
+        self.tipset_tracker.write().await.add(header);
+    }
+
     /// Writes tipset block headers to data store and updates heaviest tipset
     pub async fn put_tipset(&self, ts: &Tipset) -> Result<(), Error> {
         persist_objects(self.blockstore(), ts.blocks())?;
-        // TODO determine if expanded tipset is required; see https://github.com/filecoin-project/lotus/blob/testnet/3/chain/store/store.go#L236
-        self.update_heaviest(ts).await?;
+        let expanded = self.expand_tipset(ts.min_ticket_block().clone()).await?;
```
austinabell, line 166: this seems pretty redundant, if the heaviest is based on the expanded, why you need to persist the tipset being passed in? Shouldn't this be able to use the other blocks in this passed in tipset as well when expanding?

blockchain/chain/src/store/chain_store.rs
austinabell: Well, I still don't see the reason why the mutex scope shouldn't be reduced, since the db never needs interior mutability. After reading the other parts a bit more I do see that the way it's setup forces a lock to be held for the whole function, when Cids are very cheap to clone. I just see the issue where each add, which happens after each block is validated, holds the lock while it performs io with the database during the function.

I don't think it's an issue with the current usage, but just thinking ahead to how this is setup. And yes we currently need this tracker to be used concurrently, as there will be multiple workers doing state transitions. That's why these types are setup to be threadsafe. We just have the default workers set to 1 for testing until everything is functional as there was some redundant overhead and made it harder to debug

austinabell (approved):

Issue #767, opened 2020-10-19
Test Block Processing from Storage Miner: Perform state transitions for block received from Storage Miner


amerameen:
When a block is received over RPC (from Storage Miner), the syncer must be notified so that the system can sync towards it.

Notes
- We don't have a mechanism to notify the syncer as of now (we don't have access to the syncer from RPC). What is the best way to do this? A channel or a network event? This should be all there is to do.
- A block from RPC or pubsub should be handled the same essentially (although there are peers to deal with on pubsub and track their heads, but that doesn't affect this issue)

Issue #766, opened 2020-10-19
Test Receiving Storage Miner Block: Forest can receive and validate block produced by Storage Miner


amerameen:
**Acceptance Criteria**
Block is produced by Storage Miner, sent to Forest via RPC and the block validations are done. (Out of Scope: Performing state transitions)

PR #882, opened 2020-12-09
Storage Miner Init Interop

### LINKED ISSUE #765: Testing Block Production: Miner can produce 1 block that gets sent to Forest over RPC and published
**Acceptance Criteria**
- [x] Storage Miner & Forest are able to be started up and connected to each other 
- [x] Storage Miner is able to produce a block
- [ ] Publish block over pubsub

The Storage Miner will send a block over RPC. See Lotus for more info: https://github.com/filecoin-project/lotus/blob/382dcf9a37f037294afd4da9aa5ecfaad21b8c44/miner/miner.go#L282

ec2:
**Summary of changes**
This PR allows us to initialize a Lotus storage miner.
- Fixes to the state manager: Fix wait for message and tipset executed message
- Adds the ability to receive and process JSONRPC messages concurrently  
- Adds SyncSubmitBlock method which accepts a block over JSONRPC which gets synced to by a sync worker
- Adds some missing State API methods with the big one being MinerCreateBlock which assembles a block for the miner 
- Fixes wallet signing where the string being signed isnt converted to base64 before signing
- Fixes some misc. JSONRPC method signatures

Notes:
Although this PR is IMO ready for review, there are some fixes that I would like the reviewer to give an opinion on:
Testing conditions for this PR are such that the Forest node is the only node on the network, thus it has no peers. Since it has no peers, the sync_workers sync method will be stuck in a loop. This results in newly mined blocks being submitted through SyncSubmitBlock to not be processed. Currently I have commented out this check for peer count. Wondering what yalls opinion is on this. 

Closes #765, #766, #767


blockchain/blocks/src/lib.rs
```diff
@@ -8,11 +8,11 @@ extern crate serde;
 extern crate lazy_static;
 
 mod block;
-mod election_proof;
+pub mod election_proof;
 mod errors;
 pub mod gossip_block;
 pub mod header;
-mod ticket;
+pub mod ticket;
```
austinabell, line 15: why re-export? It's exported below?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -135,14 +135,21 @@ where
...
     // subscribing returns a future sink that we can essentially iterate over using future streams
     pub async fn subscribe(&self) -> Subscriber<HeadChange> {
-        self.publisher.write().await.subscribe()
+        let sub = self.publisher.write().await.subscribe();
+        let ts = self.heaviest_tipset().await.unwrap();
+        self.publisher
+            .write()
+            .await
+            .publish(HeadChange::Current(ts))
+            .await;
```
austinabell, line 151: this seems wrong? Why would subscribing publish a headchange event?

blockchain/chain_sync/src/sync.rs
```diff
@@ -269,8 +269,13 @@ where
...
     /// Spawns a network handler and begins the syncing process.
-    pub async fn start(mut self, num_workers: usize) {
-        let (worker_tx, worker_rx) = channel(5);
+    pub async fn start(
+        mut self,
+        worker_tx: Sender<Arc<Tipset>>,
+        worker_rx: Receiver<Arc<Tipset>>,
+        num_workers: usize,
+    ) {
+        // let (worker_tx, worker_rx) = channel(20);
```
austinabell, line 278:
```suggestion
```

blockchain/state_manager/src/lib.rs
```diff
@@ -973,28 +969,43 @@ where
...
-            let larger_height_of_head = height_of_head >= back_tipset.epoch() + confidence;
-            if !should_revert && larger_height_of_head {
-                return Ok((Some(back_tipset), Some(back_receipt)));
+        })
+        .fuse();
+
+        let mut search_back_poll = task::spawn::<
+            _,
+            Result<(Option<Arc<Tipset>>, Option<MessageReceipt>), Error>,
+        >(async move {
```
austinabell, line 979: Optional: only the error type needs to be inferred

```suggestion
        let mut search_back_poll = task::spawn::<_, Result<_, Error>>(async move {
```

vm/actor/src/builtin/miner/policy.rs
```diff
@@ -58,7 +58,10 @@ pub const SEALED_CID_PREFIX: cid::Prefix = cid::Prefix {
 /// List of proof types which can be used when creating new miner actors
 pub fn check_supported_proof_types(proof: RegisteredSealProof) -> bool {
     use RegisteredSealProof::*;
-    matches!(proof, StackedDRG32GiBV1 | StackedDRG64GiBV1)
+    matches!(
+        proof,
+        StackedDRG2KiBV1 | StackedDRG32GiBV1 | StackedDRG64GiBV1
```
austinabell, line 63: did you just add 2kib for testing? It's not supposed to be supported

vm/actor/src/builtin/power/policy.rs
```diff
@@ -18,5 +18,5 @@ pub const MAX_MINER_PROVE_COMMITS_PER_EPOCH: u64 = 200;
 lazy_static! {
     // TODO: Might need to be able to override later on based on network params
     /// Minimum power of an individual miner to meet the threshold for leader election.
-    pub static ref CONSENSUS_MINER_MIN_POWER: StoragePower = StoragePower::from_i64(10 << 40).unwrap(); // placeholder
+    pub static ref CONSENSUS_MINER_MIN_POWER: StoragePower = StoragePower::from_i64(2048).unwrap(); // placeholder
```
austinabell, line 21: same with this, should be swapped out?

blockchain/chain_sync/src/sync_worker.rs
```diff
@@ -91,15 +91,15 @@ where
...
-        }
+        // loop {
+        //     let peer_count = self.network.peer_manager().len().await;
+        //     if peer_count < MIN_PEERS {
+        //         debug!("bootstrapping peers, have {}", peer_count);
+        //         task::sleep(Duration::from_secs(2)).await;
+        //     } else {
+        //         break;
+        //     }
+        // }
```
austinabell, line 102: Also, I'll give more thought on this soon, but probably fine for now to not have the bootstrap wait and to keep this out. I'll try to think of a better long term solution though

ec2: Cool. One solution i was thinking of was having a reserved sync_worker just for sync submit block. Then we can have an `if` around the loop to indicate is a block from a miner essentially.

austinabell: I don't think that's a great solution, and would turn out fairly clunky. I haven't moved around this bootstrap check since the sync worker has been moved out, is it possible to have this check around the `ChainSyncer` so that it doesn't cue initial sync tipsets until after bootstrap, but workers can still be triggered through RPC? 

I think the fact that this is checked here is just because this used to be all connected and not detached.

vm/actor/src/builtin/miner/policy.rs
ec2: ah, yeah good catch i forgot.

blockchain/blocks/src/lib.rs
ec2: if i dont pub mod it, i cant access the json submodule

blockchain/chain/src/store/chain_store.rs
ec2: When we sub to head changes, the first event is resolved right away and must be the Current.

blockchain/chain_sync/src/sync_worker.rs
ec2: Hmm yeah... Ill  leave this commented out on this PR and add a todo for now

blockchain/chain/src/store/chain_store.rs
austinabell: Why can't you put this in the channel when it's created, not when it's subscribed to? You'd get unnecessary head change events every time something subscribes

blockchain/chain_sync/src/sync_worker.rs
austinabell: Sounds good, but yeah thinking about it more, it for sure should just happen in the queueing of initial syncs in the `ChainSyncer`

blockchain/chain/src/store/chain_store.rs
ec2: Not sure what you mean? Subscribe can be called multiple times from all over the program. And anytime we subscribe, the Subscribers first event is a current event.

austinabell: Yeah, but what I'm saying is that _all_ subscribers would receive this event, is that intended?

austinabell: Wouldn't it make sense for the subscribe fn to also pass back the current ts, instead of publishing to broadcast?

ec2: Ohhhhhhh. Shoot, big brain oof moment. I believe youre completely right. Will make that change right now

austinabell (approved):

PR #860, opened 2020-11-19
Fix code coverage

austinabell:
**Summary of changes**
Changes introduced in this pull request:
- Switches machine to linux
- builds in release mode (with debug flag enabled `-g`)
- collects coverage from `target/release` because the tests are run in release mode with `make test-all`
- Fixes badge on README
 
I don't know how to hide the patch info on the build tag, but I'll try to figure that out before this comes in

CI run on fork:
https://app.circleci.com/pipelines/github/austinabell/forest/13/workflows/f4bacf83-37f1-4002-b8c6-bb921ad58753/jobs/122
(it's slow since medium machine so I'll wait for it to finish and upload successfully before merging)

**Reference issue to close (if applicable)**

Closes #846 

**Other information and links**

PR #883, opened 2020-12-10
Remove coverage from CI until fixed

austinabell:
**Summary of changes**
- Been wasting CI resources, but I want to make this change to fix a few things:
  - New CI wasn't checking license anymore
  - Fix gitmodules https over git
  - README coverage label broken (will fix in #860 or whatever PR fixes it when we have the time)
  - Use env variable for proofs cache

Closes 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,7 @@ commands:
             echo 'export CARGO_INCREMENTAL=0' >> $BASH_ENV
             echo 'export CI_TIMEOUT="timeout 40m"' >> $BASH_ENV
             echo 'export CC=gcc' >> $BASH_ENV
+            echo 'export FIL_PROOFS_PARAMETER_CACHE=$HOME/.proof-params' >> $BASH_ENV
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -53,6 +53,7 @@ commands:
             echo 'export CARGO_INCREMENTAL=0' >> $BASH_ENV
             echo 'export CI_TIMEOUT="timeout 40m"' >> $BASH_ENV
             echo 'export CC=gcc' >> $BASH_ENV
+            echo 'export FIL_PROOFS_PARAMETER_CACHE=$HOME/.proof-params' >> $BASH_ENV
```
ec2, line 56: Whats wrong with the default path?

austinabell: it's in the tmp directory, so this is more easily accessed and persisted (it's needed when running certain commands, but not immediately needed now in the CI). Just a good idea to set imo

ec2 (approved):

PR #884, opened 2020-12-11
Implement Net API Module + Some other RPC methods

ec2:
**Summary of changes**
- Adds Net JSONRPC API module. 
- Implements the NetAddrsListen method in the Net API.
- Implements StateMinerSectorAllocated API method
- Fixes ChainGetTipSet method

With this PR, we can start producing tipsets indefinitely. There are warnings on the storage miner though. I suspect it still has something to do with how we handle ChainNotify method calls... 

Closes 


node/forest_libp2p/src/service.rs
```diff
@@ -104,6 +105,19 @@ pub enum NetworkMessage {
         cid: Cid,
         response_channel: OneShotSender<()>,
     },
+    JSONRPCRequest {
+        method: NetRPCMethods,
+        response_channel: OneShotSender<NetRPCResponse>,
```
austinabell, line 110: why not have the channel type based on the request enum? That way all responses don't have to be wrapped in an enum and be matched on when you know what the return type should be?

austinabell (approved):

PR #887, opened 2020-12-12
Version Hamt, Amt, State tree

austinabell:
**Summary of changes**
- Moved utils used by hamt into own crate so it can be used in the interface for a cleaner API
- Version `Amt`, `Hamt`, and `StateTree`
  - I was trying to put off doing this for later, but I want to track test vector progress as I update actors

Verified correct because now 1204/1294 test vectors pass (failures due to not all actors being updated, 1255/1294 now pass with the multisig actor update)

Closes 

### DIFF
--- a/types/src/state.rs
+++ b/types/src/state.rs
@@ -1,7 +1,10 @@
 use cid::Cid;
+use encoding::repr::*;
+use encoding::tuple::*;
+use serde::{Deserialize, Serialize};
 
 /// Specifies the version of the state tree
-#[derive(Debug, PartialEq, Clone, Copy, PartialOrd)]
+#[derive(Debug, PartialEq, Clone, Copy, PartialOrd, Serialize_repr, Deserialize_repr)]
 #[repr(u64)]
 pub enum StateTreeVersion {
     /// Corresponds to actors < v2
@@ -10,6 +13,7 @@ pub enum StateTreeVersion {
     V1,
 }
 
+#[derive(Deserialize_tuple, Serialize_tuple)]
 pub struct StateRoot {
     /// State tree version
     pub version: StateTreeVersion,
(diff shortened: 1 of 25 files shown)


utils/hash_utils/src/key.rs
```diff
@@ -0,0 +1,64 @@
...
+/// with the go implementation.
+#[derive(Eq, PartialOrd, Clone, Debug, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct BytesKey(#[serde(with = "serde_bytes")] pub Vec<u8>);
+
+impl PartialEq for BytesKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
+    }
+}
```
timvermeulen, line 21: Can't this be derived?

timvermeulen (approved):

types/src/state.rs
```diff
@@ -0,0 +1,33 @@
...
+use cid::Cid;
+use encoding::repr::*;
+use encoding::tuple::*;
+use serde::{Deserialize, Serialize};
+
+/// Specifies the version of the state tree
+#[derive(Debug, PartialEq, Clone, Copy, PartialOrd, Serialize_repr, Deserialize_repr)]
+#[repr(u64)]
+pub enum StateTreeVersion {
+    /// Corresponds to actors < v2
```
ec2, line 13: Doesnt this actually correspond to the version of Hamt we use?

ec2 (approved):

utils/hash_utils/src/key.rs
austinabell: Well iirc it was because clippy warns against manually impl `Hash` and deriving `PartialEq`. Or maybe this isn't an issue now that it's our own hash trait. I'll try switching

types/src/state.rs
austinabell: It does! Because it uses the versioned hamt, and that version is tied to actor versions

ec2: Makes complete sense

utils/hash_utils/src/key.rs
austinabell: Ohhh wait sorry no it can't because it's own own trait. Reason for our own trait is because of issues like: https://github.com/rust-lang/rust/issues/27108.

PR #890, opened 2020-12-14
Update seal proof types and proofs api version

austinabell:
**Summary of changes**
- Update seal proof types to include new proof version
- Update proofs api (not necessary, but adds versioning and internal proofs updates)

Changes necessary for power actor v2 update

Closes 

### DIFF
--- a/types/src/sector/registered_proof.rs
+++ b/types/src/sector/registered_proof.rs
@@ -14,6 +14,12 @@ pub enum RegisteredSealProof {
     StackedDRG8MiBV1,
     StackedDRG32GiBV1,
     StackedDRG64GiBV1,
+
+    StackedDRG2KiBV1P1,
+    StackedDRG512MiBV1P1,
+    StackedDRG8MiBV1P1,
+    StackedDRG32GiBV1P1,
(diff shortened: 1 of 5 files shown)


types/src/sector/registered_proof.rs
```diff
@@ -14,6 +14,12 @@ pub enum RegisteredSealProof {
     StackedDRG8MiBV1,
     StackedDRG32GiBV1,
     StackedDRG64GiBV1,
+
+    StackedDRG2KiBV1P1,
```
ec2, line 18: Any clues as to what the P1 at the end stands for? Just curious

ec2 (approved):

types/src/sector/registered_proof.rs
austinabell: I made that on our end, they have V1_1 but I didn't want to add underscores in the variant since it's against rust's conventions so I added P1 as in the (p)roofs version 1. Open to suggestions or if you think this should just be `_1` to match exactly and ignore the lint

ec2: Its cool i was just curious 😀

PR #891, opened 2020-12-15
Handle pubsub blocks in parallel

### LINKED ISSUE #877: Parallel pubsub event handling in ChainSyncer
**Acceptance Criteria**
- Pubsub event handling done in parallel in the ChainSyncer 

**Other Info**
- Performance needs to be updated. ChainSyncer is polling network event channel. It's slow right now processing (i.e. receiving events) because it's using the same thread for bitswap requests.
- ChainSyncer would need to be refactored to be threadsafe
- Testing this is also difficult (i.e. awaits, bitswap requests)

timvermeulen:
The simplest way to make sure the pubsub message handler doesn't block the polling thread of the chain syncer. I think we should be able to reuse the same channel used by Hello events that are passed to `inform_new_head` (which is what this PR does).

Closes #877

### DIFF
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -273,9 +273,6 @@ where
         };
         let ts = FullTipset::new(vec![block]).unwrap();
         channel.send((source, ts)).await;
-        // if self.inform_new_head(source.clone(), &ts).await.is_err() {
-        //     warn!("failed to inform new head from peer {}", source);
-        // }
     }
 
     /// Spawns a network handler and begins the syncing process.
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -19,9 +19,9 @@ use cid::{Cid, Code::Blake2b256};
 use encoding::{Cbor, Error as EncodingError};
 use fil_types::verifier::ProofVerifier;
 use forest_libp2p::{hello::HelloRequest, NetworkEvent, NetworkMessage};
-use futures::future::try_join_all;
 use futures::select;
 use futures::stream::StreamExt;
+use futures::{future::try_join_all, try_join};
 use ipld_blockstore::BlockStore;
 use libp2p::core::PeerId;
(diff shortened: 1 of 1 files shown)


blockchain/chain_sync/src/sync.rs
```diff
@@ -268,6 +216,68 @@ where
...
+        let block = Block {
+            header: block.header,
+            bls_messages,
+            secp_messages,
+        };
+        let ts = FullTipset::new(vec![block]).unwrap();
+        channel.send((source, ts)).await;
+        // if self.inform_new_head(source.clone(), &ts).await.is_err() {
+        //     warn!("failed to inform new head from peer {}", source);
+        // }
```
austinabell, line 278: Probably worth removing comment and also documenting that this will be informing new head since `channel` isn't very descriptive

blockchain/chain_sync/src/sync.rs
```diff
@@ -268,6 +216,68 @@ where
...
+            .map(|m| network.bitswap_get::<SignedMessage>(m))
+            .collect();
+
+        let secp_messages = match try_join_all(secp_messages).await {
+            Ok(msgs) => msgs,
+            Err(e) => {
+                warn!("Failed to get SignedMessage: {}", e);
+                return;
+            }
+        };
```
austinabell, line 266: These are currently done in parallel, so the first secp message request is sent only after _all_ bls messages are received. If you would like to leave for another PR, can you add a TODO so it's less likely it gets missed in the future?

blockchain/chain_sync/src/sync.rs
timvermeulen: ...oops, this wasn't meant to still be there

blockchain/chain_sync/src/sync.rs
timvermeulen: Sounds very reasonable, let me know if this was the fix that you had in mind here

austinabell: lgtm!

austinabell (approved):

PR #892, opened 2020-12-15
Add syncing configuration options

### LINKED ISSUE #801: Blocksync: Implement dynamic request length

dutterbutter:
**Summary of changes**
- Adds `SyncConfig` to the `ChainSyncer` 
- Options include `worker_tasks` and `req_window`
- Adds sync options to `DaemonOpts` for command-line usage 
- Updates tests accordingly 

Closes #801 

Command-line usage:

```
forest --req-window 100 --worker-tasks 1
```

### DIFF
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -46,9 +46,13 @@ enum ChainSyncState {
     /// Following chain with blocks received over gossipsub.
     Follow,
 }
+
+/// Struct that defines syncing configuration options
 #[derive(Debug, Deserialize, Clone)]
 pub struct SyncConfig {
+    /// Request window length for tipsets during chain exchange
     pub req_window: i64,
+    /// Number of tasks spawned for sync workers
     pub worker_tasks: usize,
 }
 
@@ -102,6 +106,7 @@ pub struct ChainSyncer<DB, TBeacon, V, M> {
 
     mpool: Arc<MessagePool<M>>,
 
+    /// Syncing configurations
     sync_config: SyncConfig,
 }
 
@@ -288,6 +293,7 @@ where
 
     /// Spawns a network handler and begins the syncing process.
     pub async fn start(mut self, worker_tx: Sender<Arc<Tipset>>, worker_rx: Receiver<Arc<Tipset>>) {
+        // TODO benchmark (1 is temporary value to avoid overlap)
         for _ in 0..self.sync_config.worker_tasks {
             self.spawn_worker(worker_rx.clone()).await;
         }
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -27,6 +27,7 @@ use libp2p::core::PeerId;
 use log::{debug, info, trace, warn};
 use message::{SignedMessage, UnsignedMessage};
 use message_pool::{MessagePool, Provider};
+use serde::Deserialize;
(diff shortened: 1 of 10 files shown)


blockchain/chain_sync/src/sync/peer_test.rs
```diff
@@ -54,14 +54,15 @@ fn peer_manager_update() {
         local_sender,
         event_receiver,
         genesis_ts.clone(),
+        SyncConfig::default(),
```
austinabell, line 57: this should spawn 0 workers, or else the test has unnecessary overhead

blockchain/chain_sync/src/sync_worker.rs
```diff
@@ -1029,6 +1031,7 @@ mod tests {
                 genesis: genesis_ts,
                 bad_blocks: Default::default(),
                 verifier: Default::default(),
+                req_window: Default::default(),
```
austinabell, line 1034: this will be 0, seems like that would cause issues with the test?

blockchain/chain_sync/src/sync_worker/full_sync_test.rs
```diff
@@ -83,6 +83,7 @@ async fn space_race_full_sync() {
         genesis,
         bad_blocks: Default::default(),
         verifier: PhantomData::<FullVerifier>::default(),
+        req_window: Default::default(),
```
austinabell, line 86: same here

blockchain/chain_sync/src/sync.rs
```diff
@@ -46,6 +47,23 @@ enum ChainSyncState {
...
+    pub req_window: i64,
+    /// Number of tasks spawned for sync workers
+    pub worker_tasks: usize,
+}
+
+impl Default for SyncConfig {
+    fn default() -> Self {
+        Self {
+            req_window: 200,
+            worker_tasks: 1,
```
austinabell, line 63: can you please add the TODO that existed to revisit this value, as this certainly shouldn't stay once side effects fixed?

blockchain/chain_sync/src/sync.rs
```diff
@@ -55,15 +55,24 @@ pub struct SyncConfig {
...
 }
-
+impl SyncConfig {
+    pub fn new(req_window: i64, worker_tasks: usize) -> Self {
+        Self {
+            req_window,
+            worker_tasks
+        }
+    }
+}
```
austinabell, line 65: I don't think a constructor like this is necessary, but doesn't matter much

austinabell (approved):

blockchain/chain_sync/src/sync.rs
dutterbutter: I can remove if you want and just reference the struct directly in `peer_test.rs` I just defaulted to adding a constructor instead. Give a thumbs up if you want that changed.

austinabell: It's just config, it can change later if more are added (probably would be swapped for a builder pattern or something so don't worry about it)

PR #895, opened 2020-12-16
Add a cfg flag to build Forest in Devnet mode

ec2:
**Summary of changes**
On enabling the devnet flag,
- Allows for 2k sectors
- Sets the min consensus power to 2048

In draft because I need write access to the actor_v0 repo to add the flag on v0 actors. @austinabell 

Closes 

### DIFF
--- a/vm/actor/src/builtin/sector.rs
+++ b/vm/actor/src/builtin/sector.rs
@@ -4,6 +4,7 @@
 use fil_types::{RegisteredSealProof, StoragePower};
 
 /// Returns the minimum storage power required for each seal proof types.
+#[cfg(not(feature = "devnet"))]
 pub fn consensus_miner_min_power(p: RegisteredSealProof) -> Result<StoragePower, String> {
     use RegisteredSealProof::*;
     match p {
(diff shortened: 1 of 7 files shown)


austinabell:
Gave permissions. As for these changes, it would be nice if the deal minimum (`MINIMUM_VERIFIED_DEAL_SIZE`) and other parameters associated with the devnet are changed. Things like block delay and other parameters might be nice to also lower for testing, and we can consider changing the upgrade schedule too maybe?

ec2:
Cool, I can do the block delay and minimum verified deal size. The upgrade schedule is a little bit more tricky. I dont think it's worth adding that right now. When im doing storage miner testing, i actually have been disabling all the upgrades.

vm/actor/src/builtin/sector.rs
```diff
@@ -4,6 +4,7 @@
 use fil_types::{RegisteredSealProof, StoragePower};
 
 /// Returns the minimum storage power required for each seal proof types.
+#[cfg(not(feature = "devnet"))]
```
austinabell, line 7: Just a note, you probably want to just have the feature functionality within the scope of the function, to be sure that there will be no differences in the future, but this is totally fine for this case

austinabell (approved):

PR #896, opened 2020-12-16
Update runtime for network version upgrades

austinabell:
**Summary of changes**
- Update runtime receiver ID resolution based on NetworkVersion 3
- Implement call depth max from NetworkVersion 6

There was some test vectors failing in paych actor, and wanted to make sure that I didn't miss anything in the update. Passes 1267/1294 now

Closes 

### DIFF
--- a/vm/interpreter/src/default_runtime.rs
+++ b/vm/interpreter/src/default_runtime.rs
@@ -39,6 +39,9 @@ use vm::{
     EMPTY_ARR_CID, METHOD_SEND,
 };
 
+/// Max runtime call depth
+const MAX_CALL_DEPTH: u64 = 4096;
+
 // This is just used for gas tracing, intentionally 0 and could be removed.
 const ACTOR_EXEC_GAS: GasCharge = GasCharge {
     name: "OnActorExec",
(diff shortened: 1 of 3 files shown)


vm/interpreter/src/default_runtime.rs
```diff
@@ -55,9 +58,15 @@ struct VMMsg {
...
     fn caller(&self) -> &Address {
+        assert!(
+            matches!(self.caller.protocol(), Protocol::ID),
+            "runtime message caller was not resolved to ID address"
+        );
         &self.caller
     }
     fn receiver(&self) -> &Address {
+        // * Can't assert that receiver is an ID address here because it was not being done
+        // * pre NetworkVersion3. Can maybe add in assertion later
```
ec2, line 69: If this is part of the v2 actors and syncing with mainnet, shouldn't you just get that in now? Or is it a big effort that should be done in another PR?

ec2 (approved):

vm/interpreter/src/default_runtime.rs
austinabell: No like we can't do it at all, because it breaks in <v3. There is no possible way to do this without not supporting those versions

austinabell: it's one assertion

PR #899, opened 2020-12-17
Storage Miner Pledging

ec2:
**Summary of changes**
- Added RPC methods pertaining to miner pledging
- Fixed some borked RPC methods.

We can now pledge a commited capacity sector and it will go through Precommit to Commit then Proving phase.

Closes 


austinabell (approved):

node/rpc/src/state_api.rs
```diff
@@ -802,16 +799,144 @@ pub(crate) async fn state_miner_sector_allocated<
...
+            let circ_supply = data
+                .state_manager
+                .get_circulating_supply(ts.epoch(), &state)?;
+            reward::State::load(data.state_manager.blockstore(), &reward_actor)?
+                .initial_pledge_for_power(
+                    &sector_weight,
+                    &total_locked,
+                    power_smoothed.into(),
+                    &circ_supply,
+                )
```
austinabell, line 922: why does this need to be split into two enums? can't you expose qa power smoothed and total pledge collateral and avoid doing this?

node/rpc/src/state_api.rs
```diff
@@ -802,16 +799,144 @@ pub(crate) async fn state_miner_sector_allocated<
...
+                .pre_commit_deposit_for_power(power_smoothed.into(), &sector_weight)
+        }
+        power::State::V2(s) => {
+            let power_smoothed = s.this_epoch_qa_power_smoothed;
+            let reward_actor = state
+                .get_actor(reward::ADDRESS)?
+                .ok_or("couldnt load market actor")?;
+            reward::State::load(data.state_manager.blockstore(), &reward_actor)?
+                .pre_commit_deposit_for_power(power_smoothed.into(), &sector_weight)
+        }
```
austinabell, line 850: same here, would be better if this was done through interface, only this epoch qa power needs to be exposed

austinabell (changes requested):
actually these should be moved to the interface

node/rpc/src/state_api.rs
ec2: Done. Testing out right now

austinabell (approved):

PR #914, opened 2021-01-06
Wrap ChainSyncer::state in an Arc<Mutex<_>> and set it to Follow accordingly

### LINKED ISSUE #876: Transition from 'Initial Sync' to 'Chain Follow'
**Acceptance Criteria**
- Once 'initial sync' is done, switch the ChainSync state to ChainFollow, which will allow pubsub messages to be processed from the network event channel

**Questions**
- How do we know when intial sync is done? May be worthwhile looking into how Lotus does it.
- Do we need to do another 'Initial Sync' after the first one is done (since it takes a while) before switching 'ChainFollow'?
- The whole time we're doing 'Initial Sync' we're getting gossipsub blocks, what is happening to these? What do we do with them once we switch to 'ChainFollow'?
- How to test this properly? May be difficult 
- Signalling to the ChainSyncer that 'Initial Sync' is done is not trivial.

timvermeulen:
**Summary of changes**
- Wraps `ChainSyncer::state` in an `Arc<Mutex<_>>`
- Passes this mutex to each sync worker, which will set the state to `::Follow` when `SyncWorker::sync` completes succesfully

Closes #876

### DIFF
--- a/blockchain/chain_sync/src/sync.rs
+++ b/blockchain/chain_sync/src/sync.rs
@@ -10,7 +10,7 @@ use super::sync_state::SyncState;
 use super::sync_worker::SyncWorker;
 use super::{Error, SyncNetworkContext};
 use amt::Amt;
-use async_std::sync::{channel, Receiver, RwLock, Sender};
+use async_std::sync::{channel, Mutex, Receiver, RwLock, Sender};
 use async_std::task::{self, JoinHandle};
 use beacon::Beacon;
 use blocks::{Block, FullTipset, GossipBlock, Tipset, TipsetKeys, TxMeta};
@@ -38,7 +38,7 @@ use std::sync::Arc;
 type WorkerState = Arc<RwLock<Vec<Arc<RwLock<SyncState>>>>>;
 
 #[derive(Debug, PartialEq)]
-enum ChainSyncState {
+pub enum ChainSyncState {
     /// Bootstrapping peers before starting sync.
     Bootstrap,
     /// Syncing chain with ChainExchange protocol.
@@ -78,7 +78,7 @@ impl Default for SyncConfig {
 /// messages to be able to do the initial sync.
 pub struct ChainSyncer<DB, TBeacon, V, M> {
     /// State of general `ChainSync` protocol.
-    state: ChainSyncState,
+    state: Arc<Mutex<ChainSyncState>>,
 
     /// Syncing state of chain sync workers.
     worker_state: WorkerState,
@@ -142,7 +142,7 @@ where
         );
 
         Ok(Self {
-            state: ChainSyncState::Bootstrap,
+            state: Arc::new(Mutex::new(ChainSyncState::Bootstrap)),
             worker_state: Default::default(),
             beacon,
             network,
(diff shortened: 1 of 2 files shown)


blockchain/chain_sync/src/sync.rs
```diff
@@ -424,10 +424,10 @@ where
             .await;
 
         // Only update target on initial sync
-        if self.state == ChainSyncState::Bootstrap {
+        if *self.state.lock().await == ChainSyncState::Bootstrap {
             if let Some(best_target) = self.select_sync_target().await {
                 self.schedule_tipset(best_target).await;
-                self.state = ChainSyncState::Initial;
+                *self.state.lock().await = ChainSyncState::Initial;
```
austinabell, line 430: I believe this keeps the lock and will deadlock here, has this been tested?

blockchain/chain_sync/src/sync.rs
```diff
@@ -78,7 +78,7 @@ impl Default for SyncConfig {
 /// messages to be able to do the initial sync.
 pub struct ChainSyncer<DB, TBeacon, V, M> {
     /// State of general `ChainSync` protocol.
-    state: ChainSyncState,
+    state: Arc<Mutex<ChainSyncState>>,
```
austinabell, line 81: is mutex over rwlock required?

austinabell (changes requested):

blockchain/chain_sync/src/sync.rs
timvermeulen: It is not, but does `RwLock` have any benefits if we're not expecting concurrent reads?

blockchain/chain_sync/src/sync.rs
timvermeulen: Not tested, but I'm quite certain that `if *self.state.lock().await` does not keep the guard around since it's not bound to a variable, if that's what you're referring to. ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6f082ff7b18953ba225181e2d96df9f9) that uses std's Mutex)

blockchain/chain_sync/src/sync.rs
austinabell: Yeah I just was asking because there doesn't seem to be a reason to have the additional constraint (just thinking for future proofing)

It's definitely fine if this stays as is, this will probably be replaced in the future anyway

blockchain/chain_sync/src/sync.rs
austinabell: I've had issues with async-std mutexes holding the lock in this exact case in the past, not sure if it's changed or only applies to `RwLock` but can double check

austinabell: nvm, there was a specific case where it deadlocked similar to this, and moving the read outside of the `if` removed the deadlock, I just tested and this works

austinabell (approved):

blockchain/chain_sync/src/sync.rs
timvermeulen: Oh yeah, I can totally see that happening. Thanks for pointing it out either way, I could have easily messed this up

PR #922, opened 2021-01-09
Update libp2p, async-std, and other deps

austinabell:
**Summary of changes**
- Migrates to new error prone async-std channels (0.1.8)
- Libp2p update to 0.28
  - Updated bitswap dep and reinitialized request-response (I kept their oneshot channels)

Closes 


austinabell:
Having issues with libp2p connections timing out with this new version, going to let this sit for a bit to debug. Could also just remove that version bump from this PR and revisit later, undecided.

ec2:
Interesting... When does it actually time out? When its trying to connect? Or when you try to send some RPC(BlockSync) req?

austinabell:
> Interesting... When does it actually time out? When its trying to connect? Or when you try to send some RPC(BlockSync) req?

When dialing trying to send rpc requests

austinabell:
I think the issue is actually because https://github.com/libp2p/rust-libp2p/pull/1867 dropped support for one-way requests (we are not responding to Hello currently) and possibly this is blocking us from being able to send chain exchange requests to that same peer.

I'm going to investigate a bit more but I'm just leaning toward updating our hello response (was going to be done later anyway) and following this pattern, so we are available for future updates of the upstream request-response

austinabell:
Going to just leave it as 0.28 update for now and open a draft PR for 0.34 changes (it's currently broken running against the go implementation)

node/forest_libp2p/src/service.rs
```diff
@@ -368,32 +348,41 @@ where
...
+            .into_authentic(&local_key)
+            .expect("Noise key generation failed");
+
+        noise::NoiseConfig::xx(dh_keys).into_authenticated()
+    };
+
+    let mplex_config = {
+        let mut mplex_config = mplex::MplexConfig::new();
+        mplex_config.max_buffer_len(usize::MAX);
+        // mplex_config.set_max_buffer_size(usize::MAX);
```
ec2, line 373: Did you forget to remove this comment?

node/forest_libp2p/src/behaviour.rs
```diff
@@ -55,6 +59,18 @@ pub struct ForestBehaviour {
     events: Vec<ForestBehaviourEvent>,
     #[behaviour(ignore)]
     peers: HashSet<PeerId>,
+    /// Keeps track of Chain exchange requests to responses
+    #[behaviour(ignore)]
+    cx_request_table: HashMap<RequestId, OneShotSender<ChainExchangeResponse>>,
+    /// Boxed futures of responses for the
```
ec2, line 65: for the.....

node/forest_libp2p/src/service.rs
austinabell: ah ya, it was what the type changed in future libp2p version, I commented for easier switching, but #928 exists

node/forest_libp2p/src/behaviour.rs
austinabell: haha

ec2 (approved):

PR #924, opened 2021-01-11
Fix ChainNotify RPC

ec2:
**Summary of changes**
- Fixes Chain Notify subscription IDs, storage miner no longer throwing warnings. 

Closes 

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -967,18 +967,14 @@ where
 
 pub async fn sub_head_changes(
     mut subscribed_head_change: Subscriber<HeadChange>,
-    heaviest_tipset: &Option<Arc<Tipset>>,
+    heaviest_tipset: &Arc<Tipset>,
     current_index: usize,
     events_pubsub: Arc<RwLock<Publisher<EventsPayload>>>,
 ) -> Result<usize, Error> {
-    let head = heaviest_tipset
-        .as_ref()
-        .ok_or_else(|| Error::Other("Could not get heaviest tipset".to_string()))?;
-
     (*events_pubsub.write().await)
         .publish(EventsPayload::SubHeadChanges(IndexToHeadChange(
             current_index,
-            HeadChange::Current(head.clone()),
+            HeadChange::Current(heaviest_tipset.clone()),
         )))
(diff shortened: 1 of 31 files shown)


blockchain/chain/src/store/chain_store.rs
```diff
@@ -992,7 +993,9 @@ pub async fn sub_head_changes(
                 .await;
         }
     });
-    let cancel_sender = events_pubsub.write().await.subscribe();
+    let mut x = events_pubsub.write().await;
+    let cancel_sender = x.subscribe();
+    drop(x);
```
austinabell, line 998: Can't you just do this in one line? You don't need to keep the lock for anything?

node/rpc/src/chain_api.rs
```diff
@@ -69,10 +69,10 @@ where
     KS: KeyStore + Send + Sync + 'static,
     B: Beacon + Send + Sync + 'static,
 {
-    let (data_subscribe, _) = data.state_manager.chain_store().subscribe().await;
+    let (data_subscribe, ts) = data.state_manager.chain_store().subscribe().await;
     let index = chain::sub_head_changes(
         data_subscribe,
-        &data.state_manager.chain_store().heaviest_tipset().await,
+        &Some(ts),
```
austinabell, line 75: why is this an option if this is the only place it's used? Should subscribe just return the option then?

austinabell (approved):

blockchain/chain/src/store/chain_store.rs
ec2: Ah yeah, i wasnt sure at one point if the lock was being held by calling subscribe, but yeah I can remove this.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -967,20 +967,17 @@ where
 
 pub async fn sub_head_changes(
     mut subscribed_head_change: Subscriber<HeadChange>,
-    heaviest_tipset: &Option<Arc<Tipset>>,
+    heaviest_tipset: &Arc<Tipset>,
```
austinabell, line 970: also, does it need to be a reference if it can just be moved? it's cloned during the subscribe?

PR #932, opened 2021-01-15
CBOR Stream Read in LibP2P RPC

ec2:
**Summary of changes**
- Before, we read libp2p RPC streams until we see an EOF. This is wrong because we can tell when cbor objects end and lotus doesnt send an EOF after writing to the stream. This PR introduces streams and decodes. 
- Must come in after #931 

Closes 

### DIFF
--- a/node/forest_libp2p/Cargo.toml
+++ b/node/forest_libp2p/Cargo.toml
@@ -44,7 +44,6 @@ tiny-cid = "0.2.0"
 ipld_blockstore = "0.1"
 async-trait = "0.1"
 lazy_static = "1.4"
-
 futures_cbor_codec = { git = "https://github.com/najamelan/futures_cbor_codec", rev = "de08e31530513f993cbf5efef7311ac5194b2256" }
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


node/forest_libp2p/Cargo.toml
```diff
@@ -45,6 +45,8 @@ ipld_blockstore = "0.1"
 async-trait = "0.1"
 lazy_static = "1.4"
 
+futures_cbor_codec = { git = "https://github.com/najamelan/futures_cbor_codec", rev = "de08e31530513f993cbf5efef7311ac5194b2256" }
```
austinabell, line 48: Did the most recent release, `0.2` not work?

austinabell (approved):
ut but looks correct. Also, what was the reason for not switching the write to this as well? Not a blocking issue since the functionality will be the same, but would avoid going through the vec buffer first. Seems like it's worth to do now while we test this specifically?

node/forest_libp2p/Cargo.toml
austinabell: Is it because of the version bump from the commit after the release?

ec2: Yeah. The 0.2 release is using futures-codec 0.3 whereas I wanted to use 0.4.

PR #944, opened 2021-01-21
Swap BLS backend for blst

austinabell:
**Summary of changes**
- something something 35x faster
- Downgrades certain versions to compatible (newer disallows zero key/addr)

In draft until it gets released (don't want two versions of blst)

Closes 


austinabell:
oops wrong branch

PR #946, opened 2021-01-22
Update consensus min power

austinabell:
**Summary of changes**

So in v2 they decided to have a hashmap of all proofs to their corresponding minimum consensus power, makes sense. Turns out the function they used to override the single value in v0 was used to iterate over every hashmap value and set it to a static `10 << 40`. Why did they change it to this variably set map if they were just going to overwrite all values to one static? Why would they not alter the setting of consensus minimum to match this change? I don't have the answers

Closes 


ec2 (approved):
huh... this is quite weird

PR #975, opened 2021-01-30
Refactor discovery, improve Hello handling/peer management

### LINKED ISSUE #966: Handle Libp2p Disconnects
**Issue summary**

Peer handling is a little buggy right now. The issue it appears is that we do not remove peers that have shutdown. This is an issue for a Forest network. Suppose there is a network with 2 Forest nodes. When Node A shuts down and is later spun back up, since Node B didn't remove Node A from its peer list, it will not send a Hello message and thus Node A will not start syncing. Currently while manually testing, I delete the Libp2p key before starting Node A back up again to regenerate a new PeerId. 

austinabell:
**Summary of changes**
- Refactors discovery into it's own behaviour and improves interactions
  - Sends random kad queries when under the discovery limit (default 50) to ensure quick bootstrapping and maintaining peers if ever drops below (but opposite issue is more common, of having too many peers)
  - Disable mdns by default. Better to not have the overhead by default, Lotus removed it altogether
  - Improve peer connect and disconnect events

- Handle hello responses as well as correctly logging failure and marking peers as bad on libp2p errors such as unsupported protocols or dial failures
  - Change peer manager api to allow for removing but not marking bad (on peer disconnect, to allow them to be reconnected)
  - On request success or failure, insert a peer in the set if doesn't exist. Before it was relying on the fact that an empty peer was added on inbound hello request, but this handled the case when we send the first request poorly

How the hello responses are handled is when a hello request is sent, a boxed future is returned. This future is added to a FuturesUnordered and polled in the chain syncer.

Closes #966 


blockchain/chain_sync/src/network_context.rs
```diff
@@ -12,17 +12,32 @@ use forest_libp2p::{
...
+
+/// Future of the response from sending a hello request. This does not need to be immediately polled
+/// because the response does not need to be handled synchronously.
+pub(crate) type HelloResponseFuture = Pin<
+    Box<
+        dyn Future<
+                Output = (
+                    PeerId,
+                    SystemTime,
+                    Option<Result<HelloResponse, RequestResponseError>>,
```
ec2, line 35: I normally really like rustfmt... This is one of the rare cases lol

ec2 (approved):

PR #977, opened 2021-02-02
Include git hash and crate version

### LINKED ISSUE #934: Include version and git hash on libp2p identity agent version
**Issue summary**

Currently hard coded as 0.1.0 here https://github.com/ChainSafe/forest/blob/a2ac9552cd21e9911ace79204b229aea6d99de03/node/forest_libp2p/src/behaviour.rs#L462

But would be much better to be formatted as something like: forest-{cargo-version}-{shortened-git-hash}

This should be done at build time, with something like `lazy_static` and we are pulling the git tag https://github.com/ChainSafe/forest/blob/3c536d0d0e43d41b87ade7208479f64e5844f27c/types/src/build_version/mod.rs#L27. I do not think this way of pulling the git tag is ideal, but in either case, it should be consolidated and put in a shared crate

nannick:
**Summary of changes**
- Uses git-version crate to get git commit
-  Uses rust env! to get package version

Closes 
#934


node/forest_libp2p/src/behaviour.rs
```diff
@@ -428,9 +432,7 @@ impl ForestBehaviour {
             ping: Ping::default(),
             identify: Identify::new(
                 "ipfs/0.1.0".into(),
-                // TODO update to include actual version
-                // https://github.com/ChainSafe/forest/issues/934
-                format!("forest-{}", "0.1.0"),
+                format!("forest-{}", *VERSION),
```
austinabell, line 435: nice! Can you add that same commit hash at the end of this as well? Something like

```suggestion
                format!("forest-{}-{}", *VERSION, *CURRENT_COMMIT),
```

Sorry if the issue was unclear. If not I can just do it in a following PR.

austinabell (approved):

node/forest_libp2p/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 utils = { path = "../utils" }
 networks = { path = "../../types/networks" }
+fil_types = { path = "../../types" }
```
austinabell, line 10: sorry for the nit, but can you avoid bringing this dependency into the libp2p crate? Since the type is as simple as just `git_version!()`, maybe just have that in place of `CURRENT_COMMIT`

austinabell, line 10:
```suggestion
```

austinabell:
Hey @nannick, this PR looks great and I'd like to pull it in. The branch just needs to be updated and that last change (https://github.com/ChainSafe/forest/pull/977#discussion_r569568138) made. If you give me privileges to modify your branch I can do this for you, if you'd like :)

node/forest_libp2p/Cargo.toml
nannick: > Hey @nannick, this PR looks great and I'd like to pull it in. The branch just needs to be updated and that last change ([#977 (comment)](https://github.com/ChainSafe/forest/pull/977#discussion_r569568138)) made. If you give me privileges to modify your branch I can do this for you, if you'd like :)

Sorry for the late fix on such a small PR

austinabell (approved):

austinabell:
Thank you so much for this! This will be nice to have

PR #980, opened 2021-02-02
Fix fork handling

ec2:
**Summary of changes**
- i started a devnet with 2 miners. Then, at block 86, i disabled mdns, so they were not connected no more. Continued mining on block 86 on both. Killed one of the miners at block 100. Then mined a couple more blocks on the other one. Then reconnected the two nodes and they both agreed on the heaviest chain.

Closes 


blockchain/chain_sync/src/sync_worker.rs
```diff
@@ -283,6 +283,8 @@ where
     /// fork detected, collect tipsets to be included in return_set sync_headers_reverse
     async fn sync_fork(&self, head: &Tipset, to: &Tipset) -> Result<Vec<Arc<Tipset>>, Error> {
         // TODO move to shared parameter (from actors crate most likely)
+        // TODO the threshold should really be 900. Not entirely sure if we can handle a 900
```
austinabell, line 286: The todo a few lines below is the same, maybe just delete that old one or combine somehow to remove confusion?

austinabell (approved):

PR #982, opened 2021-02-03
v3 Hamt update

### LINKED ISSUE #959: v3 Amt and Hamt update
**Task summary**

Most of this is already done (because we fixed the inefficient caching of each under a feature) but we still need to update a few details around the API to match the actors v3

**Specification reference**

- 

austinabell:
**Summary of changes**
- Updates pointer serialization format, all caching updates had been done previously (`!go-interop` flag)

Tested against https://github.com/austinabell/go-hamt-ipld/tree/rust/testing

Will wait on #981 and add another method to the API before opening PR for review

Closes #959 

### DIFF
--- a/ipld/hamt/src/node.rs
+++ b/ipld/hamt/src/node.rs
@@ -292,9 +292,18 @@ where
             Pointer::Values(vals) => {
                 // Update, if the key already exists.
                 if let Some(i) = vals.iter().position(|p| p.key() == &key) {
-                    if overwrite && vals[i].value() != &value {
-                        // Can overwrite, replace and return true.
-                        return Ok((Some(std::mem::replace(&mut vals[i].1, value)), true));
+                    if overwrite {
+                        // If value changed, the parent nodes need to be marked as dirty.
+                        // ! The assumption here is that `PartialEq` is implemented correctly,
+                        // ! and that if that is true, the serialized bytes are equal.
+                        // ! To be absolutely sure, can serialize each value and compare or
+                        // ! refactor the Hamt to not be type safe and serialize on entry and
+                        // ! exit. These both come at costs, and this isn't a concern.
+                        let value_changed = vals[i].value() != &value;
+                        return Ok((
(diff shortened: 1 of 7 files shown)


ec2 (approved):

ipld/hamt/src/node.rs
```diff
@@ -228,48 +241,79 @@ where
...
                 if let Some(i) = vals.iter().position(|p| p.key() == &key) {
-                    vals[i].1 = value;
-                    return Ok(());
+                    if overwrite && vals[i].value() != &value {
+                        // Can overwrite, replace and return true.
+                        return Ok((Some(std::mem::replace(&mut vals[i].1, value)), true));
+                    } else {
+                        // Can't overwrite, return None and false that the Node was not modified.
+                        return Ok((None, false));
+                    }
```
cryptoquick, line 301: @austinabell Isn't this where you were going to address that additional case you mentioned on our review call? Also, can you describe why that came to mind, so I can understand better?

austinabell: So the point I was thinking through is not relevant to consensus, because the go implementation does not return the previous value. However, thinking through it a third time, I think that in this case a few lines above should be modified a bit. 

The case is on set (overwrite = true) and the values are equal, should the value being passed in be returned back. Although it seems weird internally to do this, it would match a `HashMap` functionality by returning the value that existed. The difference would be that we are not swapping the values, and assuming PartialEq is implemented correctly. I am going to switch it to returning the value that existed, to match a `HashMap` implementation, but open to just returning the value passed in. This operation should be super quick in every case so I think it's the best decision

austinabell: Also, just to be clear, the only case which this would have unintended behaviour is when `PartialEq` doesn't match serialized bytes equality `AND` this was the only value changed, the parent roots won't be updated. Definitely nothing to worry about, but worth keeping in mind that `PartialEq` must match bytes equality to interop in every case with go.

ec2 (approved):

PR #983, opened 2021-02-03
Nightly build/audit workaround and nightly linting fixes

austinabell:
**Summary of changes**
- logs are broken without a semicolon after it, so that is the workaround
  - https://github.com/rust-lang/rust/issues/81531
- Fixes other nightly clippy warnings while I was checking
- Ignores fil dependency audit failure. It can't be triggered and is out of our control

Closes 


.circleci/config.yml
```diff
@@ -249,7 +249,8 @@ jobs:
           command: cargo install cargo-audit
       - run:
           name: Check for known security issues in dependencies
-          command: cargo audit
+          # TODO remove ignore when https://github.com/filecoin-project/rust-fil-proofs/issues/1403 resolved
```
ec2, line 252: Can you open an issue for this real quick if were gunna ignore? Def gunna forget this 😂

ec2 (approved):

PR #987, opened 2021-02-05
Framework for State Migrations

ec2:
**Summary of changes**
- Getting ready to do state migrations.
- Function signatures are not stable yet, and are subject to change.

Closes 

### DIFF
--- a/types/networks/Cargo.toml
+++ b/types/networks/Cargo.toml
@@ -9,7 +9,6 @@ fil_types = { path = "../../types" }
 clock = { package = "fil_clock", path = "../../node/clock", version = "0.1" }
 lazy_static = "1.4"
 beacon = { path = "../../blockchain/beacon" }
-blocks = { package = "forest_blocks", path = "../../blockchain/blocks" }
 serde_json = "1.0"
 cid = { package = "forest_cid", version = "0.3" }
 ipld_blockstore = { path = "../../ipld/blockstore" }
--- a/types/networks/Cargo.toml
+++ b/types/networks/Cargo.toml
@@ -9,7 +9,9 @@ fil_types = { path = "../../types" }
 clock = { package = "fil_clock", path = "../../node/clock", version = "0.1" }
 lazy_static = "1.4"
 beacon = { path = "../../blockchain/beacon" }
+blocks = { package = "forest_blocks", path = "../../blockchain/blocks" }
 serde_json = "1.0"
-
+cid = { package = "forest_cid", version = "0.3" }
+ipld_blockstore = { path = "../../ipld/blockstore" }
 [features]
 interopnet = []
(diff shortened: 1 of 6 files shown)


vm/interpreter/src/vm.rs
```diff
@@ -181,6 +181,13 @@ where
             if i > parent_epoch {
                 self.run_cron(epoch, callback.as_mut())?;
             }
+            if is_migrate_epoch(i) {
```
austinabell, line 184: I'm confused about why the need to check if migration through the networks, can't you just match the epochs inside of `migrate_state` and in that function just return an `Option<Cid>` as the return type, and overwrite on `Some`? Seems cleaner than checking through the table, then matching again in the function (when implemented) to avoid returning an option.

types/networks/src/lib.rs
```diff
@@ -80,6 +97,27 @@ pub fn get_network_version_default(epoch: ChainEpoch) -> NetworkVersion {
...
+        .find(|upgrade| epoch == upgrade.height)
+        .map(|upgrade| upgrade.migration)
+        .unwrap_or(false)
+}
+
+/// Calls the state migration functions to migrate to the new StateTree.
+/// Currently this is a TODO because we havent implemented and migrations yet.
+/// The first migration to implement is from V2 to V3 (Network Version 10), so this method
+/// signature will probably change as we discover more things we need to pass in + error handling.
+pub fn migrate_state<T>(_bs: &T, _old_state: Cid, _epoch: ChainEpoch) -> Cid
```
austinabell, line 114:
```suggestion
pub fn migrate_state<T>(_bs: &T, _old_state: &Cid, _epoch: ChainEpoch) -> Result<Cid, Box<dyn StdError>>
```

Could wait to change to error prone until it's more clear what errors can arise, but also could wrap `Cid` in an option if you decide to take previous comment.

types/networks/Cargo.toml
```diff
@@ -10,6 +10,7 @@ clock = { package = "fil_clock", path = "../../node/clock", version = "0.1" }
 lazy_static = "1.4"
 beacon = { path = "../../blockchain/beacon" }
 serde_json = "1.0"
-
+cid = { package = "forest_cid", version = "0.3" }
+ipld_blockstore = { path = "../../ipld/blockstore" }
```
austinabell, line 14: don't use the path here, it will cause issues if the patch isn't applied or when things are released. The only reason this compiles is because of the patch to switch all to local for easier development
```suggestion
ipld_blockstore = "0.1"
```

types/networks/src/lib.rs
ec2: Good idea

vm/interpreter/src/vm.rs
ec2: If we have a function that does the check and migrate then we would have to flush after the cron regardless of whether we need to migrate or not. I thought we agreed that was wasteful?

austinabell: No you wouldn't because you would still only flush when matched, I'm just suggesting this to be able to keep that original function signature, but that doesn't matter much I don't think

austinabell (approved):

PR #988, opened 2021-02-05
Improve Car file read node

austinabell:
**Summary of changes**
- Noticed when looking around, there is no need to convert the cid to bytes just to get the amount of bytes read.

Closes 


cryptoquick (approved):
Nice catch!

ec2 (approved):
Nice

Issue #706, opened 2020-09-21
Add support for HTTP on RPC Server


amerameen:
**Acceptance Criteria**
The methods in [Node/RPC/source/lib.rs](https://github.com/ChainSafe/forest/blob/main/node/rpc/src/lib.rs) (30-40 methods) can all be called through WebSocket or HTTP (except ChainNotify which is streaming only)

**Notes**
Adding websocket support broke our existing CLI commands that were using HTTP (since the new library doesn't have HTTP support). 

async-tungstenite only has websocket support and there are no libraries that support websocket and http. So we probably need to use two different libraries (and likely bind to two different ports). 

Potential HTTP Library: jsonrpsee or tide? Tide is probably what we're going to stick with. 
Would like to use upstream jsonrpc-v2 instead of fork.

PR #990, opened 2021-02-06
HTTP RPC-JSON and tide-websockets

### LINKED ISSUE #986: Refactor JSON-RPC to use tide-websockets
**ACs**
- [x] Remove direct dependency on async_tungstenite
- [x] `node/rpc/src/lib.rs` uses tide-websockets
- [x] tide server upgrades WS connection to JSON-RPC server over HTTP
- [x] tide-websockets implemented for ChainNotify (currently our only websocket endpoint)
- [x] Lotus miner still works with RPC interface (i.e. we can produce blocks on a devnet with Forest)

**To Test**
- `lotus chain head` works the same on the forest daemon
  - `export FULLNODE_API_INFO=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJhZG1pbiJdfQ.tpPyEhluYQAGdZ2nsP6oMJBCLmsqpvGvhZ5uc_2xihk:/ip4/127.0.0.1/tcp/1235/http`
- `forest -c ../config.toml -p 1235`
- `manx "ws://127.0.0.1:1235/rpc/v0"`
- `{"jsonrpc": "2.0", "method":"Filecoin.ChainHead", "params": [], "id": 0}`

<https://github.com/hwchen/manx-rs> is a fantastic alternative to `wscat`

**Related**
Prerequisite for #706

cryptoquick:
**Summary of changes**
- [x] Add tide-websockets
  - [x] Fix state generic type in websocket closure
- [x] Add tide server in front of RPC server
- [x] Update JSON-RPC to upstream
- [ ] Error handling
- [x] Update packages

Closes #986 #706 

Currently having issues with getting the state generic type to work with Tide's websocket closure.


blockchain/chain/src/store/chain_store.rs
```diff
@@ -539,26 +543,39 @@ where
...
+    pub async fn sub_head_changes(&self) -> i64 {
+        let (tx, rx) = channel::unbounded();
         let mut subscriber = self.publisher.subscribe();
 
+        let sub_id = self.subscriptions.read().await.len() as i64;
+
+        self.subscriptions
+            .write()
+            .await
+            .insert(sub_id.clone(), Some(rx));
```
creativcoder, line 558: @cryptoquick  I think we can remove `.clone()` from `sub_id` as `i64` is `Copy`

cryptoquick: @creativcoder Good point! That was leftovers from something else. It's always a good idea to minimize clones. :+1:

cryptoquick:
ChainNotify now works, but just a note, it doesn't yet handle the disconnection gracefully by ending the loop. At least, I don't think it's doing that, according to the logs I can see.

node/rpc/src/rpc_ws_handler.rs
```diff
@@ -0,0 +1,139 @@
...
+                                        .with_method(RPC_METHOD_CHAIN_HEAD_SUB)
+                                        .finish(),
+                                )
+                                .await?;
+
+                                info!(
+                                    "RPC WS ChainNotify for subscription ID: {}",
+                                    subscription_id
+                                );
+
```
ec2, line 63: When I call chainnotify, i expect the first response to be the channel id, and the second response to be a "current" headchange event. Only after those two we start streaming the incoming tipsets out.

blockchain/chain/Cargo.toml
```diff
@@ -28,7 +28,7 @@ tokio = { version = "1.0", features = ["sync"] }
 address = { package = "forest_address", version = "0.3" }
 futures = "0.3.5"
 async-std = { version = "1.9", features = ["tokio1"] }
-types = { package = "fil_types", version = "0.1" }
+types = { package = "fil_types", version = "0.1", path = "../../types" }
```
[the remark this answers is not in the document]
cryptoquick: Whoops! Good catch. Was trying a lot of different things a while back.

node/rpc/src/rpc_ws_handler.rs
cryptoquick: I don't quite understand this instruction, you mean there should be two separate JSON responses immediately after calling ChainNotify over the websocket?

cryptoquick: Alright man. I'll take a look at how it worked before.

ec2: Dope. Its easy to miss and should be an easy fix 👍

ec2: For reference, this is how we do it currently in main: 
* Send the Channel ID (https://github.com/ChainSafe/forest/blob/main/node/rpc/src/lib.rs#L448)
* sub_head_changes will return the `current` event, if you look at it.  (https://github.com/ChainSafe/forest/blob/main/node/rpc/src/lib.rs#L435)

cryptoquick:
@ec2 Still working on some stuff, so still WIP technically, but getting kinda tired so I'm gonna call it a night for now.

cryptoquick:
Should be one of the last remaining issues:
https://github.com/kardeiz/jsonrpc-v2/issues/16

node/rpc/src/lib.rs
```diff
@@ -81,6 +81,7 @@ where
                 chain_head_sub::<DB, KS, B>,
             )
             .with_method("Filecoin.ChainNotify", chain_notify::<DB, KS, B>)
+            .with_method("xrpc.ch.val", chain_notify::<DB, KS, B>)
```
ec2, line 84: Not sure why you need this.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -539,26 +543,33 @@ where
...
     /// Subscribes to head changes. This function will send the current tipset through a channel,
     /// start a task that listens to each head change event and forwards into the channel,
-    /// then returns the receiver of this channel from the function.
+    /// then returns an id corresponding to the receiver of this channel from the function.
     /// This function is not blocking on events, and does not stall publishing events as it will
     /// skip over lagged events.
-    pub async fn sub_head_changes(&self) -> Receiver<HeadChange> {
-        let (tx, rx) = bounded(16);
+    pub async fn sub_head_changes(&self) -> i64 {
+        let (tx, rx) = channel::unbounded();
```
ec2, line 550: Should probably not be unbounded...

node/rpc/src/lib.rs
```diff
@@ -88,492 +49,204 @@ where
...
+            .with_method(
+                "Filecoin.ChainTipSetWeight",
+                chain_tipset_weight::<DB, KS, B>,
+            )
+            .with_method("Filecoin.ChainGetTipSet", chain_get_tipset::<DB, KS, B>)
+            .with_method(
+                "Filecoin.ChainHeadSubscription",
+                chain_head_sub::<DB, KS, B>,
+            )
+            .with_method("Filecoin.ChainNotify", chain_notify::<DB, KS, B>)
```
ec2, line 83: You defined a const for this name, but you dont use it here. Same with the other new ones you've defined, like ChainHeadSubscription

node/rpc/src/rpc_util.rs
```diff
@@ -0,0 +1,97 @@
...
+        Ok(err_str) => err_str,
+        Err(err) => format!("Failed to serialize error data. Error was: {}", err),
+    }
+}
+
+pub const RPC_METHOD_CHAIN_HEAD_SUB: &str = "Filecoin.ChainHeadSubscription";
+pub const RPC_METHOD_CHAIN_NOTIFY: &str = "Filecoin.ChainNotify";
+pub const RPC_METHOD_CHAIN_NOTIFY_RESPONSE: &str = "xrpc.ch.val";
+
+const STREAMING_METHODS: [&str; 1] = [RPC_METHOD_CHAIN_NOTIFY];
```
ec2, line 36: I think you missed the other two as well :p

node/rpc/src/rpc_ws_handler.rs
```diff
@@ -0,0 +1,148 @@
...
+{
+    let rpc_server = request.state();
+    let (ws_sender, mut ws_receiver) = ws_stream.split();
+    let ws_sender = Arc::new(Mutex::new(ws_sender));
+    let socket_active = Arc::new(AtomicCell::new(true));
+
+    info!("Accepted WS connection!");
+
+    while let Some(message_result) = ws_receiver.next().await {
+        match message_result {
```
ec2, line 37: Something is blocking in this while loop that is preventing me from spamming ourselves with ChainNotify calls. I would consider throwing this whole block into a task. There are certain Websocket calls like StateMessageWait that can take up a few minutes before a response (as it waits X epochs for transaction confirmations). And this will prevent any new requests from going through. I think this is why I cant spam us with ChainNotify calls.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -994,28 +1022,29 @@ where
...
 
-    #[derive(Serialize)]
+    #[derive(Debug, Deserialize, Serialize)]
     #[serde(rename_all = "lowercase")]
     #[serde(tag = "type", content = "val")]
-    pub enum HeadChangeJson<'a> {
-        Current(TipsetJsonRef<'a>),
-        Apply(TipsetJsonRef<'a>),
-        Revert(TipsetJsonRef<'a>),
+    pub enum HeadChangeJson {
```
ec2, line 1031: Are you sure we cant do this by reference anymore?

ec2 (changes requested):

ec2:
I deleted the `Cargo.lock` one from this branch, copied the one from `main`, and rebuilt and there is definitely a lot less changes than what you have up there. I personally think it's kinda bad practice to nuke your Cargo.lock like that. Lemme know what you think about this, open to have my mind changed w.r.t. this

cryptoquick:
> I deleted the `Cargo.lock` one from this branch, copied the one from `main`, and rebuilt and there is definitely a lot less changes than what you have up there. I personally think it's kinda bad practice to nuke your Cargo.lock like that. Lemme know what you think about this, open to have my mind changed w.r.t. this

That's fine. Go ahead and commit it.

blockchain/chain/src/store/chain_store.rs
cryptoquick: Is it better if we do it by reference? I took out the explicit lifetime because it didn't seem necessary anymore.

node/rpc/src/rpc_util.rs
cryptoquick: There are others?

node/rpc/src/rpc_ws_handler.rs
cryptoquick: That makes sense, I'll do that in the morning.

node/rpc/src/lib.rs
cryptoquick: Yeah, it seemed inconsistent with the pattern used for the other methods on the router. I think maybe I should refactor all the method names out in a more deliberate and well-thought out way when working on the client library, which would be a separate PR.

blockchain/chain/src/store/chain_store.rs
cryptoquick: What's wrong with unbounded?

node/rpc/src/lib.rs
cryptoquick: You're right. I was experimenting with a different way to handle ChainNotify, but then I realized, the response is formatted like a request. So, I have to manually build the response struct.

I'll take this out.

ec2: It panicked btw, when i took it out, so there must be a use for it somewhere? No idea xD. It's just really confusing, and I'm ok with you leaving it, but it'd be nice to have a comment to describe what its for.

node/rpc/src/rpc_util.rs
ec2: Filecoin.ChainHeadSubscription and xrpc.ch.val?

blockchain/chain/src/store/chain_store.rs
ec2: Mmm, i think doing things by ref is better for serde serialization? I'm not sure though. Not a big concern for me. Just pointing it out.

blockchain/chain/src/store/chain_store.rs
ec2: If the websocket connection is slow, then we might end up just filling up the unbounded buffer. Is there a reason not to bound this by something reasonable like 16 (what we had before)? Seems like limiting will only benefit us if we ever have to chase down memory issues

node/rpc/src/rpc_util.rs
cryptoquick: Technically only ChainNotify is streaming, the others are traditional request-response. I can make a note of that if you like?

ec2: Youre right. Lets make a note of that. I was more making sure that you cant call those other two methods from the HTTP endpoint.

blockchain/chain/src/store/chain_store.rs
cryptoquick: Alright, I've put it back to a bounded channel.

node/rpc/src/lib.rs
cryptoquick: To remove it there's more to refactor. I've done so.

node/rpc/src/rpc_util.rs
cryptoquick: I've removed one, I don't think it's necessary to disable HTTP calls to the other.

ec2: I think we should prevent `Filecoin.ChainHeadSubscription` from being called from HTTP still.
Its not like they can poll for the subscriptions anyways.

cryptoquick: Fair enough. It also does provide more balance to the code so it doesn't look like an oversight. Done.

ec2:
Last functional change, and I think we'll be good.
So when ChainNotify is called, the first two things (the channel id response, and the current event) should be sent back right away. What I am seeing is that those dont get returned until after a new block comes in.

Great work!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

cryptoquick:
Well, it does that for the first ChainNotify call I think, but if you issue additional ChainNotify subscriptions, I don't think the old message is no longer in that broadcast channel anymore. Did it work differently before?

ec2:
Yeah, it used to/has to be the case that i can issue multiple ChainNotifies and have them all respond right away. One application can have multiple places where it subscribes to chain notifies. You can check that behaviour in `main`.

node/rpc/src/data_types.rs
```diff
@@ -0,0 +1,48 @@
...
+use state_manager::StateManager;
+use wallet::KeyStore;
+
+pub type SubscriptionHeadChange = (i64, HeadChangeJson);
+
+#[derive(Serialize)]
+pub struct StreamingData<'a> {
+    pub json_rpc: &'a str,
+    pub method: &'a str,
+    pub params: SubscriptionHeadChange,
```
ec2, line 26: This is what it should be
```suggestion
    pub params: (i64, Vec<HeadChangeJson>),
```

ec2 (changes requested):

node/rpc/src/data_types.rs
cryptoquick: That's what was causing the deserialization error earlier, which is why I changed it. I'll put it back, and try to solve the deserialization issue a different way.

ec2: Deserialization error? I have this working right now... Didn't change much to get this working.

cryptoquick: Yeah, no, it wasn't difficult to solve, now that I knew where to look. Try now.

ec2 (approved):

PR #1001, opened 2021-02-08
Actors v3 Setup

ec2:
**Summary of changes**
- Change actors interface to include new HAMT and AMT
- Update Code CIDs

Closes 

### DIFF
--- a/vm/actor_interface/Cargo.toml
+++ b/vm/actor_interface/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 actorv0 = { package = "forest_actor", version = "0.1.3" }
 actorv2 = { package = "forest_actor", version = "1.0.0" }
+actorv3 = { package = "forest_actor", path = "../actor" }
 fil_types = "0.1"
 vm = { package = "forest_vm", version = "0.3.1" }
 ipld_blockstore = "0.1"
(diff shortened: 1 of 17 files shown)


vm/actor_interface/Cargo.toml
```diff
@@ -20,7 +21,8 @@ forest_bitfield = "0.1"
 num-bigint = { version = "0.1.1", package = "forest_bigint", features = ["json"] }
 forest_hash_utils = "0.1"
 forest_json_utils = "0.1.1"
-
+ipld_hamt = { path = "../../ipld/hamt" }
+ipld_amt = { path = "../../ipld/amt", features = ["go-interop"]}
```
austinabell, line 25: these should be defined in actors, like the other versions. This will just add confusion and unnecessary duplication of this dependency reference.

I see the TODO, but what's stopping this from being done now?

austinabell (changes requested):

vm/actor_interface/Cargo.toml
ec2: Because there will be like a bazillion something errors because the interface has changed. Which means ill have to update the actors code to the new interface only to need to update it again when i actually update the actors.

austinabell: Oh I see, sure whatever works better for you then. This was opinionated I guess

austinabell (approved):

PR #1005, opened 2021-02-09
V3 HAMT and AMT for Actors

ec2:
**Summary of changes**
- Branch Name is a little weird, ignore the name 
- Adds the V3 HAMT and AMT to the actors crate
- Removes importing the V3 HAMT and AMT libraries from the interface, and import it from the actors crate instead. 

Closes 

### DIFF
--- a/vm/actor/Cargo.toml
+++ b/vm/actor/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "forest_actor"
 description = "Actors for the Filecoin protocol"
-version = "0.1.0"
+version = "1.0.0"
 license = "MIT OR Apache-2.0"
 authors = ["ChainSafe Systems <info@chainsafe.io>"]
 edition = "2018"
--- a/vm/actor/Cargo.toml
+++ b/vm/actor/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "forest_actor"
 description = "Actors for the Filecoin protocol"
-version = "1.0.0"
+version = "2.0.0"
 license = "MIT OR Apache-2.0"
 authors = ["ChainSafe Systems <info@chainsafe.io>"]
 edition = "2018"
--- a/vm/actor/Cargo.toml
+++ b/vm/actor/Cargo.toml
@@ -21,6 +21,7 @@ serde = { version = "1.0", features = ["derive"] }
 lazy_static = "1.4.0"
 ipld_blockstore = "0.1"
 ipld_hamt = { path = "../../ipld/hamt" }
+ipld_amt = { path = "../../ipld/amt", features = ["go-interop"]}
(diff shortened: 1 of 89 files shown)


vm/actor/src/util/set_multimap.rs
```diff
@@ -114,11 +116,11 @@ where
         };
 
         set.for_each(|k| {
-            let v = parse_uint_key(&k)
+            let v = parse_usize_key(&k)
```
austinabell, line 119: cases like this is confusing why the use of usize key, because it's used as a u64

vm/actor/src/lib.rs
```diff
@@ -72,11 +72,22 @@ pub fn u64_key(k: u64) -> BytesKey {
...
 pub fn parse_uint_key(s: &[u8]) -> Result<u64, UVarintError> {
     let (v, _) = unsigned_varint::decode::u64(s)?;
     Ok(v)
 }
 
+pub fn parse_usize_key(s: &[u8]) -> Result<usize, UVarintError> {
+    let (v, _) = unsigned_varint::decode::usize(s)?;
+    Ok(v)
+}
+
```
austinabell, line 90: I can't really see what the benefit is in having both of these, but if it makes it easier then that's nbd

vm/actor/Cargo.toml
```diff
@@ -20,8 +20,7 @@ cid = { package = "forest_cid", version = "0.3", features = ["cbor"] }
 serde = { version = "1.0", features = ["derive"] }
 lazy_static = "1.4.0"
 ipld_blockstore = "0.1"
-ipld_hamt = { version = "1.0", features = ["go-interop"] }
-ipld_amt = { features = ["go-interop"], version = "0.2" }
+ipld_hamt = { path = "../../ipld/hamt" }
```
austinabell, line 23: why can't these be removed (hamt and amt below)

ec2: ??? I removed the old ones from crates and added the local onces

vm/actor/src/util/set_multimap.rs
ec2: oh whoops

vm/actor/Cargo.toml
austinabell: oh oops thought I was reading a diff cargo toml

ec2: haha i had a feeling :p

vm/actor/src/lib.rs
ec2: fixed. i think i got a little haphazard with the u64 <-> usize conversions

austinabell (approved):
UT but lgtm

ec2:
Just tested. Still syncing. MERGING!

PR #1010, opened 2021-02-11
Market Actor v3

ec2:
**Summary of changes**
- Its all in the title

Closes 


vm/actor/src/builtin/market/mod.rs
```diff
@@ -229,9 +218,19 @@ impl Actor {
...
+        let mut caller_ok = caller == &worker;
+        for controller in controllers.iter() {
+            if caller_ok {
+                break;
+            }
+            caller_ok = caller == controller;
+        }
+        if !caller_ok {
+            return Err(
+                actor_error!(ErrForbidden; "caller {} is now worker or control address of provider {}", caller, provider),
```
austinabell, line 232: No need to change here, but you'll run into less formatting issues later on if you use the `,` separator instead of `;`. For some reason semicolon in a macro avoids formatting and can void formatting for the whole file if over the width limit. I had been slowly transitioning over with the plan of removing support for the `;`, but totally optional.

vm/actor/src/builtin/market/mod.rs
```diff
@@ -279,7 +278,7 @@ impl Actor {
                     .pending_deals
                     .as_ref()
                     .unwrap()
-                    .contains_key(&pcid.to_bytes())
+                    .has(&pcid.to_bytes())
```
austinabell, line 281: optional: I know not changed in this, but the `has` function could use the `contains_key` method on the Hamt that was added after the method was added on the `Set` (functionally the same)

vm/actor/src/builtin/market/mod.rs
```diff
@@ -381,29 +380,36 @@ impl Actor {
...
-            &params.deal_ids,
-            &miner_addr,
-            params.sector_expiry,
-            params.sector_start,
-        )
-        .map_err(|e| {
-            e.downcast_default(
-                ExitCode::ErrIllegalState,
-                "failed to validate deal proposals for activation",
+        let mut weights = Vec::with_capacity(params.sectors.len());
```
austinabell, line 390: this will panic, although the capacity is set here, the length needs to be set as you are indexing below. Either you need to initialize the vector with that length before iterating, or instead of indexing below just push to the vector below. I recommend the latter to avoid indexing, but up to you.

If the go initialization is confusing, it's
```
make(<type>, <length>, <capacity>)
```

so they are specifically sizing the slice in go, which is why they can index it.

vm/actor/src/builtin/market/mod.rs
```diff
@@ -563,10 +569,9 @@ impl Actor {
                 }
                 let deal = deal.unwrap();
 
-                assert_eq!(
-                    deal.provider, miner_addr,
-                    "caller is not the provider of the deal"
-                );
+                if deal.provider != miner_addr {
+                    return Err(actor_error!(ErrIllegalState; "caller {} is not the provider {} of deal {}", miner_addr, deal.provider, id));
```
austinabell, line 573: same here as the point above, formatting will be off (especially with how long this line is) and affects the rest of the file and will be confusing for people modifying in the future

vm/actor/src/builtin/market/mod.rs
```diff
@@ -813,17 +817,14 @@ impl Actor {
 
                     let (slash_amount, next_epoch, remove_deal) =
                         msm.update_pending_deal_state(&state, &deal, curr_epoch)?;
-                    assert_ne!(
-                        slash_amount.sign(),
-                        Sign::Minus,
-                        "next scheduled epoch should be undefined as deal has been removed"
-                    );
+                    if slash_amount.sign() == Sign::Minus {
```
austinabell, line 820: if you bring in the `Signed` trait from `num_traits` or whatever it's called, you can just call `.is_negative()` and it's a bit easier to read

vm/actor/src/builtin/market/mod.rs
```diff
@@ -863,6 +864,13 @@ impl Actor {
                             "deal should not be slashed and should have a schedule for next cron"
                         );
```
austinabell, line 865: Can't comment on these lines, but the assertions should be removed (you're checking this in the lines you added below). I might have missed some above, but be sure to remove all those assertions

vm/actor/src/builtin/market/state.rs
```diff
@@ -54,21 +55,39 @@ pub struct State {
...
-            deal_ops_by_epoch: empty_mset,
+            deal_ops_by_epoch: empty_deal_ops_hamt,
             last_cron: EPOCH_UNDEFINED,
 
             total_client_locked_colateral: TokenAmount::default(),
             total_provider_locked_colateral: TokenAmount::default(),
             total_client_storage_fee: TokenAmount::default(),
-        }
+        })
     }
```
austinabell, line 92: I can't comment anywhere that isn't changed, but the assertions in this file should probably be changed to state errors, as they did

vm/actor/src/lib.rs
```diff
@@ -53,6 +53,16 @@ where
...
 
+/// Create a hamt with a custom bitwidth.
+#[inline]
+pub fn make_map_with_bitwidth<BS, V>(store: &'_ BS, bitwidth: u32) -> Map<'_, BS, V>
+where
+    BS: BlockStore,
+    V: DeserializeOwned + Serialize,
+{
+    Map::<_, V>::new_with_bit_width(store, bitwidth)
+}
```
austinabell, line 64: Optional: you can have duplicates where the original uses the default hamt bit width, or you could just pass in the width for each. I'd argue it would be less prone to an error if you did the latter, so this wouldn't implicitly fail if someone didn't know about the default. Very opinionated though and this could definitely be better, but having 4 functions with different names for all combinations may be confusing

vm/actor/src/util/set.rs
```diff
@@ -53,10 +58,11 @@ where
 
     /// Deletes key from set.
     #[inline]
-    pub fn delete(&mut self, key: &[u8]) -> Result<(), Error> {
-        self.0.delete(key)?;
-
-        Ok(())
+    pub fn delete(&mut self, key: &[u8]) -> Result<Option<()>, Error> {
```
austinabell, line 61: `Option<()>` can definitely be simplified to be a `bool`, can't think of a benefit to having it as an option

vm/actor/src/builtin/market/mod.rs
ec2: ah, yeah that is a whoopsies

vm/actor/src/builtin/market/state.rs
ec2: Yeah, good call.

vm/actor/src/util/set.rs
ec2: I like the option because then we can do the method chaining when handling the error instead of checking if its Some(false). What do you think?

vm/actor/src/lib.rs
ec2: Yeah, youre right. Best thing is to mandate that a bitwidth be passed through

vm/actor/src/lib.rs
```diff
@@ -53,6 +53,16 @@ where
     Map::<_, V>::new_with_bit_width(store, HAMT_BIT_WIDTH)
 }
 
+/// Create a hamt with a custom bitwidth.
+#[inline]
+pub fn make_map_with_bitwidth<BS, V>(store: &'_ BS, bitwidth: u32) -> Map<'_, BS, V>
```
austinabell, line 58: small point, if both of these always requires a bit width, perhaps you don't need the `_with_bitwidth` suffix. Just an opinion though, this is totally fine.

vm/actor/src/util/set.rs
austinabell: Then what is the need to erase the return in the function? This may lead to having to do unnecessary work if you need to retrieve the value first?

austinabell (approved):

vm/actor/src/lib.rs
ec2: hah, makes sense. im gunna completely remove `make_map` and replace it with this.

vm/actor/src/util/set.rs
ec2: Good catch. gunna return the byteskey

ec2: Actually... Does that make sense? we should already have the key.... The value of the underlying map is always () because this is a Set that backed by a hamt

vm/actor/src/lib.rs
austinabell: yeah can definitely do later, I'm just suggesting more as something to move toward

PR #1051, opened 2021-03-19
Libp2p Connection Limits

### LINKED ISSUE #798: Limit max active peers
**Issue summary**

Technically this can be a security vulnerability because a large amount of lightweight nodes which can strain on resources. 

This is definitely low priority for now, because it would require also correctly managing and disconnecting bad peers to make sure this doesn't become an actual dos vector.

ec2:
**Summary of changes**
- Add new flag `--target-peer-count` to have a sort of bound on the number of peers we should be connected to with a default of 75. 
- Added Topic scoring to Gossipsub topics. They are currently disabled because I don't think were using them properly, but they aren't important for now, but might be in the future, so just leaving it in for posterity. 

Closes 
#798 

### DIFF
--- a/forest/src/cli/mod.rs
+++ b/forest/src/cli/mod.rs
@@ -133,8 +133,8 @@ impl DaemonOpts {
 
         cfg.network.kademlia = self.kademlia.unwrap_or(cfg.network.kademlia);
         cfg.network.mdns = self.mdns.unwrap_or(cfg.network.mdns);
-        if let Some(target_peer_count) = &self.target_peer_count {
-            cfg.network.target_peer_count = *target_peer_count;
+        if let Some(target_peer_count) = self.target_peer_count {
+            cfg.network.target_peer_count = target_peer_count;
         }
         // (where to find these flags, should be easy to do with structops)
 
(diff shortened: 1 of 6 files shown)


forest/src/cli/mod.rs
```diff
@@ -128,6 +133,9 @@ impl DaemonOpts {
 
         cfg.network.kademlia = self.kademlia.unwrap_or(cfg.network.kademlia);
         cfg.network.mdns = self.mdns.unwrap_or(cfg.network.mdns);
+        if let Some(target_peer_count) = &self.target_peer_count {
+            cfg.network.target_peer_count = *target_peer_count;
+        }
```
cryptoquick, line 138: You don't need to reference and dereference here.

cryptoquick (changes requested):
Minor nitpick, otherwise, looks good.

forest/src/cli/mod.rs
ec2: done

cryptoquick (approved):
Looks good.

creativcoder (approved):

PR #1064, opened 2021-03-29
Clippy "Fix"

ec2:
**Summary of changes**
- upper-case-acronyms is disabled in `make lint`. We could fix it, but it would require releasing a bunch of crates, so we should do it the next time we do a round of releases. 

Closes 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ clean:
 
 lint: license clean
 	cargo fmt --all
-	cargo clippy --all-features -- -D warnings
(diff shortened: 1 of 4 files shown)


olibearo (approved):
LGTM, thanks @ec2

cryptoquick (approved):
Nice! I love how clippy helps make things better over time, even if little surprises like this happen here and there. 

@ec2 mind making a tracking issue for correcting this, making a note that we'll fix it when we go to publish packages next?

PR #1066, opened 2021-03-30
NetworkVersion 11 Upgrade at Epoch 665280

ec2:
**Summary of changes**
- See FIP 0014 (https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0014.md)
- TL;DR Allows for sectors using the old (slightly insecure) proofs type to be extended to a maximum of 540 days. 

Closes 


types/networks/src/lib.rs
```diff
@@ -74,6 +74,10 @@ const VERSION_SCHEDULE: [Upgrade; 10] = [
         height: UPGRADE_ACTORS_V3_HEIGHT,
         network: NetworkVersion::V10,
     },
+    Upgrade {
+        height: UPGRADE_NORWEGIAN_HEIGHT,
+        network: NetworkVersion::V11,
+    },
```
cryptoquick, line 80: Wouldn't this make sense to order from most recent to least recent? That way it'd have to do fewer checks in the iterator? How often is `get_network_version_default` called?

cryptoquick (changes requested):

types/networks/src/lib.rs
cryptoquick: Oh cool, I missed that.

PR #1069, opened 2021-04-11
Tweaks to buffered blockstore

austinabell:
**Summary of changes**
Were just changes I noticed when scanning code for what I was working on, figured I'd PR

What changed:
- Switched trait to `Read` since there was no need for the `AsyncRead` specific functions
- Removed `BufferedReader` wrapper for the bytes
  - There is no need to have the overhead of the buffered reader when the bytes are already in memory, it should be redundant, unless I'm missing something here?
- Switched the scanning for links to attach a callback instead of allocating a vector of Cids, just to iterate and copy them
  - There are iterator based alternatives to this with the same function, but this was the least invasive change

Closes 


amerameen:
Look who's back!!! ❤️

creativcoder (approved):
Looks neat!

cryptoquick (approved):
Agreed! Good improvements. Much appreciated!

PR #1072, opened 2021-04-13
HTTP JWT validation

### LINKED ISSUE #997: JSON-RPC JWT Authentication
**ACs**
- [ ] RPC endpoint validates jwt token from Lotus client
  - [ ] Both for HTTP & Websockets
- [ ] Test cases for both unauthorized and authorized cases

**Notes**
- How do we know who has which permissions? Is there a table? @ec2 will follow up

cryptoquick:
**Summary of changes**
- HTTP JWT validation

Closes #997 


cryptoquick:
Tested wtih:

`FULLNODE_API_INFO="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIl19.3gT2mGAxugWuN8bG3Znndcv4V4cJlhZkIuE9zhp6Lh4:/ip4/127.0.0.1/tcp/1235/http" lotus chain head`

`FULLNODE_API_INFO="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.PFlC4lndH_Dw8GitzmTS6nxxGOia_kuefCWHd5FwtR0:/ip4/127.0.0.1/tcp/1235/http" lotus auth create-token --perm write`

When testing with:
`FULLNODE_API_INFO="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIl19.3gT2mGAxugWuN8bG3Znndcv4V4cJlhZkIuE9zhp6Lh4:/ip4/127.0.0.1/tcp/1234/http" lotus wallet new`

This will return a 403 Forbidden error, but lotus cannot possibly cope, and so it segfaults.

node/rpc/src/auth_api.rs
```diff
@@ -35,7 +38,9 @@ where
     B: Beacon + Send + Sync + 'static,
 {
     let ks = data.keystore.read().await;
-    let (token,) = params;
+    let (header_raw,) = params;
+    let token = header_raw.trim_start_matches("Bearer ");
+    info!("Auth verify token: {}", token);
```
ec2, line 43: I'd remove these logs. They are already captured by the logs that print when we call an RPC method

node/rpc/src/auth_api.rs
```diff
@@ -18,6 +20,7 @@ where
     B: Beacon + Send + Sync + 'static,
 {
     let (perms,) = params;
+    info!("Auth new requested for permissions: {:?}", perms);
```
ec2, line 23: I'd remove these logs. They are already captured by the logs that print when we call an RPC method

node/rpc/src/auth_api.rs
```diff
@@ -35,7 +38,9 @@ where
     B: Beacon + Send + Sync + 'static,
 {
     let ks = data.keystore.read().await;
-    let (token,) = params;
+    let (header_raw,) = params;
+    let token = header_raw.trim_start_matches("Bearer ");
```
olibearo, line 42: If we add case insensitivity here for the `Bearer ` prefix, we might save a stranger some time. 

`trim_start_matches` accepts a `&[char]` as a `Pattern`:
https://doc.rust-lang.org/std/primitive.str.html#method.trim_start_matches

node/rpc/src/rpc_util.rs
```diff
@@ -42,6 +47,62 @@ pub fn is_streaming_method(method_name: &str) -> bool {
...
+    KS: KeyStore + Send + Sync + 'static,
+    B: Beacon + Send + Sync + 'static,
+{
+    match authorization_header {
+        None => return Ok(()),
+        Some(header_values) => match header_values.get(0) {
+            None => return Ok(()),
+            Some(token) => {
+                let token = token.to_string();
+                info!("JWT from HTTP Header: {}", token);
```
ec2, line 68: I'd remove this log too. Or at least make it debug!

node/rpc/src/auth_api.rs
olibearo: but doesn't accepts a `&[str]`, so that's less helpful, It might require two calls

cryptoquick: This approach is compatible with Lotus, so that's all I'm concerned about right now

ec2 (approved):

node/rpc/src/auth_api.rs
```diff
@@ -35,7 +37,8 @@ where
     B: Beacon + Send + Sync + 'static,
 {
     let ks = data.keystore.read().await;
-    let (token,) = params;
+    let (header_raw,) = params;
```
creativcoder, line 40: Do we extract anything else from `params`, if not then we could just rename it to `header_raw`?

node/rpc/src/rpc_util.rs
```diff
@@ -42,6 +47,62 @@ pub fn is_streaming_method(method_name: &str) -> bool {
...
+    authorization_header: Option<HeaderValues>,
+) -> Result<(), tide::Error>
+where
+    DB: BlockStore + Send + Sync + 'static,
+    KS: KeyStore + Send + Sync + 'static,
+    B: Beacon + Send + Sync + 'static,
+{
+    match authorization_header {
+        None => return Ok(()),
+        Some(header_values) => match header_values.get(0) {
```
creativcoder, line 64: We could use [and_then](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then) to cleanup things here.

node/rpc/src/rpc_util.rs
```diff
@@ -42,6 +47,62 @@ pub fn is_streaming_method(method_name: &str) -> bool {
...
+            }
+        },
+    }
+}
+
+pub fn get_auth_header(
+    request: tide::Request<JsonRpcServerState>,
+) -> (Option<HeaderValues>, tide::Request<JsonRpcServerState>) {
+    match request.header("Authorization") {
+        Some(header) => (Some(header.to_owned()), request),
```
creativcoder, line 101: maybe `map` could be used here?

node/rpc/src/rpc_http_handler.rs
```diff
@@ -44,12 +45,11 @@ where
         ));
     }
 
-    let rpc_server = http_request.state();
-    let rpc_response = call_rpc_str(rpc_server.clone(), rpc_call).await?;
-    let http_response = HttpResponse::builder(200)
-        .body(rpc_response)
+    let result = call_rpc_str(rpc_server.clone(), rpc_call).await?;
+    let response = tide::Response::builder(200)
```
creativcoder, line 49: Could qualify `tide::` imports here

creativcoder (changes requested):

node/rpc/src/auth_api.rs
cryptoquick: No, it's how the JSON-RPC interface is called by Lotus; it's nested within a single array, and we destructure that from a tuple in our code.

node/rpc/src/rpc_http_handler.rs
cryptoquick: What do you mean?

creativcoder: oh i meant to import the types at top.

node/rpc/src/auth_api.rs
creativcoder: ah okay

node/rpc/src/rpc_http_handler.rs
cryptoquick: I prefer to leave it the way I changed it to so it's more explicit by not aliasing between http terms (request, response) and RPC terms (call, result).

cryptoquick:
@creativcoder Really good Rust style suggestions, btw.

creativcoder (approved):

PR #1075, opened 2021-04-17
Remove a couple unnecessary panics

huitseeker:
By folding them in the Error case of their encompassing Result

Tool-aided by [comby-rust](https://github.com/huitseeker/compby-rust)


ec2 (approved):
Thanks for this contribution :D

cryptoquick (approved):
Nice 👌

PR #1078, opened 2021-04-21
Encrypted Key Store

### LINKED ISSUE #1033: Encrypted Key Store
ACs
- [x] config.toml option for `encrypted_keystore: bool`
- [x] Pass config to keystore
- [x] Encryption methods using PBKDF2 and sodiumoxide secretbox
- [x] Unit tests for encryption methods
- [x] Separate binary `keystore` file for when using an encrypted keystore
- [x] CBOR serialization for binary encrypted keystore
- [x] Instead of a hostname salt, we need to use a random bytes salt (32 should be enough) when the key is first generated, and it should be passed an `Option<&[u8; 32]>` it can use if the `keystore` file already exists. We'll just prepend the salt to the beginning of the file, and pass a byte slice to derive the same key. This way the keystore is portable between hosts. That's important for backups. The keystore is used for wallet keys, and it'd really suck to lose access to FIL if a user lost track of their hostname.
- [x] Whenever the keystore is accessed, the keystore will need to check whether it's configured to use encrypted or plaintext keystore, and if it's encrypted, use the encryption key to:
  - [x] Decrypt on read
  - [x] Encrypt on write
- [x] When setting a passphrase for the first time, ask for it to be entered once more, and compare that they are the same. If not, prompt the user for a new one.
- [x] Error handling when unable to parse a decrypted keystore (say, if an incorrect passphrase was used to derive the encryption key)
  - [x] Prompt the user for passphrase in case they mistyped it.
- [x] Interactively ask for passphrase on daemon start if encryption is enabled
- [x] This interactive mode will block initialization until it's completed (probably using some async/await)
- [x] `keystore.unlock(passphrase: String)` method that will call the `generate_key` method, but also manage daemon init blocking, and will be called by the encryption RPC API.

Notes
- Allow users to encrypt using a password

connormullett:
**Summary of changes**
- This PR introduces methods to encrypt and decrypt messages and payloads symmetrically. This will further be utilized in the keystore to securely save keys.

Closes #1033 

Thank you 🔥


cryptoquick:
In addition to `key_info` I think we should have two additional properties on the `MemKeyStore` struct:

- `unlocked: bool`: has it been unlocked yet
- `encrypted_key: bool`: it's an encrypted key

@ec2 I'd like your opinions on these:

1. We may have to introduce checks wherever the keystore is read from to throw if an encrypted key needs to be decrypted. I'm not sure if we should have a default key that's read-only.
2. Should we introduce checks or requirements for encrypted keys to only work if an HTTP connection is secure?
3. Should we introduce a requirement for admin, sign, and write keys to require encryption?

I think 2 & 3 these are nice-to-haves, maybe I can put them in a separate issue, when we have the option for running in "production" mode vs "development" mode.

cryptoquick:
Also @connormullett , check this out:
- `target/release/forest -c ../config.toml -p 1235`
- `FULLNODE_API_INFO="/ip4/127.0.0.1/tcp/1235/http" lotus auth create-token --perm admin`
- Take the resulting admin JWT, and replace the one here:
- `FULLNODE_API_INFO="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.PFlC4lndH_Dw8GitzmTS6nxxGOia_kuefCWHd5FwtR0:/ip4/127.0.0.1/tcp/1235/http" lotus wallet new`
- Then check `keystore.json`
- All these private keys should be capable of being encrypted

Though, should we be able to encrypt a private key that's already be written to the keystore, or should we only write freshly encrypted keys?

forest/src/daemon.rs
```diff
@@ -13,8 +13,10 @@ use libp2p::identity::{ed25519, Keypair};
 use log::{debug, info, trace};
 use message_pool::{MessagePool, MpoolConfig, MpoolRpcProvider};
 use paramfetch::{get_params_default, SectorSizeOpt};
+use rpassword::read_password;
```
cryptoquick, line 16: Awesome find!

key_management/src/keystore.rs
```diff
@@ -242,19 +249,14 @@ impl KeyStore for PersistentKeyStore {
...
+        let file = OpenOptions::new()
             .write(true)
             .create(true)
             .open(&self.location)
             .map_err(|err| Error::Other(err.to_string()))?;
-        let json =
-            serde_json::to_string(&self.key_info).map_err(|err| Error::Other(err.to_string()))?;
-        let encrypted_json = PersistentKeyStore::encrypt(&generated_key, json.as_bytes())?;
-        file.write_all(&encrypted_json)?;
+        serde_json::to_writer(&file, &self.key_info)
```
cryptoquick, line 258: Do you think we could use serde_cbor here instead?

cryptoquick (changes requested):

key_management/src/keystore.rs
ec2: Makes sense to json this. Its going into a file and a user might wanna inspect it or something. Generally id say we should use cbor for when we wanna store in DB or send over the wire.

cryptoquick: We're actually using a separate file for the encrypted keystore (without the `.json` extension), and the entire file is encrypted bytes, so I think it makes more sense to serialize the data to bytes, also.

connormullett: So we don't want to use CBOR? Just keep everything as JSON and encrypt the data as a whole?

cryptoquick: I think CBOR is better for the encrypted keystore, but JSON is better for the plaintext JSON file

key_management/src/keystore.rs
```diff
@@ -226,11 +226,17 @@ impl PersistentKeyStore {
...
-        serde_json::to_writer(writer, &self.key_info)
-            .map_err(|e| Error::Other(format!("failed to serialize and write key info: {}", e)))?;
+        if let true = self.is_encrypted {
+            serde_cbor::to_writer(writer, &self.key_info).map_err(|e| {
+                Error::Other(format!("failed to serialize and write key info: {}", e))
+            })?;
+        } else {
+            serde_json::to_writer(writer, &self.key_info).map_err(|e| {
+                Error::Other(format!("failed to serialize and write key info: {}", e))
+            })?;
```
cryptoquick, line 247: Perfect!

connormullett: Well, maybe not perfect if Eric thinks we should just keep track of one file and encrypt the keys

cryptoquick: I'm not sure he understands the context. Let's keep it this way for now.

connormullett: :+1:

forest/src/daemon.rs
```diff
@@ -40,7 +42,21 @@ pub(super) async fn start(config: Config) {
         });
 
     // Initialize keystore
-    let mut ks = PersistentKeyStore::new(config.data_dir.to_string()).unwrap();
+    let passphrase = if let true = config.encrypt_keystore {
```
olibearo, line 45: You don't need the `if let`, almost reads like poetry though

cryptoquick: Oh, good point! I'm not sure I understand this usage of if let.

ec2: Wait, is this syntactically valid??? 😮

forest/src/daemon.rs
```diff
@@ -40,7 +42,21 @@ pub(super) async fn start(config: Config) {
...
+        print!("Keystore passphrase: ");
+        std::io::stdout().flush().unwrap();
+        Some(read_password().expect("Error reading passphrase"))
+    } else {
+        None
+    };
+
+    let mut ks = PersistentKeyStore::new(
```
olibearo, line 53: Just as a reader it would be nice to know why we expect this to never fail. Either a comment or switching to `.expect("reason")` would be nice

forest/src/daemon.rs
cryptoquick: I wouldn't be surprised if it was, but I honestly couldn't say what the return value would be, or how the assignment works... It's kinda breaking my brain rn, ngl :joy_cat:

connormullett: Noted. Using a match instead.

key_management/src/keystore.rs
```diff
@@ -147,18 +181,33 @@ impl KeyStore for MemKeyStore {
     fn remove(&mut self, key: String) -> Result<KeyInfo, Error> {
         self.key_info.remove(&key).ok_or(Error::KeyInfo)
     }
+
+    fn unlock(&mut self, _passphrase: &str) -> Result<(), EncryptedKeyStoreError> {
+        todo!()
```
olibearo, line 186: Did I start reviewing this too early? If not, just a reminder to fill this method.

connormullett: It's still a draft PR. Eyes on at any stage helps me out. And yes, this method will be used by the RPC methods and CLI commands later on, just haven't gotten that far yet.

forest/src/daemon.rs
olibearo: The if statement is fine, you just don't need to destructure the boolean since it has no ... structure :)

forest/src/daemon.rs
connormullett: If the keystore fails to be created, its either because of configuration issues, or creating the necessary files. The errors were propagated up in the associated function. Changed this to use `.expect()`.

key_management/src/keystore.rs
olibearo: Ahh my bad, carry on

forest/src/daemon.rs
connormullett: > Wait, is this syntactically valid???

yes it is lol

cryptoquick: @olibearo ahhh, that's how to read it...

cryptoquick (approved):
I just tested it out, it works great!
Great work, @connormullett, and a fantastic first contribution!

key_management/src/keystore.rs
```diff
@@ -113,127 +117,387 @@ pub trait KeyStore {
...
 
-impl MemKeyStore {
-    /// Return a new empty KeyStore
-    pub fn new() -> Self {
-        MemKeyStore {
-            key_info: HashMap::new(),
-        }
-    }
+pub enum KeyStoreConfig {
+    Memory(),
```
creativcoder, line 129: Nit: we don't need the `()`

key_management/src/keystore.rs
```diff
@@ -113,127 +117,387 @@ pub trait KeyStore {
...
+impl KeyStore {
+    pub fn new(config: KeyStoreConfig) -> Result<Self, Error> {
+        match config {
+            KeyStoreConfig::Memory() => Ok(Self {
+                key_info: HashMap::new(),
+                persistence: None,
+                encryption: None,
+            }),
+            KeyStoreConfig::Persistent(location) => {
+                let file_path = location.join(Path::new(KEYSTORE_NAME));
```
creativcoder, line 172: We can just pass `KEYSTORE_NAME` to join, as it accepts anything that implements AsRef<Path> and &str does: https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.join

key_management/src/keystore.rs
```diff
@@ -113,127 +117,387 @@ pub trait KeyStore {
...
+        ciphertext
+    }
+
+    fn decrypt(
+        encryption_key: Arc<secretbox::Key>,
+        msg: &[u8],
+    ) -> Result<Vec<u8>, EncryptedKeyStoreError> {
+        let ciphertext = &msg[..msg.len() - 24];
+
+        let nonce = match secretbox::Nonce::from_slice(&msg[msg.len() - 24..]) {
```
creativcoder, line 421: `ok_or` is perfect for these situations where we want to return a Result in None case.

key_management/src/keystore.rs
cryptoquick: Yeah, that's fair. I just figured it looked more consistent that way.

key_management/src/keystore.rs
connormullett: This makes it much cleaner (cause `?` op). Thanks for this.

key_management/src/keystore.rs
connormullett: Refactored.

key_management/src/keystore.rs
creativcoder: Yeah, i love how std lib APIs compose well together

creativcoder:
Awesome work! Let's get this merged

ec2 (approved):
LGTM. It does seem like if you use encrypted mode, you cant access your unencrypted keys. I wonder if we want that functionality in the future...

Issue #1023, opened 2021-02-19
Implement optimal message selection


amerameen:
Currently only greedy message selection exists for the Message Pool. Add optimal message selection.

Optimal message selection
- As a miner you want to collect as much gas as possible (so you pack the block with as much gas as possible). Add an optimal message selection algorithm.
- Greedy message selection description: @ec2 will fill in.

Notes
- Lotus has a ticket quality parameter that is set through the RPC, which allows miners to select how much optimization they want for message selection

amerameen:
Confirming that this is done @creativcoder

PR #1086, opened 2021-04-26
Implement optimal message selection

### LINKED ISSUE #1054: Implement a correct chain abstraction for optimal message selection
**Task summary**

This was a result of investigating the failing tests in #1023 (PR: ##1037)
We need to re-write the MsgChain and MsgChainNode structures to mimic closely how lotus uses it.
Lotus's message pool implementation uses a array of msgChain nodes which provides both array and list semantics.

It's not entirely clear if it's possible to approach it 1:1 because of Rust's borrow check rules. What we have been exploring is the slotmap crate.

More details to follow soon on this issue.

**Specification reference**

- 

creativcoder:
**Summary of changes**
- Closes #1054 and #1023 

Closes 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3591,13 +3591,13 @@ checksum = "3576a87f2ba00f6f106fdfcd16db1d698d648a26ad8e0573cad8537c3c362d2a"
 
 [[package]]
 name = "lexical-core"
-version = "0.7.4"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "db65c6da02e61f55dae90a0ae427b2a5f6b3e8db09f58d10efab23af92592616"
+checksum = "6607c62aa161d23d17a9072cc5da0be67cdfc89d3afb1e8d9c842bebc2525ffe"
 dependencies = [
  "arrayvec",
  "bitflags 1.2.1",
- "cfg-if 0.1.10",
+ "cfg-if 1.0.0",
  "ryu",
  "static_assertions",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3591,13 +3591,13 @@ checksum = "3576a87f2ba00f6f106fdfcd16db1d698d648a26ad8e0573cad8537c3c362d2a"
 
 [[package]]
 name = "lexical-core"
-version = "0.7.5"
+version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "21f866863575d0e1d654fbeeabdc927292fdf862873dc3c96c6f753357e13374"
+checksum = "db65c6da02e61f55dae90a0ae427b2a5f6b3e8db09f58d10efab23af92592616"
 dependencies = [
  "arrayvec",
  "bitflags 1.2.1",
- "cfg-if 1.0.0",
+ "cfg-if 0.1.10",
  "ryu",
  "static_assertions",
 ]
@@ -4186,8 +4186,9 @@ dependencies = [
  "networks",
  "num-rational 0.3.2",
  "num-traits",
- "rand 0.7.3",
+ "rand 0.8.3",
  "serde",
+ "slotmap",
  "state_manager",
  "state_tree",
  "statrs",
@@ -5216,9 +5217,9 @@ dependencies = [
 
 [[package]]
 name = "rand"
-version = "0.8.1"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c24fcd450d3fa2b592732565aa4f17a27a61c65ece4726353e000939b0edee34"
+checksum = "0ef9e7e66b4468674bfcb0c81af8b7fa0bb154fa9f28eb840da5c447baeb8d7e"
 dependencies = [
  "libc",
  "rand_chacha 0.3.0",
@@ -5953,6 +5954,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "slotmap"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ab3003725ae562cf995f3dc82bb99e70926e09000396816765bb6d7adbe740b1"
+dependencies = [
+ "version_check",
+]
+
+[[package]]
 name = "sluice"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -54,26 +72,374 @@ where
...
+            .read()
+            .await
+            .chain_compute_base_fee(&target_tipset)?;
+
+        // 0. Load messages from the target tipset; if it is the same as the current tipset in
+        //    the mpool, then this is just the pending messages
+        let mut pending = self.get_pending_messages(&cur_ts, &target_tipset).await?;
+
+        if pending.is_empty() {
+            // FIXME: lotus returns a nil, shouldn't we return an Err?
```
creativcoder, line 111: @ec2 should we?

ec2: No. If theres no pending messages, then we select no messages. Makes sense to return empty vec.

cryptoquick (commented):
slotmap looks really neat! However, I'm afraid I'm not able to provide review without further context.

blockchain/message_pool/src/msg_chain.rs
```diff
@@ -19,23 +260,50 @@ pub(crate) struct MsgChainNode {
...
+                    .cmp(&other.gas_reward)
+                    .eq(&Ordering::Greater)
+        {
+            return Ordering::Greater;
+        }
+
+        Ordering::Less
+    }
+
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
creativcoder: Right, that's gone

blockchain/message_pool/src/msg_chain.rs
```diff
@@ -1,15 +1,256 @@
...
+}
+
+/// Chains is an abstraction of a list of message chain nodes.
+/// It wraps a slotmap instance. key_vec is an additional requirement in order to satisfy
+/// optimal msg selection use cases, such as iteration in insertion order.
+/// The slotamap serves as a lookup table for nodes to get around the borrow checker rules.
+/// Each MsgChainNode contains only pointers as `NodeKey` to the entries in the map
+/// With this design, we get around the borrow checker rule issues when
+/// implementing the optimal selection algorithm.
+pub struct Chains {
```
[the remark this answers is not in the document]
creativcoder: done

blockchain/message_pool/src/msgpool/selection.rs
creativcoder: Alright!

Cargo.lock
```diff
@@ -3590,13 +3590,13 @@ checksum = "3576a87f2ba00f6f106fdfcd16db1d698d648a26ad8e0573cad8537c3c362d2a"
 
 [[package]]
 name = "lexical-core"
-version = "0.7.5"
+version = "0.7.4"
```
ec2, line 3593: run `cargo update -p lexical-core:0.7.4` and this should fix the nightly build check.

ec2 (approved):
LGTM aside from the clippy issues and nightly build check.

cryptoquick (commented):
Thank you for the HackMD primer you made on [Optimal Message Selection](https://hackmd.io/@tbdrqGmwSXiPjxgteK3hMg/rkJmTfrvu#). It was very useful to understand the context of this PR.

I don't have specific comments on this PR, but I would like to ask a few general questions.

Do you have an idea of how many messages will be in the SlotMap each time?

Because the SlotMap is frequently cleared, I wonder if we shouldn't use the HopSlotMap, with a predefined (rough estimate) of capacity, using `with_capacity_and_key`, and then using the `clear` method to explicitly clear its contents.

Also, I noticed you removed the comment "TODO: Implement guess gas". Is that implemented now, say, via chain message? If so, are we checking against the new `MIN_GAS` constant anywhere the new, dynamic `min_gas` value should be used instead?

Finally, do you have any test cases? How have you observed that this new message selection algorithm works?

creativcoder:
> Thank you for the HackMD primer you made on [Optimal Message Selection](https://hackmd.io/@tbdrqGmwSXiPjxgteK3hMg/rkJmTfrvu#). It was very useful to understand the context of this PR.
> 
> I don't have specific comments on this PR, but I would like to ask a few general questions.
> 
> Do you have an idea of how many messages will be in the SlotMap each time?

Glad you found it helpful Hunter. Um, @ec2 might have a better answer on that. When I ran the tests for optimal selection (3 of them) I get about 1437 messages picked out of the selection method. It will usually be variable though.

> 
> Because the SlotMap is frequently cleared, I wonder if we shouldn't use the HopSlotMap, with a predefined (rough estimate) of capacity, using `with_capacity_and_key`, and then using the `clear` method to explicitly clear its contents.

We don't need to delete anything from the map anytime during the algorithm, it gets de-allocated at end of the method. We just remove, invalidate pointers to them.
> 
> Also, I noticed you removed the comment "TODO: Implement guess gas". Is that implemented now, say, via chain message? If so, are we checking against the new `MIN_GAS` constant anywhere the new, dynamic `min_gas` value should be used instead?

So gas_guess is actually referring to a module from [lotus impl](https://github.com/filecoin-project/lotus/blob/master/chain/messagepool/gasguess/guessgas.go). As of now we only use two constants from it and we don't have the rest of entities being used anywhere in forest. Yeah, it's better to put the TODO comment back, actually at the top where MIN_GAS constant is defined.

> 
> Finally, do you have any test cases? How have you observed that this new message selection algorithm works?

Yes, `selection.rs` contains 3 test cases for optimal selection. Starts here: [selection.rs](https://github.com/ChainSafe/forest/pull/1086/files#diff-a783b11c9b7dbd811a7cbf71bc863d5bf68a34cfb6768c0c513dc82484b766b4R1020) They are direct ports of lotus' [selection_test.go](https://github.com/filecoin-project/lotus/blob/master/chain/messagepool/selection_test.go)

cryptoquick (approved):
I feel comfortable with those answers! Fantastic work!

I think we should get this in before we merge #1033 due to our changes to how the keystore is created (Memory, Persistent, Encrypted). We'll handle those changes when we update our branch.

cryptoquick:
Also, don't forget to make the linter happy!

https://app.circleci.com/pipelines/github/ChainSafe/forest/4132/workflows/6168d5f9-bf19-40ad-a466-55f4d693c3c4/jobs/15493/parallel-runs/0/steps/0-111

Cargo.lock
creativcoder: I cannot update to 0.7.4, it always updates to 0.7.6, also crates.io page lists 0.7.4 in gray: https://crates.io/crates/lexical-core/versions Can we go with 0.7.6?

ec2: Well the issue is that you're on 0.7.4 right now. `cargo update -p lexical-core:0.7.4` upgrades it to the newest version according to semvar. So yeah, the point of that command IS to update to 0.7.6.

creativcoder: Ah I see. It's fixed.

PR #1089, opened 2021-05-03
Remove tide-websockets-sink fork

cryptoquick:
**Summary of changes**
- Refactor to not need tide-websockets-sink fork.

N/A

https://github.com/rust-lang/wg-async-foundations/pull/97#issuecomment-808399598


node/rpc/src/rpc_ws_handler.rs
```diff
@@ -129,8 +122,8 @@ where
     let (authorization_header, request) = get_auth_header(request);
     let rpc_server = request.state();
     let socket_active = Arc::new(AtomicCell::new(true));
-    let (ws_sender, mut ws_receiver) = ws_stream.split();
-    let ws_sender = Arc::new(Mutex::new(ws_sender));
+    let mut ws_receiver = ws_stream.clone();
```
jbr, line 125: I think this is one more clone than absolutely necessary, in that this makes three where I _think_ two would suffice. clone is fairly cheap, but either the sender or the receiver could just be the same allocation as the `ws_stream`

cryptoquick: Oh! Good point, thank you!

ec2 (approved):

node/rpc/src/rpc_ws_handler.rs
cryptoquick: Fixed!

creativcoder (approved):

PR #1101, opened 2021-05-17
Implement network version 12 state migration / actors v4 migration

### LINKED ISSUE #1039: Implement state migration / network version 12
**AC**
- [ ] Implement changes that would implement state migration using the existing framework for state migrations

**Notes**
Currently we have a naive implementation of state migration, which makes it impractical to use in production. Research how to improve (processing HAMT & AMT nodes concurrently is a good place to start)

creativcoder:
**Summary of changes**
- Implements state migration scaffold in vm/state_migration crate.
- Implements network version 12 / actors v4 migration.
- *Update*: nv12 state migration takes around 13-15 secs.

Closes #1039 


vm/interpreter/src/vm.rs
```diff
@@ -20,9 +20,10 @@ use forest_encoding::Cbor;
 use ipld_blockstore::BlockStore;
 use log::debug;
 use message::{ChainMessage, Message, MessageReceipt, UnsignedMessage};
-use networks::{UPGRADE_CLAUS_HEIGHT, UPGRADE_PLACEHOLDER_HEIGHT};
+use networks::{UPGRADE_ACTORS_V4_HEIGHT, UPGRADE_CLAUS_HEIGHT};
```
cryptoquick, line 23: Not directly related to this PR, but is this spelled correctly?

cryptoquick (commented):
Very cool to see more state migration work! Just one small comment, I can't comment on the code with what I know.

vm/interpreter/src/vm.rs
creativcoder: Um, not sure

ec2: Which word? I don't see any misspellings...

cryptoquick: `UPGRADE_CLAUS_HEIGHT` ?

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+use std::rc::Rc;
+use vm::ActorState;
+use vm::TokenAmount;
+
+pub mod nv12;
+
+pub type MigrationResult<T> = Result<T, MigrationErr>;
+
+#[derive(thiserror::Error, Debug)]
+pub enum MigrationErr {
```
olibearo, line 21: `MigrationErr` -> `MigrationError`, `Err` is a `go` convention

vm/interpreter/src/vm.rs
cryptoquick: what is CLAUS?

ec2: Santa Claus lmao. It was an emergency upgrade that happened a few days before christmas last year

cryptoquick: oh
I thought, is this supposed to be CLASS? CLAUSE? was not expecting the name of a deity

ec2 (approved):

olibearo (changes requested):

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+//! Common code that's shared across all migration code.
+//! Each network upgrade / state migration code lives in their own module.
+
+use address::Address;
+use cid::Cid;
+use clock::ChainEpoch;
+use ipld_blockstore::BlockStore;
+use std::error::Error as StdError;
+use std::rc::Rc;
+use vm::{ActorState, TokenAmount};
```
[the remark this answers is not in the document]
creativcoder: Do we have a quorum on what order, we want things in?
```
use std::
use std::

use other_crate::
use some_crate::

use crate::
use crate::
```
like that?

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+
+pub mod nv12;
+
+pub type MigrationResult<T> = Result<T, MigrationError>;
+
+#[derive(thiserror::Error, Debug)]
+pub enum MigrationError {
+    #[error("Failed creating job for state migration")]
+    MigrationJobCreate(Box<dyn StdError>),
```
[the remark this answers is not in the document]
ec2: Ah yeah, i agree. Forgot that we agreed that new code we do should try out best not to use dyn errors.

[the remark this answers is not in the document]
creativcoder: Yeah, i do feel uneasy with using Box<dyn Error>, but i added them because not adding them is not propagating the underlying error source, thereby providing less context to end user.

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
ec2: @olibearo suggested it, so following his lead on this one :D

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
olibearo: Just wrap the underlying error type: https://github.com/ChainSafe/forest/blob/main/blockchain/chain_sync/src/chain_muxer.rs#L57

That way the caller can also handle different error types separately when needed.

vm/state_migration/src/nv12/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+                    .ok_or(MigrationError::MigratorNotFound(state.code))?,
+            };
+
+            jobs.push(async move { next_input.run(store, prior_epoch) });
+
+            Ok(())
+        })
+        .map_err(MigrationError::MigrationJobCreate)?;
+
+    task::block_on(async {
```
olibearo, line 116: As far as I can tell, none of this needs to be async. `MigrationJob::run` is synchronous and so is `ActorMigration::MigrateState`. I would just accumulate `MigrationJob` in the vector and in the while loop just call `jobs.next().run()`. 

Let me know if I missed some aspect of this.

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
creativcoder: Yes, could have easily done that. But, the underlying error type itself is erased and is also `Box<dyn Error>`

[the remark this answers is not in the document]
creativcoder: for example: for_each method of StateTree returns an erased error type:

```rust
    pub fn for_each<F>(&self, mut f: F) -> Result<(), Box<dyn StdError>>
    where
        F: FnMut(Address, &ActorState) -> Result<(), Box<dyn StdError>>,
        S: BlockStore,
    {
        self.hamt.for_each(|k, v| f(Address::from_bytes(&k.0)?, v))
    }
```

vm/state_migration/src/nv12/mod.rs
olibearo: I suppose you want the blocking migrations to be distributed across the threads in the async executor. 

I'm a bit naive on how state migrations get run, could you explain that bit to me?

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
olibearo: We probably should standardize, but I personally don't have a preference. Here is the order I used in my previous PR:

https://github.com/ChainSafe/forest/blob/bd90aa65a885225ac84eb874898ca7dee430d3e1/blockchain/chain_sync/src/chain_muxer.rs

vm/state_migration/src/nv12/mod.rs
creativcoder: Yeah, i feel the async thing is not of any help here, since all underlying APIs are blocking. Yes, we wanted to distribute each job to run concurrently, (maybe use rayon as an alternative.). But from what i gathered from @ec2 on this, is that the actor state should get updated one after the other but not concurrently. is that correct @ec2 ?

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
olibearo: I see. Seems like we need to address this in our core libraries before it makes sense to enforce across the board.

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
creativcoder: Yeah, i'll just follow the same then

vm/state_migration/src/nv12/mod.rs
olibearo: As it stands, the actor states will get updated concurrently. In this specific migration, all migrations are nil migrations except one, so distributing it across threads isn't much help anyways. 

Depending on when this is run, the current async setup might not be getting in the way of other tasks either. So, it's probably best to run the functions in a way that communicates the most information to the reader with respect to how these functions fit into the the rest of the logic on the node.

ec2: So what i had said before was that the StateTree itself can't be used in concurrent settings. You can definitely do everything else concurrently. Its only the set_actor call that probably wont work in a concurrent setting. Concurrency will help even if there is only one migration. In this example, the only migration is for the miner actor. But there are a lot of miner actors.

olibearo: So I've advanced my understanding here with @ec2 :

I was wrong that the concurrency isn't useful here because this routine will run one `MinerMigrator` per miner, and the `block_on` will schedule that task on the Runtime's main thread pool. Because of that, this runs the risk of exhausting the Runtime's main thread pool. A better design would be to schedule this on a separate thread pool for blocking tasks.

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
olibearo: That would be nice

vm/state_migration/src/lib.rs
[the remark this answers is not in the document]
ec2: Makes sense. Everything that implements the Error trait i think has `.to_string()`. Perhaps we can map all Box<dyn Error> to Strings until we change the underlying libraries to return Error types? Does that sound reasonable?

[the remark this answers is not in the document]
creativcoder: Yeah, can do that for the time being

vm/state_migration/src/nv12/mod.rs
creativcoder: Yeah, i missed the part that there will be multiple miners that needs to run their migrator. To replace `block_on`, we can go with `task::spawn`

creativcoder: I guess we could spawn each job with task::spawn and collect and await on their join handles in `migrate_state_tree`.

olibearo: The main suggestion is to make sure the jobs are scheduled on a thread pool for blocking tasks. Because there are many miners, there will be many blocking tasks which will otherwise exhaust the primary async_std runtime thread pool and block the entire process.

ipld/blockstore/src/buffered.rs
```diff
@@ -7,17 +7,18 @@ use super::BlockStore;
...
-use std::{cell::RefCell, convert::TryFrom, io::Cursor};
+use std::{convert::TryFrom, io::Cursor};
 
 /// Wrapper around `BlockStore` to limit and have control over when values are written.
 /// This type is not threadsafe and can only be used in synchronous contexts.
 #[derive(Debug)]
 pub struct BufferedBlockStore<'bs, BS> {
     base: &'bs BS,
-    write: RefCell<HashMap<Cid, Vec<u8>>>,
+    write: std::sync::Mutex<HashMap<Cid, Vec<u8>>>,
```
ec2, line 21: Any reason for not using a lock free map?

creativcoder: Replaced with DashMap

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+        mut actors_out: StateTree<BS>,
+    ) -> MigrationResult<Cid> {
+        if self.migrations.len() + self.deferred_code_ids.len() != ACTORS_COUNT {
+            return Err(MigrationError::IncompleteMigrationSpec(
+                self.migrations.len(),
+            ));
+        }
+
+        let cpus = num_cpus::get();
+        let chan_size = 2;
```
ec2, line 129: really? this was the optimal size? i feel like this should also be a fucntion of num cpus, no?

vm/state_migration/src/nv12/miner.rs
```diff
@@ -0,0 +1,68 @@
...
+use actor_interface::actorv3::miner::State as V3State;
+use actor_interface::actorv4::miner::State as V4State;
+use async_std::sync::Arc;
+use cid::Cid;
+use cid::Code::Blake2b256;
+use ipld_blockstore::BlockStore;
+
+pub struct MinerMigrator(Cid);
+
+pub fn miner_migrator_v4<BS: BlockStore + Send + Sync>(
```
ec2, line 21:
```suggestion
pub fn miner_migrator_v12<BS: BlockStore + Send + Sync>(
```

vm/state_migration/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+forest_bitfield = "0.1"
+num-bigint = { version = "0.1.1", package = "forest_bigint", features = ["json"] }
+forest_hash_utils = "0.1"
+forest_json_utils = "0.1.1"
+thiserror = "1.0"
+futures = "0.3.14"
+state_tree = { path = "../state_tree" }
+async-std = "1.9.0"
+actor_interface = { path = "../actor_interface" }
+crossbeam-utils = "0.8.4"
```
ec2, line 26: we still need this?

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+                    let migrator = self.migrations.get(&state.code).cloned().unwrap();
+                    scope.spawn(move |_| {
+                        let job = MigrationJob {
+                            address,
+                            actor_state: state,
+                            actor_migration: migrator,
+                        };
+
+                        let job_output = job.run(store_clone, prior_epoch).unwrap_or_else(|e| {
+                            panic!(
```
ec2, line 173: is it okay to panic here?... i dont expect there be a good way to handle a state migration failure, but i wonder if we can do this more cleanly

vm/interpreter/src/vm.rs
```diff
@@ -198,6 +207,7 @@ where
         messages: &[BlockMessages],
         parent_epoch: ChainEpoch,
         epoch: ChainEpoch,
+        arc_store: std::sync::Arc<impl BlockStore + Send + Sync>,
```
ec2, line 210: cant you get this from `self`?

vm/state_migration/src/lib.rs
creativcoder: Ah no this should be 4 (will do the change). Anything greater than that and i see performance degrade. With 4 we're seeing 13-15 secs.

vm/state_migration/src/nv12/miner.rs
creativcoder: it's an actor, so shouldn't it reflect the actor version?

vm/state_migration/Cargo.toml
creativcoder: good catch, definitely not!

vm/state_migration/src/nv12/miner.rs
ec2: mmm, yeah thats true

vm/state_migration/src/lib.rs
creativcoder: Yeah, had thought about  it, I could have created a global error variable outside and assign any `Err`s and could bail early from the spawned thread. But i guess it is safe to panic here as it's a child thread and not proceed with any further migration state change. Totally open if we have better error handling ideas.

creativcoder: I think we need to ask whether is it safe for the migration to continue if any of the spawned thread fails, or we should discard the whole migration. If it's the later case, then yeah we can bail early with an `Err`. Thoughts?

vm/interpreter/src/vm.rs
creativcoder: No AFAIR, i wasn't possible. Since vm stores DB by ref, I'd have to change it to `Arc<DB>` and that cascaded changes all the way to HAMT structures

ec2: oh yeah gotcha

vm/state_migration/src/lib.rs
ec2: Then would you say the optimal is num_cpus/2?

creativcoder: mm, right. Let's do that

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    Other,
+}
+
+pub struct StateMigration<BS> {
+    migrations: HashMap<Cid, Migrator<BS>>,
+    deferred_code_ids: HashSet<Cid>,
+}
+
+impl<BS: BlockStore + Send + Sync> Default for StateMigration<BS> {
+    fn default() -> Self {
```
ec2, line 60: I wouldnt implement default on this.

ec2 (changes requested):

vm/state_migration/src/lib.rs
creativcoder: Yeah, I had `new` implemented but clippy asked me to do it. Can we disable that lint?

cryptoquick: Either that or we could just default to default. I don't mind default over new, even though it's a little more verbose sometimes.

vm/state_migration/src/lib.rs
cryptoquick: If the ideal is 4, why not do `std::cmp::min(4, num_cpus::get() - 1);` ? Subtracting one core so the system doesn't lock up.

cryptoquick (changes requested):
Looks pretty good, but I have just one suggestion.

vm/state_migration/src/lib.rs
creativcoder: I guess it doesn't make sense to have `nil_migrator` with the `default` method. I'll keep only a `new` method. Also, migrations must be initialized in the `migrations` map specific to a given migration and not here (for this it can be in `nv12/mod.rs`). At present, we are initializing it with specific actor v3 -> v4, which doesn't make it a generic `new` method.

ec2 (approved):

cryptoquick (approved):
Looks like my feedback wasn't terribly applicable. Everything looks good.

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    pub fn migrate_state_tree(
+        &mut self,
+        store: Arc<BS>,
+        prior_epoch: ChainEpoch,
+        actors_in: StateTree<BS>,
+        mut actors_out: StateTree<BS>,
+    ) -> MigrationResult<Cid> {
+        if self.migrations.len() + self.deferred_code_ids.len() != ACTORS_COUNT {
+            return Err(MigrationError::IncompleteMigrationSpec(
```
olibearo, line 78: :+1:

cryptoquick:
@creativcoder @ec2 Can I get a link to the v12 snapshot you have, btw? I just compiled this, I wanted to see how it runs. Also, don't forget to merge main.

cryptoquick:
n/m, got the link in slack. For others:
https://fil-chain-snapshots-fallback.s3-us-west-2.amazonaws.com/mainnet/minimal_finality_stateroots_711940_2021-04-29_03-00-00.car

vm/interpreter/src/vm.rs
```diff
@@ -172,16 +174,23 @@ where
...
-                let new_state: Cid = prev_state;
+                let new_state = run_nv12_migration(arc_store, prev_state, epoch)?;
                 if new_state != prev_state {
+                    log::info!(
+                        "actors_v4 state migration successful, took: {}ms",
+                        start.elapsed().as_millis()
+                    );
                     Ok(Some(new_state))
                 } else {
                     Ok(None)
```
olibearo, line 196: Curious why this doesn't result in an error

vm/interpreter/src/vm.rs
```diff
@@ -598,6 +608,31 @@ where
     }
 }
 
+// Performs network version 12 / actors v4 state migration
+fn run_nv12_migration(
+    arc_store: std::sync::Arc<impl BlockStore + Send + Sync>,
```
olibearo, line 613: _nit_: Including `arc` in the name is redundant with the type signature

vm/interpreter/src/vm.rs
olibearo: Seems like the state roots _must not_ match after the migration

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+                    .unwrap_or_else(|e| {
+                        panic!(
+                            "Failed setting new actor state at given address: {}, Reason: {}",
+                            address, e
+                        )
+                    });
+            }
+        });
+
+        let root_cid = actors_out
```
olibearo, line 158: You can remove the assignment and the `;` to just return

vm/state_migration/src/nv12/miner.rs
```diff
@@ -0,0 +1,68 @@
...
+// each actor's state migration is read from blockstore, changes state tree, and writes back to the blocstore.
+impl<BS: BlockStore + Send + Sync> ActorMigration<BS> for MinerMigrator {
+    fn migrate_state(
+        &self,
+        store: Arc<BS>,
+        input: ActorMigrationInput,
+    ) -> MigrationResult<MigrationOutput> {
+        let v3_state: Option<V3State> = store
+            .get(&input.head)
+            .map_err(|e| MigrationError::BlockStoreRead(e.to_string()))?;
```
olibearo, line 36: I think you can chain the `ok_or_else` to the `map_err` if you add a `?` after `map_err`

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+        mut actors_out: StateTree<BS>,
+    ) -> MigrationResult<Cid> {
+        if self.migrations.len() + self.deferred_code_ids.len() != ACTORS_COUNT {
+            return Err(MigrationError::IncompleteMigrationSpec(
+                self.migrations.len(),
+            ));
+        }
+
+        let cpus = num_cpus::get();
+        let chan_size = cpus / 2;
```
olibearo, line 84: I want to make the case that `chan_size` should be set to `cpus` but I don't know why you chose `cpus / 2`, could you add a comment?

vm/state_migration/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        let (state_tx, state_rx) = crossbeam_channel::bounded(chan_size);
+        let (job_tx, job_rx) = crossbeam_channel::bounded(chan_size);
+
+        pool.scope(|s| {
+            let store_clone = store.clone();
+
+            s.spawn(move |_| {
+                actors_in
+                    .for_each(|addr, state| {
```
olibearo, line 106: It would be nice if `StateTree` just implemented `Iterator` instead of this `for_each` method which returns a `Result<(), E>`. That way you could implement your own error handling, or lack thereof since this lambda isn't fallible.

olibearo (changes requested):
Rahul, I left a few questions and comments. I didn't run the migration myself to ensure the node stays in sync with the network post migration; I  assume you did this and saw that Hunter also was running this check.

Thanks for the great workl.

vm/state_migration/src/lib.rs
ec2: I think it would require quite a bit of refactoring to get this to implement Iterator sadly

vm/state_migration/src/lib.rs
ec2: I think this was empirically calculated based on Rahuls cloud machine...

vm/state_migration/src/lib.rs
olibearo: Just speculating, but I think it would just require adding a `.iter()` or `.into_iter()` before the `for_each` and changing the `for_each` to `try_for_each`.

olibearo: Everywhere that currently uses the custom `for_each` and wants to continue iterating with a fallible closure.

cryptoquick:
I verified the code in this PR works as expected. :+1: 

Screenshot from 2021-06-15 19-29-32

vm/state_migration/src/lib.rs
creativcoder: I wanted to keep it explicit on what is being returned for readability.

vm/state_migration/src/lib.rs
creativcoder: As i was testing the migration run-times, i noticed that increasing channel size made the migration time jump by 12-13 secs. 
For example (from my raw test logs):
```
16 threads 
16 chan size
16 outchan - run-time 80 secs
```
OTOH decreasing the channel size to less than half the cpu core count increased the run-time too. I am not exactly sure why it increases in the first case. We can file an issue to tune it in a later PR.

vm/state_migration/src/lib.rs
creativcoder: StateTree doesn't implement `Iterator`. If it implements that, the Hamt type would also need to implement that, where the Node type contained within would also need to implement that. So it's probably a non-trivial change and better tackled as a separate PR. I'll add a TODO for that.

vm/interpreter/src/vm.rs
creativcoder: Right, refactored to return an error now

creativcoder:
hey @cryptoquick thanks for sharing the pain of testing state migration lol. Must have taken you about 1.5 hours!

cryptoquick:
@creativcoder Dude, it took me closer to 10 hours, lol. I'm on a dual core laptop while my rig is down. I'm not sure if the anomalous behavior I mentioned in Slack was unusual just because of my limited system resources. If I had to guess, yeah, probably!

I was glad to do it, though! I try to help with PRs when I can.

olibearo (approved):
Great contribution, LGTM

PR #1102, opened 2021-05-17
Metrics

### LINKED ISSUE #1096: Create Prometheus Server & Grafana Stack
**Task summary**

Create a container stack that manages a Prometheus server and serves the Grafana webapp

**Specification reference**

The Prometheus server should collect metrics from the Forest process. The Grafana webapp should preload dashboards and come with documentation to update the dashboards.

### LINKED ISSUE #1097: Create Initial Set of Metrics on Forest
**Task summary**

Add first set of metrics to the Forest process and establish the patterns for defining, collecting, and updating metrics in Rust.

**Specification reference**

- Histogram to measure tipset processing time in ms
- Counter of GossipSub messages received by type
- Counter of tipsets received through GossipSub which fail validation
- Gauge of RocksDB size on disk
- Count of TipsetRangeSyncer future errors

olibearo:
**Summary of changes**

- Adds the initial set of metrics for `forest` 
- Provides a simple way to set up the metrics stack (`prometheus` & `grafana`) using Docker

Closes #1096 & #1097  


cryptoquick:
Nice! This looks like it measures block time and a few other things. How would you recommend I run it?

cryptoquick:
Okay, and this works without prometheus running, correct?

olibearo:
@cryptoquick that's correct

cryptoquick (changes requested):
Got an error when trying to run this:
`
thread 'async-std/runtime' panicked at 'called `Result::unwrap()` on an `Err` value: InconsistentCardinality { expect: 7, got: 1 }', /home/hunter/.cargo/registry/src/github.com-1ecc6299db9ec823/prometheus-0.12.0/src/vec.rs:258:49
`
Screenshot from 2021-05-19 19-02-34

cryptoquick:
Also, you're missing the following metric as defined in the issue ACs:
- Gauge of RocksDB size on disk

Maybe try measuring this after every flush:
https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html#tymethod.size

olibearo:
@cryptoquick I didn't realize that moving this from draft to open would auto request review, that's my bad. I still have much to do, including the items you mentioned.  I don't see a button that will move this from open -> draft again.

cryptoquick:
Oh, gotcha! Be sure to let us know when it's RFR. Also LMK if you think I could help in some way!

.circleci/config.yml
```diff
@@ -271,7 +271,7 @@ jobs:
       - run:
           name: Check for known security issues in dependencies
           # TODO remove ignore when https://github.com/filecoin-project/rust-fil-proofs/issues/1403 resolved
-          command: cargo audit --ignore RUSTSEC-2021-0011
+          command: cargo audit --ignore RUSTSEC-2021-0011 --ignore RUSTSEC-2021-0070
```
cryptoquick, line 274: Can you leave a comment like the one on 273?
Also, be sure to label that 273 is for `RUSTSEC-2021-0011`, while the new comment is for `RUSTSEC-2021-0070`, so we know which is which.

cryptoquick (changes requested):

ec2:
i know this isnt R4R yet, but you've incorrectly grouped Libp2p metrics (eg HELLO_REQUEST)  as Gossip metrics.

blockchain/chain_sync/src/chain_muxer.rs
```diff
@@ -140,7 +146,7 @@ pub struct ChainMuxer<DB, TBeacon, V, M> {
     /// Will mark any invalid blocks and all childen as bad in this bounded cache
     bad_blocks: Arc<BadBlockCache>,
 
-    /// Incoming network events to be handled by syncer
+    ///   network events to be handled by syncer
```
cryptoquick, line 149: Typo

cryptoquick (changes requested):
Just found one obvious typo.

I'm able to compile this and it runs fine, but I'm not able to test it on my laptop because, although I have docker configured on my desktop, I'm reluctant to do similar on my laptop, so my feedback on this will be somewhat limited for now.

ec2:
It LGTM. Im just setting up Docker on BigChungus right now to see this in live action

cryptoquick (approved):
The feedback I've provided has been addressed. @ec2, be sure to let us know how your Docker deploy works out.

ec2 (approved):
Just tested out with the new metrics. Looks perfect!

forest/src/cli/mod.rs
```diff
@@ -107,6 +107,8 @@ pub struct DaemonOpts {
         help = "Amount of Peers we want to be connected to (default is 75)"
     )]
     pub target_peer_count: Option<u32>,
+    #[structopt(long, help = "Encrypt the keystore (default = true)")]
+    pub encrypt_keystore: Option<bool>,
```
creativcoder, line 111: are these changes part of this PR?

creativcoder (approved):
Looks great! The only thing I felt is if `Metrics` could be global so we wouldn't have to pass it around everywhere. Alternatively, it could be a trait with default methods, which structs could then implement.

I tried running docker compose on aws machine but it fails. Posted in detail on slack.

forest/src/cli/mod.rs
olibearo: Yes

olibearo:
> Looks great! The only thing I felt is if `Metrics` could be global so we wouldn't have to pass it around everywhere. Alternatively, it could be a trait with default methods, which structs could then implement.
> 
> I tried running docker compose on aws machine but it fails. Posted in detail on slack.

This was an issue with the `docker` installation on Rahul's machine. I've updated the documentation to make installation easier.

blockchain/chain_sync/src/metrics.rs
```diff
@@ -0,0 +1,135 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use lazy_static::lazy_static;
+use prometheus::{
+    core::{AtomicU64, GenericCounter, GenericCounterVec, GenericGauge, Opts},
+    Histogram, HistogramOpts,
+};
+
+lazy_static! {
```
creativcoder, line 10: Kudos for making this a singleton! One more suggestion, why don't we try once_cell: https://github.com/async-rs/async-std/issues/406 as people speak highly of it compared to lazy_static.

olibearo: Yeah, I looked into it, but the stdlib impl is only available in nightly and despite that, it's a container type that would require every routine that needs the `OnceCell<T>`  to call `get` to retrieve the inner `T`.

olibearo: I'm all ears if you understand the benefits we might get from adding that getter every time we collect a metric.

ec2: From my understanding, the benefits would be:
- Lack of macros
- Faster hot get
I am not sure if that is enough for us to switch over at this point for this.

olibearo: Why is it faster than accessing a global static variable?

Is the concern with macros the incremental increase in compilation cost?

creativcoder: oh, you don't need to use the OnceCell container. Instead you can use: https://docs.rs/once_cell/1.8.0/once_cell/#lazy-initialized-global-data. Anyways if you feel the refactor is quite complex, then lazy_static it is! As for perf improvements, we will have to measure to see the benefits in context to forest's access patterns.

ec2: Not sure why its faster on hot get. That's just what someone said. But i dont think its significant. 
The concern with macros is increased compilation cost as well as the fact that there is code being generated. That being said, those are "meh" concerns for me... I'm ok with lazy_static. We use lazy_static in other parts of the codebase

olibearo: I don’t feel enticed, seems more fashion than function

PR #1103, opened 2021-05-18
Resolve Stack Overflow

### LINKED ISSUE #869: Stack overflow issue
**Describe the bug**

I've now hit the stack overflow three times when syncing/validating chains

Error looks like:
```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
```

and isn't easily reproducible/consistent. I am currently hitting it when importing and validation a chain through `forest_bench` and simply restarting allows it to continue. I am not quite sure what could be taking up the stack space yet, as for each transition, the stack should be reset aside from caches during the transition increasing.

I am hitting this using Rust 1.48.0 stable

Edit: This might just be a machine specific issue, since the stack size is based on system limits, but will keep this issue open until it's verified exactly how it's hit (because aborting is an unacceptable effect)

ec2:
**Summary of changes**
- Adds a custom global Rayon threadpool so we can easily identify where panics or stack overflows happen in the future for tasks spawned in the global Rayon threadpool
- We have a minimum on the length that Rayon will split when doing parallel iteration, so that Rayon wont split into too many chunks, which is what was causing the stack overflow.

Closes 
#869

### DIFF
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -35,6 +35,11 @@ pub(super) async fn start(config: Config) {
         .set(address::Network::Mainnet)
         .unwrap();
 
+    // Configure the Rayon Threadpool
+    rayon::ThreadPoolBuilder::new().thread_name(|n| {
+        format!("rayon-thread-{}", n)
+    }).build_global().unwrap();
+
     info!("Starting Forest daemon");
     let net_keypair = get_keypair(&format!("{}{}", &config.data_dir, "/libp2p/keypair"))
         .unwrap_or_else(|| {
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -41,7 +41,7 @@ pub(super) async fn start(config: Config) {
     rayon::ThreadPoolBuilder::new()
         .thread_name(|n| format!("rayon-thread-{}", n))
         .build_global()
-        .unwrap();
+        .expect("Failed to create global Rayon ThreadPool");
 
     info!("Starting Forest daemon");
     let net_keypair = get_keypair(&format!("{}{}", &config.data_dir, "/libp2p/keypair"))
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -38,9 +38,10 @@ pub(super) async fn start(config: Config) {
         .unwrap();
 
     // Configure the Rayon Threadpool
-    rayon::ThreadPoolBuilder::new().thread_name(|n| {
-        format!("rayon-thread-{}", n)
-    }).build_global().unwrap();
+    rayon::ThreadPoolBuilder::new()
+        .thread_name(|n| format!("rayon-thread-{}", n))
+        .build_global()
+        .unwrap();
 
     info!("Starting Forest daemon");
     let net_keypair = get_keypair(&format!("{}{}", &config.data_dir, "/libp2p/keypair"))
--- a/vm/interpreter/src/default_runtime.rs
+++ b/vm/interpreter/src/default_runtime.rs
@@ -987,17 +987,15 @@ where
         dbg!(vis.len());
         let avg = std::sync::atomic::AtomicUsize::new(0);
         let max = std::sync::atomic::AtomicUsize::new(0);
-        let min = std::sync::atomic::AtomicUsize::new(0);
+        let min = std::sync::atomic::AtomicUsize::new(1);
         let out = vis
             .par_iter()
             .map(|(&addr, seals)| {
-                dbg!(seals.len());
                 avg.fetch_add(seals.len(), std::sync::atomic::Ordering::Relaxed);
                 max.fetch_max(seals.len(), std::sync::atomic::Ordering::Relaxed);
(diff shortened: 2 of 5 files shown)


vm/interpreter/src/default_runtime.rs
```diff
@@ -985,26 +985,35 @@ where
...
+        let cpus = num_cpus::get();
+        // Create a local threadpool just for running verify seal in this epoch
+        let batch_pool = rayon::ThreadPoolBuilder::new()
+            .num_threads(cpus)
+            .thread_name(|id| format!("batch_verify_seal_{}", id))
+            .build()?;
+
+        let out = batch_pool.scope(|_scope| {
+            vis.par_iter()
+                .with_min_len(cpus)
```
creativcoder, line 997: Should this be `min_len` or `max_len` ?

forest/src/daemon.rs
```diff
@@ -37,6 +37,12 @@ pub(super) async fn start(config: Config) {
         .set(address::Network::Mainnet)
         .unwrap();
 
+    // Configure the Rayon Threadpool
+    rayon::ThreadPoolBuilder::new()
+        .thread_name(|n| format!("rayon-thread-{}", n))
+        .build_global()
+        .unwrap();
```
olibearo, line 44: Bringing this up in anticipation of our style guide: When do you prefer `unwrap` over `expect`?

vm/interpreter/src/default_runtime.rs
```diff
@@ -985,26 +985,35 @@ where
...
-                            false
-                        } else {
-                            true
-                        }
-                    })
-                    .collect();
-                (addr, results)
-            })
-            .collect();
+        let cpus = num_cpus::get();
```
olibearo, line 988: Consider using `lazy_static` to evaluate this function once and set a constant, module level variable

creativcoder: or [once_cell](https://github.com/matklad/once_cell)

vm/interpreter/src/default_runtime.rs
```diff
@@ -985,26 +985,35 @@ where
...
+        // Create a local threadpool just for running verify seal in this epoch
+        let batch_pool = rayon::ThreadPoolBuilder::new()
```
olibearo, line 990: Is the Runtime object a singleton in the process? If so, I would create the threadpool in the constructor and store the object as a property on the Runtime.

vm/interpreter/src/default_runtime.rs
olibearo: `min_len` sets a floor on the chunk size which avoids scheduling too many blocking tasks across the threads in this blocking thread pool. Max would set a ceiling on the chunk size, which would have the opposite effect. We determined that number of chunks in the outer loop and inner loop is the source of the stack overflow.

ec2: Mhm, thats a better heuristic, actually.

forest/src/daemon.rs
ec2: Good point. I think expect is much more appropriate here

vm/interpreter/src/default_runtime.rs
creativcoder: I see, thanks for elaborating on that. Do we then not intend to use a bound on the inner `par_iter` as well?

vm/interpreter/src/default_runtime.rs
ec2: Mmm, like that could be done, but its kind of annoying because this Runtime trait is shared between all versions of actors. And thus we would need to change the interface, release a breaking version of the runtime, then release like 3 different version of actors too. I can add a todo and an associated issue for doing this in v5?

vm/interpreter/src/default_runtime.rs
ec2: No. The average len of the inner par_iter is 1.

vm/interpreter/src/default_runtime.rs
olibearo: Why does this change the interface? The threadpool should be internal to the Runtime object when you instantiate it in the constructor.

ec2: Oh i misunderstood what you were saying. Let me try that out. It should work

ec2: Actually, that wont really optimize anything. In fact, i think that adds even more overhead. A `Runtime` is instantiated every time we make a `Interpreter::send()` call. So if we were to make a threadpool in the `Runtime` constructor, we will be making it even when we don't use it (which will be 99%) of the time.  
This is because batch_verify_seals only gets called once at the end of the epoch by the cron actor. So currently we create a new threadpool at every epoch. What you are suggesting will create a threadpool for every transaction sent.

olibearo: Gotcha, this is why I was wondering if it is a singleton. Thanks for checking

olibearo (approved):
Nice and trim!

creativcoder (approved):

cryptoquick (approved):
Nice! I prefer once_cell, but I suppose it's fine if we use both.

PR #1122, opened 2021-05-31
RPC API w/ Permissions handling

### LINKED ISSUE #1026: RPC API w/ Permissions
**ACs**
- [ ] We are able to build a new program that imports the interface and is able to make calls to the RPC node
- [ ] RPC server & client make use of this interface

**Notes**
- Would be helpful to have name of method, the serializer for request, deserializer for response, request handler in one place
- Currently we build the server from a hardcoded builder command
- A crate for interacting with Filecoin (i.e. Forest & Lotus RPC servers) 
- This would include a refactor of RPC methods in RPC router
- This would be the first step in testing our RPC programatically
- This would be like a web3.js in Rust for Filecoin

cryptoquick:
**Summary of changes**
- [x] Unified RPC API abstraction with method name, params type, result type, and access map, with permissions matched against lotus, and used in both RPC client and server
  - [x] Auth API
  - [x] Beacon API
  - [x] Chain API
  - [x] Message Pool API
  - [x] Sync API
  - [x] Wallet API
  - [x] State API
  - [x] Gas API
  - [x] Common API
  - [x] Net API
- [x] Used in rpc-client and rpc server
- [x] Checks permissions against JWT claims
- [x] Prints JWT token to console

Closes #1026 

N/A


forest/src/daemon.rs
```diff
@@ -102,6 +103,12 @@ pub(super) async fn start(config: Config) {
         ks.put(JWT_IDENTIFIER.to_owned(), generate_priv_key())
             .unwrap();
     }
+
+    // Print admin token
+    let ki = ks.get(JWT_IDENTIFIER).unwrap();
+    let token = create_token(ADMIN.to_owned(), ki.private_key()).unwrap();
+    println!("Admin token: {}", token);
```
ec2, line 110: Not understanding why we need to create a new JWT upon launch.

node/rpc-api/src/data_types.rs
```diff
@@ -0,0 +1,215 @@
...
+// RPC State
+#[derive(Serialize)]
+pub struct StreamingData<'a> {
+    pub json_rpc: &'a str,
+    pub method: &'a str,
+    pub params: SubscriptionHeadChange,
+}
+
+/// This is where you store persistent data, or at least access to stateful data.
+pub struct RPCState<DB, B>
```
ec2, line 46:
```suggestion
pub struct RpcState<DB, B>
```

node/rpc-api/src/data_types.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub bad_blocks: Arc<BadBlockCache>,
+    pub sync_state: Arc<RwLock<SyncState>>,
+    pub network_send: Sender<NetworkMessage>,
+    pub new_mined_block_tx: Sender<Arc<Tipset>>,
+    pub network_name: String,
+    pub beacon: Arc<BeaconSchedule<B>>,
+}
+
+#[derive(Serialize)]
+pub struct RPCSyncState {
```
ec2, line 64:
```suggestion
pub struct RpcSyncState {
```

forest/src/daemon.rs
cryptoquick: It's not a new key, it's just a token derived from the existing key. Because I've added rich permissioning for every endpoint, there's no way to call `auth create-token` without providing an admin JWT. If you're using an encrypted keystore, this then becomes impossible to use.

node/rpc-api/src/data_types.rs
cryptoquick: I actually changed this in this PR to the way I prefer it. We've disabled that linter rule anyway.

node/rpc-api/src/data_types.rs
cryptoquick: I actually changed this in this PR to the way I prefer it. We've disabled that linter rule anyway.

ec2 (approved):
I really like this. Makes things in the RPC easier to follow. And can open up interesting composability patterns between different subsystems in Forest (long shot though, just a little thought)

cryptoquick:
Thanks! That's an interesting thought! Since our RPC interface is transport-agnostic, if I understand your suggestion, we could use the RPC client and server within the forest daemon, operating without the HTTP transport, and instead with a shared-memory RPC router operating as an "In-Proc" service...

Is that your suggestion? If so, I'm not sure I can recommend that, since there'd still be significant overhead from JSON serialization.

It sounds similar to ZeroMQ's `inproc` transport. It's not quite like IPC since we use threads and shared memory, thus avoiding serialization overhead. IPC is likely somewhat rare in Rust, I suspect.

node/rpc-client/src/chain_ops.rs
```diff
@@ -2,32 +2,32 @@
...
 use cid::{json::CidJson, Cid};
+use rpc_api::chain_api::*;
+
 use jsonrpc_v2::Error as JsonRpcError;
-use message::unsigned_message::json::UnsignedMessageJson;
 
 /// Returns a block with specified CID fom chain via RPC
-pub async fn block(cid: Cid) -> Result<BlockHeaderJson, JsonRpcError> {
-    filecoin_rpc::chain_get_block(CidJson(cid)).await
+pub async fn block(cid: Cid) -> Result<ChainGetBlockResult, JsonRpcError> {
```
cryptoquick, line 11: Refactor:
`Result<ChainGetBlockResult, JsonRpcError>` -> `ChainGetBlockResult`, and elsewhere:

```rust
pub type RPCResult<T> = Result<T, JsonRpcError>;
pub async fn block(cid: Cid) -> RPCResult<ChainGetBlockResult> {
    ...
}
```

node/rpc-client/src/client.rs
```diff
@@ -169,42 +168,42 @@ where
 {
     let rpc_req = jsonrpc_v2::RequestObject::request()
         .with_method(method_name)
-        .with_params(serde_json::to_value(vec![params])?)
+        .with_params(serde_json::to_value(params)?)
         .finish();
 
     call(rpc_req).await.map_err(|e| e)
 }
```
cryptoquick, line 175: Combine `call` with `call_params`, and rename to `call_method`.

node/rpc/src/lib.rs
```diff
@@ -51,155 +54,114 @@ where
         Server::new()
             .with_data(Data(state))
             // Auth API
-            .with_method("Filecoin.AuthNew", auth_new::<DB, B>)
-            .with_method("Filecoin.AuthVerify", auth_verify::<DB, B>)
+            .with_method(AUTH_NEW, auth_new::<DB, B>)
```
cryptoquick, line 57: Investigate removing the trait generic here.

connormullett (commented):
The issues addressed in the 1:1:1 with Rahul seem like a good start. Will approve once they are resolved. Otherwise, looks great!

connormullett (approved):
Despite some cleanup and minor code quality changes, it is functional and readable and gets the job done great. Approved!

PR #1125, opened 2021-06-02
Fix P2P random walk logic 

Fraccaman:
**Summary of changes**

In the p2p behaviour, when executing random query, the next execution should be triggered with an exponential delay capped at 60 seconds (as per comment). 
Instead, the actual implementation gets triggered every 10 seconds. This PR follows uses `next_kad_random_query` instead of `Duration::from_secs(10)` to schedule the next trigger.


ec2 (approved):
Nice catch! Thanks for the contribution @Fraccaman :) If you could just run `cargo fmt` on this, I can merge it. Or else I can run it and push onto this branch and merge.

cryptoquick (approved):
LGTM

PR #1132, opened 2021-06-08
Add rust-toolchain file

### LINKED ISSUE #1131: add a `rust-toolchain` file to set a specific version
**Issue summary**
for now, codes in main branch can not be compiled with stable (1.51.0, in my case) toolchain.
it could be better to add a `rust-toolchain` file to a specific  version.
[Managing versions](https://doc.rust-lang.org/edition-guide/rust-2018/rustup-for-managing-rust-versions.html#managing-versions)

this may also avoid issues caused by differences in the compilers, 
[A Polkadot Postmortem - 24.05.2021](https://polkadot.network/a-polkadot-postmortem-24-05-2021/)

cryptoquick:
**Summary of changes**

- Adds rust-toolchain file pointing to rust `1.52.1`
  - We're just using a new Rust string method, it's mostly just so our code is cleaner and modern, is all

Closes #1131 

In case anyone's curious:
- `split_once` was added in `1.52`: https://doc.rust-lang.org/std/primitive.str.html#method.split_once
- It's used here: https://github.com/ChainSafe/forest/blob/6f9edae88add3e02c57925c297017fe2a9a9fd23/node/rpc-client/src/client.rs#L120

Also note, our CI is already configured to use `1.52.1`


ec2 (approved):

cryptoquick:
Also, to review this, just use `cargo check`, it will result in slightly different behavior as it downloads the toolchain specific to this project.

olibearo (approved):

PR #1149, opened 2021-06-09
[Audit fixes] FOR-02: Inconsistent Deserialization of Address ID

### LINKED ISSUE #1134: FOR-02: Inconsistent Deserialisation of Address

cryptoquick:
**Summary of changes**
- Compares serialized leb128 bytes to deserialized bytes.
- Adds a passing test case, in addition to 2 test cases for encodings meant to fail, as recommended in the audit document.
- These changes should not alter behavior in release build; it will only make the node behavior more consistent with that of the Lotus node.

Closes #1134 

- Explanation of changes: `from_leb_128` is too permissive of how leb128 bytes are deserialized, resulting in inconsistent behavior from how Lotus deserializes bytes. Lotus throws an error if there byte encodings are inconsistent.

### DIFF
--- a/vm/address/Cargo.toml
+++ b/vm/address/Cargo.toml
@@ -13,6 +13,7 @@ num-derive = "0.3.0"
 data-encoding = "2.1.2"
 data-encoding-macro = "0.1.7"
 leb128 = "0.2.1"
+log = "0.4.8"
 encoding = { package = "forest_encoding", version = "0.2.1" }
 thiserror = "1.0"
 serde = { version = "1.0", features = ["derive"] }
--- a/vm/address/src/lib.rs
+++ b/vm/address/src/lib.rs
@@ -14,7 +14,6 @@ use data_encoding::Encoding;
 #[allow(unused_imports)]
 use data_encoding_macro::{internal_new_encoding, new_encoding};
 use encoding::{blake2b_variable, serde_bytes, Cbor};
-use log::error;
 use once_cell::sync::OnceCell;
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 use std::hash::Hash;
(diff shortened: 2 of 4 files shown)


vm/address/src/lib.rs
```diff
@@ -303,7 +304,69 @@ pub(crate) fn from_leb_bytes(bz: &[u8]) -> Result<u64, Error> {
     let mut readable = bz;
 
     // write id to buffer in leb128 format
-    Ok(leb128::read::unsigned(&mut readable)?)
+    let id = leb128::read::unsigned(&mut readable)?;
+
+    if to_leb_bytes(id)? == bz {
+        Ok(id)
+    } else {
+        error!(
```
ec2, line 312: Probably makes sense to have this be part of the InvalidAddressIDPayload variant instead of printing it out

creativcoder (approved):
:+1:

vm/address/src/lib.rs
cryptoquick: Like, maybe using a tuple in the enum variant to capture the id?

cryptoquick: Take a look at these changes. Is this what you meant?
https://github.com/ChainSafe/forest/pull/1149/commits/2e5f084c5628bff42be67441aae92d563065fcb3

vm/address/Cargo.toml
```diff
@@ -13,6 +13,7 @@ num-derive = "0.3.0"
 data-encoding = "2.1.2"
 data-encoding-macro = "0.1.7"
 leb128 = "0.2.1"
+log = "0.4.8"
```
ec2, line 16: Think we can remove this now.

ec2 (approved):

PR #1151, opened 2021-06-09
[Audit fixes] FOR-16: Unnecessary Extensive Permissions for Private Keys

### LINKED ISSUE #1148: FOR-16: Unnecessary Extensive Permissions for Private Keys

cryptoquick:
**Summary of changes**
- Restrict to only user permissions both libp2p keypair and forest keystore files

Closes #1148 

Makes full use of the outstanding Unix permission implementation so helpfully recommended by our simply delightful auditors... Capital! Most capital.
https://github.com/sigp/lighthouse/pull/551/files

### DIFF
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -56,10 +56,10 @@ pub(super) async fn start(config: Config) {
                 &format!("{}{}", &config.data_dir, "/libp2p/"),
                 "keypair",
             ) {
-                Ok(_) => {
+                Ok(file) => {
                     // Restrict permissions on files containing private keys
-                    #[cfg(linux)]
-                    utils::set_user_perm(file)?;
+                    #[cfg(unix)]
+                    utils::set_user_perm(&file).expect("Set user perms on unix systems");
                 }
                 Err(e) => {
                     info!("Could not write keystore to disk!");
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -51,13 +51,20 @@ pub(super) async fn start(config: Config) {
             let gen_keypair = ed25519::Keypair::generate();
             // Save Ed25519 keypair to file
             // TODO rename old file to keypair.old(?)
-            if let Err(e) = write_to_file(
+            match write_to_file(
                 &gen_keypair.encode(),
                 &format!("{}{}", &config.data_dir, "/libp2p/"),
                 "keypair",
             ) {
-                info!("Could not write keystore to disk!");
-                trace!("Error {:?}", e);
+                Ok(_) => {
+                    // Restrict permissions on files containing private keys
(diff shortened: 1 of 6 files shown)


ec2 (approved):

node/utils/src/lib.rs
```diff
@@ -6,15 +6,30 @@ use std::fs::{create_dir_all, File};
...
 use std::path::Path;
 
+/// Restricts permissions on a file to user-only: 0600
+#[cfg(unix)]
+pub fn set_user_perm(file: &File) -> Result<()> {
+    use log::info;
+    use std::os::unix::fs::PermissionsExt;
+
+    let mut perm = file.metadata()?.permissions();
+    perm.set_mode((libc::S_IWUSR | libc::S_IRUSR) as u32);
```
creativcoder, line 16: Do you think `0o600` would be better here?

forest/src/daemon.rs
```diff
@@ -51,13 +51,20 @@ pub(super) async fn start(config: Config) {
             let gen_keypair = ed25519::Keypair::generate();
             // Save Ed25519 keypair to file
             // TODO rename old file to keypair.old(?)
-            if let Err(e) = write_to_file(
+            match write_to_file(
                 &gen_keypair.encode(),
                 &format!("{}{}", &config.data_dir, "/libp2p/"),
```
creativcoder, line 56: probably better to use Path::join() with just "libp2p" ?

node/utils/src/lib.rs
```diff
@@ -6,15 +6,30 @@ use std::fs::{create_dir_all, File};
...
+    Ok(())
+}
+
 /// Writes a string to a specified file. Creates the desired path if it does not exist.
 /// Note: `path` and `filename` are appended to produce the resulting file path.
-pub fn write_to_file(message: &[u8], path: &str, file_name: &str) -> Result<()> {
+pub fn write_to_file(message: &[u8], path: &str, file_name: &str) -> Result<File> {
     // Create path if it doesn't exist
     create_dir_all(Path::new(path))?;
     let join = format!("{}{}", path, file_name);
```
creativcoder, line 29: with Path::join being passed, this can also be joined with `join()`

node/utils/src/lib.rs
connormullett: I think using the mask like C style #defines works fine here. Write | Read looks better than `600` from a readability perspective.

forest/src/daemon.rs
cryptoquick: Good idea!

ec2 (approved):

creativcoder (approved):
Great, let's tackle the PathBuf thing later!

PR #1154, opened 2021-06-10
Disable nightly CI

cryptoquick:
**Summary of changes**
- Nightly CI builds are broken and causing us to have a lot of :x: which makes me sad. Let's disable that for now, and maybe if someone can find time to spend on CI, it can be corrected.

N/A

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,9 +289,9 @@ workflows:
   build-check:
     jobs:
       - prefetch-crates
-      - nightly-build-check:
-          requires:
-            - prefetch-crates
+      # - nightly-build-check:
+      #     requires:
+      #       - prefetch-crates
       - msrv-build-check:
           requires:
             - prefetch-crates


.circleci/config.yml
```diff
@@ -289,9 +289,9 @@ workflows:
   build-check:
     jobs:
       - prefetch-crates
-      - nightly-build-check:
-          requires:
-            - prefetch-crates
+      # - nightly-build-check:
```
connormullett, line 292: Any reason to leave these commented out, or should we remove it entirely? Will we potentially turn this back on?

connormullett (commented):
Single question

.circleci/config.yml
cryptoquick: I think so, I just don't have time to look into it right now, and it might be something that'll get sorted out eventually. Maybe we could have someone experienced in CI to look into it. I know we have a few on staff. @GregTheGreek ?

connormullett: Well it passes in the meantime, so I guess we can go ahead and stick it through.

connormullett (approved):

.circleci/config.yml
cryptoquick: It might also be nice to fix MacOS builds too, you know?

connormullett: True, maybe they can be lumped up into a single issue?

cryptoquick:
@connormullett GitHub is being fucky with resolved conversations. Anyway, yes, I can create a tracking issue for this, to address later.

.circleci/config.yml
olibearo: The underlying problem is that a dependency is using features which are deprecated in nightly. @ec2 tried to phase out the offending crate, but it turned out to be a longer task than he expected

PR #1164, opened 2021-06-14
LumberjackDB (WIP)

### LINKED ISSUE #1124: Research Spike: LumberjackDB
**Task summary**

An Append-Only ChainStore could reduce how much data is stored within the RocksDB key-value store.

The idea is to append all data to a log file, similar to a Write-Ahead Log (WAL), but intended to be a persistent data store. That's why I call it an Append-Only Immutable Log (AOIL), and since it's used to back a ChainStore, it'd be an Append-Only ChainStore, or AOCS.

The AOILs would need to be split into separate files, especially if we wanted more compact 32-bit offsets, and not all file systems support files many terabytes in size. We would also need to be careful not to exceed an offset of 2^32 bytes: 4294967296, or 4GB. The last item could exceed that value, making the AOIL file slightly longer, as long as the offset does not exceed `u32::MAX`.

We'd use byte tuples to indicate to places within the files that we can call a `Loc`, like this:

```rust
let (offset, limit, file) :(u32, u32, u16) = loc;
```

This allows us to store up to (and maybe slightly over) 262 TB; `2^16 files * 2^32 offsets` using Loc tuples at a constant 10 bytes in size.

We could use separate AOIL files for keys and values. By separating them, keys could also be omitted from the K/V store, but this might have serious scaling concerns. However, if we needed to, this could be something that's possible, as long as the files are scanned separately. Each key file could hold up to `2^32 / 10` keys; 429,496,729 keys.

I'm not sure how state pruning might occur, but it'd likely require generating a new AOIL alongside the existing one. The Append-Only ChainStore's primary use would be just in optimizing our present use-case of snapshot imports, in addition to a full-node implementation with reduced write-amplification penalties.

Ideally this would offer an interface that's basically a drop-in replacement for a K/V store.

Memory-mapping might also provide some advantages here, but I'm unsure how well the address space would scale to the terabytes.

Sled might still be worth exploring for this use-case, since the index should be much smaller. Sled's write amplification begins to seriously impede its performance past 10GB.

cryptoquick:
**Summary of changes**
- Work on LumberjackDB
- Enabled by default only for testing in this PR, will change before making as RFR

Closes #1124 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -456,7 +456,7 @@ dependencies = [
  "async-io",
  "async-lock",
  "async-process",
- "crossbeam-utils 0.8.1",
+ "crossbeam-utils 0.8.5",
  "futures-channel",
  "futures-core",
  "futures-io",
@@ -1016,7 +1016,7 @@ dependencies = [
  "forest_car",
  "forest_cid",
  "forest_crypto",
- "forest_db",
+ "forest_db 0.1.0",
  "forest_encoding",
  "forest_ipld",
  "forest_message",
@@ -1057,7 +1057,7 @@ dependencies = [
  "forest_car",
  "forest_cid",
  "forest_crypto",
- "forest_db",
+ "forest_db 0.1.0",
  "forest_encoding",
  "forest_libp2p",
  "forest_message",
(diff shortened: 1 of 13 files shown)


q9f:
let's revisit this once we get around it again.

PR #1169, opened 2021-06-15
hotfix:metrics:change prometheus response type

detailyang:
**Summary of changes**
Change prometheus response body type to text/format (default is application/octet-stream)

More detail see https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-based-format

Signed-off-by: detailyang <detailyang@gmail.com>

Closes 

### DIFF
--- a/utils/metrics/src/lib.rs
+++ b/utils/metrics/src/lib.rs
@@ -40,6 +40,7 @@ async fn collect_metrics(req: tide::Request<Registry>) -> tide::Result {
         .encode(&metric_families, &mut metrics)
         .expect("Encoding Prometheus metrics must succeed.");
     Ok(tide::Response::builder(tide::StatusCode::Ok)
+        .header("content-type", "text/plain; version=0.0.4; charset=utf-8")
         .body(metrics)
         .build())
 }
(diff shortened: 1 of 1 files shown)


utils/metrics/src/lib.rs
```diff
@@ -40,6 +40,7 @@ async fn collect_metrics(req: tide::Request<Registry>) -> tide::Result {
         .encode(&metric_families, &mut metrics)
         .expect("Encoding Prometheus metrics must succeed.");
     Ok(tide::Response::builder(tide::StatusCode::Ok)
+        .header("content-type", "text/plain; version=0.0.4; charset=utf-8")
```
olibearo, line 43: Hey, thanks for the submission. Does the Prometheus server you are using to scrape forest require these headers?

detailyang: Actually, not sure it needs or not. But when you open the http://127.0.0.1:6116/metrics in the browser, you will get the download page instead of metrics because of the  `application/octet-stream` type.

Also, the specs say the following: 
> A missing version value will lead to a fall-back to the most recent text format version

cryptoquick (approved):
Never a bad idea to configure proper headers.

utils/metrics/src/lib.rs
olibearo: I see.

Can you update to use the appropriate constants exposed by the stdlib?
https://docs.rs/http/0.1.5/http/header/constant.CONTENT_TYPE.html

olibearo (changes requested):

detailyang:
Done. Sorry for late reply. I'm busying working for actorv5 interopnet to forest:)

cryptoquick:
@detailyang I like your fix!

olibearo (approved):

PR #1170, opened 2021-06-15
fix wallet verify

connormullett:
**Summary of changes**
- This is more of a bugfix. When refactoring the wallet cli when that PR was in review, a tuple construction was moved to another module to clean up the original file. The first two parameters of that tuple were strings and messed with the API. The address was actually a message, and a message was actually an address. All fixed now

Closes 


cryptoquick (approved):
Glad you figured that out!

ec2 (approved):
Nice find!

Issue #1166, opened 2021-06-14
[Forest CLI] Fix panic when passing a malformed address to `wallet set-default`


cryptoquick:
**Task summary**

`forest wallet set-default` works fine when providing a proper address, but if a malformed address is given (say, but adding junk to the end), `forest` quits with a panic.

**Specification reference**

- `forest wallet set-default <address>$ome3xtraJunk` no longer panics, but ideally returns a 400 (Bad Request).

**Other information and links**

https://github.com/ChainSafe/forest/pull/1128#pullrequestreview-683262538

PR #1172, opened 2021-06-16
`auth api-info`

### LINKED ISSUE #1117: `forest auth api-info`
**Task summary**

We need to have an api-info command like `lotus auth api-info`. This will help with documenting workflows. Templating the JWT and mutliaddress is not ideal.

ACs:

- [ ] `forest auth api-info` provides similar output to `lotus auth api-info`

**Specification reference**

> FULLNODE_API_INFO="/ip4/127.0.0.1/tcp/1234/http" lotus auth api-info --perm admin 

`FULLNODE_API_INFO=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.6dMalXxBGadVxVLzbxAanYV--7ihmIGK3eLrtjhEawQ:/ip4/127.0.0.1/tcp/1234/http`

connormullett:
**Summary of changes**
- Adds `auth api-info` command to forest CLI.
To test: 
- use command `FULLNODE_API_INFO="..." forest auth api-info -p <read/write/admin/sign>`
- copy/paste output as new `FULLNODE_API_INFO` and send another command. Verify that permissions are correct (cannot write with a read token for example)

also bundled in this PR is the fix for `set-default` panicking on a bad address

Closes #1117 and #1166


cryptoquick (changes requested):
Works great! Nice code, too.

I want to approve this, but there's an issue. What got me thinking about it is that lotus fails to run this with an error code 500, and also the lotus client fails to run this on a forest node; lotus segfaults. But that makes sense, actually, because after thinking about it, I'm not sure if this isn't just an unauthenticated `auth new`, since it requires only read permissions, and I looked into Lotus to see what they're using, and I was surprised, the command isn't even part of their RPC API:

https://github.com/filecoin-project/lotus/search?q=ApiInfo

So, after giving this some thought, I was thinking, we should add an additional param to the `auth api-info` command: basically, a positional or named argument for the admin JWT to use. I'm fine with supporting permissions params and generating a new JWT, but the admin JWT that's printed to the terminal when first starting the daemon should always be provided when running this command.

I'm fine with leaving this in our RPC API implementation, but I think we should call it "Forest.ApiInfo", since, as far as I understand, this is not something that's provided by Lotus's RPC API, and when I go to generate the OpenRPC schema, it'll show up as an additional API we support but they don't. I'll ignore all methods prefixed with Forest, on the other hand.

I think it's okay to have it in the RPC server since it's basically just string formatting at that point, which is the most any user should need to do.

Of course, seeing as to how it's implemented in lotus, there's no need to support lotus or test with it as part of this feature.

One more thing. With forest, you can set the -p, --port number separately from an API INFO string. I suggest that running the forest CLI with this top-level argument also produce a multiaddr with the corresponding port number. I would recommend you implement it in such a way that it would override the FULLNODE_API_INFO string on any command it's provided with.

I also recommend mentioning the following changes in the PR issue description, since it's separate from what's fixed here:

https://github.com/ChainSafe/forest/pull/1172/files#diff-4734af900a01b14a1a70ed35452444e19b505d4a51f4237f38a427fd769bc2ddR160-R167

cryptoquick (approved):
Approving this to unblock review since I requested changes earlier. I recommend others review this also, however, since I made some pretty significant changes to Connor's original PR.

ec2 (approved):

PR #1173, opened 2021-06-18
Update Runtime to Support V5 Actors

ec2:
**Summary of changes**
- Introduces 2 new methods to the Runtime Interface: `verify_aggregate_seals` and `base_fee`.
- Update the chain and beacon randomness functions in the Runtime to look forward after Hyperdrive and backwards before.
- Bump to v8 of proofs api to support aggregate seal verification.

TODO:

- [x]  Update MockRuntime
- [ ]  Update `networks` crate to set Hyperdrive height (currently using placeholder

Closes 


cryptoquick:
This still had todos and failing tests and linter. Is this RFR?

types/src/verifier/mod.rs
```diff
@@ -45,6 +53,70 @@ pub trait ProofVerifier {
...
+        aggregate: &AggregateSealVerifyProofAndInfos,
+    ) -> Result<(), Box<dyn StdError>> {
+        if aggregate.infos.is_empty() {
+            return Err("no seal verify infos".into());
+        }
+        let spt: proofs::RegisteredSealProof = aggregate.seal_proof.try_into()?;
+        let prover_id = prover_id_from_u64(aggregate.miner);
+
+        struct AggregationInputs {
+            commr: [u8; 32],
```
creativcoder, line 66: readability: better to expand `r` and `d` to `replica` and `data` respectively?

vm/interpreter/src/gas_tracker/price_list.rs
```diff
@@ -42,6 +43,52 @@ lazy_static! {
         hashing_base: 31355,
         compute_unsealed_sector_cid_base: 98647,
         verify_seal_base: 2000, // TODO revisit potential removal of this
```
creativcoder, line 45: unrelated, but can this be removed now?

vm/interpreter/src/gas_tracker/price_list.rs
```diff
@@ -42,6 +43,52 @@ lazy_static! {
...
+        verify_aggregate_seal_per: [
+            (
+                RegisteredSealProof::StackedDRG32GiBV1P1,
+                449900
+            ),
+            (
+                RegisteredSealProof::StackedDRG64GiBV1P1,
+                359272
+            )
+        ].iter().copied().collect(),
```
creativcoder, line 57: can we not simply do: `vec![ {..} {...}]` ?

vm/interpreter/src/gas_tracker/price_list.rs
```diff
@@ -322,6 +444,28 @@ impl PriceList {
     pub fn on_verify_seal(&self, _info: &SealVerifyInfo) -> GasCharge {
         GasCharge::new("OnVerifySeal", self.verify_seal_base, 0)
     }
+    #[inline]
+    pub fn on_verify_aggregate_seals(
+        &self,
+        aggregate: &AggregateSealVerifyProofAndInfos,
+    ) -> GasCharge {
```
creativcoder, line 451: If this is fallible, does it make sense to return a `Result` here?

vm/interpreter/src/gas_tracker/price_list.rs
```diff
@@ -322,6 +444,28 @@ impl PriceList {
...
+        let proof_type = aggregate.seal_proof;
+        let per_proof = self
+            .verify_aggregate_seal_per
+            .get(&proof_type)
+            .expect("There is an implementation error where proof type does not exist in table");
```
creativcoder, line 456: If this is fallible, does it make sense to return a Result here?

creativcoder (approved):
LGTM, left a few comments

detailyang:
@ec2 It looks good in my localhost. I merge this branch to my dev branch and now It can sync the whole latest actorv5 interopnet network:)

vm/interpreter/src/gas_tracker/price_list.rs
olibearo: A result doesn't make sense since we expect there to be a per proof price when verifying an aggregate. If we wanted to get rid of the `.expect(...)` though it would be nice if the look up function for the price used the compiled to return a price for each enum variant.

vm/interpreter/src/gas_tracker/price_list.rs
```diff
@@ -322,6 +444,28 @@ impl PriceList {
...
+        let step = self
+            .verify_aggregate_seal_steps
+            .get(&proof_type)
+            .expect("There is an implementation error where proof type does not exist in table");
```
olibearo, line 460: Same suggestion here as before. A `HashMap` is lacking the knowledge that there is a price for each variant of proof type.

vm/interpreter/src/gas_tracker/price_list.rs
```diff
@@ -322,6 +444,28 @@ impl PriceList {
...
+        // Should be safe because there is a limit to how much seals get aggregated
+        let num = aggregate.infos.len() as i64;
```
olibearo, line 462: Curious why we use `i64` for the `GasCharge` properties: `compute_gas` and `storage_gas`; can these be negative?

vm/interpreter/src/default_runtime.rs
```diff
@@ -611,10 +615,15 @@ where
         rand_epoch: ChainEpoch,
         entropy: &[u8],
     ) -> Result<Randomness, ActorError> {
-        let r = self
-            .rand
-            .get_chain_randomness(personalization, rand_epoch, entropy)
-            .map_err(|e| e.downcast_fatal("could not get randomness"))?;
+        let r = if rand_epoch > networks::UPGRADE_PLACEHOLDER_HEIGHT {
```
olibearo, line 618: The placeholder is currently set to `9_999_999`, when will this be set?

types/src/verifier/mod.rs
olibearo: Yeah, or add comments

types/src/verifier/mod.rs
```diff
@@ -45,6 +53,70 @@ pub trait ProofVerifier {
...
+        let seeds: Vec<[u8; 32]> = inputs.iter().map(|input| input.seed).collect();
+        if !verify_aggregate_seal_commit_proofs(
+            spt,
+            aggregate.aggregate_proof.try_into()?,
+            aggregate.proof.clone(),
+            &commrs,
+            &seeds,
+            inp,
+        )? {
+            Err("Invalid Aggregate Seal proof".into())
```
olibearo, line 115: Upsetting that this doesn't use a custom error type. Would it be worth it to create a ticket to define the error types here? Do you anticipate any complications will come up?

olibearo: This trait is fairly small.

types/src/sector/registered_proof.rs
```diff
@@ -283,6 +290,22 @@ i64_conversion! {
...
+    SnarkPackV1 => 0,
+}
+#[cfg(feature = "proofs")]
+impl TryFrom<RegisteredAggregateProof> for filecoin_proofs_api::RegisteredAggregationProof {
+    type Error = String;
+    fn try_from(p: RegisteredAggregateProof) -> Result<Self, Self::Error> {
+        use RegisteredAggregateProof::*;
+        match p {
+            SnarkPackV1 => Ok(Self::SnarkPackV1),
+            Invalid(i) => Err(format!("unsuppored aggregate proof type: {}", i)),
```
olibearo, line 304: unsupported*

olibearo (changes requested):
Generally looks good to me. I left a few questions and suggestions that I'd like to talk through before approving.

vm/interpreter/src/gas_tracker/price_list.rs
ec2: I left it there because it is still a comment in lotus

vm/interpreter/src/default_runtime.rs
ec2: Should have it set once we update the actors interface.

vm/interpreter/src/gas_tracker/price_list.rs
ec2: No this is a AHashMap

types/src/verifier/mod.rs
ec2: commr and commd are colloquially used terms, but ill add comments.

types/src/verifier/mod.rs
ec2: I dont see why its useful or important here. These methods are generally used in actors. And the way the error is handled is irrespective of what error happens here anyways. I think a refactor of error handling in this might mess up with error propagation in actors.

vm/interpreter/src/gas_tracker/price_list.rs
ec2: There exists operations that refund gas.

olibearo (approved):
Thanks Eric

PR #1194, opened 2021-07-15
OpenRPC schema parsing & generation

### LINKED ISSUE #1123: OpenRPC Schema Validation
**Task summary**

We need to validate our RPC API against the Lotus OpenRPC schema.

**Specification reference**

This will be implemented in a build script.

- [x] Parse Lotus OpenRPC
  - [x] Method name
  - [x] Params
  - [x] Result
- [x] Parse a `syn` syntax tree of the `rpc-api` crate
- [x] Print a report comparing the Forest API to Lotus's

An OpenRPC schema can be generated in the latest Lotus repo using the following commands:
- `make docsgen`
- `./docgen-openrpc "api/api_full.go" "FullNode" "api" "./api" > build/openrpc/full.json`

Currently the Lotus makefile gzips OpenRPC schemas by default, and I'm not sure how to extract the archive, it looks like an empty folder, but not an empty archive. Regardless, the last command works.

Generated OpenRPC JSON wil be committed to the repo since there's no easy way to get the definitions otherwise.

cryptoquick:
**Summary of changes**
- Build script that validates Lotus OpenRPC schema against Forest RPC API lib
- Creates a Markdown report in the root directory

Closes #1123 

Hopefully this is pretty useful:
https://github.com/ChainSafe/forest/blob/8f51cea9464e809f0768670b4fe7cf2a8df5476f/API_IMPLEMENTATION.md

We should make some of these types richer and add some type aliases at some point, but that's separate work.


node/rpc-api/build.rs
```diff
@@ -0,0 +1,522 @@
...
+                })
+                .collect::<Vec<String>>();
+
+            let forest_only_methods_list = forest_only_methods
+                .iter()
+                .map(|method| format!("- `{method}`", method = method))
+                .collect::<Vec<String>>();
+
+            let report = format!(
+                r####"# Forest API Implementation Report
```
connormullett, line 417: Maybe we can kick this out to a template text file? Seems out of place to be kept here as a magic string

node/rpc-api/static/full.json
```diff
@@ -0,0 +1,18680 @@
+{
```
connormullett, line 1: Do we have a way to pull this file from lotus? ie from github with a curl call? or is this just static and needs to be copy/pasted over

connormullett (changes requested):
1 nit and 1 question. Looks good otherwise! Love the parsing and AST logic

connormullett:
Make sure to take care of the clippy stuff too, looks like it failed

API_IMPLEMENTATION.md
```diff
@@ -0,0 +1,241 @@
...
+
+- Forest method count: 66
+- Lotus method count: 169
+- API coverage: 39.05%
+
+## Forest-only Methods
+
+These methods exist in Forest only and cannot be compared:
+
+- `Filecoin.AuthNew`
```
ec2, line 13: I'm pretty confused. I thought that these all exist in Lotus.... I mean we implemented all of these so that we can be compatible with Lotus 🤔

ec2: The only methods i can think of that would be in Forest and not in Lotus are methods prefixed with `Forest.` instead of `Filecoin.`

cryptoquick: Looking into it, I think there's a small bug, in that when I compare the method names, I need to make sure the case is consistent. But stuff like AuthNew, it's just not there in the OpenRPC file, no auth methods are. And that's what this is really comparing. It's just comparing their auto-generated API schema against ours.

https://github.com/ChainSafe/forest/blob/8f51cea9464e809f0768670b4fe7cf2a8df5476f/node/rpc-api/static/full.json

cryptoquick: An example of the inconsistent casing:
`ChainGetTipsetByHeight`
vs
`ChainGetTipSetByHeight`

Actually, I'm not sure I'm going to fix that. That's not a bug in my script, that's an actual API incompatibility. I think proper casing should matter. And this is the kind of thing that reveals that discrepancy.

node/rpc-api/static/full.json
cryptoquick: No, they don't have a generated version in their repo, and not only that, their default make command doesn't produce a proper JSON file, so I had to modify it. This can be generated via the commands detailed in the original issue.

node/rpc-api/build.rs
cryptoquick: I'm not sure the format macro would even work that way, would it?

cryptoquick:
> Make sure to take care of the clippy stuff too, looks like it failed

Haha, thanks for the heads-up.

node/rpc-api/build.rs
connormullett: That is a good question, I'm going to check.

node/rpc-api/static/full.json
connormullett: Gotcha, that makes sense. Would be great if we could fetch it with a script to execute with the RPC schema generation

node/rpc-api/build.rs
connormullett: Yup, disregard. The macro requires a string literal

node/rpc-api/static/full.json
cryptoquick: Yeah, I mean, maybe, but build.rs runs every time you build forest. Fortunately, Rust is so fast it's barely noticeable, lol.

node/rpc-api/build.rs
cryptoquick: Haha, macros continue to disappoint. Format, we still love you~

ec2 (approved):
I believe Auth commands are just not part of their OpenRPC stuff, but its definitely available on Lotus. But we can address those in another PR after talking with their team

connormullett (approved):

PR #1197, opened 2021-07-21
added the check for config file via Env Var

### LINKED ISSUE #1191: `FOREST_CONFIG_PATH` env var
**Task summary**

In the interest of a better UX, if users often have specific config options they often like to override by default, such as having the store on a different disk or disabling `encrypt_keystore`, it might be nice to create an environment variable that points to a configuration file that can be overridden by explicitly providing an explicit config argument.

**Acceptance criteria**

- `FOREST_CONFIG_PATH` env var is read when checking for config file path in forest CLI (for both client and daemon)
- `-c, --config` argument overrides this setting

It's probably best the check should go here:
https://github.com/ChainSafe/forest/blob/32656db9a28120326a4fdb328cb751e1705de761/forest/src/cli/mod.rs#L130

leviathanbeak:
**Summary of changes**
- improves UX by adding the check for config file via Env Var

Closes #1191 

### DIFF
--- a/forest/src/cli/mod.rs
+++ b/forest/src/cli/mod.rs
@@ -136,7 +136,17 @@ impl CLIOpts {
                 // Parse and return the configuration file
                 read_toml(&toml)?
             }
-            None => Config::default(),
+            None => {
+                // Check ENV VAR for config file
+                if let Ok(config_file) = std::env::var("FOREST_CONFIG_PATH") {
+                    // Read from config file
+                    let toml = read_file_to_string(&*config_file)?;
+                    // Parse and return the configuration file
+                    read_toml(&toml)?
+                } else {
+                    Config::default()
+                }
+            }
         };
         if let Some(genesis_file) = &self.genesis {
             cfg.genesis_file = Some(genesis_file.to_owned());


cryptoquick:
I'm very glad you decided to take this on! However, have you tested this? I'm not sure it works. This is what I get after running `make install`:

Screenshot from 2021-07-21 05-46-49

As you can see, it ignored my config setting to turn off keystore encryption. This occurs when running both as client and daemon.

Briefly looking at the code, I see no obvious reason why this doesn't work. If you can reproduce the issue, do you think you can figure out why the config file seems to be ignored, despite it being provided by the shell?

cryptoquick:
Wait, I just realized, after checking out your repo, you have your changes in a branch. Recompiling now.

cryptoquick:
It works perfectly! Good job!! Let's get this merged in, team, this is a substantial UX improvement (imo).

Screenshot from 2021-07-21 06-01-17

Screenshot from 2021-07-21 06-01-32

cryptoquick (approved):

ec2 (approved):

leviathanbeak:
@cryptoquick @ec2 thanks for reviewing, I went outside after PR so wasn't able to respond, cheers

PR #1199, opened 2021-07-24
 Remove loopback and duplicate addrs in `net peers` output

### LINKED ISSUE #1184: Cleanup `net peers` output
**Task summary**
`net peers` prints a lot of extraneous information that's either redundant or not useful.

**Acceptance criteria**

- [ ] `net peers` has no duplicate Multiaddrs
- [ ] `net peers` does not show localhost or loopback ip4 and ip6 routes

Compare the output of `lotus net peers`:
Screenshot from 2021-07-05 09-18-33

...with the output of `forest net peers`:
Screenshot from 2021-07-05 09-18-04

A lotta extra junk!

- Be sure to note: It's okay to have multiple routes, for example, we might have an external ip4 route and an external ip6 route to a peer, so this task isn't as simple as selecting a single route per peer, we still need to return the vector, even though the majority are a single external ip4 route.

FrancisMurillo:
**Summary of changes**
-  Remove loopback and duplicate addrs in `net peers` output
   - This does not filter `localhost` alias, onlyl loopback like `127.0.0.1`
   - Peers without connections are omitted

Closes #1184 

### DIFF
--- a/forest/src/cli/net_cmd.rs
+++ b/forest/src/cli/net_cmd.rs
@@ -3,6 +3,7 @@
 
 use forest_libp2p::{Multiaddr, Protocol};
 use rpc_api::data_types::AddrInfo;
+use std::collections::HashSet;
 use structopt::StructOpt;
 
 use crate::cli::cli_error_and_die;
@@ -50,12 +51,26 @@ impl NetCommands {
             Self::Peers => match net_peers(()).await {
                 Ok(addrs) => {
                     let output: Vec<String> = addrs
-                        .iter()
-                        .map(|info| {
-                            let addresses: Vec<String> =
-                                info.addrs.iter().map(|addr| addr.to_string()).collect();
+                        .into_iter()
+                        .filter_map(|info| {
+                            let addresses: Vec<String> = info
+                                .addrs
+                                .into_iter()
+                                .filter(|addr| match addr.iter().next().unwrap() {
+                                    Protocol::Ip4(ip_addr) => !ip_addr.is_loopback(),
+                                    Protocol::Ip6(ip_addr) => !ip_addr.is_loopback(),
+                                    _ => true,
+                                })
+                                .map(|addr| addr.to_string())
+                                .collect::<HashSet<_>>()
+                                .into_iter()
+                                .collect();
 
-                            format!("{}, [{}]", info.id, addresses.join(", "))
+                            if addresses.is_empty() {
+                                return None;
+                            }
+
+                            Some(format!("{}, [{}]", info.id, addresses.join(", ")))
                         })
                         .collect();
 


leviathanbeak:
it's up to the reviewers, personally I'd just collect to `HashSet<String>` then to `Vec<String>` in order to get unique addresses, instead of explicitly creating let `mut seen_addrs: HashSet<Multiaddr> = HashSet::default();`

cryptoquick (approved):
Very nice! I really like this approach, and it works well!

Compare:
Screenshot from 2021-07-26 12-42-17

With:
Screenshot from 2021-07-26 12-42-31

Againly, nicely done!

> it's up to the reviewers, personally I'd just collect to `HashSet<String>` then to `Vec<String>` in order to get unique addresses, instead of explicitly creating `let mut seen_addrs: HashSet<Multiaddr> = HashSet::default();`

@leviathanbeak I'm not sure I understand this feedback, do you think you could provide an example or maybe suggest a change?

connormullett:
> @leviathanbeak I'm not sure I understand this feedback, do you think you could provide an example or maybe suggest a change?

HashSets cannot contain duplicates. Personally I think that approach is much simpler then containing another iterable with seen addresses, so I would like to see if we can incorporate a `HashSet` with this approach. Nice job though!

connormullett (changes requested):
See previous comment on HashSet approach

cryptoquick:
> > @leviathanbeak I'm not sure I understand this feedback, do you think you could provide an example or maybe suggest a change?
> 
> HashSets cannot contain duplicates. Personally I think that approach is much simpler then containing another iterable with seen addresses, so I would like to see if we can incorporate a `HashSet`. Nice work!

Right, but isn't that the approach already being used?
https://github.com/ChainSafe/forest/pull/1199/files#diff-a4693340c77dcb6dcb5dd651478de1755657fb815fea64d1cc1cd4626f81002cR6

leviathanbeak:
@cryptoquick I meant something along these lines (I'm not at my PC)

```
let addresses: Vec<String> = info
                                .addrs
                                .into_iter()
                                .filter_map(|multi_addr| filter loopback addresses)
                                .collect::<HashSet<String>()
                                .into_iter()
                                .collect();
```

connormullett:
> @cryptoquick I meant something along these lines (I'm not at my PC)
> 
> ```
> let addresses: Vec<String> = info
>                                 .addrs
>                                 .into_iter()
>                                 .filter_map(|multi_addr| filter loopback addresses)
>                                 .collect::<HashSet<String>()
>                                 .into_iter()
>                                 .collect();
> ```

Yup, much cleaner.

cryptoquick:
> @cryptoquick I meant something along these lines (I'm not at my PC)
> 
> ```
> let addresses: Vec<String> = info
>                                 .addrs
>                                 .into_iter()
>                                 .filter_map(|multi_addr| filter loopback addresses)
>                                 .collect::<HashSet<String>()
>                                 .into_iter()
>                                 .collect();
> ```

Clever, I like it. @FrancisMurillo Care to make the changes?

FrancisMurillo:
> > @cryptoquick I meant something along these lines (I'm not at my PC)
> > ```
> > let addresses: Vec<String> = info
> >                                 .addrs
> >                                 .into_iter()
> >                                 .filter_map(|multi_addr| filter loopback addresses)
> >                                 .collect::<HashSet<String>()
> >                                 .into_iter()
> >                                 .collect();
> > ```
> 
> Clever, I like it. @FrancisMurillo Care to make the changes?

@cryptoquick Happy to make the changes. Just to clarify with `let mut seen_addrs`, there is no duplicate `Multiaddrs` globally or among all peers. This approach is take by `lotus` so that is where it came from. With `.collect::<HashSet<_>>()`, there is no duplicate in a peer but can occur in others. So is the filtering global or per peer?

cryptoquick:
> @cryptoquick Happy to make the changes. Just to clarify with `let mut seen_addrs`, there is no duplicate `Multiaddrs` globally or among all peers. This approach is taken by `lotus` so that is where it came from. With `.collect::<HashSet<_>>()`, there is no duplicate in a peer but can occur in others. So is the filtering global or per peer?

Ah, makes sense. Are you saying the "no duplicate Multiaddrs" portion of the AC isn't necessary if we filter for only external addresses? That makes sense, since, we shouldn't have duplicate addresses on IP, haha.

If I remove that requirement, could the code be simplified?

FrancisMurillo:
@connormullett Updated to use `.collect::<HashSet<_>>()`

cryptoquick:
Just reviewed the latest, still works as expected. Thanks for responding to the feedback so quickly, @FrancisMurillo. The code is definitely more succinct now.

@connormullett How's that look?

connormullett (approved):
Much cleaner, thanks for this work! Approved

PR #1201, opened 2021-08-05
Multi-key import feature

### LINKED ISSUE #1129: Support multi-key imports from file/stdin
The Forest CLI should be able to import multiple keys from a file and STDIN if a path flag is not present.

The command is `forest wallet import`, and it should behave similarly to how it does on lotus:

Screenshot from 2021-07-27 18-31-17

leviathanbeak:
**Summary of changes**
- instead of reading the key from the command argument, we either read the path to the key, or if there is no path provided we read it from stdin (the same way lotus does it)

Closes #1129 


cryptoquick (changes requested):
I really like it! Thanks so much for your continued involvement on the project.

I left a couple nits, but while you're at it, do you think you could make it so this error returns a code other than 0?
Screenshot from 2021-08-05 08-24-46

Do you think you could also make these work? I'm a little rusty with my bash shell, but I think this is how you do this, right?
Screenshot from 2021-08-05 08-48-21

Key import from path doesn't seem to decode correctly? Or maybe it's string trim, not sure. Either way, it doesn't work.
Screenshot from 2021-08-05 08-56-09

From a user experience standpoint, I think we should also make it also print out the key on a successful import so the user knows which key they just imported. Right now no message is returned other than exit code 0.

Also, do you think you could make it trim the string input? If you accidentally add extra space to stdin when copy/pasting, that might be nice.

In summary:
- [ ] Fix qualified std import nit
- [ ] Make key import work with stdin in addition to read_password (if possible)
- [ ] Fix path import decoding issue
- [ ] Fix "key already exists" error code so returns other than 0
- [ ] Print out key name after succesful import
- [ ] Trim string input

I know it's a lot more work than originally specified, but that'd really help us out. That's whole point, right? And thanks again for your contributions so far!

leviathanbeak:
@cryptoquick thanks for the detailed review, I didn't expect it to be that easy anyway, will improve it

cryptoquick:
Haha, great!

leviathanbeak:
> In summary:
> 
> * [x ]  Fix qualified std import nit
> * [ x]  Make key import work with stdin in addition to read_password (if possible)
fixed this by only updating `rpassword` lib, it was very, very outdated

> * [ x]  Fix path import decoding issue
added trimming - should fix all the problems
> * [ x]  Fix "key already exists" error code so returns other than 0
as discussed, the problem was in 2 places, fixed it

> * [ x]  Print out key name after succesful import
added `println!(key)`
> * [ x]  Trim string input

cryptoquick (approved):
Good catch on the outdated `rpassword` crate! Just tried it out, it all works great. Thank you so much for your solid contributions!

Also, as mentioned in `#fil-forest-dev` in the Slack, we decided the error code change isn't needed for this PR.

ec2 (approved):

PR #1228, opened 2021-09-27
rust: default to stable toolchain instead of pinned version

q9f:
**Summary of changes**
- ~~`.circleci/config.yml` msrv rust version `stable`~~
- `rust-toolchain` default rust version `stable`
- `Dockerfile` uses `1-buster` rust image
- `Dockerfile-ci` uses `1-buster` rust image

The rust compiler release cycle is a moving target, and we should _at least_ support the latest stable. 

Subsequently, we can set up pipelines for beta, to be alerted about changes coming downstream.


q9f:
I added stable build jobs to #1233 so you can see it works with stable both on macos and linux

ec2 (approved):

q9f:
This needs more work, I'll do it tomorrow.

.circleci/config.yml
```diff
@@ -241,7 +241,7 @@ jobs:
           command: make license
       - run:
           name: cargo clippy
-          command: cargo clippy -- -D warnings -A clippy::upper_case_acronyms
+          command: cargo clippy -- -D warnings -A clippy::upper_case_acronyms -A clippy::needless_collect -A clippy::enum_variant_names -A clippy::needless_borrow
```
q9f, line 244: Look at this @ec2 - this makes clippy pass on stable and instead of fixing these warnings in a single PR, I would like to remove these warnings in subsequent PRs, so you can better review the changes, and if needed, easier revert in case something goes south.

q9f:
This should be safe to merge now once CI is green. I will fix these warnings in separate pull requests.

PR #1253, opened 2021-11-05
v6 runtime mutability fix

laudiacay:
The problems with the runtime mutability situation are fixed by this, I believe.


noot (approved):
looks good, I'm still getting one compilation error that seems minor-ish but going to merge this into the base branch so we can work off that

```
error: this arithmetic operation will overflow
   --> vm/actor/src/builtin/miner/monies.rs:288:54
    |
288 |     static ref BATCH_BALANCER: BigInt = BigInt::from(5 * 1_000_000_000); // 5 * 1 nanoFIL
    |                                                      ^^^^^^^^^^^^^^^^^ attempt to compute `5_i32 * 1000000000_i32`, which would overflow
    |
    = note: `#[deny(arithmetic_overflow)]` on by default
```

PR #1294, opened 2021-11-30
Elmattic/actors review F4

### LINKED ISSUE #1268: F4: market::validate_deal checks properties out of order
Description of the issue [HERE](https://hackmd.io/eThp9-pyR0WfLJjBynFntA?view#F4-marketvalidate_deal-checks-properties-out-of-order)

elmattic:
**Summary of changes**
- Remove redundant code

Could cause a chain split.

Closes 

#1268

### DIFF
--- a/vm/actor/src/builtin/market/mod.rs
+++ b/vm/actor/src/builtin/market/mod.rs
@@ -1185,7 +1185,7 @@ where
     if proposal.end_epoch <= proposal.start_epoch {
         return Err(actor_error!(
             ErrIllegalArgument,
-            "proposal end before start"
+            "proposal end before proposal start"
         ));
     }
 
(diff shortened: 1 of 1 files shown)


vm/actor/src/builtin/market/mod.rs
```diff
@@ -1251,12 +1251,6 @@ where
     BS: BlockStore,
     RT: Runtime<BS>,
 {
-    if proposal.proposal.end_epoch <= proposal.proposal.start_epoch {
-        return Err(actor_error!(
-            ErrIllegalArgument,
-            "proposal end epoch before start epoch"
-        ));
-    }
```
noot, line 1259: I think this check still needs to happen, but me moved after the `verify_signature` step?

elmattic: mmh the check is already happening [here](https://github.com/ChainSafe/forest/blob/main/vm/actor/src/builtin/market/mod.rs#L1185-L1190).
I don't see the point to do it twice. Lotus does it only once too if I recall.

noot (approved):

PR #1322, opened 2021-12-08
Elmattic/actors review f20

### LINKED ISSUE #1286: F20: Operations performed out-of-order in ExpirationQueue::reschedule_all_as_faults
Description of the issue [HERE](https://hackmd.io/eThp9-pyR0WfLJjBynFntA?view#F20-Operations-performed-out-of-order-in-ExpirationQueuereschedule_all_as_faults)

elmattic:
**Summary of changes**
- Move validation code out of the `for_each` loop

Closes #1286

Note this still doesn't match exactly one for one Go actor logic. E.g. if we return at [this point](https://github.com/chainsafe/forest/blob/ccf1ac118aefe7143b0a97a671f241e052f79343/vm/actor/src/builtin/miner/expiration_queue.rs#L318) we never touch the state since we update it [afterwards](https://github.com/chainsafe/forest/blob/ccf1ac118aefe7143b0a97a671f241e052f79343/vm/actor/src/builtin/miner/expiration_queue.rs#L334). In contrary Go actor updates the state in the [for_each](https://github.com/filecoin-project/specs-actors/blob/97fc4480cbc9b7078c67276d8c3394e3423ef513/actors/builtin/miner/expiration_queue.go#L292) loop.


arajasek (approved):
This seems correct to me. Note this is a somewhat cosmetic fix -- if validate state fails, the chain halts (and so a subtle distinction between how Go and Rust behaved is immaterial).

Also note that this is only called by Cron, so gas is irrelevant (cron is gasless).

noot (approved):

PR #1323, opened 2021-12-08
laudiacay/actors review F24

### LINKED ISSUE #1301: F24: Forest does not reject BLS identity address during account actor creation
Description of the issue [HERE](https://hackmd.io/eThp9-pyR0WfLJjBynFntA?view#F24-Forest-does-not-reject-BLS-identity-address-during-account-actor-creation)

laudiacay:
**Summary of changes**
- adding version-gated checks for BLS zero key actor creation, and lazy-static to generate that key once per run in memory

Closes #1301 

half of this has been merged into FVM, other half of this we're waiting on because it'll need to go outside of actors in lotus-territory.


vm/interpreter/src/default_runtime.rs
```diff
@@ -468,6 +468,10 @@ where
     ) -> Result<(ActorState, Address), ActorError> {
         self.charge_gas(self.price_list().on_create_actor())?;
 
+        if addr.is_bls_zero_address() && self.network_version() >= NetworkVersion::V10 {
+            Err(actor_error!(SysErrIllegalArgument; "cannot create the bls zero address actor"))?
```
elmattic, line 472: `ErrIllegalArgument` is used by Lotus so this is a problem. Is this ok in Lotus? I would say Lotus should fix and use`SysErrIllegalArgument` too no?

elmattic: Return directly the error without `?` to make `cargo clippy` happy.

noot: updated this!

noot (approved):

PR #1325, opened 2021-12-09
laudiacay/actors review f23

### LINKED ISSUE #1300: F23: No overflow protection in GasTracker::charge_gas
Description of the issue [HERE](https://hackmd.io/eThp9-pyR0WfLJjBynFntA?view#F23-No-overflow-protection-in-GasTrackercharge_gas)

laudiacay:
**Summary of changes**
- overflow protection/handling in charge_gas, fixes actors audit f23

Closes #1300 

### DIFF
--- a/vm/interpreter/src/gas_tracker/mod.rs
+++ b/vm/interpreter/src/gas_tracker/mod.rs
@@ -25,27 +25,16 @@ impl GasTracker {
     /// enough gas remaining for charge.
     pub fn charge_gas(&mut self, charge: GasCharge) -> Result<(), ActorError> {
         let to_use = charge.total();
-        let used_or = self.gas_used.checked_add(to_use);
-        match used_or {
-            None => {
-                self.gas_used = self.gas_available;
-                Err(actor_error!(SysErrOutOfGas;
-                    "adding gas_used={} and to_use={} overflowed",
-                    self.gas_used, to_use
-                ))
-            }
-            Some(used) => {
-                if used > self.gas_available {
-                    self.gas_used = self.gas_available;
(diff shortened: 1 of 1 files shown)


vm/interpreter/src/gas_tracker/mod.rs
```diff
@@ -25,16 +25,27 @@ impl GasTracker {
...
-        if used > self.gas_available {
-            self.gas_used = self.gas_available;
-            Err(actor_error!(SysErrOutOfGas;
-                    "not enough gas (used={}) (available={})",
-               used, self.gas_available
-            ))
-        } else {
-            self.gas_used += to_use;
-            Ok(())
+        let used_or = self.gas_used.checked_add(to_use);
```
elmattic, line 28: I would not complicate the Rust code here and distinguish the two cases (`None`/`Some`).
Just use the equivalent algorithm than what Forest does (it's really overflow safe).

So the code would be something like:
```
if self.gas_used > self.gas_available-to_use {
    ...
} else {
    self.gas_used += to_use;
    Ok(())
}
```

elmattic (commented):
Simplify logic to be like in Lotus.

vm/interpreter/src/gas_tracker/mod.rs
laudiacay: it just wasn't immediately evident to me why it was overflow safe (I had to think for a second, and could have seen tired-me or someone else changing it back to the overflowable form)

but now I realize this problem is solvable with the highly advanced technology of "commenting your code"

will change

elmattic (approved):

PR #1327, opened 2021-12-10
fix market actor publish_storage_deals

noot:
**Summary of changes**
- fix market actor publish_storage_deals panic, see lotus: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go#L198

Closes n/a

### DIFF
--- a/vm/actor/src/builtin/market/mod.rs
+++ b/vm/actor/src/builtin/market/mod.rs
@@ -269,6 +269,8 @@ impl Actor {
         msm.with_pending_proposals(Permission::ReadOnly);
         msm.with_escrow_table(Permission::ReadOnly);
         msm.with_locked_table(Permission::ReadOnly);
+        msm.build()
+            .map_err(|e| e.downcast_default(ExitCode::ErrIllegalState, "failed to load msm"))?;
 
         // the for loop too uses rt immutably and also needs to mutably access it.
         // so we will need to isolate them.
--- a/vm/actor/src/builtin/market/mod.rs
+++ b/vm/actor/src/builtin/market/mod.rs
@@ -265,7 +265,8 @@ impl Actor {
         let mut state: State = rt.state::<State>()?;
 
         let store = rt.store();
-        let msm = state.mutator(store);
+        let mut msm = state.mutator(store);
+        msm.with_pending_proposals(Permission::ReadOnly);
 
         // the for loop too uses rt immutably and also needs to mutably access it.
         // so we will need to isolate them.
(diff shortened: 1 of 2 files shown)


vm/actor/src/builtin/market/mod.rs
```diff
@@ -265,7 +265,10 @@ impl Actor {
         let mut state: State = rt.state::<State>()?;
 
         let store = rt.store();
-        let msm = state.mutator(store);
+        let mut msm = state.mutator(store);
+        msm.with_pending_proposals(Permission::ReadOnly);
+        msm.with_escrow_table(Permission::ReadOnly);
+        msm.with_locked_table(Permission::ReadOnly);
```
elmattic, line 271: Not tested but pretty sure you can chain withs like this:

```
let msm = state.mutator(store)
    .with_pending_proposals(Permission::ReadOnly)
    .with_escrow_table(Permission::ReadOnly)
    .with_locked_table(Permission::ReadOnly);
```

This would allow `msm` binding staying immutable.

elmattic (commented):
lgtm

vm/actor/src/builtin/market/mod.rs
noot: hm, i get an error when i try that:
```
error[E0716]: temporary value dropped while borrowed
   --> vm/actor/src/builtin/market/mod.rs:268:23
    |
268 |           let msm = state.mutator(store)
    |                         ^^^^^^^^^^^^^^^^^^^^ creates a temporary which is freed while still in use
...
271 |           .with_locked_table(Permission::ReadOnly);
    |                                                   - temporary value is freed at the end of this statement
...
319 |               let client_balance_ok = msm
    |  _____________________________________-
320 | |                 .balance_covered(client, total_client_lockup.get(&client).unwrap())
    | |___________________________________________________________________________________- borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value
```

elmattic (approved):

vm/actor/src/builtin/market/mod.rs
elmattic: There are places like [here](https://github.com/ChainSafe/forest/blob/c53d8794477625f4fbb9189b2b2275b9bf2f8a5c/vm/actor/src/builtin/market/mod.rs#L604) where withs are chained but then a `build` is used. Not sure at all if we need this.

noot: oh nice, yeah updated!

PR #1340, opened 2021-12-16
Elmattic/actors review f26

### LINKED ISSUE #1303: F26: Both Forest and Lotus VMs allow an account to send a negative value to itself
Description of the issue [HERE](https://hackmd.io/eThp9-pyR0WfLJjBynFntA?view#F26-Both-Forest-and-Lotus-VMs-allow-an-account-to-send-a-negative-value-to-itself)

elmattic:
**Summary of changes**
- Fix testing of balance when self transfering tokens
- Add tests

Closes #1303

This fix will be activated starting `NV15`.

Equivalent fix in Lotus is [here](https://github.com/filecoin-project/lotus/pull/7637).


noot (commented):
looks good! since this is for V15 (even if it won't affect the code to merge it in), it would be nicer to merge this into a V15 branch so when we tag a release for V14 it's cleaner

noot (approved):

vm/interpreter/src/default_runtime.rs
```diff
@@ -1135,37 +1142,85 @@ fn transfer<BS: BlockStore>(
     from: &Address,
     to: &Address,
     value: &TokenAmount,
+    version: NetworkVersion,
 ) -> Result<(), ActorError> {
-    if from == to {
-        return Ok(());
-    }
+    let (to_id, mut f) = if version >= NetworkVersion::V15 {
+        if value.is_negative() {
```
LesnyRumcajs, line 1148: Out of curiosity, sending a negative value is an error, sending to same address is noop - what about sending 0? I understand that Filecoin allows such transactions? If so, do you know what's the purpose of them?

LesnyRumcajs (approved):

vm/interpreter/src/default_runtime.rs
elmattic: I don't think you can create a working transaction with 0 FIL because you will have to pay gas at some point (there's no free lunch).

LesnyRumcajs: Makes sense. I guess there won't be that many `0 FIL` transactions so potential performance gain from short-circuiting on that would be negligible. Looks good to me, thanks!

elmattic: There are some check for `!= 0` transfer but this is done after charging.

See [this line](https://github.com/ChainSafe/forest/blob/elmattic/actors-review-F26/vm/interpreter/src/default_runtime.rs#L430).

types/src/version.rs
```diff
@@ -41,6 +41,8 @@ pub enum NetworkVersion {
     V13,
     /// chocolate (specs-actors v6.0.0)
     V14,
+    // TBD
```
LesnyRumcajs, line 44: I believe the name is OhSnap?

q9f:
Please be more descriptive with PR titles in future 🙏🏼

types/src/version.rs
elmattic: That's korrekt.

elmattic: See duplicated version [here](https://github.com/filecoin-project/ref-fvm/blob/6f3db40ad7108e27146d0b9efdf7c64107507839/shared/src/version/mod.rs)
Related issue #1448.

PR #1354, opened 2022-01-06
FIP-0023: Break ties between tipsets of equal weight

### LINKED ISSUE #1258: implement FIP-0023: Break ties between tipsets of equal weight
**Task summary**

implement FIP-0023: Break ties between tipsets of equal weight

**Specification reference**

-  see https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0023.md

connormullett:
**Summary of changes**
- Introduces logic to break tipsets of equal weight by comparing the smallest ticket within them

Closes #1258 


q9f:
Do you still need this PR, Connor?

lemmih:
I think this still needs to be done.

connormullett:
Currently blocked: no reference implementation, no test cases, building test cases would be difficult as we would need tipsets with equal weight.

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -215,7 +224,23 @@ impl TipsetGroup {
...
+        match weight.cmp(otherw) {
+            Ordering::Greater => Ordering::Greater,
+            Ordering::Equal => {
+                if self.tipsets[i].break_weight_tie(&other.tipsets[j]) {
+                    Ordering::Greater
+                } else {
+                    Ordering::Equal
+                }
+            }
+            Ordering::Less => Ordering::Less,
```
elmattic, line 242: Could use `r @ _ => r`?

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -187,21 +187,30 @@ impl TipsetGroup {
...
+            .tipsets
             .iter()
-            .map(|ts| ts.weight().clone())
-            .max()
-            .unwrap()
+            .enumerate()
+            .filter(|(_, ts)| ts.weight() == max);
+
+        let first = ties.next().unwrap();
+        let (index, ts) = ties.fold(first, |(i, ts), (j, other)| {
```
elmattic, line 205: Use `reduce` instead of `next/fold`?

blockchain/blocks/src/tipset.rs
```diff
@@ -133,6 +134,21 @@ impl Tipset {
     pub fn weight(&self) -> &BigInt {
         self.min_ticket_block().weight()
     }
+    /// Returns true if self wins according to the filecoin tie-break rule
+    /// (FIP-0023)
+    pub fn break_weight_tie(&self, other: &Tipset) -> bool {
+        // blocks are already sorted by ticket
+        for (a, b) in self.blocks().iter().zip(other.blocks().iter()) {
+            let ticket = a.ticket().as_ref().unwrap();
```
LesnyRumcajs, line 142: Is it safe to unwrap here? If yes, let's put a comment on why. If not, perhaps we should take care of the error in a more gentle way?

blockchain/blocks/src/tipset.rs
```diff
@@ -133,6 +134,21 @@ impl Tipset {
...
+            let othtkt = b.ticket().as_ref().unwrap();
```
LesnyRumcajs, line 143: Let's use a name that doesn't resemble one of the Cthulhu Mythos deities. Perhaps `other_ticket`?

blockchain/blocks/src/tipset.rs
```diff
@@ -133,6 +134,21 @@ impl Tipset {
     pub fn weight(&self) -> &BigInt {
         self.min_ticket_block().weight()
     }
+    /// Returns true if self wins according to the filecoin tie-break rule
+    /// (FIP-0023)
+    pub fn break_weight_tie(&self, other: &Tipset) -> bool {
+        // blocks are already sorted by ticket
+        for (a, b) in self.blocks().iter().zip(other.blocks().iter()) {
```
LesnyRumcajs, line 141: I believe this could be written more expressively with a call to [any](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any). Not a big deal though.

blockchain/blocks/src/tipset.rs
```diff
@@ -343,3 +359,70 @@ pub mod tipset_json {
...
+    use address::Address;
+    use cid::Cid;
+    use crypto::VRFProof;
+    use num_bigint::BigInt;
+
+    use std::convert::TryFrom;
+
+    pub fn mock_block(id: u64, weight: u64, ticket_sequence: u64) -> BlockHeader {
+        let addr = Address::new_id(id);
+        let c =
```
LesnyRumcajs, line 375: let's use more expressive variable names

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -187,21 +187,31 @@ impl TipsetGroup {
...
-            .map(|ts| ts.weight().clone())
-            .max()
-            .unwrap()
+            .enumerate()
+            .filter(|(_, ts)| ts.weight() == max);
+
+        let (index, ts) = ties
+            .reduce(|(i, ts), (j, other)| {
+                // break the tie
+                if ts.break_weight_tie(other) {
```
LesnyRumcajs, line 207: Do you think we could make it more idiomatic by implementing `PartialOrd` for `Tipset` with this logic? Seems to me this logic should be contained there, I may be in the wrong though. Thoughts?

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -215,7 +225,22 @@ impl TipsetGroup {
     }
 
     fn is_heavier_than(&self, other: &Self) -> bool {
-        self.heaviest_weight() > other.heaviest_weight()
+        self.cmp(other).is_gt()
+    }
+
+    fn cmp(&self, other: &Self) -> Ordering {
```
LesnyRumcajs, line 231: Should this be rather an implementation for `PartialOrd`? It would simplify some logic in other places, e.g. we wouldn't have to call `max_by` with a lambda, just `max`.

LesnyRumcajs (changes requested):

blockchain/blocks/src/tipset.rs
elmattic: Normally yes. Function `block_sanity_checks` should have been called at this point. What you think?

blockchain/blocks/src/tipset.rs
elmattic: Yes, good idea ;)

blockchain/chain_sync/src/tipset_syncer.rs
elmattic: Yes, I see your point. In this case I would have also to implement `PartialEq` [trait](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) (not sure how). So don't makes sense to go down to this route, feels a bit too much complex for a seldom use case.

blockchain/blocks/src/tipset.rs
q9f: In that case, use `expect()`:

```suggestion
            let ticket = a.ticket().as_ref().expect("The function block_sanity_checks() should have been called at this point.");
```

lemmih (approved):

blockchain/blocks/src/tipset.rs
```diff
@@ -133,6 +134,27 @@ impl Tipset {
...
+            .zip(other.blocks().iter())
+            .any(|(a, b)| {
+                const MSG: &str =
+                    "The function block_sanity_checks should have been called at this point.";
+                let ticket = a.ticket().as_ref().expect(MSG);
+                let other_ticket = b.ticket().as_ref().expect(MSG);
+                ticket.vrfproof < other_ticket.vrfproof
+            });
+        if broken {
+            info!("weight tie broken in favour of {:?}", self.key());
```
LesnyRumcajs, line 152: should it be `info` or rather `debug`? Not sure how often should we see it and how relevant it is for anyone running the node.

LesnyRumcajs (approved):

LesnyRumcajs:
Nice!

blockchain/blocks/src/tipset.rs
lemmih: Or maybe `warn` if this is something that is highly unexpected.

elmattic: I would prefer to let it as `info` first so that it's really visible to all of us, then turn it to `debug` or `warn` yes.
It's probably an event that is very likely to happen.

LesnyRumcajs:
@elmattic mergio?

PR #1373, opened 2022-01-13
networks: Show an informative error message if the selected feature set is invalid.

lemmih:
**Summary of changes**
- In the `networks` crate, emit an informative error message if none of the required feature flags are selected.

Closes nothing.

This improves the error message if one tries to build `networks` without any feature flags set.

### DIFF
--- a/types/networks/src/lib.rs
+++ b/types/networks/src/lib.rs
@@ -15,7 +15,9 @@ mod drand;
     not(feature = "devnet"),
     not(feature = "mainnet")
 ))]
-compile_error!("Either feature \"mainnet\", \"devnet\", or \"interopnet\" must be enabled for this crate.");
+compile_error!(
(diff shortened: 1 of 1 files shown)


types/networks/src/lib.rs
```diff
@@ -19,7 +28,6 @@ pub use self::mainnet::*;
 mod mainnet;
 #[cfg(feature = "conformace")]
 pub use self::mainnet::*;
-pub use crate::mainnet::*;
```
noot, line 22: I think this might be breaking the test vectors: https://github.com/ChainSafe/forest/runs/4806953786?check_suite_focus=true

Lemmih:
Yeah, the situation was a bit more complicated then I thought.

Old behavior:
Using `mainnet` and `conformance` at the same time broke the build with no intelligible error message. Using `mainnet` and `devnet` at the same time would silently ignore `devnet`.

New behavior: One and only one network must be specified. If this isn't the case then an error will be raised.

PR #1382, opened 2022-01-14
ci: fix gh pages action version string

q9f:
I must have fatfingered this 🙄


lemmih (approved):

lemmih:
Oh, we also need to restore the `scripts/build-rust-docs.sh` script or inline it.

lemmih (changes requested):
Think I accidentally deleted `scripts/build-rust-docs.sh`.

lemmih (changes requested):
Oh, we don't need the script. That entire section should just be removed from the workflow.

q9f:
#1383

PR #1385, opened 2022-01-17
Remove unused lint exceptions.

lemmih:
**Summary of changes**
- Remove lint exceptions for `absurd_extreme_comparisons`, `suspicious_op_assign_impl`, `unreachable_code`, `unused_variables`, and `clippy::all`.

The remaining lint exceptions are related to style rather than correctness (eg. `clippy::too_many_arguments` and `clippy::type_complexity`).


q9f (approved):

elmattic:
Just a question on the removal of those lint exceptions. If I understand correctly those exceptions were out of date?
It doesn't have been disable because of some removal of clippy lints?

I would rather have all possible checks and lints done, for exemple those kind of code should raise a red flag (`suspicious_op_assign_impl`):

```
impl AddAssign for Foo {
    fn add_assign(&mut self, other: Foo) {
        *self = *self - other;
    }
}
```

lemmih:
Yes, removing the exceptions means more lints are checked. The `suspicious_op_assign_impl` lint is enabled by default but was manually suppressed. Fortunately none of the correctness-related lint exceptions are required anymore and could just be removed.

lemmih:
Why does it believe I haven't signed the CLA?

PR #1388, opened 2022-01-19
nv15: update crates for actors v6 release

noot:
**Summary of changes**
Changes introduced in this pull request:
- released crate is here https://github.com/ChainSafe/forest-actorv5 (version is 5 since i believe there was no actors v1 or something)

**Reference issue to close (if applicable)**

Closes 

**Other information and links**

Issue #1376, opened 2022-01-13
NV15: implement FIP28


noot:
**Task summary**

see https://github.com/ChainSafe/forest/issues/1352

**Specification reference**

- 

**Other information and links**

lemmih:
Part of FVM, closing.

Issue #1375, opened 2022-01-13
NV15: implement FIP19


noot:
**Task summary**

see https://github.com/ChainSafe/forest/issues/1352

**Specification reference**

- 

**Other information and links**

lemmih:
Part of FVM, closing.

PR #1392, opened 2022-01-20
[WIP] Nv15

elmattic:
**Summary of changes**
- merged #1388
- Bump local version of actors to 6.0.0 (v7 actors)
- Add new network and actor versions

Closes 


elmattic:
@lemmih What's missing before we can start implementing #1375 and #1376?

I think we need to add to each actor an new version for its state too (like [here](https://github.com/ChainSafe/forest/blob/main/vm/actor_interface/src/builtin/account/mod.rs)).

elmattic:
I'm not sure if it's intended but at some [place](https://github.com/ChainSafe/forest/blob/9ff55bf4a656351626c59769c7ef96e2d81df9b2/vm/actor_interface/src/builtin/miner/mod.rs#L278) `Deadline::V5` for `State::V6` is used and I've stick to it to for `State::V7`

lemmih:
> I'm not sure if it's intended but at some [place](https://github.com/ChainSafe/forest/blob/9ff55bf4a656351626c59769c7ef96e2d81df9b2/vm/actor_interface/src/builtin/miner/mod.rs#L278) `Deadline::V5` for `State::V6` is used and I've stick to it too for `State::V7`

I have no idea what the code is supposed to do. Can we find the corresponding go code in lotus?

lemmih:
Background information:
The test vectors were updated to run against nv14 (aka `chocolate`, actor_v6) rather than nv13 (aka `hyperdrive`, actor_v5). Only the network version was changed so the actors are still v5. Specs-actors handles this by performing a migration (see [this file](https://github.com/filecoin-project/specs-actors/blob/v6/actors/migration/nv14/top.go#L68)). We don't have a similar migration so our tests fail with the message `no code for actor at address f0101 with code [snip]`. This error happens when actors have an unknown code id (see [this file](https://github.com/ChainSafe/forest/blob/main/vm/actor/src/lib.rs#L84)).

PR #1401, opened 2022-01-26
Include network in forest version string.

lemmih:
**Summary of changes**
- Include network (eg. `mainnet` or `devnet`) in the version string shown with `forest --version`.

Target network is selected at compile time but the forest executable couldn't show which network had been selected.


q9f (approved):
I do not disagree with having this as part of the version string as long as this is a compile-time feature.

connormullett (approved):
:+1:

PR #1411, opened 2022-02-01
Fix 1333

### LINKED ISSUE #1333: C5: Missing checks when deserializing Cids
Description of the issue [HERE](https://hackmd.io/eThp9-pyR0WfLJjBynFntA?view#C5-Missing-checks-when-deserializing-Cids)

willeslau:
**Summary of changes**
- Match with the implementation in Lotus

Closes #1333 


ipld/cid/src/lib.rs
```diff
@@ -122,10 +122,23 @@ impl<'de> de::Deserialize<'de> for Cid {
             Some(CBOR_TAG_CID) | None => {
                 let mut bz = tagged.value.into_vec();
 
-                if bz.first() == Some(&MULTIBASE_IDENTITY) {
-                    bz.remove(0);
+                if bz.len() == 0 {
```
lemmih, line 125: Is this check necessary? Isn't it caught by `Cid::try_from(bz)` later on?

ipld/cid/src/lib.rs
```diff
@@ -122,10 +122,23 @@ impl<'de> de::Deserialize<'de> for Cid {
...
+                    return Err(de::Error::custom("undefined cid"));
                 }
 
+                if bz.len() < 2 {
```
lemmih, line 129: Same with this. Isn't it handled by `Cid::try_from(bz)`?

ipld/cid/src/lib.rs
```diff
@@ -122,10 +122,23 @@ impl<'de> de::Deserialize<'de> for Cid {
             Some(CBOR_TAG_CID) | None => {
```
lemmih, line 122: `None` should no longer match.

ipld/cid/src/lib.rs
```diff
@@ -122,10 +122,23 @@ impl<'de> de::Deserialize<'de> for Cid {
...
+                    return Err(de::Error::custom("undefined cid"));
                 }
 
+                if bz.len() < 2 {
+                    return Err(de::Error::custom(
+                        "cbor serialized CIDs must have at least two bytes",
+                    ));
+                }
+
+                if bz.first() != Some(&MULTIBASE_IDENTITY) {
```
lemmih, line 135: This looks good. I'd like to see unit tests covering the different behaviors.

lemmih (changes requested):
Good start. Requires slight changes and more unit tests.

ipld/cid/src/lib.rs
willeslau: @lemmih pr is updated, pls check

willeslau: Aligning with Lotus would break these tests: https://github.com/ChainSafe/forest/blob/main/ipld/tests/cbor_test.rs#L20 because 1 at index 0 is no longer allowed. Not sure if the tests should be updated.

lemmih: Yes, those tests have to be updated.

q9f:
Hi Willes, did you have a chance to rebase this on the latest main and see if you could address the tests?

willeslau:
@q9f  Yep, let me check it out

willeslau:
@q9f @lemmih I think I found the issue, the problem seems to be in IPLD deserialization. In the source code, ipld/src/de.rs:124
```
    fn visit_newtype_struct<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
    where
        D: de::Deserializer<'de>,
    {
        match current_cbor_tag() {      <=========== The tag is already read, so when call Cid::deserialize(deserializer), it's None for the tag and an error is thrown as we no longer allow None tag.
            Some(42) => {
                let cid = Cid::deserialize(deserializer)?;

                Ok(Ipld::Link(cid))
            }
            Some(tag) => Err(de::Error::custom(format!("unexpected tag ({})", tag))),
            _ => Err(de::Error::custom("tag expected")),
        }
    }
```

PR #1412, opened 2022-02-01
fix exit code mismatch

noot:
**Summary of changes**
- fix exit code mismatch causing receipt root mismatch on mainnet

Closes n/a


lemmih:
Are there some situations where we need to return `SysErrForbidden` and others where `ErrForbidden` is correct?

elmattic:
> Are there some situations where we need to return `SysErrForbidden` and others where `ErrForbidden` is correct?

The rule of thumb is that `SysErrXXX` errors must be used in non-actor code only (so here the VM) while `ErrXXX` are actor code only.

lemmih:
I think we're using test-vectors that are no longer valid. Re-generating the test vectors from scratch changes the expected exit codes.

lemmih:
Here's the patch that changed the test vector: https://github.com/filecoin-project/specs-actors/pull/1566

Note that they only updated the `specs-actors` repository and not the `test-vectors` repository.

noot:
@lemmih since we directly pull `test-vectors` as a submodule we need to either wait for the repo to be updated with the correct vectors to get CI to pass, or to fix it ourselves (which as you pointed out is cumbersome).

lemmih:
> @lemmih since we directly pull `test-vectors` as a submodule we need to either wait for the repo to be updated with the correct vectors to get CI to pass, or to fix it ourselves (which as you pointed out is cumbersome).

As a short-term solution, I say we fork `test-vectors` and update the vectors ourselves.

lemmih:
Hi noot! I disabled the borked test vectors and pull in their replacement to the `extra-vectors/` folder. Not all of the updated test vectors pass but I think we can tackle that in another PR.

noot:
@lemmih thanks! I can tackle the failing test in a follow up PR.

PR #1416, opened 2022-02-02
Add assert in debug mode and tests

### LINKED ISSUE #1361: B5: Forest bitfield rle writer will write 10-byte varints

elmattic:
**Summary of changes**
- Add assert in debug mode to avoid misuses
- Add two tests

Closes #1361

Will notice PL so they can report PR on `ref-fvm` repo.

### DIFF
--- a/utils/bitfield/src/rleplus/writer.rs
+++ b/utils/bitfield/src/rleplus/writer.rs
@@ -1,6 +1,9 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+/// The maximum varint that can be serialized on 9 bytes.
+const MAX_VARINT: u64 = 2u64.pow(63)-1;
+
 #[derive(Default, Clone, Debug)]
 /// A `BitWriter` allows for efficiently writing bits to a byte buffer, up to a byte at a time.
 pub struct BitWriter {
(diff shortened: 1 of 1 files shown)


q9f (approved):

utils/bitfield/src/rleplus/writer.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+/// The maximum varint that can be serialized on 9 bytes.
+const MAX_VARINT: u64 = 2u64.pow(63)-1;
```
LesnyRumcajs, line 5: What do you think about using `usize` here? Unless we're supporting 32-bit systems it should be fine, plus we would get rid of some casts.

elmattic: Good question. We indeed want 32-bit systems support like `wasm32`. Protocol Labs had to port our code to explicitly use `u64` (there are many commits, the one for Bitfield is [here](https://github.com/filecoin-project/ref-fvm/commit/95e1f33a4433cb55d4ad0ff75b013e4e8b539b30)). We will probably use some of their crates instead of ours (I'm commiting here but this PR will have to be done on the `ref-fvm` repo too to keep things in sync in the meantime).

Even if we don't want to support say 32-bit system I think it's a good practice to use properly size type (makes the code more readable and correct).

LesnyRumcajs: Thanks for the explanation! In this case, should we rather do an opposite cast which wouldn't be lossy? Given that potentially (on a 32-bit machine) `MAX_VARINT as usize` would get truncated to 32 bits (`u32::MAX`). In the end the behaviour would still be the same - on a 32 bit machine the condition will always pass regardless.

LesnyRumcajs (approved):

PR #1422, opened 2022-02-10
Fail CI on failed fmt or other linting file changes

LesnyRumcajs:
**Summary of changes**
- Changes the CI lint step to check for changes resulted from e.g. `cargo fmt`. Before they were silently discarded.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -98,7 +98,7 @@ jobs:
       - name: Install RustFmt
         run: rustup component add rustfmt
       - name: Run Linters
-        run: make lint-check
+        run: make lint
 
   audit:
     name: Audit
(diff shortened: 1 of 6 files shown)


lemmih:
Just change the `lint` target to run `cargo fmt --all -- --check`. It should not actually do any formatting.

LesnyRumcajs:
> Just change the `lint` target to run `cargo fmt --all -- --check`. It should not actually do any formatting.

Yes, but then anyone wanting to run the lint will have to `cargo fmt --all` himself - trying not to introduce _breaking changes_ to existing contributors :smile: . What do you think?

lemmih:
> > Just change the `lint` target to run `cargo fmt --all -- --check`. It should not actually do any formatting.
> 
> Yes, but then anyone wanting to run the lint will have to `cargo fmt --all` himself - trying not to introduce _breaking changes_ to existing contributors smile . What do you think?

I think it is a mistake that `lint` modified the code in the first place. In my mind, linting should check for problems, not automatically fix them.

lemmih (approved):

q9f (approved):

PR #1423, opened 2022-02-10
add shellcheck to CI

LesnyRumcajs:
**Summary of changes**
- adds ShellCheck to the CI to lint project's shell scripts.

### DIFF
--- a/.github/workflows/shell.yml
+++ b/.github/workflows/shell.yml
@@ -1,4 +1,4 @@
-name: ShellCheck
+name: Shell
 
 on:
   pull_request:


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches:
+      - main
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Run shellcheck
```
q9f, line 17: What about adding these two lines to the lint job to reduce complexity in the workflows?

https://github.com/ChainSafe/forest/blob/main/.github/workflows/rust.yml#L100

q9f: I agree, however, renaming it means we need to mess with all the PRs again. Maybe just keep it in `shell.yml` for now.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,18 @@
+name: ShellCheck
```
q9f, line 1:
```suggestion
name: Shell
```

Let's name it also `Shell` so in case we need to add more later.

.github/workflows/shellcheck.yml
LesnyRumcajs: Sounds good to me!

q9f (approved):

PR #1429, opened 2022-02-14
rustc: switch to rust edition 2021

q9f:
from 2018

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3539,11 +3539,11 @@ dependencies = [
 
 [[package]]
 name = "jsonwebtoken"
-version = "8.0.1"
+version = "7.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "012bb02250fdd38faa5feee63235f7a459974440b9b57593822414c31f92839e"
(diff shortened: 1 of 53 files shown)


Cargo.lock
```diff
@@ -3539,11 +3539,11 @@ dependencies = [
 
 [[package]]
 name = "jsonwebtoken"
-version = "8.0.1"
+version = "7.2.0"
```
q9f, line 3542: I will never understand what is causing this.

lemmih:
CI is failing due to transient network errors.

LesnyRumcajs (commented):
Do you know if it's possible to define this only in the workspace root and let it be propagated without repeating the same setting everywhere?

q9f:
I don't think it's possible because it will always use `2015` as the default unless indicated otherwise.

LesnyRumcajs:
> I don't think it's possible because it will always use `2015` as the default unless indicated otherwise.

Indeed, seems this is still not implemented. https://github.com/rust-lang/cargo/issues/5784

Hopefully one day we'll get rid of this boilerplate!

LesnyRumcajs (approved):

lemmih (approved):

PR #1434, opened 2022-02-15
introduce serde_generic_array

### LINKED ISSUE #1334: C6: []T serde missing max length checks for arrays of items
Description of the issue [HERE](https://hackmd.io/eThp9-pyR0WfLJjBynFntA?view#C6-T-serde-missing-max-length-checks-for-arrays-of-items)

clearloop:
**Summary of changes**
- [x] rename `serde_byte_array`
- [x] introduce `serde_generic_array`
- [x] add macro `check_generic_array_length!`

Closes #1334 


q9f:
Hey Tianyi, is this ready for review?

q9f:
> > Hey Tianyi, is this ready for review?
> 
> yeah, I'm thinking of should I provide some more tests for `BlockHeader`

Yes, I would appreciate that 🙏🏼

encoding/src/serde_byte_array.rs
```diff
@@ -5,48 +5,41 @@ use crate::BYTE_ARRAY_MAX_LEN;
...
-        if len > BYTE_ARRAY_MAX_LEN {
-            return Err(ser::Error::custom::<String>(
+/// checked if output > `crate::ByteArrayMaxLen`
+pub fn deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
+where
+    T: Deserialize<'de> + AsRef<[u8]>,
+    D: Deserializer<'de>,
+{
+    Deserialize::deserialize(deserializer).and_then(|bytes: T| {
+        if bytes.as_ref().len() > BYTE_ARRAY_MAX_LEN {
```
LesnyRumcajs, line 30: Perhaps it would be useful to add a debug log indicating what was the size of the offending array. Or perhaps put it into the error message? What do you think?

LesnyRumcajs (changes requested):

encoding/src/serde_byte_array.rs
clearloop: agree with this idea

tried to place this error into https://github.com/ChainSafe/forest/blob/main/encoding/src/errors.rs but it seems that's not where it should be lol

adding a new `Error` type is kind of luxury for this one line error, so I just placed a string which can be captured by `serde` for the minimal modification against to the last commit

or I can add an `Any` item in [`CodecProtocol`](https://github.com/ChainSafe/forest/blob/e66d0f9e2f810752bd3330d9ae2370f7197b234b/encoding/src/errors.rs#L50) to describe this error

```rust
pub enum CodecProtocol {
  // use any since this error occurs in serialize && deserialize
  //
  // add this line ⬇️
  Any,
  Cbor,
}
```

and the error would be like

```rust
let arr_limit_err = encoding::Error {
  description: "Array exceeds max length".into(),
  protocal: CodecProtocol::Any,
}
```

encoding/src/serde_byte_array.rs
```diff
@@ -5,48 +5,41 @@ use crate::BYTE_ARRAY_MAX_LEN;
...
-pub mod serde_byte_array {
-    use super::*;
+/// checked if input > `crate::BYTE_ARRAY_MAX_LEN`
+pub fn serialize<T, S>(bytes: &T, serializer: S) -> Result<S::Ok, S::Error>
+where
+    T: ?Sized + Serialize + AsRef<[u8]>,
+    S: Serializer,
+{
+    if bytes.as_ref().len() > BYTE_ARRAY_MAX_LEN {
+        return Err(ser::Error::custom::<String>(
```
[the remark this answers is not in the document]
clearloop: 0. I can add a `SerdeError` in crate `encoding`

```
pub enum SerdeError {
    ArrayExceedsMaxLength,
}
```

1. use previous `encoding::Error`

or I can add an `Any` item in [`CodecProtocol`](https://github.com/ChainSafe/forest/blob/e66d0f9e2f810752bd3330d9ae2370f7197b234b/encoding/src/errors.rs#L50) to describe this error

```rust
pub enum CodecProtocol {
  // use any since this error occurs in serialize && deserialize
  //
  // add this line ⬇️
  Any,
  Cbor,
}
```

and the error would be like

```rust
let arr_limit_err = encoding::Error {
  description: "Array exceeds max length".into(),
  protocal: CodecProtocol::Any,
}
```

encoding/src/serde_byte_array.rs
```diff
@@ -5,48 +5,41 @@ use crate::BYTE_ARRAY_MAX_LEN;
...
+            "Array exceed max length".into(),
```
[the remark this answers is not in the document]
clearloop: thanks for checking my typo! English everywhere lol

encoding/src/serde_byte_array.rs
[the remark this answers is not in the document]
LesnyRumcajs: Both are fine to me, perhaps having it is `ArrayExceedsMaxLength(usize)` would make it easy to pass the offending length to `Display`.

encoding/src/serde_byte_array.rs
elmattic: Agree with @clearloop. It's kind of luxury, in other parts of the codebase we just `format!` the string with the interesting value, e.g. [here](https://github.com/ChainSafe/forest/blob/main/utils/bitfield/src/rleplus/mod.rs#L100).

LesnyRumcajs: Sure, if you think it's an overkill then it's fine to me. It potentially makes error matching more difficult later on but I'm new to the codebase, if it's unlikely and the error rarely occurs then indeed it may be too much.

encoding/src/serde_byte_array.rs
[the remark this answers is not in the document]
clearloop: https://github.com/ChainSafe/forest/pull/1434/commits/9986f52e7438eb4f2bf01f279a22b0ea273f4b55

encoding/src/serde_byte_array.rs
```diff
@@ -68,13 +63,14 @@ mod tests {
...
+        let len = BYTE_ARRAY_MAX_LEN + 1;
         let bytes = ByteArray {
-            inner: vec![0; BYTE_ARRAY_MAX_LEN + 1],
+            inner: vec![0; len],
         };
 
         assert_eq!(
-            format!("{}", serde_cbor::to_vec(&bytes).err().unwrap()),
-            "Array exceed max length".to_string()
+            serde_cbor::to_vec(&bytes).err().unwrap().to_string(),
```
LesnyRumcajs, line 72: Just a cosmetic question, wouldn't it be equivalent to not compare displays but only errors themselves? In other words, getting rid of `to_string()` in assertions?

LesnyRumcajs (approved):

encoding/src/serde_byte_array.rs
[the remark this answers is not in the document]
LesnyRumcajs: Thank you!

encoding/src/serde_byte_array.rs
clearloop: we had converted our Error types into the associated `Error` types with `ser::Error::custom` and `de::Error::custom` in serializing and deserializing(serde requires, see https://serde.rs/convert-error.html), and the error output in `serde_cbor::to_vec` converted the associated serde Error type into `serde_cbor::Error` which is `serde_cbor::Error::Message(string)` as result...so, we could not compare this `serde_cbor::Error` with our `SerdeError` but `to_string` can make this work

LesnyRumcajs: @clearloop makes sense, thanks, I missed that. We could probably implement `From` for this type but that would be definitely an overkill, I'm fine with `String`. :+1:

lemmih (approved):

PR #1443, opened 2022-02-23
improve license check script

LesnyRumcajs:
**Summary of changes**
- The license check script would fail if you run `git submodule --init --recursive` before the check as some submodules (namely test-vectors' submodules) contain Rust source files (that do not contain Chainsafe license). Now it checks only tracked (or cached) Rust files with one exclusion that was there. I removed the other one as the file was no longer around.


noot (approved):
nice!

q9f (approved):

PR #1426, opened 2022-02-11
move from chrono to time crate

LesnyRumcajs:
**Summary of changes**
Changes introduced in this pull request:
- ~~move from `chrono` to `time` crate to resolve `RUSTSEC-2020-0071` and `RUSTSEC-2020-0159` in https://github.com/ChainSafe/forest/issues/1247~~
- remove direct usage of `chrono` crate in the project

PR #1464, opened 2022-03-04
Enable snapshot export from the Forest client

### LINKED ISSUE #1505: Chain Export
Forest currently has an existing implementation of an export command within a separate repo (`forest_bench`). We would like to move this implementation to within forest such that we can create our own snapshots without relying on lotus

connormullett:
**Summary of changes**
- 

Closes #1505 


elmattic (commented):
Great! Have you thought about writing some tests for `ChainStore::export`? How have you validated that the export command is working as expected?

forest/src/cli/chain_cmd.rs
```diff
@@ -48,6 +69,81 @@ impl ChainCommands {
...
+                    Ok(head) => head.0,
+                    Err(_) => return cli_error_and_die("Could not get network head", 1),
+                };
+
+                let output_path = match output_path {
+                    Some(path) => path.to_owned(),
+                    None => {
+                        let now = Utc::now();
+                        format!(
+                            "forest_snapshot_{}_{}_{}_{}.car",
```
elmattic, line 113: Could you add a short desc for the network and `high` before epoch to help readability?
E.g.
`forest_cali_snapshot_2021_07_14_high_73770`

connormullett: 👍

q9f (dismissed):
Please use `time` instead of `chrono`, see #1426

lemmih:
@connormullett Could you update this PR with a status report? Are you blocked? What is missing?

lemmih:
Let's put this on the back burner for now and revisit after nv16.

node/rpc/src/chain_api.rs
```diff
@@ -49,6 +50,28 @@ where
...
+pub(crate) async fn chain_export<DB, B>(
+    data: Data<RPCState<DB, B>>,
+    Params(params): Params<ChainExportParams>,
+) -> Result<ChainExportResult, JsonRpcError>
+where
+    DB: BlockStore + Send + Sync + 'static,
+    B: Beacon + Send + Sync + 'static,
+{
+    let (epoch, recent_roots, skip_old_msgs, out, TipsetKeysJson(tsk)) = params;
+    let file = File::create(out).await.unwrap();
```
LesnyRumcajs, line 62: `unwrap` is bit harsh, let's just return an error with a helpful message

forest/src/cli/chain_cmd.rs
```diff
@@ -48,6 +69,82 @@ impl ChainCommands {
                 let cid: Cid = cid.parse().unwrap();
                 print_rpc_res_pretty(chain_get_block((CidJson(cid),)).await);
             }
+            Self::Export {
+                tipset,
+                recent_stateroots,
+                skip_old_messages,
+                output_path,
+            } => {
+                let chain_finality = Policy::default().chain_finality;
```
LesnyRumcajs, line 78: shouldn't we take policy from the configuration? otherwise issues may arise in case the correct value is not the default one

forest/src/cli/chain_cmd.rs
```diff
@@ -48,6 +69,82 @@ impl ChainCommands {
...
+                let recent_stateroots = match recent_stateroots {
+                    Some(rsrs) => {
```
LesnyRumcajs, line 80: let's use full name, shadowing is fine :)

forest/src/cli/chain_cmd.rs
```diff
@@ -48,6 +69,82 @@ impl ChainCommands {
...
+                        let rsrs = rsrs.to_owned();
+                        if rsrs < chain_finality {
+                            return cli_error_and_die(
+                                &format!(
+                                    "\recent-stateroots\" must be greater than {}",
```
LesnyRumcajs, line 85: are you sure about this formatting?

forest/src/cli/chain_cmd.rs
```diff
@@ -48,6 +69,82 @@ impl ChainCommands {
...
+                            "forest_snapshot_{}_{}_{}_height_{}.car",
+                            now.year(),
+                            now.month(),
+                            now.day(),
+                            chain_head.epoch()
+                        )
+                    }
+                };
+
+                let epoch = if let Some(epoch) = tipset {
```
LesnyRumcajs, line 123: you could simplify it with `unwrap_or` or something similar

forest/src/cli/chain_cmd.rs
```diff
@@ -48,6 +69,82 @@ impl ChainCommands {
...
+                    *epoch
+                } else {
+                    chain_head.epoch()
+                };
+
+                if recent_stateroots == 0 && *skip_old_messages {
```
LesnyRumcajs, line 129: we could do this check earlier

forest/src/cli/chain_cmd.rs
```diff
@@ -48,6 +69,82 @@ impl ChainCommands {
...
+
+                let params = (
+                    epoch,
+                    recent_stateroots,
+                    *skip_old_messages,
+                    output_path.clone(),
+                    TipsetKeysJson(chain_head.key().clone()),
+                );
+
+                chain_export(params).await.map_err(handle_rpc_err).unwrap();
```
LesnyRumcajs, line 144: `unwrap` is mean

LesnyRumcajs: do we get info where the file was saved?

LesnyRumcajs (changes requested):

forest/src/cli/chain_cmd.rs
connormullett: We should, however I wasn't able to figure out a way to access the current config from the CLI. Specifically within this command. I'll look into this some more, maybe we can pass the config to each command as a refactor?

forest/src/cli/chain_cmd.rs
connormullett: :+1:

forest/src/cli/chain_cmd.rs
connormullett: good idea

forest/src/cli/chain_cmd.rs
connormullett: No, however I can make the API return the output file as a pathbuf to be displayed to the user

forest/src/cli/chain_cmd.rs
connormullett: :+1:

forest/src/cli/chain_cmd.rs
```diff
@@ -17,6 +22,22 @@ pub enum ChainCommands {
...
+    #[structopt(about = "Export chain snapshot to file")]
+    Export {
+        #[structopt(short, help = "Tipset to start the export from, default is @HEAD")]
+        tipset: Option<i64>,
+        #[structopt(
+            short,
+            help = "specify the number of recent state roots to include in the export"
+        )]
+        recent_stateroots: Option<i64>,
+        #[structopt(short, help = "default: false")]
```
elmattic, line 35: Would be more helpful to print a help message for the ```-s``` flag
(you can put it in a doc comment as well, so ```/// Skips old messages```).
Btw default value is expected to be false for a bool flag in a cli.

forest/src/cli/chain_cmd.rs
```diff
@@ -49,6 +70,81 @@ impl ChainCommands {
...
+
+                let params = (
+                    epoch,
+                    recent_stateroots,
+                    *skip_old_messages,
+                    output_path.clone(),
+                    TipsetKeysJson(chain_head.key().clone()),
+                );
+
+                let out = chain_export(params)
```
elmattic, line 141: Just wondering if we could produce a ```.sha256sum``` checksum file at the same time of creating the snapshot.
This could come up faster that reloading the whole ```.car``` file and hashing it.
But having the hash computed with some command in our DO pipeline will be totally fine (maybe this could be part of #1694).

forest/src/cli/chain_cmd.rs
connormullett: Oops, fixed.

forest/src/cli/chain_cmd.rs
connormullett: :+1:

forest/src/cli/chain_cmd.rs
connormullett: I really like this idea. I just implemented this and it will put the checksum in the same directory as the snapshot

lemmih: As guillaume said, there isn't much point in Forest generating the checksum /after/ the snapshot has been generated. If we cannot do it concurrently then it makes more sense to use an external tool.

lemmih:
> The WASM bundles aren't included in the snapshots. I'm investigating why.

Fixed.

lemmih:
This is a minimal PR for chain exporting. Error messages and ergonomics can be improved in the future (issues have been created for this).

API_IMPLEMENTATION.md
```diff
@@ -2,16 +2,17 @@
 
 ## Stats
 
-- Forest method count: 68
+- Forest method count: 69
 - Lotus method count: 173
-- API coverage: 39.31%
+- API coverage: 39.88%
```
LesnyRumcajs, line 7: :rocket:

forest/src/cli/chain_cmd.rs
```diff
@@ -17,6 +20,22 @@ pub enum ChainCommands {
         cid: String,
     },
 
+    /// Export a snapshot of the chain to <output_path>
+    #[structopt(about = "Export chain snapshot to file")]
+    Export {
+        #[structopt(short, help = "Tipset to start the export from, default is @HEAD")]
```
LesnyRumcajs, line 26: If instead of `help = "...` we could put it as a doc comment. This way it'd be visible in documentation (or IDE prompts).

forest/src/cli/chain_cmd.rs
```diff
@@ -49,6 +68,46 @@ impl ChainCommands {
...
+                            epoch,
+                        )
+                    }
+                };
+
+                let params = (
+                    epoch,
+                    *recent_stateroots,
+                    *skip_old_messages,
+                    output_path.clone(),
```
LesnyRumcajs, line 102: why do we need to clone it? It's not used anymore, no?

node/rpc/src/chain_api.rs
```diff
@@ -50,6 +51,44 @@ where
...
+    let recent_roots = recent_roots.unwrap_or(chain_finality);
+    if recent_roots < chain_finality {
+        Err(&format!(
+            "\"recent-stateroots\" must be greater than {}",
+            chain_finality
+        ))?;
+    }
+
+    if recent_roots == 0 && skip_old_msgs {
+        Err("must pass recent-stateroots along with skip-old-messages")?;
```
LesnyRumcajs, line 74: not sure why in one message we are quoting `recent-stateroots` and in another not? I'd say we can skip quoting, no strong opinion though.

node/rpc/src/chain_api.rs
```diff
@@ -50,6 +51,44 @@ where
...
+    }
+
+    let file = File::create(&out).await.map_err(JsonRpcError::from)?;
+    let writer = BufWriter::new(file.clone());
```
LesnyRumcajs, line 78: why clone?

lemmih: I'm surprised clippy didn't bark at this. Fixed.

node/rpc/src/chain_api.rs
lemmih: Fixed.

forest/src/cli/chain_cmd.rs
lemmih: Fixed.

forest/src/cli/chain_cmd.rs
lemmih: Fixed.

LesnyRumcajs (approved):

tyshko5 (approved):

PR #1468, opened 2022-03-08
Limit RocksDB to 200 open files.

lemmih:
Many systems are limited to 1k open files by default.

**Summary of changes**
- Cap RocksDB to 200 open files. This is not optimal and more tuning is required.

### DIFF
--- a/node/db/src/rocks.rs
+++ b/node/db/src/rocks.rs
@@ -30,6 +30,7 @@ impl RocksDb {
         db_opts.create_if_missing(true);
         db_opts.increase_parallelism(num_cpus::get() as i32);
         db_opts.set_write_buffer_size(256 * 1024 * 1024); // increase from 64MB to 256MB
+        db_opts.set_max_open_files(200);
         Ok(Self {
             db: DB::open(&db_opts, path)?,
         })


node/db/src/rocks.rs
```diff
@@ -30,6 +30,7 @@ impl RocksDb {
         db_opts.create_if_missing(true);
         db_opts.increase_parallelism(num_cpus::get() as i32);
         db_opts.set_write_buffer_size(256 * 1024 * 1024); // increase from 64MB to 256MB
+        db_opts.set_max_open_files(200);
```
LesnyRumcajs, line 33: Should this be perhaps a config option (that defaults to `200`), not a hardcoded value? What do you think?

lemmih: Yes, that would be nice. We should probably use a [RockDB Options File](https://github.com/facebook/rocksdb/wiki/RocksDB-Options-File) with our own defaults such that everything can be overridden. Using a configuration file would also make it easier to find optimal parameters.

lemmih: I created an issue (#1473) for using a configuration file. I'd like to still marge this PR with a hard-coded value for open files since it can cause crashes for users (it does for me when I forget to increase my ulimit).

LesnyRumcajs: Sounds good to me.

LesnyRumcajs (approved):

q9f (approved):

Issue #1473, opened 2022-03-09
Use a configuration file for RockDB options.


lemmih:
**Issue summary**

The performance of Forest depends primarily on RocksDB. Tuning the parameters of RocksDB can affect throughput by several orders of magnitude but they are highly system dependent. We should find a good set of defaults for our use-case (point-based lookups, few deletes, bulk writes at predictable times) and save them in an options file such that they can be tuned for each installation.

**Other information and links**

LesnyRumcajs:
[RocksDB tuning guide](https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide) for reference

lemmih:
Especially this section: https://github.com/facebook/rocksdb/wiki/RocksDB-Options-File

LesnyRumcajs:
@lemmih Unfortunately I don't think we can use this. RocksDB Rust wrapper is using the C API which lacks this. See [here](https://github.com/rust-rocksdb/rust-rocksdb/issues/568). Seems we'll need to implement it manually.

lemmih:
Sigh, that's a bummer.

PR #1479, opened 2022-03-15
parametrize current rocksdb settings

LesnyRumcajs:
**Summary of changes**
- Parametrized currently used `rocksdb` settings,
- refactored rocksdb tests a bit,
- some other cosmetic changes I couldn't resist,

Part of https://github.com/ChainSafe/forest/issues/1473, more settings and different defaults may come after some tinkering. Doesn't make sense to expose everything at once, there are quite a lot of different [knobs](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#) available.

### DIFF
--- a/node/db/tests/db_utils/mod.rs
+++ b/node/db/tests/db_utils/mod.rs
@@ -3,44 +3,35 @@
 
 #![cfg(feature = "rocksdb")]
 
-// Taken from
-// https://github.com/rust-rocksdb/rust-rocksdb/blob/master/tests/util/mod.rs
-use rocksdb::{Options, DB};
-use std::path::{Path, PathBuf};
-use std::time::{SystemTime, UNIX_EPOCH};
+use forest_db::rocks::{RocksDb, RocksDbConfig};
+use std::ops::Deref;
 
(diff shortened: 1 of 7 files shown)


node/db/tests/db_utils/mod.rs
```diff
@@ -3,44 +3,35 @@
...
-        DBPath {
-            path: PathBuf::from(path),
+        TempRocksDB {
+            db: RocksDb::open(&path, RocksDbConfig::default()).unwrap(),
+            _dir: dir,
         }
     }
 }
 
-impl Drop for DBPath {
```
lemmih, line 35: No Drop impl for `TempRocksDB`?

lemmih: Ah, the tempdir is automatically removed when it goes out of scope.

lemmih (approved):

node/db/tests/db_utils/mod.rs
LesnyRumcajs: I didn't really see any point in having this `destroy` to be honest, temporary dir that will obliterate itself automagically seems cleaner to me.

elmattic (approved):

PR #1494, opened 2022-03-22
Connor/default config location

connormullett:
**Summary of changes**
- Forest will look for a default config file in $XDG_CONFIG_HOME (~/.config) or in the users home directory if not available.
- If a config file is not passed to Forest when the node is started and a config file is not found, it will use a default configuration.

Closes 


lemmih:
Let's use https://crates.io/crates/directories/ instead of querying `XDG_CONFIG_HOME` directly.

lemmih (changes requested):
If `FOREST_CONFIG_PATH` isn't available, use https://crates.io/crates/directories/ to query the config path in a cross platform way.

LesnyRumcajs:
> If `FOREST_CONFIG_PATH` isn't available, use https://crates.io/crates/directories/ to query the config path in a cross platform way.

I didn't know this one, looks pretty good. Probably we should use it in some other places around Forest we want to strive for being cross-platform without containterization.

forest/src/cli/mod.rs
```diff
@@ -224,6 +217,31 @@ impl CliOpts {
     }
 }
 
+pub fn find_default_config() -> Option<Config> {
+    // check forest config path first, then look at other default locations
```
LesnyRumcajs, line 221: From a user perspective, it may be nice to check all those places and display a piece of information if e.g. `forest` will use config from `XDG_CONFIG_HOME` and not what is there in `HOME_DIR`. Otherwise, we could at least document it somewhere in README.

forest/src/cli/mod.rs
```diff
@@ -224,6 +217,31 @@ impl CliOpts {
...
+    } else if let Ok(root_config_path) = std::env::var("XDG_CONFIG_HOME") {
+        // determine if XDG_CONFIG_HOME is set for default locations, otherwise use home_dir
+        PathBuf::from(root_config_path + "/forest/config.toml")
+    } else {
+        // $XDG_CONFIG_HOME does not exist, check ~/.forest.toml
+        PathBuf::from(get_home_dir() + "/.forest/config.toml")
+    };
+
+    if path.exists() {
+        let toml = match read_file_to_string(&path) {
```
LesnyRumcajs, line 233: Perhaps you could omit `match` and `if let` later on with `?` and `ok()?`?

LesnyRumcajs (changes requested):

forest/src/cli/mod.rs
LesnyRumcajs: Not sure how others feel about it @lemmih @elmattic

elmattic:
> > If `FOREST_CONFIG_PATH` isn't available, use https://crates.io/crates/directories/ to query the config path in a cross platform way.
> 
> I didn't know this one, looks pretty good. Probably we should use it in some other places around Forest we want to strive for being cross-platform without containterization.

What are the pros of using `directories` here instead of the lower level `dirs`?

forest/src/cli/mod.rs
elmattic: Yes, I think it would help to have some `info` log that displays to the user the loaded config file location.

lemmih:
> > > If `FOREST_CONFIG_PATH` isn't available, use https://crates.io/crates/directories/ to query the config path in a cross platform way.
> > 
> > 
> > I didn't know this one, looks pretty good. Probably we should use it in some other places around Forest we want to strive for being cross-platform without containterization.
> 
> What are the pros of using `directories` here instead of the lower level `dirs`?

It's easier to follow OS conventions with `directories`. On mac and windows, we don't just want to create a `forest` folder in the config directory.

forest/src/cli/mod.rs
```diff
@@ -224,6 +217,31 @@ impl CliOpts {
...
+pub fn find_default_config() -> Option<Config> {
+    // check forest config path first, then look at other default locations
+    let path = if let Ok(config_file) = std::env::var("FOREST_CONFIG_PATH") {
+        PathBuf::from(config_file)
+    } else if let Ok(root_config_path) = std::env::var("XDG_CONFIG_HOME") {
+        // determine if XDG_CONFIG_HOME is set for default locations, otherwise use home_dir
+        PathBuf::from(root_config_path + "/forest/config.toml")
+    } else {
+        // $XDG_CONFIG_HOME does not exist, check ~/.forest.toml
+        PathBuf::from(get_home_dir() + "/.forest/config.toml")
```
elmattic, line 229: Would be better not to use strings to craft paths (see `Path` and `PathBuff` types).

forest/src/cli/mod.rs
elmattic: And in case of doubt the user still can add `config dump` to command-line to see the computed config.

forest/src/cli/mod.rs
connormullett: Gets tricky with Result<_> into Option<_>, this is the simplest way i can see it currently. I've refactored to DRY and reduce repeated code

forest/src/cli/mod.rs
connormullett: Added logging with config locations and errors that occured while trying to load the config. Resorts to default on error

forest/src/cli/mod.rs
connormullett: :+1:

forest/src/cli/mod.rs
```diff
@@ -158,17 +160,10 @@ impl CliOpts {
...
-                    // Parse and return the configuration file
-                    read_toml(&toml)?
-                } else {
-                    Config::default()
-                }
-            }
+            None => match find_default_config() {
+                Some(cfg) => cfg,
+                None => Config::default(),
+            },
```
LesnyRumcajs, line 166:
```suggestion
            None => find_default_config().unwrap_or_default()
```

LesnyRumcajs (commented):
Let's put the available default locations and their priority to the config section of README, what do you think?

forest/src/cli/mod.rs
connormullett: I think that's a good idea, will sneak it in now

lemmih (changes requested):

forest/src/cli/mod.rs
```diff
@@ -224,6 +219,53 @@ impl CliOpts {
...
+        );
+        let path = PathBuf::from(config_file);
+        if path.exists() {
+            return read_config_or_none(path);
+        }
+    };
+
+    if let Some(dir) = ProjectDirs::from("com", "ChainSafe", "Forest") {
+        let mut config_dir = dir.config_dir().to_path_buf();
+        config_dir.push("/forest/config.toml");
```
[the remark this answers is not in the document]
connormullett: Good catch, fixing

LesnyRumcajs (approved):
Thanks! It's really useful!

PR #1496, opened 2022-03-23
deps: simplify os dependencies

q9f:
Just testing if we need all of them.


lemmih:
Looks good. Seems like we didn't need all of those packages.

LesnyRumcajs (approved):

PR #1519, opened 2022-03-30
Use `anyhow::Error` where possible.

elmattic:
**Summary of changes**
- `vm/actor_interface/*` is now using `anyhow::Result` in all `load` methods

Closes 


ipld/blockstore/src/lib.rs
```diff
@@ -37,6 +37,10 @@ pub trait BlockStore: Store {
         Ok(self.read(cid.to_bytes())?)
     }
 
+    fn get_bytes2(&self, cid: &Cid) -> anyhow::Result<Option<Vec<u8>>> {
```
LesnyRumcajs, line 40: Those `name2` names are not very descriptive. Could we get rid of the original `get_bytes` to snatch the name? Alternatively, `get_bytes_anyhow` would provide more context.

vm/actor_interface/src/builtin/mod.rs
```diff
@@ -25,6 +25,51 @@ pub static BURNT_FUNDS_ACTOR_ADDR: &actorv0::BURNT_FUNDS_ACTOR_ADDR =
...
+                .context("Actor state doesn't exist in store")
+                .map(State::V2)
+        } else if $actor.code == *actorv0::$id {
+            use anyhow::Context;
+            $store
+                .get2(&$actor.state)?
+                .context("Actor state doesn't exist in store")
+                .map(State::V0)
+        } else {
+            anyhow::bail!("Unknown actor code {}", $actor.code)
```
LesnyRumcajs, line 68: I believe it would be more *gentle* to return an error (as an expression) but not force exit the function in which it was expanded. This way the macro could be used in a method that does not necessarily return `anyhow::Result`.

vm/actor_interface/src/builtin/power/mod.rs
```diff
@@ -139,77 +106,131 @@ impl State {
...
         match self {
-            State::V0(st) => Ok(st.miner_power(s, miner)?.map(From::from)),
-            State::V2(st) => Ok(st.miner_power(s, miner)?.map(From::from)),
-            State::V3(st) => Ok(st.miner_power(s, miner)?.map(From::from)),
-            State::V4(st) => Ok(st.miner_power(s, miner)?.map(From::from)),
-            State::V5(st) => Ok(st.miner_power(s, miner)?.map(From::from)),
-            State::V6(st) => Ok(st.miner_power(s, miner)?.map(From::from)),
+            State::V0(st) => {
+                let claim = st
+                    .miner_power(s, miner)
```
LesnyRumcajs, line 113: how about a small macro to reduce code duplication?

LesnyRumcajs (changes requested):
Great job in reducing some boilerplate code!

vm/actor_interface/src/builtin/mod.rs
lemmih: :+1:

vm/actor_interface/src/builtin/power/mod.rs
lemmih: Having a consistent set of macros for dealing with all the different actor versions would be great. I'll leave it for a future PR, though.

ipld/blockstore/src/lib.rs
lemmih: I believe we have to keep the original names until we've moved completely to FVM actors. I've renamed the functions to `*_anyhow` for now.

blockchain/state_manager/src/lib.rs
```diff
@@ -851,13 +855,9 @@ where
...
                     .ok_or_else(|| Error::Other("block must have parents".to_string()))?;
-                let parent: BlockHeader = self
-                    .blockstore()
-                    .get(parent_cid)
-                    .map_err(|e| Error::Other(e.to_string()))?
-                    .ok_or_else(|| {
-                        format!("Could not find parent block with cid {}", parent_cid)
-                    })?;
+                let parent: BlockHeader = self.blockstore().get(parent_cid)?.ok_or_else(|| {
+                    format!("Could not find parent block with cid {}", parent_cid)
```
LesnyRumcajs, line 859:
```suggestion
                    format!("Could not find parent block with cid {parent_cid}")
```

LesnyRumcajs (approved):

lemmih (approved):

blockchain/state_manager/src/lib.rs
lemmih: Nice.

PR #1523, opened 2022-03-30
Separate Chain DBs

### LINKED ISSUE #1418: Add network specific databases
**Issue summary**

Forest currently has one database and is adding support for calibnet at time of this issue's creation. As a quality of life feature, Forest will need to be able to have a database for each network available. Currently users/devs have to delete the database when switching networks, or changing the data_dir in config.toml. Being able to use our network features ( #[cfg(feature = "calibnet")]), we can automatically pick a database to use without having to delete the others.

**Task summary**

- [ ] use feature propogation to create and manage separate db paths based on network that was built

**Acceptance Criteria**

- [ ] when using forest, users should be able to switch freely between mainnet and calibnet without having to modify the config.toml or deleting their database and be able to start up where the nodes left on, on that network

connormullett:
**Summary of changes**
- Allows users switching between networks regularly to maintain separate blockstores.
- Also maintains separation between sled and rocksdb users such that at most (with 2 networks), there can be up to 4 different databases.

Closes #1418 

### DIFF
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -125,10 +125,11 @@ pub(super) async fn start(config: Config) {
 
     // Initialize database (RocksDb will be default if both features enabled)
     #[cfg(all(feature = "sled", not(feature = "rocksdb")))]
-    let db = db::sled::SledDb::open(format!(
-        "{}/{}/{}",
-        config.data_dir, &config.chain.name, "/sled"
-    ))
+    let db = db::sled::SledDb::open(
+        PathBuf::from(&config.data_dir)
+            .join(&config.chain.name)
+            .join("sled"),
+    )
     .expect("Opening SledDB must succeed");
 
     #[cfg(feature = "rocksdb")]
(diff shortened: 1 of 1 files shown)


forest/src/daemon.rs
```diff
@@ -125,12 +125,20 @@ pub(super) async fn start(config: Config) {
...
+    .expect("Opening SledDB must succeed");
 
     #[cfg(feature = "rocksdb")]
-    let db = db::rocks::RocksDb::open(PathBuf::from(&config.data_dir).join("db"), &config.rocks_db)
-        .expect("Opening RocksDB must succeed");
+    let db = db::rocks::RocksDb::open(
+        PathBuf::from(&config.data_dir)
+            .join(&config.chain.name)
+            .join("db"),
+        &config.rocks_db,
```
q9f, line 139: This is a breaking change. If we do not plan to offer a db migration, we should at least highlight it in the release notes.

q9f (changes requested):
works as intended, just change to PathBuf and it's good to go. :+1:

forest/src/daemon.rs
elmattic: Agree, we should definitely document this.

elmattic (approved):

PR #1530, opened 2022-04-01
log: don't override default filters (#1504)

### LINKED ISSUE #1504: Use default logging filters even if RUST_LOG is set.
**Issue summary**

Forest has a set of default logging filters. These filters are quite sensible but they're only used when the `RUST_LOG` environment variable is not set. Setting `RUST_LOG` to `chain_sync=debug` doesn't just change the logging setting for the `chain_sync` component, it unfortunately also clears all the default filters.

Forest should be changed to always start with the default set of filters and then allow `RUST_LOG` to augment them.

The logging filters are built in `forest/src/logger/mod.rs`.

jdjaustin:
**Summary of changes**
- don't override default filters when `RUST_LOG` environment variable is set

Closes #1504 .

Not sure if there is a default test suite; please advise best way to test if possible.

### DIFF
--- a/forest/src/logger/mod.rs
+++ b/forest/src/logger/mod.rs
@@ -5,20 +5,22 @@ use log::LevelFilter;
 
 pub(crate) fn setup_logger() {
     let mut logger_builder = pretty_env_logger::formatted_timed_builder();
+
+    // Assign default log level settings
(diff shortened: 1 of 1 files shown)


LesnyRumcajs (approved):
Looks good, quite useful first PR!

lemmih (approved):

PR #1533, opened 2022-04-11
Deserialize Typesafe NetworkVersion

### LINKED ISSUE #1498: Add typesafe version in configuration file
**Task summary**

This task is related to #1482.
It is possible to configure network chain using a `[chain]` section in the toml configuration file, ie:

```
[chain]
name = "calibnet"
bootstrap_peers = [
    "/dns4/bootstrap-0.calibration.fildev.network/tcp/1347/p2p/12D3KooWJkikQQkxS58spo76BYzFt4fotaT5NpV2zngvrqm4u5ow",
    "/dns4/bootstrap-1.calibration.fildev.network/tcp/1347/p2p/12D3KooWLce5FDHR4EX4CrYavphA5xS3uDsX6aoowXh5tzDUxJav",
    "/dns4/bootstrap-2.calibration.fildev.network/tcp/1347/p2p/12D3KooWA9hFfQG9GjP6bHeuQQbMD3FDtZLdW1NayxKXUT26PQZu",
    "/dns4/bootstrap-3.calibration.fildev.network/tcp/1347/p2p/12D3KooWMHDi3LVTFG8Szqogt7RkNXvonbQYqSazxBx41A5aeuVz"
]
block_delay_secs = 30
version_schedule = [
  { height = "Breeze", version = 1 },
  { height = "Smoke", version = 2 },
  # ...
]
height_infos = [
  { height = "Breeze", epoch = -1 },
  { height = "Smoke", epoch = -2 },
  { height = "Ignition", epoch = -3 },
  { height = "ActorsV2", epoch = 30 },
  { height = "Tape", epoch = 60 },
  # ...
]
```

A current limitation is that the version for a given height is expressed as an `u32` rather than a string. This means forest can panic at runtime if the conversion from `u32` to `NetworkVersion` is invalid.

We want that a non-existing version fail at config loading time. IOW we need to support for deserializing a string to `NetworkVersion`.

**Specification reference**

[Conversion code](https://github.com/ChainSafe/forest/blob/elmattic/network_selection/types/networks/src/lib.rs#L249)
[UpgradeInfo struct](https://github.com/ChainSafe/forest/blob/elmattic/network_selection/types/networks/src/lib.rs#L185)

[Network version](https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/version/mod.rs#L11)

connormullett:
**Summary of changes**
- Adds the required change to allow for typesafe deserialization of NetworkVersion for custom chain configurations

Closes #1498 

### DIFF
--- a/types/networks/src/lib.rs
+++ b/types/networks/src/lib.rs
@@ -242,7 +242,7 @@ impl ChainConfig {
             .iter()
             .find(|info| height == info.height)
             .map(|info| info.version)
-            .unwrap()
+            .expect("A network version should exist even if not specified in the config (a default exists).")
     }
 
     pub async fn get_beacon_schedule(


types/networks/src/lib.rs
```diff
@@ -302,3 +304,68 @@ impl Default for ChainConfig {
...
+            "v7" => Ok(NetworkVersion::V7),
+            "v8" => Ok(NetworkVersion::V8),
+            "v9" => Ok(NetworkVersion::V9),
+            "v10" => Ok(NetworkVersion::V10),
+            "v11" => Ok(NetworkVersion::V11),
+            "v12" => Ok(NetworkVersion::V12),
+            "v13" => Ok(NetworkVersion::V13),
+            "v14" => Ok(NetworkVersion::V14),
+            "v15" => Ok(NetworkVersion::V15),
+            _ => Ok(NetworkVersion::V15),
```
elmattic, line 340: Better produce an error here. We don't want to allow user to use an unexisting version or have erroneous version.
ie. `"nv16"` or `"V16"` should lead to a deserialization error and not return `NetworkVersion::V15`

elmattic (changes requested):

types/networks/src/lib.rs
```diff
@@ -302,3 +304,68 @@ impl Default for ChainConfig {
         }
     }
 }
+
+pub fn default_network_version() -> NetworkVersion {
+    NetworkVersion::V15
```
elmattic, line 309: Unless we have good reasons to, I think it would be better to have `NetworkVersion::V1` here.

elmattic (approved):

LesnyRumcajs (changes requested):

types/networks/src/lib.rs
```diff
@@ -302,3 +304,71 @@ impl Default for ChainConfig {
...
+        n.to_owned().serialize(serializer)
+    }
+}
+
+#[cfg(test)]
+pub mod test {
+    use super::*;
+
+    #[test]
+    pub fn test_serialize_upgrade_info() {
```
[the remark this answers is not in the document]
connormullett: sounds good, will add in the AM my time!

elmattic:
Also patch `ChainConfig::network_version` to remove last line and return directly the version.

types/networks/src/lib.rs
```diff
@@ -302,3 +301,108 @@ impl Default for ChainConfig {
...
+        "#;
+
+        assert_eq!(
+            remove_whitespace(&mut actual),
+            remove_whitespace(&mut expected.to_string())
+        );
+    }
+
+    #[test]
+    pub fn test_default_network_version_serailization() {
```
LesnyRumcajs, line 397:
```suggestion
    pub fn test_default_network_version_serialization() {
```

types/networks/src/lib.rs
```diff
@@ -302,3 +301,108 @@ impl Default for ChainConfig {
...
+    }
+
+    #[test]
+    pub fn test_deserialize_upgrade_info() {
+        let input = UpgradeInfo {
+            height: Height::Breeze,
+            version: NetworkVersion::V1,
+        };
+
+        let mut actual: String = toml::to_string(&input).unwrap();
```
LesnyRumcajs, line 383:
```suggestion
        let mut actual = toml::to_string(&input).unwrap();
```

types/networks/src/lib.rs
```diff
@@ -302,3 +301,108 @@ impl Default for ChainConfig {
...
+    {
+        n.to_owned().serialize(serializer)
+    }
+}
+
+#[cfg(test)]
+pub mod test {
+    use super::*;
+
+    fn remove_whitespace(s: &mut String) {
```
LesnyRumcajs, line 356: It'd be probably cleaner to have a method that returns a copy without whitespace instead of mutating the parameter. Then we could get rid of all the `mut`s.

types/networks/src/lib.rs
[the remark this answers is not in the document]
LesnyRumcajs: I was thinking also about a test where the serialization fails i.e. `height = "Cthulhu"`

LesnyRumcajs (changes requested):

types/networks/src/lib.rs
connormullett: Good catch lol

types/networks/src/lib.rs
```diff
@@ -236,14 +238,11 @@ impl ChainConfig {
...
-        let version = self
-            .version_schedule
+        self.version_schedule
             .iter()
             .find(|info| height == info.height)
             .map(|info| info.version)
-            .unwrap();
-
-        NetworkVersion::try_from(version).unwrap()
+            .unwrap()
```
q9f, line 245: I'd always try to avoid panics by either providing a default or formulating an expectation that helps to identify panics later.

```suggestion
            .unwrap_or(NetworkVersion::V1)
```

types/networks/src/lib.rs
```diff
@@ -302,3 +301,136 @@ impl Default for ChainConfig {
...
+            NetworkVersion::V6 => "V6",
+            NetworkVersion::V7 => "V7",
+            NetworkVersion::V8 => "V8",
+            NetworkVersion::V9 => "V9",
+            NetworkVersion::V10 => "V10",
+            NetworkVersion::V11 => "V11",
+            NetworkVersion::V12 => "V12",
+            NetworkVersion::V13 => "V13",
+            NetworkVersion::V14 => "V14",
+            NetworkVersion::V15 => "V15",
```
q9f, line 364: Deserializer uses `"v15"` and serializer a capital `"V15"` - is that on purpose?

connormullett: the deserializer converts it to lowercase before matching against network versions. Just makes it so v10 and V10 are valid

types/networks/src/lib.rs
connormullett: makes sense, the unwrap() initially was assumed because a network height should exist even if not specified in the config (a default exists). The redundancy makes sense here as a just in case. done

q9f: Yes, I saw that. I just wanted to discourage the use of a plain `unwrap()`s. In that case the following would alos look good:

```suggestion
            .expect("A network height should exist even if not specified in the config (a default exists).")
```

connormullett: That make sense. I'll commit your suggestion and it should be g2g

elmattic: Good remark but I wouldn't apply it here. `.except` with an helpinh message for the dev or node operator should be used instead here.
This should panics because it means you forget to add an `UpgradeInfo` entry in the `version_schedule` (code or .toml).
e.g. some epoch leads to `Height::Skyr` but you forget to add the entry `{ height = "Skyr", version = "V16" }`.

types/networks/src/lib.rs
```diff
@@ -236,14 +238,11 @@ impl ChainConfig {
             .map(|info| info.height)
             .unwrap_or(Height::Breeze);
```
elmattic, line 239: nit: you could use `unwrap_or_default()`.

LesnyRumcajs (approved):

q9f (approved):

PR #1541, opened 2022-06-20
WIP 1534 LLVM code coverage is more precise than tarpaulin

tyshko5:
**Summary of changes**
- 

Closes 


lemmih:
I'm not 100% sure we can use LLVM code coverage since we depend on packages that use webassembly. If it doesn't work, let's document why and then return to it in the future.

PR #1558, opened 2022-06-21
PR 1534 LLVM code coverage

tyshko5:
**Summary of changes**
- 

Closes 

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -67,14 +67,21 @@ jobs:
           toolchain: nightly
           target: wasm32-unknown-unknown
           override: true
+      - name: Add component llvm-tools-preview
+        run: rustup component add llvm-tools-preview --toolchain nightly-x86_64-unknown-linux-gnu
       - name: Rust Cache
         uses: Swatinem/rust-cache@v1.3.0
-      - name: Run Tarpaulin
-        uses: actions-rs/tarpaulin@v0.1
-        with:
-          args: '--locked --jobs 16 --timeout 600 --skip-clean -- --test-threads 16'
+      - name: Run cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Generate code coverage
+        run: cargo llvm-cov --workspace --lcov --output-path lcov.info
+        # with:
+        #   args: '--locked --jobs 16 --timeout 600 --skip-clean -- --test-threads 16'
       - name: Upload CodeCov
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -67,14 +67,21 @@ jobs:
...
-      - name: Run Tarpaulin
-        uses: actions-rs/tarpaulin@v0.1
-        with:
-          args: '--locked --jobs 16 --timeout 600 --skip-clean -- --test-threads 16'
+      - name: Run cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Generate code coverage
+        run: cargo llvm-cov --workspace --lcov --output-path lcov.info
+        # with:
+        #   args: '--locked --jobs 16 --timeout 600 --skip-clean -- --test-threads 16'
```
lemmih, line 79: `cargo llvm-cov` takes `--jobs` parameter. We might as well set it to 16.

lemmih: Oh wait, it defaults to the number of available cpus. Yeah, let's leave it then.

lemmih: Let's just remove the commented-out old arguments since we're not using them anymore.

LesnyRumcajs:
Any high-level wrap-up on the differences and how is it better than `tarpaulin`? Seems the coverage will drop, is this more correct than the current solution?

lemmih:
> Any high-level wrap-up on the differences and how is it better than `tarpaulin`? Seems the coverage will drop, is this more correct than the current solution?

llvm-cov instruments the source code and will always generate 100% accurate results, even if optimizations are enabled. Tarpaulin interrupts the program every so often to look at the call-stack. It is not 100% accurate at the best of times and optimizations ruin the results entirely. The reason we didn't use it earlier is because it just became stable in rust-1.60.

LesnyRumcajs:
@lemmih sounds good, seems the coverage dropped because more lines are now taken into consideration.

LesnyRumcajs (changes requested):
Looks good, but this file needs to be removed: https://github.com/ChainSafe/forest/blob/main/.tarpaulin.toml

lemmih:
> Looks good, but this file needs to be removed: https://github.com/ChainSafe/forest/blob/main/.tarpaulin.toml

Ah, right. @tyshko5 could you see to that?

LesnyRumcajs (approved):

Issue #1536, opened 2022-06-08
Forest maintenance tasks


lemmih:
**To be removed**

 - [ ] Remove all use of `Box<dyn Error>` in favor of `anyhow::Error`
 - [ ] Many occurrences of `Box<dyn Error>` and aliased as `Box<dyn StdError>`. Also replace these occurrences with `anyhow::Error`, and replace the final occurrences of `Box<dyn Error>` in `map.rs` that are associated with functions that currently utilize occurrences of `Box<dyn StdError>`.

**To be refactored**
 - [ ] Simplify the code in `vm_circ_supply.rs` now that we don't have to support bootstrapping from genesis
   - [x] Replace use of `lazy_static` with constant data.
   - [ ] Remove all use of `OnceCell`

PR #1571, opened 2022-06-23
[WIP]  Subset of Forest Maintenance Tasks (Part 2)

jdjaustin:
**Summary of changes**
**To be removed**
- [ ] Remove all use of `Box<dyn Error>` in favor of `anyhow::Error`

Closes corresponding task in #1536 

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3350,6 +3350,7 @@ dependencies = [
 name = "forest_libp2p"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "async-std",
  "async-trait",
  "asynchronous-codec",
@@ -5874,6 +5875,7 @@ dependencies = [
 name = "networks"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "beacon",
  "fil_clock",
  "fil_types",
@@ -7151,6 +7153,7 @@ name = "rpc-api"
 version = "0.1.0"
 dependencies = [
  "actor_interface",
+ "anyhow",
  "async-std",
  "beacon",
  "chain",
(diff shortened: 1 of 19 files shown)


jdjaustin:
Many occurrences of `Box<dyn Error>` are aliased as `Box<dyn StdError>` and changing all these occurrences would probably best be handled under a separate PR, to keep PR sizes manageable. The final occurrences of `Box<dyn Error>` in `map.rs` are associated with the occurrences of `Box<dyn StdError>` in `map.rs`. Will capture this either under a seprate issue or by modifying issue #1536 .

lemmih (changes requested):

LesnyRumcajs (changes requested):
@lemmih is right, I'd extend it to all other places where you changed `?` to `unwrap`. In general, if a method returns a `Result`, I don't expect it to panic. Please apply here the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) for poor me. :)

jdjaustin:
What do you guys suggest instead of using unwrap here? I had to remove the `?` due to compiler errors after changing `Box<dyn Error>` to `anyhow::Error`. If I leave out `.unwrap()`, the compiler gives me a warning about an unused Result.

LesnyRumcajs:
> What do you guys suggest instead of using unwrap here? I had to remove the `?` due to compiler errors after changing `Box<dyn Error>` to `anyhow::Error`. If I leave out `.unwrap()`, the compiler gives me a warning about an unused Result.

What do you think about `.map_err`? Have a look at an example here: https://github.com/filecoin-project/builtin-actors/blob/6e2a4eaa6d14d53f841b93bd84621455e4778a1c/actors/miner/src/sector_map.rs#L26

LesnyRumcajs (changes requested):

LesnyRumcajs (approved):

PR #1582, opened 2022-06-27
Refactor chain-specific database location code

### LINKED ISSUE #1580: Metrics can't find chain-specific database.
**Issue summary**

We recently moved the database into a chain-specific folder but the metrics aren't aware of this. We should refactor the code so there is only a single truth for the folders that we use.

Ie. the below two snippets of code should use the same logic for finding the database:

```rust
   let prometheus_server_task = task::spawn(metrics::init_prometheus(
        (format!("127.0.0.1:{}", config.metrics_port))
            .parse()
            .unwrap(),
        PathBuf::from(&config.data_dir)
            .join("db")
            .into_os_string()
            .into_string()
            .expect("Failed converting the path to db"),
    ));
```

```rust
#[cfg(feature = "rocksdb")]
    let db = db::rocks::RocksDb::open(
        PathBuf::from(&config.data_dir)
            .join(&config.chain.name)
            .join("db"),
        &config.rocks_db,
    )
    .expect("Opening RocksDB must succeed");
```

tyshko5:
**Summary of changes**
- 
As approach changed from db to chain, I've added chain_path function, so that the single source of truth for various db patches is established. Therefore db_path and sled_path were created

Closes #1580 

### DIFF
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -105,8 +105,7 @@ pub(super) async fn start(config: Config) {
         (format!("127.0.0.1:{}", config.metrics_port))
             .parse()
             .unwrap(),
-        PathBuf::from(&config.data_dir)
-            .join("db")
+        db_path(&config)
             .into_os_string()
             .into_string()
             .expect("Failed converting the path to db"),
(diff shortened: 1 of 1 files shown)


lemmih (approved):

lemmih:
Code looks good. Could you add a summary of the changes to the PR?

lemmih (approved):

LesnyRumcajs (approved):

PR #1595, opened 2022-06-28
Update proof parameters file to most recent version.

lemmih:
**Summary of changes**
- Some parameter files were apparently implicit in the past but have to be explicit now. Not exactly sure but this update should fix the sync issues.

### DIFF
--- a/utils/paramfetch/src/parameters.json
+++ b/utils/paramfetch/src/parameters.json
@@ -1,4 +1,74 @@
 {
+    "v28-empty-sector-update-merkletree-poseidon_hasher-8-0-0-61fa69f38b9cc771ba27b670124714b4ea77fbeae05e377fb859c4a43b73a30c.params": {
+        "cid": "Qma5WL6abSqYg9uUQAZ3EHS286bsNsha7oAGsJBD48Bq2q",
+        "digest": "c3ad7bb549470b82ad52ed070aebb4f4",
+        "sector_size": 536870912
+    },
+    "v28-empty-sector-update-merkletree-poseidon_hasher-8-0-0-61fa69f38b9cc771ba27b670124714b4ea77fbeae05e377fb859c4a43b73a30c.vk": {
+        "cid": "QmUa7f9JtJMsqJJ3s3ZXk6WyF4xJLE8FiqYskZGgk8GCDv",
+        "digest": "994c5b7d450ca9da348c910689f2dc7f",
+        "sector_size": 536870912
+    },
+    "v28-empty-sector-update-merkletree-poseidon_hasher-8-0-0-92180959e1918d26350b8e6cfe217bbdd0a2d8de51ebec269078b364b715ad63.params": {
+        "cid": "QmQiT4qBGodrVNEgVTDXxBNDdPbaD8Ag7Sx3ZTq1zHX79S",
+        "digest": "5aedd2cf3e5c0a15623d56a1b43110ad",
+        "sector_size": 8388608
+    },
+    "v28-empty-sector-update-merkletree-poseidon_hasher-8-0-0-92180959e1918d26350b8e6cfe217bbdd0a2d8de51ebec269078b364b715ad63.vk": {
+        "cid": "QmdcpKUQvHM8RFRVKbk1yHfEqMcBzhtFWKRp9SNEmWq37i",
+        "digest": "abd80269054d391a734febdac0d2e687",
+        "sector_size": 8388608
+    },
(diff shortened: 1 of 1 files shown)


elmattic:
Good! What about removing `"v28-fil-inner-product-v1.srs"` entry as well?

lemmih:
> Good! What about removing `"v28-fil-inner-product-v1.srs"` entry as well?

It failed to sync for me without it.

elmattic:
> > Good! What about removing `"v28-fil-inner-product-v1.srs"` entry as well?
> 
> It failed to sync for me without it.

I managed to sync without but maybe we can keep it and be conservative.

lemmih:
> > > Good! What about removing `"v28-fil-inner-product-v1.srs"` entry as well?
> > 
> > 
> > It failed to sync for me without it.
> 
> I managed to sync without but maybe we can keep it and be conservative.

It works for you if you remove `v28-fil-inner-product-v1.srs` from the parameters json file and delete `/var/tmp/filecoin-proof-parameters/`?

elmattic:
> > > > Good! What about removing `"v28-fil-inner-product-v1.srs"` entry as well?
> > > 
> > > 
> > > It failed to sync for me without it.
> > 
> > 
> > I managed to sync without but maybe we can keep it and be conservative.
> 
> It works for you if you remove `v28-fil-inner-product-v1.srs` from the parameters json file and delete `/var/tmp/filecoin-proof-parameters/`?

My bad, I forget to recompile the client after my change (didn't saw the `include_str`).

elmattic:
> > > > Good! What about removing `"v28-fil-inner-product-v1.srs"` entry as well?
> > > 
> > > 
> > > It failed to sync for me without it.
> > 
> > 
> > I managed to sync without but maybe we can keep it and be conservative.
> 
> It works for you if you remove `v28-fil-inner-product-v1.srs` from the parameters json file and delete `/var/tmp/filecoin-proof-parameters/`?

Yes, it works for me and the file `v28-fil-inner-product-v1.srs` is still downloaded even though
 I have removed the entry in `parameters.json `.

LesnyRumcajs:
@lemmih @elmattic any wrap-up on why it caused an error, where in the system the artifacts were persisted (and grabbed by Forest)? We should understand this in full lest we risk it happening again. And perhaps add some assertions if it's feasible.

On a testing side note, once the nightly check machine is back I'll change it to not build Forest but to take the image from the registry, ensuring clean state. We would be able to catch this class of bugs sooner.

elmattic:
Can confirm it's not working without `v28-fil-inner-product-v1.srs` (My bad, I was modifying the wrong repo previously). Accepting your commit.

elmattic (approved):

lemmih:
Brief post-mortem:

Guillaume and I were able to sync Forest to mainnet but Hubert saw state root mismatches with the exact same code. No matter how many times we cleared the database and built Forest from scratch, the two different outcomes wouldn't go a way.

We faultily assumed all our data was stored in `$DATA_DIR` and clearing this folder would give us identical environments. It turns out that important parameter files are stored in `/var/tmp/filecoin-proof-parameters`. These files are downloaded from IPFS by forest according to a builtin `parameters.json` file. This parameters file is part of rust-fil-proofs[1] and was changed five months ago in their repository. We aren't aware of this change (or even of the existence of this file) and didn't update our copy. However, anyone who had tried to sync before February 2nd would have already downloaded the needed files and wouldn't see any failures. The docker image downloads the parameters on demand which explains why it couldn't sync to mainnet while both Guillaume and I still could.

It's a significant problem that a parameter mishap leads to such hard-to-debug state root mismatches rather than fail with informative error messages. Skimming our code as well as that of rust-fil-proofs, I don't see any trivial ways of checking whether we have the most up-to-date parameters.

[1]: https://github.com/filecoin-project/rust-fil-proofs/blob/master/parameters.json

LesnyRumcajs (approved):
LGTM as a workaround, but this JSON is not really maintained by us, so in my opinion, it should be fetched during e.g. build phase or the relevant repo should be added as a submodule.

lemmih:
> LGTM as a workaround, but this JSON is not really maintained by us, so in my opinion, it should be fetched during e.g. build phase or the relevant repo should be added as a submodule.

Yeah. We can't use the un-edited json file, though. We need to add the entry for "v28-fil-inner-product-v1.srs" (reasons still unknown).

lemmih:
I'll merge this PR but this is not done. We 100% need something better.

PR #1610, opened 2022-07-04
improve docker examples

LesnyRumcajs:
**Summary of changes**
- used safer env variable expansion instead of `~`. It seems the latter not supported for mounting volumes in some docker (or shell?) versions. Thanks @lemmih 
- added an example using host db


lemmih:
I guess rust 1.62 is out with new lints. :)

lemmih (approved):

elmattic (approved):

tyshko5 (approved):

PR #1619, opened 2022-07-06
Keep proof parameters in the forest data directory 

### LINKED ISSUE #1596: Keep proof parameters in the forest data directory
**Issue summary**

Proof parameters are automatically downloaded to `/var/tmp/filecoin-proof-parameters`. Ideally we would want them together with the rest of the forest data in `$DATA_DIR/filecoin-proof-parameters/`. 

- [ ] Update path in `utils/paramfetch/src/lib.rs`
- [ ] Update `FIL_PROOFS_PARAMETER_CACHE` environment variable.

https://github.com/filecoin-project/rust-fil-proofs#parameter-file-location

tyshko5:
**Summary of changes**
- 

Closes #1596

### DIFF
--- a/utils/paramfetch/src/lib.rs
+++ b/utils/paramfetch/src/lib.rs
@@ -47,6 +47,8 @@ struct ParameterData {
     sector_size: u64,
 }
 
+// Proof parameter file directory. Defaults to %DATA_DIR/filecoin-proof-parameters unless
+// the FIL_PROOFS_PARAMETER_CACHE environment variable is set.
 fn param_dir(data_dir: &Path) -> PathBuf {
     std::env::var(PathBuf::from(DIR_ENV))
         .map(PathBuf::from)
@@ -61,6 +63,8 @@ fn param_dir(data_dir: &Path) -> PathBuf {
 /// validity. For consistency, Forest will prefer to download the files it's local data
 /// directory. To this end, the FIL_PROOFS_PARAMETER_CACHE environment variable is
 /// updated before the parameters are downloaded.
+///
+/// More information available here: https://github.com/filecoin-project/rust-fil-proofs#parameter-file-location
 pub fn set_proofs_parameter_cache_dir_env(data_dir: &Path) {
     std::env::set_var(DIR_ENV, param_dir(data_dir));
 }


lemmih:
Also set the `FIL_PROOFS_PARAMETER_CACHE` environment variable.

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,22 @@ struct ParameterData {
 }
 
 #[inline]
-fn param_dir() -> String {
-    std::env::var(DIR_ENV).unwrap_or_else(|_| PARAM_DIR.to_owned())
+fn param_dir(mut data_dir: PathBuf) -> String {
+    std::env::var(PathBuf::from(DIR_ENV)).unwrap_or_else(|_| {
+        data_dir.push(PARAM_DIR);
+        data_dir.to_str().unwrap().to_string()
+    })
```
lemmih, line 55: How about returning a `PathBuf` rather than a `String`?

lemmih: Indeed. So we either convert a `String` to a `PathBuf` or a `PathBuf` to a `String`. Which conversion can we do without the possibility of a panic?

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,23 @@ struct ParameterData {
 }
 
 #[inline]
-fn param_dir() -> String {
-    std::env::var(DIR_ENV).unwrap_or_else(|_| PARAM_DIR.to_owned())
+fn param_dir(mut data_dir: PathBuf) -> String {
```
LesnyRumcajs, line 51: Why is there so much cloning whenever `param_dir` is invoked? IIUC the lazy part will be executed at most once so better allocate once only there, the rest should get passed by reference. Or am I missing something?

Another thing, this function doesn't seem to be `inline`-friendly anymore.

LesnyRumcajs (changes requested):

utils/paramfetch/src/lib.rs
tyshko5: clippy didn't like having &PathBuf

lemmih:
@tyshko5 I don't think setting an environment variable in `param_dir()` is appropriate. Try refactoring the code such that it will be unsurprising to people reading it 10 years from now. So don't update state in functions that appear to be read-only and try to avoid panics. If you run out of ideas for how to refactor this code, ask for advice in the #fil-devs channel.

utils/paramfetch/src/lib.rs
```diff
@@ -117,18 +128,26 @@ pub async fn get_params(
...
 }
 
 async fn fetch_verify_params(
+    data_dir: PathBuf,
     name: &str,
     info: Arc<ParameterData>,
     mb: Option<Arc<MultiBar<Stdout>>>,
 ) -> Result<(), anyhow::Error> {
-    let path: PathBuf = [&param_dir(), name].iter().collect();
+    let path: PathBuf = [param_dir(data_dir.clone()).to_str().unwrap(), name].iter().collect();
```
lemmih, line 150: Refactor to avoid the `unwrap`. In general, if you use `unwrap`, you have to write a comment with a proof of why it will not panic.

forest/src/cli/fetch_params_cmd.rs
```diff
@@ -32,7 +33,11 @@ impl FetchCommands {
             panic!("Sector size option must be chosen. Choose between --all, --keys, or <size>");
         };
 
-        get_params_default(sizes, self.verbose).await.unwrap();
+        set_dir_env(config.data_dir.clone());
```
lemmih, line 36: There's no need to set the environment here in the fetch command. It's only used in the forest daemon.

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,28 @@ struct ParameterData {
...
+        PathBuf::from(path_buf_string)
+    } else {
+        data_dir.push(PARAM_DIR);
+        data_dir
+    }
+}
+
+pub fn set_dir_env(data_dir: PathBuf) {
+    std::env::set_var(DIR_ENV, data_dir.clone());
 }
```
lemmih, line 62: This seems wrong. We don't want to set `FIL_PROOFS_PARAMETER_CACHE` to be the data dir. It should be `%DATA/PARAM_DIR`.

Also, try a more descriptive name like `set_proofs_parameter_cache_dir_env()`.

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,28 @@ struct ParameterData {
...
 #[inline]
-fn param_dir() -> String {
-    std::env::var(DIR_ENV).unwrap_or_else(|_| PARAM_DIR.to_owned())
+fn param_dir(mut data_dir: PathBuf) -> PathBuf {
+    if let Ok(path_buf_string) = std::env::var(PathBuf::from(DIR_ENV)) {
+        PathBuf::from(path_buf_string)
+    } else {
+        data_dir.push(PARAM_DIR);
+        data_dir
+    }
```
lemmih, line 57: Perhaps something like:

```suggestion
    std::env::var(PathBuf::from(DIR_ENV))
        .map(PathBuf::from)
        .unwrap_or_else(|_| data_dir.join(PARAM_DIR))
```

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,28 @@ struct ParameterData {
...
+}
+
+pub fn set_dir_env(data_dir: PathBuf) {
+    std::env::set_var(DIR_ENV, data_dir.clone());
 }
 
 /// Get proofs parameters and all verification keys for a given sector size given
 /// a param JSON manifest.
 pub async fn get_params(
+    data_dir: PathBuf,
```
lemmih, line 67: `data_dir` is immutable so we don't need to own it. Try with a `&Path` instead.

lemmih (changes requested):

utils/paramfetch/src/lib.rs
tyshko5: Fixed

utils/paramfetch/src/lib.rs
tyshko5: Done through &Path

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,25 @@ struct ParameterData {
...
-    std::env::var(DIR_ENV).unwrap_or_else(|_| PARAM_DIR.to_owned())
+fn param_dir(data_dir: &Path) -> PathBuf {
+    std::env::var(PathBuf::from(DIR_ENV))
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| data_dir.join(PARAM_DIR))
+}
+
+pub fn set_proofs_parameter_cache_dir_env(data_dir: PathBuf) {
+    std::env::set_var(DIR_ENV, data_dir.join(PARAM_DIR));
 }
```
lemmih, line 59: If `FIL_PROOFS_PARAMETER_CACHE` is already set then `param_dir()` would return one thing and `set_proofs_parameter_cache_dir_env` would set the environment variable to something completely different. That's not what we want. If `FIL_PROOFS_PARAMETER_CACHE` is present, it should be honored.

```suggestion
pub fn set_proofs_parameter_cache_dir_env(data_dir: PathBuf) {
    std::env::set_var(param_dir(data_dir));
}
```

utils/paramfetch/src/lib.rs
```diff
@@ -117,18 +127,21 @@ pub async fn get_params(
...
 async fn fetch_verify_params(
+    data_dir: &Path,
     name: &str,
     info: Arc<ParameterData>,
     mb: Option<Arc<MultiBar<Stdout>>>,
 ) -> Result<(), anyhow::Error> {
-    let path: PathBuf = [&param_dir(), name].iter().collect();
+    let mut path: PathBuf = param_dir(data_dir);
+    path.push(name);
     let path: Arc<Path> = Arc::from(path.as_path());
```
lemmih, line 145: Could you refactor this to avoid a mutable variable?

tyshko5: Refactored

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,25 @@ struct ParameterData {
...
 #[inline]
-fn param_dir() -> String {
-    std::env::var(DIR_ENV).unwrap_or_else(|_| PARAM_DIR.to_owned())
+fn param_dir(data_dir: &Path) -> PathBuf {
+    std::env::var(PathBuf::from(DIR_ENV))
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| data_dir.join(PARAM_DIR))
+}
+
+pub fn set_proofs_parameter_cache_dir_env(data_dir: PathBuf) {
```
lemmih, line 57: Does `data_dir` have to be owned?

lemmih:
Code looks okay. I'll verify that it works as expected tomorrow.

forest/src/daemon.rs
```diff
@@ -158,8 +158,10 @@ pub(super) async fn start(config: Config) {
 
     sync_from_snapshot(&config, &state_manager).await;
 
+    set_proofs_parameter_cache_dir_env(&config.data_dir);
```
LesnyRumcajs, line 161: This was quite an unexpected behaviour that took some time to pinpoint. It would be nice to provide a helpful comment explaining why we actually need to do this.

utils/paramfetch/src/lib.rs
```diff
@@ -48,18 +48,25 @@ struct ParameterData {
 }
 
 #[inline]
-fn param_dir() -> String {
-    std::env::var(DIR_ENV).unwrap_or_else(|_| PARAM_DIR.to_owned())
+fn param_dir(data_dir: &Path) -> PathBuf {
```
LesnyRumcajs, line 51: Let's add some helpful method documentation like for `get_params`. Maybe adding a verb to the method name would also improve its readability (`read_params_dir`?).

LesnyRumcajs (approved):
Some documentation would be extra helpful, otherwise LGTM!

lemmih (approved):

PR #1623, opened 2022-07-07
chore:docs:install steps

ZenGround0:
**Summary of changes**
Building forest from source with `make install` as specified in the readme didn't quite work for me.  I had to setup my rust toolchain in a specific way to get this to work.  At first I was thinking of putting this toolchain stuff in the makefile but that is probably doing more than user's want to delegate to the makefile so I just added a note to the readme.


ZenGround0:
Oh duh looks like I completely missed that.  I think this should get closed down then since dependencies is the right place for that.

Issue #1601, opened 2022-06-29
Nightly check: use Forest docker image


LesnyRumcajs:
**Issue summary**

Given after CI is passed we should be confident of the build itself, we can skip it in the nightly check and try to sync with a docker image / docker-compose setup. This way we also ensure that local worker cache/settings/changes are not distorting the results.

**Task summary**

- [ ] Modify nightly check scripts to use docker image (`latest` tag will suffice)
- [ ] Install them on the nightly check machine

**Acceptance Criteria**

- [ ] Forest is using the docker image in the nightly check and not building the binary
- [ ] GH has some downtime from time to time, so it would be good to try a few times to `pull` the image (without some periods of sleep) before reporting an error.

**Other information and links**

PR #1627, opened 2022-07-11
sync check in docker

LesnyRumcajs:
**Summary of changes**
- Introduced `docker-compose` suite to continuously sync Forest on both mainnet and calibnet. This made `nightly_check` obsolete. 
- After the check, Forest will still be validating blocks until a new tag is pushed.

In essence, the setup should:
* Run containerized Forest on Mainnet and Calibnet on the same machine,
* Whenever a new `latest` tag is pushed (i.e. on merge to `main`), all relevant containers will get restarted. This is thanks to `watchtower`.
* the `tester` container is there to check Forest metrics. The check it provides is re-applied on restart.

Closes https://github.com/ChainSafe/forest/issues/1601
Closes https://github.com/ChainSafe/forest/issues/1641

I don't like the manual part where one has to uncomment code on the first run. Probably something we can iterate over later, we'd need to allow entry point override in the Forest container which shouldn't be a big deal. On the plus side, afterwards, the mount is not required.


LesnyRumcajs:
A continuation of this is https://github.com/ChainSafe/forest/issues/1695.

lemmih (approved):

tyshko5 (approved):

Issue #1631, opened 2022-07-11
Add a network policy to the chain config structure.


lemmih:
**Issue summary**

The `ChainConfig` structure should include a [Policy](https://docs.rs/fil_actors_runtime/8.0.0/fil_actors_runtime/runtime/policy/struct.Policy.html) with values appropriate for the network type (calibnet or mainnet).

The policy values can be deduced from https://github.com/filecoin-project/builtin-actors/blob/master/runtime/build.rs

**Other information and links**

PR #1642, opened 2022-07-12
Add Policy struct to ChainConfig

elmattic:
**Summary of changes**
- Add network Policy struct into ChainConfig
- Fix existing code to use correct chain Policy

Closes https://github.com/ChainSafe/forest/issues/1631


forest/src/cli/config_cmd.rs
```diff
@@ -42,6 +42,7 @@ mod tests {
 
         let actual_config: Config = toml::from_str(std::str::from_utf8(sink.buffer()).unwrap())
             .expect("Invalid configuration!");
-        assert_eq!(actual_config, expected_config);
+
+        assert!(expected_config == actual_config);
```
LesnyRumcajs, line 46: Hm, why?

types/networks/src/lib.rs
```diff
@@ -134,14 +138,138 @@ struct DrandPoint<'a> {
...
     pub bootstrap_peers: Vec<String>,
     pub block_delay_secs: u64,
     pub version_schedule: Vec<UpgradeInfo>,
     pub height_infos: Vec<HeightInfo>,
+    #[serde(default = "default_policy")]
+    #[serde(with = "serde_policy")]
+    pub policy: Policy,
+}
+
+impl PartialEq for ChainConfig {
```
LesnyRumcajs, line 154: That's kind of asking for trouble in the future (I'd imagine more fields will come). Probably as a really temporary workaround, it's fine, but should we open a PR adding a proper `PartialEq, Clone` to builtin-actors? I mean, if it's not exposed then it's a really bad API.

forest/src/cli/config_cmd.rs
elmattic: This is to avoid to implement `Debug`, needed by `assert_eq`...

LesnyRumcajs: Makes sense. I added the `Debug` in the linked PR as well.

LesnyRumcajs (approved):
Approving but once `Policy` API is improved in builtin-actors we should revisit those changes. Perhaps a comment somewhere with an issue would be helpful.

types/networks/src/lib.rs
```diff
@@ -134,14 +138,81 @@ struct DrandPoint<'a> {
...
+    fn eq(&self, other: &Self) -> bool {
```
connormullett, line 155: Can we add a FIXME here for when the builtin-actors PR is good?

connormullett (changes requested):
Looks good! Would like a fixme so we don't lose where that big ole chunky conditional is and what its being blocked by

PR #1652, opened 2022-07-13
Simplify VM code and move towards using `fvm::StateTree`

lemmih:
**Summary of changes**
- Simplify error handling.
- Use `fvm::StateTree` when possible.

Closes 


blockchain/state_manager/src/lib.rs
```diff
@@ -332,9 +332,8 @@ where
     {
         let db = self.blockstore_cloned();
         let lb_wrapper = SMLookbackWrapper {
-            sm: self,
-            store: self.blockstore(),
-            tipset,
+            sm: self.clone(),
```
LesnyRumcajs, line 335: how costly is this clone?

blockchain/state_manager/src/lib.rs
```diff
@@ -1441,36 +1438,25 @@ pub struct MiningBaseInfo {
...
-    }
-
     fn chain_epoch_root(&self) -> Box<dyn Fn(ChainEpoch) -> Cid> {
         let sm = self.sm.clone();
         let tipset = self.tipset.clone();
         Box::new(move |round| {
             let (_, st) =
                 task::block_on(sm.get_lookback_tipset_for_round::<V>(tipset.clone(), round))
-                    .unwrap();
+                    .expect("Internal Error. Failed to find root CID for specific epoch.");
```
LesnyRumcajs, line 1459: What do you think about adding the epoch value in the error message?

blockchain/state_manager/src/lib.rs
lemmih: It's an `Arc` so it's essentially free.

blockchain/state_manager/src/lib.rs
lemmih: Done. The cost of allocating the String should be minor and eventually the code will be refactored to pass through the errors instead of panicking.

blockchain/state_manager/src/lib.rs
LesnyRumcajs: `self` here is an instance of `StateManager` which has a quite bit of fields, a cache included, right? Or am I missing something?

LesnyRumcajs: Ah no, it's a funny self, my bad.

LesnyRumcajs (approved):

blockchain/state_manager/src/lib.rs
lemmih: Yep, a funny self. :)

lemmih: Maybe I should change it to `Arc::clone(self)` just to be more explicit.

LesnyRumcajs: I guess it wouldn't hurt and it would make this piece of code less surprising. I'm good with both approaches.

PR #1666, opened 2022-07-14
Print a warning if executing a message fails with an internal error

### LINKED ISSUE #1649: Print a warning if executing a message fails with an internal error.
**Issue summary**

Right now we silently ignore all types of errors from the actors:
https://github.com/ChainSafe/forest/blob/b9a5ff48ad8e708c53f5312d46736babea7d1c1b/vm/interpreter/src/vm.rs#L333-L338

If the execution fails with an internal error (such as `USR_ASSERTION_FAILED`), a warning should be logged. Other errors should be logged at the debug level.

connormullett:
**Summary of changes**
- Before returning an error to the caller of `apply_message`, give a warning/debug message to the user that a message has failed

Closes 
#1649 


vm/interpreter/Cargo.toml
```diff
@@ -43,6 +43,8 @@ cid = { version = "0.8", default-features = false, features = ["std"] }
 
 [dev-dependencies]
 interpreter = { path = "../interpreter/" }
+chain = { path = "../../blockchain/chain" }
+state_manager = { path = "../../blockchain/state_manager" }
```
lemmih, line 47:
```suggestion
```

lemmih (changes requested):

lemmih:
And you probably don't need to update the `Cargo.lock` file.

vm/interpreter/src/vm.rs
```diff
@@ -324,6 +324,29 @@ where
...
+
+        if !exit_code.is_success() {
+            match exit_code.value() {
+                1..=11 => {
+                    log::debug!(
+                        "Internal message execution failure. Exit code was {}",
+                        exit_code
+                    )
+                }
+                16..=24 => {
```
LesnyRumcajs, line 338: To my understanding of the code (I could be wrong), e.g. 12, 13, 26 or 27 are reserved codes, so the `ref-fvm` team may assign it even tomorrow. We would then need to know about it and bump the values.

I believe it'd be easier and more maintainable to have a simpler:
* if the exit code value is `1..FIRST_USER_EXIT_CODE` then it's an internal fiasco,
* else it's a fiasco in general

We wouldn't have to revisit it then in the future too often (unless a new category is created or something like that).

What do you think?

connormullett: I think this way was a lot simpler than matching over every possible exit code. You're right about the reserved codes but this just seemed more clean and practical given the FVM doesn't have a way to go from the exit code value to the name easily. I'll go ahead and change this up so we can provide more information to the user than just a number they'll have to look up

connormullett: Actually nevermind, changing up the match would be a lot simpler

vm/interpreter/Cargo.toml
connormullett: Good catch. Removed

LesnyRumcajs (approved):

lemmih (changes requested):
Could you revert the changes to `Cargo.lock`?

PR #1656, opened 2022-07-14
Refactor validation: introduce the Consensus trait

aakoshh:
**Summary of changes**
Changes introduced in this pull request:
- Added a `chain_sync::consensus::Consensus` trait which currently has a single `validate_block` method for consensus specific rules. It also has an associated error type.
- Added a `blockchain/fil_consensus` crate with `FilecoinConsensus` implementing the `Consensus` trait. Moved the errors and validations specific to PoSt, tickets, beacons, DRAND, etc. into this crate. `FilecoinConsensus` also captures the generic beacon type and the `ProofValidator`. 
- The structs in `tipset_syncer` and `ChainMuxer` get a generic `consensus: C where C: Consensus` argument so they can delegate the custom validations to it. 
- They keep the common validation which is related to the messages and the ledger (gas, account sequences, state, fees). 
- Because the `ProofValidator` was moved into `FilecoinConsensus` I was able to remove it from all but one method in the RPC and `StateManager` where a block is assembled. 

**Other information and links**
This is laying some groundworks towards implementing [Hierarchical Consensus](https://github.com/protocol/ConsensusLab/blob/hc/spec/specs/hierarchical_consensus.md)

In the future we want to expand this interface to be able to generate blocks on a schedule, adhering to some hitherto unknown consensus protocol.

PR #1680, opened 2022-07-15
Move chain::weight into Consensus

aakoshh:
**Summary of changes**
- Introduce a `Scale` trait inherited by `Consensus` to abstract away the `weight` function from the `chain` module.
- Moved the `weight` function into the `FilecoinConsensus` trait introduced in https://github.com/ChainSafe/forest/pull/1656
- Moved `blockchain/fil_consensus` to `blockchain/consensus/fil_cns`. I started working on [Delegated Consensus](https://github.com/aakoshh/forest/pull/2) and realised that adding more would pollute the `blockchain` directory. `eudico` has a [similar](https://github.com/filecoin-project/eudico/tree/eudico/chain/consensus) arrangement. Sorry that I haven't thought about this in the earlier PR.

The factoring out of the chain weight [also happens](https://github.com/filecoin-project/eudico/blob/4b8c8540b7e3a58f87950baad0f77f721233a723/cmd/eudico/filcns.go#L21) in `eudico` but I didn't notice it earlier because it's passed as a setting, rather than part of the [consensus interface](https://github.com/filecoin-project/eudico/blob/4b8c8540b7e3a58f87950baad0f77f721233a723/chain/consensus/iface.go). But it feels like it belong under the same umbrella.


lemmih (approved):

LesnyRumcajs (approved):
rock-solid, thank you!

PR #1697, opened 2022-07-20
Forest fails if user doesn't provide snapshot when the DB is empty

### LINKED ISSUE #1669: Early exit Forest if no snapshot is provided and the DB is empty
**Issue summary**

Right now we can run forest on a clean instance without providing a snapshot. We can, but we all have the tribal knowledge that this will fail miserably. Others do not (yet).

It would be great to manage the expectations of people using Forest/first-time contributors and fail fast with a helpful message, e.g. `Need to bootstrap Forest with a snapshot on the first run` plus perhaps a link to the usual source of mainnet snapshots. This will certainly improve the user/contributor experience.

https://github.com/ChainSafe/forest/pull/1656

connormullett:
**Summary of changes**
- Refer the user to the documentation if no snapshot is provided and HEAD is below the latest network upgrade

Closes 
#1669 


connormullett:
> Don't check whether the DB exists. Instead, check if the epoch is recent enough for Forest to sync correctly.

That's a great idea. Was using this as a placeholder till we talked about it. Left a comment on the issue about this earlier

lemmih (dismissed):
- Make it explicit that we support Skyr and newer (rather than just the latest version).
- The error messages you use with `expect` should be understandable on their own and not just a statement that you expect to be true.
- Loading a snapshot for a network version older than Skyr should also fail. It's fine to fail /after/ the snapshot has been loaded.

forest/src/daemon.rs
```diff
@@ -102,7 +103,9 @@ pub(super) async fn start(config: Config) {
     // Start Prometheus server port
     let prometheus_listener = TcpListener::bind(config.metrics_address)
         .await
-        .unwrap_or_else(|_| panic!("could not bind to {}", config.metrics_address));
+        .unwrap_or_else(|_| {
+            cli_error_and_die(&format!("could not bind to {}", config.metrics_address), 1)
```
LesnyRumcajs, line 107: minor suggestion, we could use `AsRef<str>` as an argument to this function to make it a bit more ergonomic to use (e.g. being able to pass just `String`)

forest/src/daemon.rs
```diff
@@ -154,6 +157,26 @@ pub(super) async fn start(config: Config) {
 
     sync_from_snapshot(&config, &state_manager).await;
 
+    // Terminate if no snapshot is provided and DB isn't recent enough
+    match chain_store.heaviest_tipset().await {
+        None => {
+            cli_error_and_die(
+                "Forest cannot sync without a snapshot. Download a snapshot from a trusted source and import with --import-snapshot=[file]",
```
LesnyRumcajs, line 164: This is great but may be a bit puzzling for new folks. They most likely don't know where is the trusted source. How about we add a section in README `Trusted snapshot sources` to make their life easier?

LesnyRumcajs (approved):

forest/src/daemon.rs
lemmih: Once we're hosting our own trusted snapshots, we can offer to automatically download them.

tyshko5 (approved):

PR #1699, opened 2022-07-20
Ensure blocks are created by a distinct set of miners. 

### LINKED ISSUE #1675: Ensure blocks are created by a distinct set of miners.
**Issue summary**

There's a check in `verify_blocks` to ensure all blocks are made by different miners, before they are assembled into a valid tipset. It compares each miner address to the first. This is incorrect and will miss the case of any subsequent miners being equal to each other, but not the first, e.g. [A,B,B].

Relevant code:
https://github.com/ChainSafe/forest/blob/46fe3ff57bf1ea634fe325dc9c17755d576fb25e/blockchain/blocks/src/tipset.rs#L227-L230

- [ ] Check what Lotus is doing. Do they have the same mistake? If so, make sure it's actually a mistake and not a misunderstanding of the filecoin specification.
- [ ] Update the check to ensure the miners are all distinct.

tyshko5:
**Summary of changes**
- 

Closes #1675

### DIFF
--- a/blockchain/blocks/src/tipset.rs
+++ b/blockchain/blocks/src/tipset.rs
@@ -531,7 +531,6 @@ mod test {
             Error::InvalidTipset("parent cids are not equal".to_string())
         );
     }
-    
 
     #[test]
     fn ensure_there_are_blocks() {
--- a/blockchain/blocks/src/tipset.rs
+++ b/blockchain/blocks/src/tipset.rs
@@ -235,8 +235,9 @@ where
     };
 
     let mut headers_set: HashSet<Address> = HashSet::new();
+    headers_set.insert(*first_header.miner_address());
 
-    let mut headers_len = 0;
+    let mut headers_len = 1;
 
     for header in headers {
         verify(
@@ -253,7 +254,10 @@ where
         )?;
 
         headers_len += 1;
-        headers_set.insert(*header.miner_address());
+        verify(
+            headers_set.insert(*header.miner_address()),
+            "miner_addresses are not distinct",
+        )?;
     }
 
     verify(
@@ -377,11 +381,14 @@ pub mod tipset_json {
 
 #[cfg(test)]
 mod test {
-    use crate::{BlockHeader, ElectionProof, Ticket, Tipset};
+    use crate::{BlockHeader, ElectionProof, Ticket, Tipset, Error, TipsetKeys};
     use cid::Cid;
     use forest_crypto::VRFProof;
+    use fvm_ipld_encoding::DAG_CBOR;
     use fvm_shared::address::Address;
     use num_bigint::BigInt;
+    use cid::multihash::Code::Identity;
+    use cid::multihash::MultihashDigest;
 
     pub fn mock_block(id: u64, weight: u64, ticket_sequence: u64) -> BlockHeader {
         let addr = Address::new_id(id);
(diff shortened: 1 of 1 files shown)


blockchain/blocks/src/tipset.rs
```diff
@@ -246,12 +251,16 @@ where
             header.epoch() == first_header.epoch(),
             "epochs are not equal",
         )?;
-        verify(
-            header.miner_address() != first_header.miner_address(),
-            "miner_addresses are not distinct",
-        )?;
+
+        headers_len += 1;
+        headers_set.insert(*header.miner_address());
```
LesnyRumcajs, line 256: we could fail early by checking if insertion succeeded, given it's a set, no?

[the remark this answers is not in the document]
lemmih: And remember to include the miner address of the `first_header`.

blockchain/blocks/src/tipset.rs
```diff
@@ -429,4 +445,68 @@ mod test {
...
+        let h0 = BlockHeader::builder()
+            .miner_address(Address::new_id(0))
+            .build()
+            .unwrap();
+        let h1 = BlockHeader::builder()
+            .miner_address(Address::new_id(0))
+            .build()
+            .unwrap();
+        assert!(Tipset::new(vec![h0, h1]).unwrap_err() == Error::InvalidTipset("miner_addresses are not distinct".to_string()));
+    }
```
lemmih, line 463: Good. Could you also add a test that would have caught the error in the previous code? Ie, headers that are not distinct but still different from the first header.

blockchain/blocks/src/tipset.rs
```diff
@@ -246,12 +252,19 @@ where
...
+            headers_set.insert(*header.miner_address()),
             "miner_addresses are not distinct",
         )?;
     }
 
+    verify(
+        headers_len == headers_set.len(),
+        "miner_addresses are not distinct",
+    )?;
+
```
lemmih, line 267: This is no longer necessary.

blockchain/blocks/src/tipset.rs
```diff
@@ -429,4 +445,82 @@ mod test {
...
+            .unwrap();
+        assert!(
+            Tipset::new(vec![h0, h1]).unwrap_err()
+                == Error::InvalidTipset("parent cids are not equal".to_string())
+        );
+    }
+
+    #[test]
+    fn ensure_there_are_blocks() {
+        assert!(Tipset::new(vec![]).unwrap_err() == Error::NoBlocks);
```
lemmih, line 524:
```suggestion
        assert_eq!(Tipset::new(vec![]).unwrap_err(), Error::NoBlocks);
```

lemmih (approved):
I'm happy. It's tempting to use a peekable iterator but that's not necessary.

LesnyRumcajs:
Does Lotus have it right and it was an issue only in Forest? I'm asking, because if it's not the case, probably we should create an issue in Lotus.

lemmih:
> Does Lotus have it right and it was an issue only in Forest? I'm asking, because if it's not the case, probably we should create an issue in Lotus.

I can't seem to find any similar code in Lotus. They don't do any checks when creating a `FullTipSet`. Maybe the checks happen at some other location but grepping gives me nothing.

tyshko5:
> > Does Lotus have it right and it was an issue only in Forest? I'm asking, because if it's not the case, probably we should create an issue in Lotus.
> 
> I can't seem to find any similar code in Lotus. They don't do any checks when creating a `FullTipSet`. Maybe the checks happen at some other location but grepping gives me nothing.

They do it https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/chain/store/store.go#L1012-L1015

lemmih:
> > > Does Lotus have it right and it was an issue only in Forest? I'm asking, because if it's not the case, probably we should create an issue in Lotus.
> > 
> > 
> > I can't seem to find any similar code in Lotus. They don't do any checks when creating a `FullTipSet`. Maybe the checks happen at some other location but grepping gives me nothing.
> 
> They do it https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/chain/store/store.go#L1012-L1015

Ah, great. Then we don't need to create an issue.

LesnyRumcajs:
> > > Does Lotus have it right and it was an issue only in Forest? I'm asking, because if it's not the case, probably we should create an issue in Lotus.
> > 
> > 
> > I can't seem to find any similar code in Lotus. They don't do any checks when creating a `FullTipSet`. Maybe the checks happen at some other location but grepping gives me nothing.
> 
> They do it https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/chain/store/store.go#L1012-L1015

Looks like a bit different behaviour than what we do, no? I'm just double-checking if we are doing the right thing. Here we do the verification later and we fail everything if there's a duplicate. There, from a quick glimpse, they're just grabbing the first one and issuing a warning. If you guys think this is correct then I believe you, but looks suspicious to me. :)

LesnyRumcajs:
> They are issuing a warning each time. What you are saying is that we need to create an issue on Lotus to replace warning with error?

It's rather us that should not fail miserably. Unless it's already the case and this verification fiasco is handled gently.

LesnyRumcajs (approved):

PR #1707, opened 2022-07-21
Don't rename crates

### LINKED ISSUE #1706: Remove package renaming
**Issue summary**

Several crates are still being renamed in some locations. Renaming crates generally makes the code more difficult to read and should be avoided.

Running `git grep "package = "` will indicate where crates are being renamed. The following crates should never be renamed:
- `forest_encoding`
- `forest_db`
- `actor_interface`
- `forest_crypto`
- `forest_blocks`
- `cs_serde_bytes`

connormullett:
**Summary of changes**
- Fix crates that have been renamed (i.e. crypto -> forest_crypto)

Closes #1706 

- To verify this PR fixes all renaming issues, use `git grep "package = "


lemmih:
Make sure that your PRs have a line saying "Closes #issue_number". If you put the issue on a separate line, GitHub doesn't understand that they are connected.

tyshko5 (approved):

LesnyRumcajs:
In general, do we know the reasoning behind renaming those crates in the first place?

encoding/Cargo.toml
```diff
@@ -13,7 +13,7 @@ fvm_shared         = { version = "0.8.0", default-features = false }
 fvm_ipld_encoding  = "0.2.1"
 blake2b_simd       = "1.0"
 serde              = { version = "1.0", features = ["derive"] }
-serde_bytes        = { package = "cs_serde_bytes", version = "0.12" }
+cs_serde_bytes     = "0.12"
```
LesnyRumcajs, line 16: In general, why do we need this custom `serde_bytes` crate? What are the differences? Shall we make a task to take a look a around? I looked at [crates.io](https://crates.io/crates/cs_serde_bytes) and seems it doesn't even point to a correct repository.

LesnyRumcajs (approved):
Nice.

encoding/Cargo.toml
lemmih: That is a very good question. It seems that we're maintaining a fork of `serde_bytes` here: https://github.com/ChainSafe/serde-bytes. This crate is also used by the FVM. We should create an issue to identify why this crate exists and if it can be removed (from both Forest and ref-fvm).

lemmih: Issue created: #1722

lemmih:
> In general, do we know the reasoning behind renaming those crates in the first place?

I assume the initial authors felt it was prettier.

PR #1714, opened 2022-07-22
Delegated Consensus

aakoshh:
**Summary of changes**
_The PR is not as big as it looks, half of it is a README about trying to make a usable Genesis file._

- Created a `deleg_cns` crate for _Delegated Consensus_ mimicking [delegcns](https://github.com/filecoin-project/eudico/blob/eudico/chain/consensus/delegcns) in Eudico
- Added a `fil_cns` and a `deleg_cns` Cargo feature flag to turn it on
- Added `MessagePool::select_messages_for_block` to pick the messages which should go in the next block. Calling it from `republish_pending_messages` as well so it gets tested.
- Moved some parts of `state_api::miner_create_block` to `chain::persist_block_messages`, namely the storing of messages as CID and the aggregation of BLS signatures. It seemed to fit in with the other helper methods there that _read_ the data stored this way, now it's also able to write it. The crate already depended on `forest_crypto`, and so implicitly the BLST library as well, so I thought adding this as an explicit dependency to do the aggregation is not that much of a leap. 
- Added a huge README about my attempt to produce a Genesis file that will work with _Delegated Consensus_. It's a simple thing: it just needs a key in it, and the corresponding private key in the wallet of one of the nodes.  

This PR is part of the effort to add [Hierarchical Consensus](https://github.com/protocol/ConsensusLab/blob/hc/spec/specs/hierarchical_consensus.md) to Forest. It build on the previous PR https://github.com/ChainSafe/forest/pull/1680 to add a new consensus type, but doesn't actually add any hierarchical machinery, its goal is limited to have a single miner proposing blocks at regular intervals, as an alternative to the default Filecoin Consensus. Because everyone knows who is supposed to create a blocks, it doesn't involve proofs. It is just something @adlrocha often uses during demos because of its simplicity. 

Unfortunately I am unable to run it at the moment, because to do so would require a Genesis file which we are unable to produce:
* Forest doesn't have the methods yet to put together the initial state tree, or to write it into a CAR file
* Lotus can do it, but the version it uses at Genesis is not what Forest expects (it wants v8)

So, for now this isn't something we can use. We have to consult others to see when the CAR writing capability is forthcoming, and also add methods to Forest to produce a Genesis file in the version it expects. It would be great to allow custom consensus to have a say in the construction of the Genesis state - like in Eudico - but as long as we adhere to the general validation in Forest, which expects for example Miner state, we can't deviate too much. 

In any case, we would much appreciate your input into whether an integration like this (e.g. the feature flags) are acceptable to you, or you have a different idea. It would be awesome if it could be merged into Forest so the code doesn't rot away as the dependencies change, even though it's not in use. 

* Original PR with some discussion: https://github.com/aakoshh/forest/pull/2
* The epic this story is part of: https://github.com/protocol/ConsensusLab/issues/127

The README about hints at some issues I discovered during my exploration of the CLI:
* The CLI seems to ignore the `--token` option and only looks for the JWT in an env var
* The keys imported by the CLI always have mainnet prefix (`f`); I had to manually edit `keystore.json` to change them to `t`.


blockchain/consensus/deleg_cns/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+[dependencies]
+anyhow          = "1.0"
+async-std = { version = "1.9", features = ["tokio1", "unstable"] }
+async-trait = "0.1"
+futures = "0.3.5"
+log = "0.4.8"
+nonempty = "0.8.0"
+thiserror = "1.0"
+
+cid = { package = "forest_cid", version = "0.3" }
```
lemmih, line 16: `forest_cid` has been deleted. You can just use `cid` instead.

blockchain/consensus/deleg_cns/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+encoding = { package = "forest_encoding", version = "0.2.1" }
```
lemmih, line 17:
```suggestion
forest_encoding = "0.2.1"
```

blockchain/consensus/deleg_cns/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+fvm_shared = { version = "0.8.0", default-features = false }
+ipld_blockstore = "0.1"
+fil_types = "0.2"
+forest_bigint = "0.1"
+
+actor = { package = "actor_interface", path = "../../../vm/actor_interface" }
```
lemmih, line 23:
```suggestion
actor_interface = { path = "../../../vm/actor_interface" }
```

blockchain/consensus/deleg_cns/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+blocks = { package = "forest_blocks", path = "../../blocks" }
```
lemmih, line 24:
```suggestion
forest_blocks = { path = "../../blocks" }
```

forest/src/daemon.rs
```diff
@@ -204,12 +209,49 @@ pub(super) async fn start(config: Config) {
...
+    let submitter = SyncGossipSubmitter::new(
+        network_name.clone(),
+        network_send.clone(),
+        tipset_sink.clone(),
+    );
+
     // Initialize Consensus
-    type FullConsensus = FilecoinConsensus<DrandBeacon, FullVerifier>;
-    let consensus: FullConsensus = FilecoinConsensus::new(state_manager.beacon_schedule());
+    #[cfg(all(feature = "fil_cns", not(any(feature = "deleg_cns"))))]
```
lemmih, line 223: If neither `fil_cns` or `deleg_cns` is defined, could we print an error message?

node/rpc/src/state_api.rs
```diff
@@ -693,7 +644,7 @@ pub(crate) async fn miner_create_block<
         .signature(None)
         .build()?;
 
-    let key = key_management::find_key(&worker, &*data.keystore.as_ref().write().await)?;
+    let key = key_management::find_key(&worker, &*data.keystore.as_ref().read().await)?;
```
lemmih, line 647: Oops, good find.

lemmih (changes requested):
Looks good. Can't say I understand the consensus algorithms, though.

A few changes required because we re-arranged some of our dependencies recently.

tyshko5 (approved):

forest/src/daemon.rs
aakoshh: Good idea!

blockchain/consensus/deleg_cns/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+log = "0.4.8"
+nonempty = "0.8.0"
+thiserror = "1.0"
+
+cid = { version = "0.8", default-features = false, features = ["std"] }
+encoding = { package = "forest_encoding", version = "0.2.1" }
+fvm_shared = { version = "0.8.0", default-features = false }
+ipld_blockstore = "0.1"
+fil_types = "0.2"
+forest_bigint = "0.1"
```
lemmih, line 21: This crate has also been removed.

lemmih (changes requested):

blockchain/consensus/deleg_cns/Cargo.toml
lemmih: The `forest_bigint` crate, that is.

aakoshh: Thanks! I saw the change when I merged but didn't check this part.

forest/src/daemon.rs
```diff
@@ -204,12 +209,51 @@ pub(super) async fn start(config: Config) {
...
+        network_name.clone(),
+        network_send.clone(),
+        tipset_sink.clone(),
+    );
+
     // Initialize Consensus
-    type FullConsensus = FilecoinConsensus<DrandBeacon, FullVerifier>;
-    let consensus: FullConsensus = FilecoinConsensus::new(state_manager.beacon_schedule());
+    #[cfg(not(any(feature = "fil_cns", feature = "deleg_cns")))]
+    error!("No consensus feature has been enabled; use e.g. `--feature fil_cns` to pick one.");
```
lemmih, line 224:
```suggestion
    compile_error!("No consensus feature has been enabled; use e.g. `--feature fil_cns` to pick one.");
```

lemmih (changes requested):

aakoshh:
> Can't say I understand the consensus algorithms, though.

@adlrocha didn't say much more than this consensus being one where there's only one miner allowed to produce blocks, so the name "delegated consensus" has nothing to with delegating stake like in Cosmos, it's just a fancy term to say "we all delegate the right to create blocks to this single miner, forever". The validations are just checking that the block has been signed by this miner and that the timestamp is aligned with the epochs.

forest/src/daemon.rs
aakoshh: Sweet!

lemmih (approved):
Looking good.

adlrocha:
> @adlrocha didn't say much more than this consensus being one where there's only one miner allowed to produce blocks, so the name "delegated consensus" has nothing to do with delegating stake like in Cosmos, it's just a fancy term to say "we all delegate the right to create blocks to this single miner, forever". The validations are just checking that the block has been signed by this miner and that the timestamp is aligned with the epochs.

I agree, it is a confusing name. We called it `delegated` for historical reasons. We can probably call it `debug` or `dummy` consensus in the next iteration. Thank you for the quick merge :pray:

PR #1718, opened 2022-07-25
Re-enable more forest_ipld json tests

### LINKED ISSUE #1713: Restore the IPLD walk tests
**Issue summary**

The `forest_ipld` crate has a set of test vectors for querying IPLD documents but the tests are currently disabled. The `submodules_tests` feature flag isn't appropriate for the IPLD tests and should be removed from `ipld/tests/walk_tests.rs`. This module has not been compiled in quite a while so some elbow grease may be needed.

lemmih:
**Summary of changes**
- Re-enable walk tests.
- Re-enable selector explore tests.
- Get rid of `json` feature flag. We always want JSON support.

Closes #1713 


tyshko5 (approved):

ipld/tests/ipld-traversal-vectors/selector_explore.json
```diff
@@ -11,7 +11,7 @@
         "explore": [
             {
                 "ipld": null,
-                "path_segment": 0
+                "path_segment": "0"
```
LesnyRumcajs, line 14: Seems like we had a breaking and suboptimal change (representing numbers in with strings). Are you sure we want to carry on with this?

lemmih: We previously allowed both strings and integers and then converted them on the fly (both from string to integer and integer to string). Disallowing integer values doesn't make it any less or more typed than before, imo.

LesnyRumcajs: Why do we want to disallow them? Wouldn't it be less flexible? I could imagine a client that "optimizes" its calls a bit that would then need to change.

lemmih: They're disallowed in `libipld`. We have to roll our own code to allow integers as path segments. Given that it doesn't actually buy us any additional type safety (I'm ready to be shown wrong on this, btw), I don't think having a custom implementation of path segments is worth it.

LesnyRumcajs: Ah, ok. If the upstream disallows them then all good. Perhaps someone will issue a special feature request in the future if this flexibility is needed.

LesnyRumcajs (approved):

PR #1725, opened 2022-07-26
Enable cargo-spellcheck for blockchain/chain_sync

### LINKED ISSUE #1687: Enable `cargo-spellcheck` for `blockchain/chain_sync`
**Issue summary**

Make sure that `cargo spellcheck` can run without errors in `blockchain/chain_sync`. For details, see the tracking issue.

tyshko5:
**Summary of changes**
-  Added async, incentivize, P2P, synchronizer, nonces to the dictionary
- Fixed the crate, so it could pass the spellcheck
- I've added async instead of fuller versions, as it became a more popular word. 

Closes  #1687


lemmih:
Always write a summary of the changes in the PR. The summary is also a good place to describe the design decisions you've made. For example, you could write about why you added `async` to the dictionary instead of using `asynchronous` or `asynchronously`.

.config/forest.dic
```diff
@@ -17,3 +17,8 @@ tuple
 mutex
 RPC
 parameterize/D
+P2P
+synchronizer
+nonces
```
lemmih, line 22:
```suggestion
nonce/S
```

tyshko5: this doesn't work for some reason

tyshko5:
Added

.config/forest.dic
```diff
@@ -17,3 +17,8 @@ tuple
...
+async
+incentivize
```
LesnyRumcajs, line 24: I'm kind of puzzled why do we need to add this. It's a normal word, no? Perhaps we can use a better "general" dictionary? Same goes for `nonce`.

LesnyRumcajs (approved):

.config/forest.dic
lemmih: Yeah, the builtin dictionary only has 48k words. My up-to-date local dictionary has 79k words (including `nonce` and `incentivize`). The dictionary is 841KiBi. Probably worth it to include it in the repository.

LesnyRumcajs: Let's do it! :muscle:

lemmih (approved):

LesnyRumcajs:
@tyshko5 mergio?

tyshko5:
> @tyshko5 mergio?

I guess. But I'm not sure if I can merge anything

PR #1735, opened 2022-07-27
IaC: Service for creating and uploading calibnet snapshots

### LINKED ISSUE #1694: Periodical calibnet snapshot upload
**Issue summary**

The goal is to get a stream of up-to-date calibnet snapshots. Steps:

This should be done on a continuous sync testing machine.

- [ ] Once per day, upload a snapshot (a car file) to DO "Spaces".
  - [ ] Tag the snapshot with date and epoch.
  - [ ] Keep only `N` snapshots in the DO space (how many? 1? 5? 10? 1 per week? 1 per month?).

~~- [ ] Once the setup works, duplicate for `mainnet`.~~ This may be too heavy on the Spaces, if we want to have archival mainnet snapshots we can use Volumes. To re-consider.

https://www.digitalocean.com/blog/introducing-spaces-object-storage

lemmih:
**Summary of changes**
- Add script for initializing a service that automatically uploads calibnet snapshots to Digital Ocean and reports failures to slack.
- Decided not to use terraform due to the complexity.
- The infrastructure consists of a single boot script that is run on a server with docker installed. The bootscript takes care of all other dependencies and automatically updates itself via watchtower.

Closes #1694 


scripts/snapshot_check/docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+    environment:
+      - SLACK_HOOK
+    command:
+      - |
+        apt-get install -y curl
+
+        report() {
+          curl -X POST -H 'Content-type: application/json' --data "{\"text\":\"❌  $FOREST_HOSTNAME snapshot export failure!\"}" "$$SLACK_HOOK"
+        }
+        trap 'report' ERR
```
LesnyRumcajs, line 24: to my understanding, this will indeed call `report()` in case of an error but then it will happily continue execution anyway, no? Shouldn't it be rather `report_and_die` with an exit?

scripts/snapshot_check/run_service.sh
```diff
@@ -0,0 +1,47 @@
...
+
+set -e
+
+S3_FOLDER=$BASE_FOLDER/s3
+
+echo "service script"
+
+error=0
+
+check_env () {
```
LesnyRumcajs, line 11: This function definitely deserves an extensive documentation on what it does. :)

scripts/snapshot_check/run_service.sh
```diff
@@ -0,0 +1,47 @@
...
+function cleanup {
+  echo "unmounting s3 folder"
+  fusermount -u -q "$S3_FOLDER"
+}
+trap cleanup EXIT
+s3fs forest-snapshots "$S3_FOLDER" \
+    -o url=https://fra1.digitaloceanspaces.com/ \
+    -o allow_other
+
+docker-compose up
```
LesnyRumcajs, line 47: I kind of find it more robust to `docker-compose up --build --force-recreate`, but it may be fine here

scripts/snapshot_check/run_service.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -e
+
+S3_FOLDER=$BASE_FOLDER/s3
+
+echo "service script"
```
LesnyRumcajs, line 7: probably a more meaningful message would be useful

LesnyRumcajs (changes requested):
I believe we need a strategy to not store all calibnet snapshots. In a year or so the total amount of those calibnet snapshots (given we'll be producing two a day) will amass to around 3TB. Are we fine with paying that (given current price it shouldn't be a lot, but it's still wasty)? Or should we rather store those with different granularity, e.g. once a month pick only one snapshot from each week, and obliterate the rest?

scripts/snapshot_check/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:22.10
```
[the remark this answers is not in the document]
lemmih: I tried with `debian:buster-slim` but the version of `s3fs` was too old. :(

scripts/snapshot_check/docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+        target: ${BASE_FOLDER}
+    entrypoint: ["/bin/bash", "-c"]
+    environment:
+      - SLACK_HOOK
+    command:
+      - |
+        apt-get install -y curl
+
+        report() {
+          curl -X POST -H 'Content-type: application/json' --data "{\"text\":\"❌  $FOREST_HOSTNAME snapshot export failure!\"}" "$$SLACK_HOOK"
```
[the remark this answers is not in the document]
lemmih: Making file available through docker-in-docker is a pain in the butt, though. I wonder if there's a simple way for container to use files from another container without sharing a volume.

scripts/snapshot_check/docker-compose.yml
lemmih: I thought it would die after reporting. My bash-fu is weak so it can't hurt to make sure.

LesnyRumcajs:
```bash
#!/bin/bash

report() {
  echo 'fiasco!'
}
trap 'report' ERR

if [ $# -eq 0 ]; then
  echo "begin"
  rmdir see-no-evil
  echo "carry on my wayward soooon"
fi
```
```
❯ bash tst.sh
begin
rmdir: failed to remove 'see-no-evil': No such file or directory
fiasco!
carry on my wayward soooon
```

lemmih: Ah, right. I'm missing the `set -e`. With that at the top, your example exits after the first error.

scripts/snapshot_check/run_service.sh
```diff
@@ -0,0 +1,46 @@
...
+# 2. Run forest script with docker-compose.
+
+## Setup s3
+mkdir --parents "$S3_FOLDER"
+function cleanup {
+  echo "unmounting s3 folder"
+  fusermount -u -q "$S3_FOLDER"
+}
+trap cleanup EXIT
+s3fs forest-snapshots "$S3_FOLDER" \
```
LesnyRumcajs, line 42: these are a bit specific, would be nice to put them as an env variable

LesnyRumcajs (approved):
@lemmih sorry for a picky review, having worked a bit with shell scripts I'm just concerned about the maintainability of them. Feel free to address the comments or not, if you believe it's not worth the hassle and I'm just pedantic or grumpy today. :)

scripts/snapshot_check/docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+        }
+        trap 'report' ERR
+
+        cd $BASE_FOLDER
+        RECENT_SNAPSHOT=s3/calibnet/`ls -Atr1 s3/calibnet/ | tail -n 1`
+        forest --encrypt-keystore false --metrics-address 0.0.0.0:6116 --chain calibnet --import-snapshot $${RECENT_SNAPSHOT} &
+        
+        sleep 1200 # Wait for the RPC endpoint to be available.
```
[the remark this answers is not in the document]
lemmih: Once the `--detach` flag is available, the sleep can be removed and no periodic checks are necessary.

tyshko5 (approved):

LesnyRumcajs:
Random idea, it'd be useful to store the checksum as well (like for mainnet snapshots on AWS). Random file corruption occurred when I tried to upload large files into DO Spaces, so it'd be nice to have some checks.

scripts/ruby_common/docker_utils.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
LesnyRumcajs, line 1: this seems to be a copy from https://github.com/ChainSafe/forest/tree/main/scripts/sync_check/src, should we re-use it?

scripts/ruby_common/slack_client.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
```
LesnyRumcajs, line 1: this seems to be a copy from https://github.com/ChainSafe/forest/tree/main/scripts/sync_check/src, should we re-use it?

scripts/ruby_common/daily_snapshot.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require_relative 'slack_client'
+require_relative 'docker_utils'
+require 'rubygems'
+require 'logger'
+require 'fileutils'
+require 'active_support/time'
```
LesnyRumcajs, line 8: It's quite a big one, coming with Ruby on Rails. What's stopping us from using standard time?

scripts/ruby_common/daily_snapshot.rb
```diff
@@ -0,0 +1,64 @@
...
+
+# Create log directory
+FileUtils.mkdir_p LOG_DIR
+
+loop do
+  client = SlackClient.new CHANNEL, SLACK_TOKEN
+
+  # Find the snapshot with the most recent modification date
+  LATEST = Dir.glob("#{BASE_FOLDER}/s3/#{CHAIN_NAME}/*").max_by {|f| File.mtime(f)}
+
```
LesnyRumcajs, line 38:
```suggestion
```

scripts/ruby_common/daily_snapshot.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  # Check if the date of the most recent snapshot is today
+  if Time.new.strftime('%F') == File.stat(LATEST).mtime.strftime('%F')
```
LesnyRumcajs, line 41: converting to `Date` and comparing would be a bit more readable than `%F`s. You can use `.to_date` method of `Time`.

scripts/ruby_common/daily_snapshot.rb
```diff
@@ -0,0 +1,64 @@
...
+
+CHAIN_NAME = ARGV[0]
+raise 'No chain name supplied. Please provide chain identifier, e.g. calibnet or mainnet' if ARGV.empty?
+
+# Current datetime, to append to the log files
+DATE = Time.new.strftime '%FT%H:%M:%S'
+LOG_EXPORT = "#{LOG_DIR}/#{CHAIN_NAME}_#{DATE}_export"
+
+# Create log directory
+FileUtils.mkdir_p LOG_DIR
```
LesnyRumcajs, line 31: not sure if it's needed if the LOG_DIR is CWD

scripts/ruby_common/daily_snapshot.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require_relative 'slack_client'
+require_relative 'docker_utils'
+require 'rubygems'
+require 'logger'
+require 'fileutils'
+require 'active_support/time'
+
+# Retrieves an environmental variable, failing if its not set or empty.
```
LesnyRumcajs, line 10: This method could probably be put in a separate file for re-use

scripts/ruby_common/daily_snapshot.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require_relative 'slack_client'
+require_relative 'docker_utils'
+require 'rubygems'
```
LesnyRumcajs, line 5: why do we need this?

scripts/daily_snapshot/upload_snapshot.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# If Forest hasn't synced to the network after 30 minutes, something has gone wrong.
+SYNC_TIMEOUT=30m
+
+CHAIN_NAME=$1
+NEWEST_SNAPSHOT=$2
```
LesnyRumcajs, line 7: shall we have a guard clause, failing early if arguments are not provided?

LesnyRumcajs (changes requested):

scripts/ruby_common/docker_utils.rb
lemmih: Yes, I'd like the two services to share the ruby files. Not sure how to do it in this PR, though. Was gonna update `sync_check` in a different PR. Is there a sensible way to re-organize the code in this PR?

scripts/ruby_common/daily_snapshot.rb
lemmih: Apparently it is required by `sleep(1.hour)`. Is there a way to sleep for an hour without using that gem?

scripts/ruby_common/docker_utils.rb
LesnyRumcajs: I believe putting the Dockerfiles in parent, i.e. `scripts/` or making them use it as the context would allow playing with `require_relative` without many issues.

lemmih: That would mean changing (or refactoring) the files in `sync_check`, right? I wary about doing that in this PR as I'm not confident that I won't break anything.

LesnyRumcajs: Yes. Otherwise you could depend on files in `sync_check`, but that wouldn't be much better. Feel free to change it inside `sync_check, then I can give you a hand at validating whether it doesn't break anything. Are you fine with that?

lemmih: I'm using `../ruby_common` as a bind in docker-compose. Not sure if that is allowed.

LesnyRumcajs: I'll check it out.

LesnyRumcajs: @lemmih In sync check compose, I added the common Ruby files to Ruby PATH, it should be fine now. I also moved the `sync_check.rb` back to `sync_check/`. Is it fine for you?

LesnyRumcajs (approved):

scripts/ruby_common/docker_utils.rb
lemmih: Absolutely. It didn't belong in `ruby_common` but I didn't know how else to get the imports to work.

PR #1741, opened 2022-07-28
Allow RPC bind address to be reconfigured

lemmih:
**Summary of changes**
- Allow the RPC address to be changed to `0.0.0.0`.
- It can be useful when accessing forest from a different container/machine.
- Security is still handled via access tokens.

Closes 

### DIFF
--- a/forest/src/cli/auth_cmd.rs
+++ b/forest/src/cli/auth_cmd.rs
@@ -57,7 +57,7 @@ impl AuthCommands {
                     Ok(token) => {
                         let mut addr = Multiaddr::empty();
                         addr.push(Protocol::Ip4(DEFAULT_HOST.parse().unwrap()));
-                        addr.push(Protocol::Tcp(cfg.rpc_port));
+                        addr.push(Protocol::Tcp(cfg.rpc_address.port()));
                         addr.push(Protocol::Http);
                         println!(
                             "FULLNODE_API_INFO=\"{}:{}\"",
--- a/forest/src/cli/auth_cmd.rs
+++ b/forest/src/cli/auth_cmd.rs
@@ -4,7 +4,7 @@
 use super::{handle_rpc_err, print_rpc_res_bytes, Config};
 use forest_libp2p::{Multiaddr, Protocol};
 use jsonrpc_v2::Error as JsonRpcError;
-use rpc_client::{auth_new, DEFAULT_HOST};
+use rpc_client::auth_new;
 use structopt::StructOpt;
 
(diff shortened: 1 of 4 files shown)


tyshko5 (approved):

forest/src/cli/config.rs
```diff
@@ -31,6 +31,8 @@ pub struct Config {
     pub encrypt_keystore: bool,
     /// Metrics bind, e.g. 127.0.0.1:6116
     pub metrics_address: SocketAddr,
+    /// RPC bind, e.g. 127.0.0.1
```
LesnyRumcajs, line 34: How about we use the same approach as in `metrics_address` and use `SocketAddr`? This way, on top of some uniformity, we'll have both the address and the port in one param.

lemmih: Good idea.

LesnyRumcajs: Thanks. Yeah, it'd be great to specify the port purpose in the name.

forest/src/cli/config.rs
```diff
@@ -31,12 +29,17 @@ pub struct Config {
...
+    /// RPC bind, e.g. 127.0.0.1:1234
+    pub rpc_address: SocketAddr,
     pub rocks_db: forest_db::rocks_config::RocksDbConfig,
     pub network: Libp2pConfig,
     pub sync: SyncConfig,
     pub chain: Arc<ChainConfig>,
 }
 
+// 127.0.0.1
+const LOCALHOST: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
```
LesnyRumcajs, line 41: Do we need it? There's a constant `std::net::Ipv4Addr::LOCALHOST` already there for grabbing.

LesnyRumcajs (approved):

forest/src/cli/config.rs
lemmih: I learn something new every day. :)

forest/src/cli/auth_cmd.rs
```diff
@@ -57,7 +57,7 @@ impl AuthCommands {
                     Ok(token) => {
                         let mut addr = Multiaddr::empty();
                         addr.push(Protocol::Ip4(DEFAULT_HOST.parse().unwrap()));
```
LesnyRumcajs, line 59: in general, shouldn't this be overridden as well?

lemmih: It should. Fixed.

PR #1743, opened 2022-07-28
Order-independent configuration for TOML serde 

### LINKED ISSUE #1488: Order-independent configuration for TOML serde
**Issue summary**

This structure:
https://github.com/ChainSafe/forest/blob/e89d8839ad4df527f458e2c3fcfc88264315dae0/forest/src/cli/config.rs#L11-L30

currently requires proper ordering of its fields, i.e. simple fields must come before complex ones. That is a bit unexpecting from a dev point of view but it's a limitation of the underlying technology. A way out of this is to put all of the *free* fields into a dedicated struct named `miscellaneous` which would put them in a separate section, thus resolving the explicit ordering of fields. Alternatively, we could put some of those fields in more specialized categories.

**Task summary**

- [ ] All fields in the `Config` struct should be structs (maybe enums are also acceptable, to be investigated)

**Acceptance Criteria**

`config dump` subcommand should have all the configuration in sections, e.g.
```
[miscellaneous] 
data_dir = "/home/rumcajs-work/.forest"
enable_rpc = true
rpc_port = "1234"
rpc_token = "adlks;ajdsakljdas"
snapshot = false
skip_load = false
encrypt_keystore = true
metrics_port = 6116

[rocks_db]
create_if_missing = true
parallelism = 16
write_buffer_size = 268435456
max_open_files = 1024
compression_type = "lz4"
enable_statistics = false
```

tyshko5:
**Summary of changes**
- 

Closes #1488


elmattic:
The key `miscellaneous` could be better I think, what do you think of `client`? (it's also used in lotus for general settings).

lemmih:
> The key `miscellaneous` could be better I think, what do you think of `client`? (it's also used in lotus for general settings).

This seems like a good idea. @LesnyRumcajs What do you think?

LesnyRumcajs:
I vote yes for Lotus naming compatibility. :lotus_position:

LesnyRumcajs (commented):
To `seal` this nice change, it'd be great to add a test that checks if the default config TOML doesn't have any free params. This way, anyone that tries to add a random int to the configuration will get immediately bashed when running tests. What do you think?

lemmih:
> To `seal` this nice change, it'd be great to add a test that checks if the default config TOML doesn't have any free params. This way, anyone that tries to add a random int to the configuration will get immediately bashed when running tests. What do you think?

How do we test for this?

LesnyRumcajs:
> > To `seal` this nice change, it'd be great to add a test that checks if the default config TOML doesn't have any free params. This way, anyone that tries to add a random int to the configuration will get immediately bashed when running tests. What do you think?
> 
> How do we test for this?

We check if the generated TOML starts with a section, i.e. `[`. The `free` params always land at the beginning of the file.

forest/src/cli/config.rs
```diff
@@ -2,60 +2,19 @@
...
-    pub snapshot: bool,
-    pub snapshot_height: Option<i64>,
-    pub snapshot_path: Option<String>,
-    /// Skips loading import CAR file and assumes it's already been loaded.
-    /// Will use the cids in the header of the file to index the chain.
-    pub skip_load: bool,
-    pub encrypt_keystore: bool,
-    /// Metrics bind, e.g. 127.0.0.1:6116
-    pub metrics_address: SocketAddr,
+    pub miscellaneous: Miscellaneous,
```
lemmih, line 15: As per @elmattic suggestion:

```suggestion
    pub client: Client,
```

lemmih (changes requested):
Good work. I'll add a new issue for testing that we don't have any free-standing configuration items.

LesnyRumcajs (approved):

elmattic (approved):

elmattic:
> Good work. I'll add a new issue for testing that we don't have any free-standing configuration items.

This last task should be done before merging the PR.

PR #1756, opened 2022-07-28
Cleanup build version module

### LINKED ISSUE #1647: Investigate the `types::build_version` module
**Issue summary**

The `types::build_version` module contains a bunch of unusual code, like this:
https://github.com/ChainSafe/forest/blob/b9a5ff48ad8e708c53f5312d46736babea7d1c1b/types/src/build_version/mod.rs#L11

To be investigated:
- [x] What do the version numbers mean? Why are they different from `forest --version`? Are they required to mimic Lotus' api?
- [x] What is `BuildType`/`NodeType`? Was this part of Lotus? Can we remove it?
- [x] Is there a reason to keep the module? If so, add the reason to the comments. If not, remove the module.

elmattic:
**Summary of changes**
- [x] Remove forest/build.rs 
- [x] Remove hardcoded user_version and use forest version
- [x] Remove api version numbers and use forest version 
- [x] Remove BuiltType
- [x] Remove NodeType

Closes #1647


types/src/build_version/mod.rs
```diff
@@ -120,3 +107,8 @@ impl std::convert::TryFrom<&NodeType> for Version {
         }
     }
 }
+
+/// Returns the version string, e.g., `0.2.2-unstable+git.21146f40`
```
LesnyRumcajs, line 111: I'd drop the git. Another thing, this `unstable` doesn't sound good. What do you think about having `rc` for the release track and `dev` for non-releases?

LesnyRumcajs (commented):
I'm kind of lost in this PR, could you please summarize this versioning strategy?

types/src/build_version/mod.rs
elmattic: The `git.` is here because Lotus used it too:
```
lotus$ ./lotus -v          
lotus version 1.17.1-dev+interopnet+git.a843c52e3
```
Of course we don't really need to follow them blindly.
This `unstable` doesn't sound good, yes. We could remove this part entirely. Or I would prefer that we add `+debug` for a debug build. This can be useful to users or us.

For the release naming convention (unstable/alpha/beta/rc etc...) this should be more something we handle with git tags.

lemmih: Will you change `unstable` to `+debug` in this PR?

types/src/build_version/mod.rs
```diff
@@ -17,13 +21,13 @@ const MINOR_ONLY_MASK: u32 = 0x00ff00;
 const PATCH_ONLY_MASK: u32 = 0x0000ff;
 
 // api versions
-const FULL_API_VERSION: Version = new_version(1, 1, 0);
-const MINER_API_VERSION: Version = new_version(0, 15, 0);
-const WORKER_API_VERSION: Version = new_version(0, 15, 0);
+const FULL_API_VERSION: Version = Version::new(1, 1, 0);
+const MINER_API_VERSION: Version = Version::new(0, 15, 0);
+const WORKER_API_VERSION: Version = Version::new(0, 15, 0);
```
lemmih, line 26: We should get rid of these, right?

types/src/build_version/mod.rs
```diff
@@ -17,13 +21,13 @@ const MINOR_ONLY_MASK: u32 = 0x00ff00;
...
+
+const GIT_HASH: &str = git_version!(args = ["--always", "--exclude", "*"], fallback = "unknown");
 
 lazy_static! {
-    pub static ref CURRENT_COMMIT: String = git_version!(fallback = "unknown").to_string();
-    pub static ref BUILD_TYPE: RwLock<BuildType> = RwLock::new(BuildType::BuildDefault);
     pub static ref RUNNING_NODE_TYPE: RwLock<NodeType> = RwLock::new(NodeType::Full);
```
lemmih, line 31: Use `parking_lot::RwLock` and get rid of the `lazy_static`?

types/src/build_version/mod.rs
elmattic: If it's ok to you, then yes. It will be more useful than `unstable`.

lemmih: Yes, please.

types/src/build_version/mod.rs
elmattic: Not really, this indicates the versions of the RPC api:
In Lotus they are:
```
// semver versions of the rpc api exposed
var (
	FullAPIVersion0 = newVer(1, 5, 0)
	FullAPIVersion1 = newVer(2, 3, 0)

	MinerAPIVersion0  = newVer(1, 5, 0)
	WorkerAPIVersion0 = newVer(1, 6, 0)
)
```
In forest we can keep `1.1.0` for the full node rpc api. The other are less important since we don't have miner nor worker.

types/src/build_version/mod.rs
elmattic: Done!

types/src/build_version/mod.rs
lemmih: I don't understand why we need them or what they mean. Are they required for interoperability with Lotus? Do we want interoperability with Lotus?

elmattic:
@lemmih Is this PR ok for merging? What's left to be done?

lemmih:
> @lemmih Is this PR ok for merging? What's left to be done?

I'm still unsure what the version numbers mean. You said they came from Lotus but why do we want Lotus' version numbers?

elmattic:
> > @lemmih Is this PR ok for merging? What's left to be done?
> 
> I'm still unsure what the version numbers mean. You said they came from Lotus but why do we want Lotus' version numbers?

The version that I said came from Lotus was `0.10.2`.
`const BUILD_VERSION: &str = "0.10.2"; `

This one was indeed removed.

LesnyRumcajs (approved):

lemmih:
I'm still quite confused:
- Why is the forest version string computed in the `fil_types` package rather than in `forest`?
- Why do we need a build script? It looks like the build script doesn't actually do anything (other than renaming a few env variables).
- Why do we define our own `Version` data structure?
- Can't we get rid of `NodeType` and `RUNNING_NODE_TYPE`?

elmattic:
> I'm still quite confused:
> 
> * Why is the forest version string computed in the `fil_types` package rather than in `forest`?

No reason anymore. Will change this.

> * Why do we need a build script? It looks like the build script doesn't actually do anything (other than renaming a few env variables).

Unfortunately a `const` binding can't be used because of the `!format` is not const fn.

> * Why do we define our own `Version` data structure?

Because we copied this from Lotus codebase. The rpc call wants an `u32` and this is a helper struct.

> * Can't we get rid of `NodeType` and `RUNNING_NODE_TYPE`?

Yes.

lemmih:
> > I'm still quite confused:
> > * Why do we need a build script? It looks like the build script doesn't actually do anything (other than renaming a few env variables).
> 
> Unfortunately a `const` binding can't be used because of the `format!` is not const fn.

Do we need the version string to be `const`? If it's just about performance, I'm fine with one extra call to `format!`.

node/rpc/src/common_api.rs
```diff
@@ -2,17 +2,17 @@
...
-    let v: Version = (&*RUNNING_NODE_TYPE.read().await).try_into()?;
+pub(crate) async fn version(
+    block_delay: u64,
+    forest_version: &'static str,
+) -> Result<VersionResult, JsonRpcError> {
     Ok(APIVersion {
-        version: user_version().await,
-        api_version: v,
+        version: forest_version.to_string(),
+        api_version: Version::new(0, 0, 0),
```
lemmih, line 15: We can use the version of the `rpc` library here?

elmattic: Sure! Let's do this.

lemmih:
You can also delete `ints`, `PartialEq`, and `Display` for `Version`, I think.

node/rpc/src/common_api.rs
lemmih: Or maybe pass in the forest version like you do with the static string.

elmattic: @david Are you suggesting that we use the `semver` crate?

lemmih: Yes.

tyshko5 (approved):

node/rpc/src/common_api.rs
```diff
@@ -2,17 +2,19 @@
...
+use semver::Version as SemVer;
 
-pub(crate) async fn version(block_delay: u64) -> Result<VersionResult, JsonRpcError> {
-    #[allow(clippy::needless_borrow)]
-    let v: Version = (&*RUNNING_NODE_TYPE.read().await).try_into()?;
+pub(crate) async fn version(
+    block_delay: u64,
+    forest_version: &'static str,
+) -> Result<VersionResult, JsonRpcError> {
+    let v = SemVer::parse(env!("CARGO_PKG_VERSION")).unwrap();
```
lemmih, line 14: Let's use the forest version rather than the rpc version.

lemmih (changes requested):

node/rpc/src/common_api.rs
elmattic: Ok but then I need to pass somehow forest semver.

lemmih: You can change the type of `forest_version` to be `semver::Version`, right?

lemmih: Or even just parse the `forest_version` string. :)

elmattic: Yes, this should work as `forest_version` is a valid semver string :). We just need to make sure it will always respect the syntax, ie. if we later use pre-release identifier, build number etc.

lemmih (approved):
Nice!

PR #1762, opened 2022-08-02
Path dependencies are not allowed by crates.io 

### LINKED ISSUE #1753: Path dependencies are not allowed by crates.io
**Issue summary**

In Forest, we use a lot of path dependencies like this:
https://github.com/ChainSafe/forest/blob/77c9689a237c4a8b89a7d3bd998962c6d6189c95/forest/Cargo.toml#L14

Such dependencies are not allowed if we want to publish on crates.io. Instead, we should use proper names and version numbers and then add overrides in the top-level `Cargo.toml` file:
https://github.com/ChainSafe/forest/blob/77c9689a237c4a8b89a7d3bd998962c6d6189c95/Cargo.toml#L43-L53

This issue is complete once all path dependencies have been removed.

tyshko5:
**Summary of changes**
- 

Closes #1753


LesnyRumcajs (commented):
I'd consider having all of our in-house crates prefixed with `forest_` for consistency. Having an override that would otherwise point to an existing crate may be a bit misleading, e.g. https://crates.io/crates/metrics is a quite a popular and we'd have to do some tricks to include it. What do you think? @tyshko5 @elmattic @jdjaustin ?

lemmih:
> I'd consider having all of our in-house crates prefixed with `forest_` for consistency. Having an override that would otherwise point to an existing crate may be a bit misleading, e.g. https://crates.io/crates/metrics is a quite a popular and we'd have to do some tricks to include it. What do you think? @tyshko5 @elmattic @jdjaustin ?

Big fan of using the `forest_` prefix for our own crates. Want to split the changes over two PRs or keep it all in this one, @tyshko5?

tyshko5:
> > I'd consider having all of our in-house crates prefixed with `forest_` for consistency. Having an override that would otherwise point to an existing crate may be a bit misleading, e.g. https://crates.io/crates/metrics is a quite a popular and we'd have to do some tricks to include it. What do you think? @tyshko5 @elmattic @jdjaustin ?
> 
> Big fan of using the `forest_` prefix for our own crates. Want to split the changes over two PRs or keep it all in this one, @tyshko5?

Let's do that.

LesnyRumcajs:
> > > I'd consider having all of our in-house crates prefixed with `forest_` for consistency. Having an override that would otherwise point to an existing crate may be a bit misleading, e.g. https://crates.io/crates/metrics is a quite a popular and we'd have to do some tricks to include it. What do you think? @tyshko5 @elmattic @jdjaustin ?
> > 
> > 
> > Big fan of using the `forest_` prefix for our own crates. Want to split the changes over two PRs or keep it all in this one, @tyshko5?
> 
> Let's do that.

`that` being the first or second alternative that @lemmih suggested? :smile: If the latter, then let's create a proper issue. If the former, let's put it into a draft PR until it's done.

tyshko5:
> > > > I'd consider having all of our in-house crates prefixed with `forest_` for consistency. Having an override that would otherwise point to an existing crate may be a bit misleading, e.g. https://crates.io/crates/metrics is a quite a popular and we'd have to do some tricks to include it. What do you think? @tyshko5 @elmattic @jdjaustin ?
> > > 
> > > 
> > > Big fan of using the `forest_` prefix for our own crates. Want to split the changes over two PRs or keep it all in this one, @tyshko5?
> > 
> > 
> > Let's do that.
> 
> `that` being the first or second alternative that @lemmih suggested? 😄 If the latter, then let's create a proper issue. If the former, let's put it into a draft PR until it's done.

second alternative

elmattic (approved):

connormullett (approved):
LGTM

lemmih:
@tyshko5 Feel free to merge this PR as-is. I can create a new issue for renaming the crates.

Issue #1752, opened 2022-07-28
Test that all configuration keys are assigned to a section.


lemmih:
**Issue summary**

TOML configuration keys that haven't been assigned a section must appear at the top of the file. This can be counter-intuitive and we try to avoid it by assigning a section header to all keys. Thus we need a test-case to ensure we don't accidentally get this wrong.

This could be a good place to get introduced to [quickcheck](https://docs.rs/quickcheck/latest/quickcheck/). Write a quickcheck test that generates a random configuration, serializes it to a string, and then verifies that the first element of the string is a section header (eg. `[something like this]`).

**Other information and links**

Original issue: #1488

PR #1764, opened 2022-08-03
check config params are always under section

LesnyRumcajs:
**Summary of changes**
- Introduce `quickcheck` test,
- implemented a test asserting that the Forest config file always starts with a section

Closes https://github.com/ChainSafe/forest/issues/1752

### DIFF
--- a/forest/src/cli/config.rs
+++ b/forest/src/cli/config.rs
@@ -9,7 +9,7 @@ use std::sync::Arc;
 
 use super::client::Client;
 
-#[derive(Serialize, Deserialize, PartialEq, Default)]
+#[derive(Clone, Serialize, Deserialize, PartialEq, Default)]
 #[serde(default)]
 pub struct Config {
     pub client: Client,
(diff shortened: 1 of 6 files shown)


lemmih (approved):

forest/src/cli/config.rs
```diff
@@ -9,7 +9,7 @@ use std::sync::Arc;
 
 use super::client::Client;
 
-#[derive(Serialize, Deserialize, PartialEq, Default)]
+#[derive(Clone, Serialize, Deserialize, PartialEq, Default)]
```
elmattic, line 12: Remove `Clone` since it's not used.

elmattic (changes requested):

forest/src/cli/config.rs
LesnyRumcajs: Thanks, should be fine now.

elmattic (approved):

PR #1774, opened 2022-08-05
Display Miner Vesting Table

jdjaustin:
**Summary of changes**
- Retrieve `MinerState` and display vesting table and display in a user-friendly format.

N/A

View vesting table by executing `forest state vesting-table <miner address>` CLI command with a valid `<miner address>` (e.g. `f01001`)


jdjaustin:
Update branch with changes from main

forest/Cargo.toml
```diff
@@ -82,7 +85,7 @@ rand              = "0.8"
 tempfile          = "3"
 
 [features]
-default          = ["forest_fil_cns"]
+default          = ["forest_fil_cns", "forest_statediff"]
```
lemmih, line 88: `forest_statediff` should probably be an non-optional dependency.

lemmih (changes requested):

forest/Cargo.toml
jdjaustin: Thanks, I see the root cause now; that was my bandaid solution because I completely missed the optional flag in the dependencies.

lemmih (approved):

tyshko5 (approved):

PR #1785, opened 2022-08-08
Add the output fields of wallet list subcommand

### LINKED ISSUE #1778: Add the output fields of wallet subcommand list
**Issue summary**

Add output fields "Default" and "Balance" to cli wallet list
```sh
❯ ./forest wallet list
Address                                    Default  Balance
t1ss6z757xy46h2qhxgnk4rummmjwmpbtwcnkdcki  X        0
```

jimluo:
**Summary of changes**
- 

Added  new fields "Default"  and "Balance" to wallet list subcommand

Closes 
#1778


forest/src/cli/wallet_cmd.rs
```diff
@@ -165,10 +165,23 @@ impl WalletCommands {
...
                 let response = wallet_list().await.map_err(handle_rpc_err).unwrap();
+                let default = wallet_default_address()
+                    .await
+                    .map_err(handle_rpc_err)
+                    .unwrap();
 
-                response.iter().for_each(|address| {
-                    println!("{}", address.0);
-                });
+                println!("Address                                    Default  Balance");
```
lemmih, line 173: The alignment could be done through `println`'s padding. Not a big deal, though.

lemmih (approved):
Looks okay to me.

forest/src/cli/wallet_cmd.rs
```diff
@@ -165,10 +165,23 @@ impl WalletCommands {
...
+
+                for address in response {
+                    let addr = address.0.to_string();
+                    let def = if addr == default { "X" } else { "" };
+                    let bal = wallet_balance((addr.clone(),))
+                        .await
+                        .map_err(handle_rpc_err)
+                        .unwrap();
```
LesnyRumcajs, line 181: `unwrap` in general is not a nice way to handle errors, would be great if you could change it to e.g. `expect("failed loading the wallet balance")` or something along those lines. At the same time, I can see that there are a bunch of unwraps across the file so it's just a suggestion to make it a bit better.

forest/src/cli/wallet_cmd.rs
```diff
@@ -165,10 +165,23 @@ impl WalletCommands {
...
+                    .unwrap();
 
-                response.iter().for_each(|address| {
-                    println!("{}", address.0);
-                });
+                println!("Address                                    Default  Balance");
+
+                for address in response {
+                    let addr = address.0.to_string();
+                    let def = if addr == default { "X" } else { "" };
```
LesnyRumcajs, line 177: `def` is kind of broad, how about `default_address_mark`?

forest/src/cli/wallet_cmd.rs
```diff
@@ -165,10 +165,23 @@ impl WalletCommands {
...
+                    let bal = wallet_balance((addr.clone(),))
+                        .await
+                        .map_err(handle_rpc_err)
+                        .unwrap();
+
+                    println!("{}  {:7}  {}", addr, def, bal);
```
LesnyRumcajs, line 183:
```suggestion
                    println!("{addr}  {def:7}  {bal}");
```

LesnyRumcajs, line 183: I would like to see this to be aligned for sure with the header. E.g. `{addr:<the entire length of address column as in header>}{def:<the same>}{bal:<the same>}`. Right now, looking at the code it's hard to say if all will fit.

LesnyRumcajs (changes requested):
Thanks for the contribution! Most things are details, but I would really like to see programmatically aligning the table.

forest/src/cli/wallet_cmd.rs
lemmih: The `unwrap` will never fail (`handle_rpc_err` returns `!`). We really want something like `into_ok` here if it was stable. Or we could refactor the code.

LesnyRumcajs: I see, it makes sense then. We should probably consider refactoring this in the future, I wouldn't expect `map_` to have such side effects.

jimluo:
Thank you for your advice, I will follow the advice soon to improve.

tyshko5 (approved):

forest/src/cli/wallet_cmd.rs
```diff
@@ -165,10 +166,28 @@ impl WalletCommands {
...
+                    let default_address_mark = if addr == default { "X" } else { "" };
+                    let balance_string = wallet_balance((addr.clone(),))
+                        .await
+                        .map_err(handle_rpc_err)
+                        .expect("Failed loading the wallet balance");
+                    let balance = balance_string
+                        .parse::<BigInt>()
+                        .expect("Couldn't convert balance string to bigint");
+                    let balance_fil =
+                        balance_to_fil(balance).expect("Couldn't convert balance to fil");
```
LesnyRumcajs, line 187:
```suggestion
                        balance_to_fil(balance).expect("Couldn't convert balance to FIL");
```

forest/src/cli/wallet_cmd.rs
```diff
@@ -165,10 +166,28 @@ impl WalletCommands {
...
+                for address in response {
+                    let addr = address.0.to_string();
+                    let default_address_mark = if addr == default { "X" } else { "" };
+                    let balance_string = wallet_balance((addr.clone(),))
+                        .await
+                        .map_err(handle_rpc_err)
+                        .expect("Failed loading the wallet balance");
+                    let balance = balance_string
+                        .parse::<BigInt>()
+                        .expect("Couldn't convert balance string to bigint");
```
LesnyRumcajs, line 185:
```suggestion
                        .expect("Couldn't convert balance string to BigInt");
```

LesnyRumcajs, line 185: let's print the error reason and the value provided that caused the failure

forest/src/cli/wallet_cmd.rs
LesnyRumcajs, line 187: let's print the error reason and the value provided that caused the failure

forest/src/cli/wallet_cmd.rs
```diff
@@ -165,10 +166,28 @@ impl WalletCommands {
...
                 let response = wallet_list().await.map_err(handle_rpc_err).unwrap();
+                let default = wallet_default_address()
+                    .await
+                    .map_err(handle_rpc_err)
+                    .unwrap();
 
-                response.iter().for_each(|address| {
-                    println!("{}", address.0);
-                });
+                println!("Address                                    Default  Balance");
```
LesnyRumcajs, line 174: @jimluo let's add the same explicit padding like for the data row

LesnyRumcajs (changes requested):

LesnyRumcajs (approved):

forest/src/cli/wallet_cmd.rs
jimluo: Thanks, I will fix it over the weekend

PR #1788, opened 2022-08-08
Enable cargo-spellcheck for `blockchain/`

### LINKED ISSUE #1744: Enable `cargo-spellcheck` for `blockchain/`
**Issue summary**

Make sure that `cargo spellcheck --cfg ../.config/spellcheck.toml` can run without errors in `blockchain/` For details, see the tracking issue.

connormullett:
**Summary of changes**
- clear all spellcheck errors in `blockchain/` directory

Closes #1744 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6204,9 +6204,9 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b22a693222d716a9587786f37ac3f6b4faedb5b80c23914e7303ff5a1d8016e9"
 dependencies = [
- "proc-macro2 1.0.40",
- "quote 1.0.20",
- "syn 1.0.98",
+ "proc-macro2 1.0.43",
+ "quote 1.0.21",
+ "syn 1.0.99",
 ]
 
 [[package]]
--- a/blockchain/blocks/src/ticket.rs
+++ b/blockchain/blocks/src/ticket.rs
@@ -4,7 +4,7 @@
 use forest_crypto::VRFProof;
 use forest_encoding::tuple::*;
 
-/// A Ticket is a marker of a tick of the blockchain's clock.  It is the source
+/// A Ticket is a marker of a tick of the block-chain's clock.  It is the source
 /// of randomness for proofs of storage and leader election.  It is generated
 /// by the miner of a block using a `VRF` and a `VDF`.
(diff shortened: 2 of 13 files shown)


Cargo.lock
```diff
@@ -6204,9 +6204,9 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b22a693222d716a9587786f37ac3f6b4faedb5b80c23914e7303ff5a1d8016e9"
 dependencies = [
- "proc-macro2 1.0.40",
- "quote 1.0.20",
- "syn 1.0.98",
+ "proc-macro2 1.0.43",
```
LesnyRumcajs, line 6207: Not sure why we need to change this. Seems like the same one from the other PR. If a lockfile changed globally for `main`, the best would be to issue a separate PR.

blockchain/blocks/src/ticket.rs
```diff
@@ -4,7 +4,7 @@
 use forest_crypto::VRFProof;
 use forest_encoding::tuple::*;
 
-/// A Ticket is a marker of a tick of the blockchain's clock.  It is the source
+/// A Ticket is a marker of a tick of the block-chain's clock.  It is the source
```
LesnyRumcajs, line 7: Just `blockchain` is not fine? Or should we add it until we have a better dictionary?

LesnyRumcajs (changes requested):

blockchain/blocks/src/ticket.rs
elmattic: Try to suffix blockchain in our dictionary with `/M`, `blockchain/M`.

elmattic: > It tells hunspell to also accept the possessive version of the word.

Cargo.lock
connormullett: Not sure exactly what caused the change. If I remember right when I checked out the new branch it checked out into unstaged changes (the cargo.lock was updated with these values) despite main not having this issue?? Not sure if we need a new PR for a 3 line change that I'm not sure where it even came from

blockchain/blocks/src/ticket.rs
connormullett: :+1:

LesnyRumcajs (approved):

elmattic (approved):

PR #1789, opened 2022-08-08
Enable spellcheck in `forest/`

### LINKED ISSUE #1745: Enable `cargo-spellcheck` for `forest/`
**Issue summary**

Make sure that `cargo spellcheck --cfg ../.config/spellcheck.toml` can run without errors in `forest/`. For details, see the tracking issue.

connormullett:
**Summary of changes**
- Clear spellcheck errors in `forest/`

Closes #1745 

### DIFF
--- a/.config/forest.dic
+++ b/.config/forest.dic
@@ -16,7 +16,6 @@ RPC
 SECP
 VM
 blockchain
-blockstore
 codec
 multisig
 mutex
@@ -32,4 +31,3 @@ tipset/SM
 tipsetkey/S
 tuple
 validator/S
-stringified
--- a/.config/forest.dic
+++ b/.config/forest.dic
@@ -6,6 +6,7 @@ BLS
 CBOR
 CID
 CIDs
+CLI
 Filecoin
 IPLD
 JSON
@@ -15,6 +16,7 @@ RPC
 SECP
 VM
 blockchain
+blockstore
 codec
 multisig
 mutex
@@ -30,3 +32,4 @@ tipset/SM
 tipsetkey/S
 tuple
 validator/S
+stringified
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6204,9 +6204,9 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b22a693222d716a9587786f37ac3f6b4faedb5b80c23914e7303ff5a1d8016e9"
 dependencies = [
- "proc-macro2 1.0.40",
- "quote 1.0.20",
- "syn 1.0.98",
+ "proc-macro2 1.0.43",
+ "quote 1.0.21",
+ "syn 1.0.99",
 ]
 
 [[package]]
--- a/forest/src/cli/chain_cmd.rs
+++ b/forest/src/cli/chain_cmd.rs
@@ -46,7 +46,7 @@ pub enum ChainCommands {
     Head,
 
     /// Reads and prints out a message referenced by the specified CID from the
-    /// chain blockstore
+    /// chain block store
     #[structopt(about = "<CID> Retrieves and prints messages by CIDs")]
     Message {
         #[structopt(short, help = "Input a valid CID")]
@@ -54,7 +54,7 @@ pub enum ChainCommands {
     },
 
     /// Reads and prints out IPLD nodes referenced by the specified CID from chain
-    /// blockstore and returns raw bytes
+    /// block store and returns raw bytes
     #[structopt(about = "<CID> Read the raw bytes of an object")]
     ReadObj {
         #[structopt(short, help = "Input a valid CID")]
--- a/forest/src/cli/chain_cmd.rs
+++ b/forest/src/cli/chain_cmd.rs
@@ -53,7 +53,7 @@ pub enum ChainCommands {
         cid: String,
     },
 
-    /// Reads and prints out `ipld` nodes referenced by the specified CID from chain
+    /// Reads and prints out IPLD nodes referenced by the specified CID from chain
     /// blockstore and returns raw bytes
     #[structopt(about = "<CID> Read the raw bytes of an object")]
     ReadObj {
--- a/forest/src/cli/chain_cmd.rs
+++ b/forest/src/cli/chain_cmd.rs
@@ -20,10 +20,10 @@ pub enum ChainCommands {
         cid: String,
     },
 
-    /// Export a snapshot of the chain to <output_path>
+    /// Export a snapshot of the chain to `<output_path>`
     #[structopt(about = "Export chain snapshot to file")]
     Export {
-        /// Tipset to start the export from, default is @HEAD
+        /// Tipset to start the export from, default is `@HEAD`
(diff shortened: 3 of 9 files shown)


.config/forest.dic
```diff
@@ -30,3 +32,4 @@ tipset/SM
 tipsetkey/S
 tuple
 validator/S
+stringified
```
LesnyRumcajs, line 35: It would be probably best to use some of the tricks (`/D`? not sure) to add just `stringify` without precise tense

Cargo.lock
```diff
@@ -6204,9 +6204,9 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
```
LesnyRumcajs, line 6204: you didn't modify any cargo file so it shouldn't be changed, no?

forest/src/cli/mod.rs
```diff
@@ -337,7 +337,7 @@ pub(super) fn format_vec_pretty(vec: Vec<String>) -> String {
     format!("[{}]", vec.join(", "))
 }
 
-/// convert bigint to size string using byte size units (ie KiB, GiB, PiB, etc)
+/// convert `bigint` to size string using byte size units (i.e. KiB, GiB, PiB, etc)
```
LesnyRumcajs, line 340:
```suggestion
/// convert `BigInt` to size string using byte size units (i.e. KiB, GiB, PiB, etc)
```

forest/src/cli/client.rs
```diff
@@ -24,7 +24,7 @@ pub struct Client {
     pub snapshot_height: Option<i64>,
     pub snapshot_path: Option<String>,
     /// Skips loading import CAR file and assumes it's already been loaded.
-    /// Will use the cids in the header of the file to index the chain.
+    /// Will use the `cids` in the header of the file to index the chain.
```
LesnyRumcajs, line 27:
```suggestion
    /// Will use the `CIDs` in the header of the file to index the chain.
```

forest/src/cli/chain_cmd.rs
```diff
@@ -53,7 +53,7 @@ pub enum ChainCommands {
         cid: String,
     },
 
-    /// Reads and prints out ipld nodes referenced by the specified CID from chain
+    /// Reads and prints out `ipld` nodes referenced by the specified CID from chain
```
LesnyRumcajs, line 56:
```suggestion
    /// Reads and prints out `IPLD` nodes referenced by the specified CID from chain
```

.config/forest.dic
```diff
@@ -6,6 +6,7 @@ BLS
 CBOR
 CID
 CIDs
+CLI
```
lemmih, line 9: I have half a mind to write this out as `command-line` or `command-line interface`.

forest/src/cli/mod.rs
LesnyRumcajs, line 340:
```suggestion
/// convert `bigint` to size string using byte size units (i.e. KiB, GiB, PiB, etc.)
```

LesnyRumcajs (changes requested):

.config/forest.dic
```diff
@@ -15,6 +16,7 @@ RPC
 SECP
 VM
 blockchain
+blockstore
```
lemmih, line 19: If we're talking about block stores in general then I'd put a space between the two words. If we're talking about a code concept then I'd write `BlockStore` or `Blockstore` or similar.

.config/forest.dic
LesnyRumcajs: Let's not go this way, otherwise, API may follow. Or even YAML, in which case we're doomed to death by recursion.

.config/forest.dic
lemmih, line 35: Replace with 'pretty-printed'? Maybe use another wording?

> /// Print a stringified JSON-RPC error and exit

> /// Print a pretty-printed JSON-RPC error and exit

> /// Output a pretty-printed JSON-RPC error and exit

> /// Write a pretty-printed JSON-RPC error and exit

> /// Pretty-print a JSON-RPC error and exit

.config/forest.dic
lemmih: I'm not sold on the usefulness of `CLI` as a word. Consider the following example:

```rust
/// CLI options
pub struct CliOpts { ... }
```

I don't think this is great documentation. Spelling it out seems slightly better to me:

```rust
/// Command-line options
pub struct CliOpts { ... }
```

The `CLI` word is also used here:

```rust
/// Process CLI subcommand
pub(super) async fn process(command: Subcommand, config: Config) {
```

In this case, I think it would be better to omit it all together:

```rust
/// Process subcommand
pub(super) async fn process(command: Subcommand, config: Config) {
```

I am aware that the documentation is basically useless in both cases but I do think we could try to make the documentation as human readable as possible and avoid acronyms unless they are vital to what is being documented. What are your thoughts on this?

LesnyRumcajs: It makes sense until you get a really exhaustive, long piece of documentation and then you are forced to use the full form. I'd say we should treat each occurrence separately, even prefer `command line` over `CLI`, but give folks a choice. At the same time, we wrote in this thread more than the total amassed comments including `CLI` in this 3 years-old project, so if you feel it's better then let's do it. :)

lemmih: Okay, we'll keep `CLI` for now and worry about it later when the documentation is re-written.

forest/src/cli/client.rs
elmattic: CID and CIDs acronyms are already in the dictionary.

forest/src/cli/chain_cmd.rs
elmattic: IPLD acronyms is already in the dictionary.

forest/src/cli/mod.rs
elmattic: Maybe use `BigInt` here?

connormullett: :+1:

forest/src/cli/chain_cmd.rs
connormullett: :+1:

forest/src/cli/client.rs
connormullett: :+1:

forest/src/cli/mod.rs
connormullett: :+1:

.config/forest.dic
connormullett: going to be using @lemmih idea to just reword it with "pretty-print"

.config/forest.dic
connormullett: spelling it out sounds better in this case and omitting also just reduces any confusion in cases like `process()`

.config/forest.dic
connormullett: Yea this is much better than throwing words in the dictionary

LesnyRumcajs (approved):

lemmih (approved):

PR #1805, opened 2022-08-11
Avoid using panic for user-facing errors. 

### LINKED ISSUE #1799: Avoid using `panic` for user-facing errors.
**Issue summary**

Panics are for internal errors and should never intentionally be shown to the end-user. The following six panics should be replaced with proper error handling:

- https://github.com/ChainSafe/forest/blob/7e470572033e269d75efbebae1579dcd2bebf35b/forest/src/cli/state_cmd.rs#L54
- https://github.com/ChainSafe/forest/blob/7e470572033e269d75efbebae1579dcd2bebf35b/forest/src/cli/state_cmd.rs#L104
- https://github.com/ChainSafe/forest/blob/7e470572033e269d75efbebae1579dcd2bebf35b/forest/src/cli/state_cmd.rs#L148
- https://github.com/ChainSafe/forest/blob/7e470572033e269d75efbebae1579dcd2bebf35b/forest/src/cli/fetch_params_cmd.rs#L33
- https://github.com/ChainSafe/forest/blob/7e470572033e269d75efbebae1579dcd2bebf35b/forest/src/cli/mod.rs#L209
- https://github.com/ChainSafe/forest/blob/7e470572033e269d75efbebae1579dcd2bebf35b/forest/src/cli/mod.rs#L344

tyshko5:
**Summary of changes**
- 

Closes #1799

### DIFF
--- a/forest/src/cli/mod.rs
+++ b/forest/src/cli/mod.rs
@@ -342,14 +342,21 @@ pub(super) fn format_vec_pretty(vec: Vec<String>) -> String {
 
 /// convert `BigInt` to size string using byte size units (i.e. KiB, GiB, PiB, etc)
 /// Provided number cannot be negative, otherwise the function will panic.
-pub(super) fn to_size_string(input: &BigInt) -> String {
-    Byte::from_bytes(
-        u128::try_from(input).unwrap_or_else(|e| {
-            cli_error_and_die(format!("error parsing the input {input}: {e}"), 1)
-        }),
-    )
-    .get_appropriate_unit(true)
-    .to_string()
+pub(super) fn to_size_string(input: &BigInt) -> Result<String, String> {
+    let bytes = match u128::try_from(input) {
+        Ok(value) => value,
+        Err(e) => {
+            return Err(format!(
+                "error parsing the input {}: {}",
+                input.clone(),
+                e.clone()
+            ))
+        }
+    };
+
+    Ok(Byte::from_bytes(bytes)
+        .get_appropriate_unit(true)
+        .to_string())
 }
 
 /// Print an error message and exit the program with an error code
@@ -456,19 +463,19 @@ mod test {
         ];
 
         for (input, expected) in cases {
-            assert_eq!(to_size_string(&input), expected.to_string());
+            assert_eq!(to_size_string(&input).unwrap(), expected.to_string());
         }
     }
 
     #[test]
     #[should_panic]
     fn to_size_string_negative_input_should_fail() {
-        to_size_string(&BigInt::from(-1i8));
+        to_size_string(&BigInt::from(-1i8)).unwrap();
     }
 
     #[test]
     #[should_panic]
     fn to_size_string_too_large_input_should_fail() {
-        to_size_string(&(BigInt::from(u128::MAX) + 1));
+        to_size_string(&(BigInt::from(u128::MAX) + 1)).unwrap();
     }
 }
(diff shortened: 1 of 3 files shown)


forest/src/cli/mod.rs
```diff
@@ -339,12 +342,21 @@ pub(super) fn format_vec_pretty(vec: Vec<String>) -> String {
...
+    let bytes = match u128::try_from(input) {
+        Ok(value) => value,
+        Err(e) => {
+            return Err(format!(
+                "error parsing the input {}: {}",
+                input.clone(),
+                e.clone()
+            ))
+        }
+    };
```
lemmih, line 355: You could use `map_err` and `?`.

forest/src/cli/mod.rs
```diff
@@ -451,19 +463,19 @@ mod test {
         ];
 
         for (input, expected) in cases {
-            assert_eq!(to_size_string(&input), expected.to_string());
+            assert_eq!(to_size_string(&input).unwrap(), expected.to_string());
```
lemmih, line 466: Maybe something like this:

```suggestion
            assert_eq!(to_size_string(&input), Ok(expected.to_string()));
```

forest/src/cli/mod.rs
```diff
@@ -451,19 +463,19 @@ mod test {
...
         }
     }
 
     #[test]
     #[should_panic]
     fn to_size_string_negative_input_should_fail() {
-        to_size_string(&BigInt::from(-1i8));
+        to_size_string(&BigInt::from(-1i8)).unwrap();
```
lemmih, line 473: This might be more idiomatic:

```suggestion
        assert!(to_size_string(&BigInt::from(-1i8)).is_err());
```

lemmih (approved):
Nice. I do wonder if the `run` method should return a `Result`. Hmm, we can think about that in another PR.

forest/src/cli/mod.rs
```diff
@@ -339,12 +342,13 @@ pub(super) fn format_vec_pretty(vec: Vec<String>) -> String {
 
 /// convert `BigInt` to size string using byte size units (i.e. KiB, GiB, PiB, etc)
 /// Provided number cannot be negative, otherwise the function will panic.
-pub(super) fn to_size_string(input: &BigInt) -> String {
-    Byte::from_bytes(
-        u128::try_from(input).unwrap_or_else(|e| panic!("error parsing the input {input}: {e}")),
-    )
-    .get_appropriate_unit(true)
-    .to_string()
+pub(super) fn to_size_string(input: &BigInt) -> Result<String, String> {
```
LesnyRumcajs, line 345: Perhaps let's use `anyhow::Result`?

forest/src/cli/mod.rs
```diff
@@ -339,12 +342,13 @@ pub(super) fn format_vec_pretty(vec: Vec<String>) -> String {
...
+    let bytes = u128::try_from(input)
+        .map_err(|e| return format!("error parsing the input {}: {}", input.clone(), e.clone()))?;
```
LesnyRumcajs, line 347: do we need `return` here?

LesnyRumcajs, line 347: why do we need to clone those values, especially the error?

LesnyRumcajs (changes requested):

forest/src/cli/mod.rs
tyshko5: Deleted

tyshko5: Deleted

forest/src/cli/mod.rs
tyshko5: For now we are using basic `Result` everywhere there, so I use it too

connormullett: `anyhow::Result` potentially could help us with those nasty map_err statements. We use regularly all over the codebase but this could be a separate issue. @lemmih thoughts?

connormullett (approved):
Aside from one comment, looks good. Approving to not hold up after next reviewer

PR #1816, opened 2022-08-11
Cleanly shutdown all tasks and make sure RocksDB is closed properly

elmattic:
**Summary of changes**
- Fix of some resource leaks due to not canceling mpool tasks
- Add some yielding for faster task interruption (during blockchain scan)
- Add of debug code to print refcounts of the db value (this can be commented later)

While this fixes the segfault we were seeing, it doesn't fix them all. The cause is more profound in forest because when we spawn tasks we don't always handle their cancelation.


lemmih:
Thanks, guillaume. I'll have a look.

elmattic:
> Thanks, guillaume. I'll have a look.

Thanks. Cleaning up and put it RFR.

lemmih (approved):

PR #1821, opened 2022-08-12
chain export: revert default value for recent stateroots

LesnyRumcajs:
**Summary of changes**
- move back to the default value 2000 for recent stateroots, seems the minimal one enforced before (chain finality) is too small and was causing issues when such snapshot is imported.


lemmih (approved):
All hail the magic numbers!

elmattic (approved):

PR #1831, opened 2022-08-16
Start RPC server before loading snapshot

### LINKED ISSUE #1800: Start the RPC server as soon as possible (before loading snapshots)
**Issue summary**

The Forest daemon initiates the RPC server after loading snapshot. This is less than ideal since loading a snapshot can take quite a while and we would like the RPC server up-and-running as soon as possible. The code should be refactored such that the RPC server is booted as soon as it can be.

connormullett:
**Summary of changes**
- Change the ordering of tasks in daemon.rs to start the RPC server prior to loading the snapshot

Closes #1800 

### DIFF
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -298,6 +298,15 @@ pub(super) async fn start(config: Config) {
     select! {
         () = sync_from_snapshot(&config, &state_manager).fuse() => {},
         _ = ctrlc_oneshot => {
+            // Cancel all async services
+            for mining_task in mining_tasks {
+                mining_task.cancel().await;
+            }
+            prometheus_server_task.cancel().await;
+            head_changes_task.cancel().await;
+            republish_task.cancel().await;
+            sync_task.cancel().await;
+            maybe_cancel(rpc_task).await;
             return;
         },
     }
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -188,6 +188,25 @@ pub(super) async fn start(config: Config) {
 
     let (tipset_sink, tipset_stream) = bounded(20);
 
+    // if bootstrap peers are not set, set them
+    let config = if config.network.bootstrap_peers.is_empty() {
+        let bootstrap_peers = config
+            .chain
+            .bootstrap_peers
+            .iter()
+            .map(|node| node.parse().unwrap())
+            .collect();
+        Config {
+            network: Libp2pConfig {
+                bootstrap_peers,
+                ..config.network
+            },
+            ..config
+        }
+    } else {
+        config
+    };
+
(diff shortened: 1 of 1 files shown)


forest/src/daemon.rs
```diff
@@ -339,6 +294,56 @@ pub(super) async fn start(config: Config) {
...
+        };
+        set_proofs_parameter_cache_dir_env(&config.client.data_dir);
+
+        get_params_default(&config.client.data_dir, SectorSizeOpt::Keys, false)
+            .await
+            .unwrap();
+    }
+
+    // Start services
+    let p2p_task = task::spawn(async {
```
LesnyRumcajs, line 343: An edge case, but IMO worth considering. What will happen if we call some of the commands like `forest net peers` while the snapshot is loading? The p2p service was initialized before RPC, right now I'm not entirely sure what will happen and do we have a nice error handling.

lemmih: Good question. Right `forest net peers` blocks until the p2p service is up and running. I feel this is alright but I haven't given it much thought.

lemmih:
I modified the code to properly shutdown if ctrl-c is detected.

forest/src/daemon.rs
LesnyRumcajs: I'd say an immediate answer that the Forest service is busy may be more convenient for whoever is integrating with it, but it's fine for now, we should revisit those calls later on when covering them with a nice suite of tests.

forest/src/daemon.rs
```diff
@@ -298,48 +265,101 @@ pub(super) async fn start(config: Config) {
...
             .unwrap_or_else(|_| cli_error_and_die("could not bind to {rpc_address}", 1));
 
-        Some(task::spawn(async move {
+        let rpc_state_manager = Arc::clone(&state_manager);
+        let rpc_chain_store = Arc::clone(&chain_store);
+
+        services.push(task::spawn(async move {
             info!("JSON-RPC endpoint started at {}", config.client.rpc_address);
-            start_rpc::<_, _, FullVerifier, cns::FullConsensus>(
+            // XXX: The JSON error message are a nightmare to print.
```
LesnyRumcajs, line 289: I'm not sure what does this comment refers to? do we need it?

LesnyRumcajs (approved):

forest/src/daemon.rs
lemmih: The line below silently ignores a json-rpc error. Ideally I would like to log it but there's no standard way of pretty-printing those errors.

lemmih: (The error was also silently ignored previously.)

lemmih (approved):

PR #1835, opened 2022-08-17
Add `--halt-after-import` flag to CLI

### LINKED ISSUE #1834: Add flag for exiting immediately after a snapshot has been imported.
**Issue summary**

Forest can populate its database from a CAR file using the `--import-snapshot` flag. For performance testing, it can be useful to import a snapshot and then immediately exit. The could possibly be named `--import-snapshot-and-exit` and it's fine to omit it from the `--help` documentation.

connormullett:
**Summary of changes**
- Add ability to import a snapshot and immediately exit after completion
- Small improvements to RocksDb configuration ergonomics with set_auto_disable_compactions
- Add log level from string for RocksDb for testing and debugging purposes

Closes #1834 


forest/src/cli/client.rs
```diff
@@ -23,6 +23,7 @@ pub struct Client {
     pub snapshot: bool,
     pub snapshot_height: Option<i64>,
     pub snapshot_path: Option<String>,
+    pub exit_after_import: bool,
```
elmattic, line 26: Would be great to use the same wording as in Lotus which is `--halt-after-import`.

https://lotus.filecoin.io/lotus/manage/chain-management/#restoring-a-custom-snapshot

Most of Forest cli args are already similar to theirs.

connormullett: Oh, I didn't know this was a thing. I'll rename it

forest/src/daemon.rs
```diff
@@ -354,6 +354,12 @@ async fn sync_from_snapshot(config: &Config, state_manager: &Arc<StateManager<Ro
         } else {
             Some(0)
         };
+
+        if config.client.halt_after_import {
+            info!("Forest finish shutdown");
+            process::exit(0);
```
lemmih, line 360: @elmattic Do you think it'll cause problems with RocksDB if we call `exit(0)` here? Do we need to properly shutdown the database?

elmattic: Sure we should avoid this and this is possible to do (from my experience, the main source of resource leaks are situated after the import).

But more problematic with this PR is that the code exits before `import_chain` in case of user `--halt-after-import`, resulting in a no-op. We need to do it after the import.

Plus we need also to run this code:
```
   match chain_store.heaviest_tipset().await { ...
```
This is needed to check that the user is not importing to old snapshots.

If you do just after:
```
    // Halt
    if config.client.halt_after_import {
        info!("Forest finished import");
        return
    }
```
this should terminate forest properly. Also make sure you check rocksdb LOG file to verify that it closed properly the db (you need to enable rocksdb debug log level).

elmattic: I think we could also promote this log `debug!("Imported snapshot in: {}s", ...` to an `info!` one.

This is useful information to know for troubleshooting, and it won't really bloat our logs.

forest/src/cli/mod.rs
```diff
@@ -143,6 +143,8 @@ pub struct CliOpts {
     pub height: Option<i64>,
     #[structopt(long, help = "Import a snapshot from a local CAR file or url")]
     pub import_snapshot: Option<String>,
+    #[structopt(long)]
+    pub halt_after_import: bool,
```
elmattic, line 147: Please add some doc comment help for this new flag. Maybe refactor to use doc comments for the other flags too.

elmattic: @lemmy do you think such self-describing flags need help comments?

lemmih: Yes. The help message could even add more details such as "Halt with exit code 0 after successfully importing a snapshot".

forest/src/cli/mod.rs
```diff
@@ -143,6 +143,11 @@ pub struct CliOpts {
     pub height: Option<i64>,
     #[structopt(long, help = "Import a snapshot from a local CAR file or url")]
     pub import_snapshot: Option<String>,
+    #[structopt(
+        long,
+        help = "Halt with exit code 0 after successfully importing a snapshot"
```
elmattic, line 148: You can put this into a doc comment like this:

```
/// Halt with exit code 0 after successfully importing a snapshot
#[structopt(long)]
pub halt_after_import: bool,
```

node/db/src/rocks_config.rs
```diff
@@ -30,9 +31,10 @@ impl Default for RocksDbConfig {
             write_buffer_size: 256 * 1024 * 1024,
             max_open_files: 1024,
             max_background_jobs: None,
-            compaction_style: None,
+            compaction_style: Some("level".into()),
             compression_type: Some("lz4".into()),
             enable_statistics: false,
+            log_level: "debug".into(),
```
elmattic, line 37: Don't activate `debug` log by default here. What do you think of `warn`?

connormullett: That makes more sense for a default, yeah

node/db/src/rocks.rs
```diff
@@ -40,13 +42,19 @@ impl RocksDb {
         }
         if let Some(compaction_style) = &config.compaction_style {
             db_opts.set_compaction_style(compaction_style_from_str(compaction_style).unwrap());
+            db_opts.set_disable_auto_compactions(false);
+        } else {
+            db_opts.set_disable_auto_compactions(true);
```
lemmih, line 47: This appears to be unrelated to the `--halt-after-import` flag. Could you explain the RocksDB changes in the PR description?

lemmih (approved):
Tentative approval. The RocksDB changes are probably fine but I think they're unrelated to the original goal of this PR. Please describe what you've changed and why in the PR description.

jdjaustin (approved):

PR #1845, opened 2022-08-22
Validate the ChainConfig struct at construction time 

### LINKED ISSUE #1579: Validate at construction time the ChainConfig struct
**Issue summary**

We would like to validate the new ChainConfig (introduced by #1482) at construction time.
- making sure all variants of the enum `Height` have a corresponding entry in `height_infos`
- number of `fvm_shared::NetworkVersion` variants should match number of `Height` variants + 1

tyshko5:
**Summary of changes**
- 

Closes #1579

### DIFF
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -50,6 +50,8 @@ pub(super) async fn start(config: Config) {
         FOREST_VERSION_STRING.as_str()
     );
 
+    config.chain.validate().unwrap();
+
     let path: PathBuf = config.client.data_dir.join("libp2p");
     let net_keypair = get_keypair(&path.join("keypair")).unwrap_or_else(|| {
         // Keypair not found, generate and save generated keypair
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -50,7 +50,12 @@ pub(super) async fn start(config: Config) {
         FOREST_VERSION_STRING.as_str()
     );
 
-    config.chain.validate().unwrap();
+    config.chain.validate().unwrap_or_else(|err_string| {
+        cli_error_and_die(
+            err_string,
+            1,
+        )
+    });
 
     let path: PathBuf = config.client.data_dir.join("libp2p");
     let net_keypair = get_keypair(&path.join("keypair")).unwrap_or_else(|| {
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -50,11 +50,6 @@ pub(super) async fn start(config: Config) {
         FOREST_VERSION_STRING.as_str()
     );
 
-    config
-        .chain
-        .validate()
-        .unwrap_or_else(|err_string| cli_error_and_die(err_string, 1));
-
     let path: PathBuf = config.client.data_dir.join("libp2p");
     let net_keypair = get_keypair(&path.join("keypair")).unwrap_or_else(|| {
         // Keypair not found, generate and save generated keypair
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -50,12 +50,10 @@ pub(super) async fn start(config: Config) {
         FOREST_VERSION_STRING.as_str()
     );
 
-    config.chain.validate().unwrap_or_else(|err_string| {
-        cli_error_and_die(
-            err_string,
-            1,
-        )
-    });
+    config
+        .chain
+        .validate()
+        .unwrap_or_else(|err_string| cli_error_and_die(err_string, 1));
 
     let path: PathBuf = config.client.data_dir.join("libp2p");
     let net_keypair = get_keypair(&path.join("keypair")).unwrap_or_else(|| {
--- a/types/networks/src/lib.rs
+++ b/types/networks/src/lib.rs
@@ -209,9 +209,7 @@ impl ChainConfig {
             .map(|info| info.height)
             .unwrap_or(Height::Breeze);
 
-        // unfallible as it is guaranteed that number of heights
-        // is equal to the number of network versions
-        TryFrom::try_from(height as u32).unwrap()
+        From::from(height)
     }
 
     pub async fn get_beacon_schedule(
--- a/types/networks/src/lib.rs
+++ b/types/networks/src/lib.rs
@@ -42,6 +42,26 @@ pub enum Height {
     Skyr,
 }
 
+pub const HEIGHT_VEC: [Height; 17] = [
+    Height::Breeze,
+    Height::Smoke,
+    Height::Ignition,
+    Height::ActorsV2,
+    Height::Tape,
+    Height::Liftoff,
+    Height::Kumquat,
+    Height::Calico,
+    Height::Persian,
+    Height::Orange,
+    Height::Trust,
+    Height::Norwegian,
+    Height::Turbo,
+    Height::Hyperdrive,
+    Height::Chocolate,
+    Height::OhSnap,
+    Height::Skyr,
+];
+
 impl Default for Height {
     fn default() -> Height {
         Self::Breeze
@@ -234,7 +254,12 @@ impl ChainConfig {
     }
 
     pub fn epoch(&self, height: Height) -> ChainEpoch {
-        let height_infos = sort_by_epoch(self.height_infos.clone());
+        let height_vec: Vec<Height> = self.height_infos.iter().map(|value| value.height).collect();
+        let height_set: HashSet<Height> = HashSet::from_iter(height_vec);
+        let mut height_infos = sort_by_epoch(self.height_infos.clone());
+        for &height in HashSet::from_iter(HEIGHT_VEC).difference(&height_set) {
+            height_infos.push(HeightInfo { height, epoch: 0 });
+        }
         height_infos
             .iter()
             .find(|info| height == info.height)
--- a/types/networks/src/lib.rs
+++ b/types/networks/src/lib.rs
@@ -1,7 +1,5 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
-#![feature(variant_count)]
-
 #[macro_use]
 extern crate lazy_static;
 
@@ -10,7 +8,6 @@ use forest_beacon::{BeaconPoint, BeaconSchedule, DrandBeacon, DrandConfig};
 use fvm_shared::clock::{ChainEpoch, EPOCH_DURATION_SECONDS};
 use fvm_shared::sector::{RegisteredPoStProof, RegisteredSealProof, StoragePower};
 use fvm_shared::version::NetworkVersion;
-use static_assertions::const_assert_eq;
 
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
@@ -19,11 +16,6 @@ use std::sync::Arc;
 mod calibnet;
 mod drand;
 mod mainnet;
-
-const_assert_eq!(
-    std::mem::variant_count::<NetworkVersion>(),
-    std::mem::variant_count::<Height>()
-);
 
 /// Newest network version for all networks
 pub const NEWEST_NETWORK_VERSION: NetworkVersion = NetworkVersion::V16;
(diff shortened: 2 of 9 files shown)


forest/src/daemon.rs
```diff
@@ -50,6 +50,8 @@ pub(super) async fn start(config: Config) {
         FOREST_VERSION_STRING.as_str()
     );
 
+    config.chain.validate().unwrap();
```
lemmih, line 53: Let's use `cli_error_and_die`.

types/networks/src/lib.rs
```diff
@@ -151,6 +173,34 @@ pub struct ChainConfig {
     pub policy: Policy,
 }
 
+impl ChainConfig {
+    pub fn validate(&self) -> Result<(), String> {
+        let number_of_network_version_variants = std::mem::variant_count::<NetworkVersion>();
+        let number_of_height_variants = std::mem::variant_count::<Height>();
+        if number_of_network_version_variants + 1 != number_of_height_variants {
+            return Err(
```
lemmih, line 181: Can we print out which network versions are missing?

lemmih (changes requested):

types/networks/src/lib.rs
tyshko5: What do you mean by missing, as we aren't storing network versions?

elmattic: @tyshko5 I think you can just a have a simple loop over all `HEIGHT_VARIANTS` and check if _this_ height is present in the `height_infos` vector of the given config. If not then bail out with an error saying that the config is missing _this_ height.

lemmih: @elmattic How about:
- we default to epoch 0 if an entry is missing from the config, 
- we sort the `height_infos` in `network_version()` such that the order doesn't matter,
- we get rid of `version_schedule` since it should be constant for all networks.

Then we wouldn't have to validate the chain configuration.

elmattic: Yes, I had this in mind too for the third bullet. We shoud remove the `version_schedule` from the config.

elmattic: I'm ok with the sort. But if we are defaulting to 0 for let's say the Skyr upgrade (ie Skyr's epoch is missing in the custom config file), what will happen in case we start importing a nv15 snapshot, then trying to upgrade to nv16?
The goal of config validation to me is to avoid any bad surprises when operating a node.

elmattic: What was the purpose of `Claus` height in the filecoin protocol? We don't really use it anymore in our codebase, should we keep it nonetheless?

lemmih: > What was the purpose of `Claus` height in the filecoin protocol? We don't really use it anymore in our codebase, should we keep it nonetheless?

Nah, let's delete it.

lemmih: @tyshko5 I've updated the original issue with a new description of what needs to be done.

types/networks/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
+#![feature(variant_count)]
```
LesnyRumcajs, line 3: This seems like a shaky feature, not only we would be bolting ourselves to nightly, but also in future it may not fly https://github.com/rust-lang/rust/issues/73662#issuecomment-1183512441

If we want to compare enum lengths, maybe it would be "more stable" with https://docs.rs/strum/0.24.0/strum/derive.EnumCount.html

types/networks/src/lib.rs
```diff
@@ -151,6 +173,34 @@ pub struct ChainConfig {
     pub policy: Policy,
 }
 
+impl ChainConfig {
+    pub fn validate(&self) -> Result<(), String> {
+        let number_of_network_version_variants = std::mem::variant_count::<NetworkVersion>();
+        let number_of_height_variants = std::mem::variant_count::<Height>();
+        if number_of_network_version_variants + 1 != number_of_height_variants {
```
LesnyRumcajs, line 180: This looks like it should be a compile check not a runtime one, no?

types/networks/src/lib.rs
```diff
@@ -151,6 +173,34 @@ pub struct ChainConfig {
...
+                "length of height_infos vector was smaller than the number of HeightInfo variants"
+                    .to_string(),
+            )
+        } else {
+            let mut height_set: HashSet<Height> = HashSet::from(HEIGHT_VARIANTS);
+            for item in &self.height_infos {
+                height_set.remove(&item.height);
+            }
+            if !height_set.is_empty() {
+                return Err("not all variants of the enum Height have a corresponding entry in height_infos".to_string());
```
LesnyRumcajs, line 197: let's print missing ones?

types/networks/src/lib.rs
```diff
@@ -151,6 +173,34 @@ pub struct ChainConfig {
...
+        let number_of_height_variants = std::mem::variant_count::<Height>();
+        if number_of_network_version_variants + 1 != number_of_height_variants {
+            return Err(
+                "number of fvm_shared::NetworkVersion variants does not match number of Height variants + 1"
+                    .to_string(),
+            );
+        }
+        if self.height_infos.len() == number_of_height_variants {
+            Err(
+                "length of height_infos vector was smaller than the number of HeightInfo variants"
```
LesnyRumcajs, line 188: let's print the lengths as well to be more helpful

LesnyRumcajs (changes requested):

types/networks/src/lib.rs
LesnyRumcajs, line 3: where is it used now?

types/networks/src/lib.rs
```diff
@@ -131,6 +64,10 @@ pub struct HeightInfo {
     pub epoch: ChainEpoch,
 }
 
+pub fn normalize(height_info_vec: &mut [HeightInfo]) {
```
LesnyRumcajs, line 67: `sort_by_epoch` would be a more meaningful name here. Also, it would be cleaner to get rid of those references and just take by value and return by value since you're cloning the `height_infos` anyway.

types/networks/src/lib.rs
```diff
@@ -242,19 +180,18 @@ impl ChainConfig {
...
             .find(|info| epoch > info.epoch)
             .map(|info| info.height)
             .unwrap_or(Height::Breeze);
 
-        self.version_schedule
-            .iter()
-            .find(|info| height == info.height)
-            .map(|info| info.version)
-            .expect("A network version should exist even if not specified in the config (a default exists).")
+        // unfallible as it is guaranteed that number of heights
```
LesnyRumcajs, line 192: where is the static check that guarantees this?

tyshko5: Added a compile time check for this

types/networks/src/lib.rs
tyshko5: For compile-time check

types/networks/src/lib.rs
tyshko5: Fixed

tyshko5:
> @tyshko5 You can write a `From<Height> for NetworkVersion`. It would be a total function and guaranteed to never fail.

I've already added a compile-time check

lemmih:
> > @tyshko5 You can write a `From<Height> for NetworkVersion`. It would be a total function and guaranteed to never fail.
> 
> I've already added a compile-time check

Right but you can't use `#![feature(variant_count)]` (we want to avoid +unstable as soon as possible) and you can't assume `Height` is in the same order as `NetworkVersion`. A trait implementation that maps from `Height` to `NetworkVersion` would be stable rust and free of magic.

types/networks/src/lib.rs
```diff
@@ -242,19 +201,17 @@ impl ChainConfig {
...
+        // is equal to the number of network versions
+        TryFrom::try_from(height as u32).unwrap()
```
lemmih, line 214: This is no longer necessary. You can use your `From` impl.

types/networks/src/lib.rs
```diff
@@ -1,6 +1,5 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
-
 #[macro_use]
 extern crate lazy_static;
```
lemmih, line 4: Can this be removed?

lemmih (changes requested):

types/networks/src/lib.rs
tyshko5: Unfortunately not

lemmih: Why not?

lemmih: Oh, we're using `lazy_static!` in drand.rs.

lemmih: I think I would prefer to import the macro explicitly but I guess that can wait until another PR.

types/networks/src/lib.rs
```diff
@@ -279,7 +234,8 @@ impl ChainConfig {
     }
 
     pub fn epoch(&self, height: Height) -> ChainEpoch {
```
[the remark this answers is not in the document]
lemmih: This seems more complicated than it needs to be. How about you just replace the `.expect("Internal error: ...")` with `.unwrap_or(0)`?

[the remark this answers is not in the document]
tyshko5: Fixed

LesnyRumcajs (approved):

types/networks/src/lib.rs
```diff
@@ -111,12 +42,56 @@ pub enum Height {
...
+    Height::Orange,
+    Height::Trust,
+    Height::Norwegian,
+    Height::Turbo,
+    Height::Hyperdrive,
+    Height::Chocolate,
+    Height::OhSnap,
+    Height::Skyr,
+];
+
```
lemmih, line 64: This is unused.

types/networks/src/lib.rs
```diff
@@ -111,12 +42,56 @@ pub enum Height {
...
 
+impl From<Height> for NetworkVersion {
+    fn from(height: Height) -> NetworkVersion {
+        match height {
+            Height::Breeze => NetworkVersion::V0,
+            Height::Smoke => NetworkVersion::V1,
+            Height::Ignition => NetworkVersion::V2,
+            Height::ActorsV2 => NetworkVersion::V3,
+            Height::Tape => NetworkVersion::V4,
+            Height::Liftoff => NetworkVersion::V5,
```
lemmih, line 79: `Liftoff` should be deleted.

types/networks/src/lib.rs
```diff
@@ -111,12 +42,56 @@ pub enum Height {
...
 }
 
+impl From<Height> for NetworkVersion {
+    fn from(height: Height) -> NetworkVersion {
+        match height {
+            Height::Breeze => NetworkVersion::V0,
+            Height::Smoke => NetworkVersion::V1,
+            Height::Ignition => NetworkVersion::V2,
+            Height::ActorsV2 => NetworkVersion::V3,
+            Height::Tape => NetworkVersion::V4,
```
lemmih, line 78: Bump everything such that `Breeze` is `V1` and `Tape` is `V5`. See the deleted `UPGRADE_INFOS` for ground truth.

types/networks/src/lib.rs
```diff
@@ -131,6 +106,11 @@ pub struct HeightInfo {
     pub epoch: ChainEpoch,
 }
 
+pub fn sort_by_epoch(mut height_info_vec: Vec<HeightInfo>) -> Vec<HeightInfo> {
+    height_info_vec.sort_by(|a, b| a.epoch.cmp(&b.epoch));
+    height_info_vec
+}
+
```
lemmih, line 113: If this function takes a slice of `HeightInfo` (and turns it into a Vec) then the rest of the code will become simpler.

lemmih (changes requested):

types/networks/src/lib.rs
```diff
@@ -242,19 +202,15 @@ impl ChainConfig {
     }
 
     pub fn network_version(&self, epoch: ChainEpoch) -> NetworkVersion {
-        let height = self
-            .height_infos
+        let height_infos = sort_by_epoch(&self.height_infos);
+        let height = height_infos
```
lemmih, line 206:
```suggestion
        let height = sort_by_epoch(&self.height_infos)
```

types/networks/src/lib.rs
```diff
@@ -279,11 +235,12 @@ impl ChainConfig {
     }
 
     pub fn epoch(&self, height: Height) -> ChainEpoch {
-        self.height_infos
+        let height_infos = sort_by_epoch(&self.height_infos);
+        height_infos
```
lemmih, line 239:
```suggestion
        sort_by_epoch(&self.height_infos)
```

types/networks/src/lib.rs
```diff
@@ -155,11 +115,12 @@ pub struct ChainConfig {
...
 impl PartialEq for ChainConfig {
     fn eq(&self, other: &Self) -> bool {
+        let height_infos = sort_by_epoch(&self.height_infos);
+        let other_height_infos = sort_by_epoch(&other.height_infos);
         self.name == other.name
             && self.bootstrap_peers == other.bootstrap_peers
             && self.block_delay_secs == other.block_delay_secs
-            && self.version_schedule == other.version_schedule
-            && self.height_infos == other.height_infos
+            && height_infos == other_height_infos
```
lemmih, line 123:
```suggestion
            && sort_by_epoch(&self.height_infos) == sort_by_epoch(&other.height_infos)
```

lemmih (approved):
Good. Only a few style suggestions.

elmattic:
@tyshko5 Could you update the delegate consensus code that is still using `Claus` height in its toml settings file?

types/networks/src/lib.rs
```diff
@@ -242,19 +202,15 @@ impl ChainConfig {
     }
 
     pub fn network_version(&self, epoch: ChainEpoch) -> NetworkVersion {
-        let height = self
-            .height_infos
+        let height_infos = sort_by_epoch(&self.height_infos);
```
elmattic, line 205: Just a question, isn't it possible to sort by epoch at config creation time, so only once?

tyshko5: As we are getting Config from file, it is not

tyshko5:
> @tyshko5 Could you update the delegate consensus code that is still using `Claus` height in its toml settings file?

Done

types/networks/src/lib.rs
elmattic: Ah I see, thanks.

PR #1851, opened 2022-08-23
Use Config File to Drive Components Log Configuration 

### LINKED ISSUE #1520: Use Config File to Drive Components Log Configuration
**Issue summary**
Originally suggested by @elmattic in https://github.com/ChainSafe/forest/issues/1504#issuecomment-1081788726 to allow Forest to read logging configuration from file as opposed to only CLI parameters (similar to Gossamer implementation). Creating this issue to decide if this is something the team wants to implement; team to determine design/approach if approved.

tyshko5:
**Summary of changes**
- 

Closes #1520


lemmih (changes requested):
I think there's some confusion here. The `--log-file` flag is for directing log messages to a file rather than stdout. It shouldn't read a log configuration from that file.

The `LogConfig` should be part of the regular configuration file. And those log options shouldn't be an alternative to the default options, they should just be applied after the default options have been set.

tyshko5:
> I think there's some confusion here. The `--log-file` flag is for directing log messages to a file rather than stdout. It shouldn't read a log configuration from that file.
> 
> The `LogConfig` should be part of the regular configuration file. And those log options shouldn't be an alternative to the default options, they should just be applied after the default options have been set.

Second part is fixed, but the logger we are using, pretty_env_logger, doesn't have a feature to save logs to a file. Should we change the logger?

forest/src/logger/mod.rs
```diff
@@ -1,9 +1,12 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use crate::cli::config::LogConfig;
+use crate::cli_error_and_die;
 use log::LevelFilter;
+use std::str::FromStr;
 
-pub(crate) fn setup_logger() {
+pub(crate) fn setup_logger(option_log_config: Option<LogConfig>) {
```
lemmih, line 9: A reference to `LogConfig` is enough, right? That way we don't need to clone it from the configuration.

```suggestion
pub(crate) fn setup_logger(option_log_config: Option<&LogConfig>) {
```

forest/src/main.rs
```diff
@@ -3,15 +3,14 @@
 
 mod cli;
 mod daemon;
-mod logger;
+pub mod logger;
```
lemmih, line 6: Does this need to be `pub`?

forest/src/main.rs
```diff
@@ -3,15 +3,14 @@
...
 mod subcommand;
 
 use cli::{cli_error_and_die, Cli};
 use structopt::StructOpt;
 
 #[async_std::main]
 async fn main() {
-    logger::setup_logger();
```
lemmih, line 14: You forgot to initialize the logger when we start the daemon.

lemmih (changes requested):
Getting closer. Still need the `--log-file` flag (or a comment saying why it isn't possible).

tyshko5:
> Getting closer. Still need the `--log-file` flag (or a comment saying why it isn't possible).

the logger we are using, pretty_env_logger, doesn't have a feature to save logs to a file. Should we change the logger?

forest/src/main.rs
tyshko5: I'm doing it in `forest/src/subcommand.rs` as I need to get the config first

forest/src/logger/mod.rs
tyshko5: In this case we need to introduce a lifetime, so cloning is better, I guess

forest/src/main.rs
tyshko5: Fixed

lemmih:
> > Getting closer. Still need the `--log-file` flag (or a comment saying why it isn't possible).
> 
> the logger we are using, pretty_env_logger, doesn't have a feature to save logs to a file. Should we change the logger?

`pretty_env_logger` is layered on top of `env_logger`. You can get the `Builder` for `env_logger` and then change the `target`.

forest/src/logger/mod.rs
lemmih: I don't think you need to introduce a lifetime.

forest/src/main.rs
tyshko5: Thanks, missed this. Fixed

forest/src/logger/mod.rs
tyshko5: Fixed

tyshko5:
> > > Getting closer. Still need the `--log-file` flag (or a comment saying why it isn't possible).
> > 
> > 
> > the logger we are using, pretty_env_logger, doesn't have a feature to save logs to a file. Should we change the logger?
> 
> `pretty_env_logger` is layered on top of `env_logger`. You can get the `Builder` for `env_logger` and then change the `target`.

The only targets that can be used by  `env_logger` are stdout and stderr

lemmih:
> > > > Getting closer. Still need the `--log-file` flag (or a comment saying why it isn't possible).
> > > 
> > > 
> > > the logger we are using, pretty_env_logger, doesn't have a feature to save logs to a file. Should we change the logger?
> > 
> > 
> > `pretty_env_logger` is layered on top of `env_logger`. You can get the `Builder` for `env_logger` and then change the `target`.
> 
> The only targets that can be used by `env_logger` are stdout and stderr

Ah, the `Pipe` option was added at a later time. Add an entry for `--log-file` to the configuration and comment it out. Then write a comment describing why it cannot yet be implemented (and link to https://github.com/seanmonstar/pretty-env-logger/issues/52).

forest/src/logger/mod.rs
lemmih: Have you pushed your code? I don't see it being fixed.

forest/src/cli/config.rs
```diff
@@ -5,18 +5,30 @@ use forest_chain_sync::SyncConfig;
...
+
+#[derive(Serialize, Deserialize, Default, PartialEq)]
 #[serde(default)]
 pub struct Config {
     pub client: Client,
     pub rocks_db: forest_db::rocks_config::RocksDbConfig,
     pub network: Libp2pConfig,
     pub sync: SyncConfig,
     pub chain: Arc<ChainConfig>,
+    pub log_config: Option<LogConfig>,
```
lemmih, line 31: What happens if this is not an `Option`?

forest/src/cli/mod.rs
```diff
@@ -4,7 +4,7 @@
 mod auth_cmd;
 mod chain_cmd;
 mod client;
-mod config;
+pub mod config;
```
lemmih, line 7: I think it would be cleaner to re-export `LogConfig` and `LogValue` in the same way that `Config` is re-exported.

forest/src/cli/config.rs
tyshko5: Parse failure

lemmih: Where does this parse failure happen and why does it happen?

forest/src/cli/mod.rs
tyshko5: Re-exported

forest/src/cli/mod.rs
```diff
@@ -175,6 +176,11 @@ pub struct CliOpts {
         possible_values = &["mainnet", "calibnet"],
     )]
     pub chain: String,
+    // having such an option isn't possible since target for env_logger can be either stderr or std out, for now. See https://github.com/seanmonstar/pretty-env-logger/issues/52
```
lemmih, line 179: `env_logger-0.7` can only redirect to stderr or stdout. Version 0.9 can redirect to a file. However, we cannot upgrade to version 0.9 because `pretty_env_logger` depends on version 0.7 and hasn't been updated in quite a while. The linked issue is for releasing a new version of `pretty_env_logger`.

forest/src/cli/mod.rs
```diff
@@ -175,6 +176,11 @@ pub struct CliOpts {
...
+    // #[structopt(
+    //     help = "target file for logs"
```
lemmih, line 181: You can use the help text from the original issue: "Specify a filename into which logging should be appended."

forest/src/cli/config.rs
lemmih: What would happen if you changed it to:

```suggestion
    pub log_config: LogConfig,
```

tyshko5: We will have to write out the internal structure of LogConfig, even when we don't want to add it

forest/src/cli/mod.rs
tyshko5: Fixed

forest/src/cli/mod.rs
tyshko5: Fixed

forest/src/cli/config.rs
lemmih: That sounds like a good thing. The log configuration will still be created from its default value when it's omitted from the configuration file, right?

tyshko5: Ok. I'll delete the Option

lemmih (approved):

forest/src/cli/config.rs
```diff
@@ -5,18 +5,30 @@ use forest_chain_sync::SyncConfig;
...
+
+#[derive(Serialize, Deserialize, Default, PartialEq)]
 #[serde(default)]
 pub struct Config {
     pub client: Client,
     pub rocks_db: forest_db::rocks_config::RocksDbConfig,
     pub network: Libp2pConfig,
     pub sync: SyncConfig,
     pub chain: Arc<ChainConfig>,
+    pub log_config: LogConfig,
```
elmattic, line 31: I would just use `log` and not `log_config`, otherwise it's a bit redundant.
Gossamer use that [too](https://chainsafe.github.io/gossamer/usage/configuration/#configuration).

forest/src/cli/config.rs
```diff
@@ -5,18 +5,30 @@ use forest_chain_sync::SyncConfig;
...
-#[derive(Serialize, Deserialize, PartialEq, Default)]
+#[derive(Serialize, Deserialize, Default, Clone, PartialEq)]
+pub struct LogConfig {
+    pub log_values: HashSet<LogValue>,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
+pub struct LogValue {
+    pub module: String,
+    pub level: String,
```
elmattic, line 20: For one given module, is it possible to have multiple levels? Like `warn` and `error`?

forest/src/cli/config.rs
```diff
@@ -5,18 +5,30 @@ use forest_chain_sync::SyncConfig;
 use forest_libp2p::Libp2pConfig;
 use forest_networks::ChainConfig;
 use serde::{Deserialize, Serialize};
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 
 use super::client::Client;
 
-#[derive(Serialize, Deserialize, PartialEq, Default)]
+#[derive(Serialize, Deserialize, Default, Clone, PartialEq)]
```
elmattic, line 12: We don't need `Clone` here.

forest/src/cli/config.rs
```diff
@@ -5,18 +5,30 @@ use forest_chain_sync::SyncConfig;
...
+pub struct LogConfig {
+    pub log_values: HashSet<LogValue>,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
```
elmattic, line 17: We don't need `Clone` here.

elmattic (changes requested):
Great!
Is there some way in Forest to see the computed log levels for each module?

forest/src/cli/mod.rs
```diff
@@ -175,6 +176,13 @@ pub struct CliOpts {
         possible_values = &["mainnet", "calibnet"],
     )]
     pub chain: String,
+    // env_logger-0.7 can only redirect to stderr or stdout. Version 0.9 can redirect to a file.
+    // However, we cannot upgrade to version 0.9 because pretty_env_logger depends on version 0.7
```
LesnyRumcajs, line 180: We could use `pretty-env-logger` from Github repo not crates.io, or am I missing something?

lemmih: That is an option. I have no strong opinions about whether we do this now or hold off until later when a release is out.

LesnyRumcajs: If we hold off, let's create a separate issue and get rid of the commented out code (and copy the snippet into the issue).

forest/src/cli/config.rs
tyshko5: This isn't needed as upper level will take lower levels by default

elmattic (approved):

lemmih (changes requested):
Something definitely went wrong while resolving conflicts.

lemmih:
@tyshko5 The dictionary is still wrong. Look at the changes to see why.

.config/forest.dic
```diff
@@ -43,7 +43,26 @@ tipset/SM
...
+BLS
+crypto
+stateful
+MDNS
+KV
+Kademlia
+LibP2P
+boolean
+TOML
+verifier
```
lemmih, line 68: This looks wrong.

lemmih (changes requested):

lemmih (approved):

PR #1856, opened 2022-08-25
Daemonize Forest node

### LINKED ISSUE #1801: Add `--detach` flag to daemonize the Forest node.
**Issue summary**

Forest should daemonize after the various services (metrics, rpc, networking) have been setup.

https://crates.io/crates/daemonize-me
https://crates.io/crates/daemonize
https://crates.io/crates/ceviche

elmattic:
**Summary of changes**
- [x] Add code to handle process detaching
- [x] Expose in config daemon settings
- [x] Add shared memory and event for coordination between parent/child

Closes #1801


forest/src/main.rs
```diff
@@ -6,23 +6,80 @@ mod daemon;
...
+        daemon.stderr(file)
+    } else {
+        daemon
+    };
+    let daemon = daemon.work_dir(&config.work_dir);
+    let daemon = if let Some(path) = &config.pid_file {
+        daemon.pid_file(path, Some(false))
+    } else {
+        daemon
+    };
```
lemmih, line 48: Maybe this instead:

```suggestion
    let mut daemon = Daemon::new().umask(config.umask).work_dir(&config.work_dir);
    if let Some(user) = &config.user {
        daemon = daemon.user(User::try_from(user)?)
    }
    if let Some(group) = &config.group {
        daemon = daemon.group(Group::try_from(group)?)
    }
    if let Some(path) = &config.stdout {
        let file = File::create(path).expect("File creation {path} must succeed");
        daemon = daemon.stdout(file)
    }
    if let Some(path) = &config.stderr {
        let file = File::create(path).expect("File creation {path} must succeed");
        daemon = daemon.stderr(file)
    }
    if let Some(path) = &config.pid_file {
        daemon = daemon.pid_file(path, Some(false))
    }
```

elmattic: That's much cleaner! Thanks.

lemmih (changes requested):
Great start. It would be helpful if Forest didn't terminate until the RPC and monitoring services are up-and-running. This can done by setting a post-fork-parent-hook that waits for a signal from the daemon. We have to make sure that the parent doesn't block infinitely if the daemon dies, though.

node/rpc/src/lib.rs
```diff
@@ -181,6 +182,7 @@ where
         .post(rpc_http_handler::<DB, B>);
 
     info!("Ready for RPC connections");
+    set_event();
```
lemmih, line 185: No need to signal the event here. Do it after the RPC had had a chance to bind to the network port in the daemon.

elmattic: Done.

forest/src/daemon.rs
```diff
@@ -280,6 +293,7 @@ pub(super) async fn start(config: Config) {
         let rpc_listen = TcpListener::bind(&config.client.rpc_address)
             .await
             .unwrap_or_else(|_| cli_error_and_die("could not bind to {rpc_address}", 1));
+        set_event();
```
lemmih, line 296: If the RPC is not enable, this will block the parent process forever. If you move this below the `if` statement, it'll always work.

elmattic: Ooops. Fixing this.

forest/src/daemon.rs
```diff
@@ -42,6 +44,17 @@ use forest_fil_cns::composition as cns;
 #[cfg(feature = "forest_deleg_cns")]
 use forest_deleg_cns::composition as cns;
 
+fn set_event() {
```
lemmih, line 47: Maybe rename to `signal_event` and describe why this function is used and what happens if we didn't detach (ie. that the event is still valid but no parent will be listening for the signals, etc.)

forest/src/daemon.rs
```diff
@@ -24,7 +24,9 @@ use fvm_shared::version::NetworkVersion;
 use async_std::{channel::bounded, net::TcpListener, sync::RwLock, task};
 use futures::{select, FutureExt};
 use log::{debug, error, info, trace, warn};
+use raw_sync::events::*;
```
lemmih, line 27: We only use `Event` and `EventState`, right? We can enumerate those.

lemmih (changes requested):

forest/src/daemon.rs
lemmih: Could even name it `unblock_parent_process`.

elmattic: Note if we don't use `--detach` the pointer is still null and we just bail out (and never call `set`).

lemmih: Great. Add that to the documentation.

forest/src/main.rs
```diff
@@ -6,23 +6,111 @@ mod daemon;
...
     // Run forest as a daemon if no other subcommands are used. Otherwise, run the subcommand.
     match opts.to_config() {
         Ok(cfg) => match cmd {
-            Some(command) => subcommand::process(command, cfg).await,
-            None => daemon::start(cfg).await,
+            Some(command) => {
+                task::block_on(subcommand::process(command, cfg));
+            }
+            None => {
+                let _shmem = if opts.detach {
```
lemmih, line 92: Why do you return `shmem`?

forest/src/main.rs
```diff
@@ -6,23 +6,111 @@ mod daemon;
...
+
+const EVENT_TIMEOUT: Timeout = Timeout::Val(Duration::from_secs(4));
+
+fn create_event() -> Shmem {
+    let shmem = ShmemConf::new()
+        .size(mem::size_of::<Event>())
+        .create()
+        .expect("create must succeed");
+    SHMEM_PTR.store(shmem.as_ptr(), Ordering::Relaxed);
+    info!("Creating event in shared memory");
```
lemmih, line 34: Probably not worth it to tell the user about this.

forest/src/main.rs
elmattic: Otherwise the shared memory segment is dropped just before calling `::process` and forest woes.

elmattic: Ah sorry, that was for the `ShmemConf`. Investigating.

elmattic: Drop method is in `MapData` and will call `munmap`.

lemmih: Hmm, I think we have to be more careful about when `Shmem` is dropped. Maybe we need `lazy_static`. Then we can set the owner to be the parent.

lemmih: Right now it works mostly by accident. We should be explicit about dropping `Shmen` in the parent process, not in the child process. Doing this will mostly likely require a slight refactor.

elmattic: Can't we just create the value on the stack and let RAII take care of dropping the value?

lemmih: Not if we want to drop the reference from the parent. We can decide that it's fine to keep the reference around until the child dies, though.

forest/src/cli/config.rs
```diff
@@ -9,6 +9,32 @@ use std::sync::Arc;
 
 use super::client::Client;
 
+/// Structure that defines daemon configuration when process is detached
+#[derive(Deserialize, Serialize, PartialEq)]
+pub struct DaemonConfig {
```
LesnyRumcajs, line 14: can those be `PathBuf`s or are `String`s imposed by the library?

forest/src/daemon.rs
```diff
@@ -42,6 +43,17 @@ use forest_fil_cns::composition as cns;
...
 use forest_deleg_cns::composition as cns;
 
+// If we did not detach no existing mapping using the current configuration exists and this function is a no-op
+fn unblock_parent_process() {
+    if let Ok(shmem) = super::ipc_shmem_conf().open() {
+        let (event, _) =
+            unsafe { Event::from_existing(shmem.as_ptr()).expect("from_existing must succeed") };
+
+        event.set(EventState::Signaled).expect("set must succeed");
+        info!("Parent unblocked");
```
LesnyRumcajs, line 53: is it maybe a `debug`? If it is an `info` we may need a less cryptic message for those not exposed to systems programming.

forest/src/main.rs
```diff
@@ -6,23 +6,118 @@ mod daemon;
...
+use shared_memory::ShmemConf;
 use structopt::StructOpt;
 
-#[async_std::main]
-async fn main() {
+use std::fs::File;
+use std::process;
+use std::time::Duration;
+
+const EVENT_TIMEOUT: Timeout = Timeout::Val(Duration::from_secs(20));
```
LesnyRumcajs, line 23: It would be nice to have this in config instead. Not a big deal though.

forest/src/main.rs
```diff
@@ -6,23 +6,118 @@ mod daemon;
...
+
+// The parent process and the daemonized child communicate through an Event in
+// shared memory. The identity of the shared memory object is written to a local
+// file named .forest_daemon_ipc. The parent process is responsible for cleaning
```
LesnyRumcajs, line 27: So by default, it will create this file in the CWD? If so, I'd say we should change it to some temp dir. E.g. something like this would fail, right? `cd / && forest --detach`. Unless it's a common practice do it like this.

forest/src/main.rs
```diff
@@ -6,23 +6,118 @@ mod daemon;
...
+fn build_daemon<'a>(config: &DaemonConfig) -> Result<Daemon<'a>, DaemonError> {
+    let mut daemon = Daemon::new().umask(config.umask).work_dir(&config.work_dir);
+    if let Some(user) = &config.user {
+        daemon = daemon.user(User::try_from(user)?)
+    }
+    if let Some(group) = &config.group {
+        daemon = daemon.group(Group::try_from(group)?)
+    }
+    if let Some(path) = &config.stdout {
+        let file = File::create(path).expect("File creation {path} must succeed");
```
LesnyRumcajs, line 56: Can we `?` here and below as well?

forest/src/main.rs
```diff
@@ -6,23 +6,118 @@ mod daemon;
...
+        drop(shmem); // Delete the local link and the shared memory object.
+        if let Err(e) = ret {
+            cli_error_and_die(
+                format!(
+                    "Error unblocking process. Error was: {e}. Check the log file for details."
+                ),
+                1,
+            );
+        }
+        info!("Forest has been detached and runs in the background.");
```
LesnyRumcajs, line 82: It would be nice to inform the user where the `stdout` and `stderr` are pushed to.

forest/src/cli/config.rs
elmattic: Good idea. No the crate works later with the `File` type. `PathBuf` is much cleaner.

forest/src/daemon.rs
elmattic: Log removed.

forest/src/main.rs
elmattic: I think this should be ok for now.

forest/src/main.rs
elmattic: Yes it was CWD. Using `temp_dir()` now.

forest/src/main.rs
LesnyRumcajs: If you'd use `anyhow::Result` in the signature, wouldn't it automagically help?

forest/src/main.rs
elmattic: Done.

forest/src/cli/config.rs
```diff
@@ -5,10 +5,37 @@ use forest_chain_sync::SyncConfig;
...
+    pub user: Option<String>,
+    pub group: Option<String>,
+    pub umask: u16,
+    pub stdout: PathBuf,
+    pub stderr: PathBuf,
+    pub work_dir: String,
```
LesnyRumcajs, line 21: this could also be `PathBuf`, or not?

forest/src/main.rs
```diff
@@ -6,23 +6,116 @@ mod daemon;
...
+
+// The parent process and the daemonized child communicate through an Event in
+// shared memory. The identity of the shared memory object is written to a local
+// file named .forest_daemon_ipc. The parent process is responsible for cleaning
+// up the local file and the shared memory object.
+fn ipc_shmem_conf() -> ShmemConf {
+    ShmemConf::new()
+        .size(Event::size_of(None))
+        .force_create_flink()
+        .flink(std::env::temp_dir().join("forest_daemon_ipc"))
```
LesnyRumcajs, line 33: this defaults to `/tmp` on Unix, right? So running two detached instances of Forest (e.g. calibnet and mainnet) would fiasco, no? How about using [this nice crate](https://docs.rs/tempdir/latest/tempdir/), it's already used in some of the tests.

forest/src/main.rs
```diff
@@ -6,23 +6,118 @@ mod daemon;
...
-            Some(command) => subcommand::process(command, cfg).await,
-            None => daemon::start(cfg).await,
+            Some(command) => {
+                task::block_on(subcommand::process(command, cfg));
+            }
+            None => {
+                if opts.detach {
+                    create_ipc_lock();
+                    info!(
+                        "Redirecting stdout and stderr to files {:?} and {:?}.",
```
LesnyRumcajs, line 101: how about a regular `Display` and not `Debug`? Not sure how it looks like in `Debug`.

forest/src/cli/config.rs
elmattic: Good catch.

forest/src/main.rs
elmattic: Why not yes! Is there any guarantee that the temp dir is stable (so same after forking)?

forest/src/main.rs
elmattic: We already doing it at other places: Ie: `info!("Recovered libp2p keypair from {:?}", &path);` resulting in `2022-09-05T14:07:40.373Z INFO  forest_libp2p::service > Recovered libp2p keypair from "/Users/guillaume/Library/Application Support/com.ChainSafe.Forest/libp2p/keypair"`

Both `PathBuff` and `Path` don't implement `Display` trait btw.

LesnyRumcajs: Ya, I know they don't implement it, but you can call e.g. `to_string_lossy`. `Debug` should be left for `debug!` :)

forest/src/main.rs
lemmih: We can pick a random temporary file name before the fork.

LesnyRumcajs: the temporary directory exists for the duration of the scope so probably it should be fine. Btw. my bad, I meant this crate https://docs.rs/tempfile/latest/tempfile/, it's more feature-rich.

LesnyRumcajs (approved):
Looks good, nice job! Perhaps let's add this outstanding feature to the README?

forest/src/main.rs
lemmih: Was this resolved? I don't see it in the code.

forest/src/main.rs
```diff
@@ -6,23 +6,129 @@ mod daemon;
...
+        let mut shmem = ipc_shmem_conf().open().expect("open must succeed");
+        shmem.set_owner(true);
+        let (event, _) =
+            unsafe { Event::from_existing(shmem.as_ptr()).expect("from_existing must succeed") };
+        let ret = event.wait(EVENT_TIMEOUT);
+        drop(shmem); // Delete the local link and the shared memory object.
+        if let Err(e) = ret {
+            cli_error_and_die(
+                format!(
+                    "Error unblocking process. Error was: {e}. Check the log file for details."
```
lemmih, line 83: The `e` error is quite useless. It would be better to not include it.

forest/src/main.rs
elmattic: Yes, just after the call creating the lock: `create_ipc_lock`.

forest/src/main.rs
elmattic: Sure.

lemmih (approved):
Looks good to me.

In the future, we will write a troubleshooting document and work on improving the most common error messages. Right now, all of the error messages in Forest are quite difficult to understand.

PR #1858, opened 2022-08-29
Use tokio tasks rather than async-std tasks.

### LINKED ISSUE #1849: Investigate on supporting Tokio Console
**Issue summary**

We shoud investigate on how to use Tokio Console alongside Forest. This kind of tooling would really help us:
- understanding Forest execution flow
- debugging concurency or memory issues
- optimizing our async code

While Tokio Console is not tight to the tokio runtime, as of today only tokio runtime supports it.

Few possible ways of supporting this tool would be (while not trivial):
- Have a native support of tokio in forest (through a feature flag)
- Implement tracing in `async_std` runtime (so the smol runtime)
- Some other shortcut?

https://tokio.rs/blog/2021-12-announcing-tokio-console

lemmih:
**Summary of changes**
- Switch from async-std with tokio compatibility to using tokio directly.
- Enable tokio-console.

This is an experimental PR to investigate how much work it would take to switch to tokio from async-std.

Closes #1849 


blockchain/state_manager/src/chain_rand.rs
```diff
@@ -232,7 +232,9 @@ where
         round: ChainEpoch,
         entropy: &[u8],
     ) -> anyhow::Result<[u8; 32]> {
-        task::block_on(self.get_chain_randomness_v2(&self.blks, pers, round, entropy))
+        Runtime::new()
```
LesnyRumcajs, line 235: just to be sure, is it equivalent in terms of operations? That is to say, did previous implementation instantiated a new runtime as well?

forest/Cargo.toml
```diff
@@ -12,6 +12,7 @@ async-std              = { version = "1.9", features = ["attributes", "tokio1"]
 base64                 = "0.13"
 byte-unit              = "4.0"
 cid                    = { version = "0.8", default-features = false, features = ["std"] }
+console-subscriber     = "*"
```
LesnyRumcajs, line 15: That's a bit optimistic in terms of backwards compatibility, maybe it would be better to pin it to the current minor, i.e. `console-subscriber = "0.1.7"`?

forest/src/daemon.rs
```diff
@@ -344,20 +347,27 @@ pub(super) async fn start(config: Config) {
 
     // Block until ctrl-c is hit
     ctrlc_oneshot.await.unwrap();
+    info!("Line {}, refs: {}", line!(), db_weak_ref.strong_count());
```
LesnyRumcajs, line 350: leftover?

forest/src/daemon.rs
```diff
@@ -344,20 +347,27 @@ pub(super) async fn start(config: Config) {
...
 
     let keystore_write = task::spawn(async move {
         keystore.read().await.flush().unwrap();
     });
+    info!("Line {}, refs: {}", line!(), db_weak_ref.strong_count());
```
LesnyRumcajs, line 355: leftover?

forest/src/daemon.rs
```diff
@@ -344,20 +347,27 @@ pub(super) async fn start(config: Config) {
...
 
     // Cancel all async services
-    prometheus_server_task.cancel().await;
-    head_changes_task.cancel().await;
-    republish_task.cancel().await;
+    prometheus_server_task.abort();
```
LesnyRumcajs, line 358: maybe let's put a lambda `cancel_task if it's feasible? Seems like some code duplication that we can get rid of

forest/src/main.rs
```diff
@@ -9,7 +9,8 @@ mod subcommand;
 use cli::{cli_error_and_die, Cli};
 use structopt::StructOpt;
 
-#[async_std::main]
+// #[async_std::main]
```
LesnyRumcajs, line 12:
```suggestion
```

node/rpc-api/src/data_types.rs
```diff
@@ -38,6 +37,8 @@ use fvm_shared::clock::ChainEpoch;
 use fvm_shared::message::Message;
 use jsonrpc_v2::{MapRouter as JsonRpcMapRouter, Server as JsonRpcServer};
 use serde::{Deserialize, Serialize};
+use std::sync::Arc;
```
LesnyRumcajs, line 40: double-checking, is this correct for sure? I mean, I assume `async_std` did come up with its own `Arc` for a reason

LesnyRumcajs (commented):
What about other occurences of `async_std`, also `async_std` tasks?
```
❯ rg "async_std::task" -l
utils/genesis/src/lib.rs
utils/net_utils/src/download.rs
node/forest_libp2p/src/discovery.rs
node/forest_libp2p/src/service.rs
node/rpc/src/rpc_ws_handler.rs
blockchain/chain_sync/src/chain_muxer.rs
blockchain/message_pool/src/msgpool/mod.rs
blockchain/message_pool/src/msgpool/selection.rs
```

blockchain/state_manager/src/chain_rand.rs
lemmih: They're equivalent. `task::block_on` creates a new Runtime and calls `rt.block_on`.

forest/Cargo.toml
lemmih: Yup.

node/rpc-api/src/data_types.rs
lemmih: They re-export `Arc` from `std::sync` for convenience.

forest/src/daemon.rs
lemmih: Done.

forest/src/daemon.rs
lemmih: Yes.

PR #1865, opened 2022-09-01
Support multiple mining tasks started by the Proposer

aakoshh:
**Summary of changes**
- `Proposer::run` was replaced by `Proposer::spawn` which returns a `Vec<JoinHandle<()>>` to allow it to run multiple background tasks.

This is part of https://github.com/aakoshh/forest/pull/4 , cherry picked so we can open a separate PR upstream, changing existing integration patterns.

### DIFF
--- a/forest/src/daemon.rs
+++ b/forest/src/daemon.rs
@@ -283,8 +283,9 @@ pub(super) async fn start(config: Config) {
     );
 
     // Initialize Consensus. Mining may or may not happen, depending on type.
-    let (consensus, mining_task) =
-        cns::consensus(&state_manager, &keystore, &mpool, submitter).await;
+    let (consensus, mining_tasks) = cns::consensus(&state_manager, &keystore, &mpool, submitter)
+        .await
+        .unwrap();
 
     // Initialize ChainMuxer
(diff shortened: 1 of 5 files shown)


lemmih (approved):

forest/src/daemon.rs
```diff
@@ -283,8 +283,9 @@ pub(super) async fn start(config: Config) {
     );
 
     // Initialize Consensus. Mining may or may not happen, depending on type.
-    let (consensus, mining_task) =
-        cns::consensus(&state_manager, &keystore, &mpool, submitter).await;
+    let (consensus, mining_tasks) = cns::consensus(&state_manager, &keystore, &mpool, submitter)
+        .await
+        .unwrap();
```
lemmih, line 288: We should eventually add more consistent error handling to the daemon.

aakoshh: Yeah, I was just following the pattern; but at least it all culminates in this file, so it's easy to turn it into a nice error message instead.

LesnyRumcajs (approved):
`rock-solid` as always :)

PR #1878, opened 2022-09-06
Update dependencies in Cargo.lock 

### LINKED ISSUE #1863: Update dependencies in `Cargo.lock`
**Issue summary**

Some of Forest's dependencies have apparently not followed semver and running `cargo update` is causing compilation errors. These errors should be investigated and fixed. Some upper version bounds on dependencies might be required.

tyshko5:
**Summary of changes**
- 

Closes #1863


lemmih:
Changing the `fvm_shared` dependency to `0.8` solves the root problem?

utils/auth/Cargo.toml
```diff
@@ -6,13 +6,14 @@ edition = "2021"
 
 [dependencies]
 # Public
-fvm_shared   = { version = "0.8", default-features = false }
+fvm_shared   = { version = "=0.8.0", default-features = false }
```
lemmih, line 9: Is this still necessary or can we go back to `"0.8"`?

utils/auth/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rand         = "0.8"
 serde        = { version = "1.0", default-features = false, features = ["derive"] }
 thiserror    = "1.0"
 # Internal
+anyhow                = "=1.0.59"
```
lemmih, line 16: Is this still required?

tyshko5: Yes. `anyhow` is the reason of our. problems

lemmih: Add the reason why `anyhow` is required as a comment in the `Cargo.toml` file. Right now the dependency looks unused and could easily be deleted.

lemmih (changes requested):

lemmih (changes requested):
This appears to be the root cause: https://github.com/dtolnay/anyhow/issues/230

We can get around this by bumping our toolchain version to `nightly-2022-09-08`. Try this rust version without the restrain on `anyhow`.

lemmih:
Looking good. We do, however, have a minimum dependency on `anyhow = "1.0.64"`. You should add that requirement to `forest/Cargo.toml`.

blockchain/chain_sync/src/chain_muxer.rs
```diff
@@ -165,7 +165,7 @@ where
     M: Provider + Sync + Send + 'static,
     C: Consensus,
 {
-    #[allow(clippy::too_many_arguments)]
+    #[allow(clippy::too_many_arguments, clippy::result_large_err)]
```
lemmih, line 168: Would it be too awkward to box the errors?

tyshko5: This doesn't work

lemmih: Please explain why.

tyshko5: it worked out

LesnyRumcajs: If it worked out then do we need this suppression?

tyshko5: It worked out partly, meaning if we were ok with `map_err`. But David wanted to get of them completely (AFAI understood) and this was not possible (as it propagated all over the codebase). So for now we decided to leave the attribute.

lemmih: I'm 90% sure we can get it to work without having to change too much code. I'll open a new issue for that.

lemmih (approved):

blockchain/chain_sync/src/chain_muxer.rs
LesnyRumcajs: Sounds good

LesnyRumcajs (approved):

PR #1879, opened 2022-09-06
Fix bug in deserialize implementation for `SyncState`

lemmih:
**Summary of changes**
- The deserialize didn't handle the case when the `base` or `target` fields were null. The prevented many RPC calls until after a snapshot had been loaded.

Closes 


LesnyRumcajs (approved):
Should we take the opportunity and write a test that this is always correct? I know there's a separate task, but putting it here would end up being a nice example to follow.

lemmih:
> Should we take the opportunity and write a test that this is always correct? I know there's a separate task, but putting it here would end up being a nice example to follow.

Good idea.

blockchain/blocks/Cargo.toml
```diff
@@ -26,6 +26,7 @@ lazy_static       = "1.4.0"
 log               = "0.4.8"
 num-bigint        = "0.4"
 once_cell         = "1.5.2"
+quickcheck        = { version = "1" }
```
LesnyRumcajs, line 29: This is a test-only dependency so it should go into dev dependencies and the `Arbitrary` implementation should be hidden under `#[cfg(test)]`. Unless it's not possible for some reason?

LesnyRumcajs (changes requested):

blockchain/blocks/Cargo.toml
lemmih: The `Arbitrary` implementations are spread out over several crates, unfortunately. I tried making the dependencies optional, thinking that they would be enabled when running tests but that doesn't work. We could make the dependencies optional but enabled by default but I don't think it is worth it (at least for now).

LesnyRumcajs (approved):

PR #1885, opened 2022-09-07
Enable dependabot

lemmih:
**Summary of changes**
- Enable dependabot for Rust dependencies.

Closes 


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
LesnyRumcajs, line 1: Probably we could leave out only two last lines in the template comment, or even get rid of those as well. Not a big deal though.

LesnyRumcajs (approved):

tyshko5 (approved):

PR #1888, opened 2022-09-08
suggest GH Discussions instead of Discord

LesnyRumcajs:
**Summary of changes**
- Suggest to use GH Discussions and not Discord where we are not really active. Perhaps we should get rid of the Discord badge as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ _Work in progress_.
 
 ## Contributing
 - Check out our contribution guidelines: [CONTRIBUTING.md](documentation/developer_documentation/CONTRIBUTING.md)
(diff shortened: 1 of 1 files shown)


lemmih (approved):
I'm okay with keeping the discord badge even if it isn't the fastest way of contacting us.

LesnyRumcajs:
@lemmih how about moving the hint to ask questions on Discussions to the top before `Run with Docker`?

lemmih:
> @lemmih how about moving the hint to ask questions on Discussions to the top before `Run with Docker`?

Sounds good.

elmattic (approved):

PR #1916, opened 2022-09-14
Automatically download snapshots

### LINKED ISSUE #1847: Offer to download snapshots if Forest is run with an old or missing database
**Issue summary**

Forest will not even attempt to sync to the FileCoin network if the database is too old or missing altogether. We only support network version >=16 and Forest knows attempts to sync from an earlier version will fail:
https://github.com/ChainSafe/forest/blob/57599f7cf3051d92dec05107691b75afb5ccd944/forest/src/daemon.rs#L190-L208

To bypass this issue, users have to download a recent snapshot and use it to initialize the Forest database. To make Forest easier to use, it should ask the user for permission to automatically download and import a snapshot when necessary.

**Task summary**

- [ ] Use methods from #1846 to fetch the appropriate snapshots.
- [ ] Implement `--assume-yes` and `--assume-no`.

**Acceptance Criteria**

- [ ] When stdin is a tty and the database is not recent enough, ask to download the appropriate snapshot (either calibnet or mainnet).
- [ ] When stdin is not a tty, assume the user said no.
- [ ] When stdin is not a tty and `--assume-yes` is given, assume the user said yes.

Blocked by #1846

connormullett:
**Summary of changes**
- Ask the user when applicable if they would like forest to download a snapshot from a trusted source using RPC methods when running without --detach
- Add --download-snapshot flag to download a snapshot prior to syncing for use when stdin is not available (such as running fresh node with --detach)

Closes #1847 


LesnyRumcajs (changes requested):

forest/src/daemon.rs
```diff
@@ -398,6 +393,37 @@ pub(super) async fn start(config: Config, detached: bool) {
...
+
+    let snapshot_path = config
+        .client
+        .data_dir
+        .join("snapshots")
+        .join(config.chain.name.clone());
+
+    match snapshot_fetch(&snapshot_path, config).await {
+        Ok(snapshot_path) => {
+            config.client.snapshot_path = Some(snapshot_path.display().to_string())
```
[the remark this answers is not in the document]
connormullett: Sure, this'll make the display easier

forest/src/daemon.rs
```diff
@@ -335,20 +338,12 @@ pub(super) async fn start(config: Config, detached: bool) {
 
     // Terminate if no snapshot is provided or DB isn't recent enough
     match chain_store.heaviest_tipset().await {
-        None => {
-            cli_error_and_die(
-                "Forest cannot sync without a snapshot. Download a snapshot from a trusted source and import with --import-snapshot=[file]",
-                1,
-            );
-        }
+        None => prompt_and_fetch_snapshot(&mut config).await,
```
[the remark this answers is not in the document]
connormullett: I like this idea, but then we run into the issue of not having RPC up while fetching the snapshot. Just like we had before when downloading headers. I don't think this is an issue that needs to be addressed but it will only apply when a snapshot is being downloaded. Looking at the code, it is possible to just stick it right before the p2p service starts up. I'll make the change but open to discussion about this

forest/src/cli/mod.rs
```diff
@@ -186,6 +186,9 @@ pub struct CliOpts {
     /// Daemonize Forest process
     #[structopt(long)]
     pub detach: bool,
+    /// Assume yes to any prompts requiring yes/no inputs after Forest starts
+    #[structopt(long, short)]
+    pub yes: bool,
```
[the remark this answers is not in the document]
connormullett: That makes sense, considering we are only using this flag for downloading snapshots, this change is ok

forest/src/daemon.rs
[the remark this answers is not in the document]
LesnyRumcajs: Either this or we would need some locking on `stdout` or pausing other tasks. 

A simple approach I can think of is having all interactive prompts before all other tasks - I don't expect forest to have RPC server running while the application is in babysitter mode. But we need to ensure this is the case across the entire codebase.

Thoughts? @lemmih @elmattic @tyshko5 @jdjaustin @connormullett ?

forest/Cargo.toml
```diff
@@ -64,6 +64,7 @@ once_cell              = "1.14"
 pbr                    = "1.0.4"
 pretty_env_logger      = "0.4"
 prometheus             = { version = "0.13", features = ["process"] }
+question               = "0.2.2"
```
LesnyRumcajs, line 67: Sorry for not spotting it earlier. This seems to be a pretty niche crate with not many downloads that hasn't been active since 2018. In general, we should rely on more established libraries. Was there a particular use-case that was missing from others? Otherwise we could use e.g. [dialoguer](https://crates.io/crates/dialoguer) which has almost 100x more downloads and the usage seems pretty similar. What do you think?

forest/src/cli/mod.rs
```diff
@@ -186,6 +186,9 @@ pub struct CliOpts {
     /// Daemonize Forest process
     #[structopt(long)]
     pub detach: bool,
+    /// Automatically download the latest snapshot from a trusted source
+    #[structopt(long, short)]
```
LesnyRumcajs, line 190: I'd omit `short` here. `-d` seems more suitable for `detach`.

LesnyRumcajs (changes requested):
@connormullett Forest crashes after downloading the snapshot, the culprit most likely being [here](https://github.com/ChainSafe/forest/blob/connor/automatic-snapshot-download/forest/src/daemon.rs#L428-L432)

```
❯ target/release/forest --chain calibnet --encrypt-keystore false --target-peer-count 50
 2022-09-28T07:25:21.283Z INFO  forest::daemon > Starting Forest daemon, version 0.2.2+git.a2903b91
 2022-09-28T07:25:21.283Z INFO  forest_libp2p::service > Recovered libp2p keypair from "/home/rumcajs-work/.local/share/forest/libp2p/keypair"
 2022-09-28T07:25:21.283Z INFO  forest::daemon         > PeerId: 12D3KooWN9wV8DVwHXAFocGFJTpf7BwpTGkEE4nrtdJkBsf1f34X
 2022-09-28T07:25:21.283Z WARN  forest::daemon         > Warning: Keystore encryption disabled!
 2022-09-28T07:25:21.284Z INFO  forest::daemon         > Prometheus server started at 127.0.0.1:6116
 2022-09-28T07:25:21.284Z INFO  forest::daemon         > Admin token: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.8IqfOBc_l9awxLRg7K6fzePe6jhkpalfrpxKDUqvljw
 2022-09-28T07:25:21.311Z INFO  forest_genesis         > Initialized genesis: BlockHeader: Cid(bafy2bzacecz3trtejxtzix4f4eebs7dekm6snfsmvffiqz2rfx7iwgsgtieq4)
 2022-09-28T07:25:21.311Z INFO  forest::daemon         > Using network :: calibrationnet
Forest needs a snapshot to sync with the network. Would you like to download one now? (y/N) y
 2022-09-28T07:25:22.711Z INFO  forest::cli::snapshot_fetch > Snapshot will be downloaded to /home/rumcajs-work/.local/share/forest/snapshots/calibnet/forest_snapshot_calibnet_2022-09-28_height_1342143.car (4.75 GiB)
Downloading snapshot 810.47 MB / 4.75 GB [======================>-------------------------------------Downloading snapshot 4.75 GB / 4.75 GB [========================================] 100.00 % 37.93 MB/s
Finished downloading the snapshot. 2022-09-28T07:27:31.178Z INFO  forest::cli::snapshot_fetch > Snapshot checksum is 570442034ee49359a420b9786b8c89073f58ff812e7a708724785ea093539d9c. Validation is not yet available.
 2022-09-28T07:27:31.180Z INFO  forest_genesis              > Importing chain from snapshot at: /home/rumcajs-work/.local/share/forest/snapshots/calibnet/forest_snapshot_calibnet_2022-09-28_height_1342143.car
 2022-09-28T07:27:31.180Z INFO  forest_chain_sync::chain_muxer > Evaluating network head...
 2022-09-28T07:27:31.180Z INFO  forest::daemon                 > JSON-RPC endpoint started at 127.0.0.1:1234
 2022-09-28T07:27:31.181Z INFO  forest_rpc                     > Ready for RPC connections
 2022-09-28T07:27:31.181Z INFO  forest_genesis                 > Reading file...
Scanning blockchain 1320283 / 1342143 [==========================================================================================================================================>--] 98.37 % 15506.66/s 1s  2022-09-28T07:29:19.825Z INFO  forest_genesis                 > Validating imported chain from height: 0
 2022-09-28T07:30:23.335Z INFO  forest_state_manager           > Computing state (height: 0, ts=[Cid(bafy2bzacecz3trtejxtzix4f4eebs7dekm6snfsmvffiqz2rfx7iwgsgtieq4)])
 2022-09-28T07:30:23.335Z INFO  forest_state_manager           > Computing state (height: 1, ts=[Cid(bafy2bzacebbzkct2j2kssnnunxxblza3jbiot3sokyslltgk7ovdh2osz52sy)])
thread 'main' panicked at 'Failed miserably while importing chain from snapshot: Invalid tipset: no msg root with cid bafy2bzacebqdl36mjrld2vkvbu56gksjyanpcteskuah6ehom6fnxvfwrbm5y', forest/src/daemon.rs:441:10
```

LesnyRumcajs:
Another thing, as user, I'd expect `--download-snapshot` to download one regardless of my current DB status, e.g. if I'd like a fresh one.

forest/src/cli/mod.rs
connormullett: Yup, that was an oversight. I'll change it so they don't clash

forest/Cargo.toml
connormullett: I like dialoguer at a quick glance. The color prompts can also help with UX and seems much better than what I was using before. Going to swap it out

connormullett:
> Another thing, as user, I'd expect `--download-snapshot` to download one regardless of my current DB status, e.g. if I'd like a fresh one.

What would you recommend as a new name for a flag? I'm currently looking into the snapshot crash. Don't have anything off the top of my head for a name for the flag

connormullett:
Can't reproduce the crash on my machine. 

image

LesnyRumcajs:
> Can't reproduce the crash on my machine.
 
@connormullett Have you tried cleaning the db, i.e. `rm -rf ~/.local/share/forest/calibnet/`?

connormullett:
> @connormullett Have you tried cleaning the db, i.e. `rm -rf ~/.local/share/forest/calibnet/`?

I have. I run before every run testing this feature. I'll try running it longer. But it got past the validation step. I've also just merged main. Maybe I missed a fix. Continuing to check

LesnyRumcajs:
@connormullett what does this config of yours contain? Have you tried without it, with defaults?

forest/Cargo.toml
LesnyRumcajs: they are nice and shiny!

connormullett:
> what does this config of yours contain? Have you tried without it, with defaults?

It just contains the encrypt-keystore flag

LesnyRumcajs:
@connormullett do you mind running it longer? In my case it's still failing, just after those first two `computing state` messages.
```
forest on  connor/automatic-snapshot-download [$?] via 🦀 v1.66.0-nightly took 4m53s
❯ target/release/forest --chain calibnet --encrypt-keystore false
 2022-09-29T19:51:08.863Z INFO  forest::daemon > Starting Forest daemon, version 0.2.2+git.27da2815
 2022-09-29T19:51:08.863Z INFO  forest_libp2p::service > Recovered libp2p keypair from "/home/rumcajs-work/.local/share/forest/libp2p/keypair"
 2022-09-29T19:51:08.864Z INFO  forest::daemon         > PeerId: 12D3KooWN9wV8DVwHXAFocGFJTpf7BwpTGkEE4nrtdJkBsf1f34X
 2022-09-29T19:51:08.864Z WARN  forest::daemon         > Warning: Keystore encryption disabled!
 2022-09-29T19:51:08.864Z INFO  forest::daemon         > Prometheus server started at 127.0.0.1:6116
 2022-09-29T19:51:08.864Z INFO  forest::daemon         > Admin token: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.8IqfOBc_l9awxLRg7K6fzePe6jhkpalfrpxKDUqvljw
 2022-09-29T19:51:08.882Z WARN  forest_chain::store::chain_store > No previous chain state found
 2022-09-29T19:51:08.887Z INFO  forest_genesis                   > Initialized genesis: BlockHeader: Cid(bafy2bzacecz3trtejxtzix4f4eebs7dekm6snfsmvffiqz2rfx7iwgsgtieq4)
 2022-09-29T19:51:08.887Z INFO  forest::daemon                   > Using network :: calibrationnet
✔ Forest needs a snapshot to sync with the network. Would you like to download one now? · yes
 2022-09-29T19:51:11.203Z INFO  forest::cli::snapshot_fetch      > Snapshot will be downloaded to /home/rumcajs-work/.local/share/forest/snapshots/calibnet/forest_snapshot_calibnet_2022-09-29_height_1345031.car (4.76 GiB)
Downloading snapshot 4.76 GB / 4.76 GB [==================================================================================================================================================================================================================================================================] 100.00 % 38.45 MB/s
Finished downloading the snapshot. 2022-09-29T19:53:18.051Z INFO  forest::cli::snapshot_fetch      > Snapshot checksum is 4c34c15817acbad45a6b9450202209fc57fe671a0232fd39951b2cc984425dcd. Validation is not yet available.
 2022-09-29T19:53:18.053Z INFO  forest_genesis                   > Importing chain from snapshot at: /home/rumcajs-work/.local/share/forest/snapshots/calibnet/forest_snapshot_calibnet_2022-09-29_height_1345031.car
 2022-09-29T19:53:18.054Z INFO  forest_chain_sync::chain_muxer   > Evaluating network head...
 2022-09-29T19:53:18.054Z INFO  forest::daemon                   > JSON-RPC endpoint started at 127.0.0.1:1234
 2022-09-29T19:53:18.054Z INFO  forest_rpc                       > Ready for RPC connections
 2022-09-29T19:53:18.056Z INFO  forest_genesis                   > Reading file...
Scanning blockchain 1054871 / 1345031 [========================================================================================================================================================================================================>-------------------------------------------------------] 78.43 % 14129.39/s 21s Scanning blockchain 1341271 / 1345031 [=================================================================================================================================================================================================================================================================] 99.72 % 15387.05/s 0s  2022-09-29T19:55:07.935Z INFO  forest_genesis                   > Validating imported chain from height: 0
 2022-09-29T19:56:08.518Z INFO  forest_state_manager             > Computing state (height: 0, ts=[Cid(bafy2bzacecz3trtejxtzix4f4eebs7dekm6snfsmvffiqz2rfx7iwgsgtieq4)])
 2022-09-29T19:56:08.519Z INFO  forest_state_manager             > Computing state (height: 1, ts=[Cid(bafy2bzacebbzkct2j2kssnnunxxblza3jbiot3sokyslltgk7ovdh2osz52sy)])
thread 'main' panicked at 'Failed miserably while importing chain from snapshot: Invalid tipset: no msg root with cid bafy2bzacebqdl36mjrld2vkvbu56gksjyanpcteskuah6ehom6fnxvfwrbm5y', forest/src/daemon.rs:440:10
```

connormullett:
Yup, that's the case. I'm getting it now

forest/src/daemon.rs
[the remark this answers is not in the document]
elmattic: @LesnyRumcajs I'm looking at it now.

forest/src/cli/chain_cmd.rs
```diff
@@ -155,7 +155,7 @@ impl ChainCommands {
                         .join("snapshots")
                         .join(config.chain.name.clone())
                 });
-                match snapshot_fetch(&snapshot_dir, config).await {
+                match snapshot_fetch(&snapshot_dir, &config).await {
                     Ok(out) => println!("Snapshot successfully downloaded at {}", out.display()),
                     Err(e) => cli_error_and_die(format!("Failed fetchning the snapshot: {e}"), 1),
```
elmattic, line 160: Small typo on `fetching`.

forest/src/daemon.rs
```diff
@@ -223,6 +225,18 @@ pub(super) async fn start(config: Config, detached: bool) {
         config
     };
 
+    // Terminate if no snapshot is provided or DB isn't recent enough
```
elmattic, line 228: There is already a lot of complexity in the `daemon.rs` file. We can do all this auto download snapshot logic in `main.rs`. We don't need any service of the daemon for this (RPC, libp2p, etc).

So basically move all the logic before testing the `detach` flag:

ie
```
            match cmd {
                Some(command) => {
                    task::block_on(subcommand::process(command, cfg));
                }
                None => {
                    task::block_on(prompt_and_fetch_snapshot(&mut cfg));
                    if opts.detach {
                        ...
```

elmattic (changes requested):

forest/src/daemon.rs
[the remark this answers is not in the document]
elmattic: Exactly Hubert, all interactive prompts before the daemon starts (in detached mode or not).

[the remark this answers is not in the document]
elmattic: @connormullett We don't need forest daemon RPC service for running the `snapshot_fetch` util (he internaly makes use of `reqwest` or `s3` crates).

elmattic:
What about implementing `--assume-yes`/`--assume-no` for case when stdin is not a tty, like proposed by David?

forest/src/cli/chain_cmd.rs
connormullett: oh thanks lol

connormullett:
> What about implementing `--assume-yes`/`--assume-no` for case when stdin is not a tty, like proposed by David?

Hubert suggested we rename the flag to `--download-snapshot` which makes sense from a user perspective. They may not want to assume-yes for everything

elmattic:
I'm also have this which seem weird:

```
✔ Forest needs a snapshot to sync with the network. Would you like to download one now? · yes
 2022-09-30T14:50:55.469Z INFO  forest::cli::snapshot_fetch > Snapshot will be downloaded to /Users/guillaume/Library/Application Support/com.ChainSafe.Forest/snapshots/calibnet/forest_snapshot_calibnet_2022-09-30_height_1347965.car (4.78 GiB)
Downloading snapshot 4.78 GB / 4.78 GB [==============================================================================================================================================================================================] 100.00 % 42.73 MB/s 
Finished downloading the snapshot. 2022-09-30T14:52:49.927Z INFO  forest::cli::snapshot_fetch > Snapshot checksum is ae8cd7949833c1029e90db2e1326a7ca005ee8a6582e5aba135788d76f394dab. Validation is not yet available.
 2022-09-30T14:52:49.932Z INFO  forest_chain_sync::chain_muxer > Evaluating network head...
 2022-09-30T14:52:49.932Z INFO  forest_genesis                 > Importing chain from snapshot at: /Users/guillaume/Library/Application Support/com.ChainSafe.Forest/snapshots/calibnet/forest_snapshot_calibnet_2022-09-30_height_1347965.car
 2022-09-30T14:52:49.932Z INFO  forest::daemon                 > JSON-RPC endpoint started at 127.0.0.1:1234
 2022-09-30T14:52:49.932Z INFO  forest_genesis                 > Reading file...
 2022-09-30T14:52:49.932Z INFO  forest_rpc                     > Ready for RPC connections
Scanning blockchain 1345585 / 1347965 [==============================================================================================================================================================================================] 99.82 % 3376.96/s 1s  2022-09-30T15:00:06.128Z
INFO  forest_genesis                 > Validating imported chain from height: 0
```

elmattic:
> > What about implementing `--assume-yes`/`--assume-no` for case when stdin is not a tty, like proposed by David?
> 
> Hubert suggested we rename the flag to `--download-snapshot` which makes sense from a user perspective. They may not want to assume-yes for everything

Hmm. I find this a bit cumbersome to understand, maybe consider renaming it to `--force-download-snapshot`.

forest/src/daemon.rs
```diff
@@ -398,6 +392,35 @@ pub(super) async fn start(config: Config, detached: bool) {
     info!("Forest finish shutdown");
 }
 
+async fn prompt_and_fetch_snapshot(config: &mut Config) {
```
elmattic, line 395: You can use this crate [atty](https://crates.io/crates/atty) to test if stdin is a tty and if it's the case ask the user for download of a snapshot.

connormullett:
> > > What about implementing `--assume-yes`/`--assume-no` for case when stdin is not a tty, like proposed by David?
> > 
> > 
> > Hubert suggested we rename the flag to `--download-snapshot` which makes sense from a user perspective. They may not want to assume-yes for everything
> 
> Hmm. I find this a bit cumbersome to understand, maybe consider renaming it to `--force-download-snapshot`.

We want this flag in particular to download a snapshot when one isn't given to Forest and the DB is out of date. I don't think that portrays the intended use of the flag and to @LesnyRumcajs's point, the current one doesn't either

forest/src/daemon.rs
connormullett: We do, we need the `chain_store` to check if the DB is out of date, which is one of the criteria for offering to download a snapshot

forest/src/daemon.rs
```diff
@@ -398,6 +392,35 @@ pub(super) async fn start(config: Config, detached: bool) {
...
+                "Forest cannot sync without a snapshot. Download a snapshot from a trusted source and import with --import-snapshot=[file]",
+                1
+            );
+        }
+    }
+
+    let snapshot_path = config
+        .client
+        .data_dir
+        .join("snapshots")
```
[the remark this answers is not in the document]
connormullett: I think this is fine as it keeps it with the rest of the default locations. If we think about users that install Forest as a binary on their machine, they may not have that one folder that contains everything related to Forest. We also have a log in place to tell user's where the snapshot is being downloaded to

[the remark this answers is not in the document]
elmattic: I'm good with this, and the future `clean` forest subcommand should help the user to take care of this.

lemmih (changes requested):
Please update the PR description to match what has been implemented.

forest/shared/src/cli/mod.rs
```diff
@@ -175,6 +178,17 @@ impl CliOpts {
...
+/// Provided number cannot be negative, otherwise the function will panic.
+pub fn to_size_string(input: &BigInt) -> anyhow::Result<String> {
+    let bytes = u128::try_from(input)
+        .map_err(|e| anyhow::anyhow!("error parsing the input {}: {}", input, e))?;
+
+    Ok(Byte::from_bytes(bytes)
+        .get_appropriate_unit(true)
+        .to_string())
+}
+
```
LesnyRumcajs, line 191: If you move a function around, make sure you move also its unit tests so they stay in the same module. These tests should also be moved https://github.com/ChainSafe/forest/blob/9d6180bb7b9faf14cd4fa9fed88b71b81ae1e034/forest/cli/src/cli/mod.rs#L216-L247

forest/daemon/src/daemon.rs
```diff
@@ -398,18 +391,51 @@ pub(super) async fn start(config: Config, detached: bool) {
...
 
+async fn prompt_and_fetch_snapshot(config: &mut Config) {
+    if !config.client.download_snapshot {
+        if let false = Confirm::with_theme(&ColorfulTheme::default())
+            .with_prompt(
+                "Forest needs a snapshot to sync with the network. Would you like to download one now?",
+            )
+            .default(false)
+            .interact()
+            .expect("infallible. yes or no with default")
```
lemmih, line 402: This is not infallible. IO errors can happen and they're unrelated to the `default` value.

forest/daemon/src/daemon.rs
```diff
@@ -398,18 +391,51 @@ pub(super) async fn start(config: Config, detached: bool) {
     info!("Forest finish shutdown");
 }
 
+async fn prompt_and_fetch_snapshot(config: &mut Config) {
+    if !config.client.download_snapshot {
+        if let false = Confirm::with_theme(&ColorfulTheme::default())
```
LesnyRumcajs, line 396: isn't `if let false = fn()` just `if !fn()`?

forest/daemon/src/daemon.rs
```diff
@@ -398,18 +391,51 @@ pub(super) async fn start(config: Config, detached: bool) {
...
+                1
+            );
+        }
+    }
+
+    let snapshot_path = config
+        .client
+        .data_dir
+        .join("snapshots")
+        .join(config.chain.name.clone());
```
lemmih, line 415: Let's merge this with the identical code in `chain_cmd.rs`.

forest/shared/src/cli/client.rs
```diff
@@ -32,6 +32,7 @@ pub struct Client {
     pub metrics_address: SocketAddr,
     /// RPC bind, e.g. 127.0.0.1:1234
     pub rpc_address: SocketAddr,
+    pub download_snapshot: bool,
```
lemmih, line 35: This should also be a cli option rather than just a configuration file option.

forest/daemon/src/daemon.rs
```diff
@@ -398,18 +391,51 @@ pub(super) async fn start(config: Config, detached: bool) {
...
+    };
+}
+
 async fn sync_from_snapshot(config: &Config, state_manager: &Arc<StateManager<RocksDb>>) {
     if let Some(path) = &config.client.snapshot_path {
         let stopwatch = time::Instant::now();
         let validate_height = if config.client.snapshot {
             config.client.snapshot_height
         } else {
+            info!("SNAPSHOT_HEIGHT WAS 0");
```
lemmih, line 432: This looks like a left-over debug statement.

forest/shared/src/cli/client.rs
lemmih: The behavior of this option should also be documented.

lemmih (changes requested):
This doesn't seem to work with `--detach`. For me, it just times out rather than the expected behavior of an immediate error.

forest/shared/src/cli/client.rs
LesnyRumcajs, line 35: It would be nice to expose it. I don't see it available from CLI, except for the configuration file. Or am I missing something?
```
❯ target/release/forest --help
forest-daemon 0.4.1+git.9d6180bb
ChainSafe Systems <info@chainsafe.io>
CLI options

USAGE:
    forest [FLAGS] [OPTIONS] [cmd]

FLAGS:
        --detach               Daemonize Forest process
        --halt-after-import    Halt with exit code 0 after successfully importing a snapshot
    -h, --help                 Prints help information
        --skip-load            Skips loading CAR file and uses header to index chain. Assumes a pre-loaded database
    -V, --version              Prints version information

OPTIONS:
        --chain <chain>
            Choose network chain to sync to [default: mainnet]  [possible values: mainnet, calibnet]

    -c, --config <config>                            A TOML file containing relevant configurations
        --encrypt-keystore <encrypt-keystore>        Encrypt the key-store (default: true)
    -g, --genesis <genesis>                          The genesis CAR file
        --height <height>                            Validate snapshot at given EPOCH
        --import-chain <import-chain>                Import a chain from a local CAR file or URL
        --import-snapshot <import-snapshot>          Import a snapshot from a local CAR file or URL
    -k, --kademlia <kademlia>                        Allow Kademlia (default: true)
        --mdns <mdns>                                Allow MDNS (default: false)
        --metrics-address <metrics-address>
            Address used for metrics collection server. By defaults binds on localhost on port 6116

        --req-window <req-window>                    Number of tipsets requested over chain exchange (default is 200)
    -r, --rpc <rpc>                                  Allow RPC to be active or not (default: true)
        --rpc-address <rpc-address>                  Address used for RPC. By defaults binds on localhost on port 1234
        --target-peer-count <target-peer-count>      Amount of Peers we want to be connected to (default is 75)
        --tipset-sample-size <tipset-sample-size>
            Number of tipsets to include in the sample that determines what the network head is

    -t, --token <token>                              Client JWT token to use for JSON-RPC authentication

ARGS:
    <cmd>
```

LesnyRumcajs (changes requested):

LesnyRumcajs:
One thing that came to my mind - it'd be great to automatically dispose of the snapshot after it is loaded in the db, so it doesn't pollute the disk space. Probably a flag `--preserve-snapshot` would be helpful then. We can do it as a separate task though.

lemmih:
Please update the PR description to match what has been implemented.

> Ask the user when applicable if they would like forest to download a snapshot from a trusted source using RPC methods when running without --detach

This is not really accurate. The code still asks even in `--detach` mode.

> Add --download-snapshot flag to download a snapshot prior to syncing for use when stdin is not available (such as running fresh node with --detach)

You haven't added a `--download-snapshot` flag. You've only added it to the configuration file.

> Known bug that doesn't hinder functionality: The prompt is drowned out in logs so its a couple lines up when it blocks and will look like Forest is stuck. I can't figure out a way around this

This is a big problem. What parts of Forest are logging while we're asking the user for input? Maybe those background processes could be started after the snapshot has been downloaded.

> Works in --detach when stdin is not a tty (only when --yes/-y is on), and when running normally through the prompt (must enter yes/no)

I don't see any code that deals with ttys nor do I see any code related to `--yes/-y`.

lemmih:
It would be nice to import the snapshot /while/ they're being downloaded. But that can wait for another PR.

forest/daemon/src/daemon.rs
connormullett: it is

forest/shared/src/cli/mod.rs
connormullett: Ah, I didn't see the tests. Good catch

connormullett:
> Please update the PR description to match what has been implemented.

Updated. The bug was fixed when we added dialoguer. Rest of the PR comments are good now. Looking into why --detach isn't working with it

forest/daemon/src/daemon.rs
connormullett: Fixed and change to a match for error handling over expect

forest/daemon/src/daemon.rs
connormullett: Fixed and change to a match for error handling over expect

forest/daemon/src/daemon.rs
connormullett: extracted it out

connormullett:
The bug where the --detach flag being present broke the logic to skip the prompt has been fixed. However the log shows an error in forest.err and you can see the progress bar in forest.out. Not sure what's going on

lemmih:
In the PR description, you talk about these flags: `--yes/-y`. Could you point to me where you've implemented these flags?

You also state `The prompt is drowned out in logs so its a couple lines up when it blocks and will look like Forest is stuck.`. Could you elaborate on this? How can I replicate this problem? Do you have any ideas how it can be fixed?

connormullett:
> In the PR description, you talk about these flags: `--yes/-y`. Could you point to me where you've implemented these flags?
> 
> You also state `The prompt is drowned out in logs so its a couple lines up when it blocks and will look like Forest is stuck.`. Could you elaborate on this? How can I replicate this problem? Do you have any ideas how it can be fixed?

Coulda swore I deleted that because the bug is fixed and --yes is no longer there, been changed to --download-snapshot

lemmih:
Running with `--detach` still fails badly for me. We should never ask for user input if stdin is not a tty.

lemmih:
> > Running with `--detach` still fails badly for me. We should never ask for user input if stdin is not a tty.
> 
> how is it failing. Is the daemon saying there is an error? have you checked forest.out? does it fail after downloading?

It waits for user input which never comes and then it times out after about 4 seconds. Does something different happen for you?

connormullett:
> It waits for user input which never comes and then it times out after about 4 seconds. Does something different happen for you?

yes, it succeeds however it displays an error in the output (when it tries to detach). The error appears but Forest.out logs show that its downloading. Does your config have disable encryption set? that will also ask for user input

lemmih:
> > It waits for user input which never comes and then it times out after about 4 seconds. Does something different happen for you?
> 
> yes, it succeeds however it displays an error in the output (when it tries to detach). The error appears but Forest.out logs show that its downloading. Does your config have disable encryption set? that will also ask for user input

I'm running this command `./target/release/forest --encrypt-keystore false --chain=calibnet --detach`.

It downloads the snapshot for you when run it in detach mode? For me it just blocks waiting for user input.

connormullett:
> 
> I'm running this command `./target/release/forest --encrypt-keystore false --chain=calibnet --detach`.
> 
> It downloads the snapshot for you when run it in detach mode? For me it just blocks waiting for user input.

Ok this is a different issue. Try adding the --download-snapshot flag to that command and you'll see what I'm getting. It should fail because you haven't opted into the flag/config option. It shouldn't ask though, not sure why it is. Looking into it

lemmih:
> > I'm running this command `./target/release/forest --encrypt-keystore false --chain=calibnet --detach`.
> > It downloads the snapshot for you when run it in detach mode? For me it just blocks waiting for user input.
> 
> Ok this is a different issue. Try adding the --download-snapshot flag to that command and you'll see what I'm getting. It should fail because you haven't opted into the flag/config option. It shouldn't ask though, not sure why it is. Looking into it

Ah, interesting. Didn't expect it to fail with the `--download-snapshot` set. We need to update the code such that it doesn't time-out in either case.

When `--download-snapshot` is set, it should detach properly and download the snapshot.

When `--download-snapshot` is not set, it should fail immediately.

connormullett:
I haven't found the reason when download snapshot is set with detach it gives a false error, and you're right I'll be looking into why it's not failing when detach is set without download snapshot set as well

lemmih:
Is this PR ready for review?

connormullett:
> Is this PR ready for review?

yes, fixing conflicts now

LesnyRumcajs:
Todo:
[ ] the prompt is stuck somewhere in logs, should be the last log.
[ ] investigate `--detach` compatibility

lemmih (approved):

elmattic (approved):

forest/daemon/src/daemon.rs
```diff
@@ -181,6 +184,20 @@ pub(super) async fn start(config: Config, detached: bool) {
     .unwrap();
     chain_store.set_genesis(&genesis.blocks()[0]).unwrap();
 
+    // Terminate if no snapshot is provided or DB isn't recent enough
+    let should_fetch_snapshot = match chain_store.heaviest_tipset().await {
+        None => prompt_snapshot_or_die(&config).await,
+        Some(tipset) => {
+            let epoch = tipset.epoch();
+            let nv = config.chain.network_version(epoch);
+            if nv < NetworkVersion::V16 {
```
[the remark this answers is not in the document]
lemmih: Eventually we'll support more versions than just the most recent one. :) It might take some time but we'll get there.

Issue #1861, opened 2022-08-30
Investigate the test in `peer_test.rs`


lemmih:
**Issue summary**

We have a module that contains a single test. This module is not used by the crate and the test is therefore not run. We should investigate what to do about this. Possible options:
- Use the module such that the test is run.
- Move the test to the `tests/` folder. This can only be done if the test doesn't depend on anything private from the crate.
- Remove the test. Maybe the test isn't relevant anymore and should be removed.

File in question: https://github.com/ChainSafe/forest/blob/main/blockchain/chain_sync/src/sync/peer_test.rs

**Other information and links**

PR #1918, opened 2022-09-15
Remove nearly all instances of `block_on`

### LINKED ISSUE #1910: Avoid using `async_std::task::block_on`
**Issue summary**

The `block_on` function blocks the current thread which is undesirable from an async context. We can often refactor the code to avoid calling `block_on`.

Sometimes we use `block_on` in tests. For example:
https://github.com/ChainSafe/forest/blob/20d58d5010b7c59165d47bf20260070a7ce66119/blockchain/blocks/src/header/mod.rs#L521
This can be fixed by adding `#[async_std::test]` here:
https://github.com/ChainSafe/forest/blob/20d58d5010b7c59165d47bf20260070a7ce66119/blockchain/blocks/src/header/mod.rs#L506-L507

Other uses (like in `check_block_messages`) can be fixed by changing the parent function to be `async`.

Other uses (like in `get_beacon_randomness`) cannot be fixed, unfortunately.

lemmih:
**Summary of changes**
- Calling async code from a sync context is a bad practice and can cause a variety of issues.
- Most cases were easy to fix but some can't be updated before we switch entirely to tokio.

Closes #1910 

### DIFF
--- a/blockchain/blocks/src/header/mod.rs
+++ b/blockchain/blocks/src/header/mod.rs
@@ -503,8 +503,9 @@ mod tests {
                 .unwrap())
             .unwrap();
     }
-    #[test]
-    fn beacon_entry_exists() {
+
+    #[async_std::test]
+    async fn beacon_entry_exists() {
         // Setup
         let block_header = BlockHeader::builder()
             .miner_address(Address::new_id(0))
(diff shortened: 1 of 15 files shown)


LesnyRumcajs (approved):

elmattic (approved):
Good refactor!
Btw have you seen the `peer_manager_update` test in [peer_test.rs](https://github.com/ChainSafe/forest/blob/main/blockchain/chain_sync/src/sync/peer_test.rs) test that could be ported too?

Seems to me that this test is never run.

lemmih:
> Good refactor! Btw have you seen the `peer_manager_update` test in [peer_test.rs](https://github.com/ChainSafe/forest/blob/main/blockchain/chain_sync/src/sync/peer_test.rs) test that could be ported too?
> 
> Seems to me that this test is never run.

I've created a separate issue for figuring out what to do with `peer_test.rs`: https://github.com/ChainSafe/forest/issues/1861. I'm not sure if it should be ported or just deleted.

Issue #1919, opened 2022-09-15
Move away from async_std mannerism


lemmih:
**Issue summary**

- Use `std::sync::Arc` instead of `async_std::sync::Arc`. The async_std version is just a re-export.
- Use `std::future::Future` instead of `async_std::future::Future`. The async_std version is just a re-export.
- Use `std::pin::Pin` instead of `async_std::pin::Pin`. The async_std version is just a re-export.
- Use `std::task::{Context, Poll}` instead of `async_std::task::{Context, Poll}`. The async_std version is just a re-export.
- Use `tokio::sync::RwLock` instead of `async_std::sync::RwLock`. We can use the tokio version from an async_std context.
- Use `tokio::timer::Interval` instead of `async_std::stream::Interval`. We can use the tokio version from an async_std context.

**Other information and links**

PR #1927, opened 2022-09-15
less async_std

LesnyRumcajs:
**Summary of changes**
- use less `async_std` in favour of `tokio` and `std`.

I believe we can also touch some more of them, e.g. `async_std::fs::File` but let's address it in another PR as it will clash with https://github.com/ChainSafe/forest/pull/1918 from what I noticed.

Closes https://github.com/ChainSafe/forest/issues/1919


lemmih (approved):
Excellent. Switching to 100% tokio will soon be a trivial matter.

elmattic (approved):
Good refactor!

PR #1933, opened 2022-09-19
Avoid 0s durations when using `tokio::time::interval`

lemmih:
**Summary of changes**
- `async_std::stream::interval` allows 0s durations but `tokio::time::interval` does not. Oh well.

Closes 


lemmih:
Hmm, a better fix would be to use `tokio::time::sleep`. But I'm hesitant to actually change the logic here because I don't quite understand it.

LesnyRumcajs (approved):

tyshko5 (approved):

PR #1941, opened 2022-09-20
Async2tokio interval

LesnyRumcajs:
**Summary of changes**
- Fixed the fiasco coming from a slightly different behaviour between `async_std::Interval` and `tokio::Interval`. While the former yields only after the specified duration, the latter does that as well...except for the first tick, which returns `Poll::Ready` immediately. In our case, where we recreate this `Interval`, it was causing a book definition of an infinite loop, blocking the entire communication layer.

Closes 

https://docs.rs/tokio/latest/tokio/time/struct.Interval.html


lemmih (approved):
Great success!

tyshko5 (approved):

PR #1943, opened 2022-09-21
Re-write batch validator

### LINKED ISSUE #1911: Re-write batch validator to download tipsets in parallel.
**Issue summary**

The `sync_messages_check_state` function downloads and validates tipset messages one at a time. It would be significantly more efficient to download messages in a background thread such that the validator never waits on IO.

Downloading messages is done by the `chain_exchange_messages` function which downloads the messages of a tipset + N of its parents (who have a lower epoch number). Right now, we process tipsets in chronological order (ie from low epoch to high epoch) which makes it difficult to download a batch of tipset messages. The code should be refactored such that batches of tipset messages are downloaded and fed to the validator. For example, if we're validating tipsets from epoch 100 to 200, we would download messages for epoch 110 + 9 parents (epochs 109,108,...,100), validate them in chronological order (100, 101,...,110), and repeat until we're done.

elmattic:
**Summary of changes**
- [x] Fix compact messages iterator direction
- [x] Check impact of `REQUEST_WINDOW` and choose a good default value
- [x] Use background task to avoid waiting on IO
- [x] Measure performance impact of the PR
- [x] Some refactor and simplification
- [x] Understand why some epochs can be missing (see logs below)
- [x] Abort cleanly the background task

**Performance**

Measuring a sync from scratch on calibnet, going from epochs 1,296,060 to 1,329,726 after snapshot import:

| Commit | Elapsed time |
| ------------- | ------------- |
| 20718b83f3106842826d488e22804d64fd11c44b  | 70 minutes |
| a96de02283f2e0ff49bfb51f52db09503629aa22  | 39 minutes*  |
| 72b28527ed58dc5f22e25fa56ed60e60224668f6 | 36 minutes |
| 0d56e5a75583153cc5c9d88d412300d06fd3684b | 34 minutes |

*~~There are still some improvements left to do because we currently perform only one `ChainExchange` request at a time in the background task.~~

Test env:
- OS: Ubuntu 20.04 (LTS) x64
- Hw: Digital Ocean Droplet, 4 vCPUs, 8GB / 160GB Disk
- File: forest_snapshot_calibnet_2022-09-12_height_1296060.car

Closes #1911

```
2022-09-21T09:25:46.067Z DEBUG  forest_chain_sync::tipset_syncer > ChainExchange message sync tipsets: epochs: 1302345 <- 1302348, len: 3
 2022-09-21T09:25:46.359Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 1302345
 2022-09-21T09:25:46.399Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 1302347
 2022-09-21T09:25:46.469Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 1302348
 2022-09-21T09:25:46.552Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 1302349
```
The log is legit, the missing epoch correspond to a null block.


elmattic:
> * The `AbortHandleRegistry` will go away when we switch to tokio. Keeping it for now is fine.

What mechanisms do we plan to use for canceling futures/tasks once we switch to tokio?

lemmih:
> > * The `AbortHandleRegistry` will go away when we switch to tokio. Keeping it for now is fine.
> 
> What mechanisms do we plan to use for canceling futures/tasks once we switch to tokio?

In tokio we can use [JoinSet](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html)

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1015,92 +1062,72 @@ async fn sync_messages_check_state<DB: BlockStore + Send + Sync + 'static, C: Co
...
+            for tipset in tipsets.into_iter().rev() {
+                match task_chainstore.fill_tipset(&tipset) {
+                    Some(full_tipset) => {
+                        if !batch.is_empty() {
+                            fetch_batch(&batch, &network, &task_chainstore, &s).await?;
+                            batch.clear();
+                        }
+                        s.send(full_tipset).await.map_err(|e| {
+                            TipsetRangeSyncerError::NetworkTipsetQueryFailed(format!("{}", e))
+                        })?;
```
lemmih, line 1090: A `From<SendError<T>>` impl might be worth it here.

lemmih:
Rough benchmark results (syncing to calibnet from yesterday's snapshot):

| main branch | batched |
|--------------|----------|
| 8m9s (2.8x) | 2m51s |

lemmih:
This PR removes the network as a bottleneck. Now we're CPU bound which is great.

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1001,6 +1008,50 @@ fn sync_tipset<DB: BlockStore + Sync + Send + 'static, C: Consensus>(
     })
 }
 
+async fn fetch_batch<DB: BlockStore + Send + Sync + 'static, C: Consensus>(
+    batch: &[Arc<Tipset>],
+    network: &SyncNetworkContext<DB>,
+    chainstore: &Arc<ChainStore<DB>>,
```
lemmih, line 1014: It's essentially always a bad idea to have a reference to an `Arc`. Either use `Arc<T>` or `&T`.
```suggestion
    chainstore: &ChainStore<DB>,
```

blockchain/state_manager/src/utils.rs
```diff
@@ -11,12 +11,35 @@ use forest_actor_interface::{
...
+    }
+
+    /// Aborts all the abortable futures.
+    pub fn abort_futures(&self) {
+        for handle in self.handles.iter().rev() {
+            handle.abort();
+        }
+    }
+}
+
```
lemmih, line 42: This isn't strictly necessary for this PR and it'll be obsolete by the time we move to tokio. I think we should leave it out.

lemmih (changes requested):

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1109,7 +1139,7 @@ async fn sync_messages_check_state<DB: BlockStore + Send + Sync + 'static, C: Co
 async fn validate_tipset<DB: BlockStore + Send + Sync + 'static, C: Consensus>(
     consensus: Arc<C>,
     state_manager: Arc<StateManager<DB>>,
-    chainstore: Arc<ChainStore<DB>>,
+    chainstore: &Arc<ChainStore<DB>>,
     bad_block_cache: Arc<BadBlockCache>,
```
lemmih, line 1143: This can just be a reference:
```suggestion
    bad_block_cache: &BadBlockCache,
```

blockchain/state_manager/src/utils.rs
elmattic: Sure, then will remove this.

lemmih:
Some tipsets have no messages and they can therefore be converted to `FullTipset` without requesting anything from the network. But if we exclude the empty tipsets from the batches then the batch size goes from, say, 8 to an average closer to 2. Keeping the batch size at 8 even if some messages are downloaded twice should be worth it. For calibnet, it changes syncing from being network-bound to being cpu-bound. For mainnet, there are fewer empty tipsets so it doesn't matter as much.

elmattic:
> Some tipsets have no messages and they can therefore be converted to `FullTipset` without requesting anything from the network. But if we exclude the empty tipsets from the batches then the batch size goes from, say, 8 to an average closer to 2. Keeping the batch size at 8 even if some messages are downloaded twice should be worth it. For calibnet, it changes syncing from being network-bound to being cpu-bound. For mainnet, there are fewer empty tipsets so it doesn't matter as much.

Thanks, yes that makes totally sense. Wondering after all if we are not using less network bandwidth like this actually.

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1001,6 +1002,50 @@ fn sync_tipset<DB: BlockStore + Sync + Send + 'static, C: Consensus>(
...
 }
 
+async fn fetch_batch<DB: BlockStore + Send + Sync + 'static, C: Consensus>(
+    batch: &[Arc<Tipset>],
+    network: &SyncNetworkContext<DB>,
+    chainstore: &ChainStore<DB>,
+    s: &channel::Sender<FullTipset>,
+) -> Result<(), TipsetRangeSyncerError<C>> {
+    // Tipsets in `batch` are already in chronological order
+    let head = batch.last().unwrap(); // This is safe because a batch can't be empty
```
lemmih, line 1012: We can avoid the panic entirely.

```suggestion
    if let head = batch.last() {
```

lemmih (changes requested):

lemmih (approved):
Love it. Faster than what we had before and much easier to understand.

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1001,6 +1002,51 @@ fn sync_tipset<DB: BlockStore + Sync + Send + 'static, C: Consensus>(
...
+        debug!("ChainExchange message sync tipsets: epoch: {epoch}, len: {len}");
+
+        let compacted_messages = network
+            .chain_exchange_messages(None, head.key(), len as u64)
+            .await
+            .map_err(TipsetRangeSyncerError::NetworkMessageQueryFailed)?;
+        let messages_iter = compacted_messages.into_iter().rev();
+
+        // Since the bundle only has messages, we have to put the headers in them
+        for (messages, tipset) in messages_iter.zip(batch.iter()) {
```
LesnyRumcajs, line 1025: One-lining this would make it a bit easier to read for me. Not a big deal though.

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1010,97 +1056,67 @@ async fn sync_messages_check_state<DB: BlockStore + Send + Sync + 'static, C: Co
...
+            if batch.len() == REQUEST_WINDOW {
+                fetch_batch(&batch, &network, &task_chainstore, &s).await?;
+                batch.clear();
             }
         }
+        // Fetch last batch
+        fetch_batch(&batch, &network, &task_chainstore, &s).await?;
+        batch.clear();
+
+        Ok::<(), TipsetRangeSyncerError<C>>(())
```
LesnyRumcajs, line 1097: Can `rustc` handle this on its own?
```suggestion
        Ok(())
```

LesnyRumcajs (approved):
I like it. A few comments that may or may not be worth addressing. Nice job.

blockchain/chain_sync/src/tipset_syncer.rs
elmattic: Done.

blockchain/chain_sync/src/tipset_syncer.rs
elmattic: Done.

PR #1948, opened 2022-09-22
Avoid all uses of async-std except when spawning tasks

### LINKED ISSUE #1947: Use tokio instead of async-std for everything except task spawning
**Issue summary**

Before switching the executor from async-std to tokio, we can gradually move away from async-std structures. This means using tokio channels, streams, etc instead of their async-std counterpart.

lemmih:
**Summary of changes**
- 

Unresolved issues:
- `tide` uses async-std.
- Some uses of `async_std::channel` are difficult to replace.

Closes #1947 


lemmih:
Unresolved question: What to do about `tide`? It requires `async_std`. Can we use a different crate for the RPC server?

LesnyRumcajs:
> Unresolved question: What to do about `tide`? It requires `async_std`. Can we use a different crate for the RPC server?

There was an issue to potentially replace `tide`, I'd look into either [warp](https://github.com/seanmonstar/warp) or [rocket](https://github.com/SergioBenitez/Rocket).

lemmih:
Or maybe `actix`. The `jsonrpc-v2` library has support for that crate but maybe that doesn't mean much.

LesnyRumcajs:
`actix` was a bit rough the last time I used it; the API wasn't as high-level as it could be. Perhaps something changed though.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -66,12 +65,13 @@ pub struct ChainStore<DB> {
     /// Publisher for head change events
     publisher: Publisher<HeadChange>,
 
-    /// Tracks head change subscription channels
-    subscriptions: Arc<LockfreeMap<i64, Option<Receiver<HeadChange>>>>,
-
-    /// Keeps track of how many subscriptions there are in order to auto-increment the subscription ID
-    subscriptions_count: AtomicCell<usize>,
+    // XXX: 'subscriptions' disabled since it is unused.
```
LesnyRumcajs, line 68: Can we either delete those or create an issue to use them again?

utils/net_utils/src/download.rs
```diff
@@ -39,18 +40,22 @@ impl<R: AsyncRead + Unpin, W: Write> AsyncRead for FetchProgress<R, W> {
...
-    ) -> Poll<Result<usize, io::Error>> {
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<std::io::Result<()>> {
+        let prev_len = buf.filled().len();
         let r = Pin::new(&mut self.inner).poll_read(cx, buf);
-        if let Poll::Ready(Ok(size)) = r {
-            self.progress_bar.add(size as u64);
+        if let Poll::Ready(Ok(())) = r {
+            self.progress_bar
+                .add((buf.filled().len() - prev_len) as u64);
```
LesnyRumcajs, line 49: Why do we need this change?

utils/paramfetch/src/lib.rs
```diff
@@ -98,40 +86,19 @@ pub async fn get_params(
...
-        mb.await;
-    } else {
-        for t in tasks {
-            t.await;
-        }
+    for t in tasks {
+        t.await;
     }
 
+    // XXX: We return OK(()) even if 'fetch_verify_params' fails. :-/
```
LesnyRumcajs, line 101: Can we do something about it? Or at least create an issue?

utils/net_utils/src/download.rs
lemmih: Because tokio's `AsyncRead` trait is different from the `AsyncRead` trait in `futures` (use by async-std).

utils/paramfetch/src/lib.rs
```diff
@@ -188,32 +149,8 @@ async fn fetch_params(
...
-            progress_bar: pb,
-        };
-        copy(&mut source, &mut writer).await?;
-        source.finish();
-    } else {
-        let mut source = req.await.map_err(|e| anyhow::anyhow!(e))?;
-        copy(&mut source, &mut writer).await?;
-    };
+    let mut source = req.await.map_err(|e| anyhow::anyhow!(e))?.compat();
+    copy(&mut source, &mut writer).await?;
```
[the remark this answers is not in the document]
lemmih: Nope, no progress bars. The previous code was non-functional. If we want progress bars, we should open a new issue.

blockchain/chain/src/store/chain_store.rs
lemmih: I'll open a tracking issue.

LesnyRumcajs (approved):
nice, we'll get there someday!

elmattic:
Nice! Could you update the branch with the new toolchain? (will be easier to switch branches).

utils/paramfetch/src/lib.rs
lemmih: Tracking issue created.

blockchain/chain/src/store/chain_store.rs
lemmih: Tracking issue opened and added to source code: https://github.com/ChainSafe/forest/issues/1976

lemmih:
> Nice! Could you update the branch with the new toolchain? (will be easier to switch branches).

Done.

elmattic:
Did you try moving to `tokio::test` too? Or maybe we will just port the tests in a separate PR?

lemmih:
> Did you try moving to `tokio::test` too? Or maybe we will just port the tests in a separate PR?

We can switch to `tokio::test` once we're not using async-std tasks anymore.

elmattic (approved):

Issue #1961, opened 2022-09-27
Snapshot exporter silently failing


LesnyRumcajs:
**Issue summary**

Snapshot exporter seems to be silently failing, creating snapshots with the same epoch as the imported one. This should be investigated and fixed so that the issue is resolved and in the future, the fiascoes are explicitly reported.

**Other information and links**

PR #1962, opened 2022-09-27
snapshot exporter error handling

LesnyRumcajs:
**Summary of changes**
- increase sync timeout from 30m to 90m, just in case of some failures, so we won't need manual intervention when there's a snapshot gap for whatever reason,
- handle errors in snapshot exporter so they are reported, and the script exits immediately,
- print the logs to container stdout so that they are easily inspectable with `docker logs`
- other minor fixes

Closes https://github.com/ChainSafe/forest/issues/1961


connormullett (approved):
LGTM

Issue #1881, opened 2022-09-06
Use property testing to ensure serialize/de-serialize pairs are correct.


lemmih:
**Issue summary**

Use [quickcheck](https://docs.rs/quickcheck/latest/quickcheck/) to test that any value can be serialized and then de-serialized without loss of information.

The list of types to be tested:
- [x] BlockHeader
- [x] SyncStage
- [x] SyncState
- [x] ChainExchangeResponseStatus
- [x] BeaconEntry
- [x] ElectionProof
- [x] GossipBlock
- [x] Ticket
- [x] TipsetKeys
- [x] Tipset
- [x] Signature
- [x] SignatureType
- [ ] ~VRFProof~
- [ ] ~`Encoding/src/checked_serde_bytes.rs`~
- [ ] ~Node~
- [ ] ~Root~
- [x] Ipld
- [x] KeyInfo
- [ ] ~Policy~
- [ ] ~NetworkVersion~
- [x] PoStProof
- [x] Address
- [x] BigInt
- [x] Cid
- [ ] ~GoVecVisitor~
- [ ] ~PeerId~
- [x] MessageReceipt
- [x] Message
- [x] SignedMessage
- [x] ActorState

**Other information and links**

Blocked by #1880

PR #1963, opened 2022-09-27
Property Test for SignedMessage

jdjaustin:
**Summary of changes**
- Use quickcheck to test our custom implementations of `SignedMessage`

N/A


lemmih:
@jdjaustin All good? It's unfortunate that `serde_json::to_string` cannot be safely used without `SignedMessageJson` but it's out of our hands.

vm/message/Cargo.toml
```diff
@@ -23,13 +23,18 @@ fvm               = "1.1"
...
+serde_json        = "1.0"
 
 [features]
 default = ["blst"]
 blst    = ["forest_crypto/blst", "fvm_shared/blst"]
 
 proofs = []
+
+[package.metadata.cargo-udeps.ignore]
+normal = ["quickcheck"]
```
lemmih, line 40: Remove this.

vm/message/src/signed_message.rs
```diff
@@ -124,6 +124,35 @@ impl Cbor for SignedMessage {
...
+        Ok(Signature::new_secp256k1(DUMMY_SIG.to_vec()))
+    }
+}
+
+// impl quickcheck::Arbitrary for dyn Signer {
+//     fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+//         let dummy_signer = Signer::sign_bytes(_, g, Address).unwrap();
+//         dummy_signer
+//     }
+// }
```
lemmih, line 142: Remove this.

vm/message/src/signed_message.rs
```diff
@@ -124,6 +124,35 @@ impl Cbor for SignedMessage {
...
 
+const DUMMY_SIG: [u8; 1] = [0u8];
+
+//#[derive(Clone)]
+struct DummySigner;
+impl Signer for DummySigner {
+    fn sign_bytes(&self, _: &[u8], _: &Address) -> Result<Signature, anyhow::Error> {
+        Ok(Signature::new_secp256k1(DUMMY_SIG.to_vec()))
+    }
+}
```
lemmih, line 135: Move this away from the top-level.

lemmih (changes requested):

vm/message/Cargo.toml
jdjaustin: The PR was failing the unused dependencies check without this (`cargo udeps` apparently does not see dependencies used only in tests?).

lemmih: If a dependency is only used for testing, it should be in `dev-dependencies`.

lemmih: Suppressing warnings is a last resort and you should always ask in #fil-devs before doing it.

jdjaustin: Ah but I see that one is not required.

jdjaustin: I don't think we can put this one under `dev-dependencies` though. See the discussion under PR #1879 .

vm/message/src/signed_message.rs
```diff
@@ -124,6 +126,25 @@ impl Cbor for SignedMessage {
     }
 }
 
+struct DummySigner;
+impl Signer for DummySigner {
+    fn sign_bytes(&self, _: &[u8], _: &Address) -> Result<Signature, anyhow::Error> {
+        Ok(Signature::new_secp256k1(DUMMY_SIG.to_vec()))
+    }
+}
```
lemmih, line 134: You can move this into the body of `arbitrary`.

vm/message/Cargo.toml
lemmih: Have you tried?

jdjaustin: I guess I didn't since I was just following the example. Seems to work!

vm/message/src/signed_message.rs
```diff
@@ -124,6 +124,27 @@ impl Cbor for SignedMessage {
...
+            fn sign_bytes(&self, _: &[u8], _: &Address) -> Result<Signature, anyhow::Error> {
+                Ok(Signature::new_secp256k1(DUMMY_SIG.to_vec()))
+            }
+        }
+
+        let msg = Message {
+            to: Address::new_id(u64::arbitrary(g)),
+            from: Address::new_id(u64::arbitrary(g)),
+            ..Message::default()
+        };
```
lemmih, line 143: Define as many fields as possible.

lemmih (changes requested):

lemmih (approved):

vm/message/src/signed_message.rs
```diff
@@ -124,6 +124,34 @@ impl Cbor for SignedMessage {
...
 
+#[cfg(test)]
+impl quickcheck::Arbitrary for SignedMessage {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        const DUMMY_SIG: [u8; 1] = [0u8];
+
+        struct DummySigner;
+        impl Signer for DummySigner {
+            fn sign_bytes(&self, _: &[u8], _: &Address) -> Result<Signature, anyhow::Error> {
+                Ok(Signature::new_secp256k1(DUMMY_SIG.to_vec()))
```
LesnyRumcajs, line 135: it's used in one place only, I don't think there's a need to make it a const. Would `new_secp256k1(vec![0u8])` work? Not a big deal though.

PR #1970, opened 2022-09-29
Property Test for Message

jdjaustin:
**Summary of changes**
- Use quickcheck to test our custom implementations of `Message`

N/A


lemmih (changes requested):
Could you refactor the `SignedMessage` testing code to use the same `Arbitrary` impl for `MessageWrapper`?

lemmih (changes requested):
- Move all the testing code into the `tests` module. You should only require one `#[cfg(test)]`.
- Refactor the code for `MessageWrapper`. We only want one implementation that generates an arbitrary message, not two.

vm/message/src/message.rs
```diff
@@ -3,7 +3,7 @@
 
 #[cfg(feature = "proofs")]
 use fvm::gas::Gas;
-#[cfg(feature = "proofs")]
+#[cfg(any(feature = "proofs", test))]
```
lemmih, line 6: This should no longer be necessary.

tyshko5 (approved):

PR #1980, opened 2022-09-30
Fix rubocop action

LesnyRumcajs:
**Summary of changes**
- gets rid of archived `rubocop` action in favour of a manual one. For our use case it should be enough.

Closes 


lemmih:
Do or do not, there is no try.

hanabi1224 (approved):
Thanks for the fix!

lemmih (approved):

PR #1981, opened 2022-09-30
Property Test for Cid

jdjaustin:
**Summary of changes**
- Use quickcheck to test our custom implementations of `Cid`

N/A


lemmih (approved):
`cid` does implement `quickcheck::Arbitrary` but unfortunately they are using a dated version of quickcheck. :-/

LesnyRumcajs (commented):
The same comment as for the `BigInt` test. I don't see a value-added here. If we want to make sure `Cid` serializes/deserializes properly, we should open a PR to `ref-fvm` or wherever that type is defined and not test it in its clients.

lemmih:
> The same comment as for the `BigInt` test. I don't see a value-added here. If we want to make sure `Cid` serializes/deserializes properly, we should open a PR to `ref-fvm` or wherever that type is defined and not test it in its clients.

Neither `ref-fvm` nor `cid` defines the custom JSON serialization that we require. As long as we have a custom serializer/de-serializer, we need to test that it is lossless. It would be great if we didn't need this JSON serialization code at all, though.

LesnyRumcajs:
I guess it's the same issue, this PR is not testing what is supposed to do, as you noted here? https://github.com/ChainSafe/forest/pull/1982#issuecomment-1265150779

lemmih:
@jdjaustin Like with the other PR, explicitly use the `serialize` and `deserialize` functions.

lemmih (changes requested):
Test the custom serialization functions.

lemmih (approved):

tyshko5 (approved):

PR #1982, opened 2022-09-30
Property Test for BigInt

jdjaustin:
**Summary of changes**
- Use quickcheck to test our custom implementations of `BigInt`

N/A

### DIFF
--- a/utils/json/Cargo.toml
+++ b/utils/json/Cargo.toml
@@ -20,3 +20,8 @@ multihash = { workspace = true, default-features = false, features = [
 ] }
 num-bigint.workspace = true
 serde = { workspace = true, features = ["derive"] }
+
+[dev-dependencies]
+quickcheck        = { version = "1" }
+quickcheck_macros = "1"
+serde_json        = "1.0"
--- a/utils/json/src/bigint.rs
+++ b/utils/json/src/bigint.rs
@@ -1,6 +1,9 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+#[cfg(test)]
+use fvm_shared::bigint::BigInt;
+
 pub mod json {
     use num_bigint::BigInt;
     use serde::{Deserialize, Serialize};
@@ -46,5 +49,33 @@ pub mod json {
             }
             Ok(None)
         }
+    }
+}
+
+#[cfg(test)]
+#[derive(Clone, Debug, PartialEq)]
+struct BigIntWrapper {
+    bigint: BigInt,
+}
+
+#[cfg(test)]
+impl quickcheck::Arbitrary for BigIntWrapper {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let bigint = BigInt::from(u32::arbitrary(g));
+        BigIntWrapper { bigint }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use quickcheck_macros::quickcheck;
+    use std::str::FromStr;
+
+    #[quickcheck]
+    fn bigint_roundtrip(bigint: BigIntWrapper) {
+        let serialized = BigInt::to_string(&bigint.bigint);
+        let parsed = BigInt::from_str(&serialized).unwrap();
+        assert_eq!(bigint.bigint, parsed);
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3014,7 +3014,10 @@ dependencies = [
  "fvm_shared",
  "multihash",
  "num-bigint 0.4.3",
+ "quickcheck",
+ "quickcheck_macros",
  "serde",
+ "serde_json",
 ]
 
 [[package]]


utils/json/src/bigint.rs
```diff
@@ -48,3 +51,31 @@ pub mod json {
...
+    bigint: BigInt,
+}
+
+#[cfg(test)]
+impl quickcheck::Arbitrary for BigIntWrapper {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let bigint = BigInt::from(u32::arbitrary(g));
+        BigIntWrapper { bigint }
+    }
+}
```
lemmih, line 67: `BigInt` already has an `Arbitrary` implementation.
```suggestion
```

utils/json/src/bigint.rs
```diff
@@ -48,3 +51,31 @@ pub mod json {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use quickcheck_macros::quickcheck;
+    use std::str::FromStr;
+
+    #[quickcheck]
+    fn bigint_roundtrip(bigint: BigIntWrapper) {
```
lemmih, line 76:
```suggestion
    fn bigint_roundtrip(bigint: BigInt) {
```

utils/json/Cargo.toml
```diff
@@ -20,3 +20,8 @@ multihash = { workspace = true, default-features = false, features = [
 ] }
 num-bigint.workspace = true
 serde = { workspace = true, features = ["derive"] }
+
+[dev-dependencies]
+quickcheck        = { version = "1" }
```
lemmih, line 25: Add `num-bigint        = { workspace = true, features = ["quickcheck"] }`

lemmih (changes requested):

LesnyRumcajs (commented):
I don't see the point in this test. It's a foreign type so it should be tested in its respective crate. Even `ref-fvm` re-exports this from `num-bigint`. Can you elaborate on the value added here?

lemmih:
> I don't see the point in this test. It's a foreign type so it should be tested in its respective crate. Even `ref-fvm` re-exports this from `num-bigint`. Can you elaborate on the value added here?

We're testing the JSON serialize/de-serialize implementation which is our own and not part of any other crate. We have the assumption that a round-trip through JSON is lossless and this test is making sure that assumption is correct.

lemmih:
> > > I don't see the point in this test. It's a foreign type so it should be tested in its respective crate. Even `ref-fvm` re-exports this from `num-bigint`. Can you elaborate on the value added here?
> > > Where
> > > We're testing the JSON serialize/de-serialize implementation which is our own and not part of any other crate. We have the assumption that a round-trip through JSON is lossless and this test is making sure that assumption is correct.
> 
> Sorry, I'm missing something. Where in this PR do we do JSON serialization in the test? Is `to_string()` invoking that?

Ah, you're right. The wrong serialization code is being tested. It should test the two above `serialize`/`deserialize` functions but it isn't.

LesnyRumcajs:
@lemmih Ya, then it would make sense. In its current state, this PR needs some touch.

lemmih:
@jdjaustin You can use these two macros to test stand-alone `serialize` and `deserialize` functions:

```rust
    macro_rules! to_string_with {
        ($obj:expr, $serializer:path) => {{
            let mut writer = Vec::new();
            $serializer($obj, &mut serde_json::ser::Serializer::new(&mut writer)).unwrap();
            String::from_utf8(writer).unwrap()
        }};
    }

    macro_rules! from_str_with {
        ($str:expr, $deserializer:path) => {
            $deserializer(&mut serde_json::de::Deserializer::from_str($str)).unwrap()
        };
    }
```

utils/json/src/bigint.rs
```diff
@@ -48,3 +51,30 @@ pub mod json {
...
+            $serializer($obj, &mut serde_json::ser::Serializer::new(&mut writer)).unwrap();
+            String::from_utf8(writer).unwrap()
+        }};
+    }
+
+    macro_rules! from_str_with {
+        ($str:expr, $deserializer:path) => {
+            $deserializer(&mut serde_json::de::Deserializer::from_str($str)).unwrap()
+        };
+    }
```
LesnyRumcajs, line 72: Re-declaring those macros for every test is a bit counter-productive (like https://github.com/ChainSafe/forest/pull/1992/files#diff-b734c251f7cad8d229fbb92930569fe8b759276951e060512b2fc2d5c25f540aR583-R595 or https://github.com/ChainSafe/forest/pull/1991/files#diff-1b7afae453511db460dc76eb895dfcf001193d438377d57b142bc3c695c90b8cR145-R157). A better approach would be to export it to e.g. `forest-test-utils` and then just import it. Otherwise, we're not really gaining anything from a single-use macros.

LesnyRumcajs (changes requested):

utils/test_utils/src/chain_structures.rs
```diff
@@ -189,3 +189,21 @@ pub fn construct_chain_exchange_response() -> ChainExchangeResponse {
...
+    }};
+}
+
+// Deserialize macro used for testing
+#[macro_export]
+macro_rules! from_str_with {
+    ($str:expr, $deserializer:path) => {
+        $deserializer(&mut serde_json::de::Deserializer::from_str($str)).unwrap()
+    };
+}
```
lemmih, line 209: I'm not sure `chain_structures.rs` is the right home. How about you put them in `lib.rs` for now?

utils/json/src/bigint.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+#[cfg(test)]
+use fvm_shared::bigint::BigInt;
+
```
lemmih, line 6: You can move this into the `tests` module.

lemmih (changes requested):

utils/json/Cargo.toml
```diff
@@ -20,3 +20,10 @@ multihash = { workspace = true, default-features = false, features = [
 ] }
 num-bigint.workspace = true
 serde = { workspace = true, features = ["derive"] }
+
+[dev-dependencies]
+forest_test_utils           = { workspace = true, features = ["test_constructors"] }
```
lemmih, line 25: `test_constructors` is not strictly necessary but that feature will be removed soon in any case so it doesn't hurt to leave it.

lemmih (approved):

lemmih (approved):

tyshko5 (approved):

LesnyRumcajs (approved):

Issue #1873, opened 2022-09-02
Long-run on calibnet fails with ChainForkLengthExceedsMaximum


LesnyRumcajs:
**Describe the bug**

It's been happening consistently in the sync check while using Forest-exported calibnet snapshots. Not sure if the behaviour repeats for all snapshots, but happened for `forest_snapshot_2022-August-30_height_1258614.car`.

On theory is that we get into bad state with `ChainForkLengthExceedsMaximum` and then we export the snapshot, thus replicating the bad state. It shouldn't happen but maybe it's worth looking into.

**To Reproduce**
1. Start Forest by importing a snapshot mentioned above.
2. Import it like in the sync check script
3. See the `ChainForkLengthExceedsMaximum` after some time.

**Log output**
[forest-calibnet_2022-09-02T09_47_36_forest.log](https://github.com/ChainSafe/forest/files/9478914/forest-calibnet_2022-09-02T09_47_36_forest.log)

**Expected behaviour**

Forest can continuously sync without issues.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Fedora 36
 - Branch/commit e.g. 1f833d078febd2f2aa068d97f51c757a468f3cd9

**Other information and links**

LesnyRumcajs:
Update: this happened for mainnet as well with `minimal_finality_stateroots_2134801_2022-09-05_04-00-30.car` after about 1 day.
```
 2022-09-07T11:08:30.874Z INFO  forest_chain_sync::chain_muxer   > Local node is behind the network, starting BOOTSTRAP from LOCAL_HEAD = 2140907 -> NETWORK_HEAD = 2141417
 2022-09-07T11:08:31.124Z INFO  forest_chain_sync::tipset_syncer > Fork detected, searching for a common ancestor between the local chain and the network chain
 2022-09-07T11:08:34.431Z ERROR forest_chain_sync::chain_muxer   > Bootstrapping failed, re-evaluating the network head to retry the bootstrap. Error = TipsetRangeSyncer(ChainForkLengthExceedsMaximum)
 2022-09-07T11:08:34.431Z INFO  forest_chain_sync::chain_muxer   > Evaluating network head...
```

LesnyRumcajs:
[forest_lotus.log](https://github.com/ChainSafe/forest/files/9546189/forest_lotus.log)

[...]

PR #1985, opened 2022-10-03
fix fork threshold issue

LesnyRumcajs:
Co-authored-by: Eric Tu <tu.eric@hotmail.com>

**Summary of changes**
- @ec2 fix seems to be working well. No thresholds were exceeded for over four days. The only issue is that we still detect lots of forks, but let's make this part of another investigation.

Closes https://github.com/ChainSafe/forest/issues/1873


lemmih (approved):
Looks good. Seems like it was an oversight in the original code.

elmattic (approved):

PR #1998, opened 2022-10-04
Tweak our RocksDB settings for better performances

### LINKED ISSUE #1841: Investigate the effect of different RocksDB options.
**Issue summary**

Importing a mainnet snapshot into RocksDB is slower than expected: The import speed starts high but then drops as low as 20MB/s. This is a known issue for RocksDB and there are several options specifically dealing with bulk data import. We should identify the options that are relevant for improving the performance of bulk imports and then run benchmarks.

**Task summary**

- [ ] Identity which RocksDB options affect bulk importing. Things like temporarily disabling compaction, etc.
- [ ] Add those options to the configuration file.
- [ ] Do rough benchmarks (remove the db, import a snapshot, quit). We're looking for 2x to 10x improvements, not 5-10%.
- [ ] Report findings in this issue.

**Acceptance Criteria**

- [ ] A recommendation of which RocksDB configuration to use by default.

https://rockset.com/blog/optimizing-bulk-load-in-rocksdb/
https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html
https://github.com/EighteenZi/rocksdb_wiki/blob/master/RocksDB-Tuning-Guide.md
https://github.com/facebook/rocksdb/blob/main/options/options.cc#L451-L494

elmattic:
**Summary of changes**
- Add `optimize_filters_for_hits` and `optimize_for_point_lookup` for better db read performances
- Add `enable_statistics` and `stats_dump_period_sec` for troubleshooting perf issues and/or inspecting db stats
- Change `compaction_style` and `compression_type` to be `String`'s
- Add a stopwatch to measure the reading and bulk writes into the db of the .car file

The new `optimize_filters_for_hits` and `optimize_for_point_lookup` are respectively set to default values `true` and `8`.

Those two optimizations bring around `~12%` performance speed-up during a snapshot import:

| Change | Import Time | Total Time |
| :---  | :---: | :---: |
| baseline (no change) | ~1367s | ~3120s |
| set `optimize_filters_for_hits` to `true` and `optimize_for_point_lookup` to `8MB` | ~1307s | ~2776s |

They should improve `RocksDB` reading latency in the long run as well.

The removal of `Option` for `compaction_style` or `compression_type` is to map the default values to our own configuration.

Closes #1841


lemmih:
Why not set `compaction_style` to `None` and `optimize_for_point_lookup` to `256`? From your data, it seems to be a huge win for a tiny cost of memory.

lemmih:
And compacting once after the DB import isn't feasible?

elmattic:
> And compacting once after the DB import isn't feasible?

For calibnet it was doable. For mainnet it would result in hours of compactions and terabytes of readback.

My bet is that ingesting ready to go `.sst`s will be much easier.

lemmih:
> > And compacting once after the DB import isn't feasible?
> 
> For calibnet it was doable. For mainnet it would result in hours of compactions and terabytes of readback.
> 
> My bet is that ingesting ready to go `.sst`'s will be much easier.

I don't understand why auto compactions are fast but a single manual compaction is slow.

elmattic:
> > > And compacting once after the DB import isn't feasible?
> > 
> > 
> > For calibnet it was doable. For mainnet it would result in hours of compactions and terabytes of readback.
> > My bet is that ingesting ready to go `.sst`'s will be much easier.
> 
> I don't understand why auto compactions are fast but a single manual compaction is slow.

Yes, I don't understand this as well. Unless reading the source code and understanding how it works, there is little documentation on `CompactRange`.

lemmih:
> > > > And compacting once after the DB import isn't feasible?
> > > 
> > > 
> > > For calibnet it was doable. For mainnet it would result in hours of compactions and terabytes of readback.
> > > My bet is that ingesting ready to go `.sst`'s will be much easier.
> > 
> > 
> > I don't understand why auto compactions are fast but a single manual compaction is slow.
> 
> Yes, I don't understand this as well. Unless reading the source code and understanding how it works, there is little documentation on `CompactRange`.

Is there a downside to setting the size of L0 files to something like 1GB? If not, that would limit the number of L0 files to a manageable number.

lemmih (approved):
Looks good. Still interested in disabling compaction entirely.

node/db/src/rocks.rs
```diff
@@ -40,19 +40,43 @@ impl RocksDb {
...
             db_opts.set_disable_auto_compactions(false);
         } else {
             db_opts.set_disable_auto_compactions(true);
         }
-        if let Some(compression_type) = &config.compression_type {
-            db_opts.set_compression_type(compression_type_from_str(compression_type).unwrap());
-        }
+        db_opts.set_compression_type(compression_type_from_str(&config.compression_type).unwrap());
         if config.enable_statistics {
+            db_opts.set_stats_dump_period_sec(config.stats_dump_period_sec);
```
LesnyRumcajs, line 52: maybe `stats_dump_period_sec` could be an `Option`, so we wouldn't need `enable_statistics`, as they convey the same logic. Not a big deal though.

node/db/src/rocks.rs
```diff
@@ -40,19 +40,43 @@ impl RocksDb {
...
+            opts.set_format_version(5);
+            opts.set_data_block_index_type(DataBlockIndexType::BinaryAndHash);
+            opts.set_data_block_hash_ratio(0.75);
+            opts.set_bloom_filter(10.0, false);
+            let cache = Cache::new_lru_cache(cache_size * 1024 * 1024).unwrap();
+            opts.set_block_cache(&cache);
+            db_opts.set_block_based_table_factory(&opts);
+            db_opts.set_memtable_prefix_bloom_ratio(0.02);
+            db_opts.set_memtable_whole_key_filtering(true);
+        }
```
LesnyRumcajs, line 71: :exploding_head:

utils/genesis/src/lib.rs
```diff
@@ -143,7 +144,12 @@ where
         info!("Reading file...");
         let file = File::open(&path).await?;
         let reader = FetchProgress::fetch_from_file(file).await?;
-        load_and_retrieve_header(sm.blockstore(), reader, skip_load).await?
+        {
+            let stopwatch = time::Instant::now();
+            let cids = load_and_retrieve_header(sm.blockstore(), reader, skip_load).await?;
+            info!("Loaded .car file in {}s", stopwatch.elapsed().as_secs());
+            cids
+        }
```
LesnyRumcajs, line 152: Not a huge deal, but you are measuring the time only for the local file branch. For consistency, it should also be measured for the remote file branch above (L138-142). It should be completely fine to include parsing `Url` and opening files to simplify logic, as those operations' overhead is negligible compared to loading the snapshot.

LesnyRumcajs (approved):

utils/genesis/src/lib.rs
elmattic: Good point, will fix this.

node/db/src/rocks.rs
elmattic: Yes, this could later be refactorized in some function.

lemmih:
Let's merge this and then we can look at larger SST files + no compaction in another PR.

PR #2007, opened 2022-10-05
[WIP] IpldJson Property Test

jdjaustin:
**Summary of changes**
- Use quickcheck to test our custom implementations of `IpldJson`

N/A


lemmih:
Remember not to use `forest_test_utils::to_string_with` when you have `*Json` structures. Let those structures do the heavy lifting for you.

jdjaustin:
> Remember not to use `forest_test_utils::to_string_with` when you have `*Json` structures. Let those structures do the heavy lifting for you.

I removed the `forest_test_utils` macros, and even though I have tried several different things, it seems like the same problem is occurring with the test; the LHS of the comparison is giving a number that's not in scientific notation while the RHS is giving a number that is in scientific notation. Not sure if it's an issue with the test or the way `IpldJson` is serialized/deserialized...

lemmih:
JSON cannot natively represent i128 numbers. We could use a scheme like `BYTES_JSON_KEY` for numbers that can't fit in an `i64`.

lemmih:
@jdjaustin Maybe we don't even want a json serializer for `Ipld`. If it is only used for debugging then we can just remove it and rely on `Debug` instead.

jdjaustin:
> @jdjaustin Maybe we don't even want a json serializer for `Ipld`. If it is only used for debugging then we can just remove it and rely on `Debug` instead.

How can I verify if it is only used for debugging?

lemmih:
> > @jdjaustin Maybe we don't even want a json serializer for `Ipld`. If it is only used for debugging then we can just remove it and rely on `Debug` instead.
> 
> How can I verify if it is only used for debugging?

First you delete the deserializer. If nothing breaks then it is unused. Otherwise you look at what breaks and check if it is debugging code or not.

jdjaustin:
See #2072

Issue #2026, opened 2022-10-10
Rename `forest fetch sync` to `forest snapshot sync`


hanabi1224:
**Issue summary**

As is discussed [here](https://github.com/ChainSafe/forest/issues/1954#issuecomment-1266885202), renaming `fetch` to `snapshot` so that we can implement `forest-cli snapshot sync | list | remove | clean` more intuitively

**Other information and links**

PR #2039, opened 2022-10-11
feat: forest-cli snapshot subcommand

hanabi1224:
**Summary of changes**
- [x] Move `chain fetch` and `chain export` to `snapshot fetch` and `snapshot export`
- [x] Add deprecation warnings to `chain fetch` and `chain export` (Assuming some automation scripts use the old commands, not deleting them in this PR)
- [x] Add `snapshot list` and `snapshot remove`
- [x] Add `snapshot prune`

Closes https://github.com/ChainSafe/forest/issues/2026


lemmih:
Maybe `snapshot prune` to delete old snapshots while keeping the most recent one around.

lemmih (approved):

LesnyRumcajs (changes requested):

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,217 @@
...
+            } => {
+                let yes = yes.unwrap_or_default();
+                let snapshot_dir = snapshot_dir
+                    .clone()
+                    .unwrap_or_else(|| default_snapshot_dir(&config));
+                let mut snapshot_path = snapshot_dir;
+                snapshot_path.push(filename);
+                if snapshot_path.exists()
+                    && snapshot_path.is_file()
+                    && snapshot_path.extension() == Some(OsStr::new("car"))
```
elmattic, line 178: Don't we need to remove the associated `.sha256sum` if there is one?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,217 @@
...
+use forest_rpc_client::chain_ops::*;
+use std::{collections::HashMap, ffi::OsStr, fs, path::PathBuf};
+use strfmt::strfmt;
+use time::OffsetDateTime;
+
+pub(crate) const OUTPUT_PATH_DEFAULT_FORMAT: &str =
+    "forest_snapshot_{chain}_{year}-{month}-{day}_height_{height}.car";
+
+#[derive(Debug, StructOpt)]
+pub enum SnapshotCommands {
```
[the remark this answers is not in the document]
LesnyRumcajs: @hanabi1224 something along those lines https://github.com/ChainSafe/forest/blob/main/forest/cli/tests/config_tests.rs

forest/cli/Cargo.toml
```diff
@@ -13,6 +13,7 @@ base64.workspace                 = true
 byte-unit                        = "4.0"
 chrono                           = "0.4"
 cid                              = { workspace = true, default-features = false, features = ["std"] }
+console                          = "0.15"
```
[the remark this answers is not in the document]
hanabi1224: I can switch to `dialogure`, according to the [doc](https://docs.rs/dialoguer/0.10.2/dialoguer/), `dialogure` re-exports `console`

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: Good catch, will handle that

forest/cli/Cargo.toml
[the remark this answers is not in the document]
hanabi1224: fixed

forest/cli/src/cli/mod.rs
```diff
@@ -105,6 +107,10 @@ pub enum Subcommand {
     /// Manage node configuration
     #[structopt(name = "config")]
     Config(ConfigCommands),
+
+    /// Manage snapshots
+    #[structopt(name = "snapshot")]
```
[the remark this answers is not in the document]
hanabi1224: Fixed.  I left `Fetch` as it is so that we keep a reference usage of renaming, hopefully we won't have to check the doc if we need it someday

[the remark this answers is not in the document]
LesnyRumcajs: rock-solid!

forest/cli/src/cli/snapshot_cmd.rs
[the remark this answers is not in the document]
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// pattern will be ignored
+    Prune {
+        /// Directory to which the snapshots are downloaded. If not provided, it will be the
+        /// default Forest data location.
+        #[structopt(short, long)]
+        snapshot_dir: Option<PathBuf>,
+
+        /// Answer yes to all forest-cli yes/no questions without prompting
+        #[structopt(short, long)]
+        yes: bool,
```
LesnyRumcajs, line 92: Should we also have `no` to always decline prompts? E.g. you can find a similar approach in e.g. `dnf` (Fedora package manager)
```
❯ dnf install --help | rg assume
                   [--rpmverbosity [debug level name]] [-y] [--assumeno]
  -y, --assumeyes       automatically answer yes for all questions
  --assumeno            automatically answer no for all questions
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,326 @@
...
+                        let path = entry.path();
+                        if path.is_file() {
+                            if let Some(filename) = entry.file_name().to_str() {
+                                if filename.ends_with(".car") {
+                                    println!("{filename}");
+                                }
+                            }
+                        }
+                    }
+                }
```
LesnyRumcajs, line 178: This snippet is a bit complex, it would be nice to move it to a separate, private method (in general, for readability it would be even better to have all those commands in different methods).

Another thing, it may become more readable if you'd a nice combo along the lines of:
```
fs::read_dir(snapshot_dir)
   .flatten()
   .map(|entry| entry.path())
   .filter(|entry| entry.is_file())
   ...
```
or maybe it wouldn't be more readable. Up to you :)

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,326 @@
...
+            }
+            Self::Remove {
+                filename,
```
LesnyRumcajs, line 181: Same comment as for `List`

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,326 @@
...
+                    delete_snapshot(&snapshot_path);
+                } else {
+                    let term = Term::stdout();
+                    term.write_line(&format!(
+                        "{:?} is not a valid snapshot file path, to list all snapshots, run forest-cli snapshot list",
+                        snapshot_path)
+                    ).unwrap();
+                }
+            }
+            Self::Prune { snapshot_dir, yes } => {
```
LesnyRumcajs, line 217: The same comment for `List`, having 10 levels of indentation begs for refactoring. Also, there seem to be a lot of opportunity for `map`, `filter` and `filter_map`.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,326 @@
...
+fn default_snapshot_dir(config: &Config) -> PathBuf {
+    config
+        .client
+        .data_dir
+        .join("snapshots")
+        .join(config.chain.name.clone())
+}
+
+fn delete_snapshot(snapshot_path: &PathBuf) {
+    let term = Term::stdout();
```
LesnyRumcajs, line 313: Why do you use `term.write_line` and not just regular print macros?

LesnyRumcajs (changes requested):
Thanks for writing the tests. I like them! It'd be great to introduce some refactoring though, the `run` method has over 200 LoC now.

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: I think --no makes sense when there are optional steps, like checking updates, for this case maybe --dry-run is better?

LesnyRumcajs: Both kind of make sense. We just need to be tactical and not implement stuff that won't be used. If you believe the 99% use case would be to use `-y` then there's no point in spending time on some other flags that won't be used outside of my imagination. :) Up to you! :)

hanabi1224: Sounds good, then I will not implement `--no` to reduce the code complexity and our maintenance cost, ppl can still use `yes` tool in `coreutils` to achieve this.  `yes n | forest-cli snapshot prune`

hanabi1224: e.g. podman image prune implements --force only

```
➜  forest git:(subcommand-snapshot) podman image prune -h
Remove unused images

Description:
  Removes dangling or unused images from local storage.

Usage:
  podman image prune [options]

Examples:
  podman image prune

Options:
  -a, --all                  Remove all images not in use by containers, not just dangling ones
      --filter stringArray   Provide filter values (e.g. 'label=<key>=<value>')
  -f, --force                Do not prompt for confirmation

➜  forest git:(subcommand-snapshot) yes n | podman image prune
WARNING! This command removes all dangling images.
Are you sure you want to continue? [y/N] %
➜  forest git:(subcommand-snapshot)
```

LesnyRumcajs: In general, `yes` could be potentially also used for `--yes` to further reduce code complexity

hanabi1224: Hmm, just found it's more complicated than I thought, rust stdlib does not provide `stdin().read_char` due to unicode reason, and the console implementation does not work with `yes` tool. I might have to switch back to stdlib to make `yes` tool work.

LesnyRumcajs: Then maybe let's go with `--force` path. Whatever we choose let's have it uniform across the entire implementation

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: Fixed. I have switch back to using stdlib, and it now works with `yes` tool

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,329 @@
...
+        .unwrap_or_else(|| default_snapshot_dir(config));
+    println!("Snapshot dir: {:?}", snapshot_dir);
+    if let Ok(dir) = fs::read_dir(snapshot_dir) {
+        println!("\nLocal snapshots:");
+        for path in dir
+            .flatten()
+            .map(|entry| entry.path())
+            .filter(|p| p.is_file())
+        {
+            if let Some(Some(filename)) = path.file_name().map(|n| n.to_str()) {
```
LesnyRumcajs, line 191: I believe you could omit one `Some` with [filter_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map). Most likely you'd be able to reach the print statement without any `if`s unless it hampers readability in your opinion.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,329 @@
...
+                if filename.ends_with(".car") {
+                    println!("{filename}");
+                }
+            }
+        }
+    }
+}
+
+fn remove(config: &Config, filename: &PathBuf, snapshot_dir: &Option<PathBuf>, yes: bool) {
```
LesnyRumcajs, line 200: Use the Force, @hanabi1224 !
```suggestion
fn remove(config: &Config, filename: &PathBuf, snapshot_dir: &Option<PathBuf>, force: bool) {
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,329 @@
...
+        if !yes && !prompt_confirm() {
+            println!("Aborted.");
+            return;
+        }
+
+        delete_snapshot(&snapshot_path);
+    } else {
+        println!(
+                "{:?} is not a valid snapshot file path, to list all snapshots, run forest-cli snapshot list",
+                snapshot_path);
```
LesnyRumcajs, line 220: debug representation of `Path` is a bit shady, what do you think about using a regular `Display` formatting with e.g. `snapshot_path.display()`?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,329 @@
...
+            }
+        }
+    }
+}
+
+fn remove(config: &Config, filename: &PathBuf, snapshot_dir: &Option<PathBuf>, yes: bool) {
+    let snapshot_dir = snapshot_dir
+        .clone()
+        .unwrap_or_else(|| default_snapshot_dir(config));
+    let mut snapshot_path = snapshot_dir;
```
LesnyRumcajs, line 204: I believe you could omit mutability by using `join`, not a big deal though. Same in some other places.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+fn prompt_confirm() -> bool {
+    println!("Do you want to continue? [y/n]");
+    let mut line = String::new();
+    std::io::stdin().read_line(&mut line).unwrap();
+    let line = line.trim().to_lowercase();
+    line == "y" || line == "yes"
+}
```
LesnyRumcajs, line 329: Can we use dialoguer API instead? It gives pretty colours and it's nicer not to re-implement the same thing.

https://github.com/mitsuhiko/dialoguer/blob/master/examples/confirm.rs

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        if snapshots_with_valid_name.len() < 2 {
+            println!("No files to delete");
+            return;
+        }
+
+        let mut snapshots_to_delete = snapshots_with_valid_name;
+        let mut index_to_keep = 0;
+        println!("Files to delete:");
+        if let Some(snapshot_to_keep) = snapshot_to_keep {
```
LesnyRumcajs, line 279: could we somehow simplify this with https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position ?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,329 @@
...
+        .unwrap();
+        if let Ok(dir) = fs::read_dir(snapshot_dir) {
+            println!("\nLocal snapshots:");
+            for path in dir
+                .flatten()
+                .map(|entry| entry.path())
+                .filter(|p| p.is_file())
+            {
+                if let Some(Some(filename)) = path.file_name().map(|n| n.to_str()) {
+                    let captures = pattern.captures(filename);
```
LesnyRumcajs, line 246: do we need to do so much `if-letting` here? If we match the regular expression, then the unwraps shouldn't be fallible, no?

LesnyRumcajs (changes requested):

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: An owned `path` is needed below, I guess using `filter_map` will lose it?

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: It's built on top of console which does not work with `yes` tool, basically they are reading from terminal instead of stdin

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: It does `println!` in the for loop, I think it's better not to call functions with a side effect in the lambda

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: I thought @lemmih mentioned we want to remove all `expect`, `unwrap` usage someday, I can do `unwrap` if that's preferable

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: fixed

forest/cli/src/cli/snapshot_cmd.rs
LesnyRumcajs: We would like not to have `expect`s or `unwrap`s that may fail, causing a panic. Infallible `unwrap`s are completely fine, e.g. if you **know** for sure that given logic will always produce `Some` or `Ok`. To my understanding, if an expression matches then you **know** for sure that the captures are not nulls, or is it not the case?

forest/cli/src/cli/snapshot_cmd.rs
LesnyRumcajs: Ah yes, I see.

forest/cli/src/cli/snapshot_cmd.rs
LesnyRumcajs: That's a bit disappointing, I expected such support from, I guess, the most popular crate for such operations. Oh well.

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: That makes sense, fixed

forest/cli/src/cli/snapshot_cmd.rs
LesnyRumcajs: How about something like this?
```
        dir.flatten()
            .map(|entry| entry.path())
            .filter(|p| p.is_file() && p.ends_with(".car"))
            .for_each(|p| println!("{}", p.display()))
```

hanabi1224: Hmm I'm not a big fan of calling functions with side effects like `println` in lambdas but I guess `for_each` is for that purpose, also I just noticed there're 2 nested `Some` usage and one of them does not require an owned path, will fix as suggested

hanabi1224: Fixed with slight modification cuz [`PathBuf::ends_with`](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.ends_with) does not work with extensions

```rust
use std::path::Path;

let path = Path::new("/etc/resolv.conf");

assert!(path.ends_with("resolv.conf"));
assert!(path.ends_with("etc/resolv.conf"));
assert!(path.ends_with("/etc/resolv.conf"));

assert!(!path.ends_with("/resolv.conf"));
assert!(!path.ends_with("conf")); // use .extension() instead
```

```rust
dir.flatten()
            .map(|entry| entry.path())
            .filter(|p| {
                p.is_file()
                    && p.extension()
                        .map(|ext| ext.to_str().unwrap_or_default() == "car")
                        .unwrap_or_default()
            })
            .for_each(|p| println!("{}", p.display()));
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -0,0 +1,325 @@
...
+    println!("Snapshot dir: {}", snapshot_dir.display());
+    if let Ok(dir) = fs::read_dir(snapshot_dir) {
+        println!("\nLocal snapshots:");
+        dir.flatten()
+            .map(|entry| entry.path())
+            .filter(|p| {
+                p.is_file()
+                    && p.extension()
+                        .map(|ext| ext.to_str().unwrap_or_default() == "car")
+                        .unwrap_or_default()
```
LesnyRumcajs, line 192: do we need this map or can we simplify this?
```suggestion
                    && p.extension().unwrap_or_default() == "car"
```

LesnyRumcajs: probably `is_file` is not needed, I guess for directories the `extension` then would return `None` which would not match witch `car`.

hanabi1224: Fixed. Didn't know `&OsStr == &str` works

Issue #2038, opened 2022-10-11
Calibnet sync check GH Action


LesnyRumcajs:
**Issue summary**

Implement a check that verifies on a push (both PR and merge to `main`) that Forest can sync to `calibnet`. This check should be done on the `buildjet` machines.
- [x] Download the latest calibnet snapshot (use `forest chain fetch`),
- [x] Run `forest`, importing the snapshot,
- [x] Wait for sync. It'll be nice to do a health check as well.

**Other information and links**

The check will most likely need to be disabled when calibnet goes NV17, as we lack proper migrations.

lemmih:
Extra bonus points if you do it in the code coverage action and include the results in the coverage report.

LesnyRumcajs:
No bonus points - the code coverage action with calibnet check would take a long time (at least 40 minutes on release mode, but then we'd get a lot of false negatives because of optimized code).

PR #2042, opened 2022-10-11
Calibnet check gh action

LesnyRumcajs:
**Summary of changes**
- add calibnet sync check to a PR,
- move the code coverage step to push to `main` only (we don't need it for PRs)

Closes https://github.com/ChainSafe/forest/issues/2038


lemmih (approved):
Looks good.

I keep seeing `--target-peer-count 50` in our documentation and scripts. I wonder why the default of 75 peers isn't good enough. Should we change the default to 50? Or have different defaults for calibnet and mainnet?

LesnyRumcajs:
> Looks good.
> 
> I keep seeing `--target-peer-count 50` in our documentation and scripts. I wonder why the default of 75 peers isn't good enough. Should we change the default to 50? Or have different defaults for calibnet and mainnet?

I'm not sure, I put it everywhere because it was everywhere so now it's even more everywhere. 🤣 It's definitely worth investigating further, good point.

tyshko5 (approved):

PR #2047, opened 2022-10-12
Tipset and TipsetKeys property tests

connormullett:
**Summary of changes**
- Add quickcheck tests for tipset and tipsetkeys using quickcheck

Closes 
part of #1881, affected tasks have been marked complete


blockchain/blocks/src/tipset.rs
```diff
@@ -47,6 +47,15 @@ impl PartialEq for Tipset {
     }
 }
 
+impl quickcheck::Arbitrary for TipsetKeys {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let header = BlockHeader::arbitrary(g);
+        Self {
+            cids: vec![*header.cid()],
```
lemmih, line 54: You don't have to create a block header just to get a CID. Instead, just generate a vector of random CIDs.

blockchain/blocks/src/tipset.rs
```diff
@@ -543,3 +552,27 @@ mod test {
...
+        let serialized = serde_json::to_string(&TipsetKeysJson(tipset_keys.clone())).unwrap();
+        let parsed: TipsetKeysJson = serde_json::from_str(&serialized).unwrap();
+        assert_eq!(tipset_keys, parsed.0);
+    }
+
+    #[quickcheck]
+    fn tipset_roundtrip(tipset: Tipset) {
+        let serialized = serde_json::to_string(&TipsetJsonRef(&tipset)).unwrap();
+        let parsed: TipsetJson = serde_json::from_str(&serialized).unwrap();
+        assert_eq!(Arc::new(tipset), parsed.0);
```
lemmih, line 576:
```suggestion
        assert_eq!(&tipset, parsed.0.as_ref());
```

lemmih (changes requested):

blockchain/blocks/src/tipset.rs
connormullett: 👍

blockchain/blocks/src/tipset.rs
```diff
@@ -47,6 +47,16 @@ impl PartialEq for Tipset {
     }
 }
 
+impl quickcheck::Arbitrary for TipsetKeys {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let cid = Cid::new_v1(
+            u64::arbitrary(g),
+            cid::multihash::Multihash::wrap(u64::arbitrary(g), &[u8::arbitrary(g)]).unwrap(),
+        );
+        Self { cids: vec![cid] }
```
lemmih, line 56: We also want to test what happens when you have lots of keys or zero keys.

lemmih (changes requested):

blockchain/blocks/src/tipset.rs
lemmih: (You could implement this by creating a wrapper around `Cid`)

connormullett: Shouldn't that be part of a normal unit test? As far as I'm aware, these tests are just for checking if serialization/deserialization works both ways.

lemmih: I'm not sure what you mean. As the code is right now, you're just testing if serialization/de-serialization works when `TipsetKeys` contains a single `Cid`. We also want to know if it works when there are zero Cids or many Cids.

lemmih (approved):

jdjaustin (commented):
Probably nit, but @LesnyRumcajs has been asking me to put the `struct` and `impl` blocks under each property test module (i.e., so there is only one `#[cfg(test)]` for all the property test code), and the code is definitely a lot cleaner that way.

LesnyRumcajs:
Good point @jdjaustin! :)

connormullett:
> Probably nit, but @LesnyRumcajs has been asking me to put the `struct` and `impl` blocks under each property test module (i.e., so there is only one `#[cfg(test)]` for all the property test code), and the code is definitely a lot cleaner that way.

Not a nitpick at all, this makes sense. Fixing now

jdjaustin (approved):

PR #2056, opened 2022-10-13
chore: upgrade github action versions

hanabi1224:
**Summary of changes**
- add `GitHub Actions Version Updater` workflow
- upgrade github action versions to suppress [warnings](https://github.com/ChainSafe/forest/actions/runs/3243847226) like
```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, Swatinem/rust-cache, actions-rs/cargo, Swatinem/rust-cache, actions/checkout
```

@lemmih Do we want to have sth like [this](https://github.com/saadmk11/github-actions-version-updater) to automatically upgrade github action versions? It requires a github token with workflow scope tho.

Closes https://github.com/ChainSafe/forest/issues/2057

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,11 +16,11 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y build-essential clang ocl-icd-opencl-dev
       - name: Checkout Sources
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: true
       - name: Rust Cache
-        uses: Swatinem/rust-cache@v1.3.0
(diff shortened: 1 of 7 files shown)


lemmih:
> Do we want to have sth like [this](https://github.com/saadmk11/github-actions-version-updater) to automatically upgrade github action versions? It requires a github token with workflow scope tho.

Yes, that would be nice. You can use the `ACTIONS_PAT` token. It has the workflow scope.

LesnyRumcajs (approved):

lemmih (approved):

PR #2061, opened 2022-10-14
slower buildjet machines for workflows

LesnyRumcajs:
**Summary of changes**
- Use cheaper buildjet machines for workflows
- entire Rust workflow stays at roughly at 21 minutes:
https://github.com/ChainSafe/forest/actions/runs/3245363626
https://github.com/ChainSafe/forest/actions/runs/3248401717

- Docker build is 50% slower but 2x cheaper, so we still gain 50%. Given that it still takes less than the Rust workflow, we should be fine.
- code coverage workflow will most likely take longer, but it happens only on merge to `main` so we don't care about the time (unless it's so slow that it times out or takes >2x longer, I'll revert the change for this if it's the case)

Closes 

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   codecov:
     name: Cover
-    runs-on: buildjet-16vcpu-ubuntu-2004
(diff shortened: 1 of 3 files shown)


lemmih (approved):
Looks good.

hanabi1224 (approved):

PR #2066, opened 2022-10-14
use shorter cache timeout

LesnyRumcajs:
**Summary of changes**
- as @lemmih suggested, if the cache action doesn't complete within 2-3 minutes then it's most likely stalled and we would timeout anyway.
- continue the job even if the cache failed for whatever reason. If it happens, we will just rebuild.

Closes 

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,6 +31,7 @@ jobs:
           fail: true
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+        timeout-minutes: 5
       - name: Install MDBook
         run: cargo install mdbook mdbook-linkcheck
       - name: Execute MDBook
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,6 +32,7 @@ jobs:
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
         timeout-minutes: 5
+        continue-on-error: true
       - name: Install MDBook
         run: cargo install mdbook mdbook-linkcheck
       - name: Execute MDBook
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -30,6 +31,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+        timeout-minutes: $CACHE_TIMEOUT_MINUTES
```
hanabi1224, line 34: I guess this has to be ${{ env.CACHE_TIMEOUT_MINUTES }}

.github/workflows/docs.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           fail: true
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+        timeout-minutes: 5
```
lemmih, line 34: `continue-on-error` here?

.github/workflows/rust.yml
LesnyRumcajs: It was a bit more complex, this piece of technology is somehow typed so there's a need to `fromJSON` https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson
> Returns a JSON object or JSON data type for value. You can use this function to provide a JSON object as an evaluated expression or to convert environment variables from a string.

Thanks for looking into it anyway!

.github/workflows/docs.yml
LesnyRumcajs: good catch, I forgot to save the file :crying_cat_face:

hanabi1224 (approved):

lemmih (approved):

lemmih:
Hurrah for @LeshyBot!

PR #2073, opened 2022-10-15
Switch to cargo nextest as test runner

### LINKED ISSUE #2040: Investigate using nextest for test runs
**Issue summary**

There is an interesting tool - [cargo nextest](https://nexte.st/index.html). So far it looks promising; the test time is cut from 129 to 82s, and the report is more readable and seems nicely customizable. On top of this, it highlights slow tests.

image

Seems like a drop-in replacement but there may be some caveats. The goal here is to investigate this subject more, try to implement it and decide whether it's reasonable to run it instead of `cargo run`. 

creativcoder:
**Summary of changes**
- Changes to Makefile to use cargo nextest as test runner

Closes #2040 

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,6 +36,7 @@ jobs:
       - name: Cargo Check
         run: cargo check
       - name: Make Test-All
+        uses: taiki-e/install-action@nextest
         run: make test-all
 
   lint:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,6 +29,8 @@ jobs:
           sudo apt-get install -y build-essential clang ocl-icd-opencl-dev
       - name: Checkout Sources
         uses: actions/checkout@v3
+      - name: install nextest
+        uses: taiki-e/install-action@nextest
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
         timeout-minutes: ${{ fromJSON(env.CACHE_TIMEOUT_MINUTES) }}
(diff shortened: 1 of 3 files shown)


LesnyRumcajs (changes requested):
Could you also update the documentation?
https://github.com/chainsafe/forest/blob/creativcoder/faster-tests/README.md?plain=1#L139
https://github.com/chainsafe/forest/blob/creativcoder/faster-tests/README.md?plain=1#L148

creativcoder:
Updated README.md

.github/workflows/rust.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       - name: Cargo Check
         run: cargo check
       - name: Make Test-All
+        uses: taiki-e/install-action@nextest
```
hanabi1224, line 39: Combination of uses and run does not seem to be valid

creativcoder: I see, there's no error msg being reported nor docs to understand what's the issue is. So do I need to move all makefile test commands under the `run:` section instead?

creativcoder: according to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun i might just use run to install nextest and run make test

hanabi1224: If u look at the CI jobs, `rust.yml` did not run

LesnyRumcajs: @creativcoder Just put
```
- name: install nextest
   uses: taiki-e/install-action@nextest
```
above the cache step, leave this step as it was; it should work. To my understanding whether you can use `run` or not, depends on the plugin you are using. Here apparently it doesn't work.

creativcoder: got it thanks! it's running now

LesnyRumcajs (approved):

lemmih (approved):

Issue #2034, opened 2022-10-10
Use the gas tracker in `fvm` rather than `GasBlockStore`


lemmih:
**Issue summary**

- [x] easy: `GasCharge` should be replaced by `fvm::gas::GasCharge`.
- [x] hard: `GasBlockStore` should be replaced by `fvm::gas::GasTracker`. This is not a one-to-one replacement. Instead, we need to somehow use the `GasTracker` in the `DefaultCallManager`.

**Other information and links**

https://github.com/ChainSafe/forest/blob/75958ae2fcb86d23ab717c7a19f2dfdcd7744205/vm/interpreter/src/gas_tracker/gas_charge.rs#L8-L12
https://github.com/ChainSafe/forest/blob/75958ae2fcb86d23ab717c7a19f2dfdcd7744205/vm/interpreter/src/gas_block_store.rs#L13-L17

lemmih:
@hanabi1224 This is not a trivial issue but it has been a thorn in our sides for a while now.

PR #2079, opened 2022-10-17
chore: refactor to remove GasBlockStore

hanabi1224:
**Summary of changes**
- Replace `GasBlockStore` with `TrackingBlockStore`

Closes https://github.com/ChainSafe/forest/issues/2034


lemmih:
Let's get rid of the `BlockStore` trait. It doesn't seem to be useful anymore.

hanabi1224:
> Let's get rid of the BlockStore trait. It doesn't seem to be useful anymore.

Sure, would u suggest doing that in the same PR or a separate one?

hanabi1224:
Also I think `ForestKernel` is not needed either, it's not instantiated anywhere in either forest for fvm code

vm/interpreter/src/fvm/externs.rs
```diff
@@ -231,3 +222,89 @@ impl<DB: BlockStore> Consensus for ForestExterns<DB> {
...
+        write_bytes: &[usize],
+    ) -> anyhow::Result<()> {
+        let network_version = NetworkVersion::V8;
+        let stats = BSStats {
+            r: read_count,
+            w: write_bytes.len(),
+            br: 0, // Not used in current logic
+            bw: write_bytes.iter().sum(),
+        };
+        println!("stats: {:?}", stats);
```
LesnyRumcajs, line 291: do we need this print? I understand it's for debugging purposes only?

vm/interpreter/src/fvm/externs.rs
```diff
@@ -231,3 +222,89 @@ impl<DB: BlockStore> Consensus for ForestExterns<DB> {
         Ok((cf, total_gas))
     }
 }
+
+fn cal_gas_used_from_stats(
```
LesnyRumcajs, line 226: It'd be nice to add some piece of documentation here. I don't get from a quick glimpse what is `r` and `w`, other than poorly named fields. We can't change them, but maybe we can at least explain stuff for those less bright like myself. :)

LesnyRumcajs (changes requested):

vm/interpreter/src/fvm/externs.rs
hanabi1224: fixed

vm/interpreter/src/fvm/externs.rs
hanabi1224: fixed

LesnyRumcajs (approved):
LGTM. I'm okay with addressing removing `BlockStore` trait either here or in a separate PR.

lemmih (approved):
Looking good. We can remove the unused trait and structs in a separate PR.

PR #2093, opened 2022-10-19
Prep work for NV17 (Shark? Candy?)

lemmih:
**Summary of changes**
- 

Closes 

### DIFF
--- a/networks/src/calibnet/mod.rs
+++ b/networks/src/calibnet/mod.rs
@@ -17,7 +17,7 @@ pub const DEFAULT_BOOTSTRAP: &[&str] = &[
 ];
 
 /// Height epochs.
-pub const HEIGHT_INFOS: [HeightInfo; 17] = [
+pub const HEIGHT_INFOS: [HeightInfo; 18] = [
     HeightInfo {
         height: Height::Breeze,
         epoch: -1,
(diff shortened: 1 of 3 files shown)


LesnyRumcajs (approved):
LGTM

hanabi1224 (approved):

.github/workflows/rust.yml
```diff
@@ -129,6 +129,7 @@ jobs:
 
   calibnet-check:
     name: Calibnet sync check
+    if: ${{ false }}
```
hanabi1224, line 132: Do we want to disable snapshot fetch and import test as well?

LesnyRumcajs: Perhaps it's worth leaving those steps enabled, good point. Not ideal but at least something would be covered.

lemmih: Where are those tests?

LesnyRumcajs: https://github.com/ChainSafe/forest/pull/2093/files#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR155-R163

LesnyRumcajs: I'm not sure the link works. In general, commenting out everything below `forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot /tmp/snapshots/*.car` because up until this point it should still pass, right? And at least we would be able to detect import/fetch fiascoes.

lemmih: Not sure that links for me for. What is it supposed to show?

PR #2095, opened 2022-10-19
Add call for raising fd limit

### LINKED ISSUE #2090: handle low FD limit
**Issue summary**

Given our current RocksDB backend, Forest needs to have opened many files. This is governed via `max_open_files` in RocksDB config.
https://github.com/ChainSafe/forest/blob/927837df2451d2a0aba79c638efb47b5b45df300/node/db/src/rocks_config.rs#L19

RocksDB is not the only part of Forest that opens files, so if the default RocksDB limit is 1024, your system FD limit is 1024 (`ulimit -n`), then one more file opened will cause Forest to fail miserably with a message pointing to exhausted file descriptors. 

The goal here would be to check, at the start of the program, the FD limit and either increase or issue an explicit warning (or even fail with an error, up to the engineer implementing this). I'd prefer the first approach with https://github.com/paritytech/fdlimit, but either of those options would be better than what we have right now.

elmattic:
**Summary of changes**
- Check current fd limit using `rlimit` crate
- Compute an estimate and errors if we are under it

Closes #2090


lemmih:
Maybe a warning if the limit is lower or close to the open file limit in rocksdb?

elmattic:
> Maybe a warning if the limit is lower or close to the open file limit in rocksdb?

So you prefer a warning and not raising automagically the limit?

forest/daemon/src/main.rs
```diff
@@ -92,6 +93,13 @@ fn build_daemon<'a>(config: &DaemonConfig) -> anyhow::Result<Daemon<'a>> {
     Ok(daemon)
 }
 
+// Protect user from low FD for its own good
+fn protect_from_low_fd() {
+    if let Some(curr) = raise_fd_limit() {
+        info!("Open-file limit raised to {curr}");
+    }
+}
+
```
LesnyRumcajs, line 102: So if I understand correctly, this will **always** be logged. There's even no point in doing `if let` as this method never returns an `Err` https://docs.rs/fdlimit/latest/src/fdlimit/lib.rs.html#88-110. 

Maybe let's do this only if e.g. soft limit is smaller/equal to rocksdb config?

elmattic: On non-unix OS (albeit we don't support Windows), `None` is returned.

elmattic: Will try to find some heuristic on how much we need FDs and raise only the limit if we can't satisfy it.

lemmih:
> > Maybe a warning if the limit is lower or close to the open file limit in rocksdb?
> 
> So you prefer a warning and not raising automagically the limit?

Can we do both?

elmattic:
> Can we do both?

Sure.

elmattic:
> I'm leaning towards using the `rlimit` crate.

So you mean something like always raising the limit before starting the node, ie:

```
if let Some(curr) = raise_fd_limit() {
    warn!("Open-file limit raised to {curr}");
}
```

elmattic:
Putting the PR back to draft because I'm seeing some flaw (cast of a negative integer `max_open_files` that will create a broken estimate).

elmattic:
Any feedback on this PR? I'm more inclined towards doing something simple like using `rlimit` crate and always raising the limit (with some `info` log). Any downsides to this?

Possibly we can provide a flag to disable the auto-raise of the FDs, if the node operator wants to configure this himself to a given limit.

creativcoder:
I'm not totally sure about this but just my observation with other daemon processes in linux: Usually modifying file limits entails affecting resource utilisation by the process. It's something that's reported by processes as an error (the processes themselves don't decide it raise it) and it is up to the user to decide how to raise the limit (either they might set a global raise limit per user group or some other policy) because their machine might have resource constraints or other reasons. 

Maybe we can just suggest the raised fd number and halt, or be explicit on raising by providing a flag to user.

elmattic:
> I'm not totally sure about this but just my observation with other daemon processes in linux: Usually modifying file limits entails affecting resource utilisation by the process. It's something that's reported by processes as an error (the processes themselves don't decide to raise it) and it is up to the user to decide how to raise the limit (either they might set a global raise limit per user group or some other policy) because their machine might have resource constraints or other reasons.
> 
> Maybe we can just suggest the raised fd number and halt, or be explicit on raising by providing a flag to user.

Agree, as a remainder the main point of this PR would be to avoid starting a Forest node and 15 minutes later seeing it fail because of the "Too many open files" error (ie, user has rebooted its computer and forgot to set `ulimit -n` to a proper value).

So I'm for the "suggest and halt if lower" proposal.

creativcoder (approved):

PR #2097, opened 2022-10-19
Fix logger setup and use anyhow error

### LINKED ISSUE #2078: Improve toml config parsing
**Describe the bug**
If Forest configuration file has type errors, this will ~~result in a hard crash~~ be silented. Also a non-existing key is a no-op.

Ie, this section:

```
[rocks_db]
create_if_missing = true
parallelism = 12
write_buffer_size = 268435456
max_open_files = 1024
compression_type = "lz4"
compaction_style = false    # <-- Value should be a string not a boolean!
invoke_outer_gods = true    # <-- 😱 
```

**To Reproduce**
1. Dump config in config.toml file
2. Patch `rocks_db` section with the one above
3. Run with: `./target/release/forest --target-peer-count 50 --encrypt-keystore false --import-snapshot minimal_finality_stateroots_2215681_2022-10-03_06-00-30.car --halt-after-import -c config.toml  `
4. Observe crash

**Expected behaviour**

We should have a proper error, indicating to the user where the problem is.

Maybe also warn the user for the non-existing keys (in case we retired some or user did typos).

**Environment:**
 - Branch: main

n/a

elmattic:
**Summary of changes**
- Fix logger setup and move to anyhow error
- ?

Closes #2078

### DIFF
--- a/forest/shared/src/cli/mod.rs
+++ b/forest/shared/src/cli/mod.rs
@@ -12,7 +12,6 @@ use forest_utils::io::{read_file_to_string, read_toml};
 use git_version::git_version;
 use log::{error, info, warn};
 use once_cell::sync::Lazy;
-use std::io::{self};
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -97,7 +96,7 @@ pub struct CliOpts {
 }
 
 impl CliOpts {
-    pub fn to_config(&self) -> Result<Config, io::Error> {
+    pub fn to_config(&self) -> Result<Config, anyhow::Error> {
         let mut cfg: Config = match &self.config {
             Some(config_file) => {
                 // Read from config file
@@ -132,10 +131,7 @@ impl CliOpts {
             cfg.client.metrics_address = metrics_address;
         }
         if self.import_snapshot.is_some() && self.import_chain.is_some() {
-            cli_error_and_die(
-                "Can't set import_snapshot and import_chain at the same time!",
-                1,
-            );
+            anyhow::bail!("Can't set import_snapshot and import_chain at the same time!")
         } else {
             if let Some(snapshot_path) = &self.import_snapshot {
                 cfg.client.snapshot_path = Some(snapshot_path.to_owned());
--- a/forest/daemon/src/main.rs
+++ b/forest/daemon/src/main.rs
@@ -135,6 +135,8 @@ fn main() {
             }
         }
         Err(e) => {
+            // We need to setup logger before using `cli_error_and_die`
+            logger::setup_logger(&[]);
             cli_error_and_die(format!("Error parsing config. Error was: {e}"), 1);
         }
     };


lemmih:
We also use the logger in `CliOpts::to_config` before the logger is initialized (which means those log messages are ignored). Maybe you could address that at the same time.

elmattic:
> We also use the logger in `CliOpts::to_config` before the logger is initialized (which means those log messages are ignored). Maybe you could address that at the same time.

Ah yes! In `find_default_config`, good catch.

forest/shared/src/logger/mod.rs
```diff
@@ -12,8 +12,10 @@ pub fn setup_logger(log_config: &[LogValue]) {
     logger_builder.filter(None, LevelFilter::Info);
 
     for item in log_config {
-        let level = LevelFilter::from_str(item.level.as_str())
-            .unwrap_or_else(|_| cli_error_and_die("could not parse LevelFilter enum value", 1));
+        let level = LevelFilter::from_str(item.level.as_str()).unwrap_or_else(|_| {
```
LesnyRumcajs, line 15: It would be helpful to also print the module that happened to have this faulty level. A rock-solid approach would be to implement `Display` for `LogValue` and then just print the entire item. :)

forest/shared/src/cli/mod.rs
```diff
@@ -171,53 +169,76 @@ impl CliOpts {
...
-            warn!(
-                "Error reading configuration file, opting to defaults. Error was {} ",
-                e
-            );
-            None
+        _ => (),
+    }
+}
+
+pub fn warn_for_unknown_keys(path: Option<String>, config: &Config) {
```
LesnyRumcajs, line 226: There is a lot of unwraps here, can we handle the errors a bit nicer?

forest/shared/src/cli/mod.rs
```diff
@@ -171,53 +169,76 @@ impl CliOpts {
...
     None
 }
 
-fn read_config_or_none(path: PathBuf) -> Option<Config> {
-    let toml = match read_file_to_string(&path) {
-        Ok(t) => t,
-        Err(e) => {
-            warn!("An error occured while reading configuration file at {}. Resorting to default configuration. Error was {}", path.display(), e);
-            return None;
+fn get_keys_flatten(value: toml::Value, keys: &mut HashSet<String>) {
```
LesnyRumcajs, line 209: I see an excellent opportunity for some unit tests, especially given that there's some recursion here.

forest/shared/src/cli/mod.rs
```diff
@@ -171,53 +169,76 @@ impl CliOpts {
...
+        let value = file.parse::<toml::Value>().unwrap();
+        let mut keys = HashSet::new();
+        get_keys_flatten(value, &mut keys);
+
+        let config_file = toml::to_string(config).unwrap();
+        let config_value = config_file.parse::<toml::Value>().unwrap();
+        let mut config_keys = HashSet::new();
+        get_keys_flatten(config_value, &mut config_keys);
+
+        for k in keys.iter() {
```
LesnyRumcajs, line 238: I see an **excellent** opportunity for https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference

forest/shared/src/logger/mod.rs
elmattic: Agree.

forest/shared/src/cli/mod.rs
elmattic: I've added some comment to explain why unwrap works here.

forest/shared/src/cli/mod.rs
```diff
@@ -171,60 +169,112 @@ impl CliOpts {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn get_keys_flatten_must_work() {
+        let value = toml::from_str(
+            r#"
```
LesnyRumcajs, line 257: I like it 🐙

forest/shared/src/cli/mod.rs
```diff
@@ -171,60 +169,112 @@ impl CliOpts {
...
+                { name = 'Cthulhu' },
+                { name = 'Azathoth' },
+                { name = 'Dagon' },
+            ]
+        "#,
+        )
+        .unwrap();
+
+        let mut keys = HashSet::new();
+        get_keys_flatten(value, &mut keys);
```
LesnyRumcajs, line 271: Perhaps it'd be more ergonomic to have a separate method that would call `get_keys_flatten` but itself would have a signature `fn fancy_name(value: toml::Value) -> HashSet<String>`. It'd just create the `HashSet` and call the recursive guy, but we wouldn't need to provide a new `HashSet` in higher-level methods. Not a big deal though.

forest/shared/src/cli/mod.rs
```diff
@@ -171,60 +169,112 @@ impl CliOpts {
...
+
+        let expected = HashSet::from_iter(
+            ["folklore", "myth", "author", "entities", "name"]
```
LesnyRumcajs, line 274: Hm, is `myth` a key? Or in other terms, will the `warn_for_unknown_keys` warn about config such as:
```
[data_dir]
client = "blabla"
```
(section trying to be a key)

What I have in mind, is the keys should be "paired" with their respective sections, so:
```
[client]
data_dir = "blabla"
```
would be a single key `client.data_dir`. This would also avoid keys in invalid sections, i.e. 
```
[client]
write_buffer_size = 42 # this should be in `rocks_db` section
```

I'm not sure it makes sense or perhaps you already covered this?

elmattic: Yeah, this poor-man's diff is not 100% accurate as you can see. I can make it better.

tyshko5 (approved):

forest/shared/src/cli/mod.rs
```diff
@@ -176,60 +174,150 @@ impl CliOpts {
...
+
+    let mut result = vec![];
+    find_unknown_keys(None, &value, &config_value, &mut result);
+    for (table, k) in result.iter() {
+        let path = path.display();
+        if let Some(table) = table {
+            warn!("{path}: Unknown key `{k}` in [{table}], ignoring");
+        } else {
+            warn!("{path}: Unknown key `{k}` in top-level table, ignoring");
         }
```
LesnyRumcajs, line 259: The message could be perhaps more precise, currently if we e.g. put a random new key in log config, i.e.
```
[[log.filters]]
module = "libp2p_bitswap"
level = "WARN"
cthulhu = "yes"
```
we only get `Unknown key `cthulhu` in [filters], ignoring`. This should rather be `[log.filters]`.

forest/shared/src/cli/mod.rs
```diff
@@ -176,60 +174,150 @@ impl CliOpts {
...
+    let config_value = config_file.parse::<toml::Value>().unwrap();
+
+    let mut result = vec![];
+    find_unknown_keys(None, &value, &config_value, &mut result);
+    for (table, k) in result.iter() {
+        let path = path.display();
+        if let Some(table) = table {
+            warn!("{path}: Unknown key `{k}` in [{table}], ignoring");
+        } else {
+            warn!("{path}: Unknown key `{k}` in top-level table, ignoring");
```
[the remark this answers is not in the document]
elmattic: Yes this would be way more safe.

forest/shared/src/cli/mod.rs
elmattic: Good idea!

creativcoder (approved):

LesnyRumcajs (approved):
rock-solid

PR #2089, opened 2022-10-18
chore: remove libnghttp2-sys from build graph

hanabi1224:
**Summary of changes**
Changes introduced in this pull request:
- Removes `libnghttp2-sys` from build graph by tweaking dependency feature flags

`libnghttp2-sys` ranks No.3 in build times

top 10 from build local build:

|-| Unit | Total | CodeGen | Features |
| --- | :-- | --: | :-: | :-: |
|1. | libsodium-sys v0.2.7 build script (run) | 123.5s |   |  
|2. | libz-sys v1.1.8 build script (run) | 90.1s |   | libc, static |
|3. | libnghttp2-sys v0.1.7+1.45.0 build script (run) | 78.0s |   |  
|4. | sha2-asm v0.6.2 build script (run) | 67.5s |   |  
|5. | zstd-sys v2.0.1+zstd.1.5.2 build script (run) | 55.0s |   | default, legacy, zdict_builder
|6. | wasmtime-runtime v0.37.0 build script (run) | 51.7s |   | pooling-allocator
|7. | bzip2-sys v0.1.11+1.0.8 build script (run) | 49.0s |   | static
|8. | blst v0.3.7 build script (run) | 45.8s |   | default
|9. | cranelift-codegen v0.84.0 | 37.1s | 16.4s (44%) | default, gimli, std, unwind
|10. | librocksdb-sys v0.8.0+7.4.4 build script (run) | 36.8s |   | bzip2, bzip2-sys, default, libz-sys, lz4, snappy, static, zlib, zstd, zstd-sys

**Reference issue to close (if applicable)**

Closes part of https://github.com/ChainSafe/forest/issues/2088

**Other information and links**

PR #2102, opened 2022-10-19
fix: clear timeout in surf http client

hanabi1224:
**Summary of changes**
- clear timeout in surf http client. The default value 60s is too small for tasks like `snapshot export`. The timeout does not apply to curl client atm because of a bug in `http-client` crate
- redo https://github.com/ChainSafe/forest/pull/2089 to remove `libnghttp2-sys` from build graph
- manually verified `snapshot export` command. TODO: automation test

Closes 


lemmih (approved):

LesnyRumcajs (approved):
Excellent, thanks for the root cause investigation!

Issue #2088, opened 2022-10-18
Investigate which dependencies can be made optional


lemmih:
**Issue summary**

Right now we have several dependencies that take a long time to build and require certain system libraries to be installed. It would be better if such dependencies were optional.

These dependencies might be optional:
- [ ] `opencl` in `fvm`. This dependency requires `ocl-icd-opencl-dev` to be installed.
- [x] `dag-pb` in `libipld`. This dependency requires `cmake` to be installed.
- [x] `snappy`, `zstd`, `zlib`, and `bzip2` in `rocksdb`. These dependencies are among the slowest to compile.

We also depend on `libcurl4`. It would be nice to avoid this and use a pure Rust alternative.

**Other information and links**

hanabi1224:
`libcurl-sys` and underlying `libnghttp2-sys` ranks the 3rd in build times, it should be easy to replace `isahc` with sth like `reqwest`

lemmih:
@connormullett Feel free to try your hand at one of these. If there's an unfixable reason for why, say, the fvm cannot be used without `opencl` then create a new issue marked as blocked with an explanation of what needs to happen for it to be unblocked.

hanabi1224:
We should be able to remove `isahc`, `h2`, `curl` and `openssl` deps with https://github.com/ChainSafe/forest/pull/2140 and https://github.com/ChainSafe/forest/pull/2141

lemmih:
@tyshko5 You can also work on this issue.

lemmih:
I doubt we can get rid of `opencl`. Closing this issue as done.

PR #2103, opened 2022-10-19
chore: do not build libzstd-sys by default

hanabi1224:
**Summary of changes**
- make `rocksdb` compression types optional
- exclude `rocksdb/zstd` from default features to reduce build time cuz it's one of the slowest, see table in https://github.com/ChainSafe/forest/pull/2089
- improve tests

Closes part of https://github.com/ChainSafe/forest/issues/2088

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2666,7 +2666,6 @@ dependencies = [
  "forest_chain_sync",
  "forest_cli_shared",
  "forest_crypto",
- "forest_db",
  "forest_encoding",
  "forest_fil_types",
  "forest_interpreter",
@@ -3248,7 +3247,6 @@ dependencies = [
  "forest_actor_interface",
  "forest_blocks",
  "forest_crypto",
- "forest_db",
  "forest_encoding",
  "forest_fil_types",
  "forest_ipld_blockstore",
@@ -3807,9 +3805,9 @@ checksum = "1847abb9cb65d566acd5942e94aea9c8f547ad02c98e1649326fc0e8910b8b1e"
 
 [[package]]
 name = "futures"
-version = "0.3.24"
+version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f21eda599937fba36daeb58a22e8f5cee2d14c4a17b5b7739c7c8e5e3b8230c"
+checksum = "38390104763dc37a5145a53c29c63c1290b5d316d6086ec32c293f6736051bb0"
 dependencies = [
  "futures-channel",
  "futures-core",
@@ -3822,9 +3820,9 @@ dependencies = [
 
 [[package]]
 name = "futures-channel"
-version = "0.3.24"
+version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "30bdd20c28fadd505d0fd6712cdfcb0d4b5648baf45faef7f852afb2399bb050"
+checksum = "52ba265a92256105f45b719605a571ffe2d1f0fea3807304b522c1d778f79eed"
 dependencies = [
  "futures-core",
  "futures-sink",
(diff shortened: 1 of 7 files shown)


lemmih:
I'm not sure the `rocksdb/io-uring` feature is stable enough that we should include it.

lemmih:
> I'm not sure the `rocksdb/io-uring` feature is stable enough that we should include it.

Or maybe it would be simple enough to install liburing on the CI servers.

lemmih (approved):
Lovely. It's tempting to go even further and disable `snappy`, `zlib` and `bzip2` by default.

LesnyRumcajs (commented):
I feel we should have some piece of documentation that complements this. Otherwise, as a user, I'd be quite let down if some of the configuration options aren't working and I only browsing the code will help. See e.g. https://docs.rs/tokio/latest/tokio/#feature-flags 

What do you think?

hanabi1224:
That makes sense for crate doc. A application I assume current compress type options are for our own convenience to measure performance, imho it should eventually be some more user friendly option like --compression [yes|no], does that make sense?

elmattic:
> That makes sense for crate doc. As for application I assume current compress type options are for our own convenience to measure performance, imho it should eventually be some more user friendly option like --compression [yes|no], does that make sense?


If we remove some compression libs, then we should remove some options here as well: https://github.com/ChainSafe/forest/blob/main/node/db/src/rocks_config.rs#L61

hanabi1224:
@elmattic this should have been addressed by `#[cfg(feature = "...")]` lines, did I miss sth?

hanabi1224:
@LesnyRumcajs 
Improved error message to list all available options:
```console
invalid compression option: lz44, valid options: bz2,lz4,lz4hc,snappy,zlib,none', node/db/src/rocks.rs:50:90
```

It's complicated to make them app level feature flags cuz we need to propagate those features all the way through the dependency chain, might not be worth it or we can do it in a separate PR.

LesnyRumcajs (approved):
It'd be nice to propagate those flags, otherwise, they're painful to utilise (i.e. one would need to edit manually some Cargo file IIUC). I'm okay with doing this in a separate PR or disregarding this comment if it doesn't make sense. :)

elmattic:
> @elmattic this should have been addressed by `#[cfg(feature = "...")]` lines, did I miss sth?

Try to run forest like this:

Dump config first in `config.toml` and change compression with `compression_type = "zstd"`.

```
$ ./target/debug/forest --target-peer-count 50 --encrypt-keystore false -c config.toml 
2022-10-20T10:25:15.228Z INFO  forest::daemon > Starting Forest daemon, version 0.4.1+git.779293c1
2022-10-20T10:25:15.230Z INFO  forest_libp2p::service > Recovered libp2p keypair from "/Users/guillaume/Library/Application Support/com.ChainSafe.Forest/libp2p/keypair"
2022-10-20T10:25:15.230Z INFO  forest::daemon         > PeerId: 12D3KooWPvXJznug7GH2PHN7ov7YMhY3LDrphzR6SXdvZYEPJjPh
2022-10-20T10:25:15.232Z WARN  forest::daemon         > Warning: Keystore encryption disabled!
2022-10-20T10:25:15.233Z INFO  forest::daemon         > Prometheus server started at 127.0.0.1:6116
2022-10-20T10:25:15.234Z INFO  forest::daemon         > Admin token: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.FfJKVLU-eF7DpIffgFixTmpPEC1XB68M-jmrTGC6qPM
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: invalid compression option: zstd, valid options: bz2,lz4,lz4hc,snappy,zlib,none', node/db/src/rocks.rs:50:90
```

elmattic:
> @LesnyRumcajs Improved error message to list all available options:
> 
> ```
> invalid compression option: lz44, valid options: bz2,lz4,lz4hc,snappy,zlib,none', node/db/src/rocks.rs:50:90
> ```
> 
> It's complicated to make them app level feature flags cuz we need to propagate those features all the way through the dependency chain, might not be worth it or we can do it in a separate PR.

Maybe also remove the `unwrap` and propate the error in `RocksDb::to_options` would be cleaner.

hanabi1224:
@elmattic `zstd` is disabled by this PR because it builds slow, and the error message says
```console
invalid compression option: zstd, valid options: bz2,lz4,lz4hc,snappy,zlib,none'
```

I didn't get your suggestion abt `remove some options here as well`, could u plz elaborate?

hanabi1224:
> It'd be nice to propagate those flags, otherwise, they're painful to utilise

@LesnyRumcajs It also adds non-trival maintainence burden to add those features, I think it will be better that we discuss and have a conclusion of whether we should expose these compression types to end users after we finish tweaking the db setup

elmattic:
> @elmattic `zstd` is disabled by this PR because it builds slow, and the error message says
> 
> ```
> invalid compression option: zstd, valid options: bz2,lz4,lz4hc,snappy,zlib,none'
> ```
> 
> I didn't get your suggestion abt `remove some options here as well`, could u plz elaborate?

Sorry, didn't saw the commit 779293c14b9b02a9f5c4746ca777d1c198e72c89, so it's all good.

PR #2104, opened 2022-10-20
chore: custom cargo config to reduce build time

hanabi1224:
**Summary of changes**
- Add workspace cargo config `.cargo/config.toml` to improve build performance, for details checkout https://jondot.medium.com/8-steps-for-troubleshooting-your-rust-build-times-2ffc965fd13e
- [CI] Replace `rust-cache` with `sccache-action`, also change the cache strategy to write on push to `main` branch only.
- Merge all lint jobs into one to make better use the the cache and save build resource.
- strip symbols from `cargo install`ed tools (e.g. spellcheck from ~200MB to ~30MB)

rust-cache saves >8GB cache on every push to `main` and PR branch, while we only have 10GB total cache size, which makes cache unreliable when github automatically purges them

With the new strategy, PR commit will read the cache generated from the latest `main` commit but not saving them. (github caches are immutable, so saving means duplicating by appending some suffix to the cache key)

The cache size each `main` commit saves is now reduced to ~1-2GB.

Regarding build time, [here](https://github.com/ChainSafe/forest/actions/runs/3295257538) is an example run with the new strategy, looks on par or slightly faster than [before](https://github.com/ChainSafe/forest/actions/runs/3290409553)

The slowest udeps job takes ~15min in previous run, and all-in-one lint job takes ~9min with the new strategy. (They might run on different CPU tho, the hosted runner has a pool of CPU models)

Closes 


.github/workflows/rust.yml
```diff
@@ -23,139 +23,112 @@ jobs:
...
+        env:
+          CC: "sccache clang"
+          CXX: "sccache clang++"
       - name: download snapshot
         run: forest-cli --chain calibnet chain fetch -s /tmp/snapshots/
       - name: import snapshot and run Forest
         run: |
           forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot /tmp/snapshots/*.car
           forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --detach
-      - name: wait for sync and check health
```
hanabi1224, line 155: This stops working, disable for now?

LesnyRumcajs: @hanabi1224 you just need to :top: :fist_right: :heavy_check_mark: this guys https://github.com/ChainSafe/forest/pull/2093

LesnyRumcajs: ...and rebase. The reason is, calibnet got a version upgrade yesterday, and we are currently lagging (for various reasons), so we are unable to sync to calibnet at the moment.

.github/workflows/docs.yml
```diff
@@ -17,28 +17,36 @@ jobs:
...
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+          key: ${{ runner.os }}-docs-cargo-bin-${{ hashFiles('rust-toolchain.toml') }}
       - name: Install MDBook
         run: cargo install mdbook mdbook-linkcheck
+        continue-on-error: true
+      - name: Setup sccache
+        uses: hanabi1224/sccache-action@v1.2.0
```
LesnyRumcajs, line 39: it would be great to include a comment on why are we using a fork

.github/workflows/rust.yml
```diff
@@ -23,130 +23,102 @@ jobs:
...
-        uses: Swatinem/rust-cache@v2
-        timeout-minutes: ${{ fromJSON(env.CACHE_TIMEOUT_MINUTES) }}
-        continue-on-error: true
-      - name: Install udeps
-        run: cargo install cargo-udeps --locked
-      - name: Run udeps
-        run: make udeps
+      - run: make test-all
+        env:
+          CC: "sccache clang"
```
LesnyRumcajs, line 39: not sure if it wouldn't clash with some other steps, but would adding those envs as global for the entire workflow be fine?

LesnyRumcajs (approved):
This looks promising!

.github/workflows/rust.yml
hanabi1224: `sccache` is not always the first step to avoid generating cache for `cargo install [tools]`, so that `sccache` is not available during `cargo install` and `apt-get install`

.github/workflows/docs.yml
hanabi1224: Done. Basically we need the `cache-update` to only save the cache on 1. `main` branch, 2. cache not found. The original action repo looks like unmaintained so we keep using the fork

lemmih (approved):

hanabi1224:
Hey @lemmih  I think it requires manual changes to the branch rule of required CI checks before this PR can be merged. Could u help with that?

lemmih:
> Hey @lemmih I think it requires manual changes to the branch rule of required CI checks before this PR can be merged. Could u help with that?

Done.

PR #2105, opened 2022-10-20
Add --color=[always;auto;never] flag

### LINKED ISSUE #1512: Add `--color=never|always|auto` flag.
**Issue summary**

- [ ] Add `--color=never|always|auto` flag.
- [ ] Resolve `auto` to enabled if stdout is a tty and `NO_COLOR` hasn't been set (see https://no-color.org/)
- [ ] Update logger to follow the color preference.
- [ ] (Feel free to create a separate issue for this) Update progress bars to use newlines instead of carry returns if stdout is not a tty.

connormullett:
**Summary of changes**
- Add flag to determine ANSI colored output
- auto will resort to always if stdout is available, or never if not (such as forest being detached)

Closes #1512 


forest.err
```diff
@@ -0,0 +1,25 @@
+ 2022-10-21T16:26:35.867Z [0m[32mINFO [0m [0m[1mforest[0m > Process detached
```
LesnyRumcajs, line 1: Let's not make our errors public.

forest.out
```diff
@@ -0,0 +1,2 @@
...
Downloading headers 1015 / 2552 [=========>---------------] 39.77 % 343.87/s 4s 
Downloading headers 1120 / 2552 [==========>--------------] 43.89 % 322.06/s 4s 
Downloading headers 1329 / 2552 [=============>-----------] 52.08 % 328.69/s 4s 
Downloading headers 1541 / 2552 [===============>---------] 60.38 % 323.87/s 3s 
Downloading headers 1758 / 2552 [=================>-------] 68.89 % 321.03/s 2s 
Downloading headers 1979 / 2552 [===================>-----] 77.55 % 309.54/s 2s 
Downloading headers 2207 / 2552 [=====================>---] 86.48 % 304.16/s 1s 
Downloading headers 2315 / 2552 [======================>--] 90.71 % 296.26/s 1s 
Downloading headers 2546 / 2552 [=========================] 99.76 % 290.29/s 0s 
Downloading headers 2553 / 2552  100.04 % 291.08/s                              Could not resolve actor states: failed to find state tree bafy2bzacea7rndbjqtqhzkbg752gof67gp7takjyvfmmi6rqcnanxqikqxmom
```
LesnyRumcajs, line 1: And the output a mystery.

forest/shared/Cargo.toml
```diff
@@ -41,6 +41,7 @@ fvm_ipld_encoding.workspace      = true
 fvm_shared                       = { workspace = true, default-features = false }
 git-version                      = "0.3"
 hex.workspace                    = true
+isatty                           = "0.1.9"
```
LesnyRumcajs, line 44: the latest seem to be `0.2`, any particular reason behind `0.1`?

LesnyRumcajs (changes requested):

forest.err
connormullett: oh whoops, haha

forest/shared/Cargo.toml
connormullett: 0.2 is failing build. 1.9 is stable and works

creativcoder (approved):

forest/shared/src/cli/mod.rs
```diff
@@ -87,6 +87,9 @@ pub struct CliOpts {
     /// Daemonize Forest process
     #[structopt(long)]
     pub detach: bool,
+    /// Enable or disable colored logging in `stdout`
+    #[structopt(long, default_value = "auto", possible_values = &["always", "auto", "never"])]
+    pub color: String,
```
LesnyRumcajs, line 92: It would be more expressive to have the `color` as an enum that implements `FromStr`. Not a huge deal though.

forest/shared/src/logger/mod.rs
```diff
@@ -2,15 +2,37 @@
...
+        "auto" => {
+            if stdout_isatty() {
+                WriteStyle::Always
+            } else {
+                WriteStyle::Never
+            }
+        }
+        "never" => WriteStyle::Never,
+        // this should never happen
+        // structopt would error before this catchall happens
```
LesnyRumcajs, line 23: If this should never happen then perhaps, we should return an `Option` or a `Result` and handle the error and not silently ignore it? Or even unwrap? Or, use an enum, then no catchall would be needed.

forest/shared/src/cli/mod.rs
connormullett: I like this, would make it much cleaner

forest/shared/src/logger/mod.rs
connormullett: Going with the enum plan

connormullett:
Alright, cleaned it up with an enum and some trait impls. Looks much cleaner, thanks @LesnyRumcajs for the tip

forest/shared/src/logger/mod.rs
```diff
@@ -2,15 +2,62 @@
...
+
+impl FromStr for LoggingColor {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "auto" => Ok(LoggingColor::Auto),
+            "always" => Ok(LoggingColor::Always),
+            "never" => Ok(LoggingColor::Never),
+            _ => Ok(LoggingColor::Auto),
```
LesnyRumcajs, line 25: I'd expect `--color=cthulhu` to report an error not silently ignore the old gods. What do you think?

forest/shared/src/cli/mod.rs
```diff
@@ -87,6 +89,9 @@ pub struct CliOpts {
     /// Daemonize Forest process
     #[structopt(long)]
     pub detach: bool,
+    /// Enable or disable colored logging in `stdout`
+    #[structopt(long)]
+    pub color: Option<LoggingColor>,
```
LesnyRumcajs, line 94: It may be a bit easier and more expressive without `Option` and a default being `auto` (which would be nicely printed even in the `--help`)

connormullett: Sure, this also allows me to clean up that helper function

forest/shared/src/logger/mod.rs
connormullett: You're right, fixed. Reports an error and errors out immediately

Issue #2101, opened 2022-10-19
Improve snapshot write error handling


LesnyRumcajs:
**Issue summary**

This issue touches both `snapshot export` and `snapshot fetch`. At the moment, both exported and downloaded snapshots are saved immediately to the target location. This is a bit unfortunate because in case of an error or a panic they linger around, potentially causing various problems.

The goal here is to change the logic in both of the commands to write the files to a temporary location and only on a successful passthrough, move them to the target location. In case of an error (except for the checksum fiasco for `snapshot fetch`), the corrupted file should be removed from the temporary directory.

**Task summary**

- [ ] `snapshot export` writes to a temporary location. Only on full success, the snapshot (and the checksum file) are moved to the target directory. Otherwise, the partial file is removed.
- [ ] `snapshot fetch` writes to a temporary location during download. In case of an error during the download, the file should be deleted. In case of an error during checksum verification, the file should not be moved to the target directory but kept in the temporary directory (and the path to it should be displayed to the user).

**Acceptance Criteria**

- [ ] in case of in-flight errors, interrupts etc. the partial snapshots are not pushed to the target directory and are saved to a temporary directory which should not survive OS reboot,
- [ ] in case of a checksum verification fiasco, the faulty snapshot stays in the temporary directory for potential investigation. An informative message is displayed.

**Other information and links**

Feel free to introduce any flags that would make sense and make the API more ergonomic.

PR #2108, opened 2022-10-21
feat: use temp file for snapshot [fetch|export] before finished

hanabi1224:
**Summary of changes**
- use temp file for `snapshot [fetch|export]` before finished
- extend `snapshot [list|remove|prune|clean|]` logic to handle those temp files

Closes https://github.com/ChainSafe/forest/issues/2101


lemmih (approved):
Looks good. I wonder if we can resume a partial download that got interrupted.

LesnyRumcajs (commented):
When creating the issue I was instead thinking about using a *more* temporary file, i.e. writing it somewhere to `/tmp`. This way, in case of an interrupt or a panic it would get automatically cleaned up by the OS on the next reboot.

hanabi1224:
> i.e. writing it somewhere to /tmp

There are some corner cases like when the temp folder and snapshot folder are on different drive, the `rename` can be expensive

> get automatically cleaned up by the OS

Not sure if this works on all OS, e.g. windows?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -316,13 +322,20 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
     let snapshots_to_delete: Vec<_> = fs::read_dir(snapshot_dir)?
         .flatten()
         .map(|entry| entry.path())
-        .filter(|p| p.extension().unwrap_or_default() == "car")
+        .filter(|p| {
+            let ext = p.extension().unwrap_or_default();
+            ext == "car" || ext == "tmp"
```
LesnyRumcajs, line 327: How about moving this predicate to a common method?

tyshko5 (approved):

LesnyRumcajs (commented):
To my understanding, this will still bloat the filesystem in case of a failure that is not even a panic. E.g. if something makes the method return an `Err` just before the rename. While I understand some artefacts may be left on a `panic` somewhere in the middle, they shouldn't be there on a gracious failure.

If you don't feel like using `TempDir` or `TempFile` are a good idea, perhaps it'd be worth writing a custom wrapper with a deleter in the destructor

For the `snapshot fetch`, it doesn't seem to me that the user is sufficiently informed about the artefact, somewhere on the disk.

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: Fixed

hanabi1224:
> perhaps it'd be worth writing a custom wrapper with a deleter in the destructor

@lemmih that's great idea, fixed

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -217,7 +222,8 @@ fn remove(config: &Config, filename: &PathBuf, snapshot_dir: &Option<PathBuf>, f
     let snapshot_path = snapshot_dir.join(filename);
     if snapshot_path.exists()
         && snapshot_path.is_file()
-        && snapshot_path.extension() == Some(OsStr::new("car"))
+        && (snapshot_path.extension() == Some(OsStr::new("car"))
+            || snapshot_path.extension() == Some(OsStr::new("tmp")))
```
lemmih, line 226: Can you use `is_car_or_tmp` here?

hanabi1224: Good catch, fixed

PR #2123, opened 2022-10-25
fix: update dockerfile to use cached images

hanabi1224:
**Summary of changes**
- Use GitHub action runner cached images in our docker file to avoid getting rate limited.
- Switch from `debian:buster-slim` to `ubuntu:20.04` which have similar image sizes

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#cached-docker-images
|image|cached|size|
|--|--|--|
|debian:10|yes|119MB|
|debian:10-slim|no|72.5MB|
|ubuntu:20.04|yes|75.2MB|

```console
➜  ~ podman image ls
REPOSITORY                    TAG           IMAGE ID      CREATED        SIZE
docker.io/library/ubuntu      20.04         680e5dfb52c7  2 days ago     75.2 MB
docker.io/library/debian      buster-slim   3045415257d9  2 days ago     72.5 MB
docker.io/library/debian      buster        47db89e06601  2 days ago     119 MB
```

Note: ~~There're problems installing APT deps on `ubuntu:20.04`, use `debian:10` for now~~ 

Closes https://github.com/ChainSafe/forest/issues/2125


Dockerfile
```diff
@@ -47,7 +50,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y ocl-icd-opencl-
 RUN update-ca-certificates
 
 # Copy forest daemon and cli binaries from the build-env
-COPY --from=build-env /usr/local/cargo/bin/forest /usr/local/bin/forest
-COPY --from=build-env /usr/local/cargo/bin/forest-cli /usr/local/bin/forest-cli
+COPY --from=build-env /root/.cargo/bin/forest /usr/local/bin/forest
+COPY --from=build-env /root/.cargo/bin/forest-cli /usr/local/bin/forest-cli
```
LesnyRumcajs, line 54: This should work without repeating the target.

```suggestion
COPY --from=build-env /root/.cargo/bin/forest /root/.cargo/bin/forest-cli /usr/local/bin/
```

Dockerfile
```diff
@@ -14,30 +14,33 @@
...
-# strip symbols to make executable smaller
-RUN strip /usr/local/cargo/bin/forest /usr/local/cargo/bin/forest-cli
-
 ##
 # Prod image for forest binary
+# Use github action runner cached images to avoid being rate limited
+# https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#cached-docker-images
 ##
-FROM debian:10-slim
+FROM debian:10
```
LesnyRumcajs, line 43: I'm not sure it's a proper way out of this. Prod image should be as small as possible, and Debian is relatively large. Should we perhaps try `alpine`? We can od it in another task though.

Dockerfile
```diff
@@ -14,30 +14,33 @@
...
+FROM buildpack-deps:buster AS build-env
 
 # Install dependencies
-RUN apt-get update && apt-get install --no-install-recommends -y build-essential clang ocl-icd-opencl-dev cmake
+RUN apt-get update && apt-get install --no-install-recommends -y build-essential clang ocl-icd-opencl-dev cmake ca-certificates curl
+RUN update-ca-certificates
+
+# Install rustup
+# https://rustup.rs/
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
LesnyRumcajs, line 28: What if we'd use ubuntu as base image for the `build-env`? It [seems](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#rust-tools) to already have `cargo` installed.

Cargo.toml
```diff
@@ -167,3 +167,7 @@ debug = 0
 inherits  = "release"
 opt-level = 1
 lto       = "off"
+
+[profile.release]
+# https://doc.rust-lang.org/cargo/reference/profiles.html#strip
+strip = true
```
LesnyRumcajs, line 173: This step should be in this case deleted, right? https://github.com/ChainSafe/forest/blob/main/.github/workflows/release.yml#L42

LesnyRumcajs (commented):
Good idea!

Dockerfile
hanabi1224: It prompted to setup timezone when installing apt deps, not sure how to get rid of that

Cargo.toml
hanabi1224: Good catch! fixed

Dockerfile
hanabi1224: Fixed

Dockerfile
hanabi1224: That's a good idea, but we don't have cached buildpack for alpine, better to investigate in another task

Dockerfile
hanabi1224: Actually setting `ENV DEBIAN_FRONTEND="noninteractive"` fixes the issue, I can switch to `ubuntu:20.04` then

Dockerfile
hanabi1224: For now let me switch to `ubuntu:20.04` whose size is similar to `debian:buster-slim`

Dockerfile
hanabi1224: So cargo is installed on gh `ubuntu` runner but not the `ubuntu` image, we still have to install `cargo` unless we don't use a multistage dockerfile

Dockerfile
```diff
@@ -14,40 +14,47 @@
...
 ##
-FROM debian:10-slim
+FROM ubuntu:20.04
 
 # Link package to the repository
 LABEL org.opencontainers.image.source https://github.com/chainsafe/forest
 
+ENV DEBIAN_FRONTEND="noninteractive"
 # Install binary dependencies
 RUN apt-get update && apt-get install --no-install-recommends -y ocl-icd-opencl-dev libssl1.1 ca-certificates libcurl4
```
hanabi1224, line 50: libcurl4 should no longer be needed after https://github.com/ChainSafe/forest/pull/2102 was merged

README.md
```diff
@@ -37,7 +37,10 @@ Have questions? Feel free to post them in [Forest Q&A](https://github.com/ChainS
 
 No need to install Rust toolchain or other dependencies, you will need only Docker.
 ```
-❯ docker run --init -it ghcr.io/chainsafe/forest:latest --help
+# daemon
+❯ docker run --init -it --rm ghcr.io/chainsafe/forest:latest --help
+# cli
+❯ docker run --init -it --rm --entrypoint forest-cli ghcr.io/chainsafe/forest:latest --help
```
hanabi1224, line 43: We need to use `--entrypoint forest-cli` to invoke cli from docker because we're using `ENTRYPOINT` instruction, should we switch to `CMD`? @lemmih @LesnyRumcajs

Dockerfile
hanabi1224: Nvm it's still needed. But the new verification step is capable of finding that now! 
https://github.com/ChainSafe/forest/actions/runs/3336256732/jobs/5521188588#step:9:6

Dockerfile
LesnyRumcajs: Great, thanks!

README.md
LesnyRumcajs: Sounds good to me. We may need to potentially update many other things, including daily snapshots and sync checks. Let's do this in a separate issue.

LesnyRumcajs (approved):
Rock-solid @hanabi1224

Dockerfile
LesnyRumcajs: Beautiful!

Dockerfile
LesnyRumcajs: Sounds good, thanks

lemmih (approved):

PR #2131, opened 2022-10-26
chore: remove unused dev-dependencies

hanabi1224:
**Summary of changes**
- update `make udeps` to call `cargo udeps --all-targets --features submodule_tests`
- cleanup unussed dev-dependencies

Closes https://github.com/ChainSafe/forest/issues/2132


forest/daemon/Cargo.toml
```diff
@@ -82,10 +82,8 @@ tokio                            = { workspace = true, features = ["sync"] }
 toml.workspace                   = true
 
 [dev-dependencies]
-assert_cmd.workspace        = true
```
hanabi1224, line 85: @lemmih Looks like the `align_entries` rule always creates diff noices for such cases, do we really need that rule?

node/db/Cargo.toml
```diff
@@ -15,7 +15,8 @@ lz4    = ["rocksdb/lz4"]
 zlib   = ["rocksdb/zlib"]
 bzip2  = ["rocksdb/bzip2"]
 # not included by default to reduce build time
-zstd = ["rocksdb/zstd"]
+# uncomment when it needs to be used by other crates
+# zstd = ["rocksdb/zstd"]
```
lemmih, line 19: Without this feature, how do we enable `zstd` support in rocksdb?

forest/daemon/Cargo.toml
lemmih: Hmm, I guess aligning the dependencies is less relevant when they're all inherited from the workspace. I still prefer the noise, though.

node/db/Cargo.toml
hanabi1224: This feature is not yet propagated to daemon. It would just slow down cargo doc --all-features atm. I assume some offline measurement of different compression types is going on and we will eventually keep only one

LesnyRumcajs (approved):
rock-solid

lemmih (approved):

PR #2133, opened 2022-10-26
[WIP] Investigate Optional Dependencies

### LINKED ISSUE #2088: Investigate which dependencies can be made optional

connormullett:
**Summary of changes**
- Investigate long build time dependencies and determine if they are optional
- when applicable, disable features not being used to reduce build times

dependencies made optional
- opencl in FVM
- dag-pb in libipld
- all default features in rocksdb

Closes #2088 


README.md
```diff
@@ -63,7 +63,7 @@ rustup install nightly
 
 ```shell
 # Ubuntu
-sudo apt install build-essential clang ocl-icd-opencl-dev libssl-dev
```
hanabi1224, line 66: Can we also remove `ocl-icd-opencl-dev` from `Dockerfile` and [ci].yml

README.md
```diff
@@ -63,7 +63,7 @@ rustup install nightly
...
+sudo apt install build-essential clang libssl-dev
 
 # Archlinux
 sudo pacman -S base-devel clang ocl-icd openssl
```
hanabi1224, line 69: And `ocl-icd` from here?

connormullett: For whatever reason, removing cmake and opencl causes it to break. I'm not sure if these changes are doing anything or if we can actually remove these deps. @lemmih can you take a look?

connormullett: It might be required for testing but I'm not sure where it is in the codebase. I'm going to re-add the deps to the dockerfile and yaml

hanabi1224: Looks like `opencl` is required for `filecoin-proofs` https://github.com/filecoin-project/rust-fil-proofs

Issue #2138, opened 2022-10-27
Replace `isahc` with `reqwest`


lemmih:
**Issue summary**

We're using `isahc` to stream snapshots from URLs. We should instead use `reqwest` since it doesn't require `libcurl` to be installed.

**Other information and links**

https://github.com/ChainSafe/forest/blob/e044cb45562f5bc5074bed89be258580af9de1c8/utils/forest_utils/src/net/download.rs#L55-L85

PR #2140, opened 2022-10-27
chore: replace `isahc` with `hyper`

hanabi1224:
**Summary of changes**
- replace `isahc` with `hyper`

Manually verified with below command, the progress bar shows up and works
`cargo run -- --chain calibnet --import-snapshot https://forest-snapshots.fra1.digitaloceanspaces.com/calibnet/forest_snapshot_calibnet_2022-10-20_height_1405521.car --encrypt-keystore false`

Not using `reqwest` here because its [stream API](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.bytes_stream) returns an opaque type that cannot be wrapped in `FetchProgress` struct. `reqwest`  is built on top of `hyper` so `hyper` is already in the dependency tree

Closes https://github.com/ChainSafe/forest/issues/2138


Cargo.lock
```diff
@@ -1693,36 +1687,6 @@ dependencies = [
...
- "libc",
- "openssl-probe",
- "openssl-sys",
- "schannel",
- "socket2",
- "winapi",
-]
-
-[[package]]
-name = "curl-sys"
```
hanabi1224, line 1712: `curl` is gone

lemmih (approved):

sudo-shashank (approved):

PR #2142, opened 2022-10-28
chore: clean up old code in the VM

lemmih:
**Summary of changes**
- The buffered blockstore is used as part of the FVM. Old comments regarding the blockstore are deleted.
- Several parameters to the VM are no longer used and are deleted.
- Update the type signature of `validate_ticket_election` to show that it doesn't use the DB.
- Move from `isatty` to `atty`.
- Simplify a bunch of trait constraints.

Closes 


hanabi1224 (approved):

blockchain/state_manager/src/lib.rs
```diff
@@ -658,17 +630,13 @@ where
...
         let lb = if version <= NetworkVersion::V3 {
             ChainEpoch::from(10)
         } else {
             self.chain_config.policy.chain_finality
         };
-
-        if round > lb {
-            lbr = round - lb
-        }
+        let lbr: ChainEpoch = (round - lb).max(0);
```
LesnyRumcajs, line 639:
```suggestion
        let lbr = (round - lb).max(0);
```

LesnyRumcajs (approved):
I like it.

PR #2155, opened 2022-11-03
Reduce boot time by skipping genesis block scan for a given snapshot

### LINKED ISSUE #2129: Short-cut the scan for genesis blocks
**Issue summary**

When loading a snapshot, Forest has to scan through the entire blockchain to find the genesis block. This scan is a significant portion of the time required to import a snapshot.

Each tipset has a pointer to the previous tipset, with the exception of the genesis tipset:
```
┌─────────┐  ┌───────────┐  ┌─────────┐
│ Epoch N ├─►│ Epoch N-1 ├─►│ Genesis │
└─────────┘  └───────────┘  └─────────┘
```

The tipsets are indexed by their content id (a hash of their content) and this hash is validated when a snapshot is loaded. It's therefore safe to shortcut the search for the genesis tipset when a known tipset hash is found. For example, we can statically know that tipset with hash XYZ leads to the genesis block for mainnet, or tipset ZYX leads to the genesis block for calibnet.

Tasks:
- [ ] Investigate how to get the hash for the heaviest tipset through the RPC. If this is not possible, add an RPC command for this.
- [ ] Store a list of known tipset hashes for calibnet and mainnet, and document how they are computed.
- [ ] Add a flag for validating that these tipset hashes are valid.
- [ ] Shortcut the genesis block scan if a known tipset hash is found.

Performance data: #1841

https://github.com/ChainSafe/forest/blob/53b20f134b029b481b8d785d49ac76542d98e7ed/utils/genesis/src/lib.rs#L158-L164

creativcoder:
**Summary of changes**

This PR helps reduce boot time for the daemon by storing hashes of validated tipsets from a snapshot file.

- [X] Add a validated tipset hash registry to keep track of tipset hashes.
- [X] Skips scanning blockchain if a known hash is found.
- [X] Docs
- [x] Add RPC command to retrieve/validate tipset hash for a given snapshot.

[Before] Mainnet snapshot import time: 59min
[After] Mainnet snapshot import time: 27min

Closes #2129 

### DIFF
--- a/API_IMPLEMENTATION.md
+++ b/API_IMPLEMENTATION.md
@@ -2,9 +2,9 @@
 
 ## Stats
 
-- Forest method count: 70
+- Forest method count: 71
 - Lotus method count: 173
-- API coverage: 40.46%
+- API coverage: 41.04%
 
 ## Forest-only Methods
 
@@ -14,6 +14,7 @@ These methods exist in Forest only and cannot be compared:
 - `Filecoin.AuthVerify`
 - `Filecoin.ChainExport`
 - `Filecoin.ChainGetName`
+- `Filecoin.ChainGetTipSetHash`
 - `Filecoin.ChainGetTipsetByHeight`
 - `Filecoin.ChainHeadSubscription`
 - `Filecoin.ChainNotify`
--- a/blockchain/chain/src/store/index.rs
+++ b/blockchain/chain/src/store/index.rs
@@ -6,6 +6,7 @@ use forest_blocks::{Tipset, TipsetKeys};
 use forest_utils::io::ProgressBar;
 use fvm_ipld_blockstore::Blockstore;
 use fvm_shared::clock::ChainEpoch;
+use log::info;
 use lru::LruCache;
 use std::{num::NonZeroUsize, sync::Arc};
 use tokio::sync::RwLock;
@@ -24,13 +25,16 @@ pub(super) mod checkpoint_tipsets {
     use std::collections::HashMap;
     use std::str::FromStr;
 
+    const CALIBNET_GENESIS_CID: &str = "bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4";
+    const MAINNET_GENESIS_CID: &str = "bafy2bzacecnamqgqmifpluoeldx7zzglxcljo6oja4vrmtj7432rphldpdmm2";
+
     macro_rules! add_calibnet {
         ($map: ident, $key_hash:expr) => {
             $map.insert(
                 $key_hash,
                 // calibnet genesis tipset keys
                 TipsetKeys::new(vec![Cid::from_str(
-                    "bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4",
+                    CALIBNET_GENESIS_CID,
                 )
                 .unwrap()]),
             );
@@ -43,7 +47,7 @@ pub(super) mod checkpoint_tipsets {
                 $key_hash,
                 // mainnet genesis tipset keys
                 TipsetKeys::new(vec![Cid::from_str(
-                    "bafy2bzacecnamqgqmifpluoeldx7zzglxcljo6oja4vrmtj7432rphldpdmm2",
+                    MAINNET_GENESIS_CID,
                 )
                 .unwrap()]),
             );
@@ -64,7 +68,9 @@ pub(super) mod checkpoint_tipsets {
         // NB: Add desired tipset checkpoints below this by using RPC command: forest-cli chain tipset-hash <cid keys>
         const CALIBNET_CHECKPOINT_1405400: TipsetKeyHash = "7930ad8bf32b35314b3bc47b9e25249af8ec6ba7f5544c05e8b5bd3b3ec09f76df8bd2278f9b318badf1a08d0a468abd55130465c6c55f99e67badc0e614ca79";
         const MAINNET_CHECKPOINT_2325300: TipsetKeyHash = "319f2351ceaf78fbcc8688dc75a19bdf8ee6e895e547ff5cc2f7b18a3a36b65ff94c1860733137d0244352f82ba6fd9672aec14deee358e7cf6e088bf89a28b1";
+        const CALIBNET_CHECKPOINT_41000: TipsetKeyHash = "1a11a07d427348cc14eaa901de1ba9c6a4e18400bb557f5a0fabbcb22352319e31a7bc988a92525339f84275c0ef6dfbffcb50bb9d9843701875eecfa3ccb069";
         add_calibnet!(map, CALIBNET_CHECKPOINT_1405400);
+        add_calibnet!(map, CALIBNET_CHECKPOINT_41000);
         add_mainnet!(map, MAINNET_CHECKPOINT_2325300);
         map
     });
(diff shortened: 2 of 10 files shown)


lemmih:
@creativcoder I don't see how this approach can work. Right now you're essentially checking if the user is loading a known snapshot and by-passing the genesis scan if they are. If the user loads a slightly newer snapshot (let's say a snapshot with a single extra tipset) then this code won't skip any part of the genesis scan.

What we want is to short-circuit the genesis scan once a known tipset hash has been found. Let's say we load a snapshot with 100 epochs but we know the tipset at epoch 80 is part of, say, mainnet. Then we don't need to scan all the way from epoch 100 to 0. We can just scan from epoch 100 to 80.

creativcoder:
Got it, i've moved the check within the loop. Let me know if this is heading the right way

blockchain/chain/src/store/index.rs
```diff
@@ -16,6 +16,46 @@ const DEFAULT_CHAIN_INDEX_CACHE_SIZE: NonZeroUsize =
...
+    type TipsetKeyHash = &'static str;
+    type GenesisTipsetCids = TipsetKeys;
+    // A map of validated checkpoints mapping to their network id
+    type ValidatedCheckpointsRegistry = HashMap<TipsetKeyHash, GenesisTipsetCids>;
+    // The numeric suffix in the variable name is the height of the snapshot.
+    // Tipset hash at height 1405400
+    const CALIBNET_CHECKPOINT_1405400: TipsetKeyHash = "7930ad8bf32b35314b3bc47b9e25249af8ec6ba7f5544c05e8b5bd3b3ec09f76df8bd2278f9b318badf1a08d0a468abd55130465c6c55f99e67badc0e614ca79";
+    // Represents a static map of validated tipset hashes which helps to remove the need to validate the tipset
+    // back to genesis if it has been validated before, thereby reducing boot times.
+    pub(super) static VALIDATED_CHECKPOINTS: Lazy<Mutex<ValidatedCheckpointsRegistry>> =
```
lemmih, line 38: We probably don't need a `Mutex` here.

creativcoder: ah, that's copy pasta nit! Fixed

lemmih (approved):
Looks like a good start. Do we have a way to validate that the checkpoint hashes are correct?

API_IMPLEMENTATION.md
```diff
@@ -14,6 +14,7 @@ These methods exist in Forest only and cannot be compared:
 - `Filecoin.AuthVerify`
 - `Filecoin.ChainExport`
 - `Filecoin.ChainGetName`
+- `Filecoin.ChainGetTipSetHash`
```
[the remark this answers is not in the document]
creativcoder: Yes

creativcoder:
Added checkpoint for mainnet

creativcoder:
@lemmih > Do we have a way to validate that the checkpoint hashes are correct?

Can use the RPC api to pass tipset keys and get the hash and compare.

lemmih:
> @lemmih > Do we have a way to validate that the checkpoint hashes are correct?
> 
> Can use the RPC api to pass tipset keys and get the hash and compare.

Alright. In the future, we want something that can be verified by the CI. But this is fine for now.

blockchain/chain/src/store/index.rs
```diff
@@ -80,6 +144,14 @@ impl<BS: Blockstore> ChainIndex<BS> {
                 self.fill_cache(std::mem::take(&mut cur)).await?
             };
 
+            if let Some(genesis_tipset_keys) =
+                checkpoint_tipsets::genesis_from_checkpoint_tipset(lbe.tipset.key())
+            {
+                let tipset =
+                    tipset_from_keys(&self.ts_cache, &self.db, &genesis_tipset_keys).await?;
+                return Ok(tipset);
+            }
```
lemmih, line 153: This should only happen when `to == 0`.

lemmih (changes requested):

blockchain/chain/src/store/index.rs
creativcoder: Fixed

lemmih (approved):
Looks good. A follow-up PR can verify the checkpoints during CI.

blockchain/chain/src/store/index.rs
```diff
@@ -17,6 +17,70 @@ const DEFAULT_CHAIN_INDEX_CACHE_SIZE: NonZeroUsize =
...
+    type GenesisTipsetCids = TipsetKeys;
+    // A map of validated checkpoints mapping to their genesis tipkeys
+    type TipsetCheckpointsRegistry = HashMap<TipsetKeyHash, GenesisTipsetCids>;
+    // Represents a static map of validated tipset hashes which helps to remove the need to validate the tipset
+    // back to genesis if it has been validated before, thereby reducing boot times.
+    pub(super) static TIPSET_CHECKPOINTS: Lazy<TipsetCheckpointsRegistry> = Lazy::new(|| {
+        let mut map = HashMap::new();
+        // The numeric suffix in the variable name is the height of the snapshot.
+        // NB: Add desired tipset checkpoints below this by using RPC command: forest-cli chain tipset-hash <cid keys>
+        const CALIBNET_CHECKPOINT_1405400: TipsetKeyHash = "7930ad8bf32b35314b3bc47b9e25249af8ec6ba7f5544c05e8b5bd3b3ec09f76df8bd2278f9b318badf1a08d0a468abd55130465c6c55f99e67badc0e614ca79";
```
hanabi1224, line 66: 1405400 does not look like the correct height in rebooted calibnet

blockchain/chain/src/store/index.rs
```diff
@@ -17,6 +17,70 @@ const DEFAULT_CHAIN_INDEX_CACHE_SIZE: NonZeroUsize =
...
+    use std::collections::HashMap;
+    use std::str::FromStr;
+
+    macro_rules! add_calibnet {
+        ($map: ident, $key_hash:expr) => {
+            $map.insert(
+                $key_hash,
+                // calibnet genesis tipset keys
+                TipsetKeys::new(vec![Cid::from_str(
+                    "bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4",
```
hanabi1224, line 34: Just to confirm, does this hash match the one from rebooted calibnet?

blockchain/chain/src/store/index.rs
creativcoder: yep, am currently rebasing and integrating changes from recent calibnet snapshot

blockchain/chain/src/store/index.rs
creativcoder: yes, genesis keys are always same.

blockchain/chain/src/store/index.rs
creativcoder: Added current snapshot's tipset hash at height: 41000 below

blockchain/chain/src/store/index.rs
```diff
@@ -16,6 +17,71 @@ const DEFAULT_CHAIN_INDEX_CACHE_SIZE: NonZeroUsize =
...
+        add_calibnet!(map, CALIBNET_CHECKPOINT_41000);
+        add_mainnet!(map, MAINNET_CHECKPOINT_2325300);
+        map
+    });
+
+    pub(super) fn genesis_from_checkpoint_tipset(tsk: &TipsetKeys) -> Option<GenesisTipsetCids> {
+        TIPSET_CHECKPOINTS.get(&tipset_hash(tsk).as_str()).cloned()
+    }
+
+    pub fn tipset_hash(tsk: &TipsetKeys) -> String {
```
LesnyRumcajs, line 78: Would it be beneficial not to operate on strings (unless displaying them to the user)? Operating on raw bytes is often cheaper, easier to compare, no UTF-8 wide chars in the back etc.

LesnyRumcajs: > Always operate on raw bytes, never on encoded strings. Only use hex and base64 for pretty-printing.

from cryptopals :)

LesnyRumcajs (approved):

blockchain/chain/src/store/index.rs
creativcoder: I agree on using byte array would be perf friendly, but unfortunately we need to store these as constant values in chain_store/index.rs so a hash was ideal, compared to storing a 128 size byte array and the tipset_hash function is used in the rpc call made by the user.

creativcoder: Also, tipset_hash will only get invoked during initial scan and only for snapshot loads and it likely won't run for long as the tipset checkpoint will be near the head of the snapshot.

Issue #2158, opened 2022-11-04
forest-cli option to use external tool like aria2 for downloading


hanabi1224:
**Issue summary**

[Aria2](https://aria2.github.io/) supports resuming incomplete downloads as well as checksum validation.

- [ ] cli option `--aria2 Use aria2 for downloading snapshots` 
- [ ] fail and show installation instructions when `--aria2` is specified but `aria2c` is not available.

**Other information and links**

PR #2159, opened 2022-11-07
feat: forest-cli snapshot fetch --aria2

hanabi1224:
**Summary of changes**
- [x] cli option `--aria2 Use aria2 for downloading snapshots` 
- [x] fail and show installation instructions when `--aria2` is specified but `aria2c` is not available.

Download took `2m35s` in [previous calibnet sync check job](https://github.com/ChainSafe/forest/actions/runs/3290409553/jobs/5423210747#step:6:1)
It takes `1m1s` in [latest job](https://github.com/ChainSafe/forest/actions/runs/3407819957/jobs/5667829744#step:6:1) with `--aria2`

Closes https://github.com/ChainSafe/forest/issues/2158


sudo-shashank (approved):
great

forest/cli/src/cli/snapshot_fetch.rs
```diff
@@ -202,6 +220,66 @@ where
...
+    snapshot_path: &Path,
+) -> anyhow::Result<()>
+where
+    C: Connect + Clone + Send + Sync + 'static,
+{
+    info!("Snapshot url: {url}");
+    info!("Snapshot will be downloaded to {}", snapshot_path.display());
+
+    if let Err(err) = which::which("aria2c") {
+        bail!("Command aria2c is not available. To use --aria2, make sure you have aria2 installed. E.g. sudo apt-get install -y aria2 or brew intall aria2 or follow the instructions on https://aria2.github.io/ \n{err}");
```
lemmih, line 235: Let's wrap this line to 100 characters.

forest/cli/src/cli/snapshot_fetch.rs
```diff
@@ -407,4 +487,28 @@ mod test {
             assert!(replace_extension_url(case.0.try_into().unwrap(), case.1).is_err())
         });
     }
+
+    #[test]
+    fn download_with_aria2_test_wrong_checksum() -> Result<()> {
```
lemmih, line 492: Do these tests work if `aria2` isn't installed?

lemmih (changes requested):

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -53,6 +53,9 @@ pub enum SnapshotCommands {
         /// in default Forest data location.
         #[structopt(short, long)]
         snapshot_dir: Option<PathBuf>,
+        /// Use aria2 for downloading, default is false.
```
lemmih, line 56: `aria2` should be a code block or, better yet, a link. Then we can remove it from the dictionary.

forest/cli/src/cli/snapshot_fetch.rs
```diff
@@ -407,4 +487,28 @@ mod test {
...
+        let url = "https://github.com/ChainSafe/forest/raw/07f9b397324f63711fd587fa2cff52a20fcb9c51/.github/forest_logo.png";
```
LesnyRumcajs, line 493: From my experience in my former project, Github infrastructure is not that reliable to use in tests. How about using a self-spawned server to make our tests independent (and able to be run offline)?

forest/cli/src/cli/snapshot_cmd.rs
LesnyRumcajs, line 56:
```suggestion
        /// Use aria2 for downloading, default is false. Requires `aria2c` in PATH.
```

forest/cli/src/cli/snapshot_fetch.rs
```diff
@@ -202,6 +220,66 @@ where
...
+    info!("Snapshot will be downloaded to {}", snapshot_path.display());
+
+    if let Err(err) = which::which("aria2c") {
+        bail!("Command aria2c is not available. To use --aria2, make sure you have aria2 installed. E.g. sudo apt-get install -y aria2 or brew intall aria2 or follow the instructions on https://aria2.github.io/ \n{err}");
+    }
+
+    let checksum_url = replace_extension_url(url.clone(), "sha256sum")?;
+    let checksum_response = client.get(checksum_url.as_str().try_into()?).await?;
+    if !checksum_response.status().is_success() {
+        bail!("Unable to get the checksum file. Snapshot downloaded but not verified. Url: {checksum_url}");
```
LesnyRumcajs, line 241: This is not precisely true given the current logic, no? The user will get the message `Snapshot downloaded but not verified (...)`, but nothing is there, right?

LesnyRumcajs (changes requested):

forest/cli/src/cli/snapshot_fetch.rs
hanabi1224: Ah good catch, I copy & paste the text w/o reading it : )

forest/cli/src/cli/snapshot_fetch.rs
hanabi1224: It will fail, would u suggest make it pass w/o aria2 installed?

forest/cli/src/cli/snapshot_fetch.rs
hanabi1224: Fixed

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: Fixed

hanabi1224: Fixed

forest/cli/src/cli/snapshot_fetch.rs
hanabi1224: Fixed

forest/cli/src/cli/snapshot_fetch.rs
lemmih: Yeah, I'd like for `aria2` to be an optional dependency for developers.

hanabi1224: Fixed

creativcoder (approved):

forest/cli/src/cli/snapshot_fetch.rs
```diff
@@ -407,4 +491,64 @@ mod test {
...
+            temp_dir().as_os_str().to_str().unwrap_or_default(),
+            "sha-256=124305d4602185addd53df5f39a35b2564baa3f51eb211b266af2db77844266d",
+        )?;
+        shutdown_tx.send(()).unwrap();
+        Ok(())
+    }
+
+    fn serve_forest_logo() -> Result<(String, tokio::sync::oneshot::Sender<()>)> {
+        let port = OsRng.gen_range(20000..40000);
+        let url = format!("http://127.0.0.1:{port}/forest_logo.png");
```
LesnyRumcajs, line 532: I am not sure about this particular server, but in general it is not needed to randomly generate a random port, it should be handled by OS to get an available one. The chance for a clash is small, so feel free to skip this suggestion.

```
use std::net::TcpListener;

fn main() {
    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
    println!("{:?}", listener.local_addr());
}
```
`> Ok(127.0.0.1:62868)`

LesnyRumcajs (approved):
Good stuff!

forest/cli/src/cli/snapshot_fetch.rs
hanabi1224: That will be great, the only problem is that I have not figured out how to retrieve the random port with `axum` api to construct the logo url

LesnyRumcajs: something along those lines? https://github.com/tokio-rs/axum/blob/main/axum/src/extract/connect_info.rs#L197-L207

hanabi1224: Fixed. Thanks for looking into it!

lemmih (approved):

PR #2170, opened 2022-11-09
fix: don't ask to download a snapshot if one is already given

lemmih:
**Summary of changes**
- Don't ask to download a snapshot if we're already trying to import one.

Closes 

### DIFF
--- a/forest/daemon/src/daemon.rs
+++ b/forest/daemon/src/daemon.rs
@@ -184,6 +184,8 @@ pub(super) async fn start(config: Config, detached: bool) {
     .unwrap();
     chain_store.set_genesis(&genesis.blocks()[0]).unwrap();
 
+    // XXX: This code has to be run before we're starting background services.
+    //      If it isn't several threads will be competing for access to stdout.
     // Terminate if no snapshot is provided or DB isn't recent enough
     let should_fetch_snapshot = match chain_store.heaviest_tipset().await {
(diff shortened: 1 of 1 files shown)


forest/daemon/src/daemon.rs
```diff
@@ -184,20 +184,6 @@ pub(super) async fn start(config: Config, detached: bool) {
     .unwrap();
     chain_store.set_genesis(&genesis.blocks()[0]).unwrap();
 
-    // Terminate if no snapshot is provided or DB isn't recent enough
```
hanabi1224, line 187: Maybe also re-enable the `download` and `import` part of `calibnet sync check` job?

lemmih: I'd rather merge support for NV17 and then re-enable the entire sync check.

creativcoder (approved):

hanabi1224 (approved):

LesnyRumcajs:
This is not exactly right, the prompt should be given before everything else. Otherwise there's a race condition in the stdout.

image

LesnyRumcajs:
Instead, perhaps the condition in the `prompt_snapshot_or_die` should be adapted to this case I missed, so omit download if `import_snapshot` is already set to something.

LesnyRumcajs (approved):
Looks good. Thanks for fixing this.

PR #2173, opened 2022-11-09
feat: verify and retry download from ipfs gateway

hanabi1224:
**Summary of changes**
- verify the content length
- add retry mechanism
- ~~verify cid~~ https://docs.ipfs.tech/concepts/hashing/#content-identifiers-are-not-file-hashes
- simplify cid usage in cargo manifest

Closes https://github.com/ChainSafe/forest/issues/2177


utils/paramfetch/src/lib.rs
```diff
@@ -147,19 +147,34 @@ async fn fetch_verify_params(
...
+            debug!("Successfully downloaded {url} to {}", path.display());
+            r = Ok(());
+            break;
+        }
+    }
+    r
+}
 
+async fn fetch_params_inner(url: impl AsRef<str>, path: &Path) -> Result<(), anyhow::Error> {
+    let client: surf::Client = surf::Config::default().set_timeout(None).try_into()?;
```
creativcoder, line 167: not crticial but can move `surf::Client` init outside of retry loop.

utils/paramfetch/src/lib.rs
```diff
@@ -147,19 +147,34 @@ async fn fetch_verify_params(
...
     let url = format!("{}{}", gw, info.cid);
 
-    let client: surf::Client = surf::Config::default().set_timeout(None).try_into()?;
-
-    let req = client.get(url.as_str());
-
-    let mut source = req.await.map_err(|e| anyhow::anyhow!(e))?.compat();
-    copy(&mut source, &mut writer).await?;
+    let mut r = Ok(());
+    for _ in 0..5 {
```
creativcoder, line 153: can be overkill, but can try using https://github.com/ihrwein/backoff

utils/paramfetch/src/lib.rs
hanabi1224: Actually, I keep it in the loop intentionally so that we create a new `Client` which might help get correct result

utils/paramfetch/src/lib.rs
hanabi1224: Fixed

.github/workflows/rust.yml
```diff
@@ -121,6 +121,8 @@ jobs:
         env:
           CC: "sccache clang"
           CXX: "sccache clang++"
+      - name: fetch params
+        run: forest-cli fetch-params --keys
```
lemmih, line 125: Doesn't this happen automatically? Do you want it to be a separate step for easier debugging?

lemmih (approved):

.github/workflows/rust.yml
hanabi1224: Yeah it might be easier to observe warnings and errors if there are regressions.

tyshko5 (approved):

Issue #2076, opened 2022-10-17
Incorrect CLI usage flags for forest and forest-cli


LesnyRumcajs:
**Describe the bug**

After the binaries split was introduced in https://github.com/ChainSafe/forest/issues/1901, some of the flags don't make any sense for some of the binaries, e.g. `--import-snapshot` for `forest-cli` or `--token` for `forest` (unless I'm missing something)

**To Reproduce**
1. Run `forest-cli`
2. See usage flags.

**Log output**

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
❯ target/debug/forest-cli --help
forest-cli 0.4.1+git.110e83ad
ChainSafe Systems <info@chainsafe.io>
Forest binary sub-commands available

USAGE:
    forest-cli [FLAGS] [OPTIONS] <SUBCOMMAND>

FLAGS:
        --detach               Daemonize Forest process
        --halt-after-import    Halt with exit code 0 after successfully importing a snapshot
    -h, --help                 Prints help information
        --skip-load            Skips loading CAR file and uses header to index chain. Assumes a pre-loaded database
    -V, --version              Prints version information

OPTIONS:
        --chain <chain>
            Choose network chain to sync to [default: mainnet]  [possible values: mainnet, calibnet]

    -c, --config <config>                            A TOML file containing relevant configurations
        --encrypt-keystore <encrypt-keystore>        Encrypt the key-store (default: true)
    -g, --genesis <genesis>                          The genesis CAR file
        --height <height>                            Validate snapshot at given EPOCH
        --import-chain <import-chain>                Import a chain from a local CAR file or URL
        --import-snapshot <import-snapshot>          Import a snapshot from a local CAR file or URL
    -k, --kademlia <kademlia>                        Allow Kademlia (default: true)
        --mdns <mdns>                                Allow MDNS (default: false)
        --metrics-address <metrics-address>
            Address used for metrics collection server. By defaults binds on localhost on port 6116

        --req-window <req-window>                    Number of tipsets requested over chain exchange (default is 200)
    -r, --rpc <rpc>                                  Allow RPC to be active or not (default: true)
        --rpc-address <rpc-address>                  Address used for RPC. By defaults binds on localhost on port 1234
        --target-peer-count <target-peer-count>      Amount of Peers we want to be connected to (default is 75)
        --tipset-sample-size <tipset-sample-size>
            Number of tipsets to include in the sample that determines what the network head is

[...]

PR #2175, opened 2022-11-10
Implement Correct Usage Flags for `forest` and `forest-cli`

jdjaustin:
**Summary of changes**
- Change flags available for `forest` and `forest-cli` to only those that make sense and are usable.

Closes associated tasks in issue #2076 

Task to change the config dump behavior will be covered under a separate PR.


forest/shared/src/cli/mod.rs
```diff
@@ -351,31 +443,31 @@ mod tests {
...
+foo = "foo"
+[myth]
+author = 'H. P. Lovecraft'
+entities = [
+{ name = 'Cthulhu' },
+{ name = 'Azathoth' },
+{ baz = 'Dagon' },
+]
+bar = "bar"
+"#,
```
lemmih, line 456: This change looks unintentional.

forest/shared/src/cli/mod.rs
```diff
@@ -204,7 +280,23 @@ impl ConfigPath {
...
+        return Some(ConfigPath::Env(PathBuf::from(s)));
+    }
+    if let Some(dir) = ProjectDirs::from("com", "ChainSafe", "Forest") {
+        let path = dir.config_dir().join("config.toml");
+        if path.exists() {
+            return Some(ConfigPath::Project(path));
+        }
+    }
+    None
+}
```
lemmih, line 297: `find_cli_config_path` and `find_daemon_config_path` should be merged. Or refactored such that they share their logic.

forest/shared/src/cli/mod.rs
```diff
@@ -30,6 +30,82 @@ pub static FOREST_VERSION_STRING: Lazy<String> =
 /// CLI options
 #[derive(StructOpt, Debug)]
 pub struct CliOpts {
+    /// A TOML file containing relevant configurations
+    #[structopt(short, long)]
+    pub config: Option<String>,
+    /// Allow RPC to be active or not (default: true)
+    #[structopt(short, long)]
+    pub rpc: Option<bool>,
```
lemmih, line 38: Does `forest-cli` use the `rpc` flag?

forest/shared/src/cli/mod.rs
```diff
@@ -83,10 +159,10 @@ pub struct CliOpts {
...
     /// Choose network chain to sync to
     #[structopt(
-        long,
-        default_value = "mainnet",
-        possible_values = &["mainnet", "calibnet"],
-    )]
+long,
+default_value = "mainnet",
+possible_values = &["mainnet", "calibnet"],
+)]
```
lemmih, line 165: This change looks unintentional.

forest/shared/src/cli/mod.rs
```diff
@@ -104,14 +180,14 @@ pub struct CliOpts {
     // However, we cannot upgrade to version 0.9 because pretty_env_logger depends on version 0.7
     // and hasn't been updated in quite a while. See https://github.com/seanmonstar/pretty-env-logger/issues/52
     // #[structopt(
-    //     help = "Specify a filename into which logging should be appended"
+    // help = "Specify a filename into which logging should be appended"
```
lemmih, line 183: This change looks unintentional.

forest/shared/src/cli/mod.rs
jdjaustin: There is a test that executes `forest-cli --rpc true --token Azazello config dump`. I (perhaps incorrectly) assumed that the test was there for a reason.

lemmih: You need to understand what that test does and then update it if necessary.

forest/shared/src/cli/mod.rs
```diff
@@ -204,20 +280,44 @@ impl ConfigPath {
...
+            if let Ok(s) = std::env::var("FOREST_CONFIG_PATH") {
+                return Some(ConfigPath::Env(PathBuf::from(s)));
+            }
+            if let Some(dir) = ProjectDirs::from("com", "ChainSafe", "Forest") {
+                let path = dir.config_dir().join("config.toml");
+                if path.exists() {
+                    return Some(ConfigPath::Project(path));
+                }
+            }
+            None
```
lemmih, line 318: You've still duplicated the logic. The logic is exactly the same in both instances.

lemmih (changes requested):

forest/shared/src/cli/mod.rs
```diff
@@ -86,7 +162,7 @@ pub struct CliOpts {
         long,
         default_value = "mainnet",
         possible_values = &["mainnet", "calibnet"],
-    )]
+     )]
```
lemmih, line 165: This looks unintentional.

forest/shared/src/cli/mod.rs
```diff
@@ -204,8 +280,8 @@ impl ConfigPath {
     }
 }
 
-fn find_config_path(opts: &CliOpts) -> Option<ConfigPath> {
-    if let Some(s) = &opts.config {
+fn find_config_path(opts: &Option<String>) -> Option<ConfigPath> {
```
lemmih, line 283: Good. But `opts` is no longer a proper name for this variable.

lemmih (changes requested):
Getting closer. You still need to review all the cli options, though. It's not too difficult to see if a command-line option is used or not. You can either use `git grep` or simply delete the field and see what breaks.

lemmih (approved):
There's a lot more that can be done to clean up the command-line handling for `forest` and `forest-cli` but this is a good start.

forest/shared/src/cli/mod.rs
```diff
@@ -30,6 +30,66 @@ pub static FOREST_VERSION_STRING: Lazy<String> =
...
+    pub config: Option<String>,
+    /// Client JWT token to use for JSON-RPC authentication
+    #[structopt(short, long)]
+    pub token: Option<String>,
+    /// Address used for RPC. By defaults binds on localhost on port 1234.
+    #[structopt(long)]
+    pub rpc_address: Option<SocketAddr>,
+    /// Address used for metrics collection server. By defaults binds on localhost on port 6116.
+    #[structopt(long)]
+    pub metrics_address: Option<SocketAddr>,
```
LesnyRumcajs, line 44: I don't think we use it anywhere in the `forest-cli` unless I'm missing something.
```suggestion
```

LesnyRumcajs (changes requested):

forest/shared/src/cli/mod.rs
```diff
@@ -30,6 +30,66 @@ pub static FOREST_VERSION_STRING: Lazy<String> =
...
+    // and hasn't been updated in quite a while. See https://github.com/seanmonstar/pretty-env-logger/issues/52
+    // #[structopt(
+    // help = "Specify a filename into which logging should be appended"
+    // )]
+    // pub log_file: Option<PathBuf>,
+}
+
+impl CliOpts {
+    pub fn to_config(&self) -> Result<(Config, Option<ConfigPath>), anyhow::Error> {
+        let path = find_config_path(&self.config);
```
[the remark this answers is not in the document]
lemmih: There's no such thing as a `forest-cli` config, though, and there probably never will be.

Given that there is apparently no consensus about this, I think we should close this PR and move the discussion to an issue.

forest/shared/src/cli/mod.rs
```diff
@@ -30,6 +30,66 @@ pub static FOREST_VERSION_STRING: Lazy<String> =
...
+    pub rpc_address: Option<SocketAddr>,
+    /// Address used for metrics collection server. By defaults binds on localhost on port 6116.
+    #[structopt(long)]
+    pub metrics_address: Option<SocketAddr>,
+    /// Choose network chain to sync to
+    #[structopt(
+long,
+default_value = "mainnet",
+possible_values = &["mainnet", "calibnet"],
+)]
```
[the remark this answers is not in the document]
jdjaustin: `fmt` doesn't force it, but it does seem to be creating these random messy formatting issues. I can manually fix it though (without failing the checks).

LesnyRumcajs:
One other thing, `DaemonOpts` should not have the `token` field. As mentioned in the comment above it, it's for the client, i.e., `forest-cli`.

forest/shared/src/cli/mod.rs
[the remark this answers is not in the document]
lemmih: Command-line options are different from a configuration file. In `forest` they are the same (which is a mistake, I believe, as it leads to some unintuitive behaviour). But for `forest-cli` there is no reason to have a configuration file.

Let's discuss in a separate issue.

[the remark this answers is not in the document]
LesnyRumcajs: I'm okay with having no configuration file for `forest-cli` and removing this logic. What I am not okay with is having the same logic for the configuration file between `forest` and `forest-cli`.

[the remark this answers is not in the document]
lemmih: Why not? We want the `forest-cli` tool to read the configuration file for the `forest` node, don't we?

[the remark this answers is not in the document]
LesnyRumcajs: I mean the same logic for loading two different files, including their default location. :)

PR #2180, opened 2022-11-14
Fix chain validation

elmattic:
**Summary of changes**
- Fix chain validation function

When we changed `FIL_PROOFS_PARAMETER_CACHE` from its default value (`/var/tmp/filecoin-proof-parameters`) to a configuration driven one, we have created a functional regression that stayed unnoticed for a while.

The regression shows up in [validate_chain](https://github.com/ChainSafe/forest/blob/main/blockchain/state_manager/src/lib.rs) function, resulting in an always broken state.

This PR fixes this issue by moving the init of proof parameters *before* `sync_from_snapshot` (that can call `validate_chain` internally), ensuring a proper init.

Closes n/a

### DIFF
--- a/forest/daemon/src/daemon.rs
+++ b/forest/daemon/src/daemon.rs
@@ -342,6 +342,18 @@ pub(super) async fn start(config: Config, detached: bool) {
         unblock_parent_process();
     }
 
+    // Fetch and ensure verification keys are downloaded
+    if cns::FETCH_PARAMS {
+        use forest_paramfetch::{
+            get_params_default, set_proofs_parameter_cache_dir_env, SectorSizeOpt,
+        };
+        set_proofs_parameter_cache_dir_env(&config.client.data_dir);
+
+        get_params_default(&config.client.data_dir, SectorSizeOpt::Keys)
+            .await
+            .unwrap();
+    }
+
     let config = maybe_fetch_snapshot(should_fetch_snapshot, config).await;
 
     select! {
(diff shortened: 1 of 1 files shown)


forest/daemon/src/daemon.rs
```diff
@@ -342,6 +342,18 @@ pub(super) async fn start(config: Config, detached: bool) {
...
+        use forest_paramfetch::{
+            get_params_default, set_proofs_parameter_cache_dir_env, SectorSizeOpt,
+        };
+        set_proofs_parameter_cache_dir_env(&config.client.data_dir);
+
+        get_params_default(&config.client.data_dir, SectorSizeOpt::Keys)
+            .await
+            .unwrap();
+    }
+
```
LesnyRumcajs, line 356: It would be nice to put it into a separate method. Not a big deal though.

lemmih:
Oh my. I guess this wasn't caught because most of us still have a copy of the parameters at their default location.

LesnyRumcajs (approved):

elmattic:
> Oh my. I guess this wasn't caught because most of us still have a copy of the parameters at their default location.

Exactly.

LesnyRumcajs:
> Oh my. I guess this wasn't caught because most of us still have a copy of the parameters at their default location.

How did it work then on sync checks? I mean, the containers were always spawned with a fresh state, no?

hanabi1224 (approved):

lemmih (approved):

PR #2154, opened 2022-11-02
chore: Upgrade libp2p and bitswap

hanabi1224:
**Summary of changes**
Changes introduced in this pull request:
- [x] Upgrade `libp2p` to `0.49.0`
- [x] Upgrade `libp2p-bitswap` to `0.24`
- [x] Fix and verify new `bitswap` logic
- [x] Install `protoc` in CI env and `Dockerfile` which is required by [`libp2p 0.47+`](https://github.com/libp2p/rust-libp2p/releases/tag/v0.47.0)  It reduces build time by not building `protoc` from source)
- [x] Remove `tiny-cid`
- [x] Fix build on MacOS
- [x] Improve prometheus metrics by splitting inbound/outbound request/response
- [x] Verify metrics at http://localhost:6116/metrics
- [x] ~~Manual test on calibnet~~ `calibnet sync check` CI job passes
- [x] Manual test on mainnet

**Reference issue to close (if applicable)**

Closes https://github.com/ChainSafe/forest/issues/1924 https://github.com/ChainSafe/forest/issues/1920

**Other information and links**

PR #2184, opened 2022-11-14
feat: ban libp2p peers on genesis hash mismatch

hanabi1224:
**Summary of changes**
- ban libp2p peers on genesis hash mismatch
- ban libp2p peers on Ping protocol not supported

Based on https://github.com/ChainSafe/forest/pull/2154  Should let https://github.com/ChainSafe/forest/pull/2154 go first

Looking at the [calibnet log](https://github.com/ChainSafe/forest/actions/runs/3489296934/jobs/5839215478#step:10:23), epoch too large errors during tipset validation are completely gone 

Closes https://github.com/ChainSafe/forest/issues/2189


node/forest_libp2p/src/service.rs
```diff
@@ -158,6 +158,7 @@ pub struct Libp2pService<DB, P: StoreParams> {
     network_receiver_out: flume::Receiver<NetworkEvent>,
     network_sender_out: flume::Sender<NetworkEvent>,
     network_name: String,
+    genesis_hash: Cid,
```
LesnyRumcajs, line 161: maybe let's use `genesis_cid`, `Cid` is not precisely a hash (it does contain it, though)

node/forest_libp2p/src/service.rs
```diff
@@ -402,9 +406,10 @@ async fn handle_network_message<P: StoreParams>(
 
 #[allow(clippy::too_many_arguments)]
 async fn handle_forest_behaviour_event<DB, P: StoreParams>(
-    bh_mut: &mut ForestBehaviour<P>,
+    st_mut: &mut Swarm<ForestBehaviour<P>>,
```
LesnyRumcajs, line 409: do we need those postfixes? For example just `swarm`, it's short enough.

node/forest_libp2p/src/service.rs
```diff
@@ -425,6 +430,7 @@ async fn handle_forest_behaviour_event<DB, P: StoreParams>(
 ) where
     DB: Blockstore + Store + BitswapStore<Params = P> + Clone + Sync + Send + 'static,
 {
+    let bh_mut = st_mut.behaviour_mut();
```
LesnyRumcajs, line 433: same as above

node/forest_libp2p/src/service.rs
```diff
@@ -644,7 +658,13 @@ async fn handle_forest_behaviour_event<DB, P: StoreParams>(
                 );
             }
             Err(err) => {
-                warn!("{err}:{}", ping_event.peer.to_base58());
+                let err = err.to_string();
```
LesnyRumcajs, line 661: is this line needed?

LesnyRumcajs (approved):
In general, it looks great. Do we ban peers permanently or only for some time? Potentially some peers may "correct" their behaviour and we should perhaps bring them back. I could imagine storing those in some kind of TLRU cache. Not for this PR of course, but maybe as a potential improvement in the future.

hanabi1224:
@LesnyRumcajs 
> Do we ban peers permanently or only for some time?

Looking at the [source code](https://docs.rs/libp2p-swarm/0.40.1/src/libp2p_swarm/lib.rs.html#286), it's banned for the entire lifetime of the swarm instance.

> Potentially some peers may "correct" their behaviour and we should perhaps bring them back.

Maybe we could propose some backward-compatible API changes to the upstream, e.g. adding `fn ban_peer_id_with_duration`

node/forest_libp2p/src/service.rs
hanabi1224: Fixed. That's a better name

node/forest_libp2p/src/service.rs
hanabi1224: Fixed

node/forest_libp2p/src/service.rs
hanabi1224: Fixed

node/forest_libp2p/src/service.rs
hanabi1224: I haven't confirmed the I guess without this line, `warn!` will make an extra allocation by calling `err.to_string` in `Display` . Not a big deal tho I can change it if you prefer fewer lines

LesnyRumcajs: No, it's okay; I didn't notice the 2nd usage.

LesnyRumcajs:
> @LesnyRumcajs
> 
> > Do we ban peers permanently or only for some time?
> 
> Looking at the [source code](https://docs.rs/libp2p-swarm/0.40.1/src/libp2p_swarm/lib.rs.html#286), it's banned for the entire lifetime of the swarm instance.
> 
> > Potentially some peers may "correct" their behaviour and we should perhaps bring them back.
> 
> Maybe we could propose some backward-compatible API changes to the upstream, e.g. adding `fn ban_peer_id_with_duration`

I guess we could start by asking if there's a need for this, seems to me it may be pretty niche and it can be handled by the client code anyway (though with some overhead of an another hashmap).

lemmih (approved):

hanabi1224:
> I guess we could start by asking if there's a need for this

@LesnyRumcajs  Yes, we need to achieve this with our code. https://github.com/libp2p/rust-libp2p/issues/3140

LesnyRumcajs:
@hanabi1224 thanks for looking into this!

PR #2192, opened 2022-11-15
fix: tokio mpsc doesn't work when exporting a snapshot

lemmih:
**Summary of changes**
- Move back to an `async_std::channel` when exporting snapshots.
- For some reason, using a tokio channel blocks indefinitely. Most likely due to how we call async functions from a non-async context.

Closes 


hanabi1224 (approved):
Should we add a snapshot export step to caliber sync check ci job?

LesnyRumcajs (approved):

lemmih:
> Should we add a snapshot export step to caliber sync check ci job?

Yes.

PR #2198, opened 2022-11-15
use split-debuginfo

LesnyRumcajs:
**Summary of changes**
- https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#splitting-linux-debuginfo
- this can potentially reduce link time a bit

Closes 


Cargo.toml
```diff
@@ -191,3 +192,5 @@ lto       = "off"
 [profile.release]
 # https://doc.rust-lang.org/cargo/reference/profiles.html#strip
 strip = true
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#split-debuginfo
+split-debuginfo = "unpacked"
```
hanabi1224, line 196: Doesn't this conflict with strip?

hanabi1224 (approved):

Cargo.toml
LesnyRumcajs: it doesn't conflict but is probably useless, given that we strip away other symbols, playing with `gdb` would be most likely a pain here. I'll remove it for `release`, thanks

lemmih (approved):

PR #2202, opened 2022-11-15
feat: rocksdb block size metrics

hanabi1224:
**Summary of changes**
- Add histogram metrics for keyed blocks

Maybe we could set rocksdb block size to 2KB (default is 4KB) according the what I get on mainnet @elmattic 

```
# HELP block_size Histogram of block size
# TYPE block_size histogram
block_size_bucket{le="32"} 37
block_size_bucket{le="64"} 388
block_size_bucket{le="128"} 1355
block_size_bucket{le="256"} 10760
block_size_bucket{le="512"} 19545
block_size_bucket{le="1024"} 68417
block_size_bucket{le="2048"} 188289
block_size_bucket{le="4096"} 188414
block_size_bucket{le="8192"} 188429
block_size_bucket{le="16384"} 188446
block_size_bucket{le="32768"} 188470
block_size_bucket{le="65536"} 188550
block_size_bucket{le="+Inf"} 188550
block_size_sum 191383117
block_size_count 188550
```

Closes https://github.com/ChainSafe/forest/issues/2206


lemmih (changes requested):

elmattic:
@hanabi1224 Maybe expose this block size setting as well in our RocksDbConfig? (`set_block_size` call).

hanabi1224:
> Maybe expose this block size setting as well 

How abt doing that in a separate PR? So that we keep this one small

lemmih (approved):

elmattic:
> @elmattic
> 
> > Maybe expose this block size setting as well
> 
> How abt doing that in a separate PR? So that we keep this one small
> 
> > With smaller blocks, the indices will be bigger though
> 
> I can do some quick experiment with calibnet, but I doubt it will increase db size, I think the general conclusion is about when the value is always splitted into multiple blocks

elmattic (approved):

PR #2205, opened 2022-11-15
feat: add rocksdb block_size to config

hanabi1224:
**Summary of changes**
- add rocksdb block_size to config
- ~~sets default `block_size` to 2KB according to metrics from https://github.com/ChainSafe/forest/pull/2202~~ This PR just uses default block size
- Slightly improve DB `exists` and `get` API by using [`key_may_exist`](https://docs.rs/rocksdb/0.19.0/rocksdb/struct.DBCommon.html#method.key_may_exist)

@elmattic FYI DB size remains the same on calibnet

DB size from [CI job](https://github.com/ChainSafe/forest/actions/runs/3472493225/jobs/5803392318#step:8:34): 242.7MB 
Previous DB size from [CI job](https://github.com/ChainSafe/forest/actions/runs/3471635530/jobs/5803058120#step:8:34): 242.7MB 

Ref: [wiki](https://github.com/EighteenZi/rocksdb_wiki/blob/master/RocksDB-Tuning-Guide.md#general-options)

> block_size -- RocksDB packs user data in blocks. When reading a key-value pair from a table file, an entire block is loaded into memory. Block size is 4KB by default. Each table file contains an index that lists offsets of all blocks. Increasing block_size means that the index contains fewer entries (since there are fewer blocks per file) and is thus smaller. Increasing block_size decreases memory usage and space amplification, but increases read amplification.

Closes 


elmattic:
> @elmattic FYI DB size remains the same on calibnet

What about mainnet db size? Did overall `Get` latency improved?

From my experience it's not easy to reason on performances using calibnet, since most of the data will fit L1 and L2.

hanabi1224:
> What about mainnet db size?

I will update numbers on mainnet later, given that both 4KB and 2KB are larger than most (P95) of the values we store, my assumption is this change has no impact on any of read amplification, write amplification, and space amplification.  The only thing it would improve is cache memory usage.

> The downside of using large blocks is that RAM is wasted in the block cache.

lemmih:
I don't think we should merge this until we have concrete evidence of the benefits. Let's measure if it really decreases memory usage.

node/db/src/rocks.rs
```diff
@@ -62,6 +62,7 @@ impl RocksDb {
         if !config.optimize_for_point_lookup.is_negative() {
             let cache_size = config.optimize_for_point_lookup as usize;
             let mut opts = BlockBasedOptions::default();
+            opts.set_block_size(config.block_size);
```
elmattic, line 65: If we set the `block_size` here this setting will be only picked when `optimize_for_point_lookup` is set (`!= -1`).

It would be more accurate to set it in the else branch too, in case the user disables `optimize_for_point_lookup` but still wants 2K blocks (otherwise it will be 4K block, the default value).

node/db/src/rocks.rs
```diff
@@ -91,7 +92,12 @@ impl Store for RocksDb {
     where
         K: AsRef<[u8]>,
     {
-        self.db.get(key).map_err(Error::from)
+        let key = key.as_ref();
+        if !self.db.key_may_exist(key) {
```
elmattic, line 96: Maybe consider adding some comment here, it may not be obvious, ie:

`// If the return value of 'key_may_exist' is false, the key definitely does not exist in the db`

elmattic (commented):
Nice! Any performance gains due to usage of `key_may_exist` function? 

If most values are using around 2K bytes, 4K blocks are allocated in `.sst` files (with current setting of 4K blocks)? So DB mainnet size with 2K blocks should be smaller no?

node/db/src/rocks.rs
hanabi1224: Fixed

hanabi1224:
> Any performance gains due to usage of key_may_exist function

@elmattic It makes read performance much worse, by playing with the bench funtion [here](https://github.com/ChainSafe/forest/pull/2207#issuecomment-1318552636). Should revert the change

tyshko5 (approved):

LesnyRumcajs:
@hanabi1224 What do we do with this? Is this a draft?

hanabi1224:
@LesnyRumcajs I have repurposed this PR to not include the block_size change so now it only exposed it to the config.  It should be safe to merge but I'm not sure if we still need this as we might want to migrate to other DB soon.

Regarding numbers, I don't see any notable change in DB size or import speed by change block size to 2KB

lemmih:
> @LesnyRumcajs I have repurposed this PR to not include the block_size change so now it only exposed it to the config. It should be safe to merge but I'm not sure if we still need this as we might want to migrate to other DB soon.
> 
> Regarding numbers, I don't see any notable change in DB size or import speed by change block size to 2KB

If it doesn't have an effect on performance then I don't think we should offer it as an option. I would be quite surprised if it would be beneficial to change the default block size in the future.

LesnyRumcajs:
Shall we close it then?

PR #2207, opened 2022-11-16
feat: add multi column family support to rocksdb wrapper

hanabi1224:
**Summary of changes**
- add multi column family support to rocksdb wrapper
- store block validation result in a separate column family and remove key prefix
- store genesis / head info in a separate column family

Closes 


lemmih:
I would be surprised if this had any impact on performance. Given that it is a significant amount of extra code, we should have concrete evidence of the benefits before merging it.

hanabi1224:
> we should have concrete evidence of the benefits before merging it.

@lemmih Sure, I do plan to add some benchmark code to compare different db settings.   I think having less unique keys in a partition should generally increate its lru cache performance and bloom filter performance, one other benefit is we could implement sth like `forest-cli db clear-cache` by deleting the partition that stores block validation result so that we can easily restart tipset validation without cache

hanabi1224:
[Here](https://github.com/ChainSafe/forest/pull/2207/files#diff-ea9116b86e8119c1d728c67f1ed1d147446147ed86dd95c394bc892c2d41cba6) I added a benchmark function that measures read performance of non-partitioned and partitioned DB with N 2KB-valued records and N key-only records.  By tuning DB config, I found some possible beneficial config changes that we can measure on real mainnet snapshot later
1. Using  `opts.set_data_block_index_type(DataBlockIndexType::BinarySearch);` reduces DB size without hurting read performance
2. Using `bloom filter(5.0)` does not hurt read performance
3. Using `ribbon filter(5.0)` does not hurt read performance
4. Using partition slightly reduces DB size, its cache partition read perf is better, while block read perf difference is negligible
5. To be added

lemmih:
> [Here](https://github.com/ChainSafe/forest/pull/2207/files#diff-ea9116b86e8119c1d728c67f1ed1d147446147ed86dd95c394bc892c2d41cba6) I added a benchmark function that measures read performance of non-partitioned and partitioned DB with N 2KB-valued records and N key-only records. By tuning DB config, I found some possible beneficial config changes that we can measure on real mainnet snapshot later.
> 
> run with `cargo bench -p forest_db -- --nocapture`
> 
>     1. Using  `opts.set_data_block_index_type(DataBlockIndexType::BinarySearch);` reduces DB size without hurting read performance
> 
>     2. Using `bloom filter(5.0)` does not hurt read performance
> 
>     3. Using `ribbon filter(5.0)` does not hurt read performance
> 
>     4. Using partition slightly reduces DB size, its cache partition read perf is better, while block read perf difference is negligible
> 
>     5. Using 2KB block size and 4KB blocks have negligible impact on DB size and read performance
> 
>     6. To be added

How about we close this PR and create a new PR for setting `data_block_index_type` to the better default?

hanabi1224:
Sounds good

PR #2214, opened 2022-11-17
feat: disable rocksdb compaction by default

### LINKED ISSUE #2135: Benchmark aggressive DB configuration during tipset validation
**Issue summary**

We can benchmark Forest's tipset validation performance by loading a snapshot and re-validating the last 2000 tipsets:
```bash
# Clear out the old database files
rm -rf ~/.local/share/forest/mainnet/
# Load the snapshot into our database and quit
./forest --import-snapshot file.car --halt-after-import
# Re-validate the last 2k tipsets in the database
./forest --import-snapshot file.car --skip-load --halt-after-import --height EPOCH
```

The in above snippet, `EPOCH` should be the height of the snapshot minus 2000.

We want to test if increasing the SST file size from the default of 64 to, say, 1G has negative effects. And we want to test if setting the `compaction_type` to `none` has negative effects. This is the configuration to benchmark:

```toml
[rocks_db]
max_open_files            = -1
compression_type          = "none"
compaction_style          = "none"
optimize_for_point_lookup = 256
```

Note that there's currently no option for changing the SST file size. It will have to be added. See [set_target_file_size_base](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_target_file_size_base).

As a bonus, it would be interested to measure the effect of `compression_type`. How much space is saved and does it change the runtime performance?

Start with benchmarking calibnet as this will be much faster. When you're confident in the numbers for calibnet, run the benchmarks on mainnet.

lemmih:
**Summary of changes**
- Compaction is expensive and only beneficial for range queries. We don't do any range queries in Forest so we can disable it for a free speed-up.
- Keeping enough open files is vital for performance. Using larger SST files can limit the number of open files we require.

Disabling compression adds about 20GB to the DB size and doesn't significantly affect performance.

Mainnet (validating 400 tipsets):

| Change | Import Time | Scan Time | Validation Time | DB Size |
| :---  | :---: | :---: | :---: | :---: |
| baseline | 3764s |427s | 2304s | 71G|
| no compaction / no compression | 764s | 250s | 2444s | 93G |
| no compaction / lz4 | 725s | 238s | 2342s | 71G |
| no compaction / lz4 / small cache | 723s | 262s | 2386s | 71G |

**no compaction / no compression**
```toml
[rocks_db]
max_open_files            = -1
compression_type          = "none"
compaction_style          = "none"
optimize_for_point_lookup = 256
```

**no compaction / lz4 **
```toml
[rocks_db]
max_open_files            = -1
compaction_style          = "none"
optimize_for_point_lookup = 256
```

**no compaction / lz4 / small cache**
```toml
[rocks_db]
max_open_files            = -1
compaction_style          = "none"
```

Closes #2135 


hanabi1224:
Nice!  To further reduce `.car` import time, maybe we could try skip writing `WAL` file in `put_many_keyed` by calling [`write_without_wal`](https://docs.rs/rocksdb/0.19.0/rocksdb/struct.DBCommon.html#method.write_without_wal) and call `flush` after importing

hanabi1224 (approved):

LesnyRumcajs (approved):

PR #2221, opened 2022-11-17
Integrate snapshot export step in CI

### LINKED ISSUE #2195: Test snapshot exporting during CI
**Issue summary**

We have a CI step that imports a calibnet snapshot and verifies that it can correctly sync to the most recent tipset. We should extend this stop by also exporting a snapshot. Even as calibnet grows larger, this should take no more than a few seconds.

Snapshot exporting is also done one per day on an external server but we want quicker feedback when something stops working.

creativcoder:
**Summary of changes**
- [X] Refactors walk_snapshot method
- [X] Adds snapshot export step in CI

Closes #2195 

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -147,7 +147,9 @@ jobs:
           forest-cli sync wait
           forest-cli --chain calibnet db stats
       - name: export snapshot
-        run: forest-cli snapshot export
+        run: |
+          forest-cli snapshot export
+          sha256sum -c *.sha256sum
       - name: print forest logs
         run: |
           wget -O metrics.log http://localhost:6116/metrics
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -145,6 +145,17 @@ jobs:
         run: |
           forest-cli sync wait
           forest-cli --chain calibnet db stats
+      - name: export snapshot and sync forest with exported snapshot
+        run: |
+          rm /tmp/snapshots/*
+          forest-cli snapshot export -o /tmp/snapshots
+          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot /tmp/snapshots/*.car
+          forest-cli --chain calibnet db stats
+          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --detach
+      - name: wait for sync and check health
+        run: |
+          forest-cli sync wait
+          forest-cli --chain calibnet db stats
       - name: kill Forest process, print forest logs
         run: |
           wget -O metrics.log http://localhost:6116/metrics
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -146,12 +146,8 @@ jobs:
         run: |
           forest-cli sync wait
           forest-cli --chain calibnet db stats
-      - name: export snapshot and sync forest with exported snapshot
-        run: |
-          rm $SNAPSHOT_DIRECTORY/*
-          forest-cli snapshot export -o $SNAPSHOT_DIRECTORY
-          pkill forest
-          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot $SNAPSHOT_DIRECTORY/*.car
+      - name: export snapshot
+        run: forest-cli snapshot export
       - name: print forest logs
         run: |
           wget -O metrics.log http://localhost:6116/metrics
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,6 +15,7 @@ env:
   CI: 1
   CARGO_INCREMENTAL: 1
   CACHE_TIMEOUT_MINUTES: 5
+  SNAPSHOT_DIRECTORY: /tmp/snapshots
 
 jobs:
   test:
@@ -135,21 +136,10 @@ jobs:
       - name: fetch params
         run: forest-cli fetch-params --keys
       - name: download snapshot
-        run: forest-cli --chain calibnet snapshot fetch --aria2 -s /tmp/snapshots/
+        run: forest-cli --chain calibnet snapshot fetch --aria2 -s $SNAPSHOT_DIRECTORY
       - name: import snapshot and run Forest
         run: |
-          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot /tmp/snapshots/*.car
-          forest-cli --chain calibnet db stats
-          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --detach
-      - name: wait for sync and check health
-        run: |
-          forest-cli sync wait
-          forest-cli --chain calibnet db stats
-      - name: export snapshot and sync forest with exported snapshot
-        run: |
-          rm /tmp/snapshots/*
-          forest-cli snapshot export -o /tmp/snapshots
-          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot /tmp/snapshots/*.car
+          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot $SNAPSHOT_DIRECTORY/*.car
           forest-cli --chain calibnet db stats
           forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --detach
       - name: wait for sync and check health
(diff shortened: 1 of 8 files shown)


lemmih:
I know it's still a draft but I'm not in love with the organization. Moving all the `ipld` utils into `chain_store` feels awkward.

Also, you're mixing async code and sync code in an unusual way.

.github/workflows/rust.yml
```diff
@@ -137,6 +137,17 @@ jobs:
         run: |
           forest-cli sync wait
           forest-cli --chain calibnet db stats
+      - name: export snapshot and sync forest with exported snapshot
+        run: |
+          rm /tmp/snapshots/*
+          forest-cli snapshot export -o /tmp/snapshots
+          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot /tmp/snapshots/*.car
```
creativcoder, line 144: I wonder if we can reuse steps within a job, by referring to their ids. Is this possible?

creativcoder:
> I know it's still a draft but I'm not in love with the organization. Moving all the `ipld` utils into `chain_store` feels awkward.
> 
> Also, you're mixing async code and sync code in an unusual way.

This has been refactored as discussed

.github/workflows/rust.yml
lemmih: You can definitely re-use the database that was created in the previous steps. No need to import another snapshot.

lemmih (changes requested):
I'm not completely convinced about the `SnapshotWalker` structure. Why not use a callback like in the old `walk_snapshot`?

creativcoder:
> I'm not completely convinced about the `SnapshotWalker` structure. Why not use a callback like in the old `walk_snapshot`?

I tried refactoring back to closure based fn, but i got a few closure capture related errors:
https://github.com/ChainSafe/forest/commit/38d3d5e5f0a5e743fbf0b6cbe77b43c85774b847  thoughts?

Besides that, the main reasons I went for this refactor were:
1. Closure was doing two things, getting the blockstore and then sending the block. It just felt it was doing a lot.
2. Readability and contributer friendly. It can be hard to follow callback based code sometimes.

.github/workflows/rust.yml
```diff
@@ -144,6 +145,11 @@ jobs:
           echo "--- Forest STDOUT ---"; cat forest.out
           echo "--- Forest STDERR ---"; cat forest.err
           echo "--- Forest Prometheus metrics ---"; cat metrics.log
+      - name: export snapshot and sync forest with exported snapshot
+        run: |
+          rm $SNAPSHOT_DIRECTORY/*
+          forest-cli snapshot export -o $SNAPSHOT_DIRECTORY
```
lemmih, line 151: You can't export a snapshot without having a running node.

lemmih: `forest-cli` does pretty much nothing on its own. It just sends messages to the Forest node.

lemmih:
> > I'm not completely convinced about the `SnapshotWalker` structure. Why not use a callback like in the old `walk_snapshot`?
> 
> I tried refactoring back to closure based fn, but i got a few closure capture related errors: [38d3d5e](https://github.com/ChainSafe/forest/commit/38d3d5e5f0a5e743fbf0b6cbe77b43c85774b847) thoughts?
> 
> Besides that, the main reasons I went for this refactor were:
> 
>     1. Closure was doing two things, getting the blockstore and then sending the block. It just felt it was doing a lot.
> 
>     2. Readability and contributer friendly. It can be hard to follow callback based code sometimes.

I'm not sure the control-flow of `SnapshotWalker` is easier to follow than that of `walk_snapshot`. And, for better or worse, we're already using the callback pattern in `recurse_links`.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -617,18 +594,63 @@ where
...
+    }
+
+    async fn send_block(&self, cid: Cid) -> Result<(), Error> {
+        let block = self.load(cid)?;
+        self.tx
+            .send_async((cid, block))
+            .await
+            .expect("failed sending block for export");
+        Ok(())
+    }
```
lemmih, line 642: Avoiding cloning a block is not worth having duplicated code.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -617,18 +594,63 @@ where
...
+
+    async fn load_and_send(&self, cid: Cid) -> Result<Vec<u8>, anyhow::Error> {
+        let block = self.load(cid)?;
```
lemmih, line 645: `self.load` can be inlined here.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -617,18 +594,63 @@ where
...
+        self.tx
+            .send_async((cid, block.clone()))
+            .await
+            .expect("failed sending block for export");
```
lemmih, line 649: Can we use `?` here?

blockchain/chain/src/store/chain_store.rs
```diff
@@ -650,8 +672,10 @@ where
                 }
             }
 
-            if !skip_old_msgs || h.epoch() > incl_roots_epoch {
-                recurse_links(&mut seen, *h.messages(), &mut load_block)?;
+            let mut loader = |cid| Box::pin(self.load_and_send(cid));
```
lemmih, line 675:
```suggestion
            let mut loader = |cid| self.load_and_send(cid);
```

lemmih (changes requested):

creativcoder:
> I'm not sure the control-flow of `SnapshotWalker` is easier to follow than that of `walk_snapshot`. And, for better or worse, we're already using the callback pattern in `recurse_links`.

Reverted back to closure based walker fn

blockchain/chain/src/store/chain_store.rs
creativcoder: Fixed.

blockchain/chain/src/store/chain_store.rs
creativcoder: Fixed.

blockchain/chain/src/store/chain_store.rs
creativcoder: Fixed.

blockchain/chain/src/store/chain_store.rs
creativcoder: Fixed.

.github/workflows/rust.yml
creativcoder: ah i reordered this step incorrectly. fixed.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -492,30 +493,24 @@ where
...
-            });
-
-            Ok(block)
-        })
-        .await?;
+            async move {
+                let block = self.blockstore().get(&cid)?.ok_or_else(|| {
+                    if skip_old_msgs {
+                        Error::Other("Cid {cid} not found in blockstore".to_string())
+                    } else {
```
lemmih, line 501: I think we should get rid of `skip_old_msgs`. Lotus is also moving away from "fat" snapshots ("fat" snapshots were used by block-explorers in the past but they can download the message receipts in other ways).

Not sure if removing the flag is appropriate for this PR or if it should be in a separate PR.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -534,6 +529,61 @@ where
         Ok(digest)
     }
 
+    /// Walks over tipset and state data and loads all blocks not yet seen.
+    /// This is tracked based on the callback function loading blocks.
+    async fn walk_snapshot<F, T>(
```
lemmih, line 534: If you move this function down a few lines then the diff will be prettier.

blockchain/chain/src/store/chain_store.rs
creativcoder: I see, can remove it but removing `skip_old_msgs` would touch other modules and probably make this PR big. I can do that in a follow up PR.

blockchain/chain/src/store/chain_store.rs
creativcoder: Done

lemmih (approved):

.github/workflows/rust.yml
```diff
@@ -135,17 +136,23 @@ jobs:
...
-          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot /tmp/snapshots/*.car
+          forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --import-snapshot $SNAPSHOT_DIRECTORY/*.car
           forest-cli --chain calibnet db stats
           forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --detach
       - name: wait for sync and check health
         run: |
           forest-cli sync wait
           forest-cli --chain calibnet db stats
-      - name: kill Forest process, print forest logs
+      - name: export snapshot and sync forest with exported snapshot
```
LesnyRumcajs, line 149: It doesn't sync in this step, no? It imports the snapshot. If we want to sync, we need to `sync wait` like above.

Also, it may be a good exercise to check if the `shasum` of the exported snapshot is correct.

LesnyRumcajs (approved):
Looks good in general. The CI steps are getting kind of huge though, it would be fantastic if we could move them to a separate Makefile and only call `make ci-blabla` from the workflow. It would make the workflow files way more readable and CI steps easier to reproduce locally.

.github/workflows/rust.yml
creativcoder: We decided with @lemmih that just snapshot export would do. We would later add a snapshot verify command to validate exported snapshots.

creativcoder: I've added the sha256 check step in CI

PR #2222, opened 2022-11-17
fix: disable strict health check

lemmih:
**Summary of changes**
- The health check was a bit too strict and sometimes incorrectly sound the alarm. For now, we can rely on `forest-cli sync wait`.

Closes 


hanabi1224 (approved):

.github/workflows/rust.yml
```diff
@@ -139,7 +139,6 @@ jobs:
         run: |
           forest-cli sync wait
           forest-cli --chain calibnet db stats
-          bash scripts/health_check.sh localhost
```
LesnyRumcajs, line 142: Let's delete this file then as it is not used anywhere anymore in this repository.

LesnyRumcajs (approved):

.github/workflows/rust.yml
lemmih: Done.

lemmih:
I've manually verified that Forest can sync and make progress. I guess it just sometimes takes more than 60 seconds to advance a tipset. There should be a tipset every 30 seconds but maybe nil tipsets are getting in the way.

Issue #2002, opened 2022-10-04
Investigate support for ParityDb


elmattic:
**Issue summary**
Investigate supporting `ParityDb` from Parity Technologies. `ParityDb` is a blockchain first, ssd optimized kv store, 100% written in Rust. It's actively developed and seems to be usable in production.

The embedded database is known to be a limiting/enabling factor of performances in blockchain clients.
We would like to see if this alternate option to `RocksDB` could work in the context of Forest.

**Task summary**
- [x] figure out how to select this new backend (via a feature flag or at runtime),
- [x] create a `ParityDb` type and implement `Store` interface,
- [x] implement similar unit tests than `RocksDB` but for `ParityDb`,
- [x] evaluate performance and memory impact,
- [x] evaluate overall node stability.
  
**Acceptance Criteria**

Forest should continue to work normally:
- [x] importing a snapshot works.
- [x] syncing up to `HEAD` works.
- [x] exporting a snapshot works.
- [x] rpc service works.
- [x] performance metrics are acceptable (ideally we want something at least as fast/ memory efficient than `RocksDB`).

**Other information and links**

https://github.com/ChainSafe/forest/tree/main/node/db/tests
https://github.com/paritytech/parity-db
https://dot.polkashots.io
`sled`, a recently removed backend in Forest: https://github.com/ChainSafe/forest/pull/1625

elmattic:
Wondering how `ParityDb` can deal with non-graceful shutdown of the node.

LesnyRumcajs:
@elmattic I believe this default option handles such cases https://github.com/paritytech/parity-db/blob/master/src/options.rs#L28

PR #2224, opened 2022-11-17
feat: support parity_db

LesnyRumcajs:
**Summary of changes**
- added optional support for [paritydb](https://github.com/paritytech/parity-db)
- build Forest with `paritydb` backend: `cargo build --release --no-default-features --features forest_fil_cns,paritydb`
- separate `rocksdb` a bit; probably a more abstract way can be devised
- configuration files with rocksdb config will not compatible with `paritydb`-backed Forest

Closes https://github.com/ChainSafe/forest/issues/2002

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ lint: license clean
 	taplo fmt --check
 	taplo lint
 	cargo clippy --all-targets -- -D warnings
-	cargo clippy --all-targets --features forest_deleg_cns -- -D warnings
+	cargo clippy --all-targets --no-default-features --features forest_deleg_cns,paritydb -- -D warnings
 
 # Formats Rust and TOML files
 fmt:
@@ -91,9 +91,11 @@ test-vectors: pull-serialization-tests run-vectors
 
 # Test all without the submodule test vectors with release configuration
 test:
-	cargo nextest run --all --exclude serialization_tests --exclude forest_message --exclude forest_crypto
+	cargo nextest run --all --exclude serialization_tests --exclude forest_message --exclude forest_crypto --exclude forest_db
 	cargo nextest run -p forest_crypto --features blst --no-default-features
 	cargo nextest run -p forest_message --features blst --no-default-features
+	cargo nextest run -p forest_db --no-default-features --features paritydb
+	cargo nextest run -p forest_db --no-default-features --features rocksdb
 
 test-slow:
 	cargo nextest run -p forest_message_pool --features slow_tests
@@ -129,6 +131,6 @@ mdbook-build:
 	mdbook build ./documentation
 
 rustdoc:
-	cargo doc --workspace --all-features --no-deps
+	cargo doc --workspace --no-deps
 
 .PHONY: clean clean-all lint build release test test-all test-release license test-vectors run-vectors pull-serialization-tests install-cli install-daemon install install-deps install-lint-tools docs run-serialization-vectors rustdoc
--- a/forest/daemon/src/daemon.rs
+++ b/forest/daemon/src/daemon.rs
@@ -11,7 +11,7 @@ use forest_cli_shared::cli::{
     cli_error_and_die, db_path, default_snapshot_dir, is_aria2_installed, snapshot_fetch, Client,
     Config, FOREST_VERSION_STRING,
 };
-use forest_db::rocks::RocksDb;
+use forest_db::Store;
 use forest_fil_types::verifier::FullVerifier;
 use forest_genesis::{get_network_name_from_genesis, import_chain, read_genesis_header};
 use forest_key_management::ENCRYPTED_KEYSTORE_NAME;
@@ -24,6 +24,7 @@ use forest_rpc_api::data_types::RPCState;
 use forest_state_manager::StateManager;
 use forest_utils::io::write_to_file;
 use futures::{select, FutureExt};
+use fvm_ipld_blockstore::Blockstore;
 use fvm_shared::version::NetworkVersion;
 use log::{debug, error, info, trace, warn};
 use raw_sync::events::{Event, EventInit, EventState};
@@ -167,8 +168,7 @@ pub(super) async fn start(config: Config, detached: bool) {
 
     let keystore = Arc::new(RwLock::new(ks));
 
-    let db = forest_db::rocks::RocksDb::open(db_path(&config), &config.rocks_db)
-        .expect("Opening RocksDB must succeed");
+    let db = open_db(&config);
 
     // Initialize ChainStore
     let chain_store = Arc::new(ChainStore::new(db.clone()).await);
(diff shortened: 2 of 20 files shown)


LesnyRumcajs:
RocksDB

import:
rocks_import
run:
rocks_run

ParityDB (before enabling compression, the DB size should be smaller now)

import:
paritydb_import
run:
paritydb_run

The performance isn't that conclusive.
Import CAR:
* RocksDB: 461s
* ParityDB: 1331s

So importing a snapshot seems way better with RocksDB. On the other hand, the `scanning blockchain` part took way less time:
* RocksDB: 311s
* ParityDB: 101s

Keeping in mind that there are no particular optimisations in place (whereas we spent weeks if not months tweaking RocksDB setup). I feel it's a viable option to include. Later on, we can play around with columns.

elmattic:
Very nice!

- What settings did you took for RocksDB? (The last one for main, so the aggressive optim? just want to make sure here what we are comparing).

- What is the db size with compression for ParityDb? Is it as good as for RocksDB?

- What does the "run" command do?

lemmih (approved):

node/db/src/parity_db.rs
```diff
@@ -0,0 +1,141 @@
...
+    {
+        self.db.get(0, key.as_ref()).map_err(Error::from)
+    }
+
+    fn write<K, V>(&self, key: K, value: V) -> Result<(), Error>
+    where
+        K: AsRef<[u8]>,
+        V: AsRef<[u8]>,
+    {
+        let tx = [(0, key.as_ref(), Some(value.as_ref().to_owned()))];
```
hanabi1224, line 75: Maybe implement `bulk_write` as well? Car import may benefit from it.

node/db/Cargo.toml
```diff
@@ -8,16 +8,19 @@ edition     = "2021"
...
 [features]
-default = ["snappy", "lz4", "zlib", "bzip2"]
+default = []
 
-snappy = ["rocksdb/snappy"]
-lz4    = ["rocksdb/lz4"]
-zlib   = ["rocksdb/zlib"]
-bzip2  = ["rocksdb/bzip2"]
+rocksdb = ["dep:rocksdb", "snappy", "lz4", "zlib", "bzip2"]
+snappy  = ["rocksdb/snappy"]
```
hanabi1224, line 14: Maybe it's better to use `dep:rocksdb?/snappy` syntax

forest/daemon/src/daemon.rs
```diff
@@ -481,6 +484,22 @@ fn get_actual_chain_name(internal_network_name: &str) -> &str {
     }
 }
 
+#[cfg(feature = "rocksdb")]
```
hanabi1224, line 487: Is it possible to avoid using orthogonal / mutually exclusive features? Maybe via --db [rocks | parity] cli option (in future PRs). Personally I don't like mutually exclusive features that makes `--all-features` not working

hanabi1224 (approved):

Makefile
```diff
@@ -91,9 +91,11 @@ test-vectors: pull-serialization-tests run-vectors
 
 # Test all without the submodule test vectors with release configuration
 test:
-	cargo nextest run --all --exclude serialization_tests --exclude forest_message --exclude forest_crypto
+	cargo nextest run --all --exclude serialization_tests --exclude forest_message --exclude forest_crypto --exclude forest_db
 	cargo nextest run -p forest_crypto --features blst --no-default-features
 	cargo nextest run -p forest_message --features blst --no-default-features
+	cargo nextest run -p forest_db --no-default-features --features paritydb
```
hanabi1224, line 97: This is another reason why I don't like mutually exclusive features, it bloats testing scripts

LesnyRumcajs: Ya, on top of it it's not really the first one, we already have `forest_fil_cns` and `forest_deleg_cns`. They just don't have any test coverage. :clown_face:

node/db/Cargo.toml
LesnyRumcajs: Can you elaborate on this? I am not really aware of this syntax

hanabi1224: > Starting in 1.60, you can add a ? as in "package-name?/feature-name" which will only enable the given feature if something else has enabled the optional dependency.

https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html
https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

forest/daemon/src/main.rs
```diff
@@ -125,6 +126,10 @@ fn check_for_low_fd(config: &Config) -> Result<(), anyhow::Error> {
 
     Ok(())
 }
+#[cfg(feature = "paritydb")]
+fn check_for_low_fd(_config: &Config) -> Result<(), anyhow::Error> {
```
elmattic, line 130: So no more checks are needed with ParityDb?

elmattic (approved):

forest/daemon/src/main.rs
LesnyRumcajs: I was not able to get to a point where FD limit struck Forest so I assume it is fine for now.

node/db/Cargo.toml
LesnyRumcajs: @hanabi1224 do you mean `rocksdb?/snappy` then and not `dep:rocksdb?/snappy`? the `dep:` prefix makes `cargo` a bit angry.

hanabi1224: Yes! I guess ? Implies rocksdb is the dep instead of the feature

node/db/src/parity_db.rs
LesnyRumcajs: > Very nice!
> 
>     * What settings did you took for RocksDB? (The last one for main, so the aggressive optim? just want to make sure here what we are comparing).
I believe it was the last default config from `main` from yesterday, as I was pretty surprised by how fast it was when compared to my previous runs. :)

>     * What is the db size with compression for ParityDb? Is it as good as for RocksDB?
>

So far I'm not impressed with the compression, it sits on 92GB, so it's worse than RocksDB with the current settings.
 
>     * What does the "run" command do?
It's just running Forest for 1h after the snapshot is already imported, i.e. `target/release/forest --target-peer-count 50 --encrypt-keystore false`

LesnyRumcajs:
>     * What settings did you took for RocksDB? (The last one for main, so the aggressive optim? just want to make sure here what we are comparing).
I believe it was the last default config from `main` from yesterday, as I was pretty surprised by how fast it was when compared to my previous runs. :)

>     * What is the db size with compression for ParityDb? Is it as good as for RocksDB?
>

So far I'm not impressed with the compression, it sits on 92GB, so it's worse than RocksDB with the current settings.
 
>     * What does the "run" command do?
It's just running Forest for 1h after the snapshot is already imported, i.e. `target/release/forest --target-peer-count 50 --encrypt-keystore false`

node/db/src/parity_db.rs
LesnyRumcajs: Splendid idea!

PR #2225, opened 2022-11-17
Add `forest config dump` Command

jdjaustin:
**Summary of changes**
- `forest` and `forest-cli` binaries were recently split and the command to dump the default TOML config (mostly the daemon configuration) was placed under the `forest-cli` command functionality. This PR adds a command to dump default TOML config using `forest config dump`.

Closes associated task in issue #2076 .

Preserved `forest-cli config dump` functionality for now.


LesnyRumcajs (approved):
Looks good.

lemmih:
Should we remove the `forest-cli config dump` command? What's the reasoning for which commands belong in `forest` vs `forest-cli`?

There are several commands which do not require a running node and could therefore be placed in `forest`. For example: `fetch-params` and all of the `snapshot` commands.

But splitting up the commands between `forest` and `forest-cli` will probably be confusing to end-users.

What do you think, @LesnyRumcajs?

LesnyRumcajs:
To my understanding, this particular command dumps the config related to the node. The `forest-cli config dump` dumps the config related to the other binary, so both should be maintained if we want to use TOML files in both places.

To not introduce an additional subcommand to the node, we could somehow store the current config and `forest-cli config dump --node` would dump the current node config. Should be straightforward to implement.

In general, as you noticed, it may get confusing. Something to consider is having the following:
* `forest` - node, no subcommands
* `forest-cli` - subcommands to manage the running node, as a rule of thumb, commands that require a running one,
* `forest-tool` - everything else, like snapshot download, printing funny ascii cats etc.

lemmih:
> To my understanding, this particular command dumps the config related to the node. The `forest-cli config dump` dumps the config related to the other binary, so both should be maintained if we want to use TOML files in both places.
> 
> To not introduce an additional subcommand to the node, we could somehow store the current config and `forest-cli config dump --node` would dump the current node config. Should be straightforward to implement.
> 
> In general, as you noticed, it may get confusing. Something to consider is having the following:
> 
>     * `forest` - node, no subcommands
> 
>     * `forest-cli` - subcommands to manage the running node, as a rule of thumb, commands that require a running one,
> 
>     * `forest-tool` - everything else, like snapshot download, printing funny ascii cats etc.

The CLI doesn't have a configuration. `forest-cli config dump` prints out the configuration for the Forest daemon. I'm not sure the CLI will ever have its own configuration file.

LesnyRumcajs:
> > To my understanding, this particular command dumps the config related to the node. The `forest-cli config dump` dumps the config related to the other binary, so both should be maintained if we want to use TOML files in both places.
> > To not introduce an additional subcommand to the node, we could somehow store the current config and `forest-cli config dump --node` would dump the current node config. Should be straightforward to implement.
> > In general, as you noticed, it may get confusing. Something to consider is having the following:
> > ```
> > * `forest` - node, no subcommands
> > 
> > * `forest-cli` - subcommands to manage the running node, as a rule of thumb, commands that require a running one,
> > 
> > * `forest-tool` - everything else, like snapshot download, printing funny ascii cats etc.
> > ```
> 
> The CLI doesn't have a configuration. `forest-cli config dump` prints out the configuration for the Forest daemon. I'm not sure the CLI will ever have its own configuration file.

The goal of https://github.com/ChainSafe/forest/issues/2076 is to split it so both have a separate configuration and not a shared one. To my understanding, this PR is a step in this direction.

I won't pronounce the usefulness of config files as I'm not using them at all. I'm okay with not having them in the `forest-cli`. Then all of the subcommand should be stripped of the global config (i.e., each subcommand should explicitly request `--chain` so `forest-cli --chain calibnet snapshot fetch` wouldn't work, only `forest-cli snapshot --calibnet fetch` or even `forest-cli snapshot fetch --calibnet`).

We should ask ourselves, what is the usage we want to see? I mean, have a consensus within the team and implement it and not change it for at least some time, as it is counter-productive. To my understanding, a part of the linked issue was to propose something like this. Maybe we should have a high-level proposal first? Thoughts @lemmih @jdjaustin ?

lemmih:
> > > To my understanding, this particular command dumps the config related to the node. The `forest-cli config dump` dumps the config related to the other binary, so both should be maintained if we want to use TOML files in both places.
> > > To not introduce an additional subcommand to the node, we could somehow store the current config and `forest-cli config dump --node` would dump the current node config. Should be straightforward to implement.
> > > In general, as you noticed, it may get confusing. Something to consider is having the following:
> > > ```
> > > * `forest` - node, no subcommands
> > > 
> > > * `forest-cli` - subcommands to manage the running node, as a rule of thumb, commands that require a running one,
> > > 
> > > * `forest-tool` - everything else, like snapshot download, printing funny ascii cats etc.
> > > ```
> > 
> > 
> > The CLI doesn't have a configuration. `forest-cli config dump` prints out the configuration for the Forest daemon. I'm not sure the CLI will ever have its own configuration file.
> 
> The goal of #2076 is to split it so both have a separate configuration and not a shared one. To my understanding, this PR is a step in this direction.

Issue #2076 only affects the CLI options which, in my mind, are completely separate from the configuration file.

> I won't pronounce the usefulness of config files as I'm not using them at all. I'm okay with not having them in the `forest-cli`. Then all of the subcommand should be stripped of the global config (i.e., each subcommand should explicitly request `--chain` so `forest-cli --chain calibnet snapshot fetch` wouldn't work, only `forest-cli snapshot --calibnet fetch` or even `forest-cli snapshot fetch --calibnet`).

Yes, indeed.

> We should ask ourselves, what is the usage we want to see? I mean, have a consensus within the team and implement it and not change it for at least some time, as it is counter-productive. To my understanding, a part of the linked issue was to propose something like this. Maybe we should have a high-level proposal first? Thoughts @lemmih @jdjaustin ?

I had assumed that the node would not have any sub-commands and all queries/interactions would go through forest-cli. It would probably be a good idea to write down a plan and open it for discussion.

elmattic:
The rational to have `forest config dump` is that you want be able to extract a config file out of a given command line:

Ie:
`./target/release/forest --chain calibnet --target-peer-count 25 config dump`

This outputs a default config with 25 peers and calibnet chain.

lemmih:
> The rational to have `forest config dump` is that you want be able to extract a config file out of a given command line:
> 
> Ie: `./target/release/forest --chain calibnet --target-peer-count 25 config dump`
> 
> This outputs a default config with 25 peers and calibnet chain.

Hmm, maybe we can find a different way of supporting that use-case.

LesnyRumcajs:
@lemmih @elmattic @jdjaustin I'll make a crude design tomorrow, and then we can start a discussion, I will let you guys posted.

PR #2229, opened 2022-11-18
Validate tipset hashes via RPC command and add it as a CI step 

creativcoder:
**Summary of changes**
- Adds an RPC command ValidateTipsetCheckpoints
- Invokes validate-tipset-checkpoints in CI

~~Open questions: Do we remove checkpoints that are out of range of the current snapshot file?~~

NOTE: To be merged after #2155 

Part of #2155 

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -306,20 +306,23 @@ where
         &self,
         from: Arc<Tipset>,
         to: ChainEpoch,
+        network: String,
     ) -> Result<(), Error> {
-        info!("Validating tipset checkpoint hashes from: {}", from.epoch());
+        info!(
+            "Validating {} tipset checkpoint hashes from: {}",
+            network,
+            from.epoch()
+        );
+
         if to > from.epoch() {
-            return Err(Error::Other(
-                "Looking for tipset with height greater than start point".to_string(),
-            ));
+            return Err(Error::Other(format!(
+                "Received tipset with height ({}) greater than start ({}) point",
+                to,
+                from.epoch()
+            )));
         }
 
-        // filter tipset checkpoints that are within the snapshot's range.
-        let mut hashes: HashSet<String> = checkpoint_tipsets::TIPSET_CHECKPOINTS
-            .keys()
-            .filter(|s| s.0 <= from.epoch())
-            .map(|s| s.1.to_string())
-            .collect();
+        let mut hashes = checkpoint_tipsets::get_tipset_hashes(&network);
 
         let mut ts = from;
         let tipset_hash = checkpoint_tipsets::tipset_hash(ts.key());
@@ -333,12 +336,13 @@ where
             if to == pts.epoch() {
                 break;
             }
+
             ts = pts;
         }
 
         if !hashes.is_empty() {
             Err(Error::Other(format!(
-                "Found invalid/unknown tipset hash(es): {:?}",
+                "Found tipset hash(es) on {network} that are no longer valid: {:?}",
                 hashes
             )))
         } else {
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -327,11 +327,11 @@ where
             let tipset_hash = checkpoint_tipsets::tipset_hash(ts.key());
             hashes.remove(&tipset_hash);
 
-            if pts.epoch() == 0 {
+            ts = pts;
+
+            if ts.epoch() == 0 {
                 break;
             }
-
-            ts = pts;
         }
 
         if !hashes.is_empty() {
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -302,6 +302,50 @@ where
         }
     }
 
+    pub async fn validate_tipset_checkpoints(
+        &self,
+        from: Arc<Tipset>,
+        to: ChainEpoch,
+    ) -> Result<(), Error> {
+        info!("Validating tipset checkpoint hashes from: {}", from.epoch());
+        if to > from.epoch() {
+            return Err(Error::Other(
+                "Looking for tipset with height greater than start point".to_string(),
+            ));
+        }
+
+        // filter tipset checkpoints that are within the snapshot's range.
+        let mut hashes: HashSet<String> = checkpoint_tipsets::TIPSET_CHECKPOINTS
+            .keys()
+            .filter(|s| s.0 <= from.epoch())
+            .map(|s| s.1.to_string())
+            .collect();
+
+        let mut ts = from;
+        let tipset_hash = checkpoint_tipsets::tipset_hash(ts.key());
+        hashes.remove(&tipset_hash);
+
+        loop {
+            let pts = self.chain_index.load_tipset(ts.parents()).await?;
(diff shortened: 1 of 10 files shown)


blockchain/chain/src/store/index.rs
```diff
@@ -16,6 +17,79 @@ const DEFAULT_CHAIN_INDEX_CACHE_SIZE: NonZeroUsize =
...
+    use forest_blocks::TipsetKeys;
+    use fvm_shared::clock::ChainEpoch;
+    use once_cell::sync::Lazy;
+    use std::collections::HashMap;
+    use std::str::FromStr;
+
+    const CALIBNET_GENESIS_CID: &str =
+        "bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4";
+    const MAINNET_GENESIS_CID: &str =
+        "bafy2bzacecnamqgqmifpluoeldx7zzglxcljo6oja4vrmtj7432rphldpdmm2";
```
LesnyRumcajs, line 32: Don't we have those already in the code?

forest/cli/src/cli/chain_cmd.rs
```diff
@@ -23,6 +24,12 @@ pub enum ChainCommands {
     /// Prints out the canonical head of the chain
     Head,
 
+    /// Prints a BLAKE2b hash of the tipset give a tipset keys. useful for setting checkpoints to speed up boot times from a snapshot
```
LesnyRumcajs, line 27:
```suggestion
    /// Prints a BLAKE2b hash of the tipset given its keys. Useful for setting checkpoints to speed up boot times from a snapshot
```

creativcoder: Fixed

blockchain/chain/src/store/chain_store.rs
```diff
@@ -301,6 +301,50 @@ where
...
+                "Looking for tipset with height greater than start point".to_string(),
+            ));
+        }
+
+        // filter tipset checkpoints that are within the snapshot's range.
+        let mut hashes: HashSet<String> = checkpoint_tipsets::TIPSET_CHECKPOINTS
+            .keys()
+            .filter(|s| s.0 <= from.epoch())
+            .map(|s| s.1.to_string())
+            .collect();
```
lemmih, line 321: A checkpoint in the future (ie with an epoch in the future) should be an error, I think.

node/rpc/src/chain_api.rs
```diff
@@ -355,6 +355,37 @@ where
...
+        .heaviest_tipset()
+        .await
+        .unwrap()
+        .key()
+        .cids()
+        .to_vec();
+    let ts = data
+        .state_manager
+        .chain_store()
+        .tipset_from_keys(&TipsetKeys::new(tipset_keys))
```
lemmih, line 380: It seems unnecessary to convert the `TipsetKeys` to a vector of cids and then convert it back to a `TipsetKeys`. Why not use the `TipsetKeys` directly?

lemmih (changes requested):
I don't understand why you added a `ChainEpoch` to the checkpoints. That seems unnecessary.

Also, we can only validate the networks that we're actually connected to. If we're connected to calibnet, we can only validate calibnet checkpoints. If we're connected to mainnet, we can only validate mainnet checkpoints. You need some logic to make sure we're validating the right checkpoints.

node/rpc/src/chain_api.rs
creativcoder: Yes, fixed.

lemmih: Did you push the fix?

creativcoder: not yet, will push after addressing remaining comments in this PR

blockchain/chain/src/store/chain_store.rs
creativcoder: What if some has a snapshot file that contains that "checkpoint in future", should we still consider that as an error?

lemmih: I'm not sure what you mean. My point is that the calibnet checkpoint for epoch 1405400 is definitely wrong since calibnet is currently at epoch 61000.

creativcoder: Removed chain_epoch. Changed validation fn to only consider current "network" checkpoints

blockchain/chain/src/store/chain_store.rs
```diff
@@ -301,6 +301,54 @@ where
         }
     }
 
+    pub async fn validate_tipset_checkpoints(
+        &self,
+        from: Arc<Tipset>,
+        to: ChainEpoch,
```
lemmih, line 307: We have to check that the genesis CID actually matches the expected genesis CID so I'm not sure if taking a `to` parameter makes sense. We always have to go to epoch 0.

blockchain/chain/src/store/index.rs
```diff
@@ -54,25 +53,45 @@ pub(super) mod checkpoint_tipsets {
...
+        match network {
+            "mainnet" => {
+                let keys = MAINNET_CHECKPOINTS.keys().map(|s| s.to_string());
+                HashSet::from_iter(keys)
+            }
+            "calibnet" => {
+                let keys = CALIBNET_CHECKPOINTS.keys().map(|s| s.to_string());
+                HashSet::from_iter(keys)
+            }
+            _ => unreachable!("invalid network id"),
```
lemmih, line 93: This is not unreachable. There are other networks than just calibnet and mainnet. We don't have any checkpoints for those networks yet so there's nothing to validate.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -301,6 +301,54 @@ where
...
+        }
+
+        if !hashes.is_empty() {
+            Err(Error::Other(format!(
+                "Found tipset hash(es) on {network} that are no longer valid: {:?}",
+                hashes
+            )))
+        } else {
+            Ok(())
+        }
```
lemmih, line 349: At this point we've only validated that the current blockchain contains all the known checkpoints. We also want to make sure that the genesis CID of the chain actually matches the genesis CID of the checkpoints.

lemmih (changes requested):

blockchain/chain/src/store/chain_store.rs
creativcoder: `to` is gone

blockchain/chain/src/store/index.rs
creativcoder: Fixed, returns a None

blockchain/chain/src/store/chain_store.rs
creativcoder: Added genesis check.

lemmih (approved):

blockchain/chain/src/store/chain_store.rs
```diff
@@ -302,6 +302,55 @@ where
...
     }
 
+    pub async fn validate_tipset_checkpoints(
+        &self,
+        from: Arc<Tipset>,
+        network: String,
+    ) -> Result<(), Error> {
+        info!(
+            "Validating {} tipset checkpoint hashes from: {}",
+            network,
```
LesnyRumcajs, line 312:
```suggestion
            "Validating {network} tipset checkpoint hashes from: {}",
```

LesnyRumcajs (approved):

PR #2231, opened 2022-11-18
Add benchmark script

### LINKED ISSUE #2241: Automate running benchmarks
**Issue summary**

Today we have to run manually Forest to evaluate roughly performance gains or loss of a specific code change or new setting.

We would like to have a script that automates this process and let us run several experiments in a row and on other machines as well.

elmattic:
**Summary of changes**
- Add a Ruby script to run forest in various configurations in one shot
- Change `forest-cli snapshot dir` to return just the path

```
./scripts/bench.rb --help
Usage: bench.rb [options] snapshot
        --dry-run                    Only print the commands that will be run
        --heights [Integer]          Number of heights to validate
        --pattern [String]           Run benchmarks that match the pattern
```

Output a serie of markdown tables, ie:

Bench | Import Time | Import RSS | DB Size
-|-|-|-
baseline | 15m 12s | 2.71GiB | 73G
paritydb | 13m 8s | 2.86GiB | 95G

Bench | Validate Time | Validate RSS
-|-|-
baseline | 41m 38s | 3.18GiB
paritydb | 18m 5s | 3.98GiB

Closes #2241


LesnyRumcajs:
All in all, at first glimpse, it looks terrific and is a huge step in the right direction. Manually invoking these commands is repetitive and error-prone. 

Few pre-review remarks:
* In general, it'd be great to make a class `Benchmark`, optionally some other classes. An excellent first feel on which methods should belong to such a class is passing the `bench` param, seems quite a few of them have this. Plus, some method and class documentation.
* Putting the files into a separate `benchmark_db` directory may provide some more organisation in the scripts,
* We need a way to prevent breaking this script. If it's not in the CI, then the next commit can easily break it.  We don't need a full-fledged benchmark in the CI; cover the executed lines. So I imagine having a configuration file for this script where you can put, i.e. only five heights to validate.
* I see you introduced a non-standard `gem`. Given that it is not intended to run in a container, we either need formal instructions on how to install it somewhere (for example, a `README` inside `benchmark_db` directory) or provide a `Gemfile` with all the dependencies (and in the `README` give a command, e.g. `bundle exec ...`).

I'm okay with addressing these in separate PRs. What do you think @elmattic?

elmattic:
> I'm okay with addressing these in separate PRs. What do you think @elmattic?

I'm okay addressing those in separate PRs too.

scripts/bench.rb
```diff
@@ -0,0 +1,310 @@
...
+# Provides human readable formatting to Numeric class
+class Numeric
+  def to_bibyte
+    syscall('numfmt', '--to=iec-i', '--suffix=B', '--format=%.2f', to_s)
+  end
+end
+
+def tmp_dir
+  BENCH_PATHS[:tmpdir]
+end
```
LesnyRumcajs, line 91: Do we need a separate method? Having `TEMP_DIR = Dir.mktmpdir('forest-benchs-')` should do the trick, or am I missing something?

scripts/bench.rb
```diff
@@ -0,0 +1,310 @@
...
+
+def syscall(*command)
+  stdout, _stderr, status = Open3.capture3(*command)
+  return stdout.chomp if status.success?
+
+  raise "#{command}, #{status}"
```
LesnyRumcajs, line 97: I believe this should work. Rubocop may complain though.
```suggestion
  status.success? ? stdout.chomp : raise "#{command}, #{status}"
```

scripts/bench.rb
```diff
@@ -0,0 +1,310 @@
...
+end
+
+def config_path(bench)
+  "#{tmp_dir}/#{bench[:name]}.toml"
+end
+
+def build_config_file(bench)
+  default = default_config
+  bench_config = bench[:config]
+  # TODO: Find a better way to merge (conserve the default keys)
```
LesnyRumcajs, line 189: What do you mean?

LesnyRumcajs (approved):
Looks good to me, but it's hard to review in its current form. I'm approving this but would like to see a refactored version in the future, otherwise, it may be difficult for others to maintain.

scripts/bench.rb
elmattic: Yes.

scripts/bench.rb
elmattic: let's say I have:

```
{ key1: "value1", key2: "value2" }
```
And want to merge inside my bench config KVs, then at the end I will lose original ones and have:

```
{ key1: "bench_value1" }
```
It's correct because the config keys are the default ones, but I would prefer to have this result:
```
{ key1: "bench_value1", key2: "value2" }
```

LesnyRumcajs: If I understand correctly, you want an immutable `merge`. In general, in Ruby, methods ending with `!` mutate the underlying collection, whereas the ones without are returning a new one. So, is just `merge` what you are looking for?

scripts/bench.rb
elmattic: Ah no rubocop doesn't like this. Removed the `tmp_dir` function though.

scripts/bench.rb
elmattic: Thanks for the explanation. But actually I want some `deep_merge` function but could not find one in the baselib 😞

elmattic: Will do!

scripts/benchmark_db/Gemfile
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
```
LesnyRumcajs, line 5: It's a bit unusual, why is it needed?

scripts/benchmark_db/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+You can create a selection of benchmarks using the `--pattern` flag. If used in conjunction with `--dry-run` you will see what commands will be run:
+
+```
+$ ./scripts/benchmark_db/bench.rb 2369040_2022_11_25t12_00_00z.car --pattern 'baseline*,paritydb' --dry-run
+Running bench: baseline
+$ cargo clean
+$ cargo build --release
+$ ./target/release/forest --config <tbd> --encrypt-keystore false --import-snapshot <tbd> --halt-after-import
+$ ./target/release/forest --config <tbd> --encrypt-keystore false --import-snapshot <tbd> --halt-after-import --skip-load --height 2368640
```
LesnyRumcajs, line 30: why do we need to import the snapshot twice?

elmattic: The second time is for the validation step (nothing is imported thanks to `--skip-load`).

scripts/benchmark_db/Gemfile
elmattic: Mhh, this was done by `bundle init`.

elmattic:
@LesnyRumcajs Do you understand why the lint checks are most of the time failing?

LesnyRumcajs:
> @LesnyRumcajs Do you understand why the lint checks are most of the time failing?

You mean the spellcheck fiasco?

elmattic:
> > @LesnyRumcajs Do you understand why the lint checks are most of the time failing?
> 
> You mean the spellcheck fiasco?

Yes.

LesnyRumcajs:
@elmattic I may have a spotted a potential issue in this workflow, I will let you know.

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel    = "nightly-2022-11-02"
+channel    = "nightly-2022-09-28"
```
lemmih, line 2: What's the issue with `nightly-2022-11-02` on macos?

scripts/benchmark_db/bench.rb
```diff
@@ -0,0 +1,304 @@
...
+require 'fileutils'
+require 'open3'
+require 'optparse'
+require 'pathname'
+require 'set'
+require 'tmpdir'
+require 'toml-rb'
+
+# This is just for capturing the snapshot height
+SNAPSHOT_REGEX = /(?<height>\d+)_.*/.freeze
```
lemmih, line 17: So we only support the naming scheme from the manually downloaded Filecoin snapshots? We don't support snapshots that we've downloaded with `forest-cli snapshot fetch`?

elmattic: You mean something like `minimal_finality_stateroots_2382481_2022-11-30_04-00-30.car`?

lemmih: Files downloaded with `forest-cli snapshot fetch` follow this pattern: `{provider}_snapshot_{network}_{date}_height_{height}.car`. Examples: `filecoin_snapshot_mainnet_2022-11-30_height_2382480.car` and `forest_snapshot_calibnet_2022-11-30_height_81393.car`.

lemmih: This naming change was made a few days ago, I think.

rust-toolchain.toml
elmattic: This nightly version was not available on this `aarch64-apple-darwin` target.

elmattic: Maybe we can move to `2022-11-30`?

scripts/benchmark_db/bench.rb
elmattic: Right now I don't support `calibnet` at all and scheme like this one too: `forest_snapshot_calibnet_2022-11-30_height_81393.car`

rust-toolchain.toml
lemmih: I'm surprised that `nightly-2022-11-02` isn't available for you. It's available for me on `aarch64-apple-darwin`. Could you show me the error message that you see?

lemmih: (If our macos build server doesn't catch stuff like this then we need to address it.)

scripts/benchmark_db/bench.rb
lemmih: We want to support the naming scheme used by `forest-cli snapshot`, right?

elmattic: Ok will fix this!

rust-toolchain.toml
elmattic: Maybe some missing `rustup update` on my end, will try again.

elmattic: I'm having this error: `error: the 'cargo' binary, normally provided by the 'cargo' component, is not applicable to the 'nightly-2022-11-02-aarch64-apple-darwin' toolchain`.

elmattic: Did you managed to run a `cargo build` on your side?

elmattic: Ok, I have fix it with `rustup component add cargo`.

lemmih: Hmm, maybe we should add `cargo` to the list of components in `rust-toolchain.toml`

elmattic: Let's leave this for now. I will investigate if it's really needed on another pristine computer.

lemmih (approved):

Issue #2218, opened 2022-11-17
Benchmark RocksDB without the write-ahead log


lemmih:
**Issue summary**

We might be able to get better snapshot import performance by disabling the write-ahead log (WAL) when writing batch values. Afterwards, `flush_wal` can be used to commit the data. Alternatively, since we don't care about data persistence, we could disable the WAL altogether.

| Change | Import Time | Scan Time | Validation Time | DB Size |
| :---  | :---: | :---: | :---: | :---: |
| baseline | | |  | 
| `write_without_wal` + `flush_wal` |  | |  |  |
| disable WAL |  | |  |  |

**Other information and links**

`write_without_wal`: https://docs.rs/rocksdb/0.19.0/rocksdb/struct.DBCommon.html#method.write_without_wal
`flush_wal`: https://docs.rs/rocksdb/0.19.0/rocksdb/struct.DBCommon.html#method.flush_wal
`set_manual_wal_flush`: https://docs.rs/rocksdb/0.19.0/rocksdb/struct.Options.html#method.set_manual_wal_flush

PR #2240, opened 2022-11-22
feat: Skip WAL file during car import with Rocksdb

hanabi1224:
**Summary of changes**
- use `write_without_wal` in `put_many_keyed`
- Since the current memtable size (write_buffer_size) is 1GB, more tests are needed to make sure the potential data loss does not cause fatal errors.  Car import could not benefit from skipping WAL in `put_keyed` either because [`load_car_inner` in `fvm_ipld_car` crate](https://docs.rs/fvm_ipld_car/0.5.0/src/fvm_ipld_car/lib.rs.html#155) calls `put_many_keyed` only.

car file: `minimal_finality_stateroots_2358961_2022-11-22_00-00-30.car`

| Change | Import Time | ratio | Scan Time | DB Size |
| :---  | :---: | :---: | :---: | :---: |
| baseline | 784s | 100% |19s | 72GB | 
| `write_without_wal` + `flush` | 693s | 88.4% | 21s | 72GB |

```rust
async fn load_car_inner<R, B>(s: &B, reader: R, verify: bool) -> Result<Vec<Cid>, Error>
where
    B: Blockstore,
    R: AsyncRead + Send + Unpin,
{
    let mut car_reader = if verify {
        CarReader::new(reader).await
    } else {
        CarReader::new_unchecked(reader).await
    }?;

    // Batch write key value pairs from car file
    // TODO: Stream the data once some of the stream APIs stabilize.
    let mut buf = Vec::with_capacity(100);
    while let Some(block) = car_reader.next_block().await? {
        buf.push((block.cid, block.data));
        if buf.len() > 1000 {
            s.put_many_keyed(buf.iter().map(|(k, v)| (*k, &*v)))
            buf.clear();
        }
    }
    s.put_many_keyed(buf.iter().map(|(k, v)| (*k, &*v)))
    Ok(car_reader.header.roots)
}
```

Closes https://github.com/ChainSafe/forest/issues/2218


lemmih (approved):

elmattic:
Did some tests on my machine with same snapshot and I have:

Bench | Import Time | Import RSS | DB Size
-|-|-|-
baseline | 6m 26s | 2.42GiB | 73G
write-without-wal-plus-flush | 6m 2s | 2.50GiB | 73G
---
Bench | Validate Time | Validate RSS
-|-|-
baseline | 18m 23s | 3.62GiB
write-without-wal-plus-flush | 18m 19s | 3.71GiB

Looks good.

hanabi1224:
@elmattic Thanks for testing!

Issue #1891, opened 2022-09-08
Dangling reference to DB on SIGINT


LesnyRumcajs:
**Describe the bug**

On ctrl-c during tipset validation I got `Dangling reference to DB detected: 6.`, several times in a row.

**To Reproduce**
1. Run `target/release/forest --encrypt-keystore false --chain calibnet --target-peer-count 50` (import snapshot if necessary)
3. ctrl-c while validating tipsets
4. See the dangling reference error

**Log output**

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
 target/release/forest --encrypt-keystore false --chain calibnet --target-peer-count 50
 2022-09-08T14:12:25.026Z INFO  forest::daemon > Starting Forest daemon, version 0.2.2+git.8b6f4c82
 2022-09-08T14:12:25.026Z INFO  forest_libp2p::service > Recovered libp2p keypair from "/home/rumcajs/.local/share/forest/libp2p/keypair"
 2022-09-08T14:12:25.026Z INFO  forest::daemon         > PeerId: 12D3KooWAVGaauqEMRCEB7zMd6QduhCEkkuXkTw49kLrPnJZBNr2
 2022-09-08T14:12:25.026Z WARN  forest::daemon         > Warning: Keystore encryption disabled!
 2022-09-08T14:12:25.026Z INFO  forest::daemon         > Prometheus server started at 127.0.0.1:6116
 2022-09-08T14:12:25.027Z INFO  forest::daemon         > Admin token: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.DVoFLz6ljPHMsTc_xLzKj0ShOBVETXeyROJXZ9UAakA
 2022-09-08T14:12:25.557Z INFO  forest_genesis         > Initialized genesis: BlockHeader: Cid(bafy2bzacecz3trtejxtzix4f4eebs7dekm6snfsmvffiqz2rfx7iwgsgtieq4)
 2022-09-08T14:12:25.557Z INFO  forest::daemon         > Using network :: calibrationnet
 2022-09-08T14:12:25.560Z INFO  forest::daemon         > JSON-RPC endpoint started at 127.0.0.1:1234
 2022-09-08T14:12:25.560Z INFO  forest_rpc             > Ready for RPC connections
[...]

PR #2257, opened 2022-11-23
Add explicit shutdown of tokio rt

### LINKED ISSUE #1848: Fix task cancelation
**Issue summary**

In Forest we use in a couple of places the `Task::spawn` method to spin up new tasks.

Unfortunately for all of them expect the ones in `daemon.rs` we don’t implement their cancellation making it possible that orphaned child tasks outlive parent tasks.
Any value moved to a future/ async block will never be dropped by the task unless the task has cleanly terminated or we are calling explicitly the `cancel` method on the task handle.

This can result in memory leaks over the lifetime of the client and also hindering us to shutdown gracefully the node (rocksdb Rust wrapper do not always manage to drop itself).

Therefore we should:

- [ ] Use the new heaptrack doc to create a report on current memory usage/ leaks (https://github.com/ChainSafe/forest/pull/1819)
- [ ] Conduct any research that could solve the issue, ie:
  - Tasks “structured concurrency"
  - Use arbortable futures
  - Other solution?
- [ ] Validate that forest can now be gracefully shutdown (https://github.com/ChainSafe/forest/issues/1587)
- [ ] Perform a second heaptrack analyse and observe if some of the leaks are gone (https://github.com/ChainSafe/forest/issues/1546)

https://docs.rs/async-std/latest/async_std/task/index.html#
https://docs.rs/futures/latest/futures/future/struct.Abortable.html
https://blog.yoshuawuyts.com/async-cancellation-1 (for “structured concurrency”)

**Relevant code where we use spawn without canceling**

- [chain_muxer](https://github.com/ChainSafe/forest/blob/main/blockchain/chain_sync/src/chain_muxer.rs)
- [tipset_syncer](https://github.com/ChainSafe/forest/blob/main/blockchain/chain_sync/src/tipset_syncer.rs)
- https://github.com/ChainSafe/forest/blob/main/blockchain/consensus/fil_cns/src/validation.rs#L125
- https://github.com/ChainSafe/forest/blob/main/node/forest_libp2p/src/service.rs#L260
- [state_manager](https://github.com/ChainSafe/forest/blob/main/blockchain/state_manager/src/lib.rs)
- https://github.com/ChainSafe/forest/blob/main/node/forest_libp2p/src/service.rs#L260

elmattic:
**Summary of changes**
- Add explicit shutdown of tokio runtime

Tasks running longer than 10s could still produce the dangling error but overall I've got much less errors.

We should probably have more insights on what those tasks are with the tokio console and add yield points.

I've check LOG file in rocksdb dir and it correctly shutdowns when there are no errors:
`2022/11/23-18:20:16.538822 4771888640 [db/db_impl/db_impl.cc:710] Shutdown complete`

Tested the `paritydb` backend as well.

Closes #1848 #1891

### DIFF
--- a/forest/daemon/src/main.rs
+++ b/forest/daemon/src/main.rs
@@ -13,7 +13,7 @@ use forest_cli_shared::{
 };
 use forest_utils::io::ProgressBar;
 use lazy_static::lazy_static;
-use log::{info, warn};
+use log::{error, info, warn};
 use raw_sync::events::{Event, EventInit};
 use raw_sync::Timeout;
 use shared_memory::ShmemConf;
@@ -23,6 +23,7 @@ use tokio::runtime::Runtime;
 
 use std::fs::File;
 use std::process;
+use std::sync::Arc;
 use std::time::Duration;
 
 const EVENT_TIMEOUT: Timeout = Timeout::Val(Duration::from_secs(20));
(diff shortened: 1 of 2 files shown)


hanabi1224 (approved):

forest/daemon/src/main.rs
```diff
@@ -190,7 +197,21 @@ fn main() {
...
+
+                    let db_weak_ref = Arc::downgrade(&db.db);
+                    drop(db);
+
+                    if db_weak_ref.strong_count() != 0 {
+                        error!(
+                            "Dangling reference to DB detected: {}. Tracking issue: https://github.com/ChainSafe/forest/issues/1891",
+                            db_weak_ref.strong_count()
+                        );
+                    }
```
lemmih, line 213: I would be fine with not having the dangling DB references check anymore. Or we could leave it in for now and then remove it later once we're 100% confident it can't happen anymore.

lemmih (approved):

PR #2259, opened 2022-11-23
fix: protential memory leak in state manager cache

hanabi1224:
**Summary of changes**
- replace hashmap with fix-sized lru cache 
- add cache hit/miss metrics

Observations: state manager tipset cache is pushed once per epoch, it's mainly used to avoid running `apply_block` repetitively during the validation, 1024 should be sufficient. 

Closes 


lemmih (approved):

LesnyRumcajs (approved):
LGTM, did you observe any memory/performance improvements (or degradations) with these changes?

hanabi1224:
> did you observe any memory/performance improvements (or degradations) with these changes?

No difference in of average tipset validation time observed within 1024 epoch validations that fulfill the cache.

PR #2271, opened 2022-11-24
Refactor async task spawns

creativcoder:
**Summary of changes**
- Refactors async tasks spawns based on feedback from tokio-console
- Replace FutureUnordered with JoinSet

Closes 

### DIFF
--- a/blockchain/chain_sync/src/tipset_syncer.rs
+++ b/blockchain/chain_sync/src/tipset_syncer.rs
@@ -3,7 +3,6 @@
 
 use forest_utils::io::ProgressBar;
 use futures::stream::FuturesUnordered;
-use futures::StreamExt;
 use futures::TryFutureExt;
 use fvm_shared::bigint::BigInt;
 use fvm_shared::crypto::signature::ops::verify_bls_aggregate;
@@ -1153,7 +1152,7 @@ async fn validate_tipset<DB: Blockstore + Store + Clone + Send + Sync + 'static,
     let epoch = full_tipset.epoch();
     let full_tipset_key = full_tipset.key().clone();
 
-    let mut validations = FuturesUnordered::new();
+    let mut validations = tokio::task::JoinSet::new();
     let blocks = full_tipset.into_blocks();
 
     info!(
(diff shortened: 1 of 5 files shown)


lemmih:
This clashes with the changes in #2269, right?

creativcoder:
Noticed it now, can we have async task changes in this PR?, i thought that was supposed to just tweak tokio-console

hanabi1224:
Sure I can remove the JoinSet change from the other PR

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1163,16 +1162,16 @@ async fn validate_tipset<DB: Blockstore + Store + Clone + Send + Sync + 'static,
...
     for b in blocks {
-        let validation_fn = tokio::task::spawn(validate_block::<_, C>(
-            consensus.clone(),
-            state_manager.clone(),
-            Arc::new(b),
-        ));
-        validations.push(validation_fn);
+        let consensus = consensus.clone();
+        let state_manager = state_manager.clone();
+        let validation_fn = tokio::task::spawn_blocking(|| {
```
elmattic, line 1167: Any reason to move on a `spawn_blocking` here? `validate_block` is an async function, no?

creativcoder: tokio-console shows warning when using spawn with a "⚠ n tasks have lost their waker" at this line. Replacing spawn with spawn_blocking makes the warns go away.

**Update**: I no longer receive those warns even with `spawn` after repeated runs with tokio-console, so I'm thinking of reverting this change.

elmattic: Not sure about the real root cause of this "have lost their wake" warning. Some other people are complaining of it too, see https://github.com/tokio-rs/console/issues/149.

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1163,16 +1163,16 @@ async fn validate_tipset<DB: Blockstore + Store + Clone + Send + Sync + 'static,
     debug!("Tipset keys: {:?}", full_tipset_key.cids);
 
     for b in blocks {
-        let validation_fn = tokio::task::spawn(validate_block::<_, C>(
+        let validation_fn = tokio::task::spawn(validate_block(
```
lemmih, line 1166: Is there a reason you're spawning things twice?

```suggestion
        let validation_fn = validate_block(
```

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -1283,21 +1283,22 @@ async fn validate_block<DB: Blockstore + Store + Clone + Sync + Send + 'static,
...
+    ));
+    validations.spawn(check_block_msgs);
 
     // Base fee check
     let smoke_height = state_manager.chain_config().epoch(Height::Smoke);
     let v_base_tipset = Arc::clone(&base_tipset);
     let v_block_store = state_manager.blockstore().clone();
     let v_block = Arc::clone(&block);
-    validations.push(tokio::task::spawn_blocking(move || {
+    let base_fee_check = tokio::task::spawn_blocking(move || {
```
lemmih, line 1301: I'm not too fond of this pattern with the double spawns. A small helper function for waiting on blocking spawns might be prettier. Or we could create our own structure for having collections of validators. That way, `collect_errs` wouldn't be a free-standing function.

lemmih (changes requested):

PR #2273, opened 2022-11-25
fix: disable WAL in `put_keyed` to fix #2260

### LINKED ISSUE #2260: Forest is not crash-resistant
**Describe the bug**

Forest is no longer crash-resistant. I strongly believe it was the case some time ago, as there were no failures in a similar sync check setup that surfaced now.

The issue happens when running Forest with `rocksdb` backend. I did not manage to reproduce it with `paritydb`. This may mean that the problem is on RocksDb handling, settings or possibly a regression.

**To Reproduce**
1. run `forest-cli --chain calibnet db clean` 
2. Run 'forest --encrypt-keystore false --chain calibnet'. When prompted, download the snapshot
3. Wait for the snapshot to be downloaded. Wait for few tipsets to be validated.
4. In a separate terminal, run `pkill -9 forest` (alternatively use SIGTERM and not SIGKILL, the result is the same).
5. Run `forest --encrypt-keystore false --chain calibnet`.
6. After few epochs, a failure like below should happen.
7. If it does not, go to step 1. I managed to reproduce it every time, but I am not convinced it's 100% deterministic.

**Log output**

```
 2022-11-24T11:31:44.946Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 64175, N blocks = 3
 2022-11-24T11:31:44.946Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 64176, N blocks = 5
 2022-11-24T11:31:44.946Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 64177, N blocks = 5
 2022-11-24T11:31:44.947Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 64178, N blocks = 3
 2022-11-24T11:31:44.947Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 64179, N blocks = 3
 2022-11-24T11:31:44.947Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 64180, N blocks = 3
 2022-11-24T11:31:45.011Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 64181, N blocks = 3
 2022-11-24T11:31:45.154Z WARN  forest_chain_sync::tipset_syncer > Validating block [CID = bafy2bzacebxamcrni7txe5dfu2orupgrtoyu6cwsr6ejjy37n4bg5enqqnuse] in EPOCH = 64181 failed: Validation error: Processing error: Error calculating weight: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Processing error: Could not update state: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Processing error: Failed to calculate state: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Validation error: Consensus error: StateManager error: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Consensus error: StateManager error: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune
 2022-11-24T11:31:45.154Z ERROR forest_chain_sync::tipset_syncer > Sync messages check state failed for tipset range
 2022-11-24T11:31:45.154Z ERROR forest_chain_sync::chain_muxer   > Bootstrapping failed, re-evaluating the network head to retry the bootstrap. Error = TipsetRangeSyncer(Validation("Processing error: Error calculating weight: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Processing error: Could not update state: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Processing error: Failed to calculate state: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Validation error: Consensus error: StateManager error: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune, Consensus error: StateManager error: failed to find state tree bafy2bzacedvc6zjxm6br3p62d5r6wbfyxaqp2olcpyswn56pvmlygbdkkeune"))
 2022-11-24T11:31:45.154Z INFO  forest_chain_sync::chain_muxer   > Evaluating network head...

```

**Expected behaviour**

Forest should carry on. Lost data is fine, we can always re-validate the lost tipsets. A corrupted database is NOT fine.

**Screenshots**

 - OS: Fedora 36
 - Rust version(e.g. `rustc --version`) see the linked commit's toolchain
 - Branch/commit e9c8fe7039d1676584b8a56feab00397f934080c (main)

hanabi1224:
**Summary of changes**
- Disable WAL in `put_keyed` and ~~`flush` db at the end of `validate_tipset` ~~
- ~~Turn on auto compaction to merge fragmented sst files.~~
- ~~Set sst file size~~

Issue: ~~This PR generated fragmented sst files, not ready to merge yet.~~

Closes https://github.com/ChainSafe/forest/issues/2260


LesnyRumcajs:
This fix seems to be working for me! I need it in the `main`!

node/db/src/rocks.rs
```diff
@@ -140,12 +149,11 @@ impl Store for RocksDb {
 
 impl Blockstore for RocksDb {
     fn get(&self, k: &Cid) -> anyhow::Result<Option<Vec<u8>>> {
-        self.read(k.to_bytes()).map_err(|e| e.into())
+        self.read(k.to_bytes()).map_err(Into::into)
     }
 
     fn put_keyed(&self, k: &Cid, block: &[u8]) -> anyhow::Result<()> {
-        metrics::BLOCK_SIZE_BYTES.observe(block.len() as f64);
```
lemmih, line 147: We don't want this metric anymore?

lemmih (approved):

node/db/src/rocks.rs
hanabi1224: Good catch! fixed

hanabi1224:
@lemmih @LesnyRumcajs I have to also turn on auto compaction otherwise very small sst files in ~20KB are generated on every flush.

lemmih:
> @lemmih @LesnyRumcajs I have to also turn on auto compaction otherwise very small sst files in ~20KB are generated on every flush.

That will kill performance. :(

lemmih (changes requested):
Hmm, we need to find a solution that doesn't require compactions.

hanabi1224:
> we need to find a solution that doesn't require compactions.

@lemmih hmm, it's still using `NoCompaction` I assume the perf impact is not as significant as level compaction. I'm doing some testing on mainnet

lemmih:
> > we need to find a solution that doesn't require compactions.
> 
> @lemmih hmm, it's still using `NoCompaction` I assume the perf impact is not as significant as level compaction. I'm doing some testing on mainnet

The only thing `NoCompaction` does is disable the auto-compactions.

hanabi1224:
@lemmih Yes it does hurt car import performance, how about enabling auto compaction after car import?

lemmih:
> @lemmih Yes it does hurt car import performance, how about enabling auto compaction after car import?

You can give it a go. I'm not too optimistic, though.

node/db/src/rocks.rs
```diff
@@ -50,7 +59,8 @@ impl RocksDb {
             db_opts.set_compaction_style(compaction_style);
             db_opts.set_disable_auto_compactions(false);
         } else {
-            db_opts.set_disable_auto_compactions(true);
+            db_opts.set_disable_auto_compactions(false);
```
elmattic, line 62: If you enable compaction then the mode will still be LEVEL, as it is the default:
see: https://github.com/rust-rocksdb/rust-rocksdb/blob/v0.19.0/src/db_options.rs#L2082

Calling in both `if` and `else` branchs `set_disable_auto_compactions(false)` will make impossible to run forest without compactions.

elmattic: IOW this will make forest run always **with** LEVEL compactions.

hanabi1224: You are right! I should revert the change here

hanabi1224:
According to the [doc](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_disable_auto_compactions)
`fn set_disable_auto_compactions` is
> Dynamically changeable through SetOptions() API.

By introducing `optimize_for_offline_workload` and `optimize_for_online_workload`, car import should be using the no compaction options now, compaction is enabled after car import I will update tipset validation metrics after keeping daemon running for a while

lemmih:
> According to the [doc](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_disable_auto_compactions) `fn set_disable_auto_compactions` is
> 
> > Dynamically changeable through SetOptions() API.
> 
> By introducing `optimize_for_offline_workload` and `optimize_for_online_workload`, car import should be using the no compaction options now, compaction is enabled after car import I will update tipset validation metrics after keeping daemon running for a while

Doing a benchmark run will give us a definitive answer.

elmattic:
> By introducing `optimize_for_offline_workload` and `optimize_for_online_workload`, car import should be using the no compaction options now, compaction is enabled after car import I will update tipset validation metrics after keeping daemon running for a while

I get what you want you want to do, as I tried somewhat this approach in the past:

But I see a major problem:

Using `"none"` ie,
```
[rocks_db]
compaction_style = "none"
```
for compaction mode will be ignored in the online workload mode, and level compactions will be use again in online-mode (the most important mode since it's which forest will run 99% of its time).

This is a big problem because a user that want to avoid rocksdb to performs compactions will still have them, and all the drawbacks that comes with this mode, like space and write amplifications.

hanabi1224:
@lemmih @LesnyRumcajs  @elmattic  I've reverted this PR to its original version except that it doesn't call `flush` after each `validate_tipset` to avoid generating fragmented sst files.  That means data in memtable will only be flushed into sst when it hits `write_buffer_size` which is 1GB atm.  Not sure if this will cause any problem on mainnet, some manually testing are appreciated.

P.S.  I've tested partial data lost case on calibnet by setting `write_buffer_size` to 8MB for a few times, looks good so far. But still I'm not confident when there's 1GB partial data lost on mainnet

elmattic:
@hanabi1224

I'm having similar kind of errors with Forest+`paritydb` backend. See issue https://github.com/ChainSafe/forest/issues/2278.

Tying to figure out what is the root cause of those errors..

lemmih (approved):
This works for me. It also fixes the corrupted database issue that @creativcoder reported.

LesnyRumcajs (approved):

PR #2281, opened 2022-11-28
fix Grafana on MacOS

LesnyRumcajs:
**Summary of changes**
- change Grafana compose to work on both Linux and MacOS,
- listen on all interfaces by default for metrics endpoint. Security shouldn't be a concern here; it's up to the operator to configure the firewall properly.

Closes https://github.com/ChainSafe/forest/issues/2288

The issue stems from the fact that MacOS uses a thin VM for running docker. This doesn't work well with `host` network, and the ports should be exposed explicitly (which is not possible with `host` network). 

A way out of this is using `host.docker.internal`... which unfortunately exists only in Docker for MacOS (and probably in Docker for Windows). The way out of this is using a recently (2021) added `host-gateway`, see more https://github.com/docker/for-linux/issues/264#issuecomment-965465879. 

Tested on MacOS Ventura 13.01 and Linux Fedora Workstation 36.

### DIFF
--- a/.maintain/monitoring/README.md
+++ b/.maintain/monitoring/README.md
@@ -4,7 +4,7 @@
 
 1. `forest` node running locally
 2. `docker` & `docker-compose` must be available in the `$PATH`
-3. Ports `3000` (for `grafana`) and `9090` (for `prometheus`) have to be free.
+3. Ports `3000` for `grafana` have to be free.
 
 ## Run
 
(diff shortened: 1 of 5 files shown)


elmattic:
When this is RFR, I can test it and review it.

LesnyRumcajs:
@elmattic Feel free to to test it on your end, be sure to recompile Forest to get it to listen on all interfaces for the metrics endpoint.

elmattic (approved):
Works perfectly now!

LesnyRumcajs:
> Works perfectly now!

Thanks for looking into this and reporting it!

lemmih (approved):

PR #2291, opened 2022-11-29
chore: upgrade serde_ipld_dagcbor

hanabi1224:
**Summary of changes**
- [x] upgrade serde_ipld_dagcbor
- [x] remove `FramedRead` from `RequestResponseCodec` impl
- [x] manually test on calibnet
- [x] manually test on mainnet 

Issue: 
- ~~`Deserializer` has been made non-public, and it no longer allows trailing data left, we need to propose a `from_reader_unstrict` API to upstream.~~
- `Counted` wrapper stops working because `from_reader` takes `BufReader` now, re-implement the logic without `FramedRead`

Update: looking at [lotus code](https://github.com/filecoin-project/lotus/blob/0770bced090aef6e9feaf33fc42198984df489d2/chain/exchange/server.go#L44), they don't do framed read, we should remove that logic as well

libp2p [metrics on CI](https://github.com/ChainSafe/forest/actions/runs/3576663862/jobs/6014745529#step:12:2603) looks good:
```
# HELP libp2p_messsage_total Total number of libp2p messages by type
# TYPE libp2p_messsage_total counter
libp2p_messsage_total{libp2p_message_kind="bitswap_block_request_out"} 1
libp2p_messsage_total{libp2p_message_kind="bitswap_block_response_in"} 1
libp2p_messsage_total{libp2p_message_kind="chain_exchange_request_out"} 95
libp2p_messsage_total{libp2p_message_kind="chain_exchange_response_in"} 95
libp2p_messsage_total{libp2p_message_kind="hello_request_in"} 46
libp2p_messsage_total{libp2p_message_kind="hello_request_out"} 52
libp2p_messsage_total{libp2p_message_kind="hello_response_in"} 28
libp2p_messsage_total{libp2p_message_kind="hello_response_out"} 38
libp2p_messsage_total{libp2p_message_kind="peer_connected"} 63
libp2p_messsage_total{libp2p_message_kind="peer_disconnected"} 22
libp2p_messsage_total{libp2p_message_kind="pubsub_message_block"} 8
libp2p_messsage_total{libp2p_message_kind="pubsub_message_message"} 2
```

Closes https://github.com/ChainSafe/forest/issues/2022


node/forest_libp2p/src/rpc/cbor_codec.rs
```diff
@@ -111,39 +80,29 @@ impl<'de, Item: Deserialize<'de>> Default for Decoder<Item> {
...
-        let mut deserializer = Deserializer::new(reader);
-
-        match Item::deserialize(&mut deserializer) {
+        // let mut deserializer = Deserializer::new(reader);
+        match serde_ipld_dagcbor::de::from_reader_unstrict(reader) {
             // If we read the item, we also need to consume the corresponding bytes.
-            //
             Ok(item) => {
-                src.advance(pos);
+                let offset = fvm_ipld_encoding::to_vec(&item).expect("Infallible").len();
```
hanabi1224, line 100: counted wrapper does not work here either because `from_reader` now takes a `BufReader`

hanabi1224: nvm this file has been removed

lemmih:
Looks promising. Will take me a while to wrap my head around the code.

node/forest_libp2p/src/rpc/mod.rs
```diff
@@ -142,12 +118,45 @@ where
...
+
+async fn read_and_decode<IO, T>(io: &mut IO) -> io::Result<T>
+where
+    IO: AsyncRead + Unpin,
+    T: serde::de::DeserializeOwned,
+{
+    const MAX_BYTES_ALLOWED: usize = 2 * 1024 * 1024; // messages over 2MB are likely malicious
+    const TIMEOUT_SECS: u64 = 30;
+
+    // Currently the protocol does not send length encoded message,
```
hanabi1224, line 133: I believe currently if some malicious nodes send cbor encoded bytes + some random bytes in a single request/response, it will deadlock forest daemon because `FramedRead` internally ensures no trailing data, failure to detect end of frame will continue the loop and end up hitting deadlock at the next `io.poll_ready`

lemmih (approved):
Good. If it works then it works.

node/forest_libp2p/src/rpc/decoder.rs
```diff
@@ -0,0 +1,74 @@
...
+        loop {
+            let n = std::task::ready!(Pin::new(&mut self.io).poll_read(cx, &mut buf))?;
+            // Terminated
+            if n == 0 {
+                let item = serde_ipld_dagcbor::de::from_reader(&self.bytes[..])
+                    .map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()));
+                return Poll::Ready(item);
+            }
+            self.bytes_read += n;
+            if self.max_bytes_allowed > 0 && self.bytes_read > self.max_bytes_allowed {
```
LesnyRumcajs, line 55: Just making sure, are we guaranteed not to have integer overflow here?

node/forest_libp2p/src/rpc/decoder.rs
```diff
@@ -0,0 +1,74 @@
...
+                return Poll::Ready(Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    format!(
+                        "Buffer size exceeds the maximum allowed {}B",
+                        self.max_bytes_allowed,
+                    ),
+                )));
+            }
+            self.bytes.extend_from_slice(&buf[..n]);
```
LesnyRumcajs, line 64: should we add a failsafe in case `n` is bigger than the buffer size?

node/forest_libp2p/src/rpc/mod.rs
```diff
@@ -142,12 +118,45 @@ where
...
     }
 }
+
+async fn read_and_decode<IO, T>(io: &mut IO) -> io::Result<T>
+where
+    IO: AsyncRead + Unpin,
+    T: serde::de::DeserializeOwned,
+{
+    const MAX_BYTES_ALLOWED: usize = 2 * 1024 * 1024; // messages over 2MB are likely malicious
+    const TIMEOUT_SECS: u64 = 30;
```
LesnyRumcajs, line 131: Perhaps we could have duration strongly typed, given that https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs is `const`

node/forest_libp2p/src/rpc/mod.rs
```diff
@@ -142,12 +118,45 @@ where
...
+
+async fn encode_and_write<IO, T>(io: &mut IO, data: T) -> io::Result<()>
+where
+    IO: AsyncWrite + Unpin,
+    T: serde::Serialize,
+{
+    let bytes = fvm_ipld_encoding::to_vec(&data)
+        .map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()))?;
+    io.write_all(&bytes).await?;
+    io.close().await?;
```
LesnyRumcajs, line 160: do we need to close manually? If yes, what happens in case of an error on the `write_all`?

hanabi1224: I don't think it's needed, because the framework calls `io.close` already, but this follows the `libp2p` example: https://github.com/libp2p/rust-libp2p/blob/master/examples/file-sharing.rs#L734

node/forest_libp2p/src/rpc/decoder.rs
hanabi1224: fixed

node/forest_libp2p/src/rpc/mod.rs
hanabi1224: great idea, fixed

hanabi1224:
> Looks good to me, but it's a faith-driven code review. I would love more coverage, though I understand it may be challenging here. Or can we contrive something simple?

It will be great to have unittest coverage, created ticket https://github.com/ChainSafe/forest/issues/2359 tracking this

tyshko5 (approved):

Issue #2204, opened 2022-11-15
Cursor is lost after ctrl+c when prompt to download snapshot is active


LesnyRumcajs:
**Describe the bug**

In terminal (checked with both `alacritty` and in-built one), after interrupting (with ctrl+c) Forest while it's asking whether to download a snapshot, the cursor is not appearing again.

**To Reproduce**
1. Run fresh forest with e.g. `forest --chain calibnet`
2. Get asked whether to download snapshot
3. ctrl+c
4. the cursor is lost

**Expected behaviour**

After interrupting, the cursor should be back.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Fedora 36
 - Rust version(e.g. `rustc --version`)
 - Branch/commit f503eb5b88c706d0e7d50a0ffb4e0ec40f8e2df8

**Other information and links**

jdjaustin:
Discovered an even worse bug while troubleshooting this, at least on my system (Ubuntu 22.04.1): hitting ctrl+c twice while at the "Enter the keystore passphrase" prompt does restore the cursor, but the terminal is now unresponsive (both in VSCode and a standalone terminal).

PR #2296, opened 2022-11-30
Restore Cursor after ctrl+c when prompt to download snapshot is active

jdjaustin:
**Summary of changes**
- Cursor disappears after receiving ctrl+c interrupt when prompt to download snapshot is active. This PR aims to resolve this errant behavior and restore the cursor.

Closes issue #2204 

N/A


jdjaustin:
`cargo clippy` was not happy about using `.unwrap()` instead of `?`, but the compiler doesn't allow the use of `?` inside of a closure that returns `()`, and modifying the closure to return a `Result` is not feasible because of the trait bounds defined in the `ctrlc` crate. If I'm missing an alternate method to resolve this, please let me know.

forest/daemon/src/daemon.rs
```diff
@@ -427,15 +427,22 @@ async fn prompt_snapshot_or_die(config: &Config) -> bool {
...
+            if result {
+                true
+            } else {
+                cli_error_and_die("Forest cannot sync without a snapshot. Download a snapshot from a trusted source and import with --import-snapshot=[file] or --download-snapshot to download one automatically", 1);
+            }
+        }
+        Err(_) => {
+            std::mem::drop(should_download);
+            cli_error_and_die("Error getting input.", 1);
+        }
```
lemmih, line 445: Are these changes related to the cursor fix? If so, could you explain it in a comment. If not then you can delete this code.

lemmih (changes requested):

forest/daemon/src/daemon.rs
jdjaustin: Yes, this was originally simply part of my troubleshooting, but I just tested it without the match statement, and it is required to properly handle the prompt resuts including the case of an interrupt. I'll add a comment.

lemmih (approved):

hanabi1224 (approved):

PR #2300, opened 2022-11-30
fix: better way of caching cargo-installed bins

hanabi1224:
**Summary of changes**
- save some lines of config by omitting the cache paths
- automatically update the cache when new versions are installed
- `cargo install` no longer fails with `binary already exists` error

Cache is still useful to handle quickinstall itself and its fallback case

Not sure if it's better to remove `which [bin]` from `Makefile` so that it can receive upgrades since the new action can detect changes and update the cache now.

We currently use `actions/cache@v3` for caching cargo-install-ed binaries, but it does not support custom save conditions, when a new version of a tool is released, it will always perform full installation without updating the cache unless we manually delete it.

Closes https://github.com/ChainSafe/forest/issues/2305


Makefile
```diff
@@ -13,14 +13,14 @@ install-deps:
 	apt-get install --no-install-recommends -y build-essential clang protobuf-compiler ocl-icd-opencl-dev aria2 cmake
 
 install-lint-tools:
-	which cargo-quickinstall || cargo install cargo-quickinstall
+	cargo install cargo-quickinstall
```
LesnyRumcajs, line 16: this may still fail outside Github workflow, no?

hanabi1224: Normally `already installed` is not a failure, when $CARGO_HOME/.crates2.json presents, previously we did not cache that file, it's now fixed in the the action

LesnyRumcajs: Ah, that's the issue and the source of many recent PRs. The `cargo install` behaved differently on my machine. I'm okay with removing `which` if you are confident that there won't be a case where there's a binary but no JSON registry (causing an annoying installation fiasco).

LesnyRumcajs (approved):

tyshko5 (approved):

Makefile
hanabi1224: @LesnyRumcajs  I've done some testing, it seems quickinstall works fine with the cache action, will just include the removing `which [bin]` change in this PR since it's not yet merged

PR #2303, opened 2022-11-30
Replacement of `pbr` crate with `indicatif`

### LINKED ISSUE #2251: Investigate replacement of `pbr` crate
**Issue summary**

`pbr` crate has not been released for 2+ years, it has some issues like [`RUSTSEC-2020-0071`](https://github.com/rustsec/advisory-db/blob/main/crates/time/RUSTSEC-2020-0071.md) with its published versions. We should investigate possible replacement

tyshko5:
**Summary of changes**
- 

Closes #2251

### DIFF
--- a/utils/forest_utils/src/io/progress_bar.rs
+++ b/utils/forest_utils/src/io/progress_bar.rs
@@ -1,12 +1,12 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 use serde::{Deserialize, Serialize};
+use std::cell::RefCell;
 use std::str::FromStr;
 use std::sync::RwLock;
-use std::time::Duration;
-use std::{cell::RefCell, io::Stdout};
 
-pub use pbr::Units;
+use console::Term;
+use indicatif::{ProgressBar as IndicatifProgressBar, ProgressDrawTarget, ProgressStyle};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "lowercase")]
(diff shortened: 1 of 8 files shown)


utils/forest_utils/src/io/progress_bar.rs
```diff
@@ -42,47 +42,63 @@ static PROGRESS_BAR_VISIBILITY: RwLock<ProgressBarVisibility> =
...
         }
     }
 
-    pub fn set_units(&self, u: Units) {
+    pub fn set_bytes(&self) {
         if self.display {
-            self.inner.borrow_mut().set_units(u)
+            self.inner.borrow_mut().set_style(
+                ProgressStyle::with_template(
+                    "{spinner:.green} [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})",
```
lemmih, line 90: The previous progress bar looked like this:
`Downloading snapshot 107.42 MB / 423.67 MB [===>--------------] 25.35 % 35.79 MB/s 9s`

And the new progress bar looks like this:
`⠤ [##>------------] 75.60 MiB/423.67 MiB (9s)`

Could you change the format to include the message, the progress as a percentage, and the rate of progress?

utils/forest_utils/src/io/progress_bar.rs
```diff
@@ -42,47 +42,63 @@ static PROGRESS_BAR_VISIBILITY: RwLock<ProgressBarVisibility> =
...
 
     pub fn message(&self, message: &str) {
         if self.display {
-            self.inner.borrow_mut().message(message);
+            self.inner.borrow_mut().set_message(message.to_string());
         }
     }
 
-    pub fn set_max_refresh_rate(&self, w: Option<Duration>) {
+    pub fn set_max_refresh_rate(&mut self, w: u8) {
```
lemmih, line 68: It's not obvious to me that `w` is given in hertz. Maybe we could change the name of the function to `set_max_refresh_rate_in_hz`.

utils/forest_utils/src/net/download.rs
```diff
@@ -76,10 +75,10 @@ impl FetchProgress<DownloadStream> {
 
         let response = client.get(url.as_str().try_into()?).await?;
 
-        let pb = ProgressBar::new(total_size);
+        let mut pb = ProgressBar::new(total_size);
         pb.message("Downloading/Importing snapshot ");
-        pb.set_units(crate::io::progress_bar::Units::Bytes);
-        pb.set_max_refresh_rate(Some(Duration::from_millis(500)));
+        pb.set_bytes();
+        pb.set_max_refresh_rate(20);
```
lemmih, line 81: Updating every 500 milliseconds is the same as 2 hertz, not 20 hz. Do we need to update 20 times per second rather than just 2 times per second?

utils/forest_utils/src/io/progress_bar.rs
tyshko5: Ok. But it's actually true. Internally indicatif really divides 1000 ms by this number

utils/forest_utils/src/net/download.rs
tyshko5: Typo. Fixed

utils/forest_utils/src/io/progress_bar.rs
tyshko5: Fixed

utils/forest_utils/src/io/progress_bar.rs
tyshko5: Fixed accordingly

lemmih (approved):

hanabi1224 (approved):

PR #2306, opened 2022-12-01
fix: use correct nv17 actor CIDs

lemmih:
**Summary of changes**
- Use the actor CIDs from lotus-1.18.1 rather than lotus-1.18.0.

Closes 


lemmih:
This PR is required for validating tipsets after the NV17 upgrade which took effect yesterday.

elmattic (approved):

PR #2307, opened 2022-12-01
chore: update calibnet tipset checkpoint

creativcoder:
**Summary of changes**
- Move checkpoint closer to current snapshot head.

Closes 


LesnyRumcajs (approved):
I wonder if we could do this in a somehow more automated way.

creativcoder:
> We might wanna keep the old checkpoints. It'll help us load older snapshots when testing migrations.
> 
> Could you also add a new checkpoint for mainnet?

Done

creativcoder:
> I wonder if we could do this in a somehow more automated way.

Ah we could move these checkpoints to a file and read and update based on snapshot filename conditionally. I'll look into how feasible it would be. Also to note that it's not something we'd do frequently.

lemmih (approved):

lemmih:
Maybe we should make it part of the release checklist to add new checkpoints.

PR #2327, opened 2022-12-05
Propagate fatal errors from background tasks 

### LINKED ISSUE #2324: Propagate fatal errors from background tasks
**Issue summary**

Forest uses a bunch of background tasks but these tasks rarely propagate fatal errors to the main task. This causes fatal to put Forest into a non-working, zombie state where it keeps running but doesn't do any processing.

Steps to reproduce:
1. Download https://forest-snapshots.fra1.digitaloceanspaces.com/forest_snapshot_calibnet_2022-12-05_height_95741.car
2. Import the snapshot: `forest --chain calibnet --encrypt-keystore false --import-snapshot forest_snapshot_calibnet_2022-12-05_height_95741.car`
3. Notice that it complains about `Invalid tipset: no msg root with cid bafy2bzaceb7yrvmubpzoltkhytajykrgbhfpgmj4yqg4nlnwqzcsnfja4zvj6` and then stops validating tipsets.
4. The failure happens here: https://github.com/ChainSafe/forest/blob/1132fe205e9dc6f27b833d855f51631e5b8958c2/blockchain/message_pool/src/msgpool/msg_pool.rs#L242

@creativcoder has suggested a few ways of propagating fatal errors from background tasks to the main task:
- https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html
- Passing a Sender<ToParent> to the worker thread.
- https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notify_waiters

Tasks:
- [ ] Evaluate the different possible solutions and pick the one you feel is best.
- [ ] Propagate errors from the message pool to the main thread such that loading a broken snapshot fails with a proper error message.
- [ ] Identifier other background tasks that may fail without propagating their errors and open a new issue to track them.

tyshko5:
**Summary of changes**
- 

Closes #2324


forest/daemon/src/daemon.rs
```diff
@@ -377,6 +380,15 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
+        if let Ok(res_inner) = res {
+            if let Err(error_message) = res_inner {
+                let msg = format!("services failure: {}", error_message);
+                cli_error_and_die(msg, 1);
+            }
+        }
+    }
+
     // Block until ctrl-c is hit
     ctrlc_oneshot.await.unwrap();
```
lemmih, line 393: I don't think this works. You're no longer listening for ctrl-c.

lemmih (changes requested):

lemmih:
You still have to handle ctrl-c correctly.

Also, show the new output for running forest on the broken snapshot.

forest/daemon/src/daemon.rs
```diff
@@ -169,6 +169,7 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
             {
                 error!("Failed to initiate prometheus server: {e}");
             }
+            Ok(())
```
lemmih, line 172: Why ignore the error here?

forest/daemon/src/daemon.rs
```diff
@@ -337,6 +339,7 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
                 FOREST_VERSION_STRING.as_str(),
             )
             .await;
+            Ok(())
```
lemmih, line 342: Why ignore the error message here?

forest/daemon/src/daemon.rs
```diff
@@ -306,6 +307,7 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
     let sync_state = chain_muxer.sync_state_cloned();
     services.spawn(async {
         chain_muxer.await;
+        Ok(())
```
lemmih, line 310: Why ignore the error message here?

tyshko5:
> You still have to handle ctrl-c correctly.
> 
> Also, show the new output for running forest on the broken snapshot.

Trying to find a way to fix ctrl-c

forest/daemon/src/daemon.rs
tyshko5: I think I fixed it

forest/daemon/src/daemon.rs
tyshko5: It's bad prometheus server failed to initiate, but it seems to to be critical to Forest

forest/daemon/src/daemon.rs
tyshko5: Fixed

forest/daemon/src/daemon.rs
tyshko5: Fixed

lemmih:
I don't think that's a great way to deal with ctrl-c. You would be better off using `select!`.

tyshko5:
> I don't think that's a great way to deal with ctrl-c. You would be better off using `select!`.

Why? Currently the interface is unified recreating the old logic completely. With select as far as I understand the situation is back to what it was.

forest/daemon/src/daemon.rs
```diff
@@ -361,11 +363,6 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
 
     select! {
         () = sync_from_snapshot(&config, &state_manager).fuse() => {},
-        _ = ctrlc_oneshot => {
-            // Cancel all async services
-            services.shutdown().await;
-            return db;
-        },
```
lemmih, line 368: You can't just delete this code. It was here for a reason.

lemmih: The code is required for a graceful shutdown. We _do_ want a graceful shutdown. That's the entire point of the ctrl-c handler.

tyshko5:
> > > I don't think that's a great way to deal with ctrl-c. You would be better off using `select!`.
> > 
> > 
> > Why? Currently the interface is unified recreating the old logic completely. With select as far as I understand the situation is back to what it was.
> 
> Imagine if one of the background services terminates early. You would interpret that as a ctrl-c and quit Forest. It's not intuitive for Forest developers that ending a background thread early is the same as a ctrl-c. And it's also not intuitive behavior for an end-user.

In case of ctrl-c, I can still return a specific error message, removing the current else clause with `break`. 
Shouldn't that work, or I'm missing something?

lemmih:
> > > > I don't think that's a great way to deal with ctrl-c. You would be better off using `select!`.
> > > 
> > > 
> > > Why? Currently the interface is unified recreating the old logic completely. With select as far as I understand the situation is back to what it was.
> > 
> > 
> > Imagine if one of the background services terminates early. You would interpret that as a ctrl-c and quit Forest. It's not intuitive for Forest developers that ending a background thread early is the same as a ctrl-c. And it's also not intuitive behavior for an end-user.
> 
> In case of ctrl-c, I can still return a specific error message, removing the current else clause with `break`. Shouldn't that work, or I'm missing something?

Then, conceptually, the background services would have three possible return values: `Ok`, `Err`, and `CtrlCInterrupt`. I don't doubt you could make it work but I'm unconvinced that it is a good solution.

Why not just use `select!`? Then you wouldn't need to introduce any new return values and the control-flow would be much easier to see in the code.

forest/daemon/src/daemon.rs
tyshko5: We are still doing a graceful shutdown using the same commands in the end of the `start` function

lemmih: If ctrl-c is hit duing `sync_from_snapshot`, how do we do a graceful shutdown?

lemmih: Keep in mind that `sync_from_snapshot` may take 10 minutes to complete.

tyshko5: Reverted

forest/daemon/src/daemon.rs
tyshko5: Refixed using `select!`

forest/daemon/src/daemon.rs
```diff
@@ -377,8 +377,22 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
     services.spawn(p2p_service.run());
 
-    // Block until ctrl-c is hit
-    ctrlc_oneshot.await.unwrap();
+    loop {
+        select! {
+            option = services.join_next().fuse() => {
+                if let Some(res) = option {
+                    if let Ok(res_inner) = res {
+                        if let Err(error_message) = res_inner {
```
creativcoder: if we're not handling other clauses for the above two matches, I think we we can do: 
```
if let Some(Ok(Err(error_message)) = option {

}
```

forest/daemon/src/daemon.rs
lemmih: We already fail if bind to the port or find the DB path. Let's fail as eagerly as possible and exit Forest if something goes wrong in prometheus. It'll then be prometheus' responsibility to properly deal with non-fatal errors.

forest/daemon/src/daemon.rs
```diff
@@ -377,8 +377,20 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
+                    let msg = format!("services failure: {}", error_message);
+                    error!("{}", msg);
+                    break
+                }
+            },
+            complete => select! {
+                _ = ctrlc_oneshot => break
+            },
+        }
+    }
```
lemmih, line 393: I'm 90% sure that this code does not work.

blockchain/chain_sync/src/consensus.rs
```diff
@@ -111,7 +111,7 @@ pub trait Proposer {
         state_manager: Arc<StateManager<DB>>,
         mpool: Arc<MP>,
         submitter: SyncGossipSubmitter,
-        services: &mut JoinSet<()>,
+        services: &mut JoinSet<Result<(), String>>,
```
hanabi1224, line 114: how about `anyhow::Result<()>`?

forest/daemon/src/daemon.rs
tyshko5: Fixed accordingly

forest/daemon/src/daemon.rs
tyshko5: Changed this logic

blockchain/chain_sync/src/consensus.rs
tyshko5: Fixed accordingly

forest/daemon/src/daemon.rs
```diff
@@ -167,7 +167,12 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
                 forest_metrics::init_prometheus(prometheus_listener, db_directory, db).await
             {
-                error!("Failed to initiate prometheus server: {e}");
+                Err(anyhow::anyhow!(
+                    "{}",
+                    format!("Failed to initiate prometheus server: {e}")
+                ))
+            } else {
+                Ok(())
             }
```
lemmih, line 176: This can be refactored.

forest/daemon/src/daemon.rs
```diff
@@ -377,8 +381,28 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
+            option = services.join_next().fuse() => {
+                if let Some(Ok(Err(error_message))) = option {
+                    let msg = format!("services failure: {}", error_message);
+                    error!("{}", msg);
+                    break
+                }
+            },
+            _ = ctrlc_oneshot => break
+        }
+    }
```
lemmih, line 399: This logic is still wrong. We want to block until an error has occurred or ctrl-c has been pressed. Let's discuss if this is still unclear.

node/forest_libp2p/src/service.rs
```diff
@@ -228,12 +228,9 @@ where
...
         let mut interval =
             IntervalStream::new(tokio::time::interval(Duration::from_secs(15))).fuse();
-        let pubsub_block_str = format!("{}/{}", PUBSUB_BLOCK_STR, self.network_name);
-        let pubsub_msg_str = format!("{}/{}", PUBSUB_MSG_STR, self.network_name);
 
-        let mut hello_request_table = HashMap::new();
-        let mut cx_request_table = HashMap::new();
-        let mut outgoing_bitswap_query_ids = HashMap::new();
+        let (mut hello_request_table, mut cx_request_table, mut outgoing_bitswap_query_ids) =
+            (HashMap::new(), HashMap::new(), HashMap::new());
```
lemmih, line 233: Try changing as few lines as possible.

blockchain/chain_sync/src/consensus.rs
```diff
@@ -111,7 +111,7 @@ pub trait Proposer {
         state_manager: Arc<StateManager<DB>>,
         mpool: Arc<MP>,
         submitter: SyncGossipSubmitter,
-        services: &mut JoinSet<()>,
+        services: &mut JoinSet<Result<(), anyhow::Error>>,
```
lemmih, line 114:
```suggestion
        services: &mut JoinSet<anyhow::Result<()>>,
```

tyshko5: Fixed

forest/daemon/src/daemon.rs
```diff
@@ -392,6 +403,24 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
 
+async fn propagate_error(services: &mut JoinSet<Result<(), anyhow::Error>>) -> anyhow::Error {
+    while !services.is_empty() {
+        select! {
+            option = services.join_next().fuse() => {
+                if let Some(Ok(Err(error_message))) = option {
+                    return error_message
+                } else {
+                    continue
+                }
```
lemmih, line 414: No need for `continue` here.

node/forest_libp2p/src/service.rs
```diff
@@ -258,7 +258,7 @@ where
                             &pubsub_block_str,
                             &pubsub_msg_str,).await;
                     },
-                    None => { break; },
+                    None => { break Ok(()); },
```
lemmih, line 261: How about moving the `Ok(())` such that you only write it once?

blockchain/consensus/deleg_cns/src/composition.rs
```diff
@@ -27,7 +27,7 @@ pub async fn consensus<DB, MP>(
     keystore: &Arc<RwLock<KeyStore>>,
     mpool: &Arc<MP>,
     submitter: SyncGossipSubmitter,
-    services: &mut JoinSet<()>,
+    services: &mut JoinSet<Result<(), anyhow::Error>>,
```
lemmih, line 30: Be consistent in using `anyhow::Result<()>`

forest/daemon/src/daemon.rs
tyshko5: Done

node/forest_libp2p/src/service.rs
tyshko5: Done

forest/daemon/src/daemon.rs
tyshko5: Removed

blockchain/consensus/deleg_cns/src/proposer.rs
```diff
@@ -102,15 +102,20 @@ impl Proposer for DelegatedProposer {
...
         services.spawn(async move {
             if let Err(e) = self.run(state_manager, mpool.as_ref(), &submitter).await {
-                error!("block proposal stopped: {}", e)
+                Err(anyhow::anyhow!(
+                    "{}",
+                    format!("block proposal stopped: {}", e)
+                ))
+            } else {
+                Ok(())
             }
```
lemmih, line 119: See `map_err` or just use '?`.

blockchain/message_pool/src/msgpool/msg_pool.rs
```diff
@@ -239,13 +239,20 @@ where
...
                         .await
-                        .unwrap_or_else(|err| warn!("Error changing head: {:?}", err));
+                        .map_err(|err| err.to_string());
+
+                        if let Err(err) = res {
+                            break Err(anyhow::anyhow!(
+                                "{}",
+                                format!("Error changing head: {:?}", err)
+                            ));
+                        }
```
lemmih, line 249: Can this be refactored to be more idiomatic? Can you use `?`?

forest/daemon/src/daemon.rs
```diff
@@ -164,11 +166,11 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
         services.spawn(async {
-            if let Err(e) =
-                forest_metrics::init_prometheus(prometheus_listener, db_directory, db).await
-            {
-                error!("Failed to initiate prometheus server: {e}");
-            }
+            forest_metrics::init_prometheus(prometheus_listener, db_directory, db)
+                .await
+                .map_err(|e| {
+                    anyhow::anyhow!("{}", format!("Failed to initiate prometheus server: {e}"))
```
lemmih, line 172: Avoid formatting twice.

blockchain/consensus/deleg_cns/src/proposer.rs
```diff
@@ -102,15 +102,20 @@ impl Proposer for DelegatedProposer {
...
     where
         DB: Blockstore + Store + Clone + Sync + Send + 'static,
         MP: MessagePoolApi + Send + Sync + 'static,
     {
         services.spawn(async move {
             if let Err(e) = self.run(state_manager, mpool.as_ref(), &submitter).await {
-                error!("block proposal stopped: {}", e)
+                Err(anyhow::anyhow!(
+                    "{}",
+                    format!("block proposal stopped: {}", e)
```
lemmih, line 115: Avoid formatting twice.

blockchain/message_pool/src/msgpool/msg_pool.rs
```diff
@@ -239,13 +239,20 @@ where
...
                     }
                     Err(RecvError::Lagged(e)) => {
                         warn!("Head change subscriber lagged: skipping {} events", e);
                     }
                     Err(RecvError::Closed) => {
-                        break;
+                        break Ok::<(), anyhow::Error>(());
```
lemmih, line 255: Is the type annotation required?

blockchain/message_pool/src/msgpool/selection.rs
```diff
@@ -693,7 +693,9 @@ mod test_selection {
 
     const TEST_GAS_LIMIT: i64 = 6955002;
 
-    async fn make_test_mpool(joinset: &mut JoinSet<()>) -> MessagePool<TestApi> {
+    async fn make_test_mpool(
+        joinset: &mut JoinSet<Result<(), anyhow::Error>>,
```
lemmih, line 697: Be consistent about `anyhow::Result`.

node/forest_libp2p/src/service.rs
```diff
@@ -216,7 +216,7 @@ where
     }
 
     /// Starts the libp2p service networking stack. This Future resolves when shutdown occurs.
-    pub async fn run(mut self) {
+    pub async fn run(mut self) -> Result<(), anyhow::Error> {
         info!("Running libp2p service");
         Swarm::listen_on(&mut self.swarm, self.config.listening_multiaddr).unwrap();
```
lemmih, line 221: Can we remove this `unwrap`?

blockchain/consensus/deleg_cns/src/proposer.rs
tyshko5: Fixed

forest/daemon/src/daemon.rs
```diff
@@ -392,6 +398,22 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
+    while !services.is_empty() {
+        select! {
+            option = services.join_next().fuse() => {
+                if let Some(Ok(Err(error_message))) = option {
+                    return error_message
+                }
+            },
+        }
+    }
+    sleep(Duration::new(3200000000, 0));
```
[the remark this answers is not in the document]
tyshko5: Changed to `tokio` analogue

forest/daemon/src/daemon.rs
```diff
@@ -377,8 +378,13 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
 
     services.spawn(p2p_service.run());
 
-    // Block until ctrl-c is hit
-    ctrlc_oneshot.await.unwrap();
+    select! {
+        err = propagate_error(&mut services).fuse() => {
+            let msg = format!("services failure: {}", err);
```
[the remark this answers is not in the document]
tyshko5: Fixed

forest/daemon/src/daemon.rs
tyshko5: Fixed

blockchain/consensus/deleg_cns/src/proposer.rs
tyshko5: Fixed

blockchain/message_pool/src/msgpool/selection.rs
tyshko5: Fixed

node/forest_libp2p/src/service.rs
tyshko5: Removed

blockchain/message_pool/src/msgpool/msg_pool.rs
```diff
@@ -239,13 +239,13 @@ where
                             app,
                         )
                         .await
-                        .unwrap_or_else(|err| warn!("Error changing head: {:?}", err));
+                        .map_err(|err| anyhow::anyhow!("Error changing head: {:?}", err))?;
```
lemmih, line 242: [context](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context) looks appropriate here.

forest/daemon/src/daemon.rs
```diff
@@ -143,7 +144,7 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
 
     let db = open_db(&config);
 
-    let mut services = JoinSet::new();
+    let mut services: JoinSet<Result<(), anyhow::Error>> = JoinSet::new();
```
lemmih, line 147: `anyhow::Result`?

forest/daemon/src/daemon.rs
```diff
@@ -164,11 +165,9 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
         let db = db.clone();
         services.spawn(async {
-            if let Err(e) =
-                forest_metrics::init_prometheus(prometheus_listener, db_directory, db).await
-            {
-                error!("Failed to initiate prometheus server: {e}");
-            }
+            forest_metrics::init_prometheus(prometheus_listener, db_directory, db)
+                .await
+                .map_err(|e| anyhow::anyhow!("Failed to initiate prometheus server: {e}"))
```
lemmih, line 170: See [context](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context)

forest/daemon/src/daemon.rs
```diff
@@ -392,6 +394,22 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
+    while !services.is_empty() {
+        select! {
+            option = services.join_next().fuse() => {
+                if let Some(Ok(Err(error_message))) = option {
+                    return error_message
+                }
+            },
+        }
+    }
+    tokio::time::sleep(Duration::new(3200000000, 0)).await;
```
lemmih, line 407: Does `tokio::time::sleep` have a max sleeping duration? What happens if you exceed the max duration?

forest/daemon/src/daemon.rs
```diff
@@ -392,6 +394,22 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
+    anyhow::Error::msg(
```
lemmih, line 408: `bail!`?

node/forest_libp2p/src/service.rs
```diff
@@ -216,9 +216,10 @@ where
     }
 
     /// Starts the libp2p service networking stack. This Future resolves when shutdown occurs.
-    pub async fn run(mut self) {
+    pub async fn run(mut self) -> Result<(), anyhow::Error> {
```
lemmih, line 219: `anyhow::Result`?

node/forest_libp2p/src/service.rs
```diff
@@ -216,9 +216,10 @@ where
...
         info!("Running libp2p service");
-        Swarm::listen_on(&mut self.swarm, self.config.listening_multiaddr).unwrap();
+        Swarm::listen_on(&mut self.swarm, self.config.listening_multiaddr)
+            .map_err(|e| anyhow::anyhow!("{e}"))?;
```
lemmih, line 222: Is this needed?

forest/daemon/src/daemon.rs
```diff
@@ -377,8 +375,12 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
 
     services.spawn(p2p_service.run());
 
-    // Block until ctrl-c is hit
-    ctrlc_oneshot.await.unwrap();
+    select! {
+        err = propagate_error(&mut services).fuse() => {
+            error!("services failure: {}", err);
+        },
```
lemmih, line 381:
```suggestion
        err = propagate_error(&mut services).fuse() => error!("services failure: {err}"),
```

lemmih (changes requested):

node/forest_libp2p/src/service.rs
tyshko5: Removed

blockchain/message_pool/src/msgpool/msg_pool.rs
tyshko5: Removed

forest/daemon/src/daemon.rs
tyshko5: Fixed

forest/daemon/src/daemon.rs
tyshko5: Approximately 2.2 years. I fixed our time to be within the limit

forest/daemon/src/daemon.rs
tyshko5: Fixed accordingly

forest/daemon/src/daemon.rs
tyshko5: Fixed accordingly

forest/daemon/src/daemon.rs
lemmih: `bail!` is shorter. Doesn't matter much.

blockchain/consensus/deleg_cns/src/proposer.rs
```diff
@@ -102,18 +102,18 @@ impl Proposer for DelegatedProposer {
...
         DB: Blockstore + Store + Clone + Sync + Send + 'static,
         MP: MessagePoolApi + Send + Sync + 'static,
     {
         services.spawn(async move {
-            if let Err(e) = self.run(state_manager, mpool.as_ref(), &submitter).await {
-                error!("block proposal stopped: {}", e)
-            }
+            self.run(state_manager, mpool.as_ref(), &submitter)
+                .await
+                .map_err(|e| anyhow::anyhow!("block proposal stopped: {}", e))
```
lemmih, line 114: Can `context` be used here?

blockchain/consensus/deleg_cns/src/proposer.rs
```diff
@@ -102,18 +102,18 @@ impl Proposer for DelegatedProposer {
...
         });
-        Ok(())
+        anyhow::Ok(())
```
lemmih, line 116: Is this necessary?

forest/daemon/src/daemon.rs
```diff
@@ -143,7 +145,7 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
 
     let db = open_db(&config);
 
-    let mut services = JoinSet::new();
+    let mut services: JoinSet<anyhow::Result<()>> = JoinSet::new();
```
lemmih, line 148: Is this type signature necessary?

forest/daemon/src/daemon.rs
```diff
@@ -392,6 +393,22 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
     db
 }
 
+async fn propagate_error(services: &mut JoinSet<Result<(), anyhow::Error>>) -> anyhow::Error {
```
lemmih, line 396: Document this function.

forest/daemon/src/daemon.rs
lemmih: I don't see your fix. Please push your changes before marking issues as resolved. That makes it much easier for reviewers to keep track of things.

forest/daemon/src/daemon.rs
tyshko5: Removed

blockchain/consensus/deleg_cns/src/proposer.rs
tyshko5: Yes, as `Proposer` trait requires `anyhow::Result` as return type

blockchain/consensus/deleg_cns/src/proposer.rs
tyshko5: Fixed

forest/daemon/src/daemon.rs
```diff
@@ -377,8 +376,10 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
 
     services.spawn(p2p_service.run());
 
-    // Block until ctrl-c is hit
-    ctrlc_oneshot.await.unwrap();
+    select! {
```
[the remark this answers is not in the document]
tyshko5: Done

forest/daemon/src/daemon.rs
tyshko5: Done

lemmih (approved):
It would be nice to return error code 1 when something goes wrong but let's do that in another PR.

hanabi1224 (approved):

lemmih:
> In which situations are we entering this code:
> 
> ```
>     tokio::time::sleep(Duration::new(64000000, 0)).await;
>     anyhow::Error::msg(
>         "More than 2 years have passed, all services are down and Forest was still running",
>     )
> ```
> 
> I'm not sure to understand the purpose of it.

The function should wait for errors. If there are no running services then it should wait forever. But there's no function for sleeping forever in tokio, I believe. So we just wait 2 years instead. Some code comments to explain this would be nice. Could even loop forever to avoid problems if someone wants to run Forest for more than 2 years.

elmattic:
> If there are no running services then it should wait forever.

That's what I don't understand, in which circumstances can you end up in a state in no services at all have been spawn?
We should always have for instance `ChainMuxer` no?

lemmih:
> > If there are no running services then it should wait forever.
> 
> That's what I don't understand, in which circumstances can you end up in a state in no services at all have been spawn? We should always have for instance `ChainMuxer` no?

Right now, yes, but that may not be true in the future. One could possibly imagine that the light-client wouldn't run any background services.

lemmih (approved):

forest/daemon/src/daemon.rs
```diff
@@ -392,6 +395,27 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
...
+                }
+            },
+        }
+    }
+    // In case all services are down without errors we are still willing
+    // to wait indefinitely for CTRL-C signal. As `tokio::time::sleep` has
+    // a limit of approximately 2.2 years we have to loop
+    loop {
+        tokio::time::sleep(Duration::new(64000000, 0)).await;
+    }
```
LesnyRumcajs, line 416: I think after 2 years of waiting we are fine to retire this process...

forest/daemon/src/daemon.rs
```diff
@@ -392,6 +395,27 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
     db
 }
 
+// returns the first error with which any of the services end
+// in case all services finished without an error sleeps for more than 2 years
+// and then returns with an error
```
LesnyRumcajs, line 400: this is not true, no? I mean, if I understand the code correctly, then if all services finished without an error then nothing is returned, ever. Or am I missing something?

forest/daemon/src/daemon.rs
tyshko5: David explains in the upper comments that a situation in which we should wait indefinitely

forest/daemon/src/daemon.rs
tyshko5: Yes, we are waiting for CTRL-C to be pressed, otherwise we are waiting indefinitely

PR #2342, opened 2022-12-08
add validate snapshot cli command

### LINKED ISSUE #2325: Implement `forest-cli snapshot validate` command.
**Issue summary**

It has happened on a few occasions that `forest-cli snapshot export` has generated an invalid snapshot with an IPLD link to a missing node. We should extend `forest-cli snapshot` with a command for validating the links in the snapshot.

Background: A CAR file is a key-value database where the keys are the hashes of the values. The values are usually IPLD nodes but may be any kind of data. The codec of the key determines if the value is an IPLD node or not. A CAR file also contains a reference to the most recent block header. A CAR file will contain block headers all the way back to genesis (eg. if it's a snapshot at epoch 10,000 then there will be 10,000 block headers). However, only the last 2,000 block headers will contain working links. The data linked to by older block headers is not included in the snapshots. This is controlled by the `--recent-stateroots` flag to `forest-cli snapshot export`.

Possible approach:
- [x] Load the snapshot into a temporary database. Possibly warn the user about the space required for this operation.
- [x] Get the reference from the CAR file to the most recent block header.
- [x] Verify that all IPLD links reachable from the 2,000 (controllable by a flag) most recent block headers are all present in the database.
- [x] Verify that all block headers are present all the way to the genesis block.

IPLD walker for exporting snapshots: https://github.com/ChainSafe/forest/blob/1132fe205e9dc6f27b833d855f51631e5b8958c2/blockchain/chain/src/store/chain_store.rs#L659

Library for loading CAR files: https://docs.rs/fvm_ipld_car/0.5.0/fvm_ipld_car/index.html

creativcoder:
**Summary of changes**
Introduces a snapshot validate subcommand.

- [X] Implements validate subcommand in snapshot flag.
- [X] Loads snapshot in a temporary db with a prompt for disk usage
- [X] Gets the most recent header and asserts number of tipsets in the snapshot
- [x] Validate ipld links in db based on `validate_height` flag.

Closes 
#2325


forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +347,105 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    let mut ts = ts.parents().cids()[0];
+    loop {
+        let data = db.get(&ts).unwrap().unwrap();
+        let h = BlockHeader::unmarshal_cbor(&data).unwrap();
+        blocks_count += 1;
+        if h.epoch() == 0 {
+            break;
+        }
+        let parents = h.parents().clone();
+        ts = parents.cids()[0];
```
lemmih, line 433: How about using `load_tipset` so you don't have to deal with CBOR?

creativcoder: Will look into it

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,123 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+            config.client.genesis_file.as_ref(),
+            config.chain.genesis_bytes(),
+            &chain_store,
+        )
+        .await
+        .unwrap();
+        let genesis_header = &genesis.blocks()[0];
+        chain_store.set_genesis(genesis_header).unwrap();
+
+        let sm = StateManager::new(
```
lemmih, line 388: Don't think you need a `StateManager`. You don't seem to be using it at all.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,123 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+{
+    let mut genesis_reachable = false;
+    let mut seen = std::collections::HashSet::<Cid>::new();
+    let upto = ts.epoch() - upto;
+
+    println!("validating ipld links upto: {}", upto);
+
+    let mut ts = ts.parents().cids()[0];
+
+    loop {
```
lemmih, line 440: I'm not sure I understand this control flow. How can this loop end without `genesis_reachable` being set to `true`?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,123 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+        )
+        .await
+        .unwrap();
+
+        let state_manager = Arc::new(sm);
+
+        let cids = {
+            let file = File::open(&snapshot).await?;
+            let reader = FetchProgress::fetch_from_file(file).await?;
+            load_and_retrieve_header(state_manager.blockstore(), reader, false).await?
```
lemmih, line 401: It's probably better just to call `load_car` here.

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: Done

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: Removed

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: Using early return from failure to get `tipset_from_keys`

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: Done

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,125 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+            "This will result in using approximately {} MB of data. Proceed?",
+            std::fs::metadata(snapshot).unwrap().len() / (1024 * 1024)
+        ))
+        .default(false)
+        .interact()
+        .unwrap_or_default();
+
+    if confirm {
+        // clear temp db if exists
+        let tmp_db_path = PathBuf::from("/tmp/validate_snapshot");
```
lemmih, line 367: You can use the `tempfile` crate to get a temporary directory which will be deleted when the handle is dropped.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,125 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
 
+async fn validate(
+    config: &Config,
+    validate_height: &i64,
+    snapshot: &PathBuf,
+) -> anyhow::Result<()> {
+    let confirm = Confirm::with_theme(&ColorfulTheme::default())
+        .with_prompt(format!(
+            "This will result in using approximately {} MB of data. Proceed?",
+            std::fs::metadata(snapshot).unwrap().len() / (1024 * 1024)
```
lemmih, line 359: Can use `?` instead of `unwrap`.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,125 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+            let file = tokio::fs::File::open(&snapshot).await?;
+            let reader = FetchProgress::fetch_from_file(file).await?;
+            load_car(&chain_store.db, reader.compat()).await?
+        };
+
+        let ts = chain_store.tipset_from_keys(&TipsetKeys::new(cids)).await?;
+
+        let snapshot_head_epoch = ts.epoch();
+
+        chain_store.db.flush()?;
```
lemmih, line 397: The database isn't used for persistence so there's no need to flush the data.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,124 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+            break;
+        }
+
+        if height > upto {
+            let mut assert_cid_exists = |cid: Cid| async move {
+                let data = db.get(&cid);
+                data.transpose().expect(&format!(
+                    "could not find data for cids in tipset at height: {}",
+                    height
+                ))
```
lemmih, line 447: Panics are not a good way of communicating with users. Software should only panic if there's a bug. But a broken IPLD link isn't a bug in the validator so a panic is not appropriate here.

You can use `?` and `ok_or` to get more intuitive behavior.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,124 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    );
+
+    let mut ts = ts.parents().clone();
+
+    loop {
+        let h = chain_store.tipset_from_keys(&ts).await?;
+        let height = h.epoch();
+
+        if height == 0 {
+            break;
```
lemmih, line 438: Let's verify that we reached the expected genesis tipset. Ie. validating a calibnet snapshot as if it was a mainnet snapshot should fail.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,124 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+
+    println!(
+        "validating cids from snapshot in db: {} <- {} ...",
+        upto,
+        ts.epoch()
+    );
+
+    let mut ts = ts.parents().clone();
+
+    loop {
```
lemmih, line 433: Recursive snapshots are difficult to create but not impossible. Let's limit the amount of recursion we're doing. We want to do at most `ts.epoch()` steps and advance by at least 1 epoch per iteration.

forest/cli/Cargo.toml
```diff
@@ -9,8 +9,10 @@ edition = "2021"
 anyhow.workspace = true
 base64.workspace = true
 cid.workspace = true
+dialoguer.workspace = true
 directories.workspace = true
 fil_actor_miner_v8.workspace = true
+forest-daemon = { path = "../daemon" }
```
lemmih, line 15: Is this dependency necessary?

creativcoder: Removed

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +348,124 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+            recurse_links(
+                &mut seen,
+                *h.blocks()[0].state_root(),
+                &mut assert_cid_exists,
+            )
+            .await?;
+        }
+
+        let header = h.blocks();
+        ts = header[0].parents().clone();
```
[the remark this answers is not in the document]
creativcoder: Added

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: ah nice, added

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: done

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: ok_or it is

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+        )
+        .await
+        .unwrap();
+        let genesis_header = &genesis.blocks()[0];
+        chain_store.set_genesis(genesis_header).unwrap();
+
+        let cids = {
+            let file = tokio::fs::File::open(&snapshot).await?;
+            let reader = FetchProgress::fetch_from_file(file).await?;
+            load_car(&chain_store.db, reader.compat()).await?
```
lemmih, line 389:
```suggestion
            load_car(chain_store.blockstore(), reader.compat()).await?
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+        };
+
+        let ts = chain_store.tipset_from_keys(&TipsetKeys::new(cids)).await?;
+
+        validate_links_and_genesis_traversal(
+            &chain_store,
+            ts,
+            &chain_store.db,
```
lemmih, line 397:
```suggestion
            chain_store.blockstore(),
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+
+async fn validate_links_and_genesis_traversal<DB>(
+    chain_store: &ChainStore<DB>,
+    ts: Arc<Tipset>,
+    db: &DB,
+    upto: ChainEpoch,
+    genesis_tipset: &Tipset,
+) -> anyhow::Result<()>
+where
+    DB: fvm_ipld_blockstore::Blockstore + Store + Send + Sync + Clone,
```
lemmih, line 415:
```suggestion
    DB: fvm_ipld_blockstore::Blockstore + Store + Send + Sync,
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+        if iteration_guard <= 0 {
+            break false;
+        }
+
+        let tipset = chain_store.tipset_from_keys(&tsk).await?;
+        if iteration_guard - tipset.epoch() < 1 {
+            bail!(
+                "tipset iteration didn't make progress backwards, possible recursion in snapshot."
+            );
+        }
```
lemmih, line 441: I don't understand this logic. Could you explain it to me or replace it with something so simple it requires no explanation?

creativcoder: We start with this invariant: `let mut iteration_guard = ts.epoch() + 1;` outside of loop; that `ts.epoch()` is always smaller than the guard by atleast 1. 

and if it violates (which can only happen when tipset.epoch()'s value didn't decrease by parent() calls), we bail.

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: added

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    );
+
+    let mut tsk = ts.parents().clone();
+
+    // Security: Recursive snapshots are difficult to create but not impossible. This
+    // limits the amount of recursion we do.
+    let mut iteration_guard = ts.epoch() + 1;
+    let genesis_reachable = loop {
+        if iteration_guard <= 0 {
+            break false;
```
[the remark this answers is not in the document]
creativcoder: done

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+                bail!("could not reach genesis from snapshot, genesis tipset in snapshot doesn't match the reference");
+            }
+
+            break true;
+        }
+
+        if height > upto {
+            let mut assert_cid_exists = |cid: Cid| async move {
+                let data = db.get(&cid);
+                data.transpose().ok_or(anyhow::anyhow!(
```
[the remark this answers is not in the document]
creativcoder: yeah, was no longer applicable, removed

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: added

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: added

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -111,6 +122,14 @@ pub enum SnapshotCommands {
         #[structopt(long)]
         force: bool,
     },
+    /// Validates the snapshot.
+    Validate {
+        /// Number of block headers to validate from the tip
+        #[structopt(long)]
+        validate_height: i64,
```
lemmih, line 129: How about we name this `recent_stateroots` to match the argument in the export command? It should also default to 2k.

forest/cli/src/cli/snapshot_cmd.rs
lemmih: How about keeping track of the previous epoch and verifying the new epoch is smaller?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+            }
+        }
+
+        tsk = tipset.parents().clone();
+        iteration_guard -= 1;
+    };
+
+    if genesis_reachable {
+        println!("genesis is reachable and valid.");
+    }
```
lemmih, line 472: `genesis_reachable` is always `true`.

utils/genesis/src/lib.rs
```diff
@@ -193,7 +193,7 @@ where
 }
 
 /// Loads car file into database, and returns the block header CIDs from the CAR header.
-async fn load_and_retrieve_header<DB, R>(
+pub async fn load_and_retrieve_header<DB, R>(
```
lemmih, line 196: Doesn't need to be `pub` anymore.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,133 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+                bail!("could not reach genesis from snapshot, genesis tipset in snapshot doesn't match the reference");
+            }
+
+            break true;
+        }
+
+        if height > upto {
+            let mut assert_cid_exists = |cid: Cid| async move {
+                let data = db.get(&cid);
+                data?.ok_or(anyhow::anyhow!(
```
lemmih, line 455: It's a bit more idiomatic to put the `?` after `db.get(&cid)`.

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: added

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: removed

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: removed

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,128 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    DB: fvm_ipld_blockstore::Blockstore + Store + Send + Sync,
+{
+    let mut seen = std::collections::HashSet::<Cid>::new();
+    let upto = ts.epoch() - recent_stateroots;
+
+    println!(
+        "validating cids from snapshot in db: {} <- {} ...",
+        upto,
+        ts.epoch()
+    );
```
lemmih, line 424: How about we use a progress bar instead of this message?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,128 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+                recurse_links(&mut seen, *h.state_root(), &mut assert_cid_exists).await?;
+            }
+        }
+
+        tsk = tipset.parents().clone();
+        prev_epoch = tipset.epoch();
+    }
+
+    println!("genesis is reachable and valid.");
+    println!("ok");
```
lemmih, line 469: Let's hide some of the details and just tell the user that the snapshot is valid.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,128 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
     Ok(())
 }
 
+async fn validate(
+    config: &Config,
+    recent_stateroots: &i64,
+    snapshot: &PathBuf,
+) -> anyhow::Result<()> {
+    let confirm = Confirm::with_theme(&ColorfulTheme::default())
```
lemmih, line 357: We should only prompt if stdin is a tty. And we also need some kind of flag to force the validation.

lemmih (changes requested):
Getting closer. Still need to hook it up in the CI to check that the calibnet snapshot is valid (and also check that validation fails if we try to validate it as a mainnet snapshot).

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,130 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+                data.ok_or_else(|| {
+                    anyhow::anyhow!(
+                        "could not find data for cids in tipset at height: {}",
+                        height
+                    )
+                })
+            };
+
+            for h in tipset.blocks() {
+                recurse_links(&mut seen, *h.state_root(), &mut assert_cid_exists).await?;
```
lemmih, line 462: You also need to check the links in `h.messages()`, btw.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,128 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+        .interact()
+        .unwrap_or_default();
+
+    if confirm {
+        let tmp_db_path = TempDir::new()?;
+
+        let db_path = tmp_db_path.path().join(&config.chain.name);
+
+        let db = forest_db::rocks::RocksDb::open(db_path, &config.rocks_db)
+            .expect("Opening RocksDB must succeed");
```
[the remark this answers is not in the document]
creativcoder: added

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +349,128 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+
+        let genesis = read_genesis_header(
+            config.client.genesis_file.as_ref(),
+            config.chain.genesis_bytes(),
+            &chain_store,
+        )
+        .await
+        .unwrap();
+        let genesis_header = &genesis.blocks()[0];
+        chain_store.set_genesis(genesis_header).unwrap();
```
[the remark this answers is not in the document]
creativcoder: removed

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: added

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: added

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +353,131 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+{
+    let mut seen = std::collections::HashSet::<Cid>::new();
+    let upto = ts.epoch() - recent_stateroots;
+
+    let mut tsk = ts.parents().clone();
+    // Security: Recursive snapshots are difficult to create but not impossible. This
+    // limits the amount of recursion we do.
+    let mut prev_epoch = ts.epoch();
+
+    let total_size = recent_stateroots;
```
lemmih, line 430: We have to scan through every tipset in the snapshot, not just the `recent_stateroots`. Scanning the recent stateroots is expensive but the cost is fixed. Scanning all the way to the genesis block gets more expensive as the snapshot grows.

Either include all the epochs in the progress bar or use two progress bars (one for the recent state roots and one for the genesis scanning).

lemmih (changes requested):

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: done

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: Went with the former

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +353,132 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    let upto = ts.epoch() - recent_stateroots;
+
+    let mut tsk = ts.parents().clone();
+    // Security: Recursive snapshots are difficult to create but not impossible. This
+    // limits the amount of recursion we do.
+    let mut prev_epoch = ts.epoch();
+
+    let total_size = ts.epoch();
+    let pb = forest_utils::io::ProgressBar::new(total_size as u64);
+    pb.message("Validating cids from snapshot in db ");
```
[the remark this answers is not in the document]
creativcoder: changed

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +353,132 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+
+    loop {
+        if prev_epoch <= 0 {
+            bail!("could not reach genesis from snapshot");
+        }
+
+        let tipset = chain_store.tipset_from_keys(&tsk).await?;
+        if tipset.epoch() >= prev_epoch {
+            bail!(
+                "tipset iteration didn't make progress backwards current epoch: {} previous_epoch: {}, possible recursion in snapshot.", tipset.epoch(), prev_epoch
```
lemmih, line 444:
```suggestion
                "Broken tipset invariant: parent epoch larger than current epoch at: {height}"
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -215,6 +237,14 @@ impl SnapshotCommands {
                 snapshot_dir,
                 force,
             } => clean(&config, snapshot_dir, *force),
+            Self::Validate {
+                recent_stateroots,
+                snapshot,
+                force,
+            } => {
+                validate(&config, recent_stateroots, snapshot, *force).await?;
+                Ok(())
```
lemmih, line 246:
```suggestion
                validate(&config, recent_stateroots, snapshot, *force).await
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +353,127 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    let mut prev_epoch = ts.epoch();
+
+    let total_size = ts.epoch();
+    let pb = forest_utils::io::ProgressBar::new(total_size as u64);
+    pb.message("Validating tipsets");
+    pb.set_max_refresh_rate(Some(std::time::Duration::from_millis(500)));
+
+    loop {
+        if prev_epoch <= 0 {
+            bail!("could not reach genesis from snapshot");
```
lemmih, line 438:
```suggestion
            bail!("Broken invariant: no genesis tipset in snapshot.");
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +353,127 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+                recurse_links(&mut seen, *h.messages(), &mut assert_cid_exists).await?;
+            }
+        }
+
+        tsk = tipset.parents().clone();
+        prev_epoch = tipset.epoch();
+        pb.set((ts.epoch() - tipset.epoch()) as u64);
+    }
+
+    pb.finish_println("snapshot is valid");
```
lemmih, line 472: Not sure why but this doesn't print a newline for me.

creativcoder:
> Getting closer. Still need to hook it up in the CI to check that the calibnet snapshot is valid (and also check that validation fails if we try to validate it as a mainnet snapshot).

Confirmed validation failure with calibnet snapshot but mainnet --chain flag: 
```
Validating tipsets: 103126 / 108720 
Error: could not reach genesis from snapshot, genesis tipset in snapshot doesn't match the reference
```

forest/cli/src/cli/snapshot_cmd.rs
lemmih: Let's work around it.

creativcoder: done

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: done

.github/workflows/rust.yml
```diff
@@ -165,6 +165,9 @@ jobs:
         run: |
           forest-cli snapshot export
           sha256sum -c *.sha256sum
+      - name: validate snapshot
+        run: |
+          forest-cli --chain calibnet snapshot validate $SNAPSHOT_DIRECTORY/*.car --recent-stateroots 2000 --force
```
lemmih, line 170: Maybe without the `--recent-stateroots` argument since we also omit it when exporting a snapshot. Also, let's check that validating the snapshot as a mainnet snapshot fails.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +350,129 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    } else {
+        true
+    };
+
+    if confirm {
+        let tmp_db_path = TempDir::new()?;
+
+        let db_path = tmp_db_path.path().join(&config.chain.name);
+
+        #[cfg(feature = "rocksdb")]
```
lemmih, line 378: This does not look right here. The cli shouldn't have features for rocksdb and paritydb.

lemmih (changes requested):

Makefile
```diff
@@ -71,7 +71,7 @@ lint: license clean
 	taplo fmt --check
 	taplo lint
 	cargo clippy --all-targets -- -D warnings
-	cargo clippy --all-targets --no-default-features --features forest_deleg_cns,paritydb,instrumented_kernel -- -D warnings
+	cargo clippy --all-targets --no-default-features --features forest_deleg_cns,rocksdb,instrumented_kernel -- -D warnings
```
lemmih, line 74: I think `rocksdb` is a default feature so it'll be checked in the line above this one. Why do you need to add it here?

lemmih: Oh, I think I see. You don't want to compile the cli without the `rocksdb` feature. We should just remove the `rocksdb` and `paritydb` features from the cli. They don't belong there.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +353,127 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+        let tipset = chain_store.tipset_from_keys(&tsk).await?;
+        if tipset.epoch() >= prev_epoch {
+            bail!(
+                "tipset iteration didn't make progress backwards current epoch: {} previous_epoch: {}, possible recursion in snapshot.", tipset.epoch(), prev_epoch
+            );
+        }
+        let height = tipset.epoch();
+        if height == 0 {
+            if tipset.as_ref() != genesis_tipset {
+                bail!("could not reach genesis from snapshot, genesis tipset in snapshot doesn't match the reference");
```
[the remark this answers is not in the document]
creativcoder: I've added the chain_name to the error log. I feel checking for known genesis hashes that would make this function more involved than it needs to be and a bit slow , isn't pointing to the alternate chain_name a good enough context?

Makefile
creativcoder: It didn't and that's why the ci lint step was always red.

forest/cli/src/cli/snapshot_cmd.rs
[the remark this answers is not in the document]
lemmih: There are more than two chains. We know about `calibnet` and `mainnet` but there are many more and people may even run their own local chain.

Checking if the genesis hash matches `calibnet` or `mainnet` shouldn't be slow but it's fine to leave it out. This error message isn't too important.

Makefile
lemmih: Do you see what needs to be changed in the cli's Cargo file?

.github/workflows/rust.yml
```diff
@@ -165,6 +165,11 @@ jobs:
         run: |
           forest-cli snapshot export
           sha256sum -c *.sha256sum
+      - name: validate snapshot
+        run: |
+          mainnet_with_calibnet_snapshot_should_fail=$(forest-cli --chain mainnet snapshot validate $SNAPSHOT_DIRECTORY/*.car --force)
+          if [ "$?" -eq 0 ]; then echo "calibnet snapshot validation on mainnet should fail" exit 1; fi
```
lemmih, line 171: Github runs bash with the `-e` flag which causes it to exit immediately on errors (unless the commands are part of an if-statement). You probably want something like:

```suggestion
          if forest-cli --chain mainnet snapshot validate $SNAPSHOT_DIRECTORY/*.car --force; then
              echo "calibnet snapshot validation on mainnet should fail";
              exit 1;
          fi
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -335,3 +491,25 @@ fn delete_snapshot(snapshot_path: &PathBuf) {
...
+    {
+        forest_db::rocks::RocksDb::open(path, &config.rocks_db)
+            .map_err(|e| anyhow::anyhow!("failed to open db: {}", e))
+    }
+    #[cfg(feature = "paritydb")]
+    {
+        let paritydb_config = forest_db::parity_db::ParityDbConfig::from_path(path);
+        forest_db::parity_db::ParityDb::open(&paritydb_config)
+    }
+}
```
LesnyRumcajs, line 515: This got a bit complex; too many compilation conditionals lurking around, even as a param - supporting another config would be a pain. How about we create two `open_db` methods, i.e.

```
#[cfg(feature = "rocksdb")]
fn open_db(someparams) { ... }
#[cfg(feature = "paritydb")]
fn open_db(otherparams) { ... }
```

LesnyRumcajs, line 515: This got a bit complex; too many compilation conditionals lurking around, even as a param - supporting another config would be a pain. How about we create two `open_db` methods, i.e.

```
#[cfg(feature = "rocksdb")]
fn open_db(someparams) { ... }
#[cfg(feature = "paritydb")]
fn open_db(otherparams) { ... }
```

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +356,129 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+
+    let total_size = ts.epoch();
+    let pb = forest_utils::io::ProgressBar::new(total_size as u64);
+    pb.message("Validating tipsets: ");
+    pb.set_max_refresh_rate(Some(std::time::Duration::from_millis(500)));
+
+    // Security: Recursive snapshots are difficult to create but not impossible. This
+    // limits the amount of recursion we do.
+    let mut prev_epoch = ts.epoch();
+    loop {
```
LesnyRumcajs, line 442: For readability, it may be worth putting the content of this loop in a separate method. Then, potentially, we could do some basic unit testing.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -215,6 +237,17 @@ impl SnapshotCommands {
...
+            Self::Validate {
+                recent_stateroots,
+                snapshot,
+                force,
+            } => {
+                if cfg!(feature = "rocksdb") {
+                    validate(&config, recent_stateroots, snapshot, *force).await
+                } else {
+                    bail!("Snapshot validation needs a database. Please compile forest with supported database features: paritydb");
+                }
```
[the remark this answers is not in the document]
lemmih: Oh, right, I think this was part of an earlier solution where `rocksdb` was required. The `if` should just be deleted.

forest/cli/src/cli/snapshot_cmd.rs
lemmih: :+1: 

I also made an issue for refactoring the dupliacte `open_db` code such that we don't have multiple copies.

forest/cli/src/cli/snapshot_cmd.rs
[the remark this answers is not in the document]
creativcoder: Was an older code, removed it

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +356,129 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+    recent_stateroots: &i64,
+    snapshot: &PathBuf,
+    force: bool,
+) -> anyhow::Result<()> {
+    let confirm = force
+        || atty::is(atty::Stream::Stdin)
+            && Confirm::with_theme(&ColorfulTheme::default())
+                .with_prompt(format!(
+                    "This will result in using approximately {} MB of data. Proceed?",
+                    std::fs::metadata(snapshot)?.len() / (1024 * 1024)
```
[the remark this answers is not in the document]
creativcoder: The [docs](https://doc.rust-lang.org/std/fs/fn.metadata.html) says it does follow symlinks

.github/workflows/rust.yml
```diff
@@ -165,6 +165,14 @@ jobs:
...
           forest-cli snapshot export
           sha256sum -c *.sha256sum
+      - name: validate snapshot
+        run: |
+          set +e
+          mainnet_with_calibnet_snapshot_should_fail=$(forest-cli --chain mainnet snapshot validate $SNAPSHOT_DIRECTORY/*.car --force)
+          if [ "$?" -eq 0 ]; then echo "calibnet snapshot validation on mainnet should fail" exit 1; else 
+            forest-cli --chain calibnet snapshot validate $SNAPSHOT_DIRECTORY/*.car --force
+          fi
+          set -e
```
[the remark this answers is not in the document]
creativcoder: what's `validate-mainnet` ? a command in forest cli, a shell functon, or a github job name?

forest/cli/src/cli/snapshot_cmd.rs
creativcoder: I'm currently looking into if we can have a common `Config` instance, if that is possible, we won't need cfg feature in params of open_db.

.github/workflows/rust.yml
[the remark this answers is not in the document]
LesnyRumcajs: @creativcoder it's me being lazy :) it was my shortcut for `forest-cli --chain mainnet snapshot validate $SNAPSHOT_DIRECTORY/*.car --force`

node/db/src/parity_db.rs
```diff
@@ -33,6 +34,15 @@ impl ParityDbConfig {
     }
 }
 
+impl Default for ParityDbConfig {
+    fn default() -> Self {
+        Self {
+            path: PathBuf::from("/tmp/paritydb"),
```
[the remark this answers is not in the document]
LesnyRumcajs: We could make the path an `Option`. Here it would be `None`. In such a case, I would expect the same behaviour as for https://github.com/ChainSafe/forest/blob/f0ac5b50f91eb23a35e1c2e533cb5e65ffe7b542/forest/shared/src/cli/mod.rs#L294-L299

.github/workflows/rust.yml
```diff
@@ -165,6 +165,13 @@ jobs:
         run: |
           forest-cli snapshot export
           sha256sum -c *.sha256sum
+      - name: validate snapshot
+        run: |
+          set +e
```
LesnyRumcajs, line 170: You don't need those anymore if I get them right

lemmih:
@creativcoder How about removing the `path` field from `ParityDbConfig` altogether? `ParityDb::open` could then take two arguments which would be more consistent with how `RocksDB` works.

.github/workflows/rust.yml
creativcoder: Yep, removed

.github/workflows/rust.yml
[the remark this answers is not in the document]
creativcoder: thanks, this is more clean.

creativcoder:
> @creativcoder How about removing the `path` field from `ParityDbConfig` altogether? `ParityDb::open` could then take two arguments which would be more consistent with how `RocksDB` works.

Done

forest/daemon/src/daemon.rs
```diff
@@ -143,7 +144,8 @@ pub(super) async fn start(config: Config, detached: bool) -> Db {
 
     let keystore = Arc::new(RwLock::new(ks));
 
-    let db = open_db(&config);
+    let db =
+        forest_cli_shared::open_db(&db_path(&config), &config).expect("failed opening database");
```
lemmih, line 148:
```suggestion
        forest_cli_shared::open_db(&db_path(&config), &config)?;
```

node/db/src/parity_db.rs
```diff
@@ -19,24 +19,21 @@ pub struct ParityDb {
...
-            columns: 1,
-        }
+impl Default for ParityDbConfig {
+    fn default() -> Self {
+        Self { columns: 1 }
     }
 }
 
 impl ParityDb {
     fn to_options(config: &ParityDbConfig) -> Options {
```
lemmih, line 34: How about adding the path parameter here?

utils/statediff/src/main.rs
```diff
@@ -46,11 +46,8 @@ fn open_db(chain_path: &Path) -> forest_db::rocks::RocksDb {
 #[cfg(feature = "paritydb")]
 fn open_db(chain_path: &Path) -> forest_db::parity_db::ParityDb {
```
lemmih, line 47: Can't this code use the shared `open_db` function?

LesnyRumcajs (approved):
LGTM. Having some unit tests for the invariants would be fantastic, but it's not a huge deal if we create a separate issue.

forest/shared/src/lib.rs
```diff
@@ -3,3 +3,42 @@
...
+/// Gets database directory
+pub fn db_path(config: &crate::cli::Config) -> PathBuf {
+    chain_path(config).join("rocksdb")
+}
+
+#[cfg(feature = "paritydb")]
+/// Gets database directory
+pub fn db_path(config: &crate::cli::Config) -> PathBuf {
+    chain_path(config).join("paritydb")
+}
```
lemmih, line 30: Maybe refactor this to a single function.

forest/shared/src/lib.rs
```diff
@@ -3,3 +3,42 @@
...
+
+pub fn open_db(path: &std::path::Path, config: &cli::Config) -> Result<Db, forest_db::Error> {
```
lemmih, line 32: It might be worth it to have `open_db(config)` and `open_db_at(path, config)`. Also, shouldn't this code be in `forest_db`?

lemmih (changes requested):

lemmih:
The configuration should know about rocksdb and paritydb even if those features aren't enabled. But is unrelated to this PR and we can deal with it later.

forest/shared/src/lib.rs
creativcoder: can't have it in forest_db as there's cyclic dependency.

forest/daemon/src/daemon.rs
lemmih: We need a review for https://github.com/ChainSafe/forest/pull/2363, then. :)

lemmih: Let's merge #2363 before we merge this PR.

utils/statediff/src/main.rs
creativcoder: Can't because of cyclic dependency. I'm wondering if there's a reason statediff is separate, rather than within cli crate as a subcommand?

forest/shared/src/lib.rs
creativcoder: done

node/db/src/parity_db.rs
creativcoder: done

LesnyRumcajs:
There are quite a few conflicts; I'll re-review once they are resolved, but the change looks good.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -323,6 +350,127 @@ fn clean(config: &Config, snapshot_dir: &Option<PathBuf>, force: bool) -> anyhow
...
+                ))
+                .default(false)
+                .interact()
+                .unwrap_or_default();
+
+    if confirm {
+        let tmp_db_path = TempDir::new()?;
```
elmattic, line 371: Maybe dropping and closing `tmpdir` explicitly would be a bit more robust here.

elmattic (approved):

hanabi1224 (approved):

lemmih (approved):

PR #2368, opened 2022-12-15
Remove `check_for_low_fd`

tyshko5:
**Summary of changes**
- 

Closes 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2856,7 +2856,6 @@ dependencies = [
  "quickcheck",
  "rand 0.8.5",
  "raw_sync",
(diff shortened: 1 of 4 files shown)


elmattic:
I believe the check is still needed if you want to warn the user about low fd. But I'm ok that this is something that must be take care outside of Forest by node operator.

Maybe have a look on `#### File Descriptor Limits` section to update it, ie `usize` should be `ulimit`.

elmattic (approved):

lemmih (approved):
@elmattic makes a good point. Could you delete the

Issue #2244, opened 2022-11-22
Forest is stuck while syncing


elmattic:
**Describe the bug**
Trying to sync from an old snapshot seems to be impossible.

Using the same snapshot with Lotus seems to work (I haven't managed to sync up to HEAD, but only because I didn't have enough storage left and stopped at 2,259,643).

**To Reproduce**
1. Download a snapshot from last month, ie `minimal_finality_stateroots_2244720_2022-10-13_08-00-00.car`
2. Import the snapshot and halt after import
3. Try to sync up to HEAD

**Log output**
```
 2022-11-21T10:30:43.335Z ERROR forest_chain_sync::tipset_syncer > Sync messages check state failed for tipset range
 2022-11-21T10:30:43.335Z ERROR forest_chain_sync::chain_muxer   > Bootstrapping failed, re-evaluating the network head to retry the bootstrap. Error = TipsetRangeSyncer(NetworkMessageQueryFailed("ChainExchange request failed for all top peers"))
 2022-11-21T10:30:43.335Z INFO  forest_chain_sync::chain_muxer   > Evaluating network head...
```

See log file for more.

**Expected behaviour**

Forest can sync from an old snapshot up to HEAD.

**Environment (please complete the following information):**
 - OS: macOS 12.5
 - Rust version: rustc 1.66.0-nightly
 - Branch/commit: main

**Other information and links**

[forest.log](https://github.com/ChainSafe/forest/files/10067979/forest.log)

LesnyRumcajs:
Happened to me as well but the snapshot isn't that old. `minimal_finality_stateroots_2340241_2022-11-15_12-00-30.`

```
 2022-11-22T19:08:26.507Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 2342243
 2022-11-22T19:08:30.512Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 2342244
 2022-11-22T19:08:40.035Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 2342245
 2022-11-22T19:08:43.502Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 2342246
 2022-11-22T19:08:47.475Z INFO  forest_chain_sync::tipset_syncer > Validating tipset: EPOCH = 2342247
[...]

Issue #2366, opened 2022-12-14
Improve dagcbor decoding flow in libp2p `Hello` and `ChainExchange` protocols


hanabi1224:
**Issue summary**

Currently, we're not decoding from a stream, and due to `rust-libp2p` implementation, the stream will be reused to send more messages, so we're not able to read the entire stream before sending it to the decoder.   Current flow is
1. create a buffer for decoding
2. try reading 8*1024 bytes from the stream, append to the buffer in step 1
3. try decode buffer in step 1, if success with no trailing data left, it's considered successful and terminate the loop. otherwise go back to step 2

With this flow, there're a few issues.
1. Vulnerable to attacks.  If some malicious nodes send random bytes, currently logic is not capable of detecting it early, it will try reading more data from the input stream and cause deadlock.  While `Lotus` can detect this with the head bytes.
2. Many unnecessary decoding operations as described above for detecting the end of the message, while `Lotus` can do stream decoding

**Other information and links**

LesnyRumcajs:
@hanabi1224 what do we do with this issue?

PR #2369, opened 2022-12-15
fix: improve dagcbor decoding flow

hanabi1224:
**Summary of changes**
- Remove hacks from response decoding in `Hello` and `ChainExchange` protocol.
- Due to how lotus implement this protocol, hack is still required for decoding requests
- Remove `mplex` protocol as lotus does not use it
- Remove `yamux` custom config, the default one should work just fine 
- This should close https://github.com/ChainSafe/forest/issues/2244 as it removes hacks from response decoding flow

See [code comment](https://github.com/ChainSafe/forest/pull/2369/files#diff-c3c7f608d834b9b4c5a06eea728176418da50e38bf127628b6749ca1025a88beR128) for detailed explanations 

This is not a problem in lotus because they perform stream decoding, which is unfortunately not supported in rust ecosystem atm

Closes https://github.com/ChainSafe/forest/issues/2244 Mitigates https://github.com/ChainSafe/forest/issues/2366

https://docs.rs/yamux/0.11.0/yamux/struct.Stream.html
https://github.com/libp2p/go-yamux/blob/master/spec.md


LesnyRumcajs (approved):
Fantastic findings!

node/forest_libp2p/src/rpc/mod.rs
```diff
@@ -122,14 +125,32 @@ where
...
+//
+async fn read_request_and_decode<IO, T>(io: &mut IO) -> io::Result<T>
 where
     IO: AsyncRead + Unpin,
     T: serde::de::DeserializeOwned,
 {
-    // FIXME: investigate the best value here, 2MB is not enough for mainnet epoch 2419928
-    // Issue: https://github.com/ChainSafe/forest/issues/2362
-    const MAX_BYTES_ALLOWED: usize = 200 * 1024 * 1024; // messages over 200MB are likely malicious
+    const MAX_BYTES_ALLOWED: usize = 2 * 1024 * 1024; // messages over 2MB are likely malicious
```
lemmih, line 153: So the message size is no longer a problem with the new way of reading responses?

hanabi1224: This function is no longer used for decoding responses, and requests should be very small

lemmih (approved):

tyshko5 (approved):

Issue #2338, opened 2022-12-07
Add log file options support to `forest-daemon`


hanabi1224:
**Issue summary**

After switching from `log` to `tracing` (https://github.com/ChainSafe/forest/pull/2326), this should be feasible.

Current blocker: https://github.com/ChainSafe/forest/blob/main/forest/shared/src/cli/mod.rs#L106

```rust
    // env_logger-0.7 can only redirect to stderr or stdout. Version 0.9 can redirect to a file.
    // However, we cannot upgrade to version 0.9 because pretty_env_logger depends on version 0.7
    // and hasn't been updated in quite a while. See https://github.com/seanmonstar/pretty-env-logger/issues/52
    // #[structopt(
    //     help = "Specify a filename into which logging should be appended"
    // )]
    // pub log_file: Option<PathBuf>,
```

**Other information and links**

PR #2371, opened 2022-12-15
feat: add daemon option `--log-dir`

hanabi1224:
**Summary of changes**
- add daemon option `--log-dir`
- update CI sync check job to print log files
- fix file logger for detached daemon by using blocking mode

CI print file log section: https://github.com/ChainSafe/forest/actions/runs/3712366880/jobs/6293818268#step:13:1

Closes https://github.com/ChainSafe/forest/issues/2338


elmattic:
Nice! When detaching the node you will have as well some redirection to files.
See https://github.com/ChainSafe/forest/blob/main/forest/daemon/src/main.rs#L65

hanabi1224:
@elmattic Ah just notice it stops writing the log file after demonizing, will investigate why

>Run ls -hl log
total 4.0K
-rw-r--r-- 1 runner docker 287 Dec 15 17:43 forest.log.2022-[12](https://github.com/ChainSafe/forest/actions/runs/3706496397/jobs/6281777554#step:13:13)-15-17
2022-12-15T17:43:[14](https://github.com/ChainSafe/forest/actions/runs/3706496397/jobs/6281777554#step:13:15).847081Z  INFO forest: Using default calibnet config    
2022-12-15T17:43:14.847367Z  INFO forest: Redirecting stdout and stderr to files forest.out and forest.err.    
2022-12-15T17:43:14.883342Z  INFO forest: Forest has been detached and runs in the background.

forest/daemon/src/main.rs
```diff
@@ -107,7 +107,7 @@ fn main() {
     // Run forest as a daemon if no other subcommands are used. Otherwise, run the subcommand.
     match opts.to_config() {
         Ok((cfg, path)) => {
-            let (loki_task, ..) = logger::setup_logger(&cfg.log, &opts);
+            let (loki_task, _file_logger_worker_guard) = logger::setup_logger(&cfg.log, &opts);
```
lemmih, line 110: `_file_logger_worker_guard` will be dropped when we detach. Is that okay?

lemmih: What are the benefits and drawbacks of the blocking and non-blocking loggers?

hanabi1224: The non blocking one queues log entries in memory and uses a dedicated thread to write files, while a blocking one is just like the console logger, writes to the file (buffer) directly

lemmih: How we always use the blocking logger, then? I'm concerned that having different logic for the detach and non-detach modes will lead to unexpected behavior.

hanabi1224: Sounds good, I don't really think it will introduce too big overhead

hanabi1224: Fix pushed

lemmih (approved):

.github/workflows/rust.yml
```diff
@@ -172,6 +172,10 @@ jobs:
           echo "--- Forest STDOUT ---"; cat forest.out
           echo "--- Forest STDERR ---"; cat forest.err
           echo "--- Forest Prometheus metrics ---"; cat metrics.log
+      - name: print forest log files
+        run: |
+          ls -hl log
+          cat log/*
```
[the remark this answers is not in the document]
LesnyRumcajs: Plus, it has the same name as the previous step. Seems redundant.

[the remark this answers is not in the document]
hanabi1224: Yes it's the same, mostly verifying it works in detach mode

hanabi1224:
> Just wondering if we still really need this.

@elmattic I guess more refactoring with IaC scripts are required to retire the redirected stdio files, we could plan it in a separate issue

PR #2374, opened 2022-12-16
chore: improve code coverage

lemmih:
**Summary of changes**
- Remove a bit of dead code.
- Hoist coverage commands to a stand-alone script.
- Increase link-checker timeout from 20s to 60s.
- Test coverage increased to ~70%.

Closes 


hanabi1224 (approved):

node/rpc/src/wallet_api.rs
```diff
@@ -234,34 +230,6 @@ where
     Ok(SignatureJson(sig))
 }
 
-/// Sign an `UnsignedMessage`, return `SignedMessage`
```
LesnyRumcajs, line 237: I am not sure about removing this one. Seems like it can be easily covered and is quite useful if we move to a standalone wallet binary.

scripts/gen_coverage_report.sh
```diff
@@ -0,0 +1,68 @@
...
+# llvm-cov can be installed by running: cargo install cargo-llvm-cov
+#
+
+set +e
+
+cargo llvm-cov --workspace clean
+cargo llvm-cov --workspace --no-report --features slow_tests
+cargo llvm-cov --no-report run --bin=forest-cli -- --chain calibnet db clean --force
+cargo llvm-cov --no-report run --bin=forest-cli -- --chain calibnet snapshot fetch -s .
+# SC2012 says to use 'find' instead of 'ls' but I don't know how to do that.
```
LesnyRumcajs, line 20: `find . -name "*.car"` doesn't work? I don't think we should expect multiple CAR files.

blockchain/state_manager/src/chain_rand.rs
```diff
@@ -92,18 +92,6 @@ where
         )
     }
 
-    /// network version 0-12
```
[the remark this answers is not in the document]
lemmih: I'd rather keep it purely in the git history than as commented-out code. It's quite unlikely we'll support network versions this old.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -189,37 +170,18 @@ impl WalletCommands {
...
             }
             Self::SetDefault { key } => {
-                let key_parse_result = Address::from_str(key);
-
-                if key_parse_result.is_err() {
-                    cli_error_and_die("Error parsing address. Verify that the address exists and is in the keystore", 1);
-                }
-
-                let key = key_parse_result?;
+                let key = Address::from_str(key)?;
```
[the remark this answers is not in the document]
lemmih: It prints out the reason why the address failed to parse which is an improvement. I'll add a context to make it even better.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -189,37 +170,18 @@ impl WalletCommands {
...
-                    let balance_int = match balance_string.parse::<BigInt>() {
-                        Ok(balance) => TokenAmount::from_atto(balance),
-                        Err(err) => {
-                            println!(
-                                "Couldn't convert balance {balance_string} to TokenAmount: {err}"
-                            );
-                            continue;
-                        }
-                    };
+                    let balance_int = TokenAmount::from_atto(balance_string.parse::<BigInt>()?);
```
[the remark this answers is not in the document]
lemmih: It's an internal RPC bug if the balance cannot be parsed. Ideally we would change the RPC type to `TokenAmount` rather than `String` but that's difficult for silly reasons.

scripts/gen_coverage_report.sh
lemmih: There might be more car files when running the script locally. Is there an easy way to sort by modification time with `find`?

node/rpc/src/wallet_api.rs
lemmih: It can be implemented using the `sign` RPC, can't it?

scripts/gen_coverage_report.sh
LesnyRumcajs: None I am aware of, only the `print` voodoo, which doesn't help with readability. Alternatively, we could save the snapshot to a temporary directory (personally, that's what I would prefer) with `mktemp`.

node/rpc/src/wallet_api.rs
LesnyRumcajs: I think so, yeah.

lemmih: Okay, let's just keep the `sign` RPC call, then. If `sign_message` becomes convenient in the future, we can restore this code.

LesnyRumcajs (approved):
rock-solid

PR #2376, opened 2022-12-16
Wallet testing in Ci and calibnet ci checks to a script

### LINKED ISSUE #2318: Wallet handling testing during CI
**Issue summary**

At present, we have a workflow job that downloads a calibnet snapshot, sync to HEAD, and then exports a new snapshot. So far, so good. However, we want to extend these checks over time and they'll quickly grow too complex for the work flow and should therefore be moved to a separate script.

Additionally, we want the script to include basic wallet handling. This means importing a wallet that has been pre-loaded with FIL (from the faucet), checking its balance, creating a new address, moving a tiny amount of FIL from the pre-loaded address to the new address, exporting the wallet, and importing it again.

**Task summary**

- [x] Move CI check to separate script.
- [x] Create wallet, load it with FIL from the faucet, and export it. Figure out how to make it available in the CI server. It doesn't need to be kept secret.
- [x] CI steps:
  - [x] Import the pre-loaded wallet.
  - [x] Check that the wallet has sufficient funds.
  - [x] Create a new address.
  - [x] Move a tiny amount of FIL to the new address.
  - [x] Verify that the transfer was recorded in the blockchain.
  - [x] Export the wallet and import it again.

**Acceptance Criteria**

- [x] Calibnet health checks have been moved to separate script.
- [x] CI still passes.
- [x] Basic wallet handling is tested.

Calibnet sync check:
https://github.com/ChainSafe/forest/blob/37340139cb8831c572c1e25357ab452c589920d1/.github/workflows/rust.yml#L125-L177

creativcoder:
**Summary of changes**
- [X] Calibnet health checks in a separate script

Closes #2318 


scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,30 @@
...
+#Import snapshot and run Forest
+forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --halt-after-import --height=-200 --import-snapshot $SNAPSHOT_DIRECTORY/*.car
+forest-cli --chain calibnet db stats
+forest --chain calibnet --target-peer-count 50 --encrypt-keystore false --detach
+
+# Validate checkpoint tipset hashes
+forest-cli chain validate-tipset-checkpoints
+
+# wait for sync and check health
+sleep 30m
```
lemmih, line 18: You probably want to use the `timeout` command rather than `sleep`.

addr_two_pkey.key
```diff
@@ -0,0 +1 @@
+7b2254797065223a22736563703235366b31222c22507269766174654b6579223a223546545633762f322f62527030753231544b466b6f386d2f566d5632434e6359334b444b303848623456633d227d
```
hanabi1224, line 1: Maybe add `test` to the file name to make it clear it's not sth important being commited by mistake

creativcoder: added

scripts/preloaded_wallet.key
```diff
@@ -0,0 +1 @@
+7b2254797065223a22736563703235366b31222c22507269766174654b6579223a223546545633762f322f62527030753231544b466b6f386d2f566d5632434e6359334b444b303848623456633d227d
```
hanabi1224, line 1: Maybe add `test` to the this key file name as well

addr_two_pkey.key
hanabi1224, line 1: Possible to move it to the scripts folder?

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -204,21 +204,33 @@ impl WalletCommands {
                 exact_balance,
                 fixed_unit,
             } => {
+                // if config.chain.name == "calibnet" {
+                //     forest_shim::address::set_current_network(forest_shim::address::Network::Testnet);
+                // }
```
lemmih, line 209: We can delete this, right?

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -204,21 +204,33 @@ impl WalletCommands {
...
+                    .map_err(handle_rpc_err);
 
                 let (title_address, title_default_mark, title_balance) =
                     ("Address", "Default", "Balance");
                 println!("{title_address:41} {title_default_mark:7} {title_balance}");
 
                 for address in response {
                     let addr = address.0.to_string();
-                    let default_address_mark = if addr == default { "X" } else { "" };
+                    let default_address_mark = if let Ok(default) = &default {
```
lemmih, line 225: If we don't always have a default address, maybe `wallet_default_address` should return an optional value.

Otherwise, we can write something like `Ok(addr) == default` to simplify the code.

key_management/src/wallet.rs
```diff
@@ -174,10 +174,11 @@ pub fn list_addrs(keystore: &KeyStore) -> Result<Vec<Address>, Error> {
     let mut out = Vec::new();
     for i in all {
         if i.starts_with("wallet-") {
```
lemmih, line 176: This `if` can be removed.

lemmih (changes requested):

documentation/src/cli.md
```diff
@@ -33,10 +33,24 @@ forest-cli --token eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwi
 ```
 
 When the Forest daemon is started, an admin token will be displayed (alternatively, use `--save-token <token>` flag to save it on disk).
+## Sending filecoin tokens (money) from your wallet.
+
+For sending filecoin token, forest daemon must be running. You can do so by running:
```
LesnyRumcajs, line 38:
```suggestion
For sending Filecoin tokens, the Forest daemon must be running. You can do so by running:
```

documentation/src/cli.md
```diff
@@ -33,10 +33,24 @@ forest-cli --token eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwi
 ```
 
 When the Forest daemon is started, an admin token will be displayed (alternatively, use `--save-token <token>` flag to save it on disk).
+## Sending filecoin tokens (money) from your wallet.
```
LesnyRumcajs, line 36: I wouldn't associate tokens with money.
```suggestion
## Sending Filecoin tokens from your wallet
```

documentation/src/cli.md
```diff
@@ -33,10 +33,24 @@ forest-cli --token eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwi
...
+
+For sending filecoin token, forest daemon must be running. You can do so by running:
+
+`forest --chain calibnet --target-peer-count 50 --encrypt-keystore false`
```
LesnyRumcajs, line 40: Do we need this peer count and disable the keystone? I mean, it's OK for the CI and testing, but in general, we should recommend and default to have it encrypted.

documentation/src/cli.md
```diff
@@ -33,10 +33,24 @@ forest-cli --token eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwi
...
+
+Next, to send filecoin token (money) to a wallet address:
```
LesnyRumcajs, line 42:
```suggestion
Next, send Filecoin tokens to a wallet address:
```

documentation/src/cli.md
```diff
@@ -33,10 +33,24 @@ forest-cli --token eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwi
...
+
+`forest-cli --token <admin_token> send <wallet-address> <amount in attoFIL>`
+
+where 1 attoFIL = 10e−18 FIL.
```
LesnyRumcajs, line 46: It's not a pretty representation and I believe it's not accurate - 10e-18 is 10 attoFIL, no? Let's utilize some markdown capabilities:
```suggestion
where 1 attoFIL = $10^{−18}$ FIL.
```

forest/cli/src/cli/wallet_cmd.rs
elmattic: I think so.

documentation/src/cli.md
```diff
@@ -33,10 +33,24 @@ forest-cli --token eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwi
...
 
 ## Wallet
 
-All wallet commands require write permissions to interact with the keystore
+Filecoin wallets are stored under forest data directory (e.g., ~/.local/share/forest in case of linux) in a `keystore` file.
```
LesnyRumcajs, line 50:
```suggestion
Filecoin wallets are stored under the Forest data directory (e.g., `~/.local/share/forest` in the case of Linux) in a `keystore` file.
```

forest/cli/src/cli/wallet_cmd.rs
elmattic: `wallet_default_address` is part of the JSON-RPC API that's why I think the result is a string (and string is empty when no default address is found).

documentation/src/cli.md
```diff
@@ -33,10 +33,24 @@ forest-cli --token eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwi
...
+
+All wallet commands require write permissions and admin token (`--token`) to interact with the keystore.
+The admin token can be retrieved from forest startup logs.
```
LesnyRumcajs, line 53:
```suggestion
All wallet commands require write permissions and an admin token (`--token`) to interact with the keystore.
The admin token can be retrieved from forest startup logs or by running it with `--save-token <PATH>`.
```

forest/cli/src/cli/wallet_cmd.rs
elmattic: Reading this code I'm a bit puzzled by usage of a Rust keyword as name of a let binding.

LesnyRumcajs (changes requested):

documentation/src/cli.md
elmattic: I think so too.

scripts/preloaded_wallet.key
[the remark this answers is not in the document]
elmattic: Good point. How to obfuscate this?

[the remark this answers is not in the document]
lemmih: We can put it in our secret environment variables.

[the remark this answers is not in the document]
LesnyRumcajs: We should store it in repo secrets.
https://docs.github.com/en/actions/security-guides/encrypted-secrets#example-using-bash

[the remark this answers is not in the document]
lemmih: I'm tempted to leave it out in the open, though. That makes it easier to run the test script locally. And who would want to steal calibnet funds? Maybe I have too much faith in humanity. :)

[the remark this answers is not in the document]
lemmih: I've added the data as a secret with the name `CALIBNET_WALLET`.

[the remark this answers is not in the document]
LesnyRumcajs: It should be trivial to run it locally - just a one-time-off hassle to save the key. Just invoke the secret, pass it to the script as an argument, and it's done. 

```
steps:
  - shell: cmd
    env:
      WALLET_KEY: ${{ secrets.CALIBNET_WALLET_KEY }}
    run: |
      ./scripts/calibnet_health_check.sh "%WALLET_KEY%"
```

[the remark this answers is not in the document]
LesnyRumcajs: Yes, I have no faith in humanity. :crying_cat_face:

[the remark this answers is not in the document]
lemmih: So to run the calibnet health script locally, you'd have to create and fund a test wallet first.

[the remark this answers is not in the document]
LesnyRumcajs: Yes. You'd do it only once, though. Unless your SSD fails and you have no backup. :clown_face:

forest/cli/src/cli/wallet_cmd.rs
lemmih: `default` is not a keyword, though.

elmattic: That's right it's not in the strict keyword list:
https://doc.rust-lang.org/reference/keywords.html

elmattic: Hmm, I tried `Ok(addr) == default` but didn't managed because `PartialEq` is not implemented for `anyhow::Error`.

LesnyRumcajs: @elmattic I believe something like this should work.
```
                    let default_address_mark =
                        if default.as_ref().map(|v| v == &addr).unwrap_or(false) {
                            "X"
                        } else {
                            ""
                        };
```

scripts/preloaded_wallet.key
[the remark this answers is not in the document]
elmattic: Should be better now. @lemmih @LesnyRumcajs

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,121 @@
...
+
+echo "Listing wallet balances"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" wallet list
+
+echo "Sending FIL to the above address"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" send "$ADDR_TWO" "$FIL_AMT"
+
+echo "Checking balance of $ADDR_TWO..."
+
+sleep 5m
```
lemmih, line 102: Is there an easy way to check every 30 second with a timeout?

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,121 @@
...
+echo "Creating a new address to send FIL to"
+ADDR_TWO=$($FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" wallet new)
+echo "$ADDR_TWO"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" wallet set-default "$ADDR_ONE"
+
+echo "Listing wallet balances"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" wallet list
+
+echo "Sending FIL to the above address"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" send "$ADDR_TWO" "$FIL_AMT"
```
lemmih, line 98: It's surprising to me that `send` expects attoFIL. It's also different from what Lotus does. Could you create an issue for aligning our `send` command with the one from Lotus?

scripts/calibnet_health_check.sh
lemmih: Sounds good.

scripts/calibnet_health_check.sh
elmattic: Yes.

lemmih:
@elmattic Oh, could you add an issue for updating the coverage report with FIL sending as well.

lemmih (approved):

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,128 @@
...
+
+# The following steps does basic wallet handling tests.
+
+# Amount to send to
+FIL_AMT=500
+# Admin token used when interacting with wallet
+ADMIN_TOKEN=$(cat admin_token)
+# Wallet addresses: 
+# A preloaded address
+ADDR_ONE=t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y
```
LesnyRumcajs, line 66: Can we get this address from the CLI? I assume that after we import the key, we should see the address in the available list?

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,128 @@
...
+
+echo "Importing preloaded wallet key"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" wallet import preloaded_wallet.key
+sleep 5s
+
+echo "Exporting key"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" wallet export "$ADDR_ONE" > preloaded_wallet.test.key
+if ! cmp -s preloaded_wallet.key preloaded_wallet.test.key; then
+    echo ".key files should match"
+    $FOREST_CLI_PATH --token "$ADMIN_TOKEN" shutdown --force
```
LesnyRumcajs, line 76: Let's verify that `forest` process is no longer running.

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,128 @@
...
+fi
+
+echo "Fetching metrics"
+wget -O metrics.log http://localhost:6116/metrics
+
+sleep 5s
+
+# Show balances
+echo "Listing wallet balances"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" wallet list
```
LesnyRumcajs, line 87: I wonder if we could have the `FULLNODE_API_INFO="<token goes here>:/ip4/<host>/tcp/<port>/http` env variable to not append the `--token $ADMIN_TOKEN` everywhere?

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,128 @@
...
+
+echo "Sending FIL to the above address"
+$FOREST_CLI_PATH --chain calibnet --token "$ADMIN_TOKEN" send "$ADDR_TWO" "$FIL_AMT"
+
+echo "Checking balance of $ADDR_TWO..."
+
+ADDR_TWO_BALANCE=0
+i=0
+while [[ $i != 10 && $ADDR_TWO_BALANCE == 0 ]]; do
+  echo "i=$i"
```
LesnyRumcajs, line 105: Let's make it more human-readable, e.g., `"Check $i", perhaps with a timestamp.

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,128 @@
...
+if [ "$ADDR_TWO_BALANCE" != "$FIL_AMT" ]; then
+  echo "FIL amount should match"
+  $FOREST_CLI_PATH --token "$ADMIN_TOKEN" shutdown --force
+  exit 1
+fi
+
+echo "Get and print metrics and logs and stop forest"
+wget -O metrics.log http://localhost:6116/metrics
+
+$FOREST_CLI_PATH --token "$ADMIN_TOKEN" shutdown --force
```
LesnyRumcajs, line 124: Let's double check that the process `forest` is no longer there.

LesnyRumcajs (changes requested):

scripts/preloaded_wallet.key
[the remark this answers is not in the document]
LesnyRumcajs: Thank you!

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "$1" > preloaded_wallet.key
+
+SNAPSHOT_DIRECTORY="/tmp/snapshots"
+LOG_DIRECTORY="/tmp/log"
```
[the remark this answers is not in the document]
elmattic: Will fix, many thanks.

scripts/calibnet_health_check.sh
elmattic: Indeed.

scripts/calibnet_health_check.sh
elmattic: At what points?

If the script fails, who is responsible with reclaiming resources? (temp dirs, killing a running Forest, etc)

elmattic: Hmm I can do. Wondering just what could make shutdown command fail?

documentation/src/cli.md
```diff
@@ -65,9 +80,13 @@ default format for importing keys is hex encoded JSON. Use the `export` command
 to get formatted keys for importing. Usage:
 `forest-cli wallet import <hex encoded json key>`
 
-Export Export a key by address. Use a wallet address to export a key. Returns a
-formatted key to be used to import on another node, or into a new keystore.
-Usage: `forest-cli wallet export <address>`
+<<<<<<< HEAD Export Export a key by address. Use a wallet address to export a
```
LesnyRumcajs, line 83: fiasco

LesnyRumcajs (changes requested):

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,151 @@
...
+function cleanup {
+  # Removing temporary directory $TMP_DIR
+  rm -rf "$TMP_DIR"
+
+  $FOREST_CLI_PATH shutdown --force
+
+  sleep 10s
+  set +e
+  if pgrep -x forest; then
+    echo "Forest is still running"
```
LesnyRumcajs, line 31: In such a case, let's exit with an error.

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,151 @@
...
+    usage
+fi
+
+function cleanup {
+  # Removing temporary directory $TMP_DIR
+  rm -rf "$TMP_DIR"
+
+  $FOREST_CLI_PATH shutdown --force
+
+  sleep 10s
```
LesnyRumcajs, line 28: It's a bit brittle and potentially too long. How about a loop with a timeout, checking every second up to a minute?

LesnyRumcajs (changes requested):

scripts/calibnet_health_check.sh
```diff
@@ -0,0 +1,147 @@
...
+}
+
+if [ -z "$1" ]
+  then
+    usage
+fi
+
+function cleanup {
+  # Removing temporary directory $TMP_DIR
+  rm -rf "$TMP_DIR"
```
LesnyRumcajs, line 24: we don't need to handle that

elmattic: Ok!

PR #2385, opened 2023-01-04
Use appropriate units when printing wallet balances. 

### LINKED ISSUE #2323: Use appropriate units when printing wallet balances.
**Issue summary**

Right now we always print wallet balances with six fractional digits. Example:
```
$ ./target/release/forest-cli wallet list
Address                                   Default Balance
f1ahjaxhd4w4qu47t2kasdescbttcgv2ha2doaney           0.000000 FIL
f1necwjawankvg7yyszvzm4oqntlculkfklyzug2q  X        0.000000 FIL
f1phxedvr5653h7gvf7jfg6plrbcgyjl2lkjlxhgi           0.000000 FIL
f1uatkwrkzfn5yuw67sxoosh37rujleu6xvb2dhwq           100.000000 FIL
```

It would improve the user experience if we default to fewer digits but use appropriate units. For example, someone has 0.001 FIL then it should be printed out as "1 milli FIL" or "1 mFIL". We should also limit the output to, say, 4 significant digits. This can be overridden by a flag, `--exact-balance`. Additionally, the `--fixed-unit` flag should force the balance to be in FIL.

Tasks:
- [ ] Use 4 significant digits.
- [ ] Use appropriate units (FIL, milli FIL, atto FIL, etc).
- [ ] Add `--exact-balance` flag.
- [ ] Add `--fixed-unit` flag.
- [ ] Inlcude `~` when the printed balance is not exact.

Example output:
```
$ ./target/release/forest-cli wallet list
Address                                   Default Balance
f1ahjaxhd4w4qu47t2kasdescbttcgv2ha2doaney           0 FIL
f1necwjawankvg7yyszvzm4oqntlculkfklyzug2q  X        8 atto FIL
f1phxedvr5653h7gvf7jfg6plrbcgyjl2lkjlxhgi           ~500.2 milli FIL
f1uatkwrkzfn5yuw67sxoosh37rujleu6xvb2dhwq           100 FIL
```

```
$ ./target/release/forest-cli wallet list --exact-balance
Address                                   Default Balance
f1ahjaxhd4w4qu47t2kasdescbttcgv2ha2doaney           0 FIL
f1necwjawankvg7yyszvzm4oqntlculkfklyzug2q  X        8 atto FIL
f1phxedvr5653h7gvf7jfg6plrbcgyjl2lkjlxhgi           500.236 milli FIL
f1uatkwrkzfn5yuw67sxoosh37rujleu6xvb2dhwq           100 FIL
```

```
$ ./target/release/forest-cli wallet list --fixed-unit
Address                                   Default Balance
f1ahjaxhd4w4qu47t2kasdescbttcgv2ha2doaney           0 FIL
f1necwjawankvg7yyszvzm4oqntlculkfklyzug2q  X        ~0 FIL
f1phxedvr5653h7gvf7jfg6plrbcgyjl2lkjlxhgi           ~0 FIL
f1uatkwrkzfn5yuw67sxoosh37rujleu6xvb2dhwq           100 FIL
```

```
$ ./target/release/forest-cli wallet list --fixed-unit --exact-balance
Address                                   Default Balance
f1ahjaxhd4w4qu47t2kasdescbttcgv2ha2doaney           0 FIL
f1necwjawankvg7yyszvzm4oqntlculkfklyzug2q  X        0.000000000000000008 FIL
f1phxedvr5653h7gvf7jfg6plrbcgyjl2lkjlxhgi           0.500236 FIL
f1uatkwrkzfn5yuw67sxoosh37rujleu6xvb2dhwq           100 FIL
```

tyshko5:
**Summary of changes**
- 

Closes #2323


lemmih:
What is missing from this PR?

LesnyRumcajs (changes requested):
@tyshko5 It is a user-facing change and should be reflected in the changelog.

Also, this feature just begs for a 100% test coverage.

LesnyRumcajs:
Seems a bit stale; what is the status of this PR?

tyshko5:
> @tyshko5 It seems a bit stale; what is the status of this PR?

I'm working on it

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -235,3 +286,35 @@ impl WalletCommands {
         }
     }
 }
+
+fn formatring_vars(balance_string: String, balance_exact: String) -> (String, String) {
```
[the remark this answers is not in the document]
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -52,7 +52,12 @@ pub enum WalletCommands {
         path: Option<String>,
     },
     /// List addresses of the wallet
-    List,
+    List {
+        #[structopt(short, long)]
+        exact_balance: bool,
+        #[structopt(short, long)]
+        fixed_unit: bool,
```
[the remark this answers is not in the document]
tyshko5: Added

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -237,3 +254,94 @@ impl WalletCommands {
...
+            balance_exact.strip_suffix(".0").unwrap().to_string()
+        } else {
+            balance_exact
+        };
+        (res, "".to_string())
+    } else {
+        //unfallible unwrap
+        let re = Regex::new(r"(\.0+|\.)$").unwrap();
+        let res = re.replace(balance_string.as_str(), "").to_string();
+        (res, "~".to_string())
```
lemmih, line 271: I don't think regular expressions are necessary. Try discussing solutions in the slack channel.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -237,3 +254,94 @@ impl WalletCommands {
...
+        } else if *atto < BigInt::from(1000000000000i64) {
+            unit = "nano FIL";
+            balance_int *= BigInt::from(1000000000);
+        } else if *atto < BigInt::from(1000000000000000i64) {
+            unit = "micro FIL";
+            balance_int *= BigInt::from(1000000);
+        } else if *atto < BigInt::from(1000000000000000000i64) {
+            unit = "milli FIL";
+            balance_int *= BigInt::from(1000);
+        }
```
lemmih, line 307: There are a few problems with this code:

1. `balance_int` has type `TokenAmount` which is denominated in FIL but you're using it for other denominations. This is a conceptual type error.
2. You're using hard-coded values that are very error-prone. It would be easy to accidentally get `1000000000000000000i64` wrong. I'm sure this can be done in a way that's less error-prone and more readable.
3. The code is fairly repetitive. Each branch in the if-else chain does nearly the same thing.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -237,3 +254,94 @@ impl WalletCommands {
...
+    );
+}
+
+#[test]
+fn not_exact_balance_not_fixed_unit() {
+    assert_eq!(
+        format_balance_string(TokenAmount::from_atto(100), &false, &false,),
+        "100 atto FIL"
+    );
+}
```
lemmih, line 347: These tests are far from being exhaustive. You can definitely test 100% of your code.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -54,7 +55,17 @@ pub enum WalletCommands {
         path: Option<String>,
     },
     /// List addresses of the wallet
-    List,
+    List {
+        //flag to show exact balance meaming
```
LesnyRumcajs, line 59: 1. This should be a documentation comment, so `///`
2. I cannot really deduce what I should expect from this and the other flag. Users probably will also struggle. Please provide examples.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -237,3 +254,94 @@ impl WalletCommands {
         }
     }
 }
+
+fn formating_vars(balance_string: String, balance_exact: String) -> (String, String) {
```
LesnyRumcajs, line 258: This method also deserves some loving tests; I have difficulty understanding what it's doing and some expected usage would greatly help.

LesnyRumcajs (changes requested):

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Added more

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Notable updates:
 * [libp2p] Use in house bitswap implementation. [#2445](https://github.com/ChainSafe/forest/pull/2445)
 * [libp2p] Ban peers with duration. Banned peers are automatically unbanned after a period of 1h. [#2396](https://github.com/ChainSafe/forest/pull/2396)
 * [config] `stats` and `compression` keys in `parity_db` section were renamed to `enable_statistics` and `compression_type` respectively. [#2444](https://github.com/ChainSafe/forest/pull/2444)
+* [forest cli] changed how balances are displayed defualting to 1) showing in milli or atto format if it's appropiate to do so; 2) 4 significant digits. Consequently cli flags to override this behaviour were added: 1) `--fixed-unit` to force to show in original `FIL` unit and 2) `--exact-balance` to force full accuracy. [#2385](https://github.com/ChainSafe/forest/pull/2385)
```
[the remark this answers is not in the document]
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
tyshko5: This is a helper function, and other than its current usages it won't have any.
It's basically removes trailing fullstops, zeros as well as decides should we use `~` symbol.

LesnyRumcajs (changes requested):
1. It doesn't seem that you are doing the correct rounding in the code. To my understanding, this test should pass
```rust
    assert_eq!(
        format_balance_string(TokenAmount::from_atto(999999999999999999i64), &true, &false,),
        "~1 FIL"
    );
```
Instead, it gives `~0 FIL`, which is a bit pessimistic.

2. Too many `to_string()`. I'm pretty sure we don't need owned strings all around this unit.
3. The usage of `&true` is pretty awkward. References to primitive types don't make much sense; grab them by value.

The code gets pretty complex; perhaps it's worth using a dedicated crate like https://github.com/iliekturtles/uom

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ Notable updates:
 * [libp2p] Use in house bitswap implementation. [#2445](https://github.com/ChainSafe/forest/pull/2445)
 * [libp2p] Ban peers with duration. Banned peers are automatically unbanned after a period of 1h. [#2396](https://github.com/ChainSafe/forest/pull/2396)
 * [config] `stats` and `compression` keys in `parity_db` section were renamed to `enable_statistics` and `compression_type` respectively. [#2444](https://github.com/ChainSafe/forest/pull/2444)
+* [forest cli] changed how balances are displayed defualting to 
+  - showing in `milli` or `atto` format if it's appropiate to do so, consequently cli flag `--fixed-unit` added to force to show in original `FIL` unit
```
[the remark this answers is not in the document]
tyshko5: Fixed

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ Notable updates:
...
+  - 4 significant digits, consequently cli flag `--exact-balance` added to force full accuracy.
```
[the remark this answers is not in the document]
tyshko5: Fixed

lemmih:
@tyshko5 How is this PR coming along?

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -291,20 +292,16 @@ impl WalletCommands {
...
-    if balance_exact.len() <= balance_string.len() {
-        let res = if balance_exact.ends_with(".0") {
-            //unfallible unwrap
-            balance_exact.strip_suffix(".0").unwrap().to_string()
-        } else {
-            balance_exact
-        };
-        (res, "".to_string())
+    //unfallible unwrap
+    let re = Regex::new(r"(?i)(0+$)|(\.0+$)|(\.+$)").unwrap();
```
hanabi1224, line 296: Make regex static?

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +276,246 @@ impl WalletCommands {
...
+            token_string.parse::<Decimal>().unwrap(),
+            bool_pair_to_mode(true, true)
+        ),
+        "0.000000000000012465 FIL"
+    );
+}
+
+#[test]
+fn not_exact_balance_fixed_unit() {
+    let mut token_string = TokenAmount::from_atto(100).atto().to_string();
```
LesnyRumcajs, line 360: That's a lot of boilerplate code. We could shorten it significantly with this pattern https://github.com/ChainSafe/forest/blob/main/forest/cli/src/cli/mod.rs#L194

LesnyRumcajs: Also, there's no point in creating a `token_string`. We can use the decimal directly; I assume `BigInt` is properly tested on its own.

LesnyRumcajs (changes requested):

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -17,15 +17,25 @@ use forest_key_management::json::KeyInfoJson;
...
 use rpassword::read_password;
+use rust_decimal::prelude::*;
+use rust_decimal_macros::dec;
 
 use super::{handle_rpc_err, Config};
 
+#[allow(clippy::enum_variant_names)]
+enum FormattingMode {
+    ExactFixed,
+    NotExactFixed,
```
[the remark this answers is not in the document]
tyshko5: I don't want to introduce new naming, as it would only confuse me. Also this won't fix the clippy warning, as is is caused by all enum variant be suffixed by `Fixed`.

forest/cli/src/cli/wallet_cmd.rs
tyshko5: It's done not to test `BigInt`, but to recreate real life usage.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ Notable updates:
 * [libp2p] Use in house bitswap implementation. [#2445](https://github.com/ChainSafe/forest/pull/2445)
 * [libp2p] Ban peers with duration. Banned peers are automatically unbanned after a period of 1h. [#2396](https://github.com/ChainSafe/forest/pull/2396)
 * [config] `stats` and `compression` keys in `parity_db` section were renamed to `enable_statistics` and `compression_type` respectively. [#2444](https://github.com/ChainSafe/forest/pull/2444)
+* [forest cli] changed how balances are displayed defualting to 
+  - adding metric prefix when it's appropriate to do so, consequently CLI flag `--fixed-unit` added to force to show in original `FIL` unit
+  - 4 significant digits, consequently CLI flag `--exact-balance` added to force full accuracy.
+  [#2385](https://github.com/ChainSafe/forest/pull/2385)
+
```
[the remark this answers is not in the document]
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
LesnyRumcajs: What real life usage? We are not passing those `BigInt`s anywhere. It's not increasing code coverage at all.

The pattern I had in mind was
```
let cases = [
(input, expected_output),
(input, expected_output),
];

assert_eq!(expected_output, tested_method(input);
```
It would shorten the code significantly. There's no point in calling `parse` or `from` everywhere.

forest/cli/src/cli/wallet_cmd.rs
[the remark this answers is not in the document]
LesnyRumcajs: Why would it confuse you? I mean, `not exact` is kind of the same as `rounded`, no?

It would fix the clippy warning if you switch places. So `FixedExact, NotFixedRounded` etc.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -17,15 +17,25 @@ use forest_key_management::json::KeyInfoJson;
 use forest_rpc_client::wallet_ops::*;
 use forest_shim::address::{Address, Protocol};
 use forest_utils::io::read_file_to_string;
+#[allow(unused_imports)]
```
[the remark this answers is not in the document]
tyshko5: Removed

forest/cli/src/cli/wallet_cmd.rs
tyshko5: I applied the pattern

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +276,246 @@ impl WalletCommands {
         }
     }
 }
+
+fn format_balance_string(num: Decimal, mode: FormattingMode) -> String {
+    let units = [
+        "atto", "femto", "pico", "nano", "micro", "milli", "", "kilo", "mega", "giga", "tera",
+        "peta", "exa", "zetta", "yotta",
```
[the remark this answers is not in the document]
tyshko5: Removed

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +272,133 @@ impl WalletCommands {
...
+
+        for (atto, result) in cases_vec {
+            test_call(atto, result, true, true);
+        }
+    }
+
+    #[test]
+    fn not_exact_balance_fixed_unit() {
+        let cases_vec = vec![
+            (100i64, "~0 FIL"),
```
LesnyRumcajs, line 351: All those `i64` suffixes are not needed anymore.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -57,7 +63,25 @@ pub enum WalletCommands {
         path: Option<String>,
     },
     /// List addresses of the wallet
-    List,
+    List {
+        /// flag to force full accuracy,
+        /// not just default 4 significant digits
```
LesnyRumcajs, line 68: This is not true in the current implementation, no? There are three significant digits.
So You won't get a value like 500.2367; you will get 500.237. Unless I am missing something, let's update the documentation and add this use case (not fixed, not exact) to ensure the documentation matches the actual behaviour.
```
            (500236779800000000, "~500.237 milli FIL"),
```

LesnyRumcajs (changes requested):

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed to to be 4 significant digits. Added the test case

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +274,140 @@ impl WalletCommands {
...
+            );
+            let mut res = format!("{} FIL", fil.trunc());
+            if fil != fil_orig {
+                res.insert(0, '~');
+            }
+            res
+        }
+        FormattingMode::ExactNotFixed => format!("{num:0} {} FIL", units[unit_index]),
+        FormattingMode::NotExactNotFixed => {
+            let mut fil = num.round_dp_with_strategy(
```
lemmih, line 308: Think we want `round_sf_with_strategy` here. Significant digits = `fs`, decimal points = `dp`.

tyshko5: Tested this function out and it forces 4 digits, even with trailing zeros, so I reverted it to former state, since it worked perfectly well

lemmih (changes requested):
I think you got the wrong idea of what 'significant digits' mean. It is not the same as the numbers after the decimal point. You can read more about them here: https://en.wikipedia.org/wiki/Significant_figures

You need to update the documentation, update the code, and update the examples to actually use 4 significant digits.

tyshko5:
> I think you got the wrong idea of what 'significant digits' mean. It is not the same as the numbers after the decimal point. You can read more about them here: https://en.wikipedia.org/wiki/Significant_figures
> 
> You need to update the documentation, update the code, and update the examples to actually use 4 significant digits.

Done

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -57,7 +65,25 @@ pub enum WalletCommands {
...
+        /// E.g. 0.0000002367798 `FIL` instead of 0 `FIL`
+        #[arg(short, long)]
+        exact_balance: bool,
+        /// flag to force the balance to be in `FIL`
+        /// meaning one won't balance in `atto` or `micro`
+        /// form even if it is appropriate
+        /// E.g. 0.5002 `FIL` instead of 500.2367 `milli FIL`
+        /// In combination with `--exact-balance` flag
+        /// it will show exact data in `FIL` units
+        /// E.g. 0.0000002367798 `FIL` instead of 0 `FIL`
```
lemmih, line 83:
```suggestion
        /// E.g. 0.0000002367798 `FIL` instead of ~0 `FIL`
```

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -57,7 +65,25 @@ pub enum WalletCommands {
...
     },
     /// List addresses of the wallet
-    List,
+    List {
+        /// flag to force full accuracy,
+        /// not just default 4 significant digits
+        /// E.g. 500.2367798 `milli FIL` instead of 500.2 `milli FIL`
+        /// In combination with `--fixed-unit` flag
+        /// it will show exact data in `FIL` units
+        /// E.g. 0.0000002367798 `FIL` instead of 0 `FIL`
```
lemmih, line 74:
```suggestion
        /// E.g. 0.0000002367798 `FIL` instead of ~0 `FIL`
```

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -57,7 +65,25 @@ pub enum WalletCommands {
...
+        #[arg(short, long)]
+        exact_balance: bool,
+        /// flag to force the balance to be in `FIL`
+        /// meaning one won't balance in `atto` or `micro`
```
lemmih, line 78:
```suggestion
        /// without SI unit prefixes (like `atto` or `micro`)
```

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -57,7 +65,25 @@ pub enum WalletCommands {
...
+        /// form even if it is appropriate
```
lemmih, line 79:
```suggestion
```

lemmih (changes requested):

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +274,147 @@ impl WalletCommands {
...
+
+        for (atto, result) in cases_vec {
+            test_call(atto, result, true, false);
+        }
+    }
+
+    #[test]
+    fn not_exact_balance_not_fixed_unit() {
+        let cases_vec = vec![
+            (100, "100 atto FIL"),
```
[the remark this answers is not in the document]
tyshko5: Added

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +274,147 @@ impl WalletCommands {
...
+            (500236779800000000, "~500.2 milli FIL"),
+        ];
+
+        for (atto, result) in cases_vec {
+            test_call(atto, result, false, false);
+        }
+    }
+
+    fn test_call(atto: i64, result: &str, exact: bool, fixed: bool) {
+        let token_string = TokenAmount::from_atto(atto).atto().to_string();
```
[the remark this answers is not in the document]
tyshko5: Since rpc calls are using `TokenAmount` and not `Decimal`, which is just a helper struct, I added it to completely replicate actual use cases.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +273,153 @@ impl WalletCommands {
         }
     }
 }
+
+fn format_balance_string(num: Decimal, mode: FormattingMode) -> anyhow::Result<String> {
```
lemmih, line 277: How about we change this function to take a `TokenAmount` and then make it fail if we cannot convert to a `Decimal`? This failure should be tested.

lemmih (changes requested):

forest/cli/src/cli/wallet_cmd.rs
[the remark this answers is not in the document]
tyshko5: RPC sends a `TokenAmount`, we parse it as `Decimal`. The tests would be just wrong if I didn't add that code.

forest/cli/src/cli/wallet_cmd.rs
tyshko5: I already propagate the parsing error on line 211 in code. And test the parsing on line 418 in tests.

lemmih: You would get better type safety if you used `TokenAmount`. The code would even be slightly more self-documenting.

lemmih (changes requested):
I don't feel you've addressed my previous concerns.

forest/cli/src/cli/wallet_cmd.rs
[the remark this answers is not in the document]
tyshko5: Changed this, as function signature changed due to your comment

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Done this

tyshko5:
> I don't feel you've addressed my previous concerns.

Hope they are addressed now.

lemmih (commented):
There are three points where `format_balance_string` can fail. Is it possible cover those cases with tests?

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +278,156 @@ impl WalletCommands {
         }
     }
 }
+
+fn format_balance_string(
+    token_amount: TokenAmount,
+    mode: FormattingMode,
+) -> anyhow::Result<String> {
+    let units = ["atto", "femto", "pico", "nano", "micro", "milli", ""];
+    let num = token_amount.atto().to_string().parse::<Decimal>()?;
```
lemmih, line 287: You can convert a `TokenAmount` to a `Decimal` by going through an `i128` instead of a `String`.

lemmih (changes requested):

lemmih:
> > There are three points where `format_balance_string` can fail. Is it possible cover those cases with tests?
> 
> What are the points you are talking about?

Once where you parse the `Decimal` and twice where you do the rounding.

tyshko5:
> > > There are three points where `format_balance_string` can fail. Is it possible cover those cases with tests?
> > 
> > 
> > What are the points you are talking about?
> 
> Once where you parse the `Decimal` and twice where you do the rounding.

Decimal parsing always work, since we are parsing only values less than 2B.
As for rounding, looking through the source code I think we can never get `None` therefore it's not an issue.

lemmih:
> > > > There are three points where `format_balance_string` can fail. Is it possible cover those cases with tests?
> > > 
> > > 
> > > What are the points you are talking about?
> > 
> > 
> > Once where you parse the `Decimal` and twice where you do the rounding.
> 
> Decimal parsing always work, since we are parsing only values less than 2B. As for rounding, looking through the source code I think we can never get `None` therefore it's not an issue.

What if someone creates a Filecoin network with more than 2B FIL? If our code has limitations, those limitations should be documented and tested.

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Done

tyshko5:
> > > > > There are three points where `format_balance_string` can fail. Is it possible cover those cases with tests?
> > > > 
> > > > 
> > > > What are the points you are talking about?
> > > 
> > > 
> > > Once where you parse the `Decimal` and twice where you do the rounding.
> > 
> > 
> > Decimal parsing always work, since we are parsing only values less than 2B. As for rounding, looking through the source code I think we can never get `None` therefore it's not an issue.
> 
> What if someone creates a Filecoin network with more than 2B FIL? If our code has limitations, those limitations should be documented and tested.

If there's more than 20B it won't work. I've added a comment and I've added the test that.

lemmih (changes requested):
This causes a panic:

```rust
    #[test]
    fn panic() {
        assert_eq!(
            format_balance_string(
                TokenAmount::from_atto(79228162514264337593543950336i128),
                bool_pair_to_mode(true, true)
            )
            .unwrap_err()
            .to_string(),
            "value is too big"
        );
    }
```

lemmih:
`79228162514264337593543950336 = 2^96`

tyshko5:
> This causes a panic:
> 
> ```rust
>     #[test]
>     fn panic() {
>         assert_eq!(
>             format_balance_string(
>                 TokenAmount::from_atto(79228162514264337593543950336i128),
>                 bool_pair_to_mode(true, true)
>             )
>             .unwrap_err()
>             .to_string(),
>             "value is too big"
>         );
>     }
> ```

rust_decimal into can actually panic, which was the cause. I fixed the code with unpanicing version and added your test case

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +278,191 @@ impl WalletCommands {
         }
     }
 }
+
+fn format_balance_string(
```
lemmih, line 282: Tests are great but this function still needs some documentation.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -23,9 +23,21 @@ use fvm_shared::{
...
 use rpassword::read_password;
+use rust_decimal::prelude::*;
+use rust_decimal_macros::dec;
 
 use super::{handle_rpc_err, Config};
 
+const NUM_SIGNIFICANT_DIGITS: u32 = 4;
+
+#[allow(clippy::enum_variant_names)]
+enum FormattingMode {
```
lemmih, line 34: Some short and concise documentation for the formatting modes would be great.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +278,191 @@ impl WalletCommands {
...
+            format!("{fil} FIL")
+        }
+        FormattingMode::NotExactFixed => {
+            let fil_orig = orig / dec!(1e18);
+            let fil = fil_orig
+                .round_sf_with_strategy(
+                    NUM_SIGNIFICANT_DIGITS,
+                    RoundingStrategy::MidpointAwayFromZero,
+                )
+                .ok_or(anyhow::Error::msg("cannot represent"))?;
```
lemmih, line 318: Is it possible to write a test for this failure case? If not, why not?

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +278,191 @@ impl WalletCommands {
...
+    fn not_exact_balance_not_fixed_unit() {
+        let cases_vec = vec![
+            (100, "100 atto FIL"),
+            (120005, "~120 femto FIL"),
+            (200000045, "~200 pico FIL"),
+            (1000000123, "~1 nano FIL"),
+            (450000008000000, "~450 micro FIL"),
+            (90000002750000000, "~90 milli FIL"),
+            (500236779800000000, "~500.2 milli FIL"),
+            (1508900000000005000, "~1.509  FIL"),
```
lemmih, line 423: I don't think we want two spaces between the amount and the unit.

```suggestion
            (1508900000000005000, "~1.509 FIL"),
```

lemmih (changes requested):

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Fixed

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Added

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -241,3 +278,191 @@ impl WalletCommands {
...
+    token_amount: TokenAmount,
+    mode: FormattingMode,
+) -> anyhow::Result<String> {
+    let units = ["atto", "femto", "pico", "nano", "micro", "milli", ""];
+    let num: Decimal = Decimal::try_from_i128_with_scale(
+        token_amount
+            .atto()
+            .to_i128()
+            // currently the amount cannot be more than 2B x 10^18 tokens
+            // the limit here is 20B
```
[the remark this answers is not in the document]
tyshko5: Fixed both

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -23,9 +23,33 @@ use fvm_shared::{
...
 
 use super::{handle_rpc_err, Config};
 
+const NUM_SIGNIFICANT_DIGITS: u32 = 4;
+
+#[allow(clippy::enum_variant_names)]
+enum FormattingMode {
+    /// mode to show data in `FIL` units
+    /// in full accuracy
+    /// E.g. 0.0000002367798 `FIL` instead of ~0 `FIL`
```
lemmih, line 37: I don't think it makes sense to say "X instead of Y" in this context.

Let's simplify the documentation to just a single, illustrative example of the output of each mode.

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Added the docs

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Looking through the source code of `round_sf_with_strategy` `None` value of `Option` enum is never returned. So basically this is infallible, therefore no tests are needed.

forest/cli/src/cli/wallet_cmd.rs
tyshko5: Done

forest/cli/src/cli/wallet_cmd.rs
lemmih: It's not exactly infallible but we can leave this for now.

forest/cli/src/cli/wallet_cmd.rs
lemmih: I meant documentation at the top of the function.

tyshko5: Added it too

lemmih (approved):
I'm not completely satisfied with how you responded to Hubert's comments about naming, I think a few more tests would be helpful, and the documentation is awkwardly phrased in places. But this PR needs to move forward so let's just merge it.

sudo-shashank (approved):

LesnyRumcajs (approved):

Issue #2356, opened 2022-12-12
Miscellaneous tasks


lemmih:
**Issue summary**

- [ ] #2679
- [X] Use the `tempfile` crate in `chain_export`. See `NamedTempFile::new_in` and `NamedTempFile::persist`. https://github.com/ChainSafe/forest/blob/364e7cb8e281db5f5ed69c5dbbf00eb89e051080/node/rpc/src/chain_api.rs#L80
- [x] Remove the `FullVerifier`. We used to have a mock verifier as well but it was rightfully removed. Now the `FullVerifier` doesn't make sense anymore. Removing will require touching quite a bit of code. https://github.com/ChainSafe/forest/blob/b82879cc6773ebf6ca747d122c35226453d87a41/types/src/verifier/mod.rs#L18
- [x] Remove all calls to `forest_statediff` in the daemon. State-diffing is handled by the standalone statediff executable. The statediff crate is used in multiple locations and they should all be removed. https://github.com/ChainSafe/forest/blob/b82879cc6773ebf6ca747d122c35226453d87a41/blockchain/state_manager/src/lib.rs#L1310
- [x] Get rid of `check_for_low_fd`. It is no longer necessary.
- [x] Move `open_db` to `forest_db`. https://github.com/ChainSafe/forest/blob/daf362e1472243edd59d5ffb17b23ff944001d04/forest/daemon/src/daemon.rs#L507-L527

This list contains tasks so small that they don't merit their own issue. The list is still growing.

**Other information and links**

tyshko5:
https://github.com/ChainSafe/forest/pull/2368

creativcoder:
I'm covering: 
> Move open_db to forest_db.

in https://github.com/ChainSafe/forest/pull/2342

tyshko5:
https://github.com/ChainSafe/forest/pull/2370

tyshko5:
https://github.com/ChainSafe/forest/pull/2380

tyshko5:
https://github.com/ChainSafe/forest/pull/2389

LesnyRumcajs:
@sudo-shashank are all tasks done, e.g.,
> Allow the forest crates to be compiled with both rocksdb and paritydb enabled. If both are enabled, either default to using rocksdb or add a flag for switching between the backends.

Or we decided not to do them?

lemmih:
Hoisted out the last issue.

PR #2389, opened 2023-01-05
Refactor open_db in the statediff crate 

### LINKED ISSUE #2387: Refactor `open_db` in the `statediff` crate
**Issue summary**

The `statediff` crate has its own copy of the `open_db` function. It should be refactored and merged with the `open_db` function used by `forest` and `forest-cli`. Some re-organization might be required to do this nicely.

https://github.com/ChainSafe/forest/blob/d5db45dfaa938c8c8135e960fa6af832f2b798e6/utils/statediff/src/main.rs#L37-L52
https://github.com/ChainSafe/forest/blob/d5db45dfaa938c8c8135e960fa6af832f2b798e6/forest/shared/src/lib.rs#L31-L46

tyshko5:
**Summary of changes**
- 

Closes #2387

### DIFF
--- a/forest/shared/src/lib.rs
+++ b/forest/shared/src/lib.rs
@@ -44,9 +44,10 @@ pub fn open_db(
 #[cfg(feature = "paritydb")]
 pub fn open_db(
     path: &std::path::Path,
-    config: &Option<cli::Config>,
+    config: Option<&cli::Config>,
 ) -> anyhow::Result<forest_db::parity_db::ParityDb> {
     use forest_db::parity_db::*;
+    use forest_db::parity_db_config::*;
     if let Some(config) = config {
         ParityDb::open(path.to_owned(), &config.parity_db)
     } else {
--- a/forest/shared/src/lib.rs
+++ b/forest/shared/src/lib.rs
@@ -4,7 +4,7 @@
 pub mod cli;
 pub mod logger;
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 #[cfg(feature = "rocksdb")]
 pub type Db = forest_db::rocks::RocksDb;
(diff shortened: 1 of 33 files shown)


hanabi1224 (approved):
I saw there's an item to make DBfeatures non-mutually-exclusive in https://github.com/ChainSafe/forest/issues/2356  should it be part of this PR?

> Allow the forest crates to be compiled with both rocksdb and paritydb enabled. If both are enabled, either default to using rocksdb or add a flag for switching between the backends.

forest/shared/src/lib.rs
```diff
@@ -31,16 +31,26 @@ pub fn db_path(config: &crate::cli::Config) -> PathBuf {
 #[cfg(feature = "rocksdb")]
 pub fn open_db(
     path: &std::path::Path,
-    config: &cli::Config,
+    config: Option<&cli::Config>,
```
lemmih, line 34: Let's use the default configuration instead of passing in `None` here.

utils/statediff/src/main.rs
```diff
@@ -26,7 +28,7 @@ impl ChainCommand {
     pub async fn run(&self) {
         let dir = ProjectDirs::from("com", "ChainSafe", "Forest").unwrap();
         let chain_path = dir.data_dir().join(&self.chain);
-        let blockstore = open_db(&chain_path);
+        let blockstore = open_db(&chain_path_suffix(&chain_path), None).unwrap();
```
lemmih, line 31: How about we use `db_path` here?

lemmih (changes requested):

forest/shared/src/lib.rs
tyshko5: Done

utils/statediff/src/main.rs
tyshko5: Done

forest/shared/src/lib.rs
```diff
@@ -4,43 +4,44 @@
 pub mod cli;
 pub mod logger;
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
```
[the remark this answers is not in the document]
LesnyRumcajs: Does it make sense @tyshko5?

utils/statediff/src/main.rs
```diff
@@ -26,31 +28,14 @@ impl ChainCommand {
     pub async fn run(&self) {
         let dir = ProjectDirs::from("com", "ChainSafe", "Forest").unwrap();
         let chain_path = dir.data_dir().join(&self.chain);
-        let blockstore = open_db(&chain_path);
+        let blockstore = open_db(&db_path(&chain_path), &DbConfig::default()).unwrap();
```
lemmih, line 31: Let's replace the `unwrap`s  with proper error handling.

lemmih (changes requested):

forest/shared/src/lib.rs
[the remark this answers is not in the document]
tyshko5: I need a struct from `forest_shared_cli` importing which creates a dependencies cycle, so such approach makes sense it's impossible. I can reorganise the code within `forest_shared_cli`, though

utils/statediff/src/main.rs
tyshko5: Fixed

forest/shared/src/lib.rs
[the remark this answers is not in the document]
LesnyRumcajs: Seems you need it only for `open_db` which accepts a `Config`, otherwise there shouldn't be any cycles, no?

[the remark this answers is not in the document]
tyshko5: Fixed accordingly to your comment

lemmih (approved):

PR #2394, opened 2023-01-06
chore: remove unnecessary async

### LINKED ISSUE #2343: Replace the lock around `LruCache` with a mutex
**Issue summary**

In several places, we wrap an `LruCache` in a `RwLock`. This makes little sense, though, as both reading and writing to the cache requires mutability. We would be better off using a `Mutex` from `parking_lot`.

- [ ] Identify all uses of `RwLock` with `LruCache`.
- [ ] Replace `RwLock` with `Mutex`.
- [ ] Remove `async` from function that are no longer async.

https://github.com/ChainSafe/forest/blob/be017aac13582734914e8aaa878f61379cdce514/blockchain/chain/src/store/chain_store.rs#L617
https://docs.rs/parking_lot/latest/parking_lot/type.Mutex.html

lemmih:
**Summary of changes**
- Keeping track of the heaviest tipset may never block and should therefore not be async.
- The TipsetCache may never block and should not be async.

Closes #2343


lemmih:
> @lemmih are you positive we need `parking_lot` mutex? As far as I remember, the `std::sync::Mutex` got better some time ago.
> 
> https://twitter.com/m_ou_se/status/1526211117651050497?lang=en (kind of a Rust celebrity, author of [Rust Atomics and Locks](https://www.amazon.com/Rust-Atomics-Locks-Low-Level-Concurrency/dp/1098119444))

We don't need it for performance. It's just more convenient to not have to deal with poisoned states.

LesnyRumcajs:
> > @lemmih are you positive we need `parking_lot` mutex? As far as I remember, the `std::sync::Mutex` got better some time ago.
> > https://twitter.com/m_ou_se/status/1526211117651050497?lang=en (kind of a Rust celebrity, author of [Rust Atomics and Locks](https://www.amazon.com/Rust-Atomics-Locks-Low-Level-Concurrency/dp/1098119444))
> 
> We don't need it for performance. It's just more convenient to not have to deal with poisoned states.

Was this an issue at any point?

lemmih:
> > > @lemmih are you positive we need `parking_lot` mutex? As far as I remember, the `std::sync::Mutex` got better some time ago.
> > > https://twitter.com/m_ou_se/status/1526211117651050497?lang=en (kind of a Rust celebrity, author of [Rust Atomics and Locks](https://www.amazon.com/Rust-Atomics-Locks-Low-Level-Concurrency/dp/1098119444))
> > 
> > 
> > We don't need it for performance. It's just more convenient to not have to deal with poisoned states.
> 
> Was this an issue at any point?

The mutexes should never be poisoned but you still have to call `unwrap` or propagate the poison error value. With parking_lot, all of that goes away. But feel free to use std mutex instead if you want.

blockchain/chain/src/store/tipset_tracker.rs
```diff
@@ -30,19 +29,19 @@ impl<DB: Blockstore> TipsetTracker<DB> {
...
+            if cids.contains(header.cid()) {
                 debug!("tried to add block to tipset tracker that was already there");
                 return;
             }
-        }
+            cids.push(*header.cid());
+            cids.clone()
+        };
 
+        // XXX: What is this code supposed to do? ~Lemmih
```
LesnyRumcajs, line 44: Maybe create a `spike` issue to document this block of code

blockchain/message_pool/src/msgpool/msg_pool.rs
```diff
@@ -168,24 +168,24 @@ where
...
-        let local_msgs = Arc::new(RwLock::new(HashSet::new()));
-        let republished = Arc::new(RwLock::new(HashSet::new()));
+        let local_addrs = Arc::new(SyncRwLock::new(Vec::new()));
+        let pending = Arc::new(SyncRwLock::new(HashMap::new()));
+        let tipset = Arc::new(Mutex::new(api.get_heaviest_tipset().ok_or_else(|| {
+            Error::Other("Failed to retrieve heaviest tipset from provider".to_owned())
+        })?));
+        let bls_sig_cache = Arc::new(Mutex::new(LruCache::new(BLS_SIG_CACHE_SIZE)));
+        let sig_val_cache = Arc::new(Mutex::new(LruCache::new(SIG_VAL_CACHE_SIZE)));
+        // let api_mutex = Arc::new(RwLock::new(api));
```
LesnyRumcajs, line 178:
```suggestion
```

LesnyRumcajs (approved):
It will probably be negligible performance-wise, but it reduces the mental overhead needed to analyze async code that doesn't need to be async. Great idea, @lemmih.

blockchain/chain/src/store/index.rs
```diff
@@ -9,9 +9,9 @@ use fvm_ipld_blockstore::Blockstore;
 use fvm_shared::clock::ChainEpoch;
 use log::info;
 use lru::LruCache;
+use parking_lot::Mutex as StdMutex;
```
lemmih, line 12:
```suggestion
use parking_lot::Mutex;
```

blockchain/state_manager/src/lib.rs
```diff
@@ -858,77 +852,59 @@ where
...
-        let state = StateTree::new_from_root(self.blockstore(), current.parent_state())
-            .map_err(|e| Err(Error::State(e.to_string())))?;
-
-        if let Some(actor_state) = state
-            .get_actor(message_from_address)
-            .map_err(|e| Err(Error::State(e.to_string())))?
-        {
-            if actor_state.sequence == 0 || actor_state.sequence < *message_sequence {
+    ) -> Result<Option<(Arc<Tipset>, Receipt)>, Error> {
+        loop {
```
elmattic, line 860: Why was the `loop` construct added? Looks more complex than just moving from `async` to `sync` here.

elmattic (approved):

blockchain/chain/src/store/chain_store.rs
```diff
@@ -75,7 +76,7 @@ pub struct ChainStore<DB> {
     pub db: DB,
 
     /// Tipset at the head of the best-known chain.
-    heaviest: RwLock<Option<Arc<Tipset>>>,
+    heaviest: Mutex<Option<Arc<Tipset>>>,
```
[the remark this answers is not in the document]
lemmih: A `RwLock` would also work but I'd rather just clone the `Arc` when reading it. That way, the mutex is never held for longer than it takes to clone an `Arc`. With a `RwLock`, one could accidentally hold the read lock for too long and thereby get in the way of the writer.

blockchain/chain_sync/src/chain_muxer.rs
```diff
@@ -438,7 +438,8 @@ where
                         .inc();
                     if let PubsubMessageProcessingStrategy::Process = message_processing_strategy {
                         // Spawn and immediately move on to the next event
-                        tokio::task::spawn(Self::handle_pubsub_message(mem_pool.clone(), m));
+                        let mem_pool = mem_pool.clone();
+                        tokio::task::spawn_blocking(|| Self::handle_pubsub_message(mem_pool, m));
```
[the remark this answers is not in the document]
lemmih: Hmm, right, yeah, I don't think it's worth spawning a blocking thread here. Will update.

blockchain/message_pool/src/msg_chain.rs
```diff
@@ -331,8 +330,8 @@ impl std::default::Default for MsgChainNode {
     }
 }
 
-pub(crate) async fn create_message_chains<T>(
-    api: &RwLock<T>,
+pub(crate) fn create_message_chains<T>(
+    api: &T,
```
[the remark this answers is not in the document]
lemmih: Yes, the api used to have methods that used `&mut self` but it turns out they could all be converted to just `&self`. Without the need for mutability, the api provider can just be shared with a reference without any locking.

blockchain/state_manager/src/lib.rs
lemmih: The `loop` used to be in `search_back_for_message`. I merged the code because the return type of `check_search` was too complicated for me to understand: `Result<Option<(Arc<Tipset>, Receipt)>, Result<Arc<Tipset>, Error>>`

By moving the loop into `check_search`, the return type can be simplified to: `Result<Option<(Arc<Tipset>, Receipt)>, Error>`

The diff is a bit convoluted but the code is more-or-less the same.

Issue #2391, opened 2023-01-05
[libp2p] Implement soft ban machenism with duration


hanabi1224:
**Issue summary**

We currently ban peers for the entire process lifetime, we should softly ban them with a duration to let them reconnect afterward.

**Other information and links**

PR #2396, opened 2023-01-09
feat: ban peer with duration

hanabi1224:
**Summary of changes**
- [x] ban peer with duration
- [x] manually test ban and unban by setting duration to 5s locally

Closes https://github.com/ChainSafe/forest/issues/2391

## Change checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## Forest unreleased
+Notable updates:
+
+### Added
+
+### Changed
+* [libp2p] Ban peers with duration: [#2396](https://github.com/ChainSafe/forest/pull/2396)
+
+### Removed
+
 ## Forest v0.6.0 (2023-01-06)
 
 Notable updates:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Notable updates:
 ### Added
 
 ### Changed
-* [libp2p] Ban peers with duration: [#2396](https://github.com/ChainSafe/forest/pull/2396)
+* [libp2p] Ban peers with duration. Banned peers are automatically unbanned after a period of 1h.: [#2396](https://github.com/ChainSafe/forest/pull/2396)
 
 ### Removed
 
--- a/node/forest_libp2p/src/peer_manager.rs
+++ b/node/forest_libp2p/src/peer_manager.rs
@@ -2,12 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use crate::*;
+use flume::{Receiver, Sender};
 use forest_blocks::Tipset;
-use log::{debug, trace};
+use log::{debug, trace, warn};
 use rand::seq::SliceRandom;
 use std::collections::HashMap;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use std::{cmp::Ordering, collections::HashSet};
 use tokio::sync::RwLock;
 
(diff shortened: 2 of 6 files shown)


LesnyRumcajs:
It is kind of a user-facing change, no? I mean, renaming traits or making async stuff sync is not worth putting into the changelog, but I do believe features like this should be mentioned.

LesnyRumcajs (approved):
Good stuff

hanabi1224:
> It is kind of a user-facing change, no? I mean, renaming traits or making async stuff sync is not worth putting into the changelog, but I do believe features like this should be mentioned.

@LesnyRumcajs  Changelog updated.

node/forest_libp2p/src/peer_manager.rs
```diff
@@ -200,6 +219,61 @@ impl PeerManager {
...
+        if let Err(e) = self
+            .peer_ops_tx
+            .send_async(PeerOperation::Ban(peer, reason.into()))
+            .await
+        {
+            warn!("ban_peer err: {e}");
+        }
+    }
+
+    pub async fn background_task(self: Arc<Self>) -> anyhow::Result<()> {
```
elmattic, line 246: Maybe a more descriptive name for this function would help? `background_task` feels a bit too generic.

node/forest_libp2p/src/peer_manager.rs
```diff
@@ -200,6 +219,61 @@ impl PeerManager {
...
+        loop {
+            let mut unban_list = vec![];
```
elmattic, line 248: You could "hoist" the temporary vector outside the loop here and clear it at each iteration.

node/forest_libp2p/src/peer_manager.rs
elmattic, line 246: The return type of the method would more be "never" `!` here, no?

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Forest unreleased
+Notable updates:
+
+### Added
+
+### Changed
+* [libp2p] Ban peers with duration: [#2396](https://github.com/ChainSafe/forest/pull/2396)
```
LesnyRumcajs, line 7:
```suggestion
* [libp2p] Ban peers with duration. Banned peers are automatically unbanned after a period of 24h.: [#2396](https://github.com/ChainSafe/forest/pull/2396)
```

elmattic (approved):

CHANGELOG.md
elmattic: Where does the 24h period come from?

node/forest_libp2p/src/peer_manager.rs
hanabi1224: This is to fit the caller signature in daemon crate, otherwise we need to wrap this with a lambda

CHANGELOG.md
hanabi1224: It's 1h atm. Would u suggest changing the code to 24h or the doc to 1h?

node/forest_libp2p/src/peer_manager.rs
hanabi1224: How about `peer_operation_event_loop_task`

node/forest_libp2p/src/peer_manager.rs
hanabi1224: Fixed

CHANGELOG.md
LesnyRumcajs: @hanabi1224 @elmattic Hm, I must've dreamt about it. In general, I'm for a 24h ban.

elmattic: Yes, 24h ban seems good.

PR #2412, opened 2023-01-11
fix: cleanup cargo dependencies

hanabi1224:
## Summary of changes

- implement ruby script for detecting unused cargo dependencies
- update CI
- cleanup cargo manifests

Closes https://github.com/ChainSafe/forest/issues/2413

## Change checklist


scripts/find_unused_deps.rb
```diff
@@ -0,0 +1,31 @@
...
+Dir.glob("#{File.dirname(File.expand_path($PROGRAM_NAME))}/../**/*.toml").each do |file|
+  crate_dir = File.dirname(file)
+  toml = TomlRB.load_file(file)
+  crates = Set.new
+  toml['dependencies']&.each do |pair|
+    crates.add pair[0]
+  end
+  toml['dev-dependencies']&.each do |pair|
+    crates.add pair[0]
+  end
```
LesnyRumcajs, line 17: This should work, but please double-check. I think it's more expressive.
```suggestion
  toml['dependencies']&.each do |crate_name, _|
    crates.add crate_name
  end
  toml['dev-dependencies']&.each do |crate_name, _|
    crates.add crate_name
  end
```

scripts/find_unused_deps.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'toml-rb'
+require 'set'
+
+exit_code = 0
+
+Dir.glob("#{File.dirname(File.expand_path($PROGRAM_NAME))}/../**/*.toml").each do |file|
```
LesnyRumcajs, line 8: This seems a bit complex, I'd say we can just `Dir.glob('**/*.toml')` and run it from the project's root.

scripts/find_unused_deps.rb
```diff
@@ -0,0 +1,31 @@
...
+    crates.add pair[0]
+  end
+  toml['dev-dependencies']&.each do |pair|
+    crates.add pair[0]
+  end
+  crates.each do |crate_raw|
+    crate = crate_raw.gsub(/-/, '_')
+    used = false
+    Dir.glob("#{crate_dir}/**/*.rs").each do |rs|
+      used = true if !used && File.read(rs).match?(Regexp.new("(\buse\\s#{crate}\\b)|(\\b#{crate}::)"))
```
LesnyRumcajs, line 22: I believe this code golf should work, but please double-check.
```suggestion
      used |= File.read(rs).match?(Regexp.new("(\buse\\s#{crate}\\b)|(\\b#{crate}::)"))
```

LesnyRumcajs (approved):
This should be shipped with `cargo`. Excellent check!

LesnyRumcajs:
In general, does it make `cargo-udeps` obsolete, and if so can we get rid of it? I mean, if this script detects a superset of what `cargo-udeps` does then there's no point in keeping the latter in the CI.

forest/shared/Cargo.toml
```diff
@@ -8,66 +8,35 @@ edition = "2021"
...
-libp2p = { workspace = true, default-features = false, features = ["identify"] }
 log = { workspace = true, features = ["serde"] }
-multibase.workspace = true
-num-rational.workspace = true
-num-traits.workspace = true
 once_cell.workspace = true
-prometheus = { workspace = true, features = ["process"] }
 regex.workspace = true
-rust-s3 = { version = "0.32", default-features = false, features = ["sync-rustls-tls", "tags"] }
+s3 = { package = "rust-s3", version = "0.32", default-features = false, features = ["sync-rustls-tls", "tags"] }
```
lemmih, line 29: This is interesting. I'm surprised it worked before.

lemmih (approved):
Kinda sad that `cargo-udeps` doesn't work reliably with workspaces.

scripts/find_unused_deps.rb
hanabi1224: Fixed, it works

scripts/find_unused_deps.rb
hanabi1224: Fixed

scripts/find_unused_deps.rb
hanabi1224: Fixed, it works

forest/shared/Cargo.toml
hanabi1224: The crate [specifies `s3` as lib name](https://docs.rs/crate/rust-s3/latest/source/Cargo.toml.orig), this change is to avoid false positive from the script cuz it's using a very simple regex rule

hanabi1224:
I believe `cargo-udeps` is using a more formal way that invokes the compiler AST parser and re-uses the cargo manifest parser, I'm not confident about whether this simple script can fully replace it.

PR #2416, opened 2023-01-11
chore: refactor ChainStore::genesis to inherit the genesis value from the chain config 

### LINKED ISSUE #2410: chore: refactor `ChainStore::genesis` to inherit the genesis value from the chain config
**Issue summary**

Right now, `ChainStore::genesis` has the option of being undefined. This causes a lot of `unwrap`s in other parts of the code which require the genesis block to be set.

We should take the genesis information from the chain configuration and use it to initialize the genesis key in the database. That way, `ChainStore::genesis` will always return the right value.

- [ ] Make sure the database always contains a reference to the genesis block (see `set_genesis`).
- [ ] Update the type of `ChainStore::genesis` and update all calls to this function.

tyshko5:
## Summary of changes

- 

Closes #2410

## Change checklist


utils/genesis/src/lib.rs
```diff
@@ -103,12 +103,9 @@ where
...
 
-    let store_genesis = chain_store.genesis()?;
+    let store_genesis = chain_store.genesis();
 
-    if store_genesis
-        .map(|store| store == genesis_block)
-        .unwrap_or_default()
-    {
+    if store_genesis.is_ok() && store_genesis.unwrap() == genesis_block {
         debug!("Genesis from config matches Genesis from store");
```
lemmih, line 109: This debug message isn't important. Let's refactor the code to always set the genesis.

lemmih (changes requested):

utils/genesis/src/lib.rs
tyshko5: Removed the message

utils/genesis/src/lib.rs
```diff
@@ -103,14 +103,9 @@ where
...
-    let store_genesis = chain_store.genesis()?;
+    let store_genesis = chain_store.genesis();
 
-    if store_genesis
-        .map(|store| store == genesis_block)
-        .unwrap_or_default()
-    {
-        debug!("Genesis from config matches Genesis from store");
-    } else {
+    if store_genesis.is_err() {
```
lemmih, line 108: You can delete `store_genesis` altogether. We always want to update the genesis in the DB.

lemmih (changes requested):

utils/genesis/src/lib.rs
tyshko5: Deleted

lemmih:
Good. You still need to go through the four check boxes in the PR description.

lemmih:
You can run forest locally to see what's going wrong.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -731,14 +731,13 @@ where
 }
 
 /// Returns the genesis block from storage.
-pub fn genesis<DB>(db: &DB) -> Result<Option<BlockHeader>, Error>
+pub fn genesis<DB>(db: &DB) -> Result<BlockHeader, Error>
```
lemmih, line 734: This function should only be used from the `ChainStore`. Let's move it into `ChainStore::genesis`.

lemmih: And let's do the same with `set_genesis`. That function should also only be used through the `ChainStore`.

blockchain/chain/src/store/chain_store.rs
```diff
@@ -731,14 +731,13 @@ where
...
 where
     DB: Blockstore + Store,
 {
-    Ok(db
-        .read(GENESIS_KEY)?
-        .map(|bz| BlockHeader::unmarshal_cbor(&bz))
-        .transpose()?)
+    db.read(GENESIS_KEY)?
+        .map(|bz| BlockHeader::unmarshal_cbor(&bz).map_err(|e| Error::Other(e.to_string())))
+        .ok_or(Error::Other("no genesis".to_string()))?
```
lemmih, line 740: Maybe the error message could be `Genesis key not defined in database`

lemmih (changes requested):

blockchain/chain/src/store/chain_store.rs
tyshko5: Changed

blockchain/chain/src/store/chain_store.rs
tyshko5: Moved

LesnyRumcajs (approved):

sudo-shashank (approved):

Issue #2405, opened 2023-01-10
Bug: forest daemon crashes during bitswap get query


creativcoder:
**Describe the bug**

Forest daemon crashes with snapshot: filecoin_snapshot_calibnet_2022-12-09_height_108720.car
using libp2p-bitswap fork at: https://github.com/hanabi1224/libp2p-bitswap?branch=forest-libp2p@0.50#6f14bc98846abb4a7e1ee8739d199a55bcfcb441

**To Reproduce**
Steps to reproduce the behavior (WIP):
2. Run '....'
3. See error

**Log output**

<details>
  <summary>Log Output</summary>
  
```
2023-01-10T11:22:59.031729Z  WARN forest_libp2p::service: Fail to send ChainExchange response    
thread 'tokio-runtime-worker' panicked at 'assertion failed: state.block.is_some()', /root/.cargo/git/checkouts/libp2p-bitswap-60205aaf66f912c5/6f14bc9/src/query.rs:216:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[1]    1484452 IOT instruction (core dumped)  target/release/forest --chain calibnet --target-peer-count 50  false
```
</details> 

**Expected behaviour**

**Screenshots**

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04 LTS x86_64, kernel: 5.15.0-52-generic
 - Rust version(e.g. `rustc 1.68.0-nightly (b70baa4f9 2022-12-14))
 - Branch/commit: bd4c43e40bf527f3cc839d5c095b65899e4031f4

**Other information and links**

hanabi1224:
I will take a look

PR #2419, opened 2023-01-12
fix: forest daemon crashes on sending bitswap requests

hanabi1224:
## Summary of changes

- check if peer list is empty before sending bitswap requests

Related code: https://github.com/ipfs-rust/libp2p-bitswap/blob/master/src/query.rs#L216
```rust
/// Starts a query to locate and retrieve a block. Panics if no providers are supplied.
    pub fn get(
        &mut self,
        parent: Option<QueryId>,
        cid: Cid,
        providers: impl Iterator<Item = PeerId>,
    ) -> QueryId {
        ...
        for peer in providers {
            if state.block.is_none() {
                state.block = Some(self.block(root, id, peer, cid));
            } else {
                state.have.insert(self.have(root, id, peer, cid));
            }
        }
        assert!(state.block.is_some());
       ...
```

Closes https://github.com/ChainSafe/forest/issues/2405

## Change checklist


LesnyRumcajs (approved):
I think it's worth putting it in the `fixed` section of the changelog

lemmih (approved):

creativcoder (approved):

hanabi1224:
> I think it's worth putting it in the `fixed` section of the changelog

Changelog updated

Issue #2408, opened 2023-01-10
chore: refactor `DrandBeacon` to not use async


lemmih:
**Issue summary**

`DrandBeacon::new` uses async to perform a sanity check when compiled in debug mode. This sanity check could be done in a background thread without tainting the function as async.

- [ ] Run the sanity check in a background thread.
- [ ] Remove the async from `DrandBeacon::new`, `get_beacon_schedule`, and `StateManger::new`.

https://github.com/ChainSafe/forest/blob/11532f358e5cddc1e13b0fd7aad1bbb665c9ed10/blockchain/beacon/src/drand.rs#L207-L217

**Other information and links**

PR #2427, opened 2023-01-13
Remove async from DrandBeacon construction

LesnyRumcajs:
## Summary of changes

- removed async from `DrandBeacon` construction - a blocking check should be acceptable there given that it's only for debugging,
- removed unnecessary async that resulted from the above,
- revived one test (though I had to disable height check because of a missing stateroot, it's probably because of our lack of support for older network versions for now)

Closes https://github.com/ChainSafe/forest/issues/2408

## Change checklist

### DIFF
--- a/blockchain/beacon/src/drand.rs
+++ b/blockchain/beacon/src/drand.rs
@@ -189,7 +189,7 @@ pub struct DrandBeacon {
 
 impl DrandBeacon {
     /// Construct a new `DrandBeacon`.
-    pub async fn new(
+    pub fn new(
         genesis_ts: u64,
         interval: u64,
         config: &DrandConfig<'_>,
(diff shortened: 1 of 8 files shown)


lemmih:
Sorry I didn't spot this earlier but could you make `MessagePool::new` be non-async as well? It just needs to have the `async` keyword removed.

blockchain/beacon/src/drand.rs
```diff
@@ -201,15 +201,18 @@ impl DrandBeacon {
...
-            let client = https_client();
-            let remote_chain_info: ChainInfo = client
-                .get(format!("{}/info", &config.server).try_into()?)
-                .await?
-                .into_body()
-                .json()
-                .await
-                .map_err(|e| anyhow::anyhow!("{}", e))?;
-            debug_assert!(&remote_chain_info == chain_info);
+            futures::executor::block_on(async {
```
hanabi1224, line 204: How about using [`tokio::runtime::Handle::current().block_on`](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on)

LesnyRumcajs: I think `futures::executor` is more clear and less tied to the brittle `tokio` interface. We have `futures` in our dependencies anyway. Is it fine for you?

LesnyRumcajs:
> Sorry I didn't spot this earlier but could you make `MessagePool::new` be non-async as well? It just needs to have the `async` keyword removed.

of course

blockchain/beacon/src/drand.rs
hanabi1224: That's fine if it works, I'm just worried that it will throw errors if "tokio::spawn" is called inside the async function

LesnyRumcajs:
> `state_setup` is also no longer async, I believe.

Indeed, I expected `clippy` to find this, but I guess my expectations were too high.

lemmih (approved):

hanabi1224 (approved):

Issue #2420, opened 2023-01-12
Lint: pedantic clippy warnings


LesnyRumcajs:
**Issue summary**

To further increase code quality, it would be interesting to look into pedantic [clippy group](https://rust-lang.github.io/rust-clippy/master/).

For each lint:
* add your GH handle to the lint so that two people are not working on the same lint,
* figure out if it's helpful and makes sense in our case,
* add it to the `lint` target in Makefile https://github.com/ChainSafe/forest/blob/d8eae69e556038af86e32a388e7e118b3bd3307d/Makefile#L73-L74
* fix the warnings if any
* tick the corresponding checkbox,

After they are all resolved, enable the entire pedantic group with perhaps some exceptions.

:warning: 
This is a low-priority issue; feel free to tackle a lint or two now and then if there's nothing more important to do.

Lint list (may differ in time & clippy version):
- [ ] `bool_to_int_with_if`
- [ ] `borrow_as_ptr`
- [ ] `case_sensitive_file_extension_comparisons `
- [ ] `cast_lossless`
- [ ] `cast_possible_truncation`
- [ ] `cast_possible_wrap`
- [ ] `cast_precision_loss`
- [ ] `cast_ptr_alignment`
- [ ] `cast_sign_loss`
- [ ] `checked_conversions`
- [ ] `cloned_instead_of_copied`
- [ ] `copy_iterator`
- [ ] `default_trait_access`
- [ ] `doc_link_with_quotes`
- [ ] `doc_markdown`
- [ ] `empty_enum`
- [ ] `enum_glob_use`
- [ ] `expl_impl_clone_on_copy`
- [ ] `explicit_deref_methods`
- [ ] `explicit_into_iter_loop`
- [ ] `explicit_iter_loop`
- [ ] `filter_map_next`
- [ ] `flat_map_option`
- [ ] `float_cmp`
- [ ] `fn_params_excessive_bools`
- [ ] `from_iter_instead_of_collect`
- [ ] ~~`if_not_else`~~
- [ ] `implicit_clone`
- [ ] `implicit_hasher`
- [ ] `inconsistent_struct_constructor`
- [ ] `index_refutable_slice`
- [ ] `inefficient_to_string`
- [ ] `inline_always`
- [ ] `invalid_upcast_comparisons`
- [ ] `items_after_statements`
- [ ] `iter_not_returning_iterator`
- [ ] `large_digit_groups`
- [ ] `large_stack_arrays`
[...]

PR #2431, opened 2023-01-13
Pedantic Clippy `if_not_else` implementation

jdjaustin:
## Summary of changes

- Add `-W clippy::if_not_else` to `lint` target in Makefile and changes to associated files to clear new clippy warnings.

Closes associated task in #2420 

Per the [Clippy Lints](https://rust-lang.github.io/rust-clippy/master/) documentation, "Negations reduce the readability of statements," and the code can easily be refactored to change these `if_not_else` statements to more readable `if_else` statements.

## Change checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,8 +70,8 @@ lint: license clean
 	cargo fmt --all --check
 	taplo fmt --check
 	taplo lint
-	cargo clippy --features slow_tests --all-targets -- -D warnings -W clippy::unused_async
-	cargo clippy --all-targets --no-default-features --features forest_deleg_cns,paritydb,instrumented_kernel -- -D warnings -W clippy::unused_async
+	cargo clippy --features slow_tests --all-targets -- -D warnings -W clippy::unused_async -W clippy::if_not_else
+	cargo clippy --all-targets --no-default-features --features forest_deleg_cns,paritydb,instrumented_kernel -- -D warnings -W clippy::unused_async -W clippy::if_not_else
 
 # Formats Rust and TOML files
 fmt:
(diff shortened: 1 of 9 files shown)


lemmih:
The `figure out if it's helpful and makes sense in our case` step is the hard part. Is `if_not_else` helpful and sensible in our case? If yes, add your arguments to the PR.

lemmih:
@jdjaustin This lint isn't enabled by default. Can you explain the reasons why it isn't enabled by default and why those reasons don't apply to Forest?

jdjaustin:
> @jdjaustin This lint isn't enabled by default. Can you explain the reasons why it isn't enabled by default and why those reasons don't apply to Forest?

The only thing I've found is that prior to October 2021, the `if_not_else` lint check was triggering false positives on `else if` statements, so perhaps that is why it isn't enabled by default. Is there some documentation somewhere that would help explain?

lemmih:
> > @jdjaustin This lint isn't enabled by default. Can you explain the reasons why it isn't enabled by default and why those reasons don't apply to Forest?
> 
> The only thing I've found is that prior to October 2021, the `if_not_else` lint check was triggering false positives on `else if` statements, so perhaps that is why it isn't enabled by default. Is there some documentation somewhere that would help explain?

There's no documentation to read. It's not easy but you just have to look at the lint and try to feel what kind of changes it entails. Sometimes you can feel that the lint is always an improvement for our code-base. Other times, there might be too many false positives (eg cases where the lint shouldn't be applied) or the improvement is so small that it's not worth the added complexity or maybe it's too difficult to predict all the ways the lint will affect the code-base.

As an example, we recently enabled `unused_async`. It's disabled by default for two major reasons: (1) It's unstable, (2) there will be false positives where we can't remove async even though clippy thinks it is unused. Hubert reasoned that using an unstable lint isn't a big deal for us. And he reasoned that we don't mind manually marking each false positive. Keeping our async surface area as small as possible was worth the cost of the lint.

In general, if you aren't sure that we want a new lint then we don't want it. I'll close this PR.

PR #2435, opened 2023-01-16
clippy::single_match_else pedantic check

tyshko5:
## Summary of changes

- 

Needed for readability. `match` statements idiomatically should be used when there are at least 3 options for an enum. In case there clauses are long it's not obvious how many branches the code has. 

Closes 

## Change checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,8 +70,8 @@ lint: license clean
 	cargo fmt --all --check
 	taplo fmt --check
 	taplo lint
-	cargo clippy --features slow_tests --all-targets -- -D warnings -W clippy::unused_async
-	cargo clippy --all-targets --no-default-features --features forest_deleg_cns,paritydb,instrumented_kernel -- -D warnings -W clippy::unused_async
+	cargo clippy --features slow_tests --all-targets -- -D warnings -W clippy::unused_async -W clippy::single_match_else
+	cargo clippy --all-targets --no-default-features --features forest_deleg_cns,paritydb,instrumented_kernel -- -D warnings -W clippy::unused_async -W clippy::single_match_else
 
 # Formats Rust and TOML files
 fmt:
--- a/blockchain/beacon/src/drand.rs
+++ b/blockchain/beacon/src/drand.rs
@@ -257,20 +257,19 @@ impl Beacon for DrandBeacon {
 
     async fn entry(&self, round: u64) -> Result<BeaconEntry, anyhow::Error> {
         let cached: Option<BeaconEntry> = self.local_cache.read().get(&round).cloned();
-        match cached {
-            Some(cached_entry) => Ok(cached_entry),
-            None => {
-                let url = format!("{}/public/{}", self.url, round);
-                let client = https_client();
-                let resp: BeaconEntryJson = client
-                    .get(url.try_into()?)
-                    .await?
-                    .into_body()
-                    .json()
-                    .await
-                    .map_err(|e| anyhow::anyhow!("{}", e))?;
-                Ok(BeaconEntry::new(resp.round, hex::decode(resp.signature)?))
-            }
+        if let Some(cached_entry) = cached {
+            Ok(cached_entry)
+        } else {
+            let url = format!("{}/public/{}", self.url, round);
+            let client = https_client();
+            let resp: BeaconEntryJson = client
+                .get(url.try_into()?)
+                .await?
+                .into_body()
+                .json()
+                .await
+                .map_err(|e| anyhow::anyhow!("{}", e))?;
+            Ok(BeaconEntry::new(resp.round, hex::decode(resp.signature)?))
         }
     }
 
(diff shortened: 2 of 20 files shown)


lemmih:
You need strong arguments for why `single_match_else` should be enabled. Add them to the PR description.

LesnyRumcajs:
Personally, I don't find this more readable, but if others prefer it this way, I'm okay with the change.

lemmih:
Your arguments:

1. Needed for readability.
2. `match` statements idiomatically should be used when there are at least 3 options for an enum.
3. In case there clauses are long it's not obvious how many branches the code has.

Counter-arguments:

1. I think matches are more readable than `if` statements.
2. I think this is just categorically incorrect. Matching on boolean values is better done with `if` but the same isn't true for other values with two options.
3. The indentation makes it obvious, no? In any case, I don't think chains of `if-else` are more obvious.

lemmih:
> > Your arguments:
> > 
> > 1. Needed for readability.
> > 2. `match` statements idiomatically should be used when there are at least 3 options for an enum.
> > 3. In case there clauses are long it's not obvious how many branches the code has.
> > 
> > Counter-arguments:
> > 
> > 1. I think matches are more readable than `if` statements.
> > 2. I think this is just categorically incorrect. Matching on boolean values is better done with `if` but the same isn't true for other values with two options.
> > 3. The indentation makes it obvious, no? In any case, I don't think chains of `if-else` are more obvious.
> 
> My argument is not readability in general, but readability in case of `Option`s and `Result`s which is what is fixed in the PR. I'm not sure I understand your second argument fully. I am not a proponent of `if-else` chains.

Whether `if` is better than `match` for `Option`s is controversial and I don't feel your arguments are overwhelmingly persuasive. As such, I'm closing this PR.

The extra clippy hints listed in the parent issue are all disabled by default because they aren't appropriate in all projects. The work is not in updating our code-base, the work is in figuring out which hints are appropriate for Forest. If you don't strongly feel a lint is appropriate (and can argue for your convictions) then the lint should not be enabled.

tyshko5:
> > > Your arguments:
> > > 
> > > 1. Needed for readability.
> > > 2. `match` statements idiomatically should be used when there are at least 3 options for an enum.
> > > 3. In case there clauses are long it's not obvious how many branches the code has.
> > > 
> > > Counter-arguments:
> > > 
> > > 1. I think matches are more readable than `if` statements.
> > > 2. I think this is just categorically incorrect. Matching on boolean values is better done with `if` but the same isn't true for other values with two options.
> > > 3. The indentation makes it obvious, no? In any case, I don't think chains of `if-else` are more obvious.
> > 
> > 
> > My argument is not readability in general, but readability in case of `Option`s and `Result`s which is what is fixed in the PR. I'm not sure I understand your second argument fully. I am not a proponent of `if-else` chains.
> 
> Whether `if` is better than `match` for `Option`s is controversial and I don't feel your arguments are overwhelmingly persuasive. As such, I'm closing this PR.
> 
> The extra clippy hints listed in the parent issue are all disabled by default because they aren't appropriate in all projects. The work is not in updating our code-base, the work is in figuring out which hints are appropriate for Forest. If you don't strongly feel a lint is appropriate (and can argue for your convictions) then the lint should not be enabled.

My argument was not that `if` is better than `match` in general, but only in specific circumstances this lint works. 

I indeed strongly feel this lint is not only appropriate, but should be mandatory for all Rust projects.

lemmih:
> > > > Your arguments:
> > > > 
> > > > 1. Needed for readability.
> > > > 2. `match` statements idiomatically should be used when there are at least 3 options for an enum.
> > > > 3. In case there clauses are long it's not obvious how many branches the code has.
> > > > 
> > > > Counter-arguments:
> > > > 
> > > > 1. I think matches are more readable than `if` statements.
> > > > 2. I think this is just categorically incorrect. Matching on boolean values is better done with `if` but the same isn't true for other values with two options.
> > > > 3. The indentation makes it obvious, no? In any case, I don't think chains of `if-else` are more obvious.
> > > 
> > > 
> > > My argument is not readability in general, but readability in case of `Option`s and `Result`s which is what is fixed in the PR. I'm not sure I understand your second argument fully. I am not a proponent of `if-else` chains.
> > 
> > 
> > Whether `if` is better than `match` for `Option`s is controversial and I don't feel your arguments are overwhelmingly persuasive. As such, I'm closing this PR.
> > The extra clippy hints listed in the parent issue are all disabled by default because they aren't appropriate in all projects. The work is not in updating our code-base, the work is in figuring out which hints are appropriate for Forest. If you don't strongly feel a lint is appropriate (and can argue for your convictions) then the lint should not be enabled.
> 
> My argument was not that `if` is better than `match` in general, but only in specific circumstances this lint works.
> 
> I indeed strongly feel this lint is not only appropriate, but should be mandatory for all Rust projects.

Your objection is noted but my decision stands.

PR #2436, opened 2023-01-16
fix unbounded tipset_tracker

LesnyRumcajs:
## Summary of changes

- 

Closes 

## Change checklist


blockchain/chain/src/store/tipset_tracker.rs
```diff
@@ -1,47 +1,55 @@
...
-        };
+        let mut map = self.entries.lock();
+        let cids = map.entry(header.epoch()).or_default();
+        if cids.contains(header.cid()) {
+            debug!("tried to add block to tipset tracker that was already there");
+            return;
+        }
+        self.check_multiple_blocks_from_same_miner(cids, header);
+        cids.push(*header.cid());
+        drop(map);
```
elmattic, line 43: I don't get why we can not use an implicit drop of the guard like before, using a block expression?

lemmih:
It would be nice to not hold the lock during DB reads.

blockchain/chain/src/store/tipset_tracker.rs
LesnyRumcajs: We can, but I found explicit to be more readable and easier to change; no indentations needed, and the same number of LoC. If you feel strongly about it, I'm okay with falling back to blocks. What do you think?

LesnyRumcajs:
> It would be nice to not hold the lock during DB reads.

It's a good point. I wanted to avoid extra allocation, but it's probably not worth the extra lock time.

blockchain/chain/src/store/tipset_tracker.rs
elmattic: I'm okay with both, just a preference style for scopes on my side.

elmattic (approved):
Very nice.

lemmih (approved):
Nicely implemented and tested. Soon we'll have a continuously running Forest node and we'll get empirical data on whether such changes significantly improve our memory usage.

PR #2439, opened 2023-01-17
chore: initial stab at an FVM shim

### LINKED ISSUE #2438: Create a compatibility layer for FVM2 and FVM3
**Issue summary**

Background:

The FVM crates (`fvm` and `fvm_shared`) are splitting into two: FVM2 and FVM3. FVM2 supports network versions 15 to 17 and FVM3 supports version 18. To support a wide range of network versions, we need to use both FVM2 and FVM3 at the same time. Unfortunately, all of the core data-types have been duplicated and have to be converted when moving data between FVM2 and FVM3.

For example, the `Address` datatype is not shared between FVM2 and FVM3. To solve this issue, the plan is to keep a Forest-specific `Address` type that can be converted to `FVM2::Address` and `FVM3::Address`. This allows us to write code that is version-agnostic.

There are a lot of types to hoist out but, fortunately, the work can be done mostly in parallel.

Hoisting out a type is usually done in two steps: First a wrapper is created in the new `forest_shim` crate with conversion methods for FVM2 and FVM3, then the rest of the Forest code-base is updated to use the shimmed type.

Feel free to take any one of these tasks. Check the checkbox once you begin to avoid duplicating work.

Tasks:

- [x] `Address`
  - [x] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [x] `Protocol`
  - [ ] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [x] `TokenAmount`
  - [ ] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [x] `ExitCode`
  - [x] Separate type that can convert to FVM2 and FVM3.
  - [x] Use it in Forest proper.
- [x] `Message`
  - [ ] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [x] `NetworkVersion`
  - [x] Separate type that can convert to FVM2 and FVM3.
  - [x] Use it in Forest proper.
- [X] `Signature`
  - [X] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [X] `SignatureType`
  - [X] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [x] `PoStProof`
  - [ ] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [X] `Randomness`
  - [X] Separate type that can convert to FVM2 and FVM3.
  - [X] Use it in Forest proper.
- [x] `RegisteredSealProof`
  - [ ] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [x] `RegisteredPoStProof`
  - [X] Separate type that can convert to FVM2 and FVM3.
  - [X] Use it in Forest proper.  
- [x] `SectorInfo`
  - [ ] Separate type that can convert to FVM2 and FVM3.
  - [ ] Use it in Forest proper.
- [x] `StateTree`
  - [x] Separate type that can convert to FVM2 and FVM3. Note: Converting between two state trees might not be easy. Consider switching the entire code-base to FVM3's `StateTree`.
  - [x] Use it in Forest proper.
- Keep as type synonym: `ChainEpoch` 
- Constants:
  - [ ] `BLOCK_GAS_LIMIT`
  - [ ] `BLOCKS_PER_EPOCH`
  - [ ] `METHOD_SEND`
  - [ ] `TICKET_RANDOMNESS_LOOKBACK`
  - [ ] `TOTAL_FILECOIN`
  - [ ] `IPLD_RAW`
  - [ ] `EPOCH_DURATION_SECONDS`
- Import num-traits directly instead of going through `fvm_shared::bigint`.

lemmih:
## Summary of changes

- Introduce the `forest_shim` crate as a compatibility layer between Forest and the FVM.
- Basic shim for `Address` (currently unused).
- Basic shim for `TokenAmount` (partly used).
- Shim for `NetworkVersion`

Doesn't close #2438 but it's a beginning.

## Change checklist


utils/forest_shim/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "forest_shim"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+#fvm.workspace = true
+#fvm3 = { package = "fvm", version = "3.0.0-alpha.19" }
+fvm_shared.workspace = true
```
hanabi1224, line 11: Maybe also move `fvm_shared` out of workspace manifect, correct me if I'm wrong, with this shim crate,  I believe `fvm_shared` should not be referenced by other crates directly

hanabi1224 (approved):

utils/forest_shim/Cargo.toml
lemmih: That's the plan. It'll have to be done incrementally as more as more types are moved out of `fvm_shared` and into `forest_shim`. Moving everything at once is a big mouthful.

utils/forest_shim/src/address.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019-2022 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use fvm_shared::address::Address as Address_v2;
+use fvm_shared3::address::Address as Address_v3;
+use std::ops::{Deref, DerefMut};
+
+pub struct Address(Address_v3);
```
LesnyRumcajs, line 8: Should we add the derives from the child? E.g. `#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)`.

LesnyRumcajs (commented):
Perhaps it would make sense to write some tests or just compiled documentation showing sample usage of this. It might be useful if we want to port it later to FVM repo.

utils/forest_shim/src/address.rs
lemmih: We should.

lemmih:
> Perhaps it would make sense to write some tests or just compiled documentation showing sample usage of this. It might be useful if we want to port it later to FVM repo.

You're right. I got struck on actually using this new `Address` is Forest. Everything that depends on `Address` have to be converted first. I'll convert something simpler and show how it is to be used in Forest-proper.

LesnyRumcajs (approved):

PR #2446, opened 2023-01-18
fix drand beacon constructor being stuck

LesnyRumcajs:
## Summary of changes

- Fix the `DrandBeacon` constructor being stuck on debug tests, introduced in https://github.com/ChainSafe/forest/pull/2427. I moved it explicitly to another thread with its own tokio runtime. The alternative was to mark big chunk of tests as `#[tokio::test(flavor = "multi_thread")]` which would be quite bug prone,
- moved regular `test-all` to `debug` mode to make sure overflow checks and debug assertions are executed. The test execution is more prolonged (30s -> 2min), but it's acceptable. On top of that, it should compile faster.
- moved `submodule tests` to non-release mode. It takes them 0.2s anyway.
- clippy for submodule tests

Closes 

## Change checklist


lemmih (approved):

tyshko5 (approved):

Makefile
```diff
@@ -70,7 +70,7 @@ lint: license clean
 	cargo fmt --all --check
 	taplo fmt --check
 	taplo lint
-	cargo clippy --features slow_tests --all-targets -- -D warnings -W clippy::unused_async -W clippy::redundant_else
```
hanabi1224, line 73: Can we move custom clippy settings to `.clippy.toml`?
https://doc.rust-lang.org/stable/clippy/configuration.html

hanabi1224 (approved):

creativcoder (approved):

Makefile
LesnyRumcajs: I don't think we can do it quickly. Only a subset of lints is exposed in this configuration which is marked as experimental anyway. We can revisit it in a separate PR if you think it's doable. Adding `unused_async = true` yields

```
error: error reading Clippy's configuration file `/home/rumcajs-work/prj/forest/.clippy.toml`: unknown field `unused_async`, expected one of
           allow-dbg-in-tests                   disallowed-names                         msrv
           allow-expect-in-tests                disallowed-types                         pass-by-value-size-limit
           allow-mixed-uninlined-format-args    doc-valid-idents                         single-char-binding-names-threshold
           allow-print-in-tests                 enable-raw-pointer-heuristic-for-send    standard-macro-braces
           allow-unwrap-in-tests                enforced-import-renames                  third-party
           allowed-scripts                      enum-variant-name-threshold              too-large-for-stack
           arithmetic-side-effects-allowed      enum-variant-size-threshold              too-many-arguments-threshold
           array-size-threshold                 ignore-interior-mutability               too-many-lines-threshold
           avoid-breaking-exported-api          large-error-threshold                    trivial-copy-size-limit
           await-holding-invalid-types          literal-representation-threshold         type-complexity-threshold
           blacklisted-names                    matches-for-let-else                     unreadable-literal-lint-fractions
           cargo-ignore-publish                 max-fn-params-bools                      upper-case-acronyms-aggressive
           cognitive-complexity-threshold       max-include-file-size                    vec-box-size-threshold
           cyclomatic-complexity-threshold      max-struct-bools                         verbose-bit-mask-threshold
           disallowed-macros                    max-suggested-slice-pattern-length       warn-on-all-wildcard-imports
           disallowed-methods                   max-trait-bounds                         
```

Issue #2441, opened 2023-01-17
Consider using AHasher for internal HashSet/HashMap


hanabi1224:
**Issue summary**

std HashSet / HashMap uses `hashbrown` with `SipHash` hasher as default, where `AHash` hasher has better performance per https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md#speed  

**Other information and links**

https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md#speed

PR #2454, opened 2023-01-19
feat: favor ahash collections over std ones when applicable

hanabi1224:
## Summary of changes

- favor ~~hashbrown~~ ahash collections over std ones when applicable
- as a side product, it caught the bug in https://github.com/ChainSafe/forest/pull/2453

Closes https://github.com/ChainSafe/forest/issues/2441

## Change checklist


lemmih:
Is this still relevant?

> Since Rust 1.36, this is now the HashMap implementation for the Rust standard library. However you may still want to use this crate instead since it works in environments without std, such as embedded systems and kernels.

hanabi1224:
> How about we use ahash::HashMap?

@lemmih that also works except that the type name from ahash is slightly different, https://docs.rs/ahash/0.8.2/ahash/struct.AHashMap.html
would u prefer `use ahash::AHashMap as HashMap` or using `AHashMap` in place?

lemmih:
Using `AHashMap` would also be fine by me.

```rust
use ahash::AHashMap;

let mut map: AHashMap<i32, i32> = AHashMap::new();
map.insert(12, 34);
```

hanabi1224:
@lemmih Fixed. Using `use ahash::{HashMap, HashMapExt};` approach to reduce num of changed lines

blockchain/chain_sync/src/tipset_syncer.rs
```diff
@@ -39,15 +22,31 @@ use forest_message::Message as MessageTrait;
 use forest_networks::Height;
 use forest_state_manager::Error as StateManagerError;
 use forest_state_manager::StateManager;
+use forest_utils::io::ProgressBar;
+use futures::stream::FuturesUnordered;
 use futures::Stream;
+use futures::StreamExt;
+use futures::TryFutureExt;
```
lemmih, line 29:
```suggestion
use futures::{Stream, StreamExt, TryFutureExt};
```

blockchain/message_pool/src/msgpool/msg_pool.rs
```diff
@@ -16,6 +16,9 @@ use crate::msgpool::{republish_pending_messages, select_messages_for_block};
 use crate::msgpool::{RBF_DENOM, RBF_NUM};
 use crate::provider::Provider;
 use crate::utils::get_base_fee_lower_bound;
+use ahash::HashMapExt;
+use ahash::HashSetExt;
+use ahash::{HashMap, HashSet};
```
lemmih, line 21:
```suggestion
use ahash::{HashMap, HashMapExt, HashSet, HashSetExt};
```

lemmih (approved):
Looks good!

LesnyRumcajs:
Do we see any measurable improvement in our case?

hanabi1224:
> Do we see any measurable improvement in our case?

@LesnyRumcajs No and I doubt it will lead to any measurable improvement since the internal hashmap/hashset we use are quite small

LesnyRumcajs:
> > Do we see any measurable improvement in our case?
> 
> @LesnyRumcajs No and I doubt it will lead to any measurable improvement since the internal hashmap/hashset we use are quite small

Is it worth the added dependency/mental overhead, then? I mean, I prefer to default to standard library collections unless there's a compelling reason to switch to something specialised.

hanabi1224:
> unless there's a compelling reason to switch to something specialised.

@LesnyRumcajs the only motivation behind this is `AHash` being ~10x faster than `SipHash` as shown in the benchmark, if that's strong enough we could revisit this after getting performance measurements (but I doubt if we could see any non-trival gain, so it would be low priority)

LesnyRumcajs (approved):
Not a huge fan of this change, but I'm okay with it.

hanabi1224:
@LesnyRumcajs I think your argument is reasonable, I'm kinda neutral about whether or not we should merge this.  @lemmih  what do you think?

lemmih:
> @LesnyRumcajs I think your argument is reasonable, I'm kinda neutral about whether or not we should merge this. @lemmih what do you think?

Might as well merge it. PR is done and we were already using `AHashMap` in one place before. It might not matter for performance but I do like the consistency of always using either `ahash::HashMap` or the standard `HashMap`.

LesnyRumcajs:
If we aim for not-using standard `HashMap` at all, should we create a `.clippy.toml` file with it disallowed like [here](https://github.com/amethyst/specs/blob/81073f340425803c99e3292fa14ceb00245bfea5/clippy.toml#L1)?

lemmih:
> If we aim for not-using standard `HashMap` at all, should we create a `.clippy.toml` file with it disallowed like [here](https://github.com/amethyst/specs/blob/81073f340425803c99e3292fa14ceb00245bfea5/clippy.toml#L1)?

Will that work? We're still using `std::collections::HashMap`. We just instantiate it with a different hasher. Ie: `std::collections::HashMap<Key, Value, ahash::RandomState>`.

hanabi1224:
I just tried `.clippy.toml`, it does not complain for `ahash::HashMap` but we do use `std::collections::HashMap` in some places where upstream crate API requires it

LesnyRumcajs:
@hanabi1224 Thanks for checking it out. Let's just merge then.

hanabi1224:
> @hanabi1224 Thanks for checking it out. Let's just merge then.

@LesnyRumcajs Actually I can explicitly allow those 2 instances and still add `.clippy.toml`, just pushed the change

PR #2459, opened 2023-01-20
chore: shim `fvm_shared::bigint`

lemmih:
## Summary of changes

- [x] Avoid using `fvm_shared::bigint` when possible.
- [x] Move `bigint_ser` and `biguint_ser` to `forest_shim`. It can use the modules from `fvm_shared_v3`.

Closes 

## Change checklist


lemmih:
@elmattic Feel like taking this PR?

elmattic:
Some questions:
- Should `weight: BigInt` in `BlockHeader` use `BigInt` from `forest_shim`?
- I didn't touch to `BigInt` in `vm_actor/interface`

LesnyRumcajs:
> Some questions:
> 
> * Should `weight: BigInt` in `BlockHeader` use `BigInt` from `forest_shim`?
I believe so.

> * I didn't touch to `BigInt` in `vm_actor/interface`
Why?

elmattic:
> > Some questions:
> > 
> > * Should `weight: BigInt` in `BlockHeader` use `BigInt` from `forest_shim`?
> 
> I believe so.

It's a bit hard because `forest_shim::bigint::BigInt` lacks of `FromStr`. In doubt I will keep the code like this.

> 
> > * I didn't touch to `BigInt` in `vm_actor/interface`
> 
> Why?

No reason. Done.

blockchain/chain/src/store/base_fee.rs
```diff
@@ -90,7 +90,7 @@ where
     // Compute next base fee based on the current gas limit and parent base fee.
     let parent_base_fee = ts.blocks()[0].parent_base_fee();
     Ok(compute_next_base_fee(
-        parent_base_fee,
+        &parent_base_fee.clone(),
```
creativcoder, line 93: I think we don't need `&` `.clone()`, just parent_base_fee

creativcoder (approved):
Looking good!

blockchain/chain/src/store/base_fee.rs
elmattic: Good catch! Done.

tyshko5 (approved):

LesnyRumcajs (approved):
@elmattic How about adding a rule that would disallow using FVM bigint directly? This can be done via clippy configuration file [here](https://github.com/ChainSafe/forest/blob/main/.clippy.toml).

elmattic:
> @elmattic How about adding a rule that would disallow using FVM bigint directly? This can be done via clippy configuration file [here](https://github.com/ChainSafe/forest/blob/main/.clippy.toml).

Won't there be issues with existing legit `fvm_shared` imports?

LesnyRumcajs:
> > @elmattic How about adding a rule that would disallow using FVM bigint directly? This can be done via clippy configuration file [here](https://github.com/ChainSafe/forest/blob/main/.clippy.toml).
> 
> Won't there be issues with existing legit `fvm_shared` imports?

You can add `#[allow(clippy::disallowed_types)]` to the shim unit. To my understanding, that is the only place where `fvm_shared` `BigInt` should be directly used.

PR #2473, opened 2023-01-25
ActorState FVM 2-3 shim

LesnyRumcajs:
## Summary of changes

- initially, it was supposed to use `StateTree` from FVM3; unfortunately, it is impossible without proper state migrations, which we lack at this point. I shimmed `ActorState` and re-exported FVM2 StateTree. It should make the transition easier in the upcoming network update.

Part of https://github.com/ChainSafe/forest/issues/2438

## Change checklist


blockchain/message_pool/src/msgpool/test_provider.rs
```diff
@@ -148,8 +148,14 @@ impl Provider for TestApi {
             }
             sequence += 1;
         }
-        let actor = ActorState::new(Cid::default(), Cid::default(), balance, sequence);
-        Ok(actor)
+        let actor = <ActorState as forest_shim::state_tree::Derive>::FVM::new(
```
creativcoder, line 151: `Derive` sounds conflicting with macro derive thing, maybe we can name the trait as: 
`trait As {
    type Type;
}`

LesnyRumcajs: This would look conflicting with both `as` and `type` keywords. I can change it to `Inner`; I have difficulty creating a meaningful name for this dummy trait. Does it sound good to you, @creativcoder?

creativcoder: Right :) Yeah, Inner sounds fine

hanabi1224 (approved):

tyshko5 (approved):

Issue #2476, opened 2023-01-27
Multi-platform Docker image for Forest


LesnyRumcajs:
**Issue summary**

Currently, we offer a Forest image for `linux/amd64`, which should work on Mac M1, provided certain virtualization prerequisites are installed and configured. This may get a bit messy, and the out-of-the-box experience for the user should be better than that.

This can be done easily with QEMU; unfortunately, the cross-compilation duration using this utility is way too slow - around 30-40 minutes using the best buildjet worker. A native build (as suggested by @elmattic, thanks) on a regular machine takes only 10 minutes.

It would be fantastic to do two native builds, one for `amd64` and another for `arm64` and end up with a single, multi-platform image. The alternative is to have multiple tags and achieve multi-platform capacity this way, e.g. `latest-arm`, `v0.6.0-arm`. Easy to implement, but not the best experience we can offer.

**Other information and links**

A failed attempt can be found here https://github.com/ChainSafe/forest/pull/2475. In the end, I managed to produce a working native image for M1, but it's far from a perfect solution.

LesnyRumcajs:
Another `trick` is to reuse built binaries from other workflows to avoid re-compilation. The YAML files might get complex, though.

LesnyRumcajs:
For reference https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

PR #2483, opened 2023-01-30
ci: add more juicy github action runners

q9f:
## Summary of changes
uses build jet external runners. should not make any difference (except for speed).

n/a

n/a

## Change checklist


LesnyRumcajs:
@q9f buildjets are billable whereas GH workers for public repositories are free. Unless a job is critical from duration point of view (e.g., takes 2h on regular GH workers) , there is no real point in doing so, no?

hanabi1224:
One downside of buildjet to note is that, it's very slow in downloading the build cache, and timeout from time to time

q9f:
It's an offer. If you are fine with GA, I will move on silently 😶‍🌫️

hanabi1224:
I think the code coverage job can definitely benefit from buildjet, and we can even enable it for PR(s). We recently moved it to GH hosted runners and it takes ~50min for each run

LesnyRumcajs:
> I think the code coverage job can definitely benefit from buildjet, and we can even enable it for PR(s). We recently moved it to GH hosted runners to save cost and it takes ~50min for each run

I don't think it is helpful to enable it for PRs. After a recent change, it's doing roughly the same stuff as the test step, and we are not tracking coverage changes. For a regular coverage step, running it for 50 minutes is fine because it's already on `main`, and it's not blocking anyone.

It would be helpful to use a juicy machine for https://github.com/ChainSafe/forest/blob/main/.github/workflows/docker.yml until https://github.com/ChainSafe/forest/issues/2476 is resolved. We could then easily support multi-platform docker images without many tricks. Then again, not many folks need a native image (for now), so it might not be worth it.

q9f:
ok let me know which of the jobs will definitely not benefit from buildjet and I will revert them to GA

LesnyRumcajs:
@q9f Looking at other runs, if we add only calibnet sync checks and lint checks then we will sit at ~10m per PR. That is to say, other jobs on GA can manage under this limit and the coverage is completely async so it's not an issue.

.github/workflows/autoupdate.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   autoupdate:
     name: autoupdate
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
q9f, line 10: latest is now 22.04

https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/

LesnyRumcajs (approved):
Thank you!

hanabi1224 (approved):

PR #2491, opened 2023-02-01
feat: ensure no multiple chain export jobs in parallel

hanabi1224:
## Summary of changes

- [chain export API] Return error to client when there is already export task running, as forest daemon apparently cannot afford running multiple mainnet export tasks in parallel on a affordable machine

Sample error:
```
➜  calibnet forest-cli snapshot export
Error: {"code":-100,"message":"Another chain export job is still in progress"}
```

Closes 

## Change checklist


node/rpc/src/chain_api.rs
```diff
@@ -55,6 +58,18 @@ where
...
     B: Beacon,
 {
+    lazy_static::lazy_static! {
+        static ref LOCK: Mutex<()> = Mutex::new(());
+    }
+
+    let _locked = LOCK.try_lock();
+    if _locked.is_err() {
+        return Err(JsonRpcError::Provided {
+            code: -100,
```
LesnyRumcajs, line 68: It would be nice to use standard error codes
```suggestion
            code: http::StatusCode::SERVICE_UNAVAILABLE,
```

LesnyRumcajs (approved):

node/rpc/src/chain_api.rs
hanabi1224: Fixed

node/rpc/src/chain_api.rs
```diff
@@ -55,6 +58,18 @@ where
     DB: Blockstore + Store + Clone + Send + Sync + 'static,
     B: Beacon,
 {
+    lazy_static::lazy_static! {
```
elmattic, line 61: What is the purpose of the `lazy_static` here?

elmattic: Ah I see.

elmattic (approved):

PR #2508, opened 2023-02-06
feat: show snapshot size in prompt

### LINKED ISSUE #2501: Display snapshot size to the user during download snapshot prompt
**Issue summary**

Currently, the user gets this message when the DB is empty.
```
✔ Forest needs a snapshot to sync with the network. Would you like to download one now?
```

Then the user gets hit with 110G out of nowhere. It'd be nice to provide a piece of helpful information along the lines of the following:
```
✔ Forest needs a snapshot to sync with the network. Would you like to download one now? Required disk space 115GB · no
```

We could be more helpful by providing an estimate on the snapshot + DB size, given that the user would need ~230G upfront.

Thanks, @stonecharioteer, for the feedback!

stonecharioteer:
* show human-readable snapshot size when user is prompted to download the snapshot.

## Summary of changes

- Added `num-bigint` to `daemon`'s dependencies so that I can reuse the `to_size_string` util.
- Added `snapshot_fetch_size`, which just copies over the logic from the other parts of the file, have not DRYed this yet, need to check with @LesnyRumcajs.

Closes #2501

## Change checklist

### DIFF
--- a/forest/daemon/src/daemon.rs
+++ b/forest/daemon/src/daemon.rs
@@ -12,8 +12,8 @@ use forest_chain_sync::{consensus::SyncGossipSubmitter, ChainMuxer};
 use forest_cli_shared::{
     chain_path,
     cli::{
-        default_snapshot_dir, is_aria2_installed, snapshot_fetch, Client, Config,
-        FOREST_VERSION_STRING,
+        default_snapshot_dir, is_aria2_installed, snapshot_fetch, snapshot_fetch_size,
+        to_size_string, Client, Config, FOREST_VERSION_STRING,
     },
 };
 use forest_db::{
@@ -155,7 +155,7 @@ pub(super) async fn start(config: Config, detached: bool) -> anyhow::Result<Db> 
     let epoch = chain_store.heaviest_tipset().epoch();
     let nv = config.chain.network_version(epoch);
     let should_fetch_snapshot = if nv < NetworkVersion::V16 {
-        prompt_snapshot_or_die(&config)?
+        prompt_snapshot_or_die(&config).await?
     } else {
         false
     };
@@ -390,15 +390,17 @@ async fn maybe_fetch_snapshot(
 
 /// Last resort in case a snapshot is needed. If it is not to be downloaded,
 /// this method fails and exits the process.
-fn prompt_snapshot_or_die(config: &Config) -> anyhow::Result<bool> {
+async fn prompt_snapshot_or_die(config: &Config) -> anyhow::Result<bool> {
     if config.client.snapshot_path.is_some() {
         return Ok(false);
     }
     let should_download = if !config.client.auto_download_snapshot && atty::is(atty::Stream::Stdin)
     {
+        let required_size: u64 = snapshot_fetch_size(config).await?;
+        let required_size = to_size_string(&required_size.into())?;
         Confirm::with_theme(&ColorfulTheme::default())
                 .with_prompt(
-                    "Forest needs a snapshot to sync with the network. Would you like to download one now?",
+                    format!("Forest needs a snapshot to sync with the network. Would you like to download one now? Required disk space {required_size}."),
                 )
                 .default(false)
                 .interact()
--- a/forest/shared/src/cli/snapshot_fetch.rs
+++ b/forest/shared/src/cli/snapshot_fetch.rs
@@ -118,6 +118,37 @@ pub fn is_car_or_tmp(path: &Path) -> bool {
     ext == "car" || ext == "tmp" || ext == "aria2"
 }
 
+/// gets the size of a snapshot from filecoin.
+pub async fn snapshot_fetch_size(config: &Config) -> anyhow::Result<u64> {
+    let service_url = match config.chain.name.to_lowercase().as_ref() {
+        "mainnet" => config.snapshot_fetch.filecoin.mainnet.clone(),
+        "calibnet" => config.snapshot_fetch.filecoin.calibnet.clone(),
+        _ => bail!("Fetch not supported for chain {}", config.chain.name,),
+    };
+    let client = https_client();
+
+    let snapshot_url = {
+        let head_response = client
+            .request(hyper::Request::head(service_url.as_str()).body("".into())?)
+            .await?;
+
+        // Use the redirect if available.
+        match head_response.headers().get("location") {
+            Some(url) => url.to_str()?.try_into()?,
+            None => service_url,
+        }
+    };
+
+    let snapshot_response = client.get(snapshot_url.as_str().try_into()?).await?;
+    let total_size = snapshot_response
+        .headers()
+        .get("content-length")
+        .and_then(|ct_len| ct_len.to_str().ok())
+        .and_then(|ct_len| ct_len.parse::<u64>().ok())
+        .ok_or_else(|| anyhow::anyhow!("Couldn't retrieve content length"))?;
+    anyhow::Ok(total_size)
+}
+
 /// Fetches snapshot from a trusted location and saves it to the given
 /// directory. Chain is inferred from configuration.
 pub async fn snapshot_fetch(
@@ -195,6 +226,7 @@ async fn snapshot_fetch_forest(
     let url = snapshot_spaces_url.join(path)?.join(filename)?;
 
     let snapshot_response = client.get(url.as_str().try_into()?).await?;
+
     if use_aria2 {
         download_snapshot_and_validate_checksum_with_aria2(client, url, &snapshot_path).await?
     } else {


forest/shared/src/cli/snapshot_fetch.rs
```diff
@@ -245,6 +325,13 @@ async fn snapshot_fetch_filecoin(
     create_dir_all(snapshot_out_dir).await?;
     let snapshot_name = normalize_filecoin_snapshot_name(&config.chain.name, &filename)?;
     let snapshot_path = snapshot_out_dir.join(&snapshot_name);
+    let total_size = snapshot_response
+        .headers()
+        .get("content-length")
+        .and_then(|ct_len| ct_len.to_str().ok())
+        .and_then(|ct_len| ct_len.parse::<u64>().ok())
+        .ok_or_else(|| anyhow::anyhow!("Couldn't retrieve content length"))?;
```
creativcoder, line 333: open question: not having a `content-length` header is not a reason to bail out i guess? A server may not provide content length but can still can send bytes using streaming responses.

forest/shared/src/cli/snapshot_fetch.rs
```diff
@@ -245,6 +325,13 @@ async fn snapshot_fetch_filecoin(
...
+    log::warn!("Total Download Size: {}", total_size);
```
creativcoder, line 334: Not a warn, but it looks like an `log::info` to me

forest/shared/src/cli/snapshot_fetch.rs
```diff
@@ -116,6 +116,85 @@ pub fn is_car_or_tmp(path: &Path) -> bool {
...
+                "calibnet" => &config.snapshot_fetch.forest.calibnet,
+                _ => bail!("Fetch not supported for chain {}", config.chain.name,),
+            };
+            let name = &snapshot_fetch_config.bucket_name;
+            let region = &snapshot_fetch_config.region;
+            let bucket = Bucket::new_public(name, region.parse()?)?;
+
+            // Grab contents of the bucket
+            let bucket_contents =
+                bucket.list(snapshot_fetch_config.path.clone(), Some("/".to_string()))?;
```
creativcoder, line 147: Do we really need the delimiter as second param?

forest/shared/src/cli/snapshot_fetch.rs
stonecharioteer: I need to remove that line, I only put that there for my own debugging. Thanks for spotting that.

forest/shared/src/cli/snapshot_fetch.rs
stonecharioteer: I just copied over the code from the other sections 😆

forest/shared/src/cli/snapshot_fetch.rs
stonecharioteer: True, but in this particular exercise, the content length is *required*, since we are not downloading the file (yet), but advertising the size.

stonecharioteer:
@LesnyRumcajs let me know if you think I should abstract away the logic that I've replicated from the other sections. I didn't want to touch the rest of the code and maybe break something just to DRY.

LesnyRumcajs:
@stonecharioteer We can simplify things here. The snapshot size should be roughly the same, regardless of the provider. I'd recommend following only the Filecoin provider path when grabbing the expected size of the snapshot. What do you think about it?

stonecharioteer:
> @stonecharioteer We can simplify things here. The snapshot size should be roughly the same, regardless of the provider. I'd recommend following only the Filecoin provider path when grabbing the expected size of the snapshot. What do you think about it?

I can work with that if you think it's approximate. When you say Filecoin provider, do you mean mainnet?

LesnyRumcajs:
> I can work with that if you think it's approximate. When you say Filecoin provider, do you mean mainnet?

No, I mean this `match`. https://github.com/ChainSafe/forest/pull/2508/files#diff-4ad267ba48229929665a2b24ed8b5b34cca74cdb97fad991980885c0d72b5942R132

We can effectively eliminate the `SnapshotServer::Forest` arm and code related to differentiating between those two. Does it make sense?

stonecharioteer:
@LesnyRumcajs Aah. I thought that the mainnet/calibnet were different in sizes. Wasn't that why you recommended the `calibnet`?

I'm making that change.

LesnyRumcajs:
> @LesnyRumcajs Aah. I thought that the mainnet/calibnet were different in sizes. Wasn't that why you recommended the `calibnet`?
> 
> I'm making that change.

Not sure I get it. They are different sizes.

stonecharioteer:
> The snapshot size should be roughly the same, regardless of the provider. I'd recommend following only the Filecoin provider path when grabbing the expected size of the snapshot.

😅 Now I'm confused. Could you check the changes I've made? This only gets the Filecoin provider path when checking for the size of the snapshot.

forest/daemon/src/daemon.rs
```diff
@@ -418,15 +420,18 @@ async fn maybe_fetch_snapshot(
...
     }
     let should_download = if !config.client.auto_download_snapshot && atty::is(atty::Stream::Stdin)
     {
+        // TODO: get the size of the snapshot.
+        let required_size: u64 = snapshot_fetch_size(&config).await.unwrap();
+        let required_size = to_size_string(&required_size.to_bigint().unwrap()).unwrap();
         Confirm::with_theme(&ColorfulTheme::default())
                 .with_prompt(
-                    "Forest needs a snapshot to sync with the network. Would you like to download one now?",
+                    format!("Forest needs a snapshot to sync with the network. Would you like to download one now? Required disk space {}.", required_size),
```
LesnyRumcajs, line 434:
```suggestion
                    format!("Forest needs a snapshot to sync with the network. Would you like to download one now? Required disk space {required_size}."),
```

forest/daemon/src/daemon.rs
```diff
@@ -418,15 +420,18 @@ async fn maybe_fetch_snapshot(
...
 /// Last resort in case a snapshot is needed. If it is not to be downloaded, this method fails and
 /// exits the process.
-fn prompt_snapshot_or_die(config: &Config) -> anyhow::Result<bool> {
+async fn prompt_snapshot_or_die(config: &Config) -> anyhow::Result<bool> {
     if config.client.snapshot_path.is_some() {
         return Ok(false);
     }
     let should_download = if !config.client.auto_download_snapshot && atty::is(atty::Stream::Stdin)
     {
+        // TODO: get the size of the snapshot.
```
LesnyRumcajs, line 429:
```suggestion
```

forest/shared/src/cli/snapshot_fetch.rs
```diff
@@ -193,6 +224,7 @@ async fn snapshot_fetch_forest(
     let url = snapshot_spaces_url.join(path)?.join(filename)?;
 
     let snapshot_response = client.get(url.as_str().try_into()?).await?;
+
```
LesnyRumcajs, line 227:
```suggestion
```

forest/daemon/src/daemon.rs
```diff
@@ -418,15 +420,18 @@ async fn maybe_fetch_snapshot(
...
+        let required_size: u64 = snapshot_fetch_size(&config).await.unwrap();
+        let required_size = to_size_string(&required_size.to_bigint().unwrap()).unwrap();
```
LesnyRumcajs, line 431: Let's not use `unwrap`s but propagate the error.

LesnyRumcajs (changes requested):
Some minor tweaks, but in general, it looks great to me! I've just tested it, and it's definitely a better user experience.

forest/daemon/src/daemon.rs
stonecharioteer: Done

forest/shared/src/cli/snapshot_fetch.rs
stonecharioteer: Done

forest/daemon/src/daemon.rs
stonecharioteer: Done

stonecharioteer:
Let me know if this is good to go, and I'll squash and rebase.

forest/daemon/src/daemon.rs
LesnyRumcajs: I meant all of the `unwrap`s. :)

lemmih:
> Let me know if this is good to go, and I'll squash and rebase.

Looks like there are still a few `unwrap`s in the code and one or two open suggestions.

stonecharioteer:
Ah I missed that. Let me do that tomorrow morning.

stonecharioteer:
@LesnyRumcajs Made the changes, take a look?

forest/daemon/src/daemon.rs
```diff
@@ -31,6 +32,7 @@ use forest_utils::io::write_to_file;
 use futures::{select, FutureExt};
 use fvm_ipld_blockstore::Blockstore;
 use log::{debug, error, info, warn};
+use num_bigint::{BigInt, ToBigInt};
```
LesnyRumcajs, line 35:
```suggestion
```

forest/daemon/src/daemon.rs
```diff
@@ -418,15 +420,21 @@ async fn maybe_fetch_snapshot(
...
         return Ok(false);
     }
     let should_download = if !config.client.auto_download_snapshot && atty::is(atty::Stream::Stdin)
     {
+        let required_size: u64 = snapshot_fetch_size(&config).await?;
+        let required_size: BigInt = required_size
+            .to_bigint()
+            .context("unable to get the required snapshot size.")?;
+        //.ok_or(Err("Unable to convert snapshot size to bigint."))?;
+        let required_size = to_size_string(&required_size)?;
```
LesnyRumcajs, line 434:
```suggestion
        let required_size = snapshot_fetch_size(config).await?;
        let required_size = to_size_string(&required_size.into())?;
```

LesnyRumcajs (changes requested):
Some cleanup and tweaks, and we should be good!

forest/daemon/src/daemon.rs
stonecharioteer: Done and Done!

forest/daemon/src/daemon.rs
```diff
@@ -418,15 +419,17 @@ async fn maybe_fetch_snapshot(
...
 /// Last resort in case a snapshot is needed. If it is not to be downloaded, this method fails and
 /// exits the process.
-fn prompt_snapshot_or_die(config: &Config) -> anyhow::Result<bool> {
+async fn prompt_snapshot_or_die(config: &Config) -> anyhow::Result<bool> {
     if config.client.snapshot_path.is_some() {
         return Ok(false);
     }
     let should_download = if !config.client.auto_download_snapshot && atty::is(atty::Stream::Stdin)
     {
+        let required_size: u64 = snapshot_fetch_size(&config).await?;
```
LesnyRumcajs, line 428:
```suggestion
        let required_size: u64 = snapshot_fetch_size(config).await?;
```

LesnyRumcajs (changes requested):

LesnyRumcajs:
Also, please run `make lint` before pushing the code.

stonecharioteer:
Spoke too soon! Thanks @LesnyRumcajs

forest/daemon/Cargo.toml
```diff
@@ -46,6 +46,7 @@ fvm_ipld_blockstore.workspace = true
 fvm_shared = { workspace = true, default-features = false }
 lazy_static.workspace = true
 log.workspace = true
+num-bigint = "0.4"
```
LesnyRumcajs, line 49:
```suggestion
```

LesnyRumcajs (changes requested):

forest/daemon/Cargo.toml
stonecharioteer: :D Was already on it.

LesnyRumcajs (approved):
LGTM. Thanks a lot, @stonecharioteer!

On a side note, please don't squash the commits once the PR is in the review phase; it's hard to keep track of the individual changes.

lemmih (approved):
Nice!

Issue #2499, opened 2023-02-02
Docker no such device error when running in non-interactive mode


LesnyRumcajs:
**Describe the bug**

When running the container in non-interactive mode AND without the `--encrypt-keystore false` option we get an error like so:

```
❯ docker run  --init --rm ghcr.io/chainsafe/forest:latest  --chain calibnet
2023-02-02T09:44:35.910039Z  INFO forest: Using default calibnet config
2023-02-02T09:44:35.910741Z  INFO forest::daemon: Starting Forest daemon, version 0.4.1+git.a708b83
2023-02-02T09:44:35.910764Z  INFO forest_libp2p::service: Networking keystore not found!
2023-02-02T09:44:35.910935Z  INFO forest_utils::io: Permissions set to 0600 on File { fd: 8, path: "/home/forest/.local/share/forest/libp2p/keypair", read: false, write: true }
2023-02-02T09:44:35.911093Z  INFO forest::daemon: PeerId: 12D3KooWCJ3WDMsvajAuCB2cLzhHn2T9RdoNnJ9eSBE5hhwJriFK
Enter the keystore passphrase: Error: No such device or address (os error 6)
```

This does not occur when running the container in interactive mode OR when issuing the `--encrypt-keystore false` flag. It's probably a permission fiasco.

**Expected behaviour**

Forest should not fail.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Fedora 37
 - Branch/commit 04130d53f7c1274d02391cd711f2d71f8a32d3bd

**Other information and links**

PR #2514, opened 2023-02-06
fix no such device error when non interactive mode

LesnyRumcajs:
## Summary of changes

- added an environmental variable for specifying encryption passphrase for Keystore,
- fixed a case where running Forest in headless mode would cause a cryptic `no such device` error. Now it will enforce setting the passphrase via the environmental variable unless the encryption is explicitly disabled.

Closes https://github.com/ChainSafe/forest/issues/2499

## Change checklist

### DIFF
--- a/forest/daemon/src/daemon.rs
+++ b/forest/daemon/src/daemon.rs
@@ -89,7 +89,7 @@ pub(super) async fn start(config: Config, detached: bool) -> anyhow::Result<Db> 
     // from.
     info!("PeerId: {}", PeerId::from(net_keypair.public()));
 
-    let mut ks = if config.client.encrypt_keystore {
+    let mut ks = if config.client.encrypt_keystore && atty::is(atty::Stream::Stdin) {
         loop {
             print!("Enter the keystore passphrase: ");
             std::io::stdout().flush()?;
--- a/forest/daemon/src/daemon.rs
+++ b/forest/daemon/src/daemon.rs
@@ -12,8 +12,8 @@ use forest_chain_sync::{consensus::SyncGossipSubmitter, ChainMuxer};
 use forest_cli_shared::{
     chain_path,
     cli::{
-        cli_error_and_die, default_snapshot_dir, is_aria2_installed, snapshot_fetch, Client,
-        Config, FOREST_VERSION_STRING,
+        default_snapshot_dir, is_aria2_installed, snapshot_fetch, Client, Config,
+        FOREST_VERSION_STRING,
     },
 };
 use forest_db::{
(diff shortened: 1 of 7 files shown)


elmattic (approved):

forest/daemon/src/daemon.rs
```diff
@@ -91,7 +91,7 @@ pub(super) async fn start(config: Config, detached: bool) -> anyhow::Result<Db>
     // Useful if others want to use this node to bootstrap from.
     info!("PeerId: {}", PeerId::from(net_keypair.public()));
 
-    let mut ks = if config.client.encrypt_keystore {
+    let mut ks = if config.client.encrypt_keystore && atty::is(atty::Stream::Stdin) {
```
hanabi1224, line 94: Should `encrypt_keystore` be used as default mode once forest is released? In this case, probably it's better to ask users to set an environment variable `FOREST_KEYSTORE_PHRASE` in noninteractive mode

hanabi1224: I don't think we should fallback to non encrypted keystore in any case cuz IMHO it's a key security feature that cannot be compromised

LesnyRumcajs: This shouldn't be a big deal at the moment. We would like to move the keys into a separate process as outlined in the architecture proposal. 

That being said, it's a long journey to get there, and, as you nicely caught it, in the non-interactive mode, it is impossible to set the phrase ourselves. I'll change it to accommodate this. Thanks.

hanabi1224 (approved):

forest/daemon/src/daemon.rs
elmattic: So what's the conclusion? Is this PR on hold now?

LesnyRumcajs: We can address the bug in this PR by merging this and addressing the encrypted-keystore feature for non-interactive users in a separate issue and a PR. Does it make sense to you @hanabi1224 @elmattic ?

LesnyRumcajs: Alternatively, let me know what you prefer. The result will be the same. :)

forest/daemon/tests/keystore_tests.rs
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::path::PathBuf;
+
+use anyhow::Result;
+use assert_cmd::Command;
+use tempfile::TempDir;
+
+// https://github.com/ChainSafe/forest/issues/2499
+#[test]
```
LesnyRumcajs, line 11: More cases should be implemented once https://github.com/ChainSafe/forest/issues/2528 is done.

forest/daemon/src/daemon.rs
```diff
@@ -488,6 +455,62 @@ fn get_actual_chain_name(internal_network_name: &str) -> &str {
     }
 }
 
+fn create_keystore(config: &Config) -> anyhow::Result<KeyStore> {
+    let passphrase = std::env::var(FOREST_KEYSTORE_PHRASE_ENV);
+    let is_interactive = atty::is(atty::Stream::Stdin);
+
+    // encrypted keystore, headless
+    if config.client.encrypt_keystore && passphrase.is_err() && !is_interactive {
+        cli_error_and_die(format!("Passphrase for the keystore was not provided and the encryption was not explicitly disabled. Please set the {FOREST_KEYSTORE_PHRASE_ENV} environmental variable and re-run the command"), 1);
```
elmattic, line 464: What about using `anyhow::bail` macro here?

Issue #2513, opened 2023-02-06
Add caching for Dockerfile


LesnyRumcajs:
**Issue summary**

At the moment, every time the image is built, it gets rebuilt from scratch. It is far from ideal as it takes 5 minutes on a pretty solid machine. We don't want to rebuild the image from scratch. 

Use something along the lines of
```
RUN --mount=type=cache,target=/usr/local/cargo/registry \
    --mount=type=cache,target=/home/root/app/target \
    cargo build --release
```

**Other information and links**

PR #2516, opened 2023-02-06
dockerfile caching + tweaks

LesnyRumcajs:
## Summary of changes

- add caching to the compilation layer - this allows seamlessly building the image and re-use the artefacts from the previous build. It is a significant improvement for folks building the image regularly on the local machine; previously, it would build everything from scratch (5 minutes for every minor change on my machine). This will reduce the CI time if on the same runner.
- minor, random tweaks to make the Dockerfile more ready for production environments

Closes https://github.com/ChainSafe/forest/issues/2513

## Change checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,19 +20,21 @@
 FROM buildpack-deps:bullseye AS build-env
 
 # Install dependencies
-RUN apt-get update && apt-get install --no-install-recommends -y build-essential clang ocl-icd-opencl-dev protobuf-compiler cmake ca-certificates curl
-RUN update-ca-certificates
+RUN apt-get update && apt-get install --no-install-recommends -y build-essential clang ocl-icd-opencl-dev protobuf-compiler cmake curl
 
-# Install rustup
-# https://rustup.rs/
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 
-WORKDIR /usr/src/forest
+WORKDIR /forest
 COPY . .
 
-# Install Forest
-RUN make install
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -20,19 +20,21 @@
...
 ENV PATH="/root/.cargo/bin:${PATH}"
 
-WORKDIR /usr/src/forest
+WORKDIR /forest
 COPY . .
 
-# Install Forest
-RUN make install
+# Install Forest. Move it out of the cache for the prod image.
+RUN --mount=type=cache,sharing=private,target=/root/.cargo/registry \
```
hanabi1224, line 32: How about using `sccache`, so that we can use it for CI as well

LesnyRumcajs: Not sure I get it. It's not in the Docker [documentation](https://docs.docker.com/engine/reference/builder/#mount-types). Could you please elaborate?

hanabi1224: I was suggesting some extra steps
In dockerfile:
1. Install sccache
2. RUSTC_WRAPPER = "sccache rustc"
3. CC = "sccache clang"
4. CXX = "sccache clang++"
5. mount ~/.cache/sccache from host
In CI manifest:
`uses: hanabi1224/sccache-action@v1.2.0`

LesnyRumcajs: I am not sure about it. It seems to would complicate the otherwise relatively straightforward Dockerfile, and the cache I added here would still need to be present (rustup reasons). Perhaps we can touch this in another issue and verify that it significantly decreases the CI duration. What do you think?

hanabi1224: Sounds good

hanabi1224 (approved):

tyshko5 (approved):

PR #2524, opened 2023-02-07
fix: paritydb mem usage patch & turn off sync_wal option

hanabi1224:
## Summary of changes

- 

Closes 

## Change checklist


elmattic:
What is the impact in performance or memory usage of `sync_wal: false`?

hanabi1224:
> What is the impact in performance or memory usage of `sync_wal: false`

@elmattic it has no impact on memory usage, slightly improve import time, I believe we only need the crash resistance guarantee provided by `sync_data` but not the extra consistency guarantees provided by `sync_wal`

hanabi1224:
I constantly get errors about `cid not found` when running `walk_snapshot` with this patch, there might be some bug with their new default sse2 optimized `find_entry` method. I will test more and report to upstream. Do not merge until further confirmation

PR #2529, opened 2023-02-09
fix cargo warnings from workspace default features not set

LesnyRumcajs:
## Summary of changes

- fix the warnings generated for `cargo build` and others after the toolchain bump.

> When specifying a workspace dependency it is not obvious, and rather surprising that, not specifying default-features = false in the workspace dependency, means that any default-features = false is ignored in any member crate.

> This I assume is because the features of workspace dependencies are additive, and by not specifying default-features = false, the default features are automatically and always enabled.

https://github.com/rust-lang/cargo/issues/11329#issue-1434113367

Closes 

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,7 @@ derive_builder = "0.11"
 dialoguer = "0.10.2"
 digest = "0.10.5"
 directories = "4.0.1"
-filecoin-proofs-api = "12.0"
+filecoin-proofs-api = { version = "12.0", default-features = false }
 flume = "0.10"
 fs_extra = "1.2"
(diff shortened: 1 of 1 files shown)


hanabi1224 (approved):

lemmih (approved):

hanabi1224:
Not sure if it's worth the effort to add the check to the ruby script

LesnyRumcajs:
> Not sure if it's worth the effort to add the check to the ruby script

I don't think so. The warning implies that it may become a hard error in the future, plus everyone compiling Forest will immediately see it anyway.

PR #2534, opened 2023-02-10
tinker js lint

LesnyRumcajs:
## Summary of changes

- 

Closes 

## Change checklist


.github/workflows/js-lint.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+
+      # ESLint and Prettier must be in `package.json`
+      - name: Install Node.js dependencies
+        run: npm ci
+
+      - name: Run linters
+        uses: wearerequired/lint-action@v2
```
hanabi1224, line 32: Different packages may use different build/lint rules, better to manage CI in a mono repo way using tools like `lerna`.
a reference setup from mina: https://github.com/ChainSafe/mina-rs/blob/main/.github/workflows/build-and-test.yml#L215

LesnyRumcajs: I like it, thanks!

PR #2540, opened 2023-02-13
Changes to `walk_snapshot` to match Forest and Lotus snapshots

### LINKED ISSUE #1884: Investigate whether we can match Lotus' snapshots byte-for-byte
**Issue summary**

Both Lotus and Forest has the ability to generate snapshots. However, it has come to light that Forest snapshots fail after a day or two due to unexplained forks in the blockchain. Therefore our snapshots must be different from the snapshots from Lotus and we need to figure out why.

Tasks:
- [ ] Start with fairly recent calibnet snapshot from Lotus. (Bootstrap with a Lotus snapshot from our DO Space and generate a new snapshot.)
- [ ] Initiate both Lotus and Forest with the calibnet snapshot.
- [ ] Export a new snapshot with the same settings (epoch, recent stateroots, etc) from both Forest and Lotus.
- [ ] Check if they are exactly the same.
- [ ] If they are not, skim through the Lotus and Forest code to find differences. Add those differences to a new issue.

Lotus snapshots for calibnet: https://cloud.digitalocean.com/spaces/forest-snapshots?i=88c522&path=lotus-calibnet%2F

jdjaustin:
## Summary of changes

- Changes to `walk_snapshot` to match Lotus logic (i.e., don't include identity CIDs in snapshot export).

Closes #1884 

## Change checklist


jdjaustin:
Calibnet snapshots are matching up now but I'm so far unable to verify mainnet. @LesnyRumcajs

LesnyRumcajs:
@jdjaustin, what's the issue with mainnet snapshots? Is it because of the hardware limitations? If so, have you tried installing Forest with an alternative, turbo fast, pioneer & bleeding edge backend with `make install-with-jemalloc`

jdjaustin:
> @jdjaustin, what's the issue with mainnet snapshots? Is it because of the hardware limitations? If so, have you tried installing Forest with an alternative, turbo fast, pioneer & bleeding edge backend with `make install-with-jemalloc`

Yes, I tried `make install-with-jemalloc` and I'm still getting `No space left on device` errors. I can clear out some disk space and try again.

LesnyRumcajs:
Ah, so your issue is with disk space, not RAM. In this case, you need around 250G of free disk space.

1. clean up forest directory
2. download and import the snapshot - this would require around 250G at least - 110G snapshot + 130G DB, more or less.
3. Delete the snapshot to get back some disk space.
4. Export the snapshot - again, you will need around 250G plus some more because of the additional chain progression.

lemmih:
@jdjaustin If you still don't have enough disk space after deleting the old files, ask Hubert to set up a droplet for you.

jdjaustin:
Good news. Verified that the Lotus and Forest mainnet snapshots also match!
Screenshot from 2023-02-15 19-55-20
Screen Shot 2023-02-16 at 8 05 38 AM

LesnyRumcajs:
@jdjaustin does the shasum of both match?

jdjaustin:
> @jdjaustin does the shasum of both match?

Yepp! 
Screen Shot 2023-02-16 at 9 41 51 AM
Screenshot from 2023-02-16 09-49-59

LesnyRumcajs:
Rock solid! Now it would be fantastic to keep it this way. One approach would be to modify the test a bit in our workflow https://github.com/ChainSafe/forest/blob/main/.github/workflows/rust.yml#L150-L172

1. Download the Lotus snapshot.
2. Import it into Forest
3. Export the snapshot at height specified in Lotus snapshot.
4. Compare the checksums.

We can do it in a separate issue if it's too complex; I may be missing some gotchas.

lemmih (approved):

blockchain/chain/src/store/chain_store.rs
```diff
@@ -560,12 +563,15 @@ where
         Self::walk_snapshot(tipset, recent_roots, |cid| {
             let tx_clone = tx.clone();
             async move {
+                let prefix = Code::try_from(Prefix::from(&cid).mh_type)?;
```
LesnyRumcajs, line 566: I don't particularly like this `Prefix` struct that ultimately got [removed](https://github.com/multiformats/rust-cid/pull/49) from the `rust-cid`.

I do believe this should work and we could then get rid of the `blockchain/chain/src/store/prefix.rs`

Could you please try it out?
```suggestion
                let prefix = Code::try_from(cid.hash().code())?;
```

LesnyRumcajs (changes requested):

blockchain/chain/src/store/chain_store.rs
LesnyRumcajs: Perhaps even better, let me tinker around.

jdjaustin: Sounds good, IIRC I did try `cid.hash().code()` but that gave me an error during snapshot export.

LesnyRumcajs: @jdjaustin try to cherry-pick this commit https://github.com/ChainSafe/forest/commit/c131ab21a7b80b39088bc62089c15abdf54b66c0

It didn't fail on export and imported the snapshot fine, but I haven't checked that the checksums still match. Hopefully, they still do. Could you please verify this?

jdjaustin: Yes they still match. Compiling Forest from that branch, here are the results for the calibnet exports:
Screenshot from 2023-02-16 12-27-51
Screen Shot 2023-02-16 at 12 28 11 PM

LesnyRumcajs:
Linking the "source" PR for future reference if needed. https://github.com/filecoin-project/lotus/pull/8691

LesnyRumcajs:
@jdjaustin Perhaps we should match the logic in Lotus for the snapshot, just in case. I am not entirely sure we need it but better safe than sorry. So this:
```go
// We only include raw and dagcbor, for now.
// Raw for "code" CIDs.
switch prefix.Codec {
case cid.Raw, cid.DagCBOR:
default:
	continue
}
```
Should also get included.

LesnyRumcajs:
> > @jdjaustin Perhaps we should match the logic in Lotus for the snapshot, just in case. I am not entirely sure we need it but better safe than sorry. So this:
> > ```go
> > // We only include raw and dagcbor, for now.
> > // Raw for "code" CIDs.
> > switch prefix.Codec {
> > case cid.Raw, cid.DagCBOR:
> > default:
> > 	continue
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should also get included.
> 
> This block of code doesn't seem to be affecting snapshot exports currently, so my biggest question/concern with adding this is: how will I test the Rust code is working as intended?

I believe just checking if the snapshots are still the same will be okay.

jdjaustin:
> I believe just checking if the snapshots are still the same will be okay.

Got it, I thought it was just going to be dead code, but I was just confused by the Go code; once I started writing the Rust code, I understood what was going on. And the snapshots still match:
Screenshot from 2023-02-17 13-19-18
Screen Shot 2023-02-17 at 1 18 57 PM

blockchain/chain/src/store/chain_store.rs
```diff
@@ -570,7 +570,15 @@ where
 
                 // Don't include identity CIDs.
                 if u64::from(Code::Identity) != cid.hash().code() {
-                    tx_clone.send_async((cid, block.clone())).await?;
+                    // We only include raw and dagcbor, for now.
```
LesnyRumcajs, line 573: It'd be more concise with a single `if` condition e.g., `if code != identity && (codec == IPLD_RAW || codec == DAG_CBOR) { send_async } `.

LesnyRumcajs (approved):

hanabi1224 (approved):

tyshko5 (approved):

blockchain/chain/src/store/chain_store.rs
jdjaustin: Good suggestion. `cargo fmt` kind of mangled the formatting of the match statement, so this way definitely looks cleaner.

PR #2547, opened 2023-02-14
Seamless StateTree

LesnyRumcajs:
## Summary of changes

- Implemented the `StateTree` variant to handle multiple state tree versions in a single place. It should make transitions smoother.

Part of https://github.com/ChainSafe/forest/issues/2438

## Change checklist

### DIFF
--- a/blockchain/state_manager/src/vm_circ_supply.rs
+++ b/blockchain/state_manager/src/vm_circ_supply.rs
@@ -106,13 +106,12 @@ impl GenesisInfoVesting {
     }
 }
 
-fn get_actor_state<DB: Blockstore>(
+fn get_actor_state<DB: Blockstore + Clone>(
     state_tree: &StateTree<DB>,
     addr: &Address,
 ) -> Result<ActorState, anyhow::Error> {
     state_tree
-        .get_actor(addr)?
-        .map(|v| v.into())
+        .get_actor(&addr.into())?
         .ok_or_else(|| anyhow::anyhow!("Failed to get Actor for address {addr}"))
 }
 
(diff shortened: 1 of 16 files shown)


blockchain/state_manager/src/vm_circ_supply.rs
```diff
@@ -106,13 +106,12 @@ impl GenesisInfoVesting {
     }
 }
 
-fn get_actor_state<DB: Blockstore>(
+fn get_actor_state<DB: Blockstore + Clone>(
```
hanabi1224, line 109: Is this `Clone` used?

utils/forest_shim/src/address.rs
```diff
@@ -69,3 +69,16 @@ impl From<Address> for Address_v2 {
             .expect("Couldn't convert between FVM2 and FVM3 addresses")
     }
 }
+
+impl From<&Address> for Address_v2 {
+    fn from(other: &Address) -> Self {
+        Address_v2::from_bytes(&other.to_bytes())
+            .expect("Couldn't convert between FVM2 and FVM3 addresses")
```
hanabi1224, line 76: Maybe add some comments on why favor panic here than implementing `TryFrom` trait

lemmih: The two address versions are identical but the internals are hidden. If the conversion fails then Forest cannot continue in any way or form.

hanabi1224: @lemmih that makes sense. My suggestion is adding this to the code comment so that ppl won't wonder why not use `TryFrom` a few months later

blockchain/state_manager/src/vm_circ_supply.rs
LesnyRumcajs: [yes](https://github.com/ChainSafe/forest/blob/seamless-statetree/utils/forest_shim/src/state_tree.rs#L33)

utils/forest_shim/src/address.rs
LesnyRumcajs: Added the comment. Hopefully, it's helpful enough.

hanabi1224 (approved):

lemmih (approved):

Issue #2554, opened 2023-02-16
Update instructions on running forest docker with volume


hanabi1224:
**Issue summary**

When I run the below command, I got permission errors. I was able to work around by doing `chmod 777 /tmp/forest` but that might not be the best solution.  We also need to update the `README.md` with detailed instructions

CC  @LesnyRumcajs @samuelarogbonlo  any ideas?

```
docker run --init -it --rm --network host -v /tmp/forest:/home/forest/.local/share/forest ghcr.io/chainsafe/forest --e
ncrypt-keystore false --chain calibnet --auto-download-snapshot
```

**Other information and links**

LesnyRumcajs:
@hanabi1224 Yeah, onto it.

LesnyRumcajs:
Or maybe not. This isn't very pleasant. https://github.com/moby/moby/issues/2259

Based on a short investigation, the only "sane" way to have it with host-mounted volumes is to make the user change ownership of his host directory. So in this example, run `chown -R 1000 /tmp/forest`. It may be a reasonable thing to ask, what do you think?

LesnyRumcajs:
Regardless, I'll add some more instructions with more cases.

hanabi1224:
@LesnyRumcajs I spent some time following up on the discussion I missed about making rootless forest images, but I'm not convinced. 
When I search for `Permission denied` in https://github.com/bitnami/containers/ I see a few unresponded tickets.  Some of the points on security from the original post (which I see was published 4+ years ago) are no longer strong enough since `podman` came out and `docker` provided a rootless mode. 
In a [post](https://www.zend.com/blog/building-rootless-docker-images-zendphp) that advertises the rootless image, it also admits
> The best way to fix these issues is to run the daemon as a non-root user

And given the majority of the dockerized applications are not built rootless, I don't think it's worth it to sacrifice ease of use by asking users to do some extra steps before mounting a volume that they don't often do with majority of the popular images.

LesnyRumcajs:
@hanabi1224 Asking users to run an extra command on a single directory shared between the host and the container is less than making them use a different engine or making some non-trivial changes to the existing one. Especially given that it's a particular case - you want to run containerized Forest with your current database created with local Forest.

I agree that `podman` and rootless docker are ultimately nicer to have, but many folks are grabbing the `Docker for Mac` and don't want to tinker around.

Instead, I suggest listing a few cases of running Forest.
* run a container with its private DB,
[...]

PR #2562, opened 2023-02-17
add docker documentation

LesnyRumcajs:
## Summary of changes

- added Docker documentation with some examples and use cases

Closes https://github.com/ChainSafe/forest/issues/2554

## Change checklist


tyshko5 (approved):

documentation/src/docker.md
```diff
@@ -0,0 +1,108 @@
...
+❯ docker exec -it forest forest-cli net peers
+12D3KooWAh4qiT3ZRZgctVJ8AWwRva9AncjMRVBSkFwNjTx3EpEr, [/ip4/10.0.2.215/tcp/1347, /ip4/52.12.185.166/tcp/1347]
+12D3KooWMY4VdMsdbFwkHv9HxX2jZsUdCcWFX5F5VGzBPZkdxyVr, [/ip4/162.219.87.149/tcp/30141, /ip4/162.219.87.149/tcp/30141/p2p/12D3KooWMY4VdMsdbFwkHv9HxX2jZsUdCcWFX5F5VGzBPZkdxyVr]
+12D3KooWFWUqE9jgXvcKHWieYs9nhyp6NF4ftwLGAHm4sCv73jjK, [/dns4/bootstrap-3.calibration.fildev.network/tcp/1347]
+```
+
+### Use a shared volume to utilise across different Forest images
+Create the volume
+```shell
+docker volume create forest-data
```
hanabi1224, line 49: It looks good so far, could we have an instruction for the desired keystore management as well, see https://github.com/ChainSafe/forest/issues/2554#issuecomment-1434946039

LesnyRumcajs: I am not sure I get it. At the moment, the shared volume is also sharing the keystore. The desired (with the keystore and wallet moved away from Forest data directory) management should not be a part of the current set of instructions - we want to document what is available now.

documentation/src/docker.md
```diff
@@ -0,0 +1,108 @@
...
+```
+Now, whenever you create a new Forest container, attach the volume to where the data is stored `/home/forest/.local/share/forest`.
+```shell
+❯ docker run --init -it --rm --network host \
```
hanabi1224, line 53: Better to set ulimit as well by `--ulimit nofile=60000`

hanabi1224: `--network host` looks scary.  Better to map all ports or we could at least suggest using rootless docker that mitigates this vulnerability.

https://docs.docker.com/engine/security/rootless/
> Host network (docker run --net=host) is also namespaced inside RootlessKit.

hanabi1224: As per my testing result, setting `--memory-swappiness 1` is beneficial, however, the parameter does not seem to work with the default docker installation, alternatively we could ask setting `sudo sysctl -w vm.swappiness=1` on host machine from which docker container will inherit.
For more details: https://github.com/ChainSafe/forest/issues/2563

LesnyRumcajs: > `--network host` looks scary. Better to map all ports or we could at least suggest using rootless docker that mitigates this vulnerability.
> 
> https://docs.docker.com/engine/security/rootless/
> 
> > Host network (docker run --net=host) is also namespaced inside RootlessKit.

It's scary because it shouldn't be here. I guess I mindlessly copied it somewhere from the depths of my shell history. Good catch.

hanabi1224 (approved):

documentation/src/docker.md
LesnyRumcajs: @hanabi1224 As per our prior GH and Slack conversation, could you create a section here with Docker setup recommendations? Your points regarding running rootless are more than valid, and mentioning all those things (podman included) would be beneficial. It'd be a shame if it ended up as another forgotten Slack thread. Mentioning the swappiness would also make sense there, given it's not necessarily required but highly recommended.

LesnyRumcajs: > 60000

Why this value exactly?

hanabi1224: > > 60000
> 
> Why this value exactly?

I just randomly chose a big value

hanabi1224: > > `--network host` looks scary. Better to map all ports or we could at least suggest using rootless docker that mitigates this vulnerability.
> > https://docs.docker.com/engine/security/rootless/
> > > Host network (docker run --net=host) is also namespaced inside RootlessKit.
> 
> It's scary because it shouldn't be here. I guess I mindlessly copied it somewhere from the depths of my shell history. Good catch.

Note that without mapping the libp2p port, the listen addr does not work, all connections to other nodes have to be initiated from the forest daemon

LesnyRumcajs: > Note that without mapping the libp2p port, the listen addr does not work, all connections to other nodes have to be initiated from the forest daemon

That's not great. What is the practical impact of that? The host network doesn't really work great on Docker for Mac as far as I remember.

Perhaps using a dedicated port for libp2p is the way to go? What do you think?

hanabi1224: > @hanabi1224 As per our prior GH and Slack conversation, could you create a section here with Docker setup recommendations? Your points regarding running rootless are more than valid, and mentioning all those things (podman included) would be beneficial. It'd be a shame if it ended up as another forgotten Slack thread. Mentioning the swappiness would also make sense there, given it's not necessarily required but highly recommended.


### Security recommendations
- We strongly recommend running the docker daemon in rootless mode ([installation instruction](https://docs.docker.com/engine/security/rootless/)), or running the daemon-less docker alternative `podman` ([installation instruction](https://podman.io/getting-started/installation)) with non-root user and put `alias docker = podman` or manually replace the `docker` commands with `podman` in below instructions

### Performance recommendations
- We recommend lowering the swappiness kernel parameter on linux to 1-10 for long running forest node by doing `sudo sysctl -w vm.swappiness=[n]`. References: [[1]](https://en.wikipedia.org/wiki/Memory_paging#Swappiness)  [[2]](https://linuxhint.com/understanding_vm_swappiness/)

hanabi1224: > > Note that without mapping the libp2p port, the listen addr does not work, all connections to other nodes have to be initiated from the forest daemon
> 
> That's not great. What is the practical impact of that? The host network doesn't really work great on Docker for Mac as far as I remember.
> 
> Perhaps using a dedicated port for libp2p is the way to go? What do you think?

We could track this in a separate issue. With the current bootstrap peer list and the max peers config, the impact of not listening to an public address is trival

LesnyRumcajs: Thanks, co-authored the recommendations.

PR #2571, opened 2023-02-21
Retries to snapshot fetch daemon and cli

### LINKED ISSUE #2544: Add retries/resume to snapshot fetching
**Issue summary**

The mainnet hosting seems volatile. Forest has been failing numerous times in the CI (so dockerized, with aria2) and locally (without aria2). The error is the same, something failed in the middle of downloading.

Dockerized (aria2)
```
2023-02-13T20:45:35.637704Z  INFO forest_cli_shared::cli::snapshot_fetch: Snapshot url: https://pub-fd31751bcb69400eb39e694385c19457.r2.dev/minimal/2599920_2023_02_13T16_00_00Z.car    
2023-02-13T20:45:35.637744Z  INFO forest_cli_shared::cli::snapshot_fetch: Snapshot will be downloaded to /volumes/forest_data/snapshots/mainnet/filecoin_snapshot_mainnet_2023-02-13_height_2599920.car (117.14 GiB)    
2023-02-13T20:46:28.219805Z ERROR forest_cli::cli: Error: Failed fetching the snapshot: error reading a body from connection: unexpected end of file    
```

No logs for the local build but it's roughly the same.

This fails the CI from time to time and is annoying when setting up a server for a night job. 

The goal is to implement a retry mechanism (or, even better, retry with a resume - something like this should be available with aria2 almost out of the box; for vanilla, we can be pragmatic and retry from scratch). Only if the download fails a few times should the daemon fail.

**Task summary**

- [X] Add retry parameter to config. It doesn't have to be exposed as a flag. Set it to a sane default.
- [X] Add a retry interval parameter. Set it to a sane default.
- [X] Use the parameters above to implement retry and retry/resume for snapshot fetch. This should work for both the daemon trying to download the snapshot and the `forest-cli snapshot fetch`.
- [X] Implement tests validating the logic

**Acceptance Criteria**

- [X] Forest daemon no longer fails after a single failure in the upstream host
- [X] Feature is properly tested
- [X] Changelog is updated

sudo-shashank:
## Summary of changes

- Added a retry! macro which can be used for volatile functions 
- Used it for snapshot_fetch in daemon and cli

Closes #2544 

## Change checklist


forest/shared/src/macros.rs
```diff
@@ -0,0 +1,28 @@
...
+#[macro_export]
+macro_rules! retry {
+    ($func:ident, $max_retries:expr, $delay:expr $(, $arg:expr)*) => {{
+        let mut retry_count = 0;
+        loop {
+            match $func($($arg),*).await {
+                Ok(val) => break Ok(val),
+                Err(err) => {
+                    retry_count += 1;
+                    info!("Retry attempt {} started with delay {:#?}.", retry_count, $delay);
```
hanabi1224, line 18: Should this be after the `break` line?

hanabi1224: I think `Retry attemp {} will start in {delay}s` would be more precise, given `sleep` happens after this line

hanabi1224 (approved):

forest/shared/src/macros.rs
sudo-shashank: right, fixed

creativcoder (approved):

forest/shared/src/macros.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::time::Duration;
+
+pub const DEFAULT_RETRIES: i32 = 3;
+pub const DEFAULT_DELAY: Duration = Duration::from_secs(60);
+
+#[macro_export]
```
LesnyRumcajs, line 9: Even though the snapshot fetch can get tricky to test, I believe this macro would be trivial to cover with unit tests. What do you think?

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -67,6 +70,12 @@ pub enum SnapshotCommands {
         /// Use [`aria2`](https://aria2.github.io/) for downloading, default is false. Requires `aria2c` in PATH.
         #[arg(long)]
         aria2: bool,
+        /// Maximum number of times to retry the fetch
+        #[arg(short, long, default_value = "3")]
+        max_retries: Option<i32>,
+        /// Duration to wait between the retries in seconds
+        #[arg(short, long, default_value = "60")]
+        delay: Option<u64>,
```
LesnyRumcajs, line 78: Could we use a strongly typed `Duration` instead?

LesnyRumcajs (changes requested):

forest/shared/src/macros.rs
sudo-shashank: yeah I can add it

LesnyRumcajs: Fantastic, thanks.

forest/cli/src/cli/snapshot_cmd.rs
sudo-shashank: @LesnyRumcajs i didn't find a way to add a custom parser inside #[arg()]

LesnyRumcajs: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=71cbf458c2ce3e21fa6312d8ace3f1db

The parsing method could be put somewhere in `utils`, so it can be re-used.

forest/daemon/src/daemon.rs
```diff
@@ -388,7 +390,15 @@ async fn maybe_fetch_snapshot(
 ) -> anyhow::Result<Config> {
     if should_fetch_snapshot {
         let snapshot_path = default_snapshot_dir(&config);
-        let path = snapshot_fetch(&snapshot_path, &config, &None, is_aria2_installed()).await?;
+        let path = retry!(
+            snapshot_fetch,
+            DEFAULT_RETRIES,
+            DEFAULT_DELAY,
```
LesnyRumcajs, line 396: How about putting those into the daemon configuration? They don't have to be exposed as CLI flags, but it'd be nice to alter them via config file.

forest/shared/tests/macros_test.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright 2019-2023 ChainSafe Systems
```
LesnyRumcajs, line 1: Could we do it without any `sleep`s or keeping them to the bare minimum? This adds around 1s, but it doesn't have to unless I miss something.

LesnyRumcajs (changes requested):

forest/cli/src/cli/snapshot_cmd.rs
LesnyRumcajs: Even better, there's a crate so I believe we could just hook it up to the parser https://docs.rs/parse_duration/latest/parse_duration/

forest/shared/tests/macros_test.rs
sudo-shashank: fixed

forest/daemon/src/daemon.rs
sudo-shashank: fixed

sudo-shashank:
@LesnyRumcajs this PR is ready for final review

utils/forest_utils/tests/macros_test.rs
```diff
@@ -0,0 +1,56 @@
...
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::time::Duration;
+
+use forest_utils::retry;
+use log::info;
+use tokio::time::sleep;
+
+async fn failing_function() -> Result<(), ()> {
+    sleep(Duration::from_nanos(0)).await;
```
LesnyRumcajs, line 11: Why do we need to `sleep(0)`? Isn't it virtually the same as not sleeping at all?

LesnyRumcajs (approved):

utils/forest_utils/tests/macros_test.rs
sudo-shashank: this is for lint check actually, I needed await command inside of async fn

LesnyRumcajs: Makes sense. It's not a huge deal, so I'm fine with this.

LesnyRumcajs: Alternatively, we could suppress the linter warning in those few methods. It'd be more self-explaining.

sudo-shashank: I suppressed the linter warning for those 3 methods

PR #2578, opened 2023-02-21
Initial support for forest-cli info

### LINKED ISSUE #2442: Implement forest-cli info command.
**Issue summary**

Forest lacks equivalent Lotus command `info` that prints various informations:

```
$ ./lotus info
Network: calibrationnet
StartTime: 2h8m39s (started at 2023-01-17 11:53:20 +0100 CET)
Chain: [sync behind! (1842h49m0s behind)] [basefee 100 pFIL] [epoch 0]
Peers to: [publish messages 47] [publish blocks 47]
Chain health: 0% [unhealthy]

Default address: address not set

Wallet: 0 address
      Total balance: 0
      Market locked: 0
      Market available: 0

Payment Channels: 0 channel

Deals: 0, 0 B

Bandwidth:
      TotalIn  TotalOut  RateIn    RateOut
      968 MB   31 MB     1.4 kB/s  1.3 kB/s
```

We should start with most basic ones, so `Network`, `StartTime`, `Chain`. Possibly `Peers to` and `Chain health`.

**Task summary**

- [ ] Study how `lotus info` is implemented
- [ ] Implement basic output

**Acceptance Criteria**

- [ ] Calling `forest-cli info` shows some basic node information

https://lotus.filecoin.io/lotus/manage/lotus-cli/#lotus-info
https://github.com/filecoin-project/lotus/blob/master/api/api_common.go#L54
https://github.com/filecoin-project/lotus/blob/master/cli/info.go#L31

creativcoder:
## Summary of changes

Add a new info subcommand to forest-cli with the following information:

* Network
* Uptime
* Chain status information
* Chain health

This is ported from https://github.com/filecoin-project/lotus/blob/master/cli/info.go"

forest-cli info output:
```
Network: calibnet
Uptime: 0h 3m 210s (Started at: 2023-05-11 09:52:09.007870302 +00:00)
Chain: [sync: Ok! (9s behind)] [basefee: 100 atto FIL] [epoch: 549085]
Chain health: 71.58%


Default wallet address: -
```

Closes #2442

## Change checklist


forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+#[derive(Debug)]
+enum SyncStatus {
+    Ok,
+    Slow,
+    Behind,
+}
+
+impl std::fmt::Display for SyncStatus {
```
hanabi1224, line 29: Maybe use `string_enum` crate instead?

lemmih:
@creativcoder Have you finished the change checklist?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        for _ in 100..chain_finality {
+            block_count += ts.blocks().len();
+            let tsk = ts.parents();
+            let tsk = TipsetKeysJson(tsk.clone());
+            if let Ok(tsjson) = chain_get_tipset((tsk,), &config.client.rpc_token).await {
+                ts = tsjson.0;
+            }
+        }
```
lemmih, line 82: What's the difference between these two `for` loops?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,149 @@
...
+    health: usize,
+    epoch: i64,
+    base_fee: String,
+    sync_status: SyncStatus,
+}
+
+pub async fn node_status(config: &Config) -> NodeStatusInfo {
+    let chain_head = match chain_head(&config.client.rpc_token).await {
+        Ok(head) => head.0,
+        Err(_) => cli_error_and_die("Could not get network head. Is the node running?", 1),
```
lemmih, line 40: Let's handle all errors in `main`.

forest/cli/src/cli/info_cmd.rs
creativcoder: This was simply a side effect of verbatim port,  except this one doesn't have the intermediary variable being set from lotus codebase:
```
status.ChainStatus.BlocksPerTipsetLast100 = float64(blockCnt) / 100
```
Didn't need that. 
I will combine them

forest/cli/src/cli/info_cmd.rs
creativcoder:
```
error[E0277]: `(dyn erased_serde::ser::Serialize + std::marker::Send + 'static)` cannot be shared between threads safely
  --> forest/cli/src/cli/info_cmd.rs:37:79
   |
37 |     let chain_head = chain_head(&config.client.rpc_token).await.map(|tp| tp.0)?;
   |                                                                               ^ `(dyn erased_serde::ser::Serialize + std::marker::Send + 'static)` cannot be shared between threads safely
   ```
   
 Not sure if i can make this work with ?

Comment (creativcoder) [reply forest/cli/src/cli/info_cmd.rs:40]: nvm, using anyhow::Error::msg

File: forest/cli/src/cli/info_cmd.rs [hunk 4]
```diff
@@ -0,0 +1,138 @@
...
+            }
+            (86..) => {
+                let chain_health = format!("{health}%\n\n").green();
+                println!("Chain health: {chain_health}");
+            }
+            _ => {}
+        }
+
+        // Wallet info
+        let default_wallet_address = wallet_default_address((), &config.client.rpc_token)
```
lemmih, line 130: Apparently we don't always have a default address. Could you open a new issue for changing this RPC call to return an optional address?

creativcoder:
@lemmih So far we have the following:
```
➜  forest2 git:(creativcoder/forest-info) ✗ ./target/release/forest-cli --chain calibnet info show
Network: calibnet
Uptime: 10h 27m 53s
Chain: [sync: Behind! (55959663 behind)] [basefee: 0.0000000000000001 pFIL] [epoch: 382117]
Chain health: 60%

Default wallet address: t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y
```

node/rpc-api/src/lib.rs
```diff
@@ -54,6 +54,7 @@ pub static ACCESS_MAP: Lazy<HashMap<&str, Access>> = Lazy::new(|| {
     // Wallet API
     access.insert(wallet_api::WALLET_BALANCE, Access::Write);
     access.insert(wallet_api::WALLET_DEFAULT_ADDRESS, Access::Write);
+    access.insert(wallet_api::WALLET_DEFAULT_ADDRESS, Access::Read);
```
lemmih, line 57: You probably want to remove one of these lines.

node/rpc-api/src/lib.rs
```diff
@@ -333,6 +335,10 @@ pub mod state_api {
     pub type StateNetworkNameParams = ();
     pub type StateNetworkNameResult = String;
 
+    pub const STATE_START_TIME: &str = "Filecoin.StateStartTime";
+    pub type StateStartTimeParams = ();
+    pub type StateStartTimeResult = String;
```
lemmih, line 340: This RPC call returns a `String` to match the API in Lotus?

forest/cli/src/cli/info_cmd.rs
creativcoder: Created: https://github.com/ChainSafe/forest/issues/2663

lemmih (changes requested):

lemmih:
> @lemmih So far we have the following:
> 
> ```
> ➜  forest2 git:(creativcoder/forest-info) ✗ ./target/release/forest-cli --chain calibnet info show
> Network: calibnet
> Uptime: 10h 27m 53s
> Chain: [sync: Behind! (55959663 behind)] [basefee: 0.0000000000000001 pFIL] [epoch: 382117]
> Chain health: 60%
> 
> Default wallet address: t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y
> ```

This should be in the PR description.

lemmih:
And concepts such as `chain health` need to be documented.

node/rpc/src/sync_api.rs
```diff
@@ -151,6 +152,7 @@ mod tests {
             .unwrap()
         };
         let (new_mined_block_tx, _) = flume::bounded(5);
+        let start_time = OffsetDateTime::now_utc();
```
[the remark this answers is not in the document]
creativcoder: Created: https://github.com/ChainSafe/forest/issues/2664

node/rpc-api/src/lib.rs
creativcoder: Ah this should be part of common_api, i've moved it. Lotus returns a time.Time, i've refactored accordingly to return a OffsetDateTime.

LesnyRumcajs:
`0.0000000000000001 pFIL` I think I lack Maths PhD to decode that on the fly. Could we use attoFIL?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,163 @@
...
+            })
+            .map_err(handle_rpc_err)?;
+
+        let network = chain_get_name((), &config.client.rpc_token)
+            .await
+            .map_err(handle_rpc_err)?;
+
+        let behind = OffsetDateTime::from_unix_timestamp(node_status.behind as i64)?;
+        let health = node_status.health;
+        let base_fee = node_status.base_fee.to_string();
```
lemmih, line 123: Instead of `to_string`, you can use `format_balance_string`.

creativcoder: I couldn't find that method for num-bigint or TokenAmount, i'm calling `atto()` on TokenAmount instead:
```
Chain: [sync: Behind! (6h 56m 0s behind)] [basefee: 100 aFIL] [epoch: 384566
```

node/rpc-api/src/lib.rs
creativcoder: Removed

forest/cli/src/cli/info_cmd.rs
lemmih: `format_balance_string` is in the `wallet_cmd.rs` file.

creativcoder: using `format_balance_string`

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,159 @@
...
+        SyncStatus::Ok
+    } else if delta < 30 * 5 {
+        SyncStatus::Slow
+    } else {
+        SyncStatus::Behind
+    };
+
+    let base_fee = chain_head.0.min_ticket_block().parent_base_fee().clone();
+    // .atto()
+    // .clone();
```
lemmih, line 67: These can be removed, I think.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,159 @@
...
+    Slow,
+    Behind,
+}
+
+pub struct NodeStatusInfo {
+    /// timestamp of how far behind the node is with respect to syncing to head
+    behind: u64,
+    /// Chain health calculated as percentage: amount of blocks in last finality
+    /// very healthy amount of blocks in a finality (900 epochs * 5 blocks per
+    /// tipset)
```
lemmih, line 35: I struggle to parse this sentence. Do you mean it's the number of blocks in the last 900 epochs divided by 900*5?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,159 @@
...
+                ts = tsjson.0;
+            }
+        }
+
+        block_count / chain_finality as usize
+    } else {
+        0
+    };
+
+    let health = 100 * (900 * blocks_per_tipset_last_finality) / (900 * 5);
```
lemmih, line 88: Is this a typo or can it be simplified to:
```suggestion
    let health = 100 * blocks_per_tipset_last_finality / 5;
```

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,159 @@
...
+            0..=85 => {
+                let chain_health = format!("{health}%\n\n").red();
+                println!("Chain health: {chain_health}");
+            }
+            (86..) => {
+                let chain_health = format!("{health}%\n\n").green();
+                println!("Chain health: {chain_health}");
+            }
+            _ => {}
+        }
```
lemmih, line 149: This can be refactored.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -31,7 +31,7 @@ use super::{handle_rpc_err, Config};
 const NUM_SIGNIFICANT_DIGITS: u32 = 4;
 
 #[allow(clippy::enum_variant_names)]
-enum FormattingMode {
+pub enum FormattingMode {
```
lemmih, line 34: Let's move this to the utility crate.

forest/daemon/Cargo.toml
```diff
@@ -49,7 +49,7 @@ rpassword.workspace = true
 serde_json.workspace = true
 shared_memory = "0.12"
 tempfile.workspace = true
-time.workspace = true
+time = { workspace = true }
```
lemmih, line 52: Let's keep this unchanged.

lemmih (changes requested):

forest/daemon/Cargo.toml
creativcoder: reverted

forest/cli/src/cli/wallet_cmd.rs
creativcoder: moved

forest/cli/src/cli/info_cmd.rs
creativcoder: done.

forest/cli/src/cli/info_cmd.rs
creativcoder: I could have simplified this, but lotus impl has kept it this way to as it's written in the doc: https://github.com/filecoin-project/lotus/blob/eed1a098c625cbbd29e490ea4f7e5286b8f74f76/cli/info.go#L66

forest/cli/src/cli/info_cmd.rs
creativcoder: Ah I omitted the divide symbol, or maybe fmt did. Updated. 
taken from here: https://github.com/filecoin-project/lotus/blob/eed1a098c625cbbd29e490ea4f7e5286b8f74f76/cli/info.go#L65

forest/cli/src/cli/info_cmd.rs
creativcoder: removed

forest/cli/src/cli/info_cmd.rs
lemmih: I think the comment in Lotus could be a lot better. Let's rewrite our comment to be more readable.

forest/cli/src/cli/info_cmd.rs
lemmih: Let's ignore the Lotus code and then implement this functionality in the way you think makes the most sense.

forest/cli/src/cli/wallet_cmd.rs
```diff
@@ -305,7 +285,7 @@ impl WalletCommands {
 /// in full accuracy for `ExactNotFixed` mode,
 /// mode to show data in SI units
 /// with 4 significant digits for `NotExactNotFixed` mode
-fn format_balance_string(
+pub fn format_balance_string(
```
lemmih, line 288: This function belongs in the utility crate together with `FormattingMode`

lemmih (changes requested):

forest/cli/src/cli/wallet_cmd.rs
creativcoder: moved

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,152 @@
...
+    Slow,
+    Behind,
+}
+
+pub struct NodeStatusInfo {
+    /// timestamp of how far behind the node is with respect to syncing to head
+    behind: u64,
+    /// Chain health is defined as the amount of blocks in last finality divided
+    /// by healthy amount of blocks in a finality (900 epochs * 5 blocks per
+    /// tipset).
```
lemmih, line 36: Let's rewrite this to something simpler. What we're really measuring is how close we are to having an average of 5 blocks per tipset in the last couple of hours. You could write that the number of blocks per tipset is non-deterministic but averaging at 5 is considered healthy.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,152 @@
...
+#[derive(Debug, Subcommand)]
+pub enum InfoCommand {
+    Show,
+}
+
+#[derive(Debug, strum_macros::Display)]
+enum SyncStatus {
+    Ok,
+    Slow,
+    Behind,
```
lemmih, line 28: When are we `slow` and when are we `behind`?

forest/cli/src/cli/info_cmd.rs
lemmih: Also, we might want to copy `fvm_shared::BLOCKS_PER_EPOCH` over to Forest and use it. It defines the desired number of blocks per epoch.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,152 @@
...
+        .map_err(|_| anyhow::Error::msg("couldn't fetch chain head, is the node running?"))?;
+
+    let chain_finality = config.chain.policy.chain_finality;
+    let epoch = chain_head.0.epoch();
+    let ts = chain_head.0.min_timestamp();
+    let now = Instant::now().elapsed().as_secs();
+    let delta = ts - now;
+    let behind = delta;
+
+    let sync_status = if delta < 30 * 3 / 2 {
```
lemmih, line 58: Is the `30` here related to `EPOCH_DURATION_SECONDS` ?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,152 @@
...
+        for _ in 0..chain_finality {
+            block_count += ts.blocks().len();
+            let tsk = ts.parents();
+            let tsk = TipsetKeysJson(tsk.clone());
+            if let Ok(tsjson) = chain_get_tipset((tsk,), &config.client.rpc_token).await {
+                ts = tsjson.0;
+            }
+        }
+
+        block_count / chain_finality as usize
```
lemmih, line 82: Don't you get rounding issues when you compute the answer in a `usize`? Looks to me like the health will always be a multiple of 20% (eg 0%, 20%, 40%, etc).

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,152 @@
...
+    } else if delta < 30 * 5 {
+        SyncStatus::Slow
+    } else {
+        SyncStatus::Behind
+    };
+
+    let base_fee = chain_head.0.min_ticket_block().parent_base_fee().clone();
+
+    // chain health
+    let blocks_per_tipset_last_finality = if epoch > chain_finality {
```
lemmih, line 69: Instead of returning a health of `0%` if epoch is less than `chain_finality`, how about we look at the average block count for those tipsets that are available? If 100 tipsets are available, find the average block count for those 100 tipsets, etc.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,152 @@
...
+        behind,
+        health,
+        epoch,
+        base_fee,
+        sync_status,
+    })
+}
+
+impl InfoCommand {
+    pub async fn run(&self, config: Config) -> anyhow::Result<()> {
```
lemmih, line 99: Can we refactor this code to separate the code with side-effects and the code without side-effects? That way, we can write more unit tests.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,152 @@
...
+        .blue();
+
+        println!("Network: {}", network.green());
+        println!("Uptime: {start_time}");
+        println!("Chain: {chain_status}");
+
+        let mut chain_health = format!("{health}%\n\n").red();
+        if health > 85 {
+            chain_health = chain_health.green();
+        }
```
lemmih, line 139: Could we take the `--color` flag into account?

lemmih (changes requested):

forest/cli/src/cli/info_cmd.rs
creativcoder: Updated health definition, and using BLOCKS_PER_EPOCH

forest/cli/src/cli/info_cmd.rs
creativcoder: When delta `(current_ts - head.min_timestamp)` is within 5 seconds => `slow`
When delta is > 5 seconds => `behind`.

taken from:
https://github.com/filecoin-project/lotus/blob/bf666a3f7e35317b7400d2059685065eba7ec860/cli/info.go#L215

I've added additional comments at the conditionals.

forest/cli/src/cli/info_cmd.rs
creativcoder: yes, modified to use from fvm_shared.

forest/cli/src/cli/info_cmd.rs
creativcoder: modified to use `f64`.

forest/cli/src/cli/info_cmd.rs
creativcoder: I'm not sure which --color flag is being refered to but chain_health already takes color into account and starts as a red formatted string and then if health > 85, it's green.

lemmih: It would be nice if `forest-cli --color never info show` didn't use colors.

creativcoder: Added color flag conditional.

forest/cli/src/cli/info_cmd.rs
creativcoder: Added average block count in else block

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,223 @@
...
+    let epoch = chain_head.0.epoch();
+    let ts = chain_head.0.min_timestamp();
+    let now = Instant::now().elapsed().as_secs();
+    let delta = ts - now;
+    let behind = delta;
+
+    let sync_status = if delta < EPOCH_DURATION_SECONDS as u64 * 3 / 2 {
+        // within 1.5 epochs
+        SyncStatus::Ok
+    } else if delta < 30 * 5 {
```
lemmih, line 64: I guess the `30` also refers to `EPOCH_DURATION_SECONDS`.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,223 @@
...
+use crate::cli::handle_rpc_err;
+
+#[derive(Debug, Subcommand)]
+pub enum InfoCommand {
+    Show,
+}
+
+#[derive(Debug, strum_macros::Display)]
+enum SyncStatus {
+    Ok,
```
lemmih, line 28: Do we have any user-facing documentation that explains `Ok` is within 1.5 epochs, `Slow` is within 5 epochs and `Behind` is >5 epochs after `HEAD`?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,223 @@
...
+    behind: u64,
+    epoch: i64,
+    base_fee: TokenAmount,
+    default_wallet_address: Option<String>,
+    color: &LoggingColor,
+) -> anyhow::Result<()> {
+    let use_color = match color {
+        LoggingColor::Auto | LoggingColor::Always => true,
+        LoggingColor::Never => false,
+    };
```
lemmih, line 165: This is buggy. Use `color.coloring_enabled()`.

lemmih (changes requested):
You still haven't separated the formatting code and the IO code, and you haven't written any tests.

forest/cli/src/cli/info_cmd.rs
hanabi1224: Fixed.

forest/cli/src/cli/info_cmd.rs
hanabi1224: Fixed

forest/cli/src/cli/info_cmd.rs
hanabi1224: Fixed

node/rpc-api/src/lib.rs
```diff
@@ -53,7 +53,7 @@ pub static ACCESS_MAP: Lazy<HashMap<&str, Access>> = Lazy::new(|| {
 
     // Wallet API
     access.insert(wallet_api::WALLET_BALANCE, Access::Write);
-    access.insert(wallet_api::WALLET_DEFAULT_ADDRESS, Access::Write);
+    access.insert(wallet_api::WALLET_DEFAULT_ADDRESS, Access::Read);
```
hanabi1224, line 56: @lemmih Is this change expected or only for testing?

lemmih: Expected, I think. Should be okay to have the default wallet address be publically readable.

forest/cli/src/cli/info_cmd.rs
creativcoder: Nope, lotus impl doesn't have any docs for that: https://github.com/filecoin-project/lotus/blob/76776e5cc1ce288e7fa6c60980949ed74633f24f/cli/info.go#L215 Neither i could find anything in filecoin spec.

lemmih (changes requested):

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,375 @@
...
+
+fn get_node_status(
+    chain_head: &Arc<Tipset>,
+    block_count: usize,
+    num_tipsets: u64,
+) -> anyhow::Result<NodeStatusInfo> {
+    let epoch = chain_head.epoch();
+    let ts = chain_head.min_timestamp();
+    let now = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
+    // log::info!("ts: {ts}, now: {now}");
```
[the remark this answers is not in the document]
creativcoder: removed

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,375 @@
...
+    health: f64,
+    /// epoch the node is currently at
+    epoch: i64,
+    /// base fee
+    base_fee: TokenAmount,
+    /// sync status information
+    sync_status: SyncStatus,
+}
+
+fn get_node_status(
```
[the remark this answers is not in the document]
creativcoder: I've moved it outside of get_node_status.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,375 @@
...
+    let NodeStatusInfo {
+        behind,
+        epoch,
+        base_fee,
+        sync_status,
+        ..
+    } = node_status;
+    let base_fee = format_balance_string(base_fee.clone(), FormattingMode::NotExactNotFixed)?;
+    let behind = {
+        let b = Utc.timestamp_millis_opt(*behind as i64).unwrap();
```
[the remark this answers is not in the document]
creativcoder: fixed

forest/cli/src/cli/send_cmd.rs
```diff
@@ -186,12 +186,13 @@ impl SendCommand {
...
     use std::str::FromStr;
 
+    use forest_utils::io::parser::{format_balance_string, FormattingMode};
     use fvm_shared::econ::TokenAmount;
     use jsonrpc_v2::ErrorLike;
     use quickcheck_macros::quickcheck;
 
     use super::*;
-    use crate::cli::wallet_cmd::{format_balance_string, FormattingMode};
+    // use crate::cli::wallet_cmd::{format_balance_string, FormattingMode};
```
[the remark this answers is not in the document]
creativcoder: removed

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,429 @@
...
+    use colored::*;
+    use forest_blocks::{BlockHeader, Tipset};
+    use forest_cli_shared::logger::LoggingColor;
+    use forest_shim::{address::Address, econ::TokenAmount};
+    use fvm_shared::clock::EPOCH_DURATION_SECONDS;
+
+    use super::{fmt_info, get_node_status, NodeStatusInfo};
+    use crate::cli::info_cmd::{chain_status, SyncStatus};
+
+    fn mock_tipset_at(duration: u64) -> Arc<Tipset> {
```
lemmih, line 277: To clarify it, I'd name this argument `timestamp` or `seconds_since_unix_epoch`.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,429 @@
...
+        assert_eq!(node_status.sync_status, SyncStatus::Ok);
+    }
+
+    #[test]
+    fn block_sync_timestamp() {
+        let start_time = Local::now();
+        let network = "calibnet";
+        let default_wallet_address = Some("x".to_string());
+        let color = LoggingColor::Never;
+        let duration_now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
```
lemmih, line 333: You don't need to worry about the system time here.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,429 @@
...
+            epoch: 0,
+            base_fee: TokenAmount::from_whole(1),
+            sync_status: SyncStatus::Behind,
+        }
+    }
+
+    #[test]
+    fn node_status_with_null_tipset() {
+        // out of sync
+        let duration_now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
```
lemmih, line 310: Refrain from using system time in the tests. If you want to test a wide range of timestamps, write a quickcheck test.

lemmih (dismissed):
I'm not sold on the `NodeInfoOutput` structure. In my eyes, it makes the code longer and more difficult to test.

Let's move the FIL parsing and pretty-printing into the `forest_shim::econ` module.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,429 @@
...
+        if ts <= cur_duration_secs + 1 {
+            0
+        } else {
+            anyhow::bail!(
+                "System time should not be behind tipset timestamp, please sync the system clock."
+            );
+        }
+    } else {
+        cur_duration_secs - ts
+    };
```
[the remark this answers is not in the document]
creativcoder: done

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,429 @@
...
+        SyncStatus::Behind
+    };
+
+    let base_fee = chain_head.min_ticket_block().parent_base_fee().clone();
+
+    let health = (100 * block_count) as f64 / (num_tipsets * BLOCKS_PER_EPOCH) as f64;
+
+    log::debug!(
+        "[Health data] health: {health}, block_count: {block_count}, num_tipsets: {num_tipsets}"
+    );
```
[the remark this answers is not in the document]
creativcoder: done

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// epoch the node is currently at
+    epoch: i64,
+    /// base fee
+    base_fee: TokenAmount,
+    /// sync status information
+    sync_status: SyncStatus,
+}
+
+fn get_node_status(
+    chain_head: &Arc<Tipset>,
```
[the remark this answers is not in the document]
creativcoder: modified

forest/cli/src/cli/info_cmd.rs
creativcoder: done

forest/cli/src/cli/info_cmd.rs
creativcoder: removed

forest/cli/src/cli/info_cmd.rs
creativcoder: converted them to quickcheck tests.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    fn node_status_good() -> NodeStatusInfo {
+        super::NodeStatusInfo {
+            behind: Duration::from_secs(0),
+            health: 90.,
+            epoch: i64::MAX,
+            base_fee: TokenAmount::from_whole(1),
+            sync_status: SyncStatus::Ok,
+            start_time: Local::now(),
```
lemmih, line 275: It would be great if you didn't use non-deterministic code in the tests.

lemmih:
I dismissed my review, so you'll be able to merge this while I'm away.

elmattic:
Good, after your RPC calls optimisation it goes from 9 minutes to 4s:
```
$ time ./target/release/forest-cli info show    
Network: calibnet
Uptime: 0h 1m 90s (Started at: 2023-05-17 15:10:04.582378915 +02:00)
Chain: [sync: Ok! (34s behind)] [basefee: 100 atto FIL] [epoch: 566756]
Chain health: 69.30%


Default wallet address: -
./target/release/forest-cli info show  0.62s user 0.07s system 17% cpu 3.975 total
```

forest/cli/src/cli/info_cmd.rs
creativcoder: Using `DateTime::<Utc>::MIN_UTC`

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,376 @@
...
+            network: String::from("unknown"),
+            default_wallet_address: None,
+        })
+    }
+
+    fn display(&self, color: &LoggingColor) -> NodeInfoOutput {
+        let NodeStatusInfo { health, .. } = self;
+
+        let use_color = color.coloring_enabled();
+        let uptime_duration = Local::now() - self.start_time;
```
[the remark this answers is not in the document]
creativcoder: moved as param

elmattic:
```
$ ./target/release/forest-cli info show  
Network: calibnet
Uptime: 2h 162m 9739s (Started at: 2023-05-18 11:20:25.612908350 +02:00)
Chain: [sync: Ok! (44s behind)] [basefee: 100 atto FIL] [epoch: 569498]
Chain health: 76.28%


Default wallet address: t12orcrxghpr4saivecupbzvnqby35g6zwpbu54ia
```

Wondering why it shows 162 minutes and 9739s? Feels weird.

creativcoder:
> Wondering why it shows 162 minutes and 9739s? Feels weird.

Added formating to it, should be good now.

node/rpc-api/src/data_types.rs
```diff
@@ -40,6 +41,7 @@ where
     pub sync_state: Arc<SyncRwLock<SyncState>>,
     pub network_send: flume::Sender<NetworkMessage>,
     pub network_name: String,
+    pub start_time: chrono::DateTime<Local>,
```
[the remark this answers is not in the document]
creativcoder: I've changed it to Utc, and on the forest-cli side, i convert and present it as Local.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,391 @@
...
+            self.start_time.with_timezone(&chrono::offset::Local)
+        )
+        .normal();
+
+        let chain_status = self.chain_status().blue();
+        let network = self.network.green();
+        let wallet_address = self
+            .default_wallet_address
+            .clone()
+            .unwrap_or("-".to_string())
```
elmattic, line 141: I would use `"address not set"` here, instead of `"-"` to be more explicit to the user.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+impl InfoCommand {
+    pub async fn run(&self, config: Config, opts: &CliOpts) -> anyhow::Result<()> {
+        // uptime
+        let start_time = start_time(&config.client.rpc_token)
+            .await
+            .map_err(handle_rpc_err)?;
+
+        let network = chain_get_name((), &config.client.rpc_token)
```
elmattic, line 213: Our endpoint `ChainGetName` could be replaced by the one from the Filecoin api `Filecoin.StateGetNetworkParams`. This could be done in another PR though.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,391 @@
...
+        let fmt_uptime = fmt_duration(uptime);
+        let uptime = format!(
+            "{fmt_uptime} (Started at: {})",
+            self.start_time.with_timezone(&chrono::offset::Local)
+        )
+        .normal();
+
+        let chain_status = self.chain_status().blue();
+        let network = self.network.green();
+        let wallet_address = self
```
elmattic, line 138: Could be nice to add balance as well since we're here, like in lotus:

`fmt.Printf("      %s [%s]\n", addr.String(), types.FIL(balance).Short())`

elmattic (changes requested):

elmattic:
I was thinking that we kept the `NodeStatusInfo` inside of the info command, it's not really something that should be part of the RPC API.

Conceptually if we implement `Filecoin.NodeStatus` then `NodeStatusResult` should be of the type `NodeStatus` and not `NodeStatusInfo`.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,391 @@
...
+            .await
+            .map_err(handle_rpc_err)?;
+
+        let tipsets = chain_get_tipsets_finality((), &config.client.rpc_token)
+            .await
+            .map_err(handle_rpc_err)?;
+
+        let cur_duration_secs = SystemTime::now().duration_since(UNIX_EPOCH)?;
+
+        let mut node_status = NodeStatusInfo::new(
```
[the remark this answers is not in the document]
creativcoder: Added NodeStatus endpoint

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+        node_status_info.default_wallet_address_balance = default_wallet_address_balance;
+
+        let info_output = NodeInfoOutput::from(node_status_info).set_color(&opts.color);
+
+        println!("Network: {}", info_output.network);
+        println!("Uptime: {}", info_output.uptime);
+        println!("Chain: {}", info_output.chain_status);
+        println!("Chain health: {}", info_output.health);
+        println!(
+            "Default wallet address: {} [balance: {}]",
```
elmattic, line 252: This will results in an error if we don't have a default address.

elmattic (changes requested):

node/rpc-client/src/common_ops.rs
```diff
@@ -19,3 +19,7 @@ pub async fn shutdown(
 ) -> Result<ShutdownResult, Error> {
     call(SHUTDOWN, params, auth_token).await
 }
+
+pub async fn start_time(auth_token: &Option<String>) -> Result<StartTimeResult, Error> {
```
elmattic, line 23: Not a big deal but for automatic js binding to work we need to have two arguments for all of our endpoints:

So use prototype for XXX endpoint:
`(XXXParams, &Option<String>) -> Result<XXXResult, Error>`

elmattic (changes requested):

forest/cli/src/cli/info_cmd.rs
elmattic: You could print instead: `"Default wallet address: {}",`

And handle all the output formatting logic inside `NodeInfoOutput::from`.

And you could print `[balance: xxx]` string only if there is a default address set.

node/rpc-client/src/common_ops.rs
creativcoder: Added

forest/cli/src/cli/info_cmd.rs
creativcoder: Good catch, thx. Fixed.

elmattic (approved):

utils/forest_utils/src/misc/mod.rs
```diff
@@ -2,6 +2,46 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "auto" => Ok(LoggingColor::Auto),
+            "always" => Ok(LoggingColor::Always),
+            "never" => Ok(LoggingColor::Never),
+            _ => Err(Self::Err::msg(
+                "Invalid logging color output. Must be one of Auto, Always, Never",
+            )),
+        }
+    }
```
aatifsyed, line 46: strum?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+    pub start_time: DateTime<Utc>,
+    /// Current network the node is running on
+    pub network: String,
+    /// Default wallet address selected.
+    pub default_wallet_address: Option<String>,
+    /// Default wallet address balance
+    pub default_wallet_address_balance: Option<String>,
+}
+
+pub struct NodeInfoOutput {
```
aatifsyed, line 56: this seems a bit bloaty - can you think of any ways to slim it down?

aatifsyed: I've noticed https://docs.rs/colored/latest/colored/control/struct.ShouldColorize.html - should we be integrating this into our code instead of presumably writing this a bunch of times?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+            .await
+            .map_err(handle_rpc_err)?;
+
+        let network = chain_get_name((), &config.client.rpc_token)
+            .await
+            .map_err(handle_rpc_err)?;
+
+        let start_time = start_time(&config.client.rpc_token)
+            .await
+            .map_err(handle_rpc_err)?;
```
aatifsyed, line 218: This is ugly, can you refactor using `try_join` please

forest/cli/Cargo.toml
```diff
@@ -66,6 +69,7 @@ toml.workspace = true
 assert_cmd.workspace = true
 pretty_assertions = "1.3.0"
 quickcheck.workspace = true
+quickcheck_macros.workspace = true
```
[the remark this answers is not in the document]
creativcoder: Yes https://github.com/BurntSushi/quickcheck#the-quickcheck-attribute
> To use the #[quickcheck] attribute, you must import the quickcheck macro from the quickcheck_macros crate

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// sync status information
+    pub sync_status: SyncStatus,
+    /// Start time of the node
+    pub start_time: DateTime<Utc>,
+    /// Current network the node is running on
+    pub network: String,
+    /// Default wallet address selected.
+    pub default_wallet_address: Option<String>,
+    /// Default wallet address balance
+    pub default_wallet_address_balance: Option<String>,
```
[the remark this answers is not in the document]
creativcoder: Added context on base_fee from https://spec.filecoin.io/systems/filecoin_vm/gas_fee/

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+pub struct NodeInfoOutput {
+    pub chain_status: ColoredString,
+    pub network: ColoredString,
+    pub uptime: ColoredString,
+    pub health: ColoredString,
+    pub wallet_address: ColoredString,
+    pub wallet_balance: ColoredString,
+}
+
+#[derive(Debug, strum_macros::Display, PartialEq)]
```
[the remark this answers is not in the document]
creativcoder: Fixed

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+use crate::humantoken::TokenAmountPretty;
+
+#[derive(Debug, Subcommand)]
+pub enum InfoCommand {
+    Show,
+}
+
+#[derive(Debug)]
+pub struct NodeStatusInfo {
+    /// duration in seconds of how far behind the node is with respect to
```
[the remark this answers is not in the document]
creativcoder: true. removed `in seconds`

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+    let duration = duration.to_string();
+    let duration = duration.split(' ');
+    let format_duration = duration
+        .filter(|s| !s.ends_with("us"))
+        .filter(|s| !s.ends_with("ns"))
+        .filter(|s| !s.ends_with("ms"))
+        .map(|s| s.to_string());
+    let format_duration: Vec<String> = format_duration.collect();
+    format_duration.join(" ")
+}
```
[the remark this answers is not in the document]
creativcoder: It strips the format string from humantime::format_duration to exclude `us` `ns` `ms` parts, which would be overkill and noise to end user. Sad that format_duration doesn't have APIs to exclude those. `Duration` doesn't have APIs that would give me something like `2hr 30min 6secs 45ms ...` 

Open to alternatives.

utils/forest_utils/src/misc/mod.rs
creativcoder: yes

forest/cli/src/cli/info_cmd.rs
creativcoder: neat, thx.

forest/cli/src/cli/info_cmd.rs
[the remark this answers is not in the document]
creativcoder: Added a doc: `// Limits the format string from humantime::format_duration to exclude us ns ms parts.`

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    pub fn chain_status(&self) -> String {
+        let base_fee_fmt = self.base_fee.pretty();
+        let behind = format!("{}", humantime::format_duration(self.behind));
+        format!(
+            "[sync: {}! ({} behind)] [basefee: {base_fee_fmt}] [epoch: {}]",
+            self.sync_status, behind, self.epoch
+        )
+    }
```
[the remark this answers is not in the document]
creativcoder: Display impl on which struct?

forest/cli/src/cli/info_cmd.rs
creativcoder: Not sure about this. I wanted to have this for keeping individual formatting of NodeStatusInfo fields separate and also to aid in writing formatting tests against.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// sync status information
+    pub sync_status: SyncStatus,
+    /// Start time of the node
+    pub start_time: DateTime<Utc>,
+    /// Current network the node is running on
+    pub network: String,
+    /// Default wallet address selected.
+    pub default_wallet_address: Option<String>,
+    /// Default wallet address balance
+    pub default_wallet_address_balance: Option<String>,
```
aatifsyed: You've added documentation for `base_fee`, but my initial comment was that a lot of the documentation is superfluous, highlighting this struct, and can probably be removed.

If your reader is expected to know what base fee is already, then you can leave this out.

Basically every comment here (aside from `base_fee` now, just restates the variable name underneath.
Check out this change - what information has been lost? I'd argue none - the struct is already called `NodeInfo`, so we know that the information relates to a node.

```suggestion
    pub epoch: ChainEpoch,
    pub base_fee: TokenAmount,
    pub sync_status: SyncStatus,
    pub start_time: DateTime<Utc>,
    pub network: String,
    pub default_wallet_address: Option<String>,
    pub default_wallet_address_balance: Option<String>,
```

forest/cli/src/cli/info_cmd.rs
aatifsyed: > I wanted to have this for keeping individual formatting of NodeStatusInfo fields separate and also to aid in writing formatting tests against.

It's good to make sure that your code is testable, but if you're writing colored-string aware tests, that might be the sign that the test approach is wrong.

Think of other aspects of UI - do we write progress bar tests? How about testing log output? These are all out of scope for our codebase - we rely on libraries for these. Test your data layer, but (save for user-reported bugs), presentation layer tests aren't really _interesting_

Our tests should cover our logic, and basic implementation details.

I think we should move the colored printing to `run`, keep the `NodeStatusInfo::new` tests, and remove presentation layer tests

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+    let format_duration: Vec<String> = format_duration.collect();
+    format_duration.join(" ")
+}
+
+impl NodeStatusInfo {
+    pub fn new(
+        cur_duration: Duration,
+        blocks_per_tipset_last_finality: f64,
+        head: TipsetJson,
+    ) -> anyhow::Result<NodeStatusInfo> {
```
aatifsyed, line 90:
```suggestion
    ) -> Option<NodeStatusInfo> {
```

Not an appropriate error type here - there's only one reason this can fail, so document it.

Also please document what `blocks_per_tipset_last_finality` means, it's not clear from the name

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+        let health = 100. * blocks_per_tipset_last_finality / BLOCKS_PER_EPOCH as f64;
+
+        Ok(Self {
+            behind: Duration::from_secs(behind),
+            health,
+            epoch: head.0.epoch(),
+            base_fee,
+            sync_status,
+            start_time: Utc::now(),
+            network: String::from("unknown"),
```
aatifsyed, line 122: This seems wrong - why even have the field if it's only ever set to `unknown`?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        let health = 100. * blocks_per_tipset_last_finality / BLOCKS_PER_EPOCH as f64;
+
+        Ok(Self {
+            behind: Duration::from_secs(behind),
+            health,
+            epoch: head.0.epoch(),
+            base_fee,
+            sync_status,
+            start_time: Utc::now(),
```
aatifsyed, line 121: This also seems wrong - I expect `start time` to be whenever the node started running. Think of `uptime`. Either give it a new name, or remove - does the user really need to know the time they ran a command?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+
+#[derive(Debug, Subcommand)]
+pub enum InfoCommand {
+    Show,
+}
+
+#[derive(Debug)]
+pub struct NodeStatusInfo {
+    /// How far behind the node is with respect to syncing to head
+    pub behind: Duration,
```
aatifsyed, line 33:
```suggestion
    pub lag: Duration,
```

forest/cli/src/cli/info_cmd.rs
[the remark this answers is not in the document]
aatifsyed: `NodeStatusInfo`

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+        )
+    }
+}
+
+impl From<NodeStatusInfo> for NodeInfoOutput {
+    fn from(node_status_info: NodeStatusInfo) -> NodeInfoOutput {
+        let health = node_status_info.health;
+        let uptime = (Utc::now() - node_status_info.start_time)
+            .to_std()
+            .expect("failed converting to std duration");
```
aatifsyed, line 143: `From` impls should *not* be nondeterministic

aatifsyed (changes requested):

forest/cli/src/cli/info_cmd.rs
lemmih: In addition to integration tests, we can and should write unit tests for this code. We have an algorithm for computing the network health and a pretty-printer for formatting the status of a node. Both functions are pure and should be relatively easy to test.

I don't see how `NodeInfoOutput` relates to testing, though. As I said in a previous review, the code would be smaller and easier to test without that struture.

forest/cli/src/cli/info_cmd.rs
creativcoder: Removed superflous docs where applicable.

forest/cli/src/cli/info_cmd.rs
creativcoder: It's an `Option<..>` now.

forest/cli/src/cli/info_cmd.rs
creativcoder: This is an `Option<..>` now. It's needed to calculate uptime.

forest/cli/src/cli/info_cmd.rs
creativcoder: Removed

forest/cli/src/cli/info_cmd.rs
creativcoder: I agree, but i'd like to keep it this way for this instance, even though there's only one `Result` type here as it makes things a bit convenient within `run` function. Also, shouldn't we be reporting the error: `"System time should not be behind tipset timestamp, please sync the system clock."` to the user instead of just returning a `None`?

documented `blocks_per_tipset_last_finality` at usage.

forest/cli/src/cli/info_cmd.rs
creativcoder: Removed NodeInfoOutput, fixed presentation layer unit tests and added integration tests, but the integration tests fails (https://github.com/ChainSafe/forest/actions/runs/5196426787/jobs/9370116138#step:8:1585), because it needs the daemon to be running. Not sure if we still want to have integration test for this part of forest-cli.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,402 @@
...
+    /// but averaging at 5 is considered healthy.
+    pub health: f64,
+    /// epoch the node is currently at
+    pub epoch: ChainEpoch,
+    /// Base fee is the set price per unit of gas (measured in attoFIL/gas unit) to be burned (sent to an unrecoverable address) for every message execution
+    pub base_fee: TokenAmount,
+    pub sync_status: SyncStatus,
+    /// Start time of the node
+    pub start_time: Option<DateTime<Utc>>,
+    pub network: Option<String>,
```
lemmih, line 46: Logically, we always have a start time and a network name. You can refactor the code to reflect this.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,402 @@
...
+            "[sync: {}! ({} behind)] [basefee: {base_fee_fmt}] [epoch: {}]",
+            self.sync_status, behind, self.epoch
+        )
+        .blue();
+
+        if !self.use_color {
+            chain_status.clear()
+        } else {
+            chain_status
+        }
```
lemmih, line 66: You repeat this snippet of code six times. Let's refactor such that we only clear the color once.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,402 @@
...
+        use_color: bool,
+    ) -> anyhow::Result<NodeStatusInfo> {
+        let ts = head.0.min_timestamp();
+        let cur_duration_secs = cur_duration.as_secs();
+        let lag = if ts <= cur_duration_secs + 1 {
+            cur_duration_secs.saturating_sub(ts)
+        } else {
+            anyhow::bail!(
+                "System time should not be behind tipset timestamp, please sync the system clock."
+            );
```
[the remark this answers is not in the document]
creativcoder: What issues do we foresee with this bail in future? imo this is ok to have; it's a code in a binary crate and we should crash and report error if there's a failure case out of our control; no?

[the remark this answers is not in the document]
lemmih: A misconfigured clock is not an error in the same way that, say, failing to connect to the backend is an error. Forest can still run with a misconfigured clock, and it'll still have a status we can print out.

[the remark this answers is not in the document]
creativcoder: Added SyncStatus::Fast and changes accordingly.

forest/cli/src/cli/info_cmd.rs
creativcoder: Fixed

forest/cli/src/cli/info_cmd.rs
creativcoder: Fixed

forest/cli/src/cli/info_cmd.rs
lemmih: This function can no longer fail.

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,447 @@
...
+            chain_get_name((), &config.client.rpc_token),
+            start_time(&config.client.rpc_token)
+        );
+        match res {
+            Ok((node_status, head, network, start_time)) => {
+                let cur_duration: Duration = SystemTime::now().duration_since(UNIX_EPOCH)?;
+                let blocks_per_tipset_last_finality =
+                    node_status.chain_status.blocks_per_tipset_last_finality;
+
+                let mut node_status_info = NodeStatusInfo::new(
```
lemmih, line 210: This doesn't need to be `mut` if you refactor it slightly.

lemmih (changes requested):

forest/cli/src/cli/info_cmd.rs
creativcoder: I get https://github.com/ChainSafe/forest/actions/runs/5256500061/jobs/9497916282?pr=2578#step:6:1508 when moving wallet and balance fields as params

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,447 @@
...
+            writeln!(&mut output, "Uptime: {}", self.uptime(now).clear())?;
+            writeln!(&mut output, "Chain: {}", self.chain_status().clear())?;
+            writeln!(&mut output, "Chain health: {}", self.health().clear())?;
+            writeln!(
+                &mut output,
+                "Default wallet address: {} {}",
+                self.wallet_address().clear(),
+                self.wallet_balance().clear()
+            )?;
+        }
```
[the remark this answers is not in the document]
creativcoder: done

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,453 @@
...
+            sync_status,
+            start_time,
+            network,
+            use_color,
+            default_wallet_address,
+            default_wallet_address_balance,
+        }
+    }
+
+    fn format(&self, now: DateTime<Utc>) -> anyhow::Result<String> {
```
lemmih, line 168: Ideally, this function should be infallible.

lemmih (dismissed):

forest/cli/src/cli/info_cmd.rs
creativcoder: refactored to remove `use_color` field

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,385 @@
...
+            )
+        };
+
+        output.push(network);
+        output.push(uptime);
+        output.push(chain);
+        output.push(chain_health);
+        output.push(wallet_info);
+
+        output
```
aatifsyed, line 160: Just construct a vec with `vec![..]` here.
It's surprising to return `Vec<String>` - could you add a comment explaining why?

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+fn balance(balance: &Option<String>) -> anyhow::Result<String> {
+    if let Some(bal) = balance {
+        let balance_token_amount = TokenAmount::from_atto(bal.parse::<BigInt>()?);
+        Ok(format!("{:.4}", balance_token_amount.pretty()))
+    } else {
+        Err(anyhow::anyhow!("could not find balance"))
+    }
+}
```
aatifsyed, line 218: It's very odd to have a function take in a single `&Option<T>` - just make the inner function, and call one of the `Option` combinators on the outer option

aatifsyed (approved):
nits

forest/cli/src/cli/info_cmd.rs
aatifsyed: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8131fccabf844bba1255f586298b2ac7

Of course, in this example, the mapped function is trivial and should be inlined.

forest/cli/src/cli/info_cmd.rs
creativcoder: was needed for color strings fmt, but not needed atm. Changed it to a `String`

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,370 @@
...
+            let wallet_balance = match balance(&self.default_wallet_address_balance) {
+                Ok(bal) => format!("[balance: {}]", bal),
+                Err(e) => e.to_string(),
+            };
+
+            format!(
+                "Default wallet address: {} [{}]",
+                wallet_address, wallet_balance
+            )
+        };
```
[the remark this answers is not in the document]
aatifsyed: I'm not going to hold up the PR on this, because you'll be doing follow on work, and this has been open a while anyway

forest/cli/src/cli/info_cmd.rs
```diff
@@ -0,0 +1,374 @@
...
+        let status = node_status(duration, tipset.as_ref());
+
+        let expected_status_fmt =
+            "[sync: Behind! (8h 20m behind)] [basefee: 0 FIL] [epoch: 0]".to_string();
+        assert!(status
+            .format(DateTime::<chrono::Utc>::MIN_UTC)
+            .contains(&expected_status_fmt));
+    }
+
+    // FIXME: enable/debug this failing when color output is supported.
```
lemmih, line 337: I'd greatly prefer it if this uncommented code was removed before merging.

forest/cli/src/cli/info_cmd.rs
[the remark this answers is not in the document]
creativcoder: Noted. Thanks for the detailed review. We did decide to refactor RPC APIs to return strongly typed values, that'll be part of the follow-up PR and also accordingly the wallet fmt code.

forest/cli/src/cli/info_cmd.rs
creativcoder: removed the commented test

Issue #2143, opened 2022-10-28
Consume `quickcheck` implementation in upstream crates


connormullett:
**Issue summary**

Several quickcheck implementations have been pushed upstream to their respective crates and may soon be available for use. See tracking issue #2130. Once new versions have been released, we should get rid of our custom quickcheck implementations. The types and crates in question are:

- [x] `Cid`: https://github.com/multiformats/rust-cid
- [x] `Ipld`: https://crates.io/crates/libipld-core
- [x] `ActorState`: https://crates.io/crates/fvm
- [x] `Address`, `PoStProof`, `Message`, `Signature`, `SignatureType`: https://crates.io/crates/fvm_shared

**Other information and links**

lemmih:
This issue can also track the `Arbitrary` implementations in `rust-cid`, `libipld-core`, and `fvm`.

lemmih:
Blocked until new versions of `rust-cid`, `fvm`, `libipld-core`, and `fvm_shared` have been released.

lemmih:
New version of `rust-cid` has been released. @tyshko5

lemmih:
@jdjaustin Unblocked now that we're using `cid-0.10` and `libipld-core-0.16`.

PR #2590, opened 2023-02-23
Consume upstream `quickcheck` implementation for `Address`

jdjaustin:
## Summary of changes

- Remove custom quickcheck implementation for `Address` and instead use the implementation in `fvm_shared`.

N/A

## Change checklist


utils/json/src/address.rs
```diff
@@ -124,29 +124,11 @@ mod tests {
...
-        }
-    }
-
     #[quickcheck]
-    fn address_roundtrip(address: AddressWrapper) {
-        let serialized = serde_json::to_string(&json::AddressJsonRef(&address.address)).unwrap();
-        let parsed: json::AddressJson = serde_json::from_str(&serialized).unwrap();
-        assert_eq!(address.address, parsed.0);
+    fn address_roundtrip(id: u64) {
+        let address = Address::new_id(id);
```
lemmih, line 129: `Address_v3` _does_ implement `Arbitrary` so you can use it directly:

```suggestion
    fn address_roundtrip(address: Address_v3) {
        let address: Address = address.into();
```

Now, you shouldn't really do this. It would be nicer if you implement `Arbitrary` for `Address` instead.

lemmih (changes requested):

utils/json/src/address.rs
```diff
@@ -119,34 +119,21 @@ pub mod json {
...
-        }
-    }
-
     #[quickcheck]
-    fn address_roundtrip(address: AddressWrapper) {
-        let serialized = serde_json::to_string(&json::AddressJsonRef(&address.address)).unwrap();
-        let parsed: json::AddressJson = serde_json::from_str(&serialized).unwrap();
-        assert_eq!(address.address, parsed.0);
+    fn address_roundtrip(address: Address_v3) {
+        let address = address.into();
```
lemmih, line 130: Let's implement `Arbitrary` for `Address`. Then we can just write this:

```suggestion
    fn address_roundtrip(address: Address) {
```

lemmih (changes requested):

utils/forest_shim/src/address.rs
```diff
@@ -45,6 +45,12 @@ impl Address {
     }
 }
 
+impl quickcheck::Arbitrary for Address {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        Address(Address_v3::new_id(u64::arbitrary(g)))
```
lemmih, line 50: This is not right. Give some thought to what this function should be doing and ping me in slack if you get stuck.

lemmih (changes requested):

utils/forest_shim/src/address.rs
jdjaustin: The way I understand it, we want to create an `Address(Address_v3)`. So it seems like my options would be this, `Address::from(Address_v3::new_id(u64::arbitrary(g)))`, or `Address::new_id(u64::arbitrary(g))`.

utils/json/src/address.rs
```diff
@@ -119,34 +119,20 @@ pub mod json {
...
     #[quickcheck]
-    fn address_roundtrip(address: AddressWrapper) {
-        let serialized = serde_json::to_string(&json::AddressJsonRef(&address.address)).unwrap();
-        let parsed: json::AddressJson = serde_json::from_str(&serialized).unwrap();
-        assert_eq!(address.address, parsed.0);
+    fn address_roundtrip(address: Address) {
+        let serialized = forest_test_utils::to_string_with!(&address, json::serialize);
+        let parsed = forest_test_utils::from_str_with!(&serialized, json::deserialize);
+        // Skip delegated addresses for now
+        match address.payload() {
```
lemmih, line 133: Might be better to use `address.protocol()` here.

jdjaustin:
CI checks failing: need to update `Arbitrary` implementation for `ActorState` in `ref-fvm`. `ActorState` `address` field was changed to `delegated_address` but `Arbitrary` implementation still uses old field `address`.

lemmih (approved):
Yay, `AddressWrapper` is gone. :)

LesnyRumcajs (approved):

hanabi1224 (approved):

PR #2591, opened 2023-02-23
Consume upstream `quickcheck` implementation for `Message` and `SignedMessage`

jdjaustin:
## Summary of changes

- Remove custom quickcheck implementation for `Message` and `SignedMessage` and instead use the implementation in `fvm_shared`.

N/A

## Change checklist


hanabi1224 (approved):

utils/json/src/message.rs
```diff
@@ -131,43 +131,22 @@ pub mod json {
...
-        }
-    }
+    use crate::message::json;
 
     #[quickcheck]
-    fn message_roundtrip(message: MessageWrapper) {
-        let serialized = serde_json::to_string(&MessageJsonRef(&message.message)).unwrap();
-        let parsed: MessageJson = serde_json::from_str(&serialized).unwrap();
-        assert_eq!(message.message, parsed.0);
+    fn message_roundtrip(id: u64) {
```
lemmih, line 141: You can get rid of all the cruft. It should look something like this:

```suggestion
    fn message_roundtrip(message: Message) {
```

lemmih (changes requested):

utils/json/src/message.rs
jdjaustin: I tried that approach, but the compiler gave me this error (and a similar error if I modify the test use declaration to `use fvm_shared::message::Message`; in that case the compiler also complains about mismatched types in the `json` serialization). So I used the `TokenAmount` [quickcheck test](https://github.com/ChainSafe/forest/blob/ec2881094dad512f9cc90868af2a5c43ffd60fcb/utils/json/src/token_amount.rs#L66) as a guide. I'm not sure how to use `Message` directly at this point and could use some guidance.
Screen Shot 2023-02-24 at 9 12 01 AM

jdjaustin: I made the changes but now this quickcheck is also failing on a serialization error ("BigInt too large"). I'll push the changes.

jdjaustin: Interesting, the CI check failed with a different message. This is the failure I see when I run the test locally:
Screen Shot 2023-02-24 at 2 14 35 PM

jdjaustin: As this test passes before the changes, it seems like it could be an error in the test logic. However, it also doesn't seem to make sense why it would pass with the `MessageWrapper` and not without. Although looking at the `Arbitrary` implementation for `Message` in ref-fvm, it apperas the issue could be how the `TokenAmount` values are generated. The forest `Arbitrary` uses `TokenAmount::from_atto(u64::arbitrary(g)).into()`, whereas the ref-fvm implementation uses `TokenAmount::arbitrary(g)`. There is a `check` function for `Message` that indicates these values should not exceed `i64::MAX as u64`. The method used in the forest implementation would definitely limit the `TokenAmount` values.

jdjaustin: That does seem to be the issue. I created a [branch](https://github.com/ChainSafe/ref-fvm/tree/jdjaustin-branch) that modifies the `Arbitrary` implementation and the error has been resolved with these changes.

lemmih: > That does seem to be the issue. I created a [branch](https://github.com/ChainSafe/ref-fvm/tree/jdjaustin-branch) that modifies the `Arbitrary` implementation and the error has been resolved with these changes.

"Fixing" the issue is less important than understanding the issue. Do you understand why making the `TokenAmount` smaller "fixed" the test? And what is the size limit for a `TokenAmount`?

jdjaustin: IIUC `TokenAmount` is a wrapper for `BigInt`, so the size limit for `TokenAmount` should be the same as `BigInt`, which I found [here](https://stackoverflow.com/questions/50504503/is-there-a-limit-to-the-size-of-a-bigint-or-biguint-in-rust). However, it appears the limitation for serialization/deserialization is actually defined [here](https://github.com/filecoin-project/ref-fvm/blob/dd79a13a6d422be28f550a0d8569b869c69c52ea/shared/src/bigint/mod.rs):

```
/// The maximum number of bytes we accept to serialize/deserialize for a single BigInt.
pub const MAX_BIGINT_SIZE: usize = 128;
```

So for serialization/deserialization, we need to limit the arbitrary generation of `g` in `TokenAmount` to 128 bits.

lemmih: Bits or bytes?

lemmih: Good, so this limitation is not strictly speaking inherent to `TokenAmount` but rather a limitation on the serialization. What would it look like if you were to update the test with this limitation in mind (ie that serialization of large token amounts isn't valid)?

jdjaustin: It appears to be bytes. The `BigInt` serialization in `fvm_shared` is checking to make sure the length of the byte representation of the `BigInt` (plus a leading positive sign byte) is less than `MAX_BIGINT_SIZE`. I have added similar logic to the arbitrary generation of the `BigInt` in the `TokenAmount::arbitrary` implementation on my branch, and similar to #2590, I had to add logic to the `Message` implementation to exclude addresses using the `Delegated` protocol as well.

jdjaustin: What is the best approach to get feedback on the changes to my branch and incorporate them into our current `fvm_shared3` working branch? Should I open a PR over at `filecoin-project/ref-fvm`?

lemmih: Open a PR in https://github.com/ChainSafe/ref-fvm against `lemmih-wibbles`.

jdjaustin: I've changed this PR to draft until the PR to update the `TokenAmount` `quickcheck::Arbitrary` implementation is reviewed and merged (ready to review now): https://github.com/ChainSafe/ref-fvm/pull/1

lemmih (approved):

Issue #2458, opened 2023-01-20
Verify param files that are downloaded via ipfs http gateway


hanabi1224:
**Issue summary**

Currently, a forest daemon fetches param files via ipfs http gateway without content verification, while IPFS http gateway now offers a trustless mode ([SPEC](https://github.com/ipfs/specs/blob/main/http-gateways/TRUSTLESS_GATEWAY.md)) in which content is downloaded in `car` format so that a client can verify the integrity of the downloaded content.  

And here is a [web demo](https://github.com/2color/verified-ipfs-retrieval).

Related forest code: https://github.com/ChainSafe/forest/blob/main/utils/paramfetch/src/lib.rs#L20

**Other information and links**

https://github.com/ipfs/specs/blob/main/http-gateways/TRUSTLESS_GATEWAY.md
https://github.com/2color/verified-ipfs-retrieval

MuhtasimTanmoy:
I would like to work in this issue.

dapplion:
Just published `rs-car` as first building block for this issue https://crates.io/crates/rs-car

I found on crates that there's a brand new implementation for CAR format in https://github.com/blocklessnetwork/rust-car will look into it too

dapplion:
@hanabi1224 What's the usage pattern of paramfetch crate?
- How often are files downloaded (on start-up, on network event, on interval)
- How many files are downloaded?
- What's the size distribution of those files?

hanabi1224:
@dapplion 
> How often are files downloaded (on start-up, on network event, on interval)
only for the first run

> How many files are downloaded?
23 files (as of writing)

> What's the size distribution of those files?
1 file at ~293MB, 2 at ~2MB, others at <1MB

dapplion:
Ok so it would be ok to buffer in memory? I assume the eventual memory required by the node after startup is more than ~500MB

LesnyRumcajs:
Re-opening. This caused hash mismatches that unfortunately passed the CI so I had to revert https://github.com/ChainSafe/forest/commit/2b5515d29834be56816fb5447d7a57abb977da90

https://github.com/ChainSafe/forest/actions/runs/4374624518/jobs/7654288564
```
[...]

PR #2604, opened 2023-02-27
Read data as trustless gateway with car stream

dapplion:
## Summary of changes

- Download files from IPFS gateway using a car stream to guarantee the integrity of the data. Uses `rs-car-ipfs` library which constructs the file's block DAG in memory and writes to and out AsyncWrite when all nodes are available. This strategy should be sufficient as long as the downloaded files are relatively small.

Closes https://github.com/ChainSafe/forest/issues/2458

## Notes to mantainers
**Gateway safety**

Gateways usually return the content of the file as-is. To ask the gateway to return a CAR stream one must add the query `?format=car`. Since the query is optional old gateways will just return the content as-is which will fail to decode as car.

The trustless gateway feature was shipped a while ago so most providers will support it by now. To test is a gateway supports car streams you can pull the example readme and expect to see binary output
```
curl https://ipfs.io/ipfs/QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc?format=car 
```

To test that a specific file can be verified with the underlying car library you can install it as a binary and do
```
cargo install rs-car-ipfs --features bin
curl https://ipfs.io/ipfs/QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc?format=car | car-ipfs
```

**Cargo.lock diff**

The libraries `rs-car` and `rs-car-ipfs` depend on some building blocks which the forest dependencies already use like `cid` or `multihash`. Since the dependent versions are different that's the cause of the diff. **Question**: Should I depend on a more flexible version of those deps in my library to allow cargo to resolve it to the same? On proper semver notation `0.8.*` and `0.10.*` and major jumps. Otherwise would you recommend me to downgrade the libraries to using the same set of versions than forest?

## Change checklist


utils/paramfetch/Cargo.toml
```diff
@@ -11,10 +11,12 @@ ahash.workspace = true
 anyhow.workspace = true
 backoff = { workspace = true, features = ["tokio"] }
 blake2b_simd.workspace = true
+cid = "0.10.1"
```
hanabi1224, line 14: Better to re-export `cid` from `rs-car` in case there is a version mismatch

utils/paramfetch/src/lib.rs
```diff
@@ -20,14 +22,16 @@ use tokio::{
...
+use tokio_util::compat::TokioAsyncWriteCompatExt;
 
 const GATEWAY: &str = "https://proofs.filecoin.io/ipfs/";
 const PARAM_DIR: &str = "filecoin-proof-parameters";
 const DIR_ENV: &str = "FIL_PROOFS_PARAMETER_CACHE";
 const GATEWAY_ENV: &str = "IPFS_GATEWAY";
 const TRUST_PARAMS_ENV: &str = "TRUST_PARAMS";
 const DEFAULT_PARAMETERS: &str = include_str!("parameters.json");
+// Arbitrary limit at 10MB for DOS protection
+const MAX_PARAM_FILE_SIZE: usize = 10_000_000;
```
hanabi1224, line 34: One the files is ~280MB,  10MB might not be sufficient here

dapplion:
Previous implementation piped the HTTP response body to a file, since all bytes are received sequentially thus memory footprint is negligible.

In a CAR stream that may not be the case, since IPLD blocks are de-duplicated. If a latter node in the DAG tree references a previous block of data repeated appearances will not be streamed. If one wants to keep the memory footprint of a CAR streamer low it must have the capability to do random access on any previously streamed data. The naive implementation is to buffer in memory on a hash-map. A better design is possible but requires more engineering complexity. It's doable but would like to know it's really necessary since this issue already took decent involvement.

hanabi1224:
@dapplion Memory usage is not a big deal in the forest-fetch-param case. 

As a side note, if you'd like to extend your lib to support very large files with low memory usage, maybe you could use [parity-db](https://github.com/paritytech/parity-db) which is essentially a hash map on disk

utils/paramfetch/Cargo.toml
dapplion: Done

utils/paramfetch/src/lib.rs
dapplion: Constant removed

dapplion:
@hanabi1224 Switched to `read_single_file_seek` which does not buffer data blocks in memory. If there's a need to recover a previous block it fetches it from the target file. See this comment for reference https://github.com/dapplion/rs-car-ipfs/blob/60a6423f27445315e95c43f8286dbd7f74d4ed6c/src/single_file/single_file_seek.rs#L1-L36

hanabi1224:
@dapplion thanks! LGTM. Will approve once u resolve the merge conflict and have CI passing

Cargo.lock
```diff
@@ -1305,10 +1305,23 @@ checksum = "f6ed9c8b2d17acb8110c46f1da5bf4a696d745e1474a16db0cd2b49cd0249bf2"
...
- "multihash",
+ "multihash 0.16.3",
  "serde",
  "serde_bytes",
- "unsigned-varint",
+ "unsigned-varint 0.7.1",
+]
+
+[[package]]
+name = "cid"
```
hanabi1224, line 1315: @dapplion After thinking about it, I think it would be better if you could relax `cid` version in `rs-car` by using `cid = ">=0.8, <0.11"` in `Cargo.toml`, that way it will automatically resolve to the same version forest uses

dapplion: Looked into it and `libipld-core` depends on cid 0.10.0. That library contains a cbor decoder which may return Cid data structures from the car stream header. libipld-core v0.14.0.

The API of libipld-core v0.14.0 and v0.16.0 (latest that requires cid v0.10.0) is incompatible. The same source can work for both versions.

So I don't see an easy way to support the relaxed range you suggest

hanabi1224: That makes sense. Thanks for investigating

dapplion:
@hanabi1224 downgraded the set of deps of rs-car to match this repo. Once you bump to more recent cid and multihash dependencies I can re-bump too

hanabi1224:
@dapplion looks like `rs-car` introduces many unnecessary indirect dependencies, I made a PR to clean them up, please take a look  https://github.com/dapplion/rs-car/pull/8

Cargo.lock
```diff
@@ -6549,6 +6594,17 @@ dependencies = [
  "data-encoding-macro",
 ]
 
+[[package]]
+name = "multicodec"
```
hanabi1224, line 6598: This crate looks stale

Cargo.lock
```diff
@@ -7616,6 +7672,16 @@ dependencies = [
  "prost-derive 0.9.0",
 ]
 
+[[package]]
+name = "prost"
```
hanabi1224, line 7676: This should not be needed

Cargo.lock
```diff
@@ -7626,6 +7692,28 @@ dependencies = [
  "prost-derive 0.11.8",
 ]
 
+[[package]]
+name = "prost-build"
```
hanabi1224, line 7696: This should not be needed

Cargo.lock
```diff
@@ -7798,6 +7909,15 @@ version = "1.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
 
+[[package]]
+name = "quick-protobuf"
```
hanabi1224, line 7913: This should not be needed

dapplion:
> @dapplion looks like `rs-car` introduces many unnecessary indirect dependencies, I did some investigation and made a PR to clean them up, please take a look [dapplion/rs-car#8](https://github.com/dapplion/rs-car/pull/8)

Thank you for taking a look! Github Actions is currently offline, will merge and release once service is back up

dapplion:
@hanabi1224 upstreamed your changes, lock file looks good now. Did a pass with @ec2 today on both libraries he :+1:

hanabi1224 (approved):
Nice work, thanks!

dapplion:
Hey @LesnyRumcajs could you provide more details on the error that this PR caused so I can look into it? Including:
- How to reproduce?
- Can you link to the code that does the integrity check?
- Can you link to the set of files / or IPFS hashes so I can download locally?

elmattic:
Hey @dapplion 

I had those errors when trying to sync from scratch. So

1. Make sure you delete the whole Forest directory before.
2. Import a new snapshot calibnet.
3. Run the node.

LesnyRumcajs:
You only need to remove the `filecoin-proof-parameters` directory, no?

elmattic:
I think so, yes. But in my case I was just describing how I have encountered those errors.

PR #2616, opened 2023-02-28
Remove forest-specific dependencies from the actor-interface crate

### LINKED ISSUE #2575: Remove forest-specific dependencies from the `actor-interface` crate
**Issue summary**

The `actor-interface` crate is merely a wrapper around the different versions of the builtin-actors that live in `fil-actor-states`. As such, the `actor-interface` should also live in the same repository. However, we can't move the crate before all Forest specific dependencies have been resolved.

Right now, the `actor-interface` depends on `forest_json`, `forest_shim`, and `forest_utils`. Each of these dependencies have to be removed and it's not always obvious how to do so.

- [ ] Remove `forest_utils`. This should be fairly straightforward to remove since the crate is only using the `BlockstoreExt` convenience trait.
- [ ] Remove `forest_json`. We previously maintained a very specific JSON format for our RPC but we've since relaxed this restriction. Furthermore, if we _do_ need a specifc JSON format, this should be defined in Forest proper and not in the `actor-interface` crate. As such, let's just remove the JSON specific serialization and replace it with the default `serde` behavior.
- [ ] Remove `forest_shim`. This will be the most difficult dependency to remove and I'm not entirely sure how to deal with this correctly in the future.

Success criteria:
- [ ] The `actor-interface` crate is moved to the `fil-actor-states` repository.
- [ ] Forest still works.

https://github.com/ChainSafe/fil-actor-states

sudo-shashank:
## Summary of changes

- [X]  Remove `forest_utils`.
- [X] Remove `forest_json`.
- [X] Remove `forest_shim`.

Closes #2575 

## Change checklist


vm/actor_interface/src/lib.rs
```diff
@@ -3,4 +3,25 @@
...
 
+use cid::Cid;
+use fvm_ipld_blockstore::Blockstore;
+use fvm_ipld_encoding::from_slice;
+use serde::de::DeserializeOwned;
+
 pub use self::builtin::*;
+
+/// Extension methods for inserting and retrieving IPLD data with CIDs
+pub trait BlockstoreExt: Blockstore {
```
lemmih, line 14: `BlockstoreExt` does not belong here. You _could_ make it part of the library but it definitely shouldn't be public. I think the best solution would be to not include it at all.

lemmih (changes requested):

vm/actor_interface/src/lib.rs
sudo-shashank: removed

node/rpc/src/state_api.rs
```diff
@@ -123,9 +124,9 @@ pub(crate) async fn state_market_deals<
     let ts = data.chain_store.tipset_from_keys(&tsk)?;
     let actor = data
         .state_manager
-        .get_actor(&market::ADDRESS, *ts.parent_state())?
+        .get_actor(&Address::new_id(5), *ts.parent_state())?
```
lemmih, line 127: Let's keep using human readable addresses.

vm/actor_interface/src/builtin/account/mod.rs
```diff
@@ -60,19 +60,25 @@ impl State {
     {
         if is_v8_account_cid(&actor.code) {
             return store
-                .get_obj(&actor.state)?
+                .get(&actor.state)?
+                .map(|bz| from_slice(&bz))
+                .transpose()?
```
lemmih, line 65: Let's refactor this to a separate function.

vm/interpreter/src/vm.rs
```diff
@@ -43,6 +43,10 @@ use crate::{fvm::ForestExternsV2, fvm3::ForestExterns as ForestExterns_v3};
 pub(crate) type ForestMachine<DB> = DefaultMachine<DB, ForestExternsV2<DB>>;
 pub(crate) type ForestMachineV3<DB> = DefaultMachine_v3<DB, ForestExterns_v3<DB>>;
 
+const SYSTEM_ADDR: Address = Address::new_id(0);
+const CRON_ADDR: Address = Address::new_id(3);
+const REWARD_ADDR: Address = Address::new_id(2);
```
lemmih, line 48: I'm not against hoisting these out of the actor interface but this is not the place for them. If you want to copy the addresses, put them in `forest_shim`.

lemmih (changes requested):

node/rpc/src/state_api.rs
sudo-shashank: fixed

vm/actor_interface/src/builtin/account/mod.rs
sudo-shashank: done

vm/interpreter/src/vm.rs
sudo-shashank: moved to forest_shim

utils/forest_shim/src/consts.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use crate::address::Address;
+
+pub const SYSTEM_ADDR: Address = Address::new_id(0);
+pub const CRON_ADDR: Address = Address::new_id(3);
+pub const REWARD_ADDR: Address = Address::new_id(2);
```
lemmih, line 8: Let's copy all of them or none of them. And you can put them in the address namespace. Eg: `Address::SYSTEM_ACTOR`.

lemmih (changes requested):

utils/forest_shim/src/consts.rs
sudo-shashank: okay

sudo-shashank: fixed

node/rpc/src/state_api.rs
```diff
@@ -123,9 +123,9 @@ pub(crate) async fn state_market_deals<
     let ts = data.chain_store.tipset_from_keys(&tsk)?;
     let actor = data
         .state_manager
-        .get_actor(&market::ADDRESS, *ts.parent_state())?
+        .get_actor(&market::ADDRESS.into(), *ts.parent_state())?
```
lemmih, line 126: We can use the new constants, right?

```suggestion
        .get_actor(&Address::MARKET_ACTOR, *ts.parent_state())?
```

lemmih (changes requested):

node/rpc/src/state_api.rs
sudo-shashank: fixed

lemmih (approved):

PR #2623, opened 2023-03-01
Consume upstream `quickcheck` implementation for `Signature` and `SignatureType`

jdjaustin:
## Summary of changes

- Remove custom quickcheck implementation for `Signature` and `SignatureType` and instead use the implementation in `fvm_shared`.

N/A

## Change checklist


utils/json/src/signature.rs
```diff
@@ -119,46 +119,22 @@ mod tests {
...
+        assert_eq!(signature, parsed);
     }
 
     #[quickcheck]
-    fn signaturetype_roundtrip(sigtype: SignatureTypeWrapper) {
-        let serialized = serde_json::to_string(&SignatureTypeJson(sigtype.sigtype)).unwrap();
-        let parsed: SignatureTypeJson = serde_json::from_str(&serialized).unwrap();
-        assert_eq!(sigtype.sigtype, parsed.0);
+    fn signaturetype_roundtrip(sigtype: SignatureType) {
+        let serialized1: String =
```
lemmih, line 132: Why rename `serialized` to `serialized1`?

utils/json/src/signature.rs
```diff
@@ -119,46 +119,22 @@ mod tests {
...
-        }
-    }
-
     #[quickcheck]
-    fn signature_roundtrip(signature: SignatureWrapper) {
-        let serialized = serde_json::to_string(&SignatureJsonRef(&signature.signature)).unwrap();
-        let parsed: SignatureJson = serde_json::from_str(&serialized).unwrap();
-        assert_eq!(signature.signature, parsed.0);
+    fn signature_roundtrip(signature: Signature) {
+        let serialized: String =
```
lemmih, line 124: Is the `: String` type signature necessary?

utils/json/src/signature.rs
jdjaustin: Ah that was just part of my troubleshooting before I discovered I needed to rename `super::json::serialize` to `super::json::signature_type::serialize` (and deserialize respectively). I'll fix that.

utils/json/src/signature.rs
jdjaustin: I know we've been using the explicit type signature in other tests, but apparently we don't need it here!

hanabi1224 (approved):

utils/json/src/signature.rs
```diff
@@ -119,46 +119,21 @@ mod tests {
...
-    }
-
     #[quickcheck]
-    fn signature_roundtrip(signature: SignatureWrapper) {
-        let serialized = serde_json::to_string(&SignatureJsonRef(&signature.signature)).unwrap();
-        let parsed: SignatureJson = serde_json::from_str(&serialized).unwrap();
-        assert_eq!(signature.signature, parsed.0);
+    fn signature_roundtrip(signature: Signature) {
+        let serialized = forest_test_utils::to_string_with!(&signature, super::json::serialize);
+        let parsed = forest_test_utils::from_str_with!(&serialized, super::json::deserialize);
```
lemmih, line 125: Why switch to `to_string_with!` rather than using `SignatureJsonRef`?

jdjaustin: I thought we had preferred to use the `forest_test_utils` macros in previous `quickcheck` tests, but perhaps I'm mistaken. I would argue that it improves readability a bit, but I can change it back to using `SignatureJsonRef` if that's better.

lemmih: The macros are there to assist if you do not have a helper structure such as `SignatureJsonRef`.

LesnyRumcajs (approved):

PR #2624, opened 2023-03-01
Use tempfile in chain_export

### LINKED ISSUE #2356: Miscellaneous tasks

creativcoder:
## Summary of changes

- Use `NamedTempFile` crate in chain export for better management of tmp file.

Partially Closes #2356

## Change checklist


LesnyRumcajs (approved):
good stuff

node/rpc/src/chain_api.rs
```diff
@@ -109,24 +109,13 @@ where
             .await
     } {
         Ok(checksum) if !dry_run => {
-            std::fs::rename(&out_tmp, &output_path)?;
```
hanabi1224, line 112: I believe there's similar logic in `snapshot fetch`, would u create a tracking issue if not include that in this PR?

hanabi1224 (approved):

node/rpc/src/chain_api.rs
sudo-shashank: yes I'll create a new ticket for handling similar cases for `snapshot fetch`

PR #2631, opened 2023-03-06
Deduce network type from daemon and set cli config

### LINKED ISSUE #2609: forest-cli - deduce network type
**Issue summary**

Currently, we need to add `--chain calibnet` to a bunch of `forest-cli` subcommands. This is not required for most, except, e.g., `snapshot fetch`, which does not require a running node.

Instead, we could ask the Forest daemon via RPC what network it runs. Then do tricks like [here](https://github.com/ChainSafe/forest/blob/main/forest/cli/src/subcommand.rs#L8-L10) based on that.

**Task summary**

- [X] Figure out how to get the network type via RPC call. Something may already be available.
- [X] Don't use the explicit flag in subcommands that communicate with the daemon over RPC,
- [X] The `Address` trick above shouldn't require an explicit flag as well.

**Acceptance Criteria**

- [X] `--chain calibnet` is not required for commands that use the RPC daemon endpoint,
- [X] mainnet is also correctly deduced,
- [X] a basic integration test showing that this works as expected (simply removing the `--chain calibnet` from `forest-cli sync wait` in the CI will do).

sudo-shashank:
## Summary of changes

- Identify network type from forest daemon for `forest-cli` subcommands

Closes #2609 

## Change checklist


lemmih (changes requested):
Hmm, doing this right will probably require re-thinking the way we handle CLI options. Let's shelve this for now.

lemmih:
More specifically, the client should have an option for `--rpc false`. That makes no sense. And we need to re-think the `--chain` flag.

LesnyRumcajs:
@lemmih I'd still merge it for now. It kind of is an improvement over what we have now - some obscure serialisation error if the client and the daemon don't match.

LesnyRumcajs (approved):

lemmih:
> @lemmih I'd still merge it for now. It kind of is an improvement over what we have now - some obscure serialisation error if the client and the daemon network don't match. It's not ideal and moving forward we should prioritise implementing the new architecture proposed some time ago. Or come up with another and implement it.

But we just replaced one set of obscure errors for another. If you try to validate a snapshot without using the `--rpc false` flag, forest-cli will fail badly.

lemmih:
@sudo-shashank If we change this PR to only query the daemon when it's actually required then I'll re-review.

lemmih (approved):
Approved.

This PR solves our immediate problems but the CLI situation is still a mess and will need some love at a later point in time.

PR #2636, opened 2023-03-07
Port lotus eth changes

LesnyRumcajs:
## Summary of changes

- 

Closes 

## Change checklist


elmattic:
Nice. Could you link the corresponding PR/commit in Lotus that you have used?

LesnyRumcajs:
> Nice. Could you link the corresponding PR/commit in Lotus that you have used?

Most of them come from this giga PR https://github.com/filecoin-project/lotus/commit/cdf3812e4081d35692b5ab81c573ab82d0df69cf

There is also this that caught my attention https://github.com/filecoin-project/lotus/commit/a9a65e05d8564d0b5bfe35416f50d0ea7b5fad51

lemmih (approved):

blockchain/state_manager/src/utils.rs
```diff
@@ -104,3 +105,116 @@ where
...
+
+    // Only allow such actors to send if their delegated address is in the EAM's
+    // namespace.
+    return if let Payload::Delegated(address) = actor
+        .delegated_address
+        .as_ref()
+        .expect("unfallible")
+        .payload()
+    {
+        address.namespace() == Address::ETHEREUM_ACCOUNT_MANAGER_ACTOR.id().unwrap()
```
elmattic, line 148: Unwrap is safe here?

blockchain/state_manager/src/utils.rs
```diff
@@ -104,3 +105,116 @@ where
...
+    //
+    // But it's a safeguard.
+    //
+    // Note 2: ad-hoc checks for network versions like this across the codebase
+    // will be problematic with networks with diverging version lineages
+    // (e.g. Hyperspace). We need to revisit this strategy entirely.
+    if network_version < NetworkVersion::V18 {
+        return is_account_actor(&actor.code);
+    }
+
```
elmattic, line 126: I would keep this Lotus comment as well here.

`// After nv18, we also support other kinds of senders.`

vm/interpreter/src/lib.rs
```diff
@@ -40,6 +40,11 @@ where
         .get_actor(addr)?
         .ok_or_else(|| anyhow::anyhow!("Failed to retrieve actor: {}", addr))?;
 
+    // If there _is_ an f4 address, return it as "key" address
+    if let Some(address) = act.delegated_address {
```
elmattic, line 44: We don't need to test for `StateTree::V5` here, since we don't want to support previous network version, right?

elmattic:
What about supporting Ethereum API RPC endpoints? Do we want to support this in the near future?

elmattic:
What about support of equivalent `resolve_to_deterministic_address`/`resolve_to_deterministic_address_at_finality` functions?

Do you think it's not needed right now?

vm/interpreter/src/lib.rs
LesnyRumcajs: It's a new field. Previous versions should have it set to `None`. Checking the state tree version is redundant. Feel free to correct me if you think otherwise.

LesnyRumcajs:
> What about supporting Ethereum API RPC endpoints? Do we want to support this in the near future?

No clue how high supporting Ethereum is on our priorities list. This PR certainly is not intended to do so - only do some basic validations. @lemmih ?

> What about support of equivalent resolve_to_deterministic_address/resolve_to_deterministic_address_at_finality functions?
> Do you think it's not needed right now?

I only grabbed a few changes I encountered and tried to apply them when investigating stateroot mismatches, so they don't sit in my local stash. There are certainly more to port from this 15k LoC PR; this would require a more extended investigation.

elmattic (approved):

vm/interpreter/src/lib.rs
elmattic: No, I think it's good like this.

PR #2635, opened 2023-03-07
feat: save chain metadata to files instead of database

hanabi1224:
## Summary of changes

Changes introduced in this pull request:

As part of https://github.com/ChainSafe/forest/issues/2292

-  Save genesis block header and heaviest tipset keys to files instead of database

## Reference issue to close (if applicable)

Closes 

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG](https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md) is up-to-date. All user-facing changes should be reflected in this document.

PR #2638, opened 2023-03-08
feat: database garbage collection

hanabi1224:
## Summary of changes

- [x] Implement a DB wrapper as a drop-in replacement that can be garbage collected.
- [ ] Hook the DB wrapper into `ChainStoreo` (Blocked by https://github.com/ChainSafe/forest/pull/2635)

Closes 

## Change checklist


forest/daemon/src/daemon.rs
```diff
@@ -294,11 +311,13 @@ pub(super) async fn start(opts: CliOpts, config: Config) -> anyhow::Result<Db> {
                     sync_state,
                     network_send,
                     network_name,
-                    beacon: rpc_state_manager.beacon_schedule(), /* TODO: the RPCState can fetch
+                    beacon: rpc_state_manager.beacon_schedule(), /* TODO: the RPCState can
+                                                                  * fetch
                                                                   * this itself from the
                                                                   * StateManager */
```
lemmih, line 317: let's reflow this comment.

ipld/src/util.rs
```diff
@@ -78,3 +80,70 @@ where
 
     Ok(())
 }
+
+pub const DEFAULT_RECENT_ROOTS: ChainEpoch = 2000;
```
lemmih, line 84: This is not a `ChainEpoch`, it's a delta.

node/db/src/rolling/impls.rs
```diff
@@ -0,0 +1,333 @@
...
+impl Store for RollingDB {
+    fn read<K>(&self, key: K) -> Result<Option<Vec<u8>>, crate::Error>
+    where
+        K: AsRef<[u8]>,
+    {
+        for db in self.db_queue.read().iter() {
+            if let Ok(Some(v)) = Store::read(db, key.as_ref()) {
+                return Ok(Some(v));
+            }
+        }
```
lemmih, line 80: This code suppresses DB errors. Let's propagate them instead.

node/db/src/rolling/impls.rs
```diff
@@ -0,0 +1,333 @@
...
+    {
+        Store::write(&self.current(), key, value)
+    }
+
+    fn delete<K>(&self, key: K) -> Result<(), crate::Error>
+    where
+        K: AsRef<[u8]>,
+    {
+        Store::delete(&self.current(), key)
+    }
```
lemmih, line 111: This is not right. If the key exists in the older partition and it could still be read after a delete. However, if we move all the metadata out of the blockstore then we can get rid of the `delete` method altogether.

node/db/src/rolling/index.rs
```diff
@@ -0,0 +1,2 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
```
lemmih, line 2: Is this file supposed to be empty?

node/db/src/rolling/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use crate::db_engine::{open_db, Db, DbConfig};
+
+#[derive(Clone)]
+pub struct RollingDB {
+    db_root: Arc<PathBuf>,
+    db_config: Arc<DbConfig>,
+    db_index: Arc<RwLock<FileBacked<DbIndex>>>,
+    /// A queue of active databases, from youngest to oldest
+    db_queue: Arc<RwLock<VecDeque<Db>>>,
```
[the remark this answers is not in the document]
hanabi1224: fixed

node/db/src/rolling/index.rs
hanabi1224: removed

ipld/src/util.rs
hanabi1224: changed to i64

forest/daemon/src/daemon.rs
hanabi1224: fixed

node/db/src/rolling/impls.rs
hanabi1224: fixed

node/db/src/rolling/mod.rs
```diff
@@ -0,0 +1,33 @@
...
```
lemmih, line 27: Good job making `db_queue` abstract.

Since we always want at least one DB, let's use a tuple for the queue: `db_queue: Arc<RwLock<(Db, VecDeque<Db>)>>`

node/db/src/rolling/impls.rs
```diff
@@ -0,0 +1,371 @@
...
+
+impl Blockstore for RollingDB {
+    fn has(&self, k: &Cid) -> anyhow::Result<bool> {
+        let mut errors = AggregatedError::new();
+        for db in self.db_queue().iter() {
+            match Blockstore::has(db, k) {
+                Ok(true) => return Ok(true),
+                Ok(false) => {}
+                Err(e) => errors.push(e),
+            }
```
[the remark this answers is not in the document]
lemmih: Actually, I think it's incorrect to collect errors like this. Imagine if the first db returned an error and the second db returned `true`. Then this code would return `true` and silently ignore the error.

node/db/src/rolling/impls.rs
```diff
@@ -0,0 +1,371 @@
...
+        )?)
+    }
+
+    pub fn db_count(&self) -> usize {
+        self.db_queue().len()
+    }
+
+    pub fn current(&self) -> Db {
+        self.db_queue
+            .read()
```
lemmih, line 317:
```suggestion
        self.db_queue()
```

lemmih (changes requested):

node/db/src/rolling/impls.rs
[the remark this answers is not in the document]
hanabi1224: Depending on the perspective of seeing it. From store users' perspective, as long as the store gives the correct result, its inner errors are not important. In scenarios where data is duplicated in multiple shards to survive single-point failure, it is intended to omit internal errors as long as the correct data can be served.

I don't think either is right or wrong, but more of a design choice based on the tolerance level of internal errors, e.g. we could return the correct result to the caller and log the internal errors for future investigation. And of course, we could fail the call on any single-point failure.  

Let me know your preference and I can make the change accordingly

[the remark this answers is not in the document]
lemmih: The errors are important (and fatal). I strongly prefer not ignoring them.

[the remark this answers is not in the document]
hanabi1224: > The errors are important (and fatal). I strongly prefer not ignoring them.

In this case, I think it's better to change the API of `Store` trait to not return errors but panic internally, what do u think?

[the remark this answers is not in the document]
lemmih: I don't think so. Panics are only for bugs. If the DB fails, it might just be that your hard-drive failed or something else which isn't a bug.

node/db/src/rolling/mod.rs
hanabi1224: Could u elaborate on the benefit?   

I agree `(Db, VecDeque<Db>)` is a much better byte for public API. However, to implement this public API, I would still use `VecDeque<Db>` as internal implementation as it's a straightforward and direct mapping to the actual folder structure, making the code that maintains the relationship between the in-memory data structure and the actual underlining directories more straightforward and less error-prone.

In this case, `db_queue` is an internal field that is only accessible in the same module, I don't see why we should change it

lemmih: It's an invariant that there should always be at least 1 DB. We can statically enforce that invariant through the type-system.

The rest of the code doesn't have to care and can keep using a single `VecDeque<Db>`.

Later, I want to show that we always want 1 older Db. Then we can use `(Db, Db)`.

node/db/src/rolling/impls.rs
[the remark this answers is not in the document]
hanabi1224: OK I can make the change, although I still don't like this design cuz it uses more assumptions

- `Store` users want data
- The underlying DB or DB collection does not have any fault tolerance mechanism, like data backup/sharding
- `Store` users care more about any single point failures that are internal to the `Store` implementation than getting the data they query.

The last point is especially bad IMO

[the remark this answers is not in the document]
lemmih: I do not understand you points but I think it's vital that we're on the same page. Let's discuss in Slack.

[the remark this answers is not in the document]
hanabi1224: fixed

node/db/src/rolling/mod.rs
hanabi1224: fixed

node/db/src/rolling/impls.rs
hanabi1224: fixed

node/db/src/rolling/mod.rs
hanabi1224: Changed the DB fields as requested, although I would still prefer decoupling DB design from the specific GC approach, designing DB in a GC agnostic way, and building it from bottom up

lemmih: Your objection has been noted.

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,137 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
```
lemmih, line 3: The details of the algorithm should be documented here.

lemmih (changes requested):

ipld/src/util.rs
```diff
@@ -78,3 +79,70 @@ where
 
     Ok(())
 }
+
+pub const DEFAULT_RECENT_ROOTS: i64 = 2000;
```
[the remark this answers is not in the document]
hanabi1224: Would u suggest including it in this PR or doing it separately?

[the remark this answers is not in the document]
lemmih: Doing it immediately after this PR is fine. This PR kinda breaks the `--recent-stateroots` export flag (ie exporting with more than 2k stateroots becomes impossible). Not a huge deal but it should be fixed as soon as possible.

utils/genesis/src/lib.rs
```diff
@@ -191,22 +194,19 @@ where
 pub async fn forest_load_car<DB, R>(store: DB, reader: R) -> anyhow::Result<Vec<Cid>>
 where
     R: futures::AsyncRead + Send + Unpin,
-    DB: Store,
+    DB: Store + Sync + Send + 'static,
```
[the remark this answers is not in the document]
hanabi1224: That will be much better, but the change could bloat this PR with many unrelated refactorings, would u prefer doing it here or separately?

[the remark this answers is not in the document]
lemmih: Feel free to do it separately.

ipld/src/util.rs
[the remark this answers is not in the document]
hanabi1224: I can remove that flag in this PR if that's sufficient. Let me know if the fix is more than just removing it then I will do it separately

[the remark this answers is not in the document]
lemmih: Removing the `--recent-stateroots` flag would be good. Then we can make the number of state roots configurable in a separate PR.

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,137 @@
...
+                Ok(block)
+            }
+        })
+        .await?;
+        drop(tx);
+        write_task.await??;
+
+        info!(
+            "Garbage collection finished at epoch {}, took {}s",
```
lemmih, line 130: Let's count bytes copied and entries copied as well. This could go in a debug statement.

node/db/src/rolling/mod.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
```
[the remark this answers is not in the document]
hanabi1224: Done

node/db/src/rolling/gc.rs
hanabi1224: Done

utils/genesis/src/lib.rs
```diff
@@ -191,22 +194,19 @@ where
...
 {
     // 1GB
     const BUFFER_CAPCITY_BYTES: usize = 1024 * 1024 * 1024;
 
+    let (tx, rx) = flume::unbounded();
```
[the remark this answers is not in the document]
hanabi1224: Fixed

utils/forest_utils/src/common/mod.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::fmt::Display;
+
+#[derive(Debug, thiserror::Error)]
+pub struct AggregatedError<T: Display>(Vec<T>);
```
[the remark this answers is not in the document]
hanabi1224: Removed

node/db/src/rolling/gc.rs
hanabi1224: Done

ipld/src/util.rs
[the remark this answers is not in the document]
hanabi1224: Beside `--recent-stateroots`, I believe `--tipset` is also broken,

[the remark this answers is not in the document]
lemmih: yeah

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,170 @@
...
+//!
+//! ## GC workflow
+//! 1. Walk back from the current heaviest tipset to the genesis block, collect
+//! all the blocks that are reachable from the snapshot
+//! 2. writes blocks that are absent from the `current` database to it
+//! 3. delete `old` database(s)
+//! 4. sets `current` database to a newly created one
+//!
+//! ## Correctness
+//! This alroghrim considers all blocks that are visited during the snapshot
```
lemmih, line 16: algorithm

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,170 @@
...
+//! export task reachable, and ensures they are all transfered and kept in the
```
lemmih, line 17: transferred

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,170 @@
...
+//! current DB space. A snapshot can be used to bootstrap the node from
+//! scratch thus the algorithm is considered approapriate when the post-GC
```
lemmih, line 19: appropriate

ipld/src/util.rs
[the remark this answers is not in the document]
hanabi1224: both options are removed, created https://github.com/ChainSafe/forest/issues/2655

node/db/src/rolling/mod.rs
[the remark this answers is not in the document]
lemmih: You didn't write about the nature of our data (it's vital for our design that the data is a persistent DAG), our design goals, or the reasoning for why we chose a concurrent, semi-space collector.

lemmih (changes requested):

node/db/src/rolling/gc.rs
hanabi1224: Good catch! But why is it not caught by the spellcheck CI task

lemmih: Hmm, something is wrong with the spellchecker...

node/db/src/rolling/mod.rs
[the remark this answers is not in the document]
hanabi1224: > it's vital for our design that the data is a persistent DAG

Could you help me with properly phrasing this?  Design goals and reasonings are added

node/db/src/rolling/gc.rs
hanabi1224: fixed

node/db/src/rolling/gc.rs
hanabi1224: fixed

node/db/src/rolling/mod.rs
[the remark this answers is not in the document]
hanabi1224: And maybe also the design goals and reasoning in case what I wrote is not correct

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,180 @@
...
+                if lock.is_err() {
+                    continue;
+                }
+            }
+
+            if let (Ok(total_size), Ok(current_size)) = (
+                self.db.total_size_in_bytes(),
+                self.db.current_size_in_bytes(),
+            ) {
+                // Collect when size of young partition > 0.5 * size of old partition
```
lemmih, line 112: I'm not convinced by this metric for when to trigger a garbage collection. It's quite difficult to reason about the size of the disk space required in relation to the live data.

How about we track how much live data there was at the last collection and then base the calculation around that number?

node/db/src/rolling/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+//!
+//! This DB wrapper is specially designed for supporting the concurrent,
+//! semi-space GC algorithm that is implemented in [DbGarbageCollector],
+//! containing a reference to the `old` DB space and a reference to the
+//! `current` DB space. Both underlying key-vale DB are supposed to contain only
+//! block data as value and its content-addressed CID as key
```
lemmih, line 9: Here's how I would write it:
```rust
//! The state of the Filecoin Blockchain is a persistent, directed acyclic
//! graph. Data in this graph is never mutated nor explicitly deleted but may
//! become unreachable over time.
//! 
//! This module contains a concurrent, semi-space garbage collector. The garbage
//! collector is guaranteed to be non-blocking and can be expected to run with a
//! fixed memory overhead and require disk space proportional to the size of the
//! reachable graph. For example, if the size of the reachable graph is 100GiB,
//! expect this garbage collector to use 3x100GiB = 300GiB of storage.
```

lemmih (changes requested):

node/db/src/rolling/gc.rs
hanabi1224: Any suggestions on what formula to use? I use 0.5 because if the first GC is triggered when `size(current)` >= `size(old)`, the live data size at GC ends will be `size(current)` + `size(new data during GC)` the threshold will be ever-growing

node/db/src/rolling/mod.rs
hanabi1224: Updated

node/db/src/rolling/gc.rs
hanabi1224: P.S. As I understand, `size(old)` is the live data size after previous GC, let me know if u were refering to sth else

lemmih: `size(old)` is not the size of the reachable data after a GC run. When a GC run is started, both the `old` and `new` partition will contain some unreachable data. Therefore, after a GC run has finished, the `old` partition will contain both reachable and unreachable data.

hanabi1224: Ok I can get the exact size of the reachable data, any suggestion on how to decide the threshold with it?

lemmih: Maybe when the size of the total database is twice the reachable data.

node/db/src/rolling/mod.rs
hanabi1224: Moved

node/db/src/rolling/gc.rs
hanabi1224: Done

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,202 @@
...
+            gc_rx,
+            last_reachable_bytes: AtomicU64::new(0),
+        }
+    }
+
+    pub fn get_tx(&self) -> flume::Sender<flume::Sender<anyhow::Result<()>>> {
+        self.gc_tx.clone()
+    }
+
+    pub async fn collect_loop_passive(&self) -> anyhow::Result<()> {
```
lemmih, line 95: Would like some high-level documentation here.

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,202 @@
...
+                if should_collect {
+                    if let Err(err) = self.collect_once(tipset).await {
+                        warn!("Garbage collection failed: {err}");
+                    }
+                }
+            }
+        }
+    }
+
+    pub async fn collect_loop_event(self: &Arc<Self>) -> anyhow::Result<()> {
```
lemmih, line 134: More documentation here.

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,202 @@
...
+                if let Err(e) = responder.send(result) {
+                    warn!("{e}");
+                }
+            });
+        }
+
+        Ok(())
+    }
+
+    async fn collect_once(&self, tipset: Tipset) -> anyhow::Result<()> {
```
lemmih, line 149: And documentation here.

lemmih (changes requested):
Could you also include some benchmark numbers? What's the cost, if any, of doing two DB lookups instead of one.

Also, `next_partition()` is incredibly unsafe and can be accidentally called from anywhere. We definitely need to refactor that but maybe it can wait for a follow-up PR.

node/db/src/rolling/gc.rs
hanabi1224: Added

node/db/src/rolling/gc.rs
hanabi1224: Added

node/db/src/rolling/gc.rs
hanabi1224: Added

utils/genesis/src/lib.rs
```diff
@@ -159,18 +159,21 @@ where
 
     info!("Accepting {:?} as new head.", ts.cids());
 
+    // Creates a new partition after snapshot import if supported
+    sm.blockstore().next_partition()?;
+
```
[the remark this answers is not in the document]
hanabi1224: `fn next_partition` removed

lemmih:
Ran benchmarks with the entire snapshot in the "old" partition and it doesn't affect performance in any way I can measure.

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,230 @@
...
+
+            if let (Ok(total_size), Ok(current_size), last_reachable_bytes) = (
+                self.db.total_size_in_bytes(),
+                self.db.current_size_in_bytes(),
+                self.last_reachable_bytes.load(atomic::Ordering::Relaxed),
+            ) {
+                let should_collect = if last_reachable_bytes > 0 {
+                    total_size > 2 * last_reachable_bytes
+                } else {
+                    total_size > 0 && current_size * 3 > total_size
```
lemmih, line 140: When we import a new snapshot, `last_reachable_bytes` will be 0 and `current_size` will be the same as `total_size`, right? Doesn't that mean that the GC will never be triggered?

lemmih: Oh, right, so it'll start a GC run immediately?

hanabi1224: The check interval is 10min, so it will be triggered 0-10min after snapshot being imported and HEAD being set

lemmih (approved):
Alright!

I'll add issues for the following improvements:

- Load data directly into the old space rather than the new space.
- Show progress during garbage collection (and chain exporting).
- Possibly get rid of the `Store` trait altogether.
- Yield during GC to allow Forest to shutdown properly.

forest/daemon/src/daemon.rs
```diff
@@ -162,6 +161,23 @@ pub(super) async fn start(opts: CliOpts, config: Config) -> anyhow::Result<Db> {
...
+    #[allow(clippy::redundant_async_block)]
+    services.spawn({
+        let db_garbage_collector = db_garbage_collector.clone();
+        async move { db_garbage_collector.collect_loop_passive().await }
+    });
+    #[allow(clippy::redundant_async_block)]
+    services.spawn({
+        let db_garbage_collector = db_garbage_collector.clone();
+        async move { db_garbage_collector.collect_loop_event().await }
+    });
```
LesnyRumcajs, line 180: It'd be nice to move it to a separate method. To my understanding, those calls in general need each other.

forest/daemon/src/daemon.rs
```diff
@@ -583,28 +600,28 @@ mod test {
 
     use super::*;
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread")]
```
LesnyRumcajs, line 603: Why do we need it?

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,230 @@
...
+//!
+//! The current implementation of the garbage collector is a concurrent,
+//! semi-space one.
+//!
+//! ## Design goals
+//! Implement a correct GC algorithm that is simple and efficient for forest
+//! scenarios.
+//!
+//! ## GC algorithm
+//! We choose `semi-space` GC algorithm for simplicity and sufficiency
```
LesnyRumcajs, line 13:
```suggestion
//! We chose the `semi-space` GC algorithm for simplicity and sufficiency
```

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,230 @@
...
+//! This algorithm considers all blocks that are visited during the snapshot
+//! export task reachable, and ensures they are all transferred and kept in the
+//! current DB space. A snapshot can be used to bootstrap a node from
+//! scratch thus the algorithm is considered appropriate when the post-GC
+//! database contains blocks that are sufficient for exporting a snapshot
+//!
+//! ## Disk usage
+//! During `walk_snapshot`, data from the `old` DB is duplicated in the
+//! `current` DB, which uses extra disk space of up-to 100% of the snapshot file
+//! size
```
LesnyRumcajs, line 35:
```suggestion
//! During `walk_snapshot`, data from the `old` DB is duplicated in the
//! `current` DB, which uses extra disk space of up to 100% of the snapshot file
//! size
```

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,230 @@
...
+//!
+//! ## Memory usage
+//! During the data carry-over process, a memory buffer with a fixed capacity is
+//! used to speed up the database write operation
+//!
+//! ## Scheduling
+//! 1. GC is triggered automatically when total DB size is greater than 2x of
```
LesnyRumcajs, line 42: It'd be nice to add an actual mainnet example.

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,230 @@
...
+//! the last reachable data size
+//! 2. GC can be triggered manually by `forest-cli db gc` command
+//! 3. There's global GC lock to ensure at most one GC job is running
```
LesnyRumcajs, line 45:
```suggestion
//! the last reachable data size
//! 2. GC can be triggered manually by `forest-cli db gc` command
//! 3. There's a global GC lock to ensure at most one GC job is running
```

LesnyRumcajs: Do we need to specify the chain?

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,230 @@
...
+//! DB lookup performance is almost on-par between from single DB and two DBs.
+//! Time cost of `forest-cli snapshot export --dry-run` on DO droplet with 16GiB
+//! ram is between `9000s` to `11000s` for both scenarios, no significant
+//! performance regression has been observed
+//!
+//! ### Write performance
+//! DB write performance is typically on-par with `snapshot import`, note that
+//! when the `current` DB space is very large, it tends to trigger DB re-index
+//! more frequently, each DB re-index could pause the GC process for a few
+//! minutes. The same behaviour is observed during snapshot import as well.
```
LesnyRumcajs, line 61:
```suggestion
//! ### Write performance
//! DB write performance is typically on par with `snapshot import`. Note that
//! when the `current` DB space is very large, it tends to trigger DB re-index
//! more frequently, each DB re-index could pause the GC process for a few
//! minutes. The same behaviour is observed during snapshot import as well.
```

node/db/src/rolling/gc.rs
```diff
@@ -0,0 +1,230 @@
...
+            let db = db.clone();
+            let tx = tx.clone();
+            let reachable_bytes = reachable_bytes.clone();
+            async move {
+                let block = db
+                    .get(&cid)?
+                    .ok_or_else(|| anyhow::anyhow!("Cid {cid} not found in blockstore"))?;
+
+                let pair = (cid, block.clone());
+                // Key size is 32 bytes in paritydb
```
LesnyRumcajs, line 204: So, does this work at all with RocksDb? If not, how does Forest behave when RocksDb is used?

node/db/src/rolling/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+
+//! The state of the Filecoin Blockchain is a persistent, directed acyclic
+//! graph. Data in this graph is never mutated nor explicitly deleted but may
+//! become unreachable over time.
+//!
+//! This module contains a concurrent, semi-space garbage collector. The garbage
+//! collector is guaranteed to be non-blocking and can be expected to run with a
+//! fixed memory overhead and require disk space proportional to the size of the
+//! reachable graph. For example, if the size of the reachable graph is 100GiB,
+//! expect this garbage collector to use 3x100GiB = 300GiB of storage.
```
LesnyRumcajs, line 12: Before merging this, we should update the specs of our DO droplets.

utils/forest_utils/src/db/mod.rs
```diff
@@ -68,3 +72,40 @@ pub trait BlockstoreExt: Blockstore {
...
+            total_bytes += 32 + value.len();
+            total_entries += 1;
+            buffer.push((key, value));
+            if estimated_buffer_bytes >= buffer_capacity_bytes {
+                self.put_many_keyed(std::mem::take(&mut buffer))?;
+                estimated_buffer_bytes = 0;
+            }
+        }
+        self.put_many_keyed(buffer)?;
+        info!(
```
LesnyRumcajs, line 101: Is this instead a debug? Also, I'm not sure calling timers by default makes sense in a performance-related method.

utils/forest_utils/src/db/mod.rs
```diff
@@ -68,3 +72,40 @@ pub trait BlockstoreExt: Blockstore {
...
+        buffer_capacity_bytes: usize,
+    ) -> anyhow::Result<()> {
+        let start = Utc::now();
+        let mut total_bytes = 0;
+        let mut total_entries = 0;
+        let mut estimated_buffer_bytes = 0;
+        let mut buffer = vec![];
+        while let Ok((key, value)) = rx.recv_async().await {
+            // Key is stored in 32 bytes in paritydb
+            estimated_buffer_bytes += 32 + value.len();
```
LesnyRumcajs, line 91: we used this `32` a few times already, let's make it a named, global const.

LesnyRumcajs (commented):
It'd be fantastic to make a blogpost about it :)

node/db/src/rolling/gc.rs
hanabi1224: All functions work with rocksdb, however, I did not test thru the crash resilience story for rocksdb so it might not be able to recover when forest is shutdown improperly. I believe we had crash resilience issue with rocksdb before so I think the potential regression is not introduced by this change

node/db/src/rolling/mod.rs
hanabi1224: Agreed, 320GB is not sufficient if docker and rust toolchains need to be installed

hanabi1224: But we could mount a volume of ~500GB on the fly and symbol link the data folder

forest/daemon/src/daemon.rs
hanabi1224: `collect_loop_passive` and `collect_loop_event` work independently, either or both can be disabled

forest/daemon/src/daemon.rs
lemmih: Why is the single threaded runtime insufficient? Surely it can run multiple tasks in parallel.

node/db/src/rolling/gc.rs
hanabi1224: Do u mean an example output in log?

forest/daemon/src/daemon.rs
lemmih: I think we need to debug why that happened. We shouldn't be doing anything blocking.

node/db/src/rolling/gc.rs
LesnyRumcajs: Just example numbers with mainnet.

forest/daemon/src/daemon.rs
LesnyRumcajs: The last time I encountered deadlock with tokio was because there were multiple runtimes. It was a bug. :)

node/db/src/rolling/gc.rs
LesnyRumcajs: So you tested that the GC works with rocksdb? How long does it take? I'm asking because the key size value is hardcoded here.

utils/forest_utils/src/db/mod.rs
LesnyRumcajs: If it's helpful, perhaps let's have a proper Prometheus metric? Like, average speed or something like this.

node/db/src/rolling/gc.rs
hanabi1224: Yes for calibnet. But I think it will no longer be needed once the other chain name deduction pr is merged

node/db/src/rolling/gc.rs
hanabi1224: Example numbers added

utils/forest_utils/src/db/mod.rs
hanabi1224: In this case, I think log of individual runs is more helpful than aggregated metrics to understand details and nature of the database and for troubleshooting.  How about keeping it as it is, given it only emits 2-3 lines every 1.5-2 days?

LesnyRumcajs: Ok

PR #2652, opened 2023-03-10
Use backing file for storing validated block cids

### LINKED ISSUE #2642: Remove block validation meta-data from the blockstore
**Issue summary**

Forest keeps track of which blocks have been validated by storing a set of validated CIDs in the blockstore. In the future, we want the blockstore to only contain the blockchain itself. The set of validated blocks should therefore be moved somewhere else.

The store of validated blocks does not need to be crash proof. Losing data just means more blocks will be re-validated. We could use an approach similar to #2635 but perhaps only sync the data to disk once every 10 minutes.

It would be tempting to get rid of the set of validated blocks and go by the heaviest tipset. This would require significant changes to the code base, though.

elmattic:
## Summary of changes

- Use backing file for storing validated blocks instead of store
- Fix typo in documentation

Closes #2642

## Change checklist


lemmih (changes requested):
A lot of `Store` constraints can be removed from the rest of the code now. Let's remove as many as possible.

lemmih:
Oh, and the checklist needs to be completed.

utils/forest_utils/src/db/file_backed_obj.rs
```diff
@@ -1,36 +1,61 @@
...
 use cid::Cid;
 use log::warn;
 
 pub struct FileBacked<T: FileBackedObject> {
     inner: T,
     path: PathBuf,
+    last_sync: Option<SystemTime>,
 }
 
+const SYNC_PERIOD: Duration = Duration::from_secs(600);
```
LesnyRumcajs, line 20: Shall we add it to the configuration?

utils/forest_utils/src/db/file_backed_obj.rs
```diff
@@ -1,36 +1,61 @@
...
+        if let Some(last_sync) = self.last_sync {
+            if now.duration_since(last_sync)? > SYNC_PERIOD {
+                self.last_sync = Some(now);
+                self.flush_to_file()?;
+            }
+            return Ok(());
+        }
+        self.last_sync = Some(now);
+        Ok(())
+    }
```
LesnyRumcajs, line 50: Fancy adding a unit test for this method?

utils/forest_utils/src/db/file_backed_obj.rs
elmattic: I'm not sure if it's really critical for our users to be configurable. I prefer to leave it like this for now (ten minutes should be a good start).

utils/forest_utils/src/db/file_backed_obj.rs
elmattic: Let me think.

utils/forest_utils/src/db/file_backed_obj.rs
LesnyRumcajs: If you make it configurable then you will be able to add some unit tests. :)

LesnyRumcajs: But modifying `last_sync` directly is also fine.

elmattic: I've made it configurable.

utils/forest_utils/src/db/file_backed_obj.rs
elmattic: Added two tests.

utils/genesis/src/lib.rs
```diff
@@ -145,7 +145,7 @@ where
     // Update head with snapshot header tipset
     sm.chain_store().set_heaviest_tipset(ts.clone())?;
 
-    sm.blockstore().flush()?;
+    //sm.blockstore().flush()?;
```
lemmih, line 148: We still need this flush, unfortunately.

elmattic: Will add back.

vm/state_migration/src/nv12/miner.rs
```diff
@@ -13,7 +13,7 @@ use cid::multihash::Code::Blake2b256;
 use cid::Cid;
 use fil_actor_interface::actorv3::miner::State as V3State;
 use fil_actor_interface::actorv4::miner::State as V4State;
-use forest_db::Store;
+
```
lemmih, line 16: spaces?

lemmih (approved):

LesnyRumcajs (approved):

PR #2681, opened 2023-03-17
Fix muxer that can be stucked into an infinite loop

### LINKED ISSUE #2672: Forest chain muxer error
**Issue summary**

I got hit by this several times already. In short, sometimes, in case of an error (e.g., state mismatch) and after throwing a ctrl-c (I am not sure it's 100% needed, though) Forest will start spamming the terminal with:
```
2023-03-15T12:00:49.755785Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755788Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755789Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755792Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755793Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755796Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755798Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755802Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755804Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755806Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755808Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755810Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755812Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755815Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755817Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
2023-03-15T12:00:49.755819Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-15T12:00:49.755821Z ERROR forest_chain_sync::chain_muxer: Evaluating the network head failed, retrying. Error = P2PEventStreamReceive("receiving on a closed channel")
```

It is impossible to kill it with a regular ctrl-c; one has to open another tab and `pkill -9 forest`. To make it even worse, this uses a significant amount of CPU (probably coming from endless logs flooding the terminal without sleep). It's hard to reproduce, but it happened quite a few times when I sent SIGINT at the wrong moment.

We need to handle it better, but I have no clear solution. The simple idea is to shut down Forest after some amount of errors in a given interval.

elmattic:
## Summary of changes

- Fix muxer that can go into an infinite loop during shutdown

Closes #2672

82ca23c does run `Muxer` future separately to reproduce the issue.

## Change checklist

      (if possible),
- [x] I have made sure the


lemmih:
Could you explain the problem and how this fixes it?

elmattic:
When creating the daemon `chain_muxer` task, we are giving it the `network_rx` channel just created [before](https://github.com/ChainSafe/forest/blob/main/forest/daemon/src/daemon.rs#L237).

The root cause of the problem lies in the order of cancelation of those tasks during tokio rt shutdown.

If the `chain_muxer` is the first to be cancelled, so before `p2p_service`, everything should be fine. Otherwise the `network_rx` channel is closed and some error will be received in the `ChainMuxer` promise.

My fix consists into catching this error and returning `Poll::Pending`. Maybe we should test for the string `"receiving on a closed channel"` as well, just to be sure.

elmattic:
> So this won't loop at all now and will exit? Or will it continue running, just without the error logs?

This gives the hand to the tokio runtime. And since the task has be cancelled it will then exit.

lemmih:
I still have a hard time understand both the problem and the solution. Could you explain it in a simpler way and then add that explanation to the code?

lemmih:
> > I still have a hard time understand both the problem and the solution. Could you explain it in a simpler way and then add that explanation to the code?
> 
> I've added a comment and testing as well for the specific error string. I'm not sure if it's the best but I'm open to some better fix.

It looks to me that the root of the problem is that we cannot presently tell the difference between incorrect input (such as malformed tipsets) and fatal errors (such as a DB error or a closed channel). Right now, we treat all errors as recoverable and enter a tight, infinite loop if the errors are persistent.

I propose we bail on errors by default and then keep a white-list of recoverable errors (such as receiving bad data from peers). This would be a band-aid solution until we can re-write the muxer code.

LesnyRumcajs (approved):
I need this.

PR #2687, opened 2023-03-20
fix(ci): cover statediff

hanabi1224:
## Summary of changes

- Cover statediff tool in CI

Closes

## Change checklist

      (if possible),
- [x] I have made sure the

### DIFF
--- a/cc.sh
+++ b/cc.sh
@@ -1,0 +1,9 @@
+#! /bin/bash
+
+TIPSET_CID=$(curl -X POST \
+    --header 'Content-Type: application/json' \
+    --data '{"jsonrpc":"2.0","method":"Filecoin.SyncState","params":null,"id":null}' \
+    http://localhost:1234/rpc/v0 \
+    | jq -r --arg SLASH "/" .result.ActiveSyncs[0].Base.Cids[0][$SLASH])
+echo "TIPSET_CID: $TIPSET_CID"
+cargo run -p forest_statediff --release -- chain --chain calibnet $TIPSET_CID $TIPSET_CID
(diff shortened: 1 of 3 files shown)


lemmih:
How did you select the state root CID and what happens when it is no longer retained?

hanabi1224:
I'm using genesis CID and it does not throw errors. What should be the correct usage?

lemmih:
> I'm using genesis CID and it does not throw errors. What should be the correct usage?

The correct usage requires a state-root mismatch. And right now, the diff tool cannot inspect state-roots that are older than NV17. The first line of output when you pass in the genesis CID is: `Could not resolve actor states: Can't create a valid state tree from the given root. This error may indicate unsupported version.`

Maybe a better test would be to diff the most recent state-root CID with itself.

hanabi1224:
@lemmih I've made the change to use latest tipset CID but the tool crashes

https://github.com/ChainSafe/forest/actions/runs/4477646093/jobs/7869432528?pr=2687#step:6:4075
```
Running `target/release/forest_statediff chain --chain calibnet bafy2bzacebu22zdq7paa3amyhkrvu2icrxzxhmopebatpywdwalcwgghtjcxy bafy2bzacebu22zdq7paa3amyhkrvu2icrxzxhmopebatpywdwalcwgghtjcxy`
StateDiff:
  Expected: bafy2bzacebu22zdq7paa3amyhkrvu2icrxzxhmopebatpywdwalcwgghtjcxy
  Root: bafy2bzacebu22zdq7paa3amyhkrvu2icrxzxhmopebatpywdwalcwgghtjcxy

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
./scripts/calibnet_health_check.sh: line 156: 31970 Aborted      
```

lemmih:
I think we should test the state diff command in other ways. How about we close this PR and re-group?

PR #2696, opened 2023-03-22
feat: add --track-peak-rss to daemon

hanabi1224:
## Summary of changes

RSS is included in Prometheus metrics but it's still helpful to track and print it when forest is not in long-running mode (e.g. `--halt-after-import`), so that we could easily get peak RSS usage from log when running `forest --import-snapshot [path] --halt-after-snapshot --track-peak-rss` without using an external monitoring script, to fill in benchmark tables

Closes

## Change checklist

      (if possible),
- [x] I have made sure the


forest/daemon/src/daemon.rs
```diff
@@ -73,6 +73,16 @@ pub(super) async fn start(opts: CliOpts, config: Config) -> anyhow::Result<Db> {
...
 
     set_sigint_handler();
+
+    let mem_stats_tracker = if opts.track_peak_rss {
+        Some(MemStatsTracker::default())
+    } else {
+        None
+    };
+    if let Some(mem_stats_tracker) = &mem_stats_tracker {
+        mem_stats_tracker.run_async();
```
lemmih, line 83: Can this background task be added to the services?

hanabi1224: Yes, but that will make its usage a bit complicated for other tools, e.g. statediff

lemmih: How about exposing the async loop and then letting the caller spawn a task? Or maybe just use `std::thread::spawn`.

hanabi1224: Fixed

lemmih: Thanks! Now I won't accidentally run the function from a non-async context. :)

LesnyRumcajs (approved):
I guess this may be helpful to track the daemonized process, but in general, for benchmark, I'd use `/usr/bin/time -v`. Application self-reporting memory metrics of itself is shady. I'm okay with merging this, though.

```
        Command being timed: "target/release/forest --chain calibnet"
        User time (seconds): 1.18
        System time (seconds): 0.57
        Percent of CPU this job got: 8%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:19.70
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 120804
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 32
        Minor (reclaiming a frame) page faults: 36876
        Voluntary context switches: 16040
        Involuntary context switches: 347
        Swaps: 0
        File system inputs: 900872
        File system outputs: 89856
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

hanabi1224:
@LesnyRumcajs Thanks! That's a much better option, I kinda confused `/usr/bin/time` with `time` which does not support `-v`  : )

lemmih (approved):

Issue #2029, opened 2022-10-10
Document `forest send` in the user's-guide and in the README


lemmih:
**Issue summary**

We need document how to get started with Forest. The guides should include: how to install Forest, how to sync to calibnet (with an automatically downloaded snapshot), how to create a new address in the wallet, how to fund it from the faucet, and how to send FIL to other addresses.

**Other information and links**

Blocked by #2009 and #2000.

PR #2700, opened 2023-03-22
Align Forest and Lotus `send` behavior

### LINKED ISSUE #2668: Align Forest send command behavior with the one from Lotus
**Issue summary**

Lotus send [command](https://lotus.filecoin.io/lotus/manage/lotus-cli/#lotus-send) allows to specify the token amount in either `FIL` or `attoFIL` denominations.
That's not the case for Forest send command where it expects `attoFIL`.

- [x] Add support of similar behaviour.
- [x] Add tests covering the different denominations.
- [x] Update the documentation accordingly.
- [x] Make sure `sendFIL` still works in the JavaScript console

https://github.com/filecoin-project/lotus/blob/master/cli/send.go#L96
https://github.com/filecoin-project/lotus/blob/master/chain/types/fil.go#L86

jdjaustin:
## Summary of changes

- Add support to Forest to specify `FIL` in all units supported in Forest wallet in the amount specified in the `send` command. Subtasks:
- [x] Add support of similar behaviour.
- [x] Add tests covering the different denominations.
- [x] ~~Update the documentation accordingly.~~ (will be covered under issue #2029 )
- [x] ~~Make sure `sendFIL` still works in the JavaScript console~~ (will be covered under another PR)

Closes associated tasks in issue #2668 

Task to update documentation will be covered under issue #2029 and task to add support for multiple units in JavaScript console will be covered under another PR tied to issue #2668.

## Change checklist

- [x] I have made corresponding changes to the documentation, (see note above)
      (if possible),
- [x] I have made sure the


forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,98 @@ use forest_json::message::json::MessageJson;
...
+
+        let parsed_val = match Decimal::from_str(val) {
+            Ok(value) => value,
+            Err(_) => {
+                return Err(anyhow::anyhow!(
+                    "failed to parse {} as a decimal number",
+                    val
+                ))
+            }
+        };
```
lemmih, line 83:
```suggestion
        let parsed_val = Decimal::from_str(val).map_err(|_| anyhow::anyhow!(
                    "failed to parse {} as a decimal number",
                    val)
                )?;
```

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,105 @@
...
+        //fil amount with length>50 fails
+        let amount = "100000000000000000000000000000000000000000000000000FIL";
+        assert!(FILAmount::from_str(amount).is_err());
+    }
+
+    #[test]
+    fn convert_fil_to_attofil() {
+        //expected attofil amount matches actual amount after conversion from FIL
+        let fil_amount = "1FIL";
+        let attofil_amount = TokenAmount::from_atto(BigInt::from(FILECOIN_PRECISION));
```
lemmih, line 76:
```suggestion
        let attofil_amount = TokenAmount::from_whole(1);
```

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,98 @@ use forest_json::message::json::MessageJson;
...
+impl FromStr for FILAmount {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let suffix_idx = s.rfind(char::is_numeric);
+        let (val, suffix) = match suffix_idx {
+            Some(idx) => s.split_at(idx + 1),
+            None => return Err(anyhow::anyhow!("failed to parse string: {}", s)),
+        };
+        let suffix = suffix.replace(' ', "");
```
lemmih, line 29: Does this mean that, say, `fem to fil` is valid?

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,105 @@
...
+        //fails with negative value
+        let amount = "-1FIL";
+        assert!(FILAmount::from_str(amount).is_err());
+    }
+
+    #[quickcheck]
+    fn fil_quickcheck_test(n: u64) {
+        let token_amount = TokenAmount::from_atto(n);
+        let formatted =
+            format_balance_string(token_amount.clone(), bool_pair_to_mode(true, false)).unwrap();
```
lemmih, line 101: This might be easier to read:
```suggestion
            format_balance_string(token_amount.clone(), FormattingMode::ExactNotFixed).unwrap();
```

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,98 @@ use forest_json::message::json::MessageJson;
...
+
+        let mut multiplier = dec!(1.0);
+        let prefix = if !suffix.is_empty() {
```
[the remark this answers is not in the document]
jdjaustin: Removing it causes the `valid_fil_amount_without_suffix` test to fail. Adding a `None` arm to the match statement causes the `invalid_fil_suffix` test to fail.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,88 @@ use forest_json::message::json::MessageJson;
...
+
+        if val.chars().count() > 50 {
+            anyhow::bail!("string length too large: {}", val.chars().count());
+        }
+
+        let parsed_val = Decimal::from_str(val)
+            .map_err(|_| anyhow::anyhow!("failed to parse {} as a decimal number", val))?;
+
+        let attofil_val = if (parsed_val * multiplier).fract() != dec!(0.0) {
+            return Err(anyhow::anyhow!("invalid {}FIL value: {}", prefix, val));
```
lemmih, line 76: Here you can just say that the FIL amount has to be a whole number of atto FIL. This lets you remove the `prefix` variable which will simplify your code a lot.

forest/cli/src/cli/send_cmd.rs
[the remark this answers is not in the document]
lemmih: How about something like:
```rust
        let suffix = suffix.trim().to_lowercase();
        let prefix = match suffix.strip_suffix("fil").map(str::trim).unwrap_or(&suffix) {
            "atto" | "a" => "atto",
            "femto" => {
                multiplier *= dec!(1_000);
                "femto"
            }
            "pico" => {
                multiplier *= dec!(1_000_000);
                "pico"
            }
            "nano" => {
                multiplier *= dec!(1_000_000_000);
                "nano"
            }
            "micro" => {
                multiplier *= dec!(1_000_000_000_000);
                "micro"
            }
            "milli" => {
                multiplier *= dec!(1_000_000_000_000_000);
                "milli"
            }
            "" => {
                multiplier *= dec!(1_000_000_000_000_000_000);
                ""
            }
            _ => anyhow::bail!("unrecognized suffix: {}", suffix)
        };
```

forest/cli/src/cli/send_cmd.rs
jdjaustin: It did erroneously allow an amount with units `fem to fil`, but I have corrected that.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,85 @@ use forest_json::message::json::MessageJson;
...
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let suffix_idx = s.rfind(char::is_numeric);
+        let (val, suffix) = match suffix_idx {
+            Some(idx) => s.split_at(idx + 1),
+            None => anyhow::bail!("failed to parse string: {}", s),
+        };
+
+        let mut multiplier = dec!(1.0);
```
lemmih, line 30: This doesn't need to be mutable.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,85 @@ use forest_json::message::json::MessageJson;
...
+        }
+
+        let parsed_val = Decimal::from_str(val)
+            .map_err(|_| anyhow::anyhow!("failed to parse {} as a decimal number", val))?;
+
+        let attofil_val = if (parsed_val * multiplier).normalize().scale() != 0 {
+            anyhow::bail!("{} must convert to a whole attoFIL value", &s);
+        } else {
+            (parsed_val * multiplier).trunc().to_u128()
+        };
```
lemmih, line 74: You can refactor this code to be smaller. You can also change it to compute an `i128` instead of an `Option<i128>`.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,85 @@ use forest_json::message::json::MessageJson;
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let suffix_idx = s.rfind(char::is_numeric);
+        let (val, suffix) = match suffix_idx {
+            Some(idx) => s.split_at(idx + 1),
+            None => anyhow::bail!("failed to parse string: {}", s),
+        };
+
+        let mut multiplier = dec!(1.0);
+        let suffix = suffix.replacen(' ', "", 1).trim().to_lowercase();
```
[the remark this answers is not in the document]
jdjaustin: It was previously `replace(' ', "")`, but as you pointed out, that allowed amounts with units such as `fem to fil` to be parsed as valid inputs. This was my solution for that issue.

forest/cli/src/cli/send_cmd.rs
jdjaustin: If I remove the `mut`, I get a compiler error that it needs to be mutable due to the mutable borrows in the match statement.

jdjaustin: But I think I see what you're suggesting now.

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    #[test]
+    fn negative_fil_value() {
+        //fails with negative value
+        let amount = "-1FIL";
+        assert!(FILAmount::from_str(amount).is_err());
+    }
+
+    #[quickcheck]
+    fn fil_quickcheck_test(n: u64) {
```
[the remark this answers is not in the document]
jdjaustin: I've created a quickcheck test that generates random values between 0 and 1000. I'm not certain I understand what you mean by "a value that is equally likely to be measured in atto fil, femto fil, pico fil, nano fil, micro fil, milli fil, fil, kilo fil, etc." IIUC, we can't have fractional attofil values, but we can have fractional femto/pico/nano, etc., right?

forest/cli/src/cli/send_cmd.rs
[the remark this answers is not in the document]
lemmih: But why is it there in the first place? Isn't trimming the suffix enough?

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,120 @@
...
+        //attoFIL with fractional value fails (fractional FIL values allowed)
+        let amount = "1.234attofil";
+        assert!(FILAmount::from_str(amount).is_err());
+    }
+
+    #[test]
+    fn valid_attofil_amount() {
+        //valid attofil amount passes
+        let amount = "1234 attofil";
+        assert!(FILAmount::from_str(amount).is_ok());
```
lemmih, line 26:
```suggestion
        assert_eq!(FILAmount::from_str(amount).unwrap(), TokenAmount::from_atto(1234));
```

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    #[test]
+    fn invalid_attofil_amount() {
+        //attoFIL with fractional value fails (fractional FIL values allowed)
+        let amount = "1.234attofil";
+        assert!(FILAmount::from_str(amount).is_err());
+    }
+
+    #[test]
+    fn valid_attofil_amount() {
```
lemmih, line 23: Let's also test if `1234 afil` and `1234 a fil` parse.

lemmih: Oh, and also just `1234 a`.

forest/cli/src/cli/send_cmd.rs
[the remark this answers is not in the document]
jdjaustin: Ah yepp apparently trimming is sufficient, I guess this was an artifact from when I was using the `prefix` in the error message.

forest/cli/tests/send_tests.rs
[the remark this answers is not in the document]
lemmih: Yep, exactly.

forest/cli/tests/send_tests.rs
lemmih: By looking at your code, I would have guessed all three of them to be valid.

jdjaustin: The initial spec for my code was to attempt to align with the code/behavior in Lotus as closely as possible, but that changed during development, and the code is still a WIP. So the question now is: what behavior do we want from Forest?

jdjaustin: What do you think about `a` or `a fil`?

lemmih: I think both are fine.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,76 @@ use forest_json::message::json::MessageJson;
...
+        }
+
+        let parsed_val = Decimal::from_str(val)
+            .map_err(|_| anyhow::anyhow!("failed to parse {} as a decimal number", val))?;
+
+        let val = parsed_val * multiplier;
+        if val.normalize().scale() > 0 {
+            anyhow::bail!("{} must convert to a whole attoFIL value", &s);
+        }
+        let attofil_val = val.trunc().to_u128().ok_or(anyhow::Error::msg(
```
lemmih, line 68: It's a bit of a hack but you can also do this:
```suggestion
        let attofil_val = val.normalize().mantissa();
```

lemmih: If you do use it, write in the comments that it is only valid because `scale=0`.

lemmih: Okay, then let's change the error message to say something about negative fil amounts are not valid.

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn suffix_with_no_amount() {
+        //fails if no amount specified
+        let amount = "fil";
+        assert!(FILAmount::from_str(amount).is_err());
+    }
+    #[test]
+    fn valid_fil_amount_without_suffix() {
+        //defaults to FIL if no suffix is provided
+        let amount = "1234";
+        assert!(FILAmount::from_str(amount).is_ok());
```
lemmih, line 73: Can assert that it parses to `TokenAmount::from_whole(1234)`.

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,155 @@
...
+        let token_amount = TokenAmount::from_atto(n);
+        let formatted =
+            format_balance_string(token_amount.clone(), FormattingMode::ExactNotFixed).unwrap();
+        let parsed = FILAmount::from_str(&formatted).unwrap().value;
+        assert_eq!(token_amount, parsed);
+    }
+
+    #[quickcheck]
+    fn scaled_fil_quickcheck_test(n: u64) {
+        let rand_num = rand::thread_rng().gen_range(0..20);
```
lemmih, line 147: You're not allowed to use non-determinism in quickcheck tests. The tests have to be determined by their arguments. Otherwise, they won't be reproducible.

You already get one random value: `n`. How about getting `rand_num` in the same way?

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,76 @@ use forest_json::message::json::MessageJson;
...
+            Some(idx) => s.split_at(idx + 1),
+            None => anyhow::bail!("failed to parse string: {}", s),
+        };
+
+        let suffix = suffix.trim().to_lowercase();
+        let multiplier = match suffix.strip_suffix("fil").map(str::trim).unwrap_or(&suffix) {
+            "atto" | "a" => dec!(1e0),
+            "femto" => {
+                dec!(1e3)
+            }
```
lemmih, line 35: All the branches can be shortened:
```suggestion
            "femto" => dec!(1e3),
```

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,76 @@ use forest_json::message::json::MessageJson;
...
+        };
+
+        // string length check to match Lotus logic
+        if val.chars().count() > 50 {
+            anyhow::bail!("string length too large: {}", val.chars().count());
+        }
+
+        let parsed_val = Decimal::from_str(val)
+            .map_err(|_| anyhow::anyhow!("failed to parse {} as a decimal number", val))?;
```
lemmih, line 62: The error returned by `Decimal::from_str` can be quite useful. Let's include it in the error message.

forest/cli/tests/send_tests.rs
```diff
@@ -0,0 +1,155 @@
...
+        let scaled_n = n % u64::pow(10, rand_num);
+        let token_amount = TokenAmount::from_atto(scaled_n);
+        let formatted =
+            format_balance_string(token_amount.clone(), FormattingMode::ExactNotFixed).unwrap();
+        let parsed = FILAmount::from_str(&formatted).unwrap().value;
+        assert_eq!(token_amount, parsed);
```
lemmih, line 153: This code looks like `fil_quickcheck_test`. How about we call that function here?

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,76 @@ use forest_json::message::json::MessageJson;
...
+            }
+            _ => {
+                anyhow::bail!("unrecognized suffix: {}", suffix);
+            }
+        };
+
+        // string length check to match Lotus logic
+        if val.chars().count() > 50 {
+            anyhow::bail!("string length too large: {}", val.chars().count());
+        }
```
lemmih, line 59: This check could be moved up in the code. There's no need to parse the suffix if we're just gonna bail due to the input length.

lemmih (changes requested):
Getting close. Just needs a few more things:

- Documentation for `FILAmount::from_str`.
- More consistent error messages. Try adding a local helper function with an error format like this `failed to parse fil amount: {input_string}. {failure_reason}`. Then the other parts of the code can emit different `failure_reason`s: `Unrecognized suffix: {}`, `Too many digits (maximum 50)`, `Fractional attoFILs are not allowed`, `Negative fil amounts are not allowed`, etc. You can directly use the error message from `Decimal::from_str` as an error reason.
- Check that the error messages are correct in the test cases.
- Documentation for `fil_quickcheck_test` and `scaled_fil_quickcheck_test`.
- Extend `fil_quickcheck_test` to also format the input number using `ExactFixed`. This new string should be parsable to the same input number.
- Add a quickcheck tests that generate random strings and checks if `FILAmount::from_str` panics.

forest/cli/src/cli/mod.rs
```diff
@@ -14,12 +14,12 @@ mod db_cmd;
...
 mod mpool_cmd;
 mod net_cmd;
-mod send_cmd;
+pub mod send_cmd;
 mod shutdown_cmd;
 mod snapshot_cmd;
 mod state_cmd;
 mod sync_cmd;
-mod wallet_cmd;
+pub mod wallet_cmd;
```
lemmih, line 22: These modules shouldn't be public. Instead of using public modules, you can move your tests into the modules.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,76 @@ use forest_json::message::json::MessageJson;
...
+
+impl FromStr for FILAmount {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let suffix_idx = s.rfind(char::is_numeric);
+        let (val, suffix) = match suffix_idx {
+            Some(idx) => s.split_at(idx + 1),
+            None => anyhow::bail!("failed to parse string: {}", s),
+        };
```
lemmih, line 28:
```suggestion
        let suffix_idx = s.rfind(char::is_numeric).ok_or_else(|| anyhow::anyhow!("failed to parse fil amount: {}. No digits.", s))?;
        let (val, suffix) = s.split_at(suffix_idx + 1);
```

lemmih:
In my not-so-humble opinion, `rfind+is_numeric+split_at` is greatly preferable to a regex. Better performance, fewer failure points, fewer dependencies, fewer lines of code.

LesnyRumcajs:
> In my not-so-humble opinion, `rfind+is_numeric+split_at` is greatly preferable to a regex. Better performance, fewer failure points, fewer dependencies, fewer lines of code.

Performance is better without regular expressions, but it doesn't matter here unless I'm missing something. Dependencies - we already have `regex` in our tree. 

Personally, I find this more readable. But I guess it's a matter of preferences. :)
https://gist.github.com/LesnyRumcajs/1f431903c0acd845c4992c07efc15cd4

lemmih:
> > In my not-so-humble opinion, `rfind+is_numeric+split_at` is greatly preferable to a regex. Better performance, fewer failure points, fewer dependencies, fewer lines of code.
> 
> Performance is better without regular expressions, but it doesn't matter here unless I'm missing something. Dependencies - we already have `regex` in our tree.
> 
> Personally, I find this more readable. But I guess it's a matter of preferences. :) https://gist.github.com/LesnyRumcajs/1f431903c0acd845c4992c07efc15cd4

So it would be:
```rust
    let re = Regex::new(r"^(\d+(?:\.\d+)?)\s*([[:alpha:]]+)?$")?;
    if let Some(captures) = re.captures(&s.to_lowercase()) {
        let value = captures
            .get(1)
            .ok_or_else(|| anyhow::anyhow!("fiasco"))?
            .as_str();
        let suffix = captures.get(2).map(|m| m.as_str());
        ...
    }
    bail!("Invalid input")
```
vs:
```rust
    let suffix_idx = s.rfind(char::is_numeric).ok_or_else(||
            anyhow::anyhow!("failed to parse fil amount: {}. No digits.", s))?;
    let (value, suffix) = s.split_at(suffix_idx + 1);
    ...
```

The rest of the code is the same. 10 lines vs 4 lines. I like the shorter version. :)

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,81 @@ use forest_json::message::json::MessageJson;
...
+/// 'amount' field of the the `send` command) with all units supported by forest
+/// wallet ("attoFIL", "femtoFIL", "picoFIL", "nanoFIL", "microFIL", "milliFIL",
+/// and "FIL") and converting the amount into an attoFIL `TokenAmount`. To match
+/// the current behavior in Lotus, the default units (if no units are specified
+/// in the command line) are "FIL".
+impl FromStr for FILAmount {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        fn error_call<T: std::fmt::Display>(s: &str, e: T) -> anyhow::Error {
```
lemmih, line 30: Why is `e` not a `&str`?

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,81 @@ use forest_json::message::json::MessageJson;
...
+            anyhow::anyhow!("failed to parse fil amount: {}. {}.", &s, e)
+        }
+
+        let suffix_idx = s
+            .rfind(char::is_numeric)
+            .ok_or_else(|| anyhow::anyhow!("failed to parse fil amount: {}. No digits.", s))?;
```
lemmih, line 36: Looks like you want to use `error_call` here.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,81 @@ use forest_json::message::json::MessageJson;
...
+        let (val, suffix) = s.split_at(suffix_idx + 1);
+
+        // string length check to match Lotus logic
+        if val.chars().count() > 50 {
+            return Err(error_call(
+                s,
+                anyhow::Error::msg(format!("string length too large: {}", val.chars().count())),
```
lemmih, line 43: I don't think you want to use `anyhow::Error::msg` here.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,81 @@ use forest_json::message::json::MessageJson;
...
+            "nano" => dec!(1e9),
+            "micro" => dec!(1e12),
+            "milli" => dec!(1e15),
+            "" => dec!(1e18),
+            _ => {
+                return Err(error_call(
+                    s,
+                    anyhow::Error::msg(format!("unrecognized suffix: {}", suffix)),
+                ));
+            }
```
lemmih, line 61: You can skip the curly brackets just like in the other branches.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,81 @@ use forest_json::message::json::MessageJson;
...
+
+        let val = parsed_val * multiplier;
+        if val.normalize().scale() > 0 {
+            return Err(error_call(
+                s,
+                anyhow::Error::msg(format!("{} must convert to a whole attoFIL value", &val)),
+            ));
+        }
+        let attofil_val = val.trunc().to_u128().ok_or(error_call(s, anyhow::Error::msg(
+            "Could not covert amount input to send into an attoFIL value (note that negative FIL amounts are not valid)"),
```
lemmih, line 74: This error can _only_ happen when the user types in negative numbers. Let's update the error message to be more direct and concise.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,81 @@ use forest_json::message::json::MessageJson;
...
+        ))?;
+
+        let token_amount = TokenAmount::from_atto(BigInt::from(attofil_val));
```
lemmih, line 77: You don't need the `BigInt` and `token_amount` is only used once so you can just inline it.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,81 @@ use forest_json::message::json::MessageJson;
...
+pub struct FILAmount {
+    pub value: TokenAmount,
+}
+
+/// `FILAmount::from_str` adds support for parsing a string (input to the
+/// 'amount' field of the the `send` command) with all units supported by forest
+/// wallet (`attoFIL`, `femtoFIL`, `picoFIL`, `nanoFIL`, `microFIL`, `milliFIL`,
+/// and `FIL`) and converting the amount into an attoFIL `TokenAmount`. To match
+/// the current behavior in Lotus, the default units (if no units are specified
+/// in the command line) are `FIL`.
```
lemmih, line 25: This documentation is for the `form_str` function and it should just explain what the function does. In other words, the function doesn't "add support". The function just parses a string in a certain way.

forest/cli/src/cli/send_cmd.rs
jdjaustin: So `{failure_reason}` would be something like `format!("Unrecognized suffix: {}", suffix).as_str()`?

lemmih: The type of `&format!(...)` is `&str`.

jdjaustin: Clippy didn't like `format!` anyway so I used `to_string()` and as the closure is able to take a `String`.

jdjaustin:
> Getting close. Just needs a few more things:
> 
>     * Documentation for `FILAmount::from_str`.
> 
>     * More consistent error messages. Try adding a local helper function with an error format like this `failed to parse fil amount: {input_string}. {failure_reason}`. Then the other parts of the code can emit different `failure_reason`s: `Unrecognized suffix: {}`, `Too many digits (maximum 50)`, `Fractional attoFILs are not allowed`, `Negative fil amounts are not allowed`, etc. You can directly use the error message from `Decimal::from_str` as an error reason.
> 
>     * Check that the error messages are correct in the test cases.
> 
>     * Documentation for `fil_quickcheck_test` and `scaled_fil_quickcheck_test`.
> 
>     * Extend `fil_quickcheck_test` to also format the input number using `ExactFixed`. This new string should be parsable to the same input number.
> 
>     * Add a quickcheck tests that generate random strings and checks if `FILAmount::from_str` panics.

Remaining to do from this list:
- Documentation for `fil_quickcheck_test` and `scaled_fil_quickcheck_test`.
- Add a quickcheck tests that generate random strings and checks if `FILAmount::from_str` panics.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,69 @@ use forest_json::message::json::MessageJson;
...
+
+impl FromStr for FILAmount {
+    type Err = anyhow::Error;
+
+    /// `FILAmount::from_str` parses a string (input to the 'amount' field of
+    /// the the `send` command) with all units supported by forest wallet
+    /// (`attoFIL`, `femtoFIL`, `picoFIL`, `nanoFIL`, `microFIL`,
+    /// `milliFIL`, and `FIL`) and converts the amount into an `attoFIL`
+    /// `TokenAmount`. To match the current behavior in Lotus, the default
+    /// units (if no units are specified in the command line) are `FIL`.
```
lemmih, line 28: The documentation should describe what this function does. This means you should not mention the `send` command at all. That this function happens to be used by the `send` command is irrelevant.

Describe what this function does. Show some positive examples. Show some negative examples. Talk about constraints, the relationship to `format_balance_string`, and mention that we want to accept all input accepted by Lotus as a footnote.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,69 @@ use forest_json::message::json::MessageJson;
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let error_call = |e| anyhow::anyhow!("failed to parse fil amount: {}. {}.", &s, e);
```
lemmih, line 30: Is it better for `e` to be a `String` or `&str`?

lemmih (changes requested):

forest/cli/src/cli/send_cmd.rs
jdjaustin: I phoned a friend, who told me it's better to be a`&str` to avoid the heap allocation (and as a bonus, this removes all the calls to `to_string()`). Just seems that `e` needs to be explicity typed in the closure or the borrow in the `parsed_val` call complains.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -7,9 +7,75 @@ use forest_json::message::json::MessageJson;
...
+    /// attoFIL (e.g., `1.234 attoFIL` is invalid while `1234 attoFIL` is valid
+    /// and parses to a `TokenAmount` corresponding to `1234` `attoFIL`). To
+    /// match the current behavior in Lotus, the default units are `FIL`
+    /// (e.g., `1` parses as `FIL` with a resulting `TokenAmount` corresponding
+    /// to `1e18` `attoFIL`). The function also accepts all units currently
+    /// supported by forest wallet in the `format_balance_string` function
+    /// (`attoFIL`, `femtoFIL`, `picoFIL`, `nanoFIL`, `microFIL`, `milliFIL`,
+    /// and `FIL`), and the output of `FILAmount::from_str` can be used as input
+    /// to the `format_balance_string` function.
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
lemmih, line 37: - This documentation is attached to `FILAmount::from_str` so you don't need to refer back to the function.
- I don't know what this means: "an attoFIL TokenAmount".
- It's best if you cover the positive definition first and then briefly cover the negative definition as a separate paragraph.
- This is not correct: "Fractional input values are valid unless the input units are attoFIL". For example, this input is invalid "0.0001 femto fil". That is because "0.0001 femto fil" is the same as "0.1 atto fil".
- The note about Lotus isn't wrong but it's not the only behaviour that matches Lotus. It's shorter and more precise: We accept all input strings accepted by Lotus.
- You got the relationship between `FILAmount::from_str` and `format_balance_string` the wrong way around. The type-system statically guarantees that the output of `FILAmount::from_str` can be used by `format_balance_string`. The interesting relationship is whether the output of `format_balane_string` can be parsed by `FILAmount::from_str`.
- In general, text documentation should be short and to-the-point.

Try something like this:

Parse a string like `10 attoFIL` or `0.5 milliFIL` into a `TokenAmount`. Known units are `aFIL`, `attoFIL`, `femtoFIL`, `picoFIL`, `nanoFIL`, `microFIL`, and `milliFIL`. These are the same units used by `format_balance_string`.

The `FIL` suffix may be omitted. The parsed `TokenAmount` may not be negative, contain a fractional number of attoFIL, or be longer than 50 digits. Any input parseable by Lotus is considered to be valid.

Examples:
```rust
// Without a unit, numbers are interpreted as FIL
assert_eq!(
  FILAmount::from_str("1.5").unwrap(),
  FILAmount::from_str("1.5 FIL").unwrap(),
);
```

```rust
// aFIL is a synonym of attoFIL
assert_eq!(
  FILAmount::from_str("10 attoFIL").unwrap(),
  FILAmount::from_str("10 aFIL").unwrap(),
);
```

```rust
// Suffixes are case insensitive and may omit `FIL`
assert_eq!(
  FILAmount::from_str("0.5 milli").unwrap(),
  FILAmount::from_str("0.5 MiLlI FIL").unwrap(),
);
```
```rust
// Negative values are invalid
assert!(FILAmount::from_str("-0.5 FIL").is_err());
```

```rust
// Fractional attoFIL amounts are invalid
assert!(FILAmount::from_str("0.0001 picoFIL").is_err());
```

lemmih (changes requested):

.config/forest.dic
```diff
@@ -44,6 +44,8 @@ MiB
 multisig
 mutex
 P2P
+parsable
+parseable
```
lemmih, line 48: They mean the same thing, right? Let's pick one or the other.

lemmih (changes requested):

.config/forest.dic
jdjaustin: Yepp they mean the same thing and both are valid spellings with similar usage frequency. I'll go with `parsable` as [it's the preferred standard form](https://en.wiktionary.org/wiki/parsable).

lemmih (approved):

forest/cli/src/cli/send_cmd.rs
```diff
@@ -71,3 +181,208 @@ impl SendCommand {
...
+        );
+    }
+
+    #[test]
+    fn invalid_fil_suffix() {
+        //fails with bad suffix
+        let amount = "42fiascos";
+        assert_eq!(
+            FILAmount::from_str(amount).unwrap_err().message(),
+            "failed to parse fil amount: 42fiascos. Unrecognized suffix."
```
elmattic, line 328: It would be nice to put in the error message the offending suffix as well.

forest/cli/src/cli/send_cmd.rs
```diff
@@ -71,3 +181,208 @@ impl SendCommand {
...
+        assert_eq!(
+            FILAmount::from_str(amount).unwrap().value,
+            TokenAmount::from_atto(1234)
+        );
+    }
+
+    #[test]
+    fn valid_attofil_amount_test3() {
+        //valid attofil amount passes
+        let amount = "1234 a fil";
```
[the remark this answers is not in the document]
lemmih: Nope, they only support `fil`, `attofil` and `afil`. We're deliberately more liberal in what we accept.

elmattic (approved):

Issue #2656, opened 2023-03-13
Fix spellchecking for every member crate


lemmih:
**Issue summary**

Our spellchecker has stopped working correctly and only spellchecks a few source files. This can be seen by running `cargo spellcheck list-files`. It should list every single source file in our repository but it doesn't. We need to figure out what went wrong and fix it.

**Other information and links**

tinoh9:
Hi bro, have you resolved this issue? I also had a look to the repo but still did not find any invalid config or conflict. By the way, I found this alternative (https://github.com/crate-ci/typos) which is also wrote in Rust if this get you any attention.

lemmih:
Still not fixed. The `typos` repository looks interesting but it is orthogonal to spellchecking in the documentation.

LesnyRumcajs:
Hey @tinoh9, it seems the culprit in our case is `edition.workspace` which is not handled well by the spellcheck or its dependencies. The workaround is to put the edition explicitly. I hope it helps.

https://github.com/ChainSafe/forest/pull/2702

tinoh9:
It is great that this issue has been solved. Really appreciate that I learn new debugging concept from you.

PR #2702, opened 2023-03-23
do not use workspace edition

LesnyRumcajs:
## Summary of changes

- It seems the `spellcheck` or its dependency do not play well with `edition.workspace`. It's a single `sed` change anyway, so let's put 2021 everywhere explicitly to re-enable spellcheck.

Closes https://github.com/ChainSafe/forest/issues/2656

## Change checklist

      (if possible),
- [x] I have made sure the


lemmih:
Thank you for fixing the spellchecking. Do you want me to address the spelling issues?

LesnyRumcajs:
> Thank you for fixing the spellchecking. Do you want me to address the spelling issues?

I added the `edition.workspace = true` some time ago, so it's only fair I clean up the mess. :)

.config/forest.dic
```diff
@@ -27,28 +35,34 @@ JWT
...
 mempool
 Merkle
+MiB
 multisig
 mutex
 P2P
 pointer/SM
 PoSt
 RPC
+RRS
```
lemmih, line 49: What is `RRS` and where is it used?

lemmih: Oh, in the mem tracker.

lemmih: I think that is a typo. It should be RSS (resident set size). And I'd prefer if we wrote out the abbreviation (because I always forget what it means).

hanabi1224 (approved):
Is it possible to add a regression test for the spellchecker?

LesnyRumcajs:
> Is it possible to add a regression test for the spellchecker?

It is possible, but it'll be like testing clippy, udeps or any other linter. Aside from that, the maintainer has just fixed this https://github.com/drahnr/cargo-spellcheck/issues/301#issuecomment-1480924512

PR #2723, opened 2023-03-29
feat: add --no-gc option to forest daemon

hanabi1224:
## Summary of changes

-

Closes

## Change checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
      (if possible),
- [ ] I have made sure the


lemmih (approved):

forest/daemon/src/daemon.rs
```diff
@@ -180,11 +180,13 @@ pub(super) async fn start(opts: CliOpts, config: Config) -> anyhow::Result<Rolli
         Arc::new(DbGarbageCollector::new(db, get_tipset))
     };
 
-    #[allow(clippy::redundant_async_block)]
-    services.spawn({
-        let db_garbage_collector = db_garbage_collector.clone();
-        async move { db_garbage_collector.collect_loop_passive().await }
-    });
+    if !opts.no_gc {
+        #[allow(clippy::redundant_async_block)]
```
LesnyRumcajs, line 184: Why do we need to add those `redundant_async_block` exceptions in so many places? I don't think it was the case before.

LesnyRumcajs (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@
 
 ### Added
 
+- `--no-gc` flag to daemon.
+  [#2721](https://github.com/ChainSafe/forest/issues/2721)
```
LesnyRumcajs, line 31: It'd be great to follow the guidelines above, that is to say, issue first, then a short description.

forest/daemon/src/daemon.rs
hanabi1224: That's a new clippy rule but the suggested fix does not work because of some lifetime issue

CHANGELOG.md
hanabi1224: Hmm, the format I use follows what is in the 0.7.0 section. Would u suggest keep them consistent for now and make changes altogether or just change it in this PR?

LesnyRumcajs: I added the guideline after v.0.7.0 was released. It's too much hassle to rewrite everything from the past, but following the guideline now would be great.

PR #2724, opened 2023-03-29
fix(gc): do not run GC when heaviest tipset is not stored in current db space

hanabi1224:
## Summary of changes

Changes introduced in this pull request:

- do not run GC when heaviest tipset is not stored in current db space, otherwise, forest will run into some unrecoverable state

## Reference issue to close (if applicable)

Closes

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works
      (if possible),
- [x] I have made sure the
      [CHANGELOG](https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md) is
      up-to-date. All user-facing changes should be reflected in this document.

PR #2725, opened 2023-03-29
feat: add `fn heaviest_finalized_tipset` to `ChainStore`

hanabi1224:
## Summary of changes

Add `fn heaviest_finalized_tipset` to `ChainStore` so that during GC we could check and ensure `heaviest_finalized_tipset` exists in the current DB space and never get deleted.   Change the [tipset check logic](https://github.com/ChainSafe/forest/pull/2724/files#diff-b36dc5aaddbd079edb780b9259815c30048d1a0fd67aa2b0201494af6edd17bdR184) here once the other PR is merged

### Assumptions:
1. The chain can be split into 2 parts:
  a. epoch(genesis) to epoch(head - FINALITY(900)), which is immutable.
  b. epoch(head - FINALITY(900)) to epoch(head), which is mutable.
2. It could cause problems when GC deletes data from the mutable part of the chain
3. It will not cause problems when GC only deletes data from the immutable part of the chain

### Solution:
Run GC only when the entire mutable part of the chain is stored in the current DB space and will not be deleted.
(Above will be included in the subsequent PR that implements the check)

Closes

## Change checklist

      (if possible),
- [x] I have made sure the

### DIFF
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -242,6 +242,19 @@ where
             .expect("Failed to load heaviest tipset")
     }
 
+    /// Returns the heaviest finalized tipset by walking back from the currently
+    /// tracked heaviest tipset by FINALITY(900) epochs.
+    pub fn heaviest_finalized_tipset(&self) -> anyhow::Result<Arc<Tipset>> {
+        const FINALITY_EPOCHS: i64 = 900;
+
+        let mut ts = self.heaviest_tipset();
+        for _i in 0..FINALITY_EPOCHS {
+            ts = self.tipset_from_keys(ts.parents())?;
+        }
+
+        Ok(ts)
+    }
+
     /// Returns a reference to the publisher of head changes.
     pub fn publisher(&self) -> &Publisher<HeadChange> {
         &self.publisher


lemmih:
I struggle to understand what you're trying to do here. Could you explain your plan to me?

lemmih:
The implications of your code are not obvious to me. Getting these things exactly right is very difficult and I would greatly appreciate if you wrote a design document before moving ahead with the implementation. The design document should explain the problem and the convince the reader that your solution is correct, robust, and won't cause surprises in the future.

hanabi1224:
@lemmih I've updated the assumptions and solution in PR desc, let me know if it looks good.

lemmih:
> @lemmih I've updated the assumptions and solution in PR desc, let me know if it looks good.

It's a good start.

What if I accidentally copy the heaviest finalized tipset from the "old" partition to the "new" partition? There's nothing stopping me from doing this but it would cause the GC to incorrectly believe it's safe to do a collection.

hanabi1224:
I think it's not too expensive to implement the check in a more defensive way by checking all 900 epochs. Sounds good?

hanabi1224:
> However, it is probably quite unlikely that someone will copy over all the spines.

@lemmih I think that's more of an intended attack instead of a manual mistake, which I'm not sure if it's within the scope of DB module.  BTW, paritydb won't allow opening the same directory for writing, so that to achive such kind of attack, one has to shutdown forest, do malicious DB ops then restart forest

lemmih:
> > However, it is probably quite unlikely that someone will copy over all the spines.
> 
> @lemmih I think that's more of an intended attack instead of a manual mistake, which I'm not sure is within the scope of DB module or not. BTW, paritydb won't allow opening the same directory for writing, so that to achive such kind of attack, one has to shutdown forest, do malicious DB ops then restart forest

Why would it have to be an attack? It can easily be done with the current DB interface and doesn't require any evil looking code.

hanabi1224:
Superseded by https://github.com/ChainSafe/forest/pull/2733

PR #2730, opened 2023-03-29
chore: Add a closer tipset checkpoint on calibnet

creativcoder:
## Summary of changes

- Adds a closer tipset checkpoint at epoch 424000 for calibnet.

Closes

## Change checklist

- [X] I have added tests that prove my fix is effective or that my feature works
      (if possible),
- [x] I have made sure the


lemmih (approved):
Nice. Could you add a new checkpoint for mainnet as well?

LesnyRumcajs (approved):
An idea for a future task - can we automate it somehow? Make it use a YAML file with all those definitions, then contrive a CLI tool to update it with shiny, new checkpoints. What do you think? @lemmih @creativcoder

PR #2743, opened 2023-04-03
feat: improve checkpoint procedure

### LINKED ISSUE #2735: Load checkpoint hashes from config and automate updates via github action
**Issue summary**

To speed up boot times, we use hashes of tipset keys to shortcut the scan of the snapshot files loaded when booting forest daemon. These checkpoints are hardcoded in the codebase and updating them requires adding hard coding the new checkpoint in the code. To ease with updating the checkpoints, it will be good to put these checkpoints in a config file (yaml?) and automate the updates via gh action.

AC: 

- [ ] Update the `Filecoin.ChainGetTipSetHash` RPC command to fetch hash by height.
- [ ] Move hardcoded checkpoint hashes to a config file that is read when forest boots up.
- [ ] Automate updating of the checkpoints via a cronjob in github actions at a monthly or an good enough interval.

Additionally 

https://github.com/ChainSafe/forest/blob/5c03f45dc69e3bf58548d9fd4b956a28752bb190/blockchain/chain/src/store/index.rs#L70

lemmih:
## Summary of changes

- Modify `forest-cli chain tipset-hash` to take an epoch argument rather than a set of CIDs.
- Move the known mainnet/calibnet checkpoints to a separate yaml file which is loaded at compile-time.
- Schedule a new issue every month with instructions for how to keep the checkpoints yaml file up-to-date.

Closes #2735 

## Change checklist


blockchain/chain/src/store/index.rs
```diff
@@ -26,97 +26,63 @@ const SKIP_LENGTH: ChainEpoch = 20;
...
+        if KNOWN_CHECKPOINTS.calibnet.contains(&key) {
+            return Some(TipsetKeys::new(vec![Cid::from_str(
+                forest_networks::calibnet::GENESIS_CID,
+            )
+            .unwrap()]));
+        }
+        None
     }
 
     pub fn get_tipset_hashes(network: &str) -> Option<HashSet<String>> {
```
hanabi1224, line 70: Is this function still being used? If so how abt using `Option<&HashSet<String>>` as return type to avoid the `clone` operation

hanabi1224 (approved):

blockchain/chain/src/store/index.rs
lemmih: Still used. And the HashSet will be modified so it'll have to be cloned at one point or another. The data is tiny, though.

.github/workflows/checkpoints.yml
```diff
@@ -0,0 +1,17 @@
+name: Checkpoints
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 1 * *" # The 1st of every month
+
+jobs:
+  stuff:
```
LesnyRumcajs, line 9:
```suggestion
  checkpoint_issue:
```

scripts/calibnet_health_check.sh
```diff
@@ -81,6 +81,20 @@ echo "Print forest log files"
 ls -hl "$LOG_DIRECTORY"
 cat "$LOG_DIRECTORY"/*
 
+# Get the checkpoint hash at epoch 424000. This output isn't affected by the
```
LesnyRumcajs, line 84: rock solid

LesnyRumcajs (approved):

Issue #2741, opened 2023-03-31
Snapshots compression for import/export/download


LesnyRumcajs:
**Issue summary**

We could speed up things by downloading compressed snapshots. Mainnet compressed snapshots are significantly smaller (70G vs 130G), and it would make a huge difference for bandwidth-constrained users.

**Task summary**

We should handle both importing and exporting compressed snapshots. Compression could be on by default. This **must** be configurable via configuration and/or flags.

- [x] add `zstd` decompression functionality for importing snapshots. A potential crate to use https://crates.io/crates/zstd
- [x] add a configuration option to use compressed snapshots for downloading (both the `forest` and the `forest-cli snapshot fetch`) and importing,
- [x] add `zstd` compression functionality for exporting snapshots.
- [x] update the snapshot service to upload both compressed and uncompressed files.
- [x] #2873
- [x] #2874
- [x] add a configuration option to use compression when exporting, 

**Acceptance Criteria**

- [x] Limit disk space usage. That is to say; it should be possible to write directly to the `zstd` stream and not create a car file, compress it and then delete it.
- [x] Forest can export compressed snapshots,
- [x] Forest can import compressed snapshots,
- [x] Forest can download compressed snapshots,
- [x] `forest-cli` can download compressed snapshots (right now it's downloading uncompressed ones only)

**Other information and links**

elmattic:
I like it.

lemmih:
@hanabi1224 Is this done?

hanabi1224:
@lemmih updated the checkboxes, `modify forest and forest-cli to prefer compressed files` one is not done

lemmih:
Done.

PR #2748, opened 2023-04-04
feat: support importing zstd compressed snapshot file directly

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/2741

- Automatically decompress a snapshot file if it is zstd compressed
- Unittests
- CI sync check tests

Closes

## Change checklist


scripts/gen_coverage_report.sh
```diff
@@ -26,10 +26,13 @@ function cov {
...
 
+aria2c -x5 -d "$TMP_DIR" -o calibnet_snapshot.zst https://snapshots.calibrationnet.filops.net/minimal/latest.zst
 cargo llvm-cov --workspace clean
 cargo llvm-cov --workspace --no-report --features slow_tests
 cov forest-cli --chain calibnet db clean --force
 cov forest-cli --chain calibnet snapshot fetch --aria2 -s "$TMP_DIR"
+SNAPSHOT_PATH=$(find "$TMP_DIR" -name \*.zst | head -n 1)
+cov forest --chain calibnet --encrypt-keystore false --import-snapshot "$SNAPSHOT_PATH" --height=-200 --detach --track-peak-rss
 SNAPSHOT_PATH=$(find "$TMP_DIR" -name \*.car | head -n 1)
 cov forest --chain calibnet --encrypt-keystore false --import-snapshot "$SNAPSHOT_PATH" --height=-200 --detach --track-peak-rss
```
lemmih, line 37: This code tries to detach two instances of Forest. I don't think that will work (and the errors will be hidden in the output files). You probably just want to verify that Forest can load the compressed snapshot.

lemmih: You could even clean the DB between the two invocations.

lemmih (changes requested):
Looks good.

scripts/gen_coverage_report.sh
hanabi1224: Good catch, fix. Thanks!

lemmih (approved):

utils/forest_utils/src/net/download.rs
```diff
@@ -107,3 +104,87 @@ impl FetchProgress<BufReader<async_fs::File>> {
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        if let Some(left) = self.left_mut() {
+            Pin::new(left).poll_read(cx, buf)
+        } else if let Some(right) = self.right_mut() {
+            Pin::new(right).poll_read(cx, buf)
+        } else {
+            unimplemented!("This branch should never be hit")
```
LesnyRumcajs, line 187: I don't think we are planning to implement it :)
```suggestion
            panic!("This branch should never be hit")
```

LesnyRumcajs (approved):
I like it!

utils/forest_utils/src/net/download.rs
hanabi1224: Fixed

hanabi1224: Hmm, saw this on my phone and it did not show as suggested change : )

PR #2751, opened 2023-04-06
add exceptions to unused deps script

LesnyRumcajs:
## Summary of changes

- Added a special case to the unused dependencies script. @jdjaustin found a case where a crate is an implicit dependency of another crate (`quickcheck` is needed by `quickcheck_macros`) though it is not explicitly called in the code, thus making the check fail.

Closes

## Change checklist


hanabi1224:
FYI the [script here](https://github.com/ChainSafe/fil-actor-states/blob/main/scripts/find_unused_deps.rb) supports `--ignore` option. For this specific case, can we add `use quickcheck;` instead of making an exception?

LesnyRumcajs:
> FYI the [script here](https://github.com/ChainSafe/fil-actor-states/blob/main/scripts/find_unused_deps.rb) supports `--ignore` option. For this specific case, can we add `use quickcheck;` instead of making an exception?

Explained on Slack why it is unfeasible to add `use quickcheck;` and why this exception would be the least invasive on the Rust codebase.

hanabi1224 (approved):

jdjaustin (approved):

PR #2754, opened 2023-04-12
feat: add `--compressed` to forest-cli snapshot export

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/2741

- Support exporting zstd compressed snapshots
- Remove `tokio::io::AsyncWrite` compat layer from the entire snapshot export flow
- Do not calculate checksum when either `--compressed` or `--skip-checksum` is supplied
- Update CI scripts

Closes

## Change checklist


forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -30,6 +30,9 @@ use crate::cli::{cli_error_and_die, handle_rpc_err};
 pub(crate) const OUTPUT_PATH_DEFAULT_FORMAT: &str =
     "forest_snapshot_{chain}_{year}-{month}-{day}_height_{height}.car";
 
+pub(crate) const OUTPUT_PATH_DEFAULT_COMPRESSED_FORMAT: &str =
+    "forest_snapshot_{chain}_{year}-{month}-{day}_height_{height}.car.zst";
+
 #[derive(Debug, Subcommand)]
 pub enum SnapshotCommands {
     /// Export a snapshot of the chain to `<output_path>`
```
lemmih, line 38: Maybe change the documentation from `forest_snapshot_{chain}_{year}-{month}-{day}_height_{height}.car` to `forest_snapshot_{chain}_{year}-{month}-{day}_height_{height}.car(.zst)`

lemmih (approved):

forest/cli/src/cli/snapshot_cmd.rs
hanabi1224: Fixed

node/rpc/src/chain_api.rs
```diff
@@ -58,14 +56,20 @@ pub(crate) async fn chain_export<DB, B>(
...
     }): Params<ChainExportParams>,
 ) -> Result<ChainExportResult, JsonRpcError>
 where
     DB: Blockstore + Clone + Send + Sync + 'static,
     B: Beacon,
 {
+    // Skip checksum when in compression mode
+    if !skip_checksum && compressed {
+        skip_checksum = true;
+    }
```
LesnyRumcajs, line 71:
```suggestion
    skip_checksum |= compressed;
```

node/rpc/src/chain_api.rs
```diff
@@ -98,22 +102,32 @@ where
...
-        let file = File::from_std(tmp_file.reopen()?);
-        data.chain_store
-            .export(
-                &start_ts,
-                recent_roots,
-                AsyncWriterWithChecksum::<Sha256, _>::new(BufWriter::new(file)),
-            )
-            .await
+        let file = async_fs::File::from(tmp_file.reopen()?);
+        if compressed {
```
LesnyRumcajs, line 106: Can we do it without duplicating this snippet by supplying an abstract writer?

utils/forest_utils/src/io/writer_checksum.rs
```diff
@@ -3,24 +3,24 @@
...
 pin_project! {
-    /// Wrapper `AsyncWriter` implementation that calculates the checksum on the fly.
+    /// Wrapper `AsyncWriter` implementation that calculates the optional checksum on the fly.
     /// Both `Writer` and `Digest` parameters are generic so one can use freely the relevant
     /// structures, e.g. `BufWriter` and `Sha256`.
     pub struct AsyncWriterWithChecksum<D, W> {
         #[pin]
         inner: W,
-        hasher: D,
+        hasher: Option<D>,
```
LesnyRumcajs, line 16: I'm not sure about it. Having `AsyncWriterWithChecksum` without checksum makes it just an `AsyncWriter`. This introduces hasher checks on every write. Do you think we can do it in a less invasive way?

hanabi1224: I've simplified the checksum logic to only use `AsyncWriterWithOptionalChecksum` inside `export` fn, let me know if that's ok or did u suggest anything else

node/rpc/src/chain_api.rs
LesnyRumcajs: I meant something like this 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=980b6908a0c22f335c6120a07ebb1374

hanabi1224: I c, but having a `Box` or `Arc` wrapper conflicts with the other change that makes `AsyncWriteWithChecksum` usage internal to the `export` function.

hanabi1224: Could u elaborate on ur suggestion [here](https://github.com/ChainSafe/forest/pull/2754#discussion_r1164161729)?

utils/forest_utils/src/io/writer_checksum.rs
hanabi1224: OK if u don't like `Option<Hasher>`, maybe I can restore the logic and do `Either<AsyncWriter, AsyncWriterWithChecksum>`, sounds good?

LesnyRumcajs: I'm fine with both, provided the solution with `Option<Hasher>` doesn't get a performance hit (can be checked by exporting mainnet snapshot).

node/rpc/src/chain_api.rs
LesnyRumcajs: Alright, if it's a massive pain then I'm good with the existing solution.

utils/forest_utils/src/io/writer_checksum.rs
LesnyRumcajs: Does this mean an additional `if` condition makes it faster? We should put `if`s everywhere :rofl: 

Anyway, it's fine.

LesnyRumcajs (approved):

node/rpc/src/chain_api.rs
hanabi1224: I've reverted the other change and try to apply this one, but it's still hard to make it compile as `W` has 2 non-auto traits `AsyncWrite + Checksum<D>` 

Got `error[E0225]: only auto traits can be used as additional traits in a trait object` to compile the below code.  How about just keep the code as it was? 

```rust
let writer: Arc<TokioMutex<dyn AsyncWrite + Checksum<Sha256> + Send + Unpin>> =
            if compressed {
                Arc::new(TokioMutex::new(AsyncWriterWithChecksum::<Sha256, _>::new(
                    ZstdEncoder::new(BufWriter::new(file)),
                    false,
                )))
            } else {
                Arc::new(TokioMutex::new(AsyncWriterWithChecksum::<Sha256, _>::new(
                    BufWriter::new(file),
                    !skip_checksum,
                )))
            };
```

LesnyRumcajs:
@hanabi1224 mergio?

hanabi1224:
> @hanabi1224 mergio?

Not yet, let me add more tests with real files to make sure we don't have checksum missmatch

Issue #2768, opened 2023-04-14
Fix importing compressed snapshots from urls


lemmih:
**Issue summary**

Forest supports importing a snapshot from a URL like this:

    forest --import-snapshot https://forest-snapshots.fra1.digitaloceanspaces.com/calibnet/forest_snapshot_calibnet_2023-04-14_height_470270.car

Intuitively, it should also support compressed snapshots but that doesn't work. It can be seen by running

    forest --import-snapshot https://snapshots.calibrationnet.filops.net/minimal/latest.zst

The download logic should be updated to support zstd.

**Other information and links**

PR #2769, opened 2023-04-14
fix: support importing zstd compressed snapshot from url

hanabi1224:
## Summary of changes

- support importing zstd compressed snapshot from url
- update sync check CI

Closes https://github.com/ChainSafe/forest/issues/2768

## Change checklist


utils/forest_utils/src/miscs/either.rs
```diff
@@ -0,0 +1,46 @@
...
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        if let Some(left) = self.left_mut() {
+            Pin::new(left).poll_read(cx, buf)
+        } else if let Some(right) = self.right_mut() {
+            Pin::new(right).poll_read(cx, buf)
+        } else {
+            panic!("This branch should never be hit")
+        }
```
lemmih, line 44: Is something like this possible?
```suggestion
        match self.deref_mut() {
            Self::Left(left) => Pin::new(left).poll_read(cx, buf),
            Self::Right(right) => Pin::new(right).poll_read(cx, buf),
        }
```

utils/forest_utils/src/miscs/either.rs
```diff
@@ -0,0 +1,43 @@
...
+impl<L: AsyncRead + Unpin, R: AsyncRead + Unpin> AsyncRead for Either<L, R> {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        match self.deref_mut() {
+            Self::Left(left) => Pin::new(left).poll_read(cx, buf),
+            Self::Right(right) => Pin::new(right).poll_read(cx, buf),
+        }
```
lemmih, line 26: Is this code correct? Can you freely create new pins like this?

lemmih: Don't you need [map_unchecked_mut](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.map_unchecked_mut) or something? I'm not an expert on this.

hanabi1224: It does compile, so I guess it works. Will c if sync check CI job passes

lemmih: @LesnyRumcajs @elmattic What are your thoughts on this? I feel this documentation is relevant: https://doc.rust-lang.org/std/pin/index.html#projections-and-structural-pinning

lemmih: Oh, I see. You require both `L` and `R` to be `Unpin`.

hanabi1224: `map_unchecked_mut` definitely looks like a better fit but it's `unsafe`, is it allowed to use `unsafe` in `forest`?

lemmih: You probably want to use `Pin::into_inner` instead of `deref_mut`, then.

utils/forest_utils/src/miscs/either.rs
```diff
@@ -0,0 +1,28 @@
...
+    Right(R),
+}
+
+impl<L: AsyncRead + Unpin, R: AsyncRead + Unpin> AsyncRead for Either<L, R> {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        match self.deref_mut() {
```
lemmih, line 23:
```suggestion
        match Pin::into_inner(self) {
```

utils/forest_utils/src/miscs/either.rs
lemmih: You can use unsafe functions as long as you prove the correctness yourself. :)

utils/forest_utils/src/miscs/either.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::{
+    ops::DerefMut,
```
lemmih:
```suggestion
```

utils/forest_utils/src/miscs/either.rs
```diff
@@ -0,0 +1,28 @@
...
+use futures::AsyncRead;
+
+pub enum Either<L, R> {
+    Left(L),
+    Right(R),
+}
+
+impl<L: AsyncRead + Unpin, R: AsyncRead + Unpin> AsyncRead for Either<L, R> {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
```
lemmih:
```suggestion
        self: Pin<&mut Self>,
```

lemmih (approved):
Looks good.

LesnyRumcajs (approved):

PR #2782, opened 2023-04-20
fix: BufWriter flush issue in AsyncWriterWithCheckSum 

hanabi1224:
## Summary of changes

- Add unit test for `AsyncWriterWithCheckSum` that writes to files
- Flush BufWriter before finalizing the hasher
- Migrate from `tokio::io::AsyncWrite` to `futures::AsyncWrite`
- ~~Remove `tokio` compat layer from snapshot export flow~~ `async_fs::File` does not work with `TempFile::persist` properly(it's likely the file is not fully flushed before being moved), keep getting checksum mismatch error, fixed by switching back to `tokio::fs::File` 

Closes

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3667,6 +3667,7 @@ name = "forest-cli"
 version = "0.8.1"
 dependencies = [
  "ahash 0.8.3",
+ "anes",
  "anyhow",
  "assert_cmd",
  "atty",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ resolver = "2"
 
 [workspace.dependencies]
 ahash = "0.8"
+anes = "0.1.6"
 anyhow = "1.0"
 assert_cmd = "2"
 async-compression = { version = "0.3", features = ["futures-io", "zstd"] }
--- a/forest/cli/Cargo.toml
+++ b/forest/cli/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 
 [dependencies]
 ahash.workspace = true
+anes.workspace = true
 anyhow.workspace = true
 atty = "0.2"
 base64.workspace = true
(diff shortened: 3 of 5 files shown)


utils/forest_utils/src/io/writer_checksum.rs
```diff
@@ -106,11 +106,40 @@ impl<D: Digest> Checksum<D> for VoidAsyncWriterWithNoChecksum<D> {
...
     use tokio::io::{AsyncWriteExt, BufWriter};
 
     use super::*;
 
+    #[tokio::test]
+    async fn file_writer_tokio_fs_buf_writer() -> anyhow::Result<()> {
+        let temp_file_path = tempfile::Builder::new().tempfile()?;
+        let temp_file = tokio::fs::File::create(temp_file_path.path()).await?;
+        let mut temp_file_writer =
+            AsyncWriterWithChecksum::<Sha256, _>::new(tokio::io::BufWriter::new(temp_file));
```
hanabi1224, line 121: @LesnyRumcajs So this is a minimal repro of the checksum missmatch problem. The UT passes w/o `BufWriter` but fails with it.

LesnyRumcajs (approved):
rock-solid

node/rpc/src/chain_api.rs
```diff
@@ -102,21 +100,17 @@ where
...
             )
             .await
     } {
         Ok(checksum) if !dry_run => {
-            if let Err(e) = tmp_file.persist(&output_path) {
-                // Temporary files cannot be persisted across filesystems, so we fallback to a
-                // copy in case it happens.
-                tokio::fs::copy(e.file, &output_path).await?;
-            }
+            temp_path.persist(&output_path)?;
```
lemmih, line 113: Maybe we should write a note here saying that this rename is expected to succeed because we make sure the temp-file is in the same folder as the final file.

lemmih (approved):

node/rpc/src/chain_api.rs
hanabi1224: Fixed

Issue #2786, opened 2023-04-20
compressed snapshot fetch without aria2 and filecoin provider fails


LesnyRumcajs:
**Describe the bug**

Without using `aria2c` the snapshot fetch of a compressed snapshot fetch fails.

**To Reproduce** Steps to reproduce the behaviour:

Run `cargo run --release --bin forest-cli -- --chain calibnet snapshot fetch --provider filecoin --compressed`

**Log output**
```
❯ cargo run --release --bin forest-cli -- --chain calibnet snapshot fetch --provider filecoin --compressed
    Finished release [optimized] target(s) in 0.93s
     Running `target/release/forest-cli --chain calibnet snapshot fetch --provider filecoin --compressed`
2023-04-20T22:32:03.064327Z  INFO forest_cli_shared::cli::snapshot_fetch: Snapshot url: https://snapshots.calibrationnet.filops.net/minimal/latest.zst
2023-04-20T22:32:04.144173Z  INFO forest_cli_shared::cli::snapshot_fetch: Snapshot url: https://pub-f851d666b9a54f60bd93202926618941.r2.dev/minimal/489840_2023_04_20T20_13_00Z.car.zst
2023-04-20T22:32:04.144216Z  INFO forest_cli_shared::cli::snapshot_fetch: Snapshot will be downloaded to /home/rumcajs/.local/share/forest/snapshots/calibnet/filecoin_snapshot_calibnet_2023-04-20_height_489840.car.zs
t (989.44 MiB)
Downloading snapshot 989.44 MB / 989.44 MB [======================================================================================================================================================] 100.00 % 44.48 MB/s
Finished downloading the snapshot.2023-04-20T22:32:26.390137Z  INFO forest_cli_shared::cli::snapshot_fetch: Validating checksum...
2023-04-20T22:32:26.748762Z  WARN forest_cli::cli::snapshot_cmd: Unable to get the checksum file. Snapshot downloaded but not verified.
2023-04-20T22:32:26.748831Z  INFO forest_cli::cli::snapshot_cmd: Retry attempt 1 started with delay 60s.
2023-04-20T22:33:26.750346Z  INFO forest_cli_shared::cli::snapshot_fetch: Snapshot url: https://snapshots.calibrationnet.filops.net/minimal/latest.zst
[...]

PR #2814, opened 2023-05-02
fix: snapshot fetch without aria2

hanabi1224:
## Summary of changes

-

Closes https://github.com/ChainSafe/forest/issues/2786

## Change checklist

### DIFF
--- a/forest/shared/src/cli/snapshot_fetch.rs
+++ b/forest/shared/src/cli/snapshot_fetch.rs
@@ -254,12 +254,13 @@ async fn snapshot_fetch_forest(
         .await?
     } else {
         let total_size = last_modified.size;
-        download_snapshot_and_validate_checksum(
+        download_snapshot_and_validate_checksum_if_needed(
             client,
             url,
             &snapshot_path,
             snapshot_response,
             total_size,
+            use_compressed,
         )
         .await?;
     }
@@ -333,12 +334,13 @@ async fn snapshot_fetch_filecoin(
             .and_then(|ct_len| ct_len.parse::<u64>().ok())
             .ok_or_else(|| anyhow::anyhow!("Couldn't retrieve content length"))?;
 
-        download_snapshot_and_validate_checksum(
+        download_snapshot_and_validate_checksum_if_needed(
             client,
             snapshot_url,
             &snapshot_path,
             snapshot_response,
             total_size,
+            use_compressed,
         )
         .await?;
     }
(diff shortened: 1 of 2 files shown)


lemmih (approved):

scripts/calibnet_health_check.sh
```diff
@@ -30,12 +30,22 @@ echo "$1" > preloaded_wallet.key
...
 echo "Fetching params"
 $FOREST_CLI_PATH fetch-params --keys
+echo "Downloading zstd compressed snapshot without aria2"
+$FOREST_CLI_PATH --chain calibnet snapshot fetch --provider filecoin --compressed -s "$SNAPSHOT_DIRECTORY"
+echo "Downloading snapshot without aria2"
+$FOREST_CLI_PATH --chain calibnet snapshot fetch --provider filecoin -s "$SNAPSHOT_DIRECTORY"
+echo "Cleaning up snapshots"
+$FOREST_CLI_PATH --chain calibnet snapshot clean -s "$SNAPSHOT_DIRECTORY" --force
+echo "Cleaning up snapshots again"
+$FOREST_CLI_PATH --chain calibnet snapshot clean -s "$SNAPSHOT_DIRECTORY" --force
```
creativcoder, line 40: Why do we need to repeat `snapshot clean` again?

forest/shared/src/cli/snapshot_fetch.rs
```diff
@@ -375,13 +378,20 @@ where
     let mut downloaded: u64 = 0;
     let mut stream = snapshot_response.into_body();
 
-    let mut snapshot_hasher = Sha256::new();
+    let mut snapshot_hasher = if use_compressed {
+        None
```
elmattic, line 382: What is the reason behind not checking against checksum in case of compressed snapshot and without aria2? Don't we have a digest in both case? (with and without aria2)

hanabi1224: Checksum is not provided for a compressed snapshot. Also zstd has checksum mechanism internally

elmattic: Hmm if I look into the digest file I have:
```
cat 468720_2023_04_13T12_13_00Z.sha256sum
ae6d21b3bd88fa0d21528343b05e6f323a7131fad5d3e1148a5ed7c603d915a5 *468720_2023_04_13T12_13_00Z.car
180819491d0dfd37cc09ebc5ceded668c09fa6ec3f1fefe54ad20ca0fa71805a *468720_2023_04_13T12_13_00Z.car.zst
```

But if zstd does it internally, then we can skip it. 👍

elmattic (approved):

forest/shared/src/cli/snapshot_fetch.rs
hanabi1224: Did not realize checksum is there. I can investigate adding back checksum check then. Thx for the info

elmattic: 👍

hanabi1224: Just checked the [zstd spec](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#zstandard-frames) again, its checksum mechanism is optional, we should indeed verify checksum if it's available

scripts/calibnet_health_check.sh
LesnyRumcajs: It's a good question. It might be worth adding this issue in the code to make it clear from the start.

PR #2838, opened 2023-05-08
Optimize the CI

LesnyRumcajs:
## Summary of changes

- don't trigger Rust workflows on Markdown or Ruby changes,
- don't trigger Docker workflow and Rust test on draft PRs,
- don't trigger build for `amd64` (docker) in a PR (this is basically done in the Rust workflow so it's fine if we check only `arm64`
- use less buildjet to cut costs,
- upload binary artifacts to re-use them in calibnet check.

Closes

## Change checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,8 +137,7 @@ test-vectors: pull-serialization-tests run-vectors
 
 # Test all without the submodule test vectors with release configuration
 test:
-	cargo nextest run --all --exclude serialization_tests --exclude forest_message
-	cargo nextest run -p forest_message --features blst --no-default-features
+	cargo nextest run --all --exclude serialization_tests
 	cargo nextest run -p forest_db --no-default-features --features paritydb
 	cargo nextest run -p forest_db --no-default-features --features rocksdb
 	cargo nextest run -p forest_libp2p_bitswap --all-features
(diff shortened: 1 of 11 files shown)


lemmih:
Summary of notes: Moving away from buildjet increases the CI times but I think we can make up most of the difference just by optimizing our build scripts. Should be possible to save between 2 to 18 minutes in each CI job. I say we merge this and work on further optimizations in follow-up PRs.


Notes for the `All lint check`:

- Linting the RocksDB feature takes 6 minutes (4m14s + 1m44s). Maybe we should drop RocksDB support altogether.
- I don't think running `apt-get update` is necessary or beneficial. It takes 10s.
- Installing `aria2` is probably unnecessary for linting.
- Installing `clang` and `cmake` might be unnecessary if we get rid of RocksDB. Do we still need `cmake` for the protocol buffers?
- Running `cargo clean` is unnecessary and takes 26s.
- The `forest_deleg_cns` feature would be better as a runtime flag. This would make Forest easier to use in testing and it would reduce the number of feature combinations we need to lint.

Notes for `Test`:

- Unnecessary check of slow_tests: `cargo check --tests --features slow_tests`. Takes 1m 23s.
- Slow tests take up 2m 22s. Many slow tests are waiting for network input and are not run in parallel.
- 342 calls to `sccache` are not cacheable because incremental compilation is turned on. We should always turn off incremental compilation in CI.
- RocksDB testing takes 43s.

Notes for `Build Ubuntu`:

- Linking takes 2m 12s (a little over 20% of the total time). Maybe `mold` would be faster.
- It looks like all our actor crates cannot be cached because they use a build script. :(

Notes for `Calibnet health check`:

- Downloading snapshots from filops can take a long time: More than 18 minutes is spent just downloading snapshots. We should move away from filops and only depend on our own snapshots. Furthermore, download tests can be done in the background (or in a separate job).

lemmih (approved):

creativcoder (approved):

PR #2844, opened 2023-05-09
feat(ci): cache target folder on critical path and on buildjet

hanabi1224:
## Summary of changes

-

Closes

## Change checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,11 +34,15 @@ jobs:
     runs-on: buildjet-8vcpu-ubuntu-2204
     # Run the job only if the PR is not a draft.
     # This is done to limit the runner cost.
-    if: github.event.pull_request.draft == false
+    # if: github.event.pull_request.draft == false
     timeout-minutes: 30
     steps:
       - name: Checkout Sources
         uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: target
+          key: cache-test-${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
       - name: Setup sccache
         uses: mozilla-actions/sccache-action@v0.0.3
         timeout-minutes: ${{ fromJSON(env.CACHE_TIMEOUT_MINUTES) }}
@@ -108,6 +112,10 @@ jobs:
     steps:
       - name: Checkout Sources
         uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: target
+          key: cache-build-${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
       - name: Setup sccache
         uses: mozilla-actions/sccache-action@v0.0.3
         timeout-minutes: ${{ fromJSON(env.CACHE_TIMEOUT_MINUTES) }}


lemmih (changes requested):
I don't understand this PR. Could you explain to me what you're aiming to achieve?

lemmih (commented):
I'm not convinced this is the right approach. Could you convince @LesnyRumcajs first and then I'll re-review.

LesnyRumcajs:
It's actually what we are doing in the Dockerfile to reduce subsequent build times.
https://github.com/ChainSafe/forest/blob/main/Dockerfile#L50

That being said, the `target` is quite big - 4G on a clean install. According to [docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) the limit is 10G on the entire repository, so there'd be a bunch of evictions happening all the time when working with more than 2 branches at once. Or does the limit not apply here?

LesnyRumcajs:
Plus, the current approach is not that bad.

image

lemmih:
> It's actually what we are doing in the Dockerfile to reduce subsequent build times. https://github.com/ChainSafe/forest/blob/main/Dockerfile#L50
> 
> That being said, the `target` is quite big - 4G on a clean install. According to [docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) the limit is 10G on the entire repository, so there'd be a bunch of evictions happening all the time when working with more than 2 branches at once. Or does the limit not apply here?

We disabled this kind of caching exactly because we ran into the 10GiB limit. Hailong has an idea for getting around that limit. I'm not sold on the idea but I'm willing to consider it if he can convince you. I'll let Hailong explain the details.

hanabi1224:
@LesnyRumcajs @lemmih here's detailed explaination

- The compressed size of target folder for `build-ubuntu` is ~1GB  see [log](https://github.com/ChainSafe/forest/actions/runs/4925456297/jobs/8799711019#step:13:3)
```
Post job cleanup.
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/forest/forest --files-from manifest.txt --use-compress-program zstdmt
Cache Size: ~1145 MB (1[2](https://github.com/ChainSafe/forest/actions/runs/4925456297/jobs/8799711019#step:13:2)00900405 B)
Cache saved successfully
Cache saved with key: cache-build-Linux-fad17427c9ba2901e2a876b2259[3](https://github.com/ChainSafe/forest/actions/runs/4925456297/jobs/8799711019#step:13:3)992a9[4](https://github.com/ChainSafe/forest/actions/runs/4925456297/jobs/8799711019#step:13:4)2977d86c09fc14[5](https://github.com/ChainSafe/forest/actions/runs/4925456297/jobs/8799711019#step:13:5)65994e8ba7f768b
```
- Only branches that change `Cargo.lock` could upload the cache (GH action cache are immutable, and Cargo.lock hash is used as cache key)
- GH action cache prune cache nightly, we have seen cache usage at ~130GB when we were using `rust-cache` action, that being said, during active development of a PR, cache can be pruned at most once a day, non-default branch first, then from oldest to latest
- GH action cache should have better IO performance than other cache backend for `build-ubuntu` job that runs on GH hosted runners
- As I estimate, the latest target cache for `main` could survive the nightly prune, and that should be sufficient for the purpose,  If we give it a try and observe otherwise, we can always switch to other cache backends like s3

LesnyRumcajs:
@hanabi1224 Thanks for the write-up. Is the cache shared across different branches? That is to say, does `main` having the same `Cargo.lock` means that my branch will re-use its cache?

hanabi1224:
@LesnyRumcajs The behavior is described in this [doc](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)

> The cache action first searches for cache hits for key and the cache version in the branch containing the workflow run. If there is no hit, it searches for restore-keys and the version. If there are still no hits in the current branch, the cache action retries same steps on the default branch. Please note that the scope restrictions apply during the search. For more information, see "[Restrictions for accessing a cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)."

LesnyRumcajs:
> @LesnyRumcajs The behavior is described in this [doc](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)
> 
> > The cache action first searches for cache hits for key and the cache version in the branch containing the workflow run. If there is no hit, it searches for restore-keys and the version. If there are still no hits in the current branch, the cache action retries same steps on the default branch. Please note that the scope restrictions apply during the search. For more information, see "[Restrictions for accessing a cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)."

But given
> Only branches that change Cargo.lock could upload the cache (GH action cache are immutable, and Cargo.lock hash is used as cache key)

it means that if the features don't change the `Cargo.lock` but they would change a file on which most of the code relies (or add a different compiler option / use different toolchain) they would need to be completely re-compiled every time?

hanabi1224:
> it means that if the features don't change the `Cargo.lock` but they would change a file on which most of the code relies (or add a different compiler option / use different toolchain) they would need to be completely re-compiled every time?

@LesnyRumcajs We could call rest API to delete cache on main before save so that it's always up-to-date, and in this case, we don't need to include `Cargo.lock` hash in cache key 

https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-github-actions-caches-for-a-repository-using-a-cache-key

LesnyRumcajs:
Each branch should keep its own cache, which shouldn't be reusable across branches (even `main`). Generally, each branch should be new and pristine, otherwise, some weird compilation issues may arise. Do you agree?

LesnyRumcajs:
But why do we have to re-use the default branch? Could we simply put `github.ref` as the key?

hanabi1224:
> But why do we have to re-use the default branch's cache? Could we simply put `github.ref` as the key?

That way the first batch of CI jobs for that PR could benefit from the cache, then we could trade off whether or not save the new cache after each run. Otherwise, the first batch of CI jobs always starts from scratch and we have to always save the cache which consumes more space.  In practise I never see a GH action cache setup that uses `github.ref` as the key

LesnyRumcajs:
> > But why do we have to re-use the default branch's cache? Could we simply put `github.ref` as the key?
> 
> That way the first batch of CI jobs for that PR could benefit from the cache, then we could trade off whether or not save the new cache after each run. Otherwise, the first batch of CI jobs always starts from scratch and we have to always save the cache which consumes more space. In practise I never see a GH action cache setup that uses `github.ref` as the key

We can try on this particular job (build ubuntu) and see how it goes. It's a minor thing to roll back in case issues arise. If it behaves well we can use it in other jobs. What do you think?

Issue #2843, opened 2023-05-09
Make fetching proof parameters optional


lemmih:
**Issue summary**

The `importing_bad_snapshot_should_fail` test (linked below) takes a long time to run because it downloads proof parameters from ipfs. However, the test has nothing to do with proof parameters. We should have a way to suppress the downloading of new proof parameters.

**Other information and links**

https://github.com/ChainSafe/forest/blob/19666d7de0c5166cbd666a3257fa1e12d2725fec/forest/daemon/tests/import_snapshot_tests.rs#L12

PR #2849, opened 2023-05-09
fix: fetch param files on demand

hanabi1224:
## Summary of changes

-

Closes https://github.com/ChainSafe/forest/issues/2843

## Change checklist


utils/paramfetch/src/lib.rs
```diff
@@ -72,6 +75,40 @@ pub fn set_proofs_parameter_cache_dir_env(data_dir: &Path) {
...
+        rt.block_on(ensure_params_downloaded_async())
+    } else {
+        Ok(())
+    }
+}
+
+/// Ensures the parameter files are downloaded to cache dir (async version)
+pub async fn ensure_params_downloaded_async() -> anyhow::Result<()> {
+    lazy_static::lazy_static! {
+        static ref LOCK: tokio::sync::Mutex<()> = tokio::sync::Mutex::new(());
```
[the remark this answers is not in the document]
lemmih: If yes, add the reasons as comments in the code.

lemmih (changes requested):
Could you describe your changes in the PR description? This is especially important since your solution differs from the solution proposed in the issue.

utils/paramfetch/src/lib.rs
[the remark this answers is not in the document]
hanabi1224: This function is not longer needed

hanabi1224:
> Could you describe your changes in the PR description? This is especially important since your solution differs from the solution proposed in the issue.

Updated desc and moved the check to right after snapshot import and before block validation as discussed

hanabi1224:
> This change does not suppress downloading param files, right? It only pushes it for later to speed-up the tests?

Yes

LesnyRumcajs:
> > This change does not suppress downloading param files, right? It only pushes it for later to speed-up the tests?
> 
> Yes

I am not 100% convinced that this is the way. I mean, we accommodated the tests, but what is likely more important is to accommodate the overall Forest experience. Did you consider downloading the param files while downloading/importing the snapshot? For a regular Forest user, it would be a startup time improvement.

hanabi1224:
> Let's implement that directly.

Done.  @lemmih  @LesnyRumcajs

utils/genesis/src/lib.rs
```diff
@@ -114,17 +115,42 @@ where
...
+            reader,
+            skip_load,
+        ))
     };
 
+    // Not checking `validate_height` to decide whether to download parameter files
+    // here because it defaults to `None` in config, making parallelization
+    // disabled by default
+    let ensure_params_downloaded_task_handle =
+        tokio::spawn(forest_utils::proofs_api::paramfetch::ensure_params_downloaded());
```
LesnyRumcajs, line 141: I am not convinced that `import_chain` is a good place for spawning the params download task. I mean, as a developer, I wouldn't expect such a method to download lots of relatively large files.

Could we somehow refactor this to have those separated? I don't have anything precise in mind. Perhaps splitting the method into smaller ones and a bit of renaming will do.

lemmih (changes requested):
As discussed in Slack, `import_chain` can be refactored to simplify the code significantly.

utils/genesis/src/lib.rs
hanabi1224: Done

forest/daemon/src/daemon.rs
```diff
@@ -538,37 +538,54 @@ async fn prompt_snapshot_or_die(
 async fn sync_from_snapshot<DB>(
     config: &Config,
     state_manager: &Arc<StateManager<DB>>,
+    halt_after_import: bool,
```
LesnyRumcajs, line 541: This is a cosmetic thing, but this makes it hard to understand for the fellow devs. If you are passing `halt_after_import` boolean, I'd expect this method would do the halting itself, which isn't the case (and rightly so).

`sync_from_snapshot(&config, &state_manager, opts.halt_after_import).`

A bit more expressive form would be

`sync_from_snapshot(&config, &state_manager, ProofParametersFetch::EnsureDownloaded).`

what do you think?

hanabi1224: Updated

forest/daemon/src/daemon.rs
```diff
@@ -538,37 +541,48 @@ async fn prompt_snapshot_or_die(
...
             config.client.snapshot_height
         } else {
             Some(0)
         };
 
-        match import_chain::<_>(
+        let ensure_params_downloaded_task_handle = if proof_params_required {
+            Some(tokio::spawn(async {
+                // Add a small delay to avoid import chain progress bar being interrupted.
+                tokio::time::sleep(Duration::from_secs(1)).await;
```
LesnyRumcajs, line 561: This is a bit brittle approach, but let's investigate how to sync this properly in another issue, shall we?

LesnyRumcajs (approved):

forest/daemon/src/daemon.rs
```diff
@@ -538,37 +541,48 @@ async fn prompt_snapshot_or_die(
...
+        .context("Failed miserably while importing chain from snapshot")?;
+        info!("Imported snapshot in: {}s", stopwatch.elapsed().as_secs());
+
+        if let Some(ensure_params_downloaded_task_handle) = ensure_params_downloaded_task_handle {
+            ensure_params_downloaded_task_handle.await??;
+        }
+
+        if let Some(validate_height) = validate_height {
+            validate_chain(state_manager, validate_height).await?;
         }
```
lemmih, line 583: The `sync_from_snapshot` shouldn't re-validate the chain. You can hoist it out.

lemmih (changes requested):

forest/daemon/src/daemon.rs
hanabi1224: As I understand, the whole point of this function is to combine the process of snapshot import and validation and make it cancellable. Would it be better to rename the function to `sync_from_snapshot_and_validate`?

```rust
tokio::select! {
        ret = sync_from_snapshot(&config, &state_manager, proof_params_required).fuse() => {
            if let Err(err) = ret {
                services.shutdown().await;
                return Err(err);
            }
        },
        _ = tokio::signal::ctrl_c() => {
            services.shutdown().await;
            return Ok(db);
        },
        _ = terminate.recv() => {
            services.shutdown().await;
            return Ok(db);
        },
        _ = shutdown_recv.recv() => {
            services.shutdown().await;
            return Ok(db);
        },
    }
```

lemmih: Hmm, the ctrl-c handling is bothering me to no end. It happens so often that ctrl-c is ignored. How about we move the ctrl-c handling out into `main.rs`? That way, the daemon doesn't have to worry about it.

hanabi1224: @lemmih When I search in the code, I can find 3 places that handle ctrl-c signal in forest daemon, I assume it's duplicated here for a reason.  @elmattic could u share ur insights on whether it is possible to remove the ctrl-c handling from here?

lemmih: There's no good reason to handle interrupts in three places. I'll have a go at refactoring it in a separate PR.

hanabi1224: > There's no good reason to handle interrupts in three places. I'll have a go at refactoring it in a separate PR.

@lemmih That makes sense, created an issue https://github.com/ChainSafe/forest/issues/2947 and mark this PR as blocked

hanabi1224:
Blocked by https://github.com/ChainSafe/forest/issues/2947

forest/daemon/src/daemon.rs
lemmih: The interrupt code has been merged. Now we can continue to refactor the proof parameter handling.

hanabi1224: Done

forest/daemon/src/daemon.rs
```diff
@@ -396,20 +398,43 @@ pub(super) async fn start(
...
+    let proof_params_required = !opts.halt_after_import || validate_height.is_some();
+    let ensure_params_downloaded_task_handle = if proof_params_required {
+        Some(tokio::spawn(async {
+            // Add a small delay to avoid import chain progress bar being interrupted.
+            tokio::time::sleep(Duration::from_millis(100)).await;
+            forest_utils::proofs_api::paramfetch::ensure_params_downloaded().await
+        }))
+    } else {
+        None
+    };
```
lemmih, line 426: I don't think the performance benefits are worth the extra complexity. Let's just ensure that the parameters have been downloaded right before `validate_chain` and right before `p2p_service` is spawned.

lemmih (changes requested):

forest/daemon/src/daemon.rs
```diff
@@ -396,20 +398,43 @@ pub(super) async fn start(
...
+            &config.client.data_dir,
+        );
     }
 
     let config = maybe_fetch_snapshot(should_fetch_snapshot, config).await?;
 
+    let validate_height = if config.client.snapshot {
+        config.client.snapshot_height
+    } else {
+        Some(0)
```
lemmih, line 414: Oh, and let's set this to `None`.

forest/daemon/src/daemon.rs
hanabi1224: Fixed.

forest/daemon/src/daemon.rs
hanabi1224: Fixed.

forest/daemon/src/daemon.rs
```diff
@@ -539,31 +553,17 @@ where
 {
     if let Some(path) = &config.client.snapshot_path {
```
lemmih, line 554: Let's inline `sync_from_snapshot`.

utils/forest_utils/src/proofs_api/paramfetch.rs
```diff
@@ -72,6 +76,28 @@ pub fn set_proofs_parameter_cache_dir_env(data_dir: &Path) {
     std::env::set_var(DIR_ENV, param_dir(data_dir));
 }
 
+/// Ensures the parameter files are downloaded to cache dir
+pub async fn ensure_params_downloaded() -> anyhow::Result<()> {
```
lemmih, line 80: Let's remove `CHECKED` and `LOCK`. We can add them back if/when we decide on downloading proof-parameters in the background.

lemmih (approved):

forest/daemon/src/daemon.rs
hanabi1224: Done.

forest/daemon/src/daemon.rs
```diff
@@ -388,19 +389,41 @@ pub(super) async fn start(
...
+
+    let validate_height = if config.client.snapshot {
+        config.client.snapshot_height
+    } else {
+        None
+    };
+    let proof_params_required = !opts.halt_after_import || validate_height.is_some();
+    if proof_params_required {
+        forest_utils::proofs_api::paramfetch::ensure_params_downloaded().await?;
+    }
```
lemmih, line 422: Instead of this logic, you can just call `ensure_params_downloaded` twice. Then the logic is encoded in the control flow.

forest/daemon/src/daemon.rs
```diff
@@ -388,19 +389,41 @@ pub(super) async fn start(
...
+
+    if let Some(validate_height) = validate_height {
+        validate_chain(&state_manager, validate_height).await?;
+    }
```
lemmih, line 426: We can inline `validate_height` and just have two `if` statements here.

hanabi1224: Done.

lemmih (approved):
Before:
```
        PASS [  20.027s] forest-daemon::import_snapshot_tests importing_bad_snapshot_should_fail
```
After:
```
        PASS [   3.336s] forest-daemon::import_snapshot_tests importing_bad_snapshot_should_fail
```

PR #2863, opened 2023-05-12
fix markdown and ruby ignore for Rust workflows

LesnyRumcajs:
## Summary of changes

- the ignores were not recursive,
- Ruby scripts linting the Rust code shall not be ignored in the Rust lint workflow,

Closes

## Change checklist


hanabi1224 (approved):
If a markdown file is included in rust as doc, there is chance the markdown change could break rust doc test right?

LesnyRumcajs:
> If a markdown file is included in rust as doc, there is chance the markdown change could break rust doc test right?

Hm, I wasn't aware of this feature. That being said, it's present only in: https://github.com/ChainSafe/forest/blob/fix-markdown-ruby-ignore-actions/node/forest_libp2p/bitswap/src/lib.rs

It's a valid point, though. Perhaps we should only ignore markdown files in the project's root directory and under `documentation`? What do you think?

lemmih (approved):

Issue #2801, opened 2023-04-27
Implement NV17 state migration


lemmih:

lemmih:
Here's a snapshot for testing the migration: https://forest-snapshots.fra1.digitaloceanspaces.com/debug/filecoin_calibnet_height_16000.car.zst

The migration happens at epoch 16_800.

PR #2866, opened 2023-05-15
fix: refactor state_migration to expose single pub api

hanabi1224:
## Summary of changes

- Expose single pub API from `state_migration` crate so that changes are (mostly, except `BunbleInfo`) only required to be made inside the crate for implementing new migrations

Closes

## Change checklist


vm/state_migration/src/lib.rs
```diff
@@ -1,9 +1,51 @@
...
+) -> anyhow::Result<Option<Cid>>
+where
+    DB: 'static + Blockstore + Clone + Send + Sync,
+{
+    let mappings: [(_, RunMigration<DB>); 2] = [
+        (Height::Hygge, nv18::run_migration::<DB>),
+        (Height::Lightning, nv19::run_migration::<DB>),
+    ];
+
+    for (height, migrate) in mappings {
```
elmattic, line 33: Before we had a `match` and now we have a `for` loop. Could this be avoided (i.e. using const code)?

hanabi1224: I don't think so because `DB` is a generic type, it would require some fancy compile-time reflection to achieve the goal, or any suggestions?

hanabi1224:
> Unrelated but could you rename the `miscs` module to `misc`?

@lemmih Done

lemmih (approved):

vm/state_migration/src/lib.rs
```diff
@@ -1,9 +1,72 @@
...
+    static BUNDLE_CHECKED: AtomicBool = AtomicBool::new(false);
+    if !BUNDLE_CHECKED.load(atomic::Ordering::Relaxed) {
+        BUNDLE_CHECKED.store(true, atomic::Ordering::Relaxed);
+        for info in &chain_config.height_infos {
+            for (height, _) in &mappings {
+                if height == &info.height {
+                    anyhow::ensure!(
+                        info.bundle.is_some(),
+                        "Actor bundle info for height {} needs to be defined in `networks/src/lib.rs` to run state migration",
+                        height
```
LesnyRumcajs, line 46:
```suggestion
                        "Actor bundle info for height {height} needs to be defined in `networks/src/lib.rs` to run state migration"
```

LesnyRumcajs (commented):
Nice. Let's update the state migration guide to reflect these changes. Alternatively, we can do it after you finish the NV17 migration. What do you think?

hanabi1224:
> Nice. Let's update the state migration guide to reflect these changes. Alternatively, we can do it after you finish the NV17 migration. What do you think?

I would prefer updating it altogether

Issue #2826, opened 2023-05-04
Devs can format their code without installing formatters getting in the way


aatifsyed:
> Oh, and let's reset the `yarn.lock` file.

_Originally posted by @lemmih in https://github.com/ChainSafe/forest/pull/2821#pullrequestreview-1412302970_

Some users can't run `yarn install --immutable && yarn run prettier **/*.md --write` without dirtying their lockfile.

Find a solution

aatifsyed:
References:
https://github.com/Jille/dockpin

aatifsyed:
Should also address rate limiting/other ci nonsense like
```
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.npmjs.org/@zkochan/js-yaml/-/js-yaml-0.0.[6](https://github.com/ChainSafe/forest/actions/runs/4979290978/jobs/8910701002?pr=2846#step:3:7).tgz: Request failed \"404 Not Found\"".
info If you think this is a bug, please open a bug report with the information provided in "/home/runner/work/forest/forest/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
https://github.com/ChainSafe/forest/actions/runs/4979290978/jobs/8910701002?pr=2846

LesnyRumcajs:
The simplest solution would be to create a Docker image with all the tools installed and host it on GHCR.

aatifsyed:
If we were to do this, I'd be worried about a couple of things:
- multiarch
- non-reproducible builds (e.g ci uses forest-devcontainer:1.0.1) but when we `git checkout 1.0.1 && docker build` we get an error :/

LesnyRumcajs:
I don't really see an issue. We could lock the tooling image tag in the repo. Not sure about GH Actions integration, it's just running a docker command so it's doable, no? Multiarch may be painful but we were able to provide multiarch images for Forest so unless there's some strange binary we manually download, we should be good there as well. 

I'm open to other suggestions.

aatifsyed:
(I've fleshed out some comments above)

aatifsyed:
different version of clippy in CI to local: https://github.com/ChainSafe/forest/actions/runs/5154069457/jobs/9282067031?pr=2893

LesnyRumcajs:
@aatifsyed what do we do with this?

PR #2367, opened 2022-12-15
Generate Daily Benchmark Results

jdjaustin:
**Summary of changes**
Changes introduced in this pull request:
Generate the following daily benchmark results:

1. Time to load a mainnet snapshot.
2. Validation time in tipsets per second.

**Reference issue to close (if applicable)**

Closes #2208 

**Other information and links**

This PR is just for the script. The script will be added to `forest-iac` under a separate PR.

PR #2876, opened 2023-05-15
stable rust toolchain

aatifsyed:
## Summary of changes

- Switch to stable rust toolchain
- Shim nightly for rustdoc (to generate an index page for self-hosted rustdoc)
- Drop rustfmt config (we included nightly-only features)
- Minor code changes to fixup formatting and unknown lints
- Drop most custom compile flags from `/cargo/config.toml`
- Surgical update of dependencies that we need to update to run on stable

Closes

## Change checklist

  - N/A

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,15 +2,7 @@
 incremental = true
 
 # In all cases, pass --cfg=tokio_unstable for tokio console integration
-# See(https://github.com/ChainSafe/forest/pull/2245)
-# on x86_64-linux-*, most CPUs support the specified vector instructions, so use those
-
-# Note, rustflags aren't additive, so we can't specify tokio_unstable in `build.rustflags`
-# So we use a `cfg(foo)` and `cfg(not(foo))` for our common flags.
-[target.'cfg(all(target_arch = "x86_64", target_os = "linux"))']
-rustflags = ["--cfg=tokio_unstable", "-Ctarget-feature=+avx2,+fma"]
-
-[target.'cfg(not(all(target_arch = "x86_64", target_os = "linux")))']
+# See (https://github.com/ChainSafe/forest/pull/2245)
 rustflags = ["--cfg=tokio_unstable"]
 
 [net]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,38 +1,17 @@
-[alias]
-b = "build"
-c = "check"
-d = "doc"
-t = "test"
-r = "run"
-
 [build]
 incremental = true
-# might be helpful but we don't enforce sccache installation atm
-# https://github.com/mozilla/sccache
-# rustc-wrapper = "sccache"
 
-# For details checkout https://jondot.medium.com/8-steps-for-troubleshooting-your-rust-build-times-2ffc965fd13e
-[target.x86_64-unknown-linux-gnu]
-# clang has been listed as prerequisite in the doc
-linker = "clang"
-# enable avx2 by default since it's available on almost all x86_64 CPUs
-rustflags = ["-Ctarget-feature=+avx2,+fma", "-Zshare-generics=y", "--cfg", "tokio_unstable"]
+# In all cases, pass --cfg=tokio_unstable for tokio console integration
+# See(https://github.com/ChainSafe/forest/pull/2245)
+# on x86_64-linux-*, most CPUs support the specified vector instructions, so use those
 
-[target.x86_64-unknown-linux-musl]
-# clang has been listed as prerequisite in the doc
-linker = "clang"
-# enable avx2 by default since it's available on almost all x86_64 CPUs
-rustflags = ["-Ctarget-feature=+avx2,+fma", "-Zshare-generics=y", "--cfg", "tokio_unstable"]
+# Note, rustflags aren't additive, so we can't specify tokio_unstable in `build.rustflags`
+# So we use a `cfg(foo)` and `cfg(not(foo))` for our common flags.
+[target.'cfg(all(target_arch = "x86_64", target_os = "linux"))']
+rustflags = ["--cfg=tokio_unstable", "-Ctarget-feature=+avx2,+fma"]
 
-[target.x86_64-apple-darwin]
-# zld might help here
-# brew install michaeleisel/zld/zld
-# "-Clink-arg=-fuse-ld=zld"
-# For details checkout https://jondot.medium.com/8-steps-for-troubleshooting-your-rust-build-times-2ffc965fd13e
-rustflags = ["-Zshare-generics=y", "--cfg", "tokio_unstable"]
-
-[target.aarch64-apple-darwin]
-rustflags = ["-Zshare-generics=y", "--cfg", "tokio_unstable"]
+[target.'cfg(not(all(target_arch = "x86_64", target_os = "linux")))']
+rustflags = ["--cfg=tokio_unstable"]
 
 [net]
 git-fetch-with-cli = true
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,8 @@
+# TODO(aatifsyed): remove this file once we have a dev/ci container
 [build]
 incremental = true
 
+# TODO(aatifsyed): remove - this can be pushed out to readme
 # In all cases, pass --cfg=tokio_unstable for tokio console integration
 # See (https://github.com/ChainSafe/forest/pull/2245)
 rustflags = ["--cfg=tokio_unstable"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,3 @@
-# TODO(aatifsyed): remove this file once we have a dev/ci container
 [build]
 incremental = true
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,7 +14,6 @@ on:
       - main
 
 env:
-  RUSTDOCFLAGS: "-Dwarnings"
   CACHE_TIMEOUT_MINUTES: 5
   SCCACHE_ENDPOINT: ${{ vars.SCCACHE_ENDPOINT }}
   SCCACHE_BUCKET: ${{ vars.SCCACHE_BUCKET }}
@@ -54,8 +53,13 @@ jobs:
       - name: Execute MDBook
         run: make mdbook-build
       - name: Build rustdoc
-        run: make rustdoc
+        run: cargo +nightly-2023-04-19 doc --workspace --no-deps
         env:
+        # When you visit https://chainsafe.github.io/forest/rustdoc you get an index.html
+        # listing all crates that are documented (copied out below).
+        # This isn't included by default, so we use the unstable `--enable-index-page` option
+        # https://doc.rust-lang.org/nightly/rustdoc/unstable-features.html#--index-page-provide-a-top-level-landing-page-for-docs
+          RUSTDOCFLAGS: -Dwarnings -Zunstable-options --enable-index-page
           CC: "sccache clang"
           CXX: "sccache clang++"
       - name: Link Checker (Docs)
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -55,8 +55,7 @@ jobs:
       - name: Build rustdoc
         run: |- 
           rustup toolchain install $toolchain --component rustfmt
-          cargo +$toolchain --version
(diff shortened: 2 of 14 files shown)


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2023-04-19"
+channel = "1.69"
```
LesnyRumcajs, line 2: nice

.github/workflows/docs.yml
```diff
@@ -14,8 +14,7 @@ on:
       - main
 
 env:
-  # https://doc.rust-lang.org/nightly/rustdoc/unstable-features.html#--index-page-provide-a-top-level-landing-page-for-docs
-  RUSTDOCFLAGS: "-Dwarnings -Zunstable-options --enable-index-page"
```
hanabi1224, line 18: So it looks like the self-hosted rustdoc will be broken

.rustfmt.toml
aatifsyed, line 1: These features have been unstable for a while, don't look likely to get stabilized soon (unless we e.g drive it...)

https://github.com/rust-lang/rustfmt/issues/4991
https://github.com/rust-lang/rustfmt/issues/3347
https://github.com/rust-lang/rustfmt/issues/5081
https://github.com/rust-lang/rustfmt/issues/5083

.cargo/config.toml
```diff
@@ -1,38 +1,9 @@
...
 [build]
 incremental = true
-# might be helpful but we don't enforce sccache installation atm
-# https://github.com/mozilla/sccache
-# rustc-wrapper = "sccache"
-
-# For details checkout https://jondot.medium.com/8-steps-for-troubleshooting-your-rust-build-times-2ffc965fd13e
-[target.x86_64-unknown-linux-gnu]
-# clang has been listed as prerequisite in the doc
-linker = "clang"
```
LesnyRumcajs, line 17: What is the implication of removing this? I believe it was added to reduce linking time. If we still want to remove it, the dependencies in both README and Dockerfiles (and Makefile) should be updated.

LesnyRumcajs (dismissed):

.cargo/config.toml
aatifsyed: Do we own linking time? Or is it owned by the developer?

I'd argue that we don't own any cargo config:
- Devs will have their own linkers set up in in `~/.cargo/config.toml`. E.g if they've install `mold`
- Fetching with CLI, cargo target directory etc, that's very developer-owned

The gray area here is `--cfg=tokio_unstable`, but I think that's a minor feature, and if users want to include it, we should provide instructions.

Note if we supported a devcontainer, we _would_ then own more of the developer experience - linking building etc, but at the moment we're a mess of both (I guess this is why I'm spending so much time here).

We don't really want to be messing with build config too much, because it ends up being pain - we're not build experts, we're rust experts.
One sample is the vectorised instruction support which we've mentioned earlier. This is another excellent example
- #2104 Added this https://github.com/ChainSafe/forest/pull/2104/files#diff-9a4f3e4537ebb7474452d131b0d969d89a51286f4269aac5ef268e712be17268R19
- #2200 Removes the link arg that actually speeds up linking, leaving us with technical debt in our build args 'linker - "clang"' that we're not sure works

BTW this arg does nothing on my machine, you actually need a linker if you want a speedup:
```
  './do-link.bash -C link-arg=-fuse-ld=mold' ran
    1.02 ± 0.11 times faster than './do-link.bash -C linker=clang -C link-arg=-fuse-ld=mold'
    1.21 ± 0.08 times faster than './do-link.bash -C linker=clang -C link-arg=-fuse-ld=lld'
    1.22 ± 0.08 times faster than './do-link.bash -C link-arg=-fuse-ld=lld'
    4.81 ± 0.32 times faster than './do-link.bash -C linker=clang'
    4.95 ± 0.42 times faster than './do-link.bash'
```
See benchmarking at https://github.com/ChainSafe/forest/commit/54fb5e8c29f3e28c3414cf5e63f2acddfec65b39

I'm generally all for build-time tuning, and developer experience, but I think the way we handle it at the moment is a bit haphazard, and we end up dropping balls outside of the actual source code

lemmih: I second this motion. I'd have more confidence in the correctness of Forest if we used the default configuration.

LesnyRumcajs: I mean, I overall agree that using defaults is the way to go. That being said if a couple of lines of configuration that are rarely changed bring 10-20% faster builds for thousands of workflow runs in the future - IMO it's worth having it as long as it is not really exotic piece of technology that changes a lot.

aatifsyed: > we do own the CI
Yeah that's an important point I missed.

# Current suggestion
- remove `linker = "clang"`, it doesn't do anything
- @elmattic is doing profiling on the avx2 stuff. I'm partial to removing it in any case

# Looking forward
cc #2826 

I think we should split responsibilites:
- forest only deals with writing and shipping minimal, correct rust (and javascript???) code
  (we bin our custom build config)

Then I'd lean towards a CI container with:
- custom build config - sccache, custom linkers, (maybe) cached deps
- cached installs of tooling and dev-deps, so we're never stuck on apt dependencies etc
- a nightly job that ensures our CI container is taking the latest tooling
- support for developers using CI containers as a devcontainer

I think we should then remove everything that we can here. Even the tokio console support should just be some `#[cfg(..)]` gated code, there's no need for us to touch that in-repo

There's some outstanding thinking on "where do locks" live - I don't really want to have to coordinate lockfiles across two repos, but maybe if we see the CI container as more of a "caching layer" then we can help ourselves a little bit

aatifsyed: P.S

# What is performance engineering?
I think vectorizing our code/dependencies is a matter of "performance engineering", maybe that should live in our CI repo - we compile and build/distribute an optimised/stripped binary. But where our product is now, I think we're a long way off - we don't have benchmarking infra, and our code quality/organisation isn't disciplined enough to be worried about e.g vectorisation - we should focus on a portable, correct and ergonomic implementation first and foremost

LesnyRumcajs: @aatifsyed Regarding "performance engineering", to give you a bit of context. We did quite a bit of benchmarking some time ago. Before that, the time of importing and syncing a snapshot was counted in many hours, right now it's below 1 hour on a regular developer machine. Most of the gains came from non-flag optimisations (changing the DB, checkpoints) but even then I feel uneasy when something may possibly deteriorate this. We have a benchmarking script (and its more advanced version coming up in https://github.com/ChainSafe/forest/pull/2367).

LesnyRumcajs: I'm fine with removing `clang` given the benchmarks you've submitted. If we can remove it from the dependencies, let's do it as well.

LesnyRumcajs: Removing `avx2` - I'm fine as long as the performance hit is not big, below 5%. Hopefully @elmattic will provide numbers.

LesnyRumcajs: As for other points - no strong opinion. I didn't have issues with the current tooling on either Linux Fedora or MacOS. NPM is sometimes down, but so are Github and Ubuntu repositories. I'm good with minimising the impact of such outages but the only ones that are really annoying me are Github outages. Little we can do about it.

LesnyRumcajs: I have spoken. :mountain:

LesnyRumcajs (commented):
Dismissing review so it's not blocked

.cargo/config.toml
aatifsyed: # Impact of `avx2`
I've not got enough hard drive space to use mainnet snapshots, so I've done it with calibnet instead.
## Methodology
I built the code with and without avx here: https://github.com/ChainSafe/forest/blob/c5b75126b84b1c5af98057187fe43a3f9a96cdff/build.bash
I ran the benchmark script here https://github.com/ChainSafe/forest/blob/c5b75126b84b1c5af98057187fe43a3f9a96cdff/benchmark.bash
I didn't have enough hard drive space to run against mainnet snapshots, so I used calibnet
## Importing
```
```
## Validating
```
```

.cargo/config.toml
```diff
@@ -1,38 +1,9 @@
...
-# enable avx2 by default since it's avaiable on almost all x86_64 CPUs
-rustflags = ["-Ctarget-feature=+avx2,+fma", "-Zshare-generics=y", "--cfg", "tokio_unstable"]
-
-[target.x86_64-unknown-linux-musl]
-# clang has been listed as prerequisite in the doc
-linker = "clang"
-# enable avx2 by default since it's avaiable on almost all x86_64 CPUs
-rustflags = ["-Ctarget-feature=+avx2,+fma", "-Zshare-generics=y", "--cfg", "tokio_unstable"]
```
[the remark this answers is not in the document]
aatifsyed: Resolved in https://github.com/ChainSafe/forest/pull/2876#discussion_r1197074372

aatifsyed:
We've pinned to `fvm 2.3.x` https://github.com/ChainSafe/fil-actor-states/pull/102
We need `fvm 2.4.x` to take this cross-compilation fix: https://github.com/filecoin-project/rust-fil-proofs/pull/1679
Otherwise cross compilation fails: https://github.com/ChainSafe/forest/actions/runs/5008225935/jobs/8975856604?pr=2876#step:9:2115

We should do whatever work we need to to take the latest `fvm 2.x.x` release.

.cargo/config.toml
elmattic: I have on my side:

```
Network: mainnet
Snapshot used: 2866560_2023_05_17T06_00_00Z.car
OS: Fedora 37
CPU: Intel Core i3-6300
```

| Command | Import Snapshot | Validation of 30 heights
|:---|---:|---:|
| `avx2` | 14m22s | 2m16s |
| `no_avx2` | 14m13s | 2m12s |

elmattic: If you can confirm this on a `mainnet` snapshot as well.

elmattic: I would say we can leave those instructions away, and re-assess performance later when needed.

.rustfmt.toml
```diff
@@ -1,5 +0,0 @@
-# Doc: https://rust-lang.github.io/rustfmt/
-imports_granularity = "Crate"
```
hanabi1224, line 2: I find these grouping rules pretty helpful. Any ideas on keeping the functionalities after the change?

aatifsyed: It would be nice, but not possible on stable - would be pretty easy to write a custom linter, but probably not a priority right now

aatifsyed: (see https://github.com/ChainSafe/forest/pull/2876#discussion_r1194520439)

.cargo/config.toml
LesnyRumcajs: Thanks for assessing this, LGTM.

.github/workflows/docs.yml
```diff
@@ -55,8 +53,16 @@ jobs:
...
+        run: |- 
+          rustup toolchain install $toolchain
+          cargo +$toolchain doc --workspace --no-deps
         env:
+          # When you visit https://chainsafe.github.io/forest/rustdoc you get an index.html
+          # listing all crates that are documented (copied out below).
+          # This isn't included by default, so we use the unstable `--enable-index-page` option
+          # https://doc.rust-lang.org/nightly/rustdoc/unstable-features.html#--index-page-provide-a-top-level-landing-page-for-docs
+          toolchain: nightly-2023-04-19
+          RUSTDOCFLAGS: -Dwarnings -Zunstable-options --enable-index-page
```
LesnyRumcajs, line 65: Any particular reason for moving it outside of the Makefile? I find it helpful to have the targets available locally, especially in case of a CI failure.

.cargo/config.toml
```diff
@@ -1,38 +1,11 @@
-[alias]
-b = "build"
-c = "check"
-d = "doc"
-t = "test"
-r = "run"
-
+# TODO(aatifsyed): remove this file once we have a dev/ci container
```
LesnyRumcajs, line 1: Let's have a tracking issue instead of this comment.

.github/workflows/docs.yml
aatifsyed: having `rustup install toolchain` in a makefile felt a bit smelly, but I suppose it's no different to the standard `make deps`. I'll move

lemmih (approved):
Nice!

PR #2885, opened 2023-05-16
Aatifsyed/2883 build profiles are too aggressive

aatifsyed:
## Summary of changes

- Don't abort the process on panic
- Keep debug info in debug builds
- Remove unused `quick` profile

Closes

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,6 @@ members = [
   "utils/forest_utils",
   "utils/forest_shim",
 ]
-resolver = "2"
 
 [workspace.dependencies]
 ahash = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,10 @@ members = [
   "utils/forest_shim",
 ]
 
+# We need this even though we have edition = "2021" because we're a virtual manifest
+# https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
+resolver = "2"
+
 [workspace.dependencies]
 ahash = "0.8"
 anes = "0.1.6"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -43,7 +43,6 @@ members = [
   "utils/forest_utils",
   "utils/forest_shim",
 ]
-resolver = "2"
```
aatifsyed, line 46: this is implied

Cargo.toml
```diff
@@ -191,16 +190,9 @@ forest_test_utils = { path = "./utils/test_utils" }
...
 
 [profile.dev]
-debug = 0
 split-debuginfo = "unpacked"
 
-[profile.quick]
-inherits = "release"
-opt-level = 1
-lto = "off"
-
```
aatifsyed, line 201: does anyone actually use this?

Cargo.toml
```diff
@@ -191,16 +190,9 @@ forest_test_utils = { path = "./utils/test_utils" }
 forest_utils = { path = "./utils/forest_utils" }
 
 [profile.dev]
-debug = 0
```
aatifsyed, line 194: why wouldn't we want debug info in a dev build?

Cargo.toml
aatifsyed: Thanks!

Cargo.toml
LesnyRumcajs: @aatifsyed https://github.com/ChainSafe/forest/pull/2918

Long story short, a `make test-all` would bloat to 71G (or more) with this. Given the overhead and how seldom one needs to go full GDB, it makes sense to disable it. Do you agree?

Issue #2811, opened 2023-05-01
`forest-cli db clean` shouldn't fail if there's nothing to clean


lemmih:
**Issue summary**

Running `forest-cli db clean` twice always fails (ie. forest returns error code 1) because the database folder cannot be deleted if it is missing. Forest-cli shouldn't fail in this case. We should either ignore it silently or print a message saying the database was already cleaned.

**Other information and links**

hanabi1224:
I'm not able to reproduce the issue. Any more detailed repro steps?
```console
➜  forest git:(main) cargo run -p forest-cli -- --chain calibnet db clean --force
    Finished dev [unoptimized] target(s) in 0.37s
     Running `target/debug/forest-cli --chain calibnet db clean --force`
Aborted. Database path /home/me/.local/share/forest/calibnet is not a valid directory
➜  forest git:(main) cargo run -p forest-cli -- --chain calibnet db clean --force
    Finished dev [unoptimized] target(s) in 0.39s
     Running `target/debug/forest-cli --chain calibnet db clean --force`
Aborted. Database path /home/me/.local/share/forest/calibnet is not a valid directory
➜  forest git:(main) cargo run -p forest-cli -- --chain mainnet db clean --force
    Finished dev [unoptimized] target(s) in 0.37s
     Running `target/debug/forest-cli --chain mainnet db clean --force`
Aborted. Database path /home/me/.local/share/forest/mainnet is not a valid directory
➜  forest git:(main) cargo run -p forest-cli -- --chain mainnet db clean --force
    Finished dev [unoptimized] target(s) in 0.38s
     Running `target/debug/forest-cli --chain mainnet db clean --force`
Aborted. Database path /home/me/.local/share/forest/mainnet is not a valid directory
➜  forest git:(main)
```

lemmih:
Hmm, indeed. I could have sworn it exited with code 1 on my droplet. Guess I must have been confused.

PR #2893, opened 2023-05-18
mark: 2873-modify-forest-and-forest-cli-to-prefer-compressed-files

aatifsyed:
## Summary of changes

-

Closes

## Change checklist


.clippy.toml
```diff
@@ -1,4 +1,5 @@
 # https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_types
+# TODO(aatifsyed): this doesn't have any justification
```
LesnyRumcajs, line 2: https://github.com/ChainSafe/forest/pull/2454

.config/forest.dic
aatifsyed, line 1: This was a real waste of time - it feels like I'm getting punished for writing documentation :/

.config/forest.dic
```diff
@@ -42,29 +46,41 @@ mainnet
...
+parse_filename
 pointer/SM
 PoSt
+recognised
 RPC
 SECP
 SECP256k1
 serializable
 serializer/SM
+snapshot_dir
```
LesnyRumcajs, line 66: In general, you can avoid putting those in the dictionary by surrounding them with backticks. This would allow removing several entries you've added here.

.config/forest.dic
```diff
@@ -42,29 +46,41 @@ mainnet
...
 milliGas
 multihash
 multisig
 mutex
 P2P
 parsable
+parse_filename
 pointer/SM
 PoSt
+recognised
```
LesnyRumcajs, line 60: Let's just use the US version in the documentation. We are using EN-US dictionary.
```suggestion
```

.config/forest.dic
```diff
@@ -13,8 +13,10 @@ callee
 CBOR
 CID/M
 CIDs
+cli
```
LesnyRumcajs, line 16: It would be nice to use only one form of this. I'm fine with either `cli` or `CLI`. Unless there's a reason for this?

forest/cli/src/cli/mod.rs
```diff
@@ -187,6 +187,7 @@ pub(super) fn print_stdout(out: String) {
         .unwrap();
 }
 
+// TODO(aatifsyed): replace with dialoguer
```
LesnyRumcajs, line 190: Let's have an issue with this - TODOs without a linked issue are expected to be resolved within the same PR. At least, that's my practice. 

Also, as far as I remember, we re-implemented it here for this reason: https://github.com/console-rs/dialoguer/issues/170

But it would be nice not to use such workarounds.

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -1,147 +1,85 @@
...
         #[arg(short, default_value = ".", verbatim_doc_comment)]
         output_path: PathBuf,
-        /// Export in zstd compressed format
-        #[arg(long)]
-        compressed: bool,
-        /// Skip creating the checksum file. Only valid when `--compressed` is
-        /// not supplied.
+        /// Skip creating the checksum file.
         #[arg(long)]
         skip_checksum: bool,
```
LesnyRumcajs, line 43: In general, given that this compression has in-built poor man's checksum, we could eliminate this altogether. Let's address it in another issue, though.

LesnyRumcajs (commented):
Partial review (I have to jump to a meeting)

.config/forest.dic
aatifsyed: It was a module name I think - the backticks tip was really important, thank you :) I'll add it to a README

forest/cli/src/cli/mod.rs
aatifsyed: https://github.com/ChainSafe/forest/issues/2928

scripts/gen_coverage_report.sh
```diff
@@ -30,7 +30,7 @@ function cov {
 cargo llvm-cov --workspace clean
 cargo llvm-cov --workspace --no-report
 cov forest-cli --chain calibnet db clean --force
-cov forest-cli --chain calibnet snapshot fetch --aria2 --provider filecoin --compressed -s "$TMP_DIR"
+cov forest-cli --chain calibnet snapshot fetch --aria2 --provider filecoin -s "$TMP_DIR"
```
LesnyRumcajs, line 33: Are there any subcommands that you introduced in this PR but didn't cover in Rust tests?

scripts/tests/forest_cli_check.sh
```diff
@@ -1,45 +1,70 @@
...
+
+function clean-snapshot-dir() {
+    rm --recursive --force --verbose -- "$snapshot_dir"
+    mkdir --parents -- "$snapshot_dir"
 }
-set -e
 
-echo "Validating as calibnet snapshot (uncompressed)"
-$FOREST_CLI_PATH --chain calibnet snapshot validate "$SNAPSHOT_DIRECTORY"/*.car --force
+# TODO(aatifsyed): I don't really understand what this does
```
LesnyRumcajs, line 30: It just calls one of the subcommands that we know we can call and should not fail.

utils/forest_utils/src/net/mod.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
+// TODO(aatifsyed): this module shouldn't exist
```
LesnyRumcajs, line 3: tracking issue

utils/forest_utils/src/lib.rs
```diff
@@ -1,14 +1,135 @@
...
+};
+use tokio::time::sleep;
+use tracing::error;
+
+/// Keep running the future created by `make_fut` until the timeout or retry
+/// limit in `args` is reached.
+/// `F` _must_ be cancel safe.
+// TODO(aatifsyed): store most recent error in RetryError
+// REFACTOR?(aatifsyed): use `futures-retry` or `again`
+// REFACTOR?(aatifsyed): return num retries
```
LesnyRumcajs, line 27: Tracking issue?

utils/forest_utils/src/lib.rs
```diff
@@ -1,14 +1,135 @@
...
+    #[default(Some(Duration::from_secs(1)))]
+    pub timeout: Option<Duration>,
+    #[default(Some(5))]
+    pub max_retries: Option<usize>,
+    #[default(Some(Duration::from_millis(200)))]
+    pub delay: Option<Duration>,
+}
+
+impl RetryArgs {
+    pub fn new_ms(
```
LesnyRumcajs, line 74: Does it simplify the client code a lot? Otherwise, I'd keep it general and accept any `Duration`. It would also allow smaller values for tests.

utils/forest_utils/src/io/progress_bar.rs
```diff
@@ -16,6 +27,17 @@ pub enum ProgressBarVisibility {
     Never,
 }
 
+impl ProgressBarVisibility {
+    /// Checks if stdout is a TTY
+    pub fn should_display(&self) -> bool {
```
LesnyRumcajs, line 32: I guess we could add a test that ensures it's false during unit tests. Not sure if we can instrument our tests to check the other condition. Not a big deal, though.

utils/forest_utils/src/io/progress_bar.rs
```diff
@@ -1,11 +1,22 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
+// JANK(aatifsyed): I don't really understand why this module exists, a lot of
```
LesnyRumcajs, line 3: If I recall correctly, it exists to not use `pbr` or other progress bar crates directly but to have it abstracted. This way, if/when our progress bar backend becomes abandonware or there's a new cool kid on crates.io, we don't have to change it everywhere. Plus, we ensure it's uniform - we don't want one part of the code to use one type of progress bar and another to use something else.

scripts/tests/forest_cli_check.sh
```diff
@@ -1,45 +1,70 @@
...
+    : : validating under mainnet chain should fail
+    if "$FOREST_CLI_PATH" --chain mainnet snapshot validate "$validate_me" --force; then
+        exit 1
+    fi
+
+    clean-snapshot-dir
+
+: intern a snapshot downloaded with aria2c
+    download_dir=$(mktemp --directory)
+    aria2c https://forest.chainsafe.io/calibnet/snapshot-latest.car.zst --dir="$download_dir"
```
LesnyRumcajs, line 62: Is the Forest fetch no longer using `aria2c` if possible?

forest/shared/src/snapshot.rs
```diff
@@ -0,0 +1,569 @@
...
+        .context("couldn't enumerate paths in snapshot directory")?;
+
+    // Sort them into pairs
+    let mut blobs_and_metadata = Vec::new();
+    while let Some(path) = paths.pop_first() {
+        match path.strip_suffix(METADATA_FILE_SUFFIX) {
+            Some(blob_path) => {
+                // we've popped the metadata file, try and pop the blob path
+                let blob_was_present = paths.remove(blob_path);
+                match blob_was_present {
```
LesnyRumcajs, line 117: How is a match on a boolean beneficial here, instead of a regular `if-else`?

forest/shared/src/snapshot.rs
```diff
@@ -0,0 +1,569 @@
...
+                    false => {
+                        warn!(%path, "deleting metadata without corresponding blob");
+                        let _ = std::fs::remove_file(path);
```
LesnyRumcajs, line 120: why do we need `let _` ?

forest/shared/src/snapshot.rs
```diff
@@ -0,0 +1,569 @@
...
+                        blobs_and_metadata.push((blob_path, metadata_path));
+                    }
+                }
+            }
+            None => {
+                // this is the blob path
+                let metadata_path = format!("{path}{METADATA_FILE_SUFFIX}");
+                let blob_path = path;
+                let metadata_was_present = paths.remove(&metadata_path);
+                match metadata_was_present {
```
LesnyRumcajs, line 134: Same as above, why do we need a `match` on a boolean?

forest/shared/src/snapshot.rs
```diff
@@ -0,0 +1,569 @@
...
+    // We use an enum to handle forward-incompatible changes in the future
+    V1 {
+        height: i64,
+        date: NaiveDate,
+        // The `forest` vendor doesn't include time
+        time: Option<chrono::NaiveTime>,
+        chain: String,
+        vendor: String,
+        // The stable url used
+        source_url: Option<String>,
```
LesnyRumcajs, line 201: Perhaps we could use a strong type here, like `Url`? Not a big deal, though.

forest/shared/src/snapshot.rs
```diff
@@ -0,0 +1,569 @@
...
+    let (std_file, path) =
+        tokio::task::spawn_blocking(|| anyhow::Ok(NamedTempFile::new()?.keep()?))
+            .await
+            .expect("NamedTempFile::new doesn't panic, and we didn't cancel/abort this task")
+            .context("couldn't create temporary file for download")?;
+    let mut dst = tokio::fs::File::from_std(std_file);
+    match tokio::io::copy(&mut src, &mut dst).await {
+        Ok(_) => Ok((path, url)),
+        Err(e) => {
+            // TODO(aatifsyed): we've maybe leaked the download here
```
LesnyRumcajs, line 357: tracking issue?

LesnyRumcajs (commented):
Nice PR. I'll test it a bit on my end.

It'd be great not to use `indicatif` directly but hide it behind our abstraction and remove `pbr`. It's not a good user experience when there are two types of progress bars used.

utils/forest_utils/src/net/mod.rs
aatifsyed: https://github.com/ChainSafe/forest/issues/2928

LesnyRumcajs (changes requested):
Blockers for now:
* can't download a snapshot on a machine with the temporary directory on a separate FS
* can't interrupt `forest` while it's downloading a snapshot with SIGINT.

.config/forest.dic
aatifsyed: Added spellcheck docs so this is less hellish

scripts/tests/forest_cli_check.sh
aatifsyed: Correct - if users want to use aria2c, they could (should?) do that and then `intern`.

Note a quick benchmark found no statistically significant different between the two, at least for the calibnet snapshot (n=20)

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `reqwest` | 9.045 ± 3.620 | 6.818 | 20.194 | 1.04 ± 0.51 |
| `aria2c` | 8.693 ± 2.491 | 5.722 | 14.846 | 1.00 |

```bash
hyperfine \
    --command-name=reqwest \
    './target/release/benchmark --out=/dev/null https://forest.chainsafe.io/calibnet/snapshot-latest.car.zst' \
    --command-name=aria2c \
    'aria2c --dir=/dev --out=null --file-allocation=none --allow-overwrite=true https://forest.chainsafe.io/calibnet/snapshot-latest.car.zst' \
    --export-markdown=benchmark.md \
    --export-json=benchmark.json \
    --export-csv=benchmark.csv \
    --show-output \
    --runs=20
```

```rust
use std::path::PathBuf;

use clap::Parser as _;
use futures::TryStreamExt as _;
use indicatif::ProgressBar;
use tap::Pipe as _;
use tokio::fs::File;
use url::Url;

#[derive(clap::Parser)]
pub struct Args {
    pub url: Url,
    #[arg(short, long)]
    pub out: PathBuf,
}

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let args = Args::parse();
    let progress_bar = ProgressBar::new(0);
    let mut dst = File::create(args.out).await?;
    let mut src = reqwest::get(args.url)
        .await?
        .error_for_status()?
        .bytes_stream()
        .map_err(|reqwest_error| std::io::Error::new(std::io::ErrorKind::Other, reqwest_error))
        .pipe(tokio_util::io::StreamReader::new)
        .pipe(|it| progress_bar.wrap_async_read(it));
    tokio::io::copy(&mut src, &mut dst).await?;
    Ok(())
}
```

forest/cli/src/cli/snapshot_cmd.rs
aatifsyed: #2933

scripts/tests/forest_cli_check.sh
lemmih: You missing the `--split=5` argument to aria2c. With this flag, `aria2c` is more than twice as fast with my internet (1Gbit/s wired connection). The difference is even bigger when using wifi. And in India and China, the difference should be bigger still.

LesnyRumcajs: If we follow this path, we could argue they could do it with `wget` on their machine anyway. As @lemmih noted, `aria2c` is significantly faster with correct settings, especially when downloading massive mainnet snapshots. I do believe it also supports resuming failed/interrupted downloads, which we supported with the `aria2c` path as far as I remember (but not native).

If anything, I'd consider having `aria2c` as a necessary dependency for Forest. That would annoy potential users that would get a runtime error, though.

forest/cli/tests/snapshot.rs
```diff
@@ -0,0 +1,245 @@
...
+    path::{Component, Path},
+};
+
+use assert_cmd::{assert::Assert, Command};
+use itertools::Itertools;
+use predicates::boolean::PredicateBooleanExt as _;
+use tap::Pipe as _;
+use tempfile::TempDir;
+
+/// Just check that the dir command produces some output
```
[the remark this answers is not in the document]
aatifsyed: Theoretically it's an implementation detail, and depends on xdg dirs. I don't want the integration tests to know too much about the inner workings of forest, that partly misses the point of integration testing

forest/daemon/src/daemon.rs
```diff
@@ -383,7 +383,46 @@ pub(super) async fn start(opts: CliOpts, config: Config) -> anyhow::Result<Rolli
...
+                )
+            },
+        )
+        .await
+        {
+            Ok(path) => {
+                config.client.snapshot_path = Some(path);
+                config.client.snapshot = true
+            }
+            Err(_) => todo!(),
```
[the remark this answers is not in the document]
aatifsyed: 😅

forest/daemon/src/daemon.rs
```diff
@@ -470,56 +509,16 @@ async fn propagate_error(services: &mut JoinSet<Result<(), anyhow::Error>>) -> a
...
     if config.client.snapshot_path.is_some() {
         return Ok(false);
     }
     let should_download = if !auto_download_snapshot && atty::is(atty::Stream::Stdin) {
-        let required_size: u64 = snapshot_fetch_size(config).await?;
-        let required_size = to_size_string(&required_size.into())?;
         Confirm::with_theme(&ColorfulTheme::default())
                 .with_prompt(
-                    format!("Forest needs a snapshot to sync with the network. Would you like to download one now? Required disk space {required_size}."),
+                    "Forest needs a snapshot to sync with the network. Would you like to download one now?",
```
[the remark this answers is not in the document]
aatifsyed: I removed it because it simplified the API, and didn't seem _that_ important. Adding back :)

forest/daemon/src/daemon.rs
```diff
@@ -647,39 +646,34 @@ mod test {
     use super::*;
 
     #[tokio::test]
-    async fn import_snapshot_from_file_valid() -> anyhow::Result<()> {
-        anyhow::ensure!(import_snapshot_from_file("test_files/chain4.car")
+    async fn import_snapshot_from_file_valid() {
+        import_snapshot_from_file("test_files/chain4.car")
```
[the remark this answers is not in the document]
aatifsyed: It's valid, just unneeded - `unwrap`-ing is clearer, especially for `unwrap_err` cases

utils/forest_utils/src/lib.rs
aatifsyed: It does simplify code - I've moved this to the test module for now, which is where it was used.

utils/forest_utils/src/io/progress_bar.rs
aatifsyed: That's an odd decision to me - we don't do that kind of wrapping for our other dependencies like dialoguer, and if we wanted to, we should use `impl Trait`. 

Agreed on uniformity though :P

scripts/tests/forest_cli_check.sh
aatifsyed: ack, adding back

forest/shared/src/snapshot.rs
aatifsyed: addressed elsewhere

forest/cli/tests/snapshot.rs
```diff
@@ -0,0 +1,245 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+//! Basic UI tests for the snapshot subcommand
+//! We only cover the `testable` aspects, so no fetching snapshots (and no validating snapshots)
+//!
+//! Where applicable, take care to write utility functions so that they can
+//! accept both &[TempDir] and &[std::path::PathBuf] (so they can be debugged if needed)
+//!
+//! See `forest_cli_check.sh` for those tests
+//! TODO(aatifsyed): remove that shellscript, and create some segregated tests
```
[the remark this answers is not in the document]
aatifsyed: #2945

scripts/tests/forest_cli_check.sh
```diff
@@ -1,45 +1,70 @@
...
+# TODO(aatifsyed): I don't really understand what this does
+"$FOREST_CLI_PATH" fetch-params --keys
+
+: fetch snapshot
+"$FOREST_CLI_PATH" --chain calibnet snapshot fetch --snapshot-dir "$snapshot_dir"
+clean-snapshot-dir
+
+
+: "clean database (twice)"
+# TODO(aatifsyed): I don't really understand what this tests
```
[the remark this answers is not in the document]
aatifsyed: It was a hypothetical bug apparently: https://github.com/ChainSafe/forest/issues/2811

Looks like you predicted my confusion before I even joined the company: https://github.com/ChainSafe/forest/pull/2814#discussion_r1184874451

forest/cli/src/cli/snapshot_cmd.rs
```diff
@@ -254,170 +199,141 @@ impl SnapshotCommands {
...
+                    // groupby works only on consecutive entries, so sort first
+                    .sorted_by_key(|(_, SnapshotMetadata::V1 { chain, .. })| chain)
+                    .group_by(|(_, SnapshotMetadata::V1 { chain, .. })| chain)
+                    .into_iter()
+                    .flat_map(|(_chain, snapshots_in_chain)| {
+                        snapshots_in_chain
+                            .max_by_key(|(_, SnapshotMetadata::V1 { height, .. })| height)
+                    })
+                    .map(|(path, _)| path)
+                    .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
aatifsyed: 🤷

.config/forest.dic
```diff
@@ -58,7 +59,9 @@ serializable
 serializer/SM
 statediff
 stateful
+stdout
 synchronizer
+syscalls
```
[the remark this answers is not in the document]
aatifsyed: I think we need `/SM` - I've added a README to actually explain some of this config

forest/cli/tests/snapshot.rs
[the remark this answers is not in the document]
LesnyRumcajs: Is it an implementation detail? I'm not saying we should compare the result to `$HOME/.local/forest/snapshots`. But checking that the given output is a valid path and it exists would avoid potential bugs, no? I recall we had some `directory does not exist` in the past.

https://github.com/ChainSafe/forest/blob/aatifsyed/2873-modify-forest-and-forest-cli-to-prefer-compressed-files/forest/shared/src/cli/config.rs#L125-L129

utils/forest_utils/src/io/progress_bar.rs
LesnyRumcajs: We should do this for dialoguer as well. The way we do it is a detail, I'm fine with either approach. What we want to achieve is a uniform UI.

forest/shared/src/snapshot.rs
LesnyRumcajs: Thanks, I didn't know of this idiom. Why are we ignoring the error? Shouldn't we at least log the error? Otherwise, potential issues may fly under our radar.

forest/cli/tests/snapshot.rs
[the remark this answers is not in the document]
aatifsyed: But this command is speculative - "show where snapshots _would_ be stored if I fetched them" - it doesn't promise to create the directory.

I'd argue the command itself needs to go - a user shouldn't care, we've probably only added it for testing.

Closing this as we're gonna rip all this out anyway

utils/forest_utils/src/io/progress_bar.rs
aatifsyed: #2928

forest/shared/src/snapshot.rs
aatifsyed: Closing this as we're gonna rip all this out anyway

forest/shared/src/snapshot.rs
aatifsyed: Closing this as this will all be ripped out anyway

blockchain/chain_sync/src/bad_block_cache.rs
```diff
@@ -17,7 +17,7 @@ pub struct BadBlockCache {
 
 impl Default for BadBlockCache {
     fn default() -> Self {
-        Self::new(forest_utils::const_option!(NonZeroUsize::new(1 << 15)))
+        Self::new(NonZeroUsize::new(1 << 15).unwrap())
```
lemmih, line 20: Can we use `nonzero!` here?

lemmih (changes requested):
Let's get rid of the `list`, `prune`,  `clean`, and `intern` snapshot commands.

blockchain/chain_sync/src/bad_block_cache.rs
aatifsyed: It wouldn't do anything - see https://chainsafeio.slack.com/archives/C0144LHN732/p1684843121482679

We _could_ make this const, but it's just as readable (and probably just as performant) as is

lemmih: It would fail at compile time rather than runtime. This compiles just fine `NonZeroUsize::new(0 << 15).unwrap()`, whereas this does not `nonzero!(0_usize << 15)`.

lemmih (changes requested):
We can also get rid of the `DefaultLocation` command as it should just be the current working directory.

lemmih (approved):

Makefile
```diff
@@ -82,7 +82,7 @@ audit:
 	cargo audit --ignore RUSTSEC-2020-0071
 
 spellcheck:
-	cargo spellcheck --code 1
+	cargo spellcheck --code 1 || echo "spellcheck failed. See .config/spellcheck.md for tips"
```
LesnyRumcajs, line 85: This will silently ignore any spellcheck errors, also in the CI. We want the CI to fail on this.

forest/daemon/tests/keystore_tests.rs
```diff
@@ -42,18 +40,17 @@ fn forest_headless_no_encrypt_no_passphrase_should_succeed() -> Result<()> {
     Ok(())
 }
 
+/// We should prompt the user if they want to create a password
```
LesnyRumcajs, line 43: I don't understand the reason behind the change in the logic here. Even if I provide the keystore phrase via an environmental variable, I should create my own phrase?

```
❯ FOREST_KEYSTORE_PHRASE=cthulhu target/release/forest --chain calibnet
2023-06-12T10:38:05.625877Z  INFO forest: Using default calibnet config
2023-06-12T10:38:05.626667Z  INFO forest::daemon: Starting Forest daemon, version 0.8.2+git.aaef338f0
2023-06-12T10:38:05.626688Z  INFO forest_libp2p::service: Could not decode networking keystore!
2023-06-12T10:38:05.626745Z  INFO forest_utils::io: Permissions set to 0600 on File { fd: 9, path: "/home/rumcajs/.local/share/forest/libp2p/keypair", read: false, write: true }
2023-06-12T10:38:05.626766Z  INFO forest::daemon: PeerId: 12D3KooWRsWRe4EGJ6B7UxZnXEJXcUP6p3v97dRAA1hHBmaFKbgs
2023-06-12T10:38:05.626772Z  WARN forest::daemon: Ignoring passphrase provided in FOREST_KEYSTORE_PHRASE for keystore creation
Create a password for Forest's keystore:
```

forest/shared/src/cli/config.rs
```diff
@@ -217,6 +141,7 @@ mod test {
 
     /// Partial configuration, as some parts of the proper one don't implement
     /// required traits (i.e. Debug)
+    // TODO(aatifsyed): the above comment is wrong
```
LesnyRumcajs, line 144: If it is wrong now (as the documentation tends to lag behind the code changes) then please modify it to reflect the current state. A wild TODO is, in general, not very helpful.

node/rpc/src/chain_api.rs
```diff
@@ -95,20 +94,14 @@ where
                 &start_ts,
                 recent_roots,
                 VoidAsyncWriter::default(),
-                compressed,
+                true,
```
LesnyRumcajs, line 97: For the sake of readability, I'd avoid `true/false` in non-obvious method calls. When you read this code as someone who does not know the parameters to `export` by heart, do you know what `true` stands for here? Let's just put a `let compressed = true` somewhere. What do you think? Alternatively we could use more explicit enums, but it may be an overkill.

LesnyRumcajs (changes requested):
Do we still need `is_zstd_compressed` and the logic in `get_fetch_progress*`?

Thanks for addressing the `tempfs` issue.

forest/shared/src/cli/config.rs
LesnyRumcajs: Or, even better, let's remove this structure altogether. It can be done in a different PR, but let's not leave untracked TODOs behind.

forest/daemon/tests/keystore_tests.rs
aatifsyed: I was surprised that we'd implicitly *create* (not load) a passphrase with no confirmation. It's relatively easy to lose the value of an environment variable (that's why they're often used for storing secrets), and I wouldn't want the user to end up with a system they didn't know the password to unless they had confirmed their password. I will back this change out if you think this is important.

Makefile
aatifsyed: whoops, sorry that wasn't my intention :D

forest/daemon/tests/keystore_tests.rs
LesnyRumcajs: Unless something changed, this was also the only way to provide the passphrase in headless mode, i.e., setting up Forest and running it with some scripts. This is not really meant for a "front-end" user.

node/rpc/src/chain_api.rs
aatifsyed: Yep, `rust-analyzer` just make this code look like `compressed: true`

forest/shared/src/cli/config.rs
aatifsyed: #2965

node/rpc/src/chain_api.rs
LesnyRumcajs: I don't believe we should rely on the IDE features when it comes to readability. Browsing via GitHub should provide a similar experience (especially useful when reviewing PRs).

forest/daemon/tests/keystore_tests.rs
aatifsyed: Yep, added back :)

Makefile
```diff
@@ -82,7 +82,7 @@ audit:
 	cargo audit --ignore RUSTSEC-2020-0071
 
 spellcheck:
-	cargo spellcheck --code 1
+	if ! cargo spellcheck --code 1; then echo "See .config/spellcheck.md for tips" && false; fi
```
LesnyRumcajs, line 85: I'd make it less verbose (and, for me, more readable), but it's just a preference.
```suggestion
	cargo spellcheck --code 1 || (echo "spellcheck failed. See .config/spellcheck.md for tips"; false)
```

LesnyRumcajs (approved):

PR #2908, opened 2023-05-24
feat: impl CborStoreExt trait to omit code parameter in `pub_cbor`

hanabi1224:
## Summary of changes

- Add `fn put_cbor_default` extension method to `Blockstore` that omits `code` parameter to match [store put API in `go`](https://github.com/ipfs/go-ipld-cbor/blob/v0.0.6/store.go#L92), to simplify code and make translation easier

Closes

## Change checklist

### DIFF
--- a/.config/forest.dic
+++ b/.config/forest.dic
@@ -1,4 +1,4 @@
-68
+73
 Algorand/M
 API/M
 APIs
@@ -44,6 +44,7 @@ mempool
 Merkle
 MiB
 milliGas
+multihash
 multisig
 mutex
 P2P
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3416,6 +3416,7 @@ dependencies = [
  "futures",
  "fvm_ipld_blockstore",
  "fvm_ipld_car",
+ "fvm_ipld_encoding 0.2.3",
  "log",
  "tokio",
  "tokio-util",
--- a/blockchain/chain/src/store/chain_store.rs
+++ b/blockchain/chain/src/store/chain_store.rs
@@ -7,7 +7,7 @@ use ahash::{HashMap, HashMapExt, HashSet};
 use anyhow::Result;
 use async_compression::futures::write::ZstdEncoder;
 use bls_signatures::Serialize as SerializeBls;
-use cid::{multihash::Code::Blake2b256, Cid};
+use cid::Cid;
 use digest::Digest;
 use forest_beacon::{BeaconEntry, IGNORE_DRAND_VAR};
 use forest_blocks::{Block, BlockHeader, FullTipset, Tipset, TipsetKeys, TxMeta};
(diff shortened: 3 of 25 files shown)


lemmih (changes requested):
I'm having difficulties understanding this code, and I don't think being aligned with the Golang code is an end goal. Does this trait replace `BlockstoreExt`? If not, why not?

hanabi1224:
>  I don't think being aligned with the Golang code is an end goal.

@lemmih Depending on the desired process for state migration. If we implement state migrations against a spec, I don't think we need to align with Go code, but if the process is to translate Go code, it will be easier to have common utilities aligning their APIs with the same utilities in Go, to avoid repetitive code and the effort to figure out the difference. In this case, `store.put(obj)` is a common utility to generate CID with hard-coded DAG-CBOR code and blake2b256 hasher on Go side, having a `pub_cbor_default` that takes the same API on the rust side is much more convenient, does that make sense?

lemmih:
> > I don't think being aligned with the Golang code is an end goal.
> 
> @lemmih Depending on the desired process for state migration. If we implement state migrations against a spec, I don't think we need to align with Go code, but if the process is to translate Go code, it will be easier to have common utilities aligning their APIs with the same utilities in Go, to avoid repetitive code and the effort to figure out the difference. In this case, `store.put(obj)` is a common utility to generate CID with hard-coded DAG-CBOR code and blake2b256 hasher on Go side, having a `pub_cbor_default` that takes the same API on the rust side is much more convenient, does that make sense?

Our code has to make sense without looking at the Golang code. If you want an interface that matches the Golang code, it requires a lot of documentation to ensure that it won't be refactored in the future.

So, I see two options:
1. We mirror the Golang interface, keep that code separate from the rest of the Forest code, and write documentation that makes it states not to refactor the code even though it looks non-idiomatic.
2. We focus on writing an idiomatic Rust solution and keep a Markdown document describing how the Golang implementation relates to the Rust implementation.

I think option 2 is the most sustainable way forward.

LesnyRumcajs (approved):

hanabi1224:
> not to refactor the code even though it looks non-idiomatic.

@lemmih  Could u elaborate on why omitting a parameter makes rust code non-idiomatic?

lemmih:
> > not to refactor the code even though it looks non-idiomatic.
> 
> @lemmih Could u elaborate on why omitting a parameter makes rust code non-idiomatic?

It has nothing to do with omitting a parameter. With this PR, we have two functions for doing the same thing: `put_obj` and `put_cbor_default`. Why not get rid of `put_obj`?

lemmih:
> > Why not get rid of `put_obj`?
> 
> @lemmih IMO `put_obj` is a duplicate of `put_cbor`that is defined in `CborStore`, it's fine to remove it and I'd love to if u ask but it's not related to the goal of this PR which adds another function that does not take `code` parameter.
> 
> But removing `pub_obj` requires larger refactoring than expected because `CborStore` has `Sized` constraint while `BlockStoreExt` is implemented for `&Blockstore`

This PR adds a function that replaces `put_obj` but doesn't delete `put_obj`. Both should be done in the same PR. Let's discuss this in Slack.

lemmih (approved):
Let's replace all calls to `put_cbor` with `put_cbor_default`.

hanabi1224:
> Let's replace all calls to `put_cbor` with `put_cbor_default`.

@lemmih  Done.

PR #2909, opened 2023-05-24
feat: add `fn new_dag_cbor_default` to `Cid`

hanabi1224:
## Summary of changes

- Add extension method `fn new_dag_cbor_default` to `Cid` to match the default filecoin CID builder in go. See [code](https://github.com/filecoin-project/go-state-types/blob/master/abi/cid.go#L49)

Closes

## Change checklist


lemmih (changes requested):
Can you explain a little more about why we want this?

hanabi1224:
> Can you explain a little more about why we want this?

@lemmih In state migration code, sometimes it needs to calculate CID for a cor-serializable object, without calling `CborStore.put_cbor_default` API. One example is calculating CID for DealProposal (see https://github.com/ChainSafe/fil-actor-states/pull/120), I feel it unsustainable to add a method to each struct, it would be much easier to do it with `Cid::new_dag_cbor_default(deal_proposal)`

utils/forest_utils/src/cid/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use cid::{
+    multihash::{Code::Blake2b256, MultihashDigest},
+    Cid,
+};
+
+/// Extension methods for constructing `dag-cbor` [Cid]
+pub trait CidCborExt {
+    /// Default CID builder for Filecoin
```
LesnyRumcajs, line 11: Such a lovely new trait! It would be a shame not to have 100% coverage on it! :)

LesnyRumcajs:
@hanabi1224, what do we do with this PR?

utils/forest_utils/src/cid/mod.rs
hanabi1224: test added

utils/forest_utils/src/cid/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    ///
+    /// - The default codec is CBOR
```
lemmih, line 13: The codec is `DAG_CBOR`, right?

utils/forest_utils/src/cid/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// - The default hash function is 256 bit BLAKE2b
+    ///
+    /// This matches [`abi.CidBuilder`](https://github.com/filecoin-project/go-state-types/blob/master/abi/cid.go#L49) in go
+    fn new_dag_cbor_default<S: serde::ser::Serialize>(obj: &S) -> anyhow::Result<Cid> {
```
lemmih, line 17: The naming feels non-intuitive to me. I can see that the `default` part refers to `blake256,` but I don't see any non-default counterpart. How about naming this function `from_cbor` or `from_cbor_blake256`?

lemmih (changes requested):

utils/forest_utils/src/cid/mod.rs
hanabi1224: Fixed.

LesnyRumcajs:
@lemmih @hanabi1224, what do we do with this PR?

lemmih (approved):

lemmih:
> @lemmih @hanabi1224, what do we do with this PR?

I'm okay with merging it.

PR #2911, opened 2023-05-25
fix: replace BlockstoreExt::{put_obj, get_obj} with CborStore::{put_cbor, get_cbor}

hanabi1224:
## Summary of changes

Changes introduced in this pull request:

- replace `BlockstoreExt::{put_obj, get_obj}` with `CborStore::{put_cbor, get_cbor}`

## Reference issue to close (if applicable)

Closes

## Change checklist

- [ ] I have performed a self-review of my own code,
- [ ] I have made corresponding changes to the documentation,
- [ ] I have added tests that prove my fix is effective or that my feature works (if possible),
- [ ] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

Issue #2718, opened 2023-03-28
Forest run into unrecoverable error state in evaluating network head


hanabi1224:
**Describe the bug**

Forest run into unrecoverable error state in evaluating network head in sync check container
epoch: 2719200 

The error is transient, another forest instance of mine worked fine

**Log output**

<details>
  <summary>Log Output</summary>
  
  ```
2023-03-27T12:49:31.887120Z  INFO forest_chain_sync::chain_muxer: Local node is behind the network, starting BOOTSTRAP from LOCAL_HEAD = 2719200 -> NETWORK_HEAD = 2720499
2023-03-27T12:49:31.887214Z ERROR forest_chain_sync::chain_muxer: Bootstrapping failed, re-evaluating the network head to retry the bootstrap. Error = TipsetRangeSyncer(TipsetRangeWithBadBlock(Cid(bafy2bzacecozeyvky3hqbaf5taswa53yz6blyckcoxinmovqtgjcp4nir7upe), "chain contained bafy2bzaceaobakmfdjtsw7b4twvrvjgvhhix7hwxwpfwpwb2fb5sd7lsil5pk"))
2023-03-27T12:49:31.887224Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-27T12:50:15.793043Z  INFO forest_chain_sync::chain_muxer: Local node is behind the network, starting BOOTSTRAP from LOCAL_HEAD = 2719200 -> NETWORK_HEAD = 2720500
2023-03-27T12:50:15.793220Z ERROR forest_chain_sync::chain_muxer: Bootstrapping failed, re-evaluating the network head to retry the bootstrap. Error = TipsetRangeSyncer(TipsetRangeWithBadBlock(Cid(bafy2bzaceckveinsh2jtlzxtsap2yvywgjoewhuqwy2ximrs3h6jdiq5fncuk), "chain contained bafy2bzacecozeyvky3hqbaf5taswa53yz6blyckcoxinmovqtgjcp4nir7upe"))
2023-03-27T12:50:15.793236Z  INFO forest_chain_sync::chain_muxer: Evaluating network head...
2023-03-27T12:50:37.229879Z  WARN forest_libp2p::hello::behaviour: sending on a closed channel
2023-03-27T12:51:07.749109Z  INFO forest_chain_sync::chain_muxer: Local node is behind the network, starting BOOTSTRAP from LOCAL_HEAD = 2719200 -> NETWORK_HEAD = 2720502
[...]

PR #2912, opened 2023-05-25
Revert "fix: replace BlockstoreExt::{put_obj, get_obj} with CborStore::{put_cbor, get_cbor}"

hanabi1224:
Reverts ChainSafe/forest#2911 in case it causes state missmatch on mainnet


LesnyRumcajs:
It's a pre-existing issue. https://github.com/ChainSafe/forest/issues/2718

PR #2941, opened 2023-06-06
chore: add blocking version of compute_tipset_state

lemmih:
## Summary of changes

- Split `compute_tipset_state` into blocking and non-blocking variants. The blocking version will be used in the devnet PR.

Closes

## Change checklist

### DIFF
--- a/blockchain/state_manager/src/lib.rs
+++ b/blockchain/state_manager/src/lib.rs
@@ -463,7 +463,7 @@ where
                     let no_func =
                         None::<fn(&Cid, &ChainMessage, &ApplyRet) -> Result<(), anyhow::Error>>;
                     let ts_state = self
-                        .compute_tipset_state_non_blocking(Arc::clone(tipset), no_func)
+                        .compute_tipset_state(Arc::clone(tipset), no_func)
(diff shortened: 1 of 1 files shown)


blockchain/state_manager/src/lib.rs
```diff
@@ -734,9 +738,24 @@ where
     /// Performs a state transition, and returns the state and receipt root of
     /// the transition.
     #[instrument(skip(self, callback))]
-    pub async fn compute_tipset_state<CB: 'static>(
+    pub async fn compute_tipset_state_non_blocking<CB: 'static>(
```
elmattic, line 741: Any rationale about changing name for `async` version and not using `compute_tipset_state_blocking` for the blocking call?

Not saying it's better (I'm not sure if there are official naming schemes for this), just curious.

elmattic (approved):

blockchain/state_manager/src/lib.rs
lemmih: Hmm, not much thought went into it. I think `compute_tipset_state_blocking` is nicer.

Issue #2726, opened 2023-03-29
Send FIL failed


LesnyRumcajs:
**Describe the bug**

The wallet test failed, and the FIL was not sent. The target wallet didn't receive anything and the transaction is not visible in the source wallet. https://calibration.filscan.io/address/general?address=t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y
(it should've happened around 2023-03-23 17:47:17, target t1fye76pfymru4y2fqbgdxswpwqt3ukeems2nzfna)

This means that either we have a bug in the test setup or the send command doesn't always send the funds, which would be a **significant** issue. 

**To Reproduce** Steps to reproduce the behaviour:
It happened in this PR https://github.com/ChainSafe/forest/pull/2709 (which didn't introduce any logic changes)

The job failed: https://github.com/ChainSafe/forest/actions/runs/4503553182/jobs/7926889785

**Log output**

```
Wallet tests
Importing preloaded wallet key
t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y
Exporting key
Fetching metrics
--2023-03-23 17:47:17--  http://localhost:6116/metrics
Resolving localhost (localhost)... ::1, 127.0.0.1
Connecting to localhost (localhost)|::1|:6116... failed: Connection refused.
Connecting to localhost (localhost)|127.0.0.1|:6116... connected.
HTTP request sent, awaiting response... 200 OK
Length: 28751 (28K) [text/plain]
Saving to: ‘metrics.log’

     0K .......... .......... ........                        100%  404M=0s

2023-03-23 17:47:17 (404 MB/s) - ‘metrics.log’ saved [28751/28751]

Listing wallet balances
Address                                   Default Balance
t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y           ~100 FIL
Creating a new address to send FIL to
t1fye76pfymru4y2fqbgdxswpwqt3ukeems2nzfna
Listing wallet balances
Address                                   Default Balance
t1ac6ndwj6nghqbmtbovvnwcqo577p6ox2pt52q2y  X        ~100 FIL
t1fye76pfymru4y2fqbgdxswpwqt3ukeems2nzfna           0 atto FIL
Sending FIL to the above address
Checking balance of t1fye76pfymru4y2fqbgdxswpwqt3ukeems2nzfna...
Checking balance 1/10
Checking balance 2/10
Checking balance 3/10
Checking balance 4/10
[...]

PR #2942, opened 2023-06-06
Add back wallet check test

elmattic:
## Summary of changes

This PR is adding back the check to the CI for trying to get more context about this Send FIL issue.

- Add warning in case of unpublished messages
- Add `--show` flag to send command for printing the message
- Add back check of balance so we know when it happens
- Add 3 rounds of testing

Closes

## Change checklist


lemmih:
Hmm, I'm not keen on using the CI for this kind of debugging. How about we do this testing on a droplet?

elmattic:
> Hmm, I'm not keen on using the CI for this kind of debugging. How about we do this testing on a droplet?

I agree, it's not ideal. Let me propose something else. I can remove all the 3 rounds so it would not be more computation heavy than what we have today.
Once the check is failing again, then I will disable it so it doesn't get in the way (hopefully we will have more insights this time).
I've been looking at a dozen of Wallet tests lately and couldn't see the issue again.

lemmih:
> > Hmm, I'm not keen on using the CI for this kind of debugging. How about we do this testing on a droplet?
> 
> I agree, it's not ideal. Let me propose something else. I can remove all the 3 rounds so it would not be more computation heavy than what we have today. Once the check is failing again, then I will disable it so it doesn't get in the way (hopefully we will have more insights this time). I've been looking at a dozen of Wallet tests lately and couldn't see the issue again.

Wouldn't it be just as good to run the `calibnet_wallet_check.sh` script in a loop on a droplet? We could set up 10 Forest nodes in parallel to run this script in a loop.

elmattic:
> > > Hmm, I'm not keen on using the CI for this kind of debugging. How about we do this testing on a droplet?
> > 
> > 
> > I agree, it's not ideal. Let me propose something else. I can remove all the 3 rounds so it would not be more computation heavy than what we have today. Once the check is failing again, then I will disable it so it doesn't get in the way (hopefully we will have more insights this time). I've been looking at a dozen of Wallet tests lately and couldn't see the issue again.
> 
> Wouldn't it be just as good to run the `calibnet_wallet_check.sh` script in a loop on a droplet? We could set up 10 Forest nodes in parallel to run this script in a loop.

Yes, I can have a try, but I'm not sure if I will be able to reproduce exact same conditions that on the CI.

For example, the error Hubert managed to produce yesterday (https://github.com/ChainSafe/forest/issues/2726#issuecomment-1581147480) is not the same one we had on the CI. Because in his case there's an error but on our side we managed to print the message `CID` (meaning the message was successfully added to the pool).

lemmih:
@elmattic Is the `--show` command something we want long-term? If not, let's close this PR and set up the testing separately on a droplet.

elmattic:
> @elmattic Is the `--show` command something we want long-term? If not, let's close this PR and set up the testing separately on a droplet.

No. Once we have `mpool pending` we won't need it because we will be able to find it thanks to message CID (returned by send subcommand).

PR #2959, opened 2023-06-09
Fix chainxchg race issue

elmattic:
## Summary of changes

- Fix race issue when waiting for `chainxchg` results

Closes

## Change checklist

### DIFF
--- a/blockchain/chain_sync/src/network_context.rs
+++ b/blockchain/chain_sync/src/network_context.rs
@@ -191,6 +191,7 @@ where
         let global_pre_time = SystemTime::now();
         let network_failures = Arc::new(AtomicU64::new(0));
         let lookup_failures = Arc::new(AtomicU64::new(0));
+        let total_failures = Arc::new(AtomicU64::new(0));
         let chain_exchange_result = match peer_id {
             // Specific peer is given to send request, send specifically to that peer.
             Some(id) => Self::chain_exchange_request(
@@ -208,7 +209,10 @@ where
                 let (result_tx, result_rx) = flume::bounded::<Vec<T>>(1);
                 // No specific peer set, send requests to a shuffled set of top peers until
                 // a request succeeds.
-                let peers = self.peer_manager.top_peers_shuffled().await;
+                let mut peers = self.peer_manager.top_peers_shuffled().await;
+                peers.truncate(1);
+                let num_peers = peers.len();
+
                 let mut tasks = JoinSet::new();
                 for peer_id in peers.into_iter() {
                     let n_task_control_tx = n_task_control_tx.clone();
@@ -219,6 +223,7 @@ where
                     let request = request.clone();
                     let network_failures = network_failures.clone();
                     let lookup_failures = lookup_failures.clone();
+                    let total_failures = total_failures.clone();
                     tasks.spawn(async move {
                         if n_task_control_tx.send_async(()).await.is_ok() {
                             match Self::chain_exchange_request(
@@ -236,6 +241,7 @@ where
                                         }
                                         Err(e) => {
                                             lookup_failures.fetch_add(1, Ordering::Relaxed);
+                                            total_failures.fetch_add(1, Ordering::Relaxed);
                                             _ = n_task_control_rx.recv_async().await;
                                             debug!("Failed chain_exchange response: {e}");
                                         }
@@ -243,18 +249,35 @@ where
                                 }
                                 Err(e) => {
                                     network_failures.fetch_add(1, Ordering::Relaxed);
+                                    total_failures.fetch_add(1, Ordering::Relaxed);
                                     _ = n_task_control_rx.recv_async().await;
                                     debug!(
                                         "Failed chain_exchange request to peer {peer_id:?}: {e}"
                                     );
                                 }
                             }
+                        } else {
+                            total_failures.fetch_add(1, Ordering::Relaxed);
                         }
                     });
                 }
 
-                async fn wait_all<T: 'static>(tasks: &mut JoinSet<T>) {
-                    while tasks.join_next().await.is_some() {}
+                async fn wait_all<T: 'static>(
+                    tasks: &mut JoinSet<T>,
+                    total_failures: Arc<AtomicU64>,
+                    num_peers: usize,
+                ) {
+                    loop {
+                        if tasks.join_next().await.is_none() {
+                            let failures = total_failures.load(Ordering::Relaxed);
+                            if failures == num_peers as u64 {
+                                // So far every task have failed
+                                return;
+                            } else {
+                                tokio::task::yield_now().await;
+                            }
+                        }
+                    }
                 }
 
                 let make_failure_message = || {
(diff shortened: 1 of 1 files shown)


blockchain/chain_sync/src/network_context.rs
```diff
@@ -62,6 +65,47 @@ impl<DB: Clone> Clone for SyncNetworkContext<DB> {
     }
 }
 
+type RaceBatchFuture<T> = Pin<Box<dyn Future<Output = Result<T, String>> + Send>>;
```
LesnyRumcajs, line 68: Let's have some documentation and possibly a set of unit tests for this. It's not clear from a first look of this for someone unfamiliar with synchronization primitives.

CHANGELOG.md
```diff
@@ -59,6 +59,8 @@
 
 - [#2950](https://github.com/ChainSafe/forest/pull/2950): Fix cases where ctrl-c
   would be ignored.
+- [#2934](https://github.com/ChainSafe/forest/issues/2934): Fix race condition
+  when calling chain exchange in case of low peers
```
lemmih, line 63: We can inform users that this bug won't be triggered by normal use:
```suggestion
- [#2934](https://github.com/ChainSafe/forest/issues/2934): Fix race condition
  when connecting to development blockchains.
```

blockchain/chain_sync/src/network_context.rs
```diff
@@ -62,6 +65,47 @@ impl<DB: Clone> Clone for SyncNetworkContext<DB> {
...
+            semaphore: Arc::new(Semaphore::new(max_concurrent_jobs)),
+        }
+    }
+
+    pub fn add(&mut self, future: RaceBatchFuture<T>) {
+        let sem = self.semaphore.clone();
+        self.tasks.spawn(async move {
+            let permit = sem.acquire_owned().await.unwrap();
+            let result = future.await;
+            drop(permit);
```
lemmih, line 91: This slightly differs from the previous code as it doesn't guarantee that the futures are run in order. But it's OK because we've already shuffled the peers.

blockchain/chain_sync/src/network_context.rs
```diff
@@ -272,13 +300,11 @@ where
...
-                        log::debug!("Succeed: handle_chain_exchange_request");
-                        result.map_err(|e| e.to_string())?
-                    },
-                    _ = wait_all(&mut tasks) => return Err(make_failure_message()),
+                if let Some(v) = batch.get_ok().await {
+                    log::debug!("Succeed: handle_chain_exchange_request");
+                    v
+                } else {
+                    return Err(make_failure_message());
                 }
```
lemmih, line 308:
```suggestion
                let v = batch.get_ok().await.ok_or_else(make_failure_message)?;
                log::debug!("Succeed: handle_chain_exchange_request");
                v
```

blockchain/chain_sync/src/network_context.rs
```diff
@@ -62,6 +65,47 @@ impl<DB: Clone> Clone for SyncNetworkContext<DB> {
...
+    T: Send + 'static,
+{
+    pub fn new(max_concurrent_jobs: usize) -> Self {
+        RaceBatch {
+            tasks: JoinSet::new(),
+            semaphore: Arc::new(Semaphore::new(max_concurrent_jobs)),
+        }
+    }
+
+    pub fn add(&mut self, future: RaceBatchFuture<T>) {
```
lemmih, line 86: I don't think the `Pin` is necessary. Something like this should also work:
```suggestion
    pub fn add(&mut self, future: impl Future<Output = Result<T, String>> + Send + 'static) {
```

lemmih (commented):
It may be simpler just to directly use a `JoinSet` and `Semaphore` in `handle_chain_exchange_request`.

elmattic:
> It may be simpler just to directly use a `JoinSet` and `Semaphore` in `handle_chain_exchange_request`.

It could help to keep the code in a dedicated struct for testing purposes, like Hubert is suggesting, no?

blockchain/chain_sync/src/network_context.rs
```diff
@@ -62,6 +64,45 @@ impl<DB: Clone> Clone for SyncNetworkContext<DB> {
...
+        RaceBatch {
+            tasks: JoinSet::new(),
+            semaphore: Arc::new(Semaphore::new(max_concurrent_jobs)),
+        }
+    }
+
+    pub fn add(&mut self, future: impl Future<Output = Result<T, String>> + Send + 'static) {
+        let sem = self.semaphore.clone();
+        self.tasks.spawn(async move {
+            let permit = sem.acquire_owned().await.unwrap();
```
lemmih, line 86: Unwraps make me nervous.

```suggestion
            let permit = sem.acquire_owned().await.map_err(|_| "Semaphore unexpectedly closed")?;
```

blockchain/chain_sync/src/network_context.rs
elmattic: Added some unit tests!

blockchain/chain_sync/src/network_context.rs
```diff
@@ -62,6 +64,48 @@ impl<DB: Clone> Clone for SyncNetworkContext<DB> {
     }
 }
 
+struct RaceBatch<T> {
```
lemmih, line 67: Would be great with some documentation here as well.

blockchain/chain_sync/src/network_context.rs
```diff
@@ -62,6 +64,53 @@ impl<DB: Clone> Clone for SyncNetworkContext<DB> {
...
+                .await
+                .map_err(|_| "Semaphore unexpectedly closed")?;
+            let result = future.await;
+            drop(permit);
+            result
+        });
+    }
+
+    /// Return first finishing `Ok` future else return `None` if all jobs failed
+    pub async fn get_ok(&mut self) -> Option<T> {
```
lemmih, line 102: This should take ownership of `self.` Then `abort_all` won't be necessary (it happens automatically when the `JoinSet` is dropped.)

lemmih (approved):
The top-level documentation could be shorter: `Race tasks to completion while limiting the number of tasks that may execute concurrently. Once a task finishes without error, the rest of the tasks are canceled.`

blockchain/chain_sync/src/network_context.rs
elmattic: Yes, that's much better.

LesnyRumcajs (approved):

LesnyRumcajs:
I like this implementation, by the way. :)

blockchain/chain_sync/src/network_context.rs
```diff
@@ -62,6 +64,49 @@ impl<DB: Clone> Clone for SyncNetworkContext<DB> {
     }
 }
 
+/// Race tasks to completion while limiting the number of tasks that may execute concurrently.
+/// Once a task finishes without error, the rest of the tasks are canceled.
+struct RaceBatch<T> {
```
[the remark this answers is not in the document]
elmattic: Yeah why not if we need it at some other place in the future, it's not fully generic yet (i.e. error is a fixed `String`).

Issue #2967, opened 2023-06-12
Forest fails to fetch params. 


ruseinov:
**Describe the bug**

Forest fails to fetch params due to an error fetching root certificates.

**To Reproduce** Steps to reproduce the behavior:

1. `make install-with-rustalloc`
2. Run '.forest  --chain calibnet --no-gc --encrypt-keystore false'
3. See error

**Log output**

```
thread 'tokio-runtime-worker' panicked at 'could not load platform certs: Custom { kind: Other, error: Error { code: -25300, message: "OSStatus -25300" } }', /Users/romanuseinov/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-rustls-0.23.2/src/config.rs:31:62
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'could not load platform certs: Custom { kind: Other, error: Error { code: -25300, message: "OSStatus -25300" } }', /Users/romanuseinov/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-rustls-0.23.2/src/config.rs:31:62
```

**Expected behaviour**

Forest is able to fetch the params successfully and move on to syncing with the network.

**Environment (please complete the following information):**

- OS: Mac OSX Ventura 13.0
- Rust version(e.g. `rustc --version`): rustc 1.69.0 (84c898d65 2023-04-16)
- Branch/commit: main/a1ca2fe7ed5fbf0dd27661abc4c5aa64bcf87354

**Other information and links**

The problem is that we're building http clients in a loop, which in turn fetches root certificates for each client built. That operation is expensive, so it fails eventually.

lemmih:
For context: Cannot replicate with Mac OSX Big Sur 11.6 or Linux 6.2.6.

I'd love to know more about why it fails. Just because an operation is expensive shouldn't mean that it will eventually fail.

elmattic:
On my side I've managed to reproduce it:
```
thread 'thread 'tokio-runtime-workertokio-runtime-worker' panicked at '' panicked at 'no CA certificates foundno CA certificates found', ', /Users/guillaume/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-rustls-0.23.2/src/config.rs/Users/guillaume/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-rustls-0.23.2/src/config.rs::4848:9
[...]

PR #2968, opened 2023-06-12
[Fix] HttpClient concurrent usage.

ruseinov:
## Summary of changes

- Hyper http client is only built once and reused via shallow clones as it should be.

Fixes https://github.com/ChainSafe/forest/issues/2967

## Change checklist


utils/forest_utils/src/net/http.rs
```diff
@@ -4,17 +4,25 @@
...
+use lazy_static::lazy_static;
 
-/// Constructs [hyper::Client] that supports both `http` and `https`.
-/// Note that only `http1` is supported.
-pub fn https_client() -> hyper::Client<HttpsConnector<HttpConnector>> {
-    hyper::Client::builder().build::<_, Body>(
+lazy_static! {
+    /// A default [hyper::Client]. It's imperative that the builder is only
+    /// called once, because fetching root certificates is expensive.
+    static ref CLIENT: hyper::Client<HttpsConnector<HttpConnector>> =
```
LesnyRumcajs, line 12: Perhaps a `OnceCell` would be a bit cleaner. Not a huge deal, though.

ruseinov: > Perhaps a `OnceCell` would be a bit cleaner. Not a huge deal, though.

Totally, but we are not on the 1.70 toolchain yet as far as I know.

ruseinov: I think perhaps `Lazy` is a better option here, I'll incorporate that after raising a PR to bump the version.

aatifsyed:
I'd like to see our client deprecated and replaced with `reqwest`.
- We don't need to (re)-implement standard logic for redirects and well-known headers - see https://github.com/ChainSafe/forest/blob/5532ab90cd69f3dce313501d29d99c5ae99a0de6/utils/forest_utils/src/net/download.rs#L181, https://github.com/ChainSafe/forest/blob/5532ab90cd69f3dce313501d29d99c5ae99a0de6/utils/forest_utils/src/proofs_api/paramfetch.rs#L195, and https://github.com/ChainSafe/forest/blob/5532ab90cd69f3dce313501d29d99c5ae99a0de6/utils/forest_utils/src/net/download.rs#L170

ruseinov:
> I'd like to see our client deprecated and replaced with `reqwest`.

Totally agreed!

It'd be nice to have an issue for that and do it in a dedicated PR. This PR aims at fixing something that totally breaks node startup on some machines, including mine and is a pretty easy and fast one to review and merge, therefore I'd like to keep it's scope as small as possible.

ruseinov:
@aatifsyed https://github.com/ChainSafe/forest/issues/2977

LesnyRumcajs (approved):

lemmih (approved):

PR #2973, opened 2023-06-12
One crate

aatifsyed:
## Summary of changes
- Move all `forest_*` crates into a root crate, `forest_filecoin`
- Move all `tests/` folders of `forest_*` crates into `forest_filecoin`, EXCEPT `daemon` and `cli` tests, which live in the integration tests directory (`/tests`)
- Fixup imports
- Refactor features
- RPC API build script is removed

Note that some todos are left in the code for #2991 
This is to keep this PR as simple as possible
- Allowing dead code
- Allowing deprecated symbols
- Various clippy lints
- Makefile redundancies
- The api being public

I've left the python scripts in for easy review - I'll remove them after approval


.github/workflows/forest.yml
aatifsyed, line 1: This don't need to be separate jobs anymore, fix in #2991

src/bin/forest.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+fn main() -> anyhow::Result<()> {
+    forest_filecoin::forestd_main()
```
aatifsyed, line 5: rename in #2991

src/blocks/header/mod.rs
```diff
@@ -32,12 +32,12 @@ pub mod json;
 ///
 /// Usage:
 /// ```
-/// use forest_blocks::{BlockHeader, TipsetKeys, Ticket};
-/// use forest_shim::address::Address;
+/// use forest_filecoin::blocks::{BlockHeader, TipsetKeys, Ticket};
```
aatifsyed, line 35: We can't have doc-comments _and_ private internals...

lemmih (approved):
What should we do about doc tests?

ruseinov (approved):
Looks good to me.

nit: I would, perhaps, consider moving all the migration-related python scripts to the `scripts/<one_crate>` folder.
I'm not sure how useful those scripts are going to be in this repo going forward, now that this migration is done, but it's great to keep them in the commit history for historical reasons, it would be a pity to see all this great effort go to waste. Perhaps a separate repo could work as well?

PR #2974, opened 2023-06-12
Prune direct `fvm_shared` `ChainEpoch` usage

### LINKED ISSUE #2962: Prune direct `fvm_shared` usage from Forest
**Issue summary**

Similar to issue #2713, we also need to remove all direct usage of `fvm_shared` and `fvm_shared3`.
- [x] Re-export all `fvm_shared` structs/constants in the Forest shim and use them in Forest. In the places where we use v2 of `fvm_shared`, we should switch to `fvm_shared3` or use a shim.
- [ ] Impose a strict rule of not using `fvm_shared` and `fvm_shared3` directly via `find_banned_deps.rb`. All in all, `fvm_shared` and `fvm_shared3` should be removed from the workspace (only exceptions are `forest_shim` and `forest_interpreter`).

jdjaustin:
## Summary of changes

- Replace direct `fvm_shared` `ChainEpoch` usage with a new `ChainEpoch` implementation in `forest_shim`

Closes associated task in #2962 

## Change checklist


utils/forest_shim/src/clock.rs
```diff
@@ -1,8 +1,86 @@
...
+use std::ops::{Add, Div, Mul, Sub};
+use fvm_shared::clock::ChainEpoch as ChainEpochV2;
 use fvm_shared::clock::EPOCH_DURATION_SECONDS;
+use serde::{Deserialize, Serialize};
 
 pub const SECONDS_IN_DAY: i64 = 86400;
 pub const EPOCHS_IN_DAY: i64 = SECONDS_IN_DAY / EPOCH_DURATION_SECONDS;
+
+#[derive(Clone, Copy, Debug, Default, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub struct ChainEpoch(pub ChainEpochV2);
```
lemmih, line 14: Here you can just do `type ChainEpoch = i64`

lemmih (approved):
Would be nice to test that our `EPOCH_DURATION_SECONDS` matches the one in `fvm_shared2`/`fvm_shared3`.

utils/forest_shim/src/clock.rs
```diff
@@ -1,8 +1,27 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-pub use fvm_shared::clock::ChainEpoch;
-use fvm_shared::clock::EPOCH_DURATION_SECONDS;
-
 pub const SECONDS_IN_DAY: i64 = 86400;
 pub const EPOCHS_IN_DAY: i64 = SECONDS_IN_DAY / EPOCH_DURATION_SECONDS;
+pub const EPOCH_DURATION_SECONDS: i64 = 30;
```
LesnyRumcajs, line 6: We should not use this constant at all because it's valid only for mainnet and calibnet. But that's to be addressed elsewhere.

LesnyRumcajs (approved):

PR #2978, opened 2023-06-13
Revert "Support only compressed snapshots on stable urls"

lemmih:
Reverts ChainSafe/forest#2893

Regressions: Fetched snapshots are always marked as being from Forest even when the Filops vendor is used. Old snapshot files from other vendors are not parsed correctly.


ruseinov (approved):
lgtm

Issue #2955, opened 2023-06-08
Replace progress bars with logging output


lemmih:
**Issue summary**

Progress bars have several drawbacks: They behave poorly together with regular logging, they're only used when connected to a tty, and they're often too verbose.

Let's use the logging infrastructure for progress information:
- [ ] Progress logs should be printed at intervals (say, every 5 seconds).
- [ ] Progress logs shouldn't be verbose. If a process takes less than, say, 5 seconds, don't print anything.
- [ ] Unless the progress bars, progress logs should always be enabled.
- [ ] Support both bytes measures and countable measures.
- [ ] Display elapsed time, remaining time, and speed.

**Other information and links**

lemmih:
Done. The progress indicators can be prettier (and use sensible units) but that can be tracked in another issue.

PR #2979, opened 2023-06-13
feat: fetch arbitrary state-trees

lemmih:
## Summary of changes

- New command: `forest-cli state fetch {CID}`
- The command uses libp2p-bitswap to fetch an IPLD graph in parallel and store it in the database.

Closes

## Change checklist


node/rpc/src/state_api.rs
```diff
@@ -186,3 +190,91 @@ pub(crate) async fn state_wait_msg<DB: Blockstore + Clone + Send + Sync + 'stati
...
+    task_set.spawn(async move { Ok(Ipld::Link(root_cid)) });
+
+    // Iterate until there are no more ipld nodes to traverse
+    while let Some(result) = task_set.join_next().await {
+        match result? {
+            Ok(ipld) => {
+                for new_cid in ipld.iter().filter_map(&mut get_ipld_link) {
+                    counter += 1;
+                    if counter % 1_000 == 0 {
+                        // set RUST_LOG=forest_rpc::state_api=debug to enable these printouts.
```
hanabi1224, line 235: Maybe leave a TODO and refer to https://github.com/ChainSafe/forest/issues/2955 if it's meant to be replaced by the desired progress bar

node/rpc/src/state_api.rs
```diff
@@ -186,3 +190,91 @@ pub(crate) async fn state_wait_msg<DB: Blockstore + Clone + Send + Sync + 'stati
...
+                                let permit = sem.acquire_owned().await?;
+                                let (tx, rx) = flume::bounded(1);
+                                network_send
+                                    .send_async(NetworkMessage::BitswapRequest {
+                                        epoch: 0,
+                                        cid: new_cid,
+                                        response_channel: tx,
+                                    })
+                                    .await?;
+                                // Bitswap requests do not fail. They are just ignored if no-one has
```
hanabi1224, line 258: FYI there is a `send_dont_have` option in `BitswapRequest` to ask peers to respond even if they don't have the requested block, if that could be beneficial here.

node/rpc/src/state_api.rs
```diff
@@ -186,3 +190,91 @@ pub(crate) async fn state_wait_msg<DB: Blockstore + Clone + Send + Sync + 'stati
...
+/// consensus rules.
+pub(crate) async fn state_fetch_root<DB: Blockstore + Clone + Sync + Send + 'static, B: Beacon>(
+    data: Data<RPCState<DB, B>>,
+    Params((CidJson(root_cid),)): Params<StateFetchRootParams>,
+) -> Result<StateFetchRootResult, JsonRpcError> {
+    const MAX_CONCURRENT_REQUESTS: usize = 16;
+    const REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
+
+    let sem = Arc::new(Semaphore::new(MAX_CONCURRENT_REQUESTS));
+    let mut seen: HashSet<Cid> = HashSet::new();
```
hanabi1224, line 216: nit: maybe `CidHashSet`  here if collision is not critical

lemmih: I would use `CidHashSet` if `insert` didn't take an extra callback function. :)

node/rpc/src/state_api.rs
lemmih: Showing a progress bar requires knowing how much work is left. We should show some progress indicators in the client, though. Maybe as a separate PR.

lemmih:
> 1. Can we add some tests for this feature? Both where the Cid we know is present and where we know it won't be there.

I added a test that fetches the state-root of epoch 1. It _should_ always be available but it's not 100% guaranteed to stay available in the future. If it becomes a problem, we can switch to walking the genesis block which we always have available.

> 2. Could this be used instead of downloading actor bundles from Github? Not in this PR, but just a thought. It would give us some resilience from GH outages (but also make us dependent on other peers).

Yes, we should be able to fetch the bundles. I'll leave it for another PR.

LesnyRumcajs (approved):

PR #2994, opened 2023-06-15
fix: use a depth-first search when fetching IPLD graphs

lemmih:
## Summary of changes

- The previous code tried to scan the entire graph and would queue up hundreds of thousands of worker tasks. This meant that the required memory quickly went over 64GiB.
- The new algorithm uses a depth-first search where the search pauses when the max number of worker tasks has been reached. This approach limits the total memory usage to less than 3GiB when walking the current mainnet state.

Closes

## Change checklist


LesnyRumcajs (approved):
:pray:

node/rpc/src/state_api.rs
```diff
@@ -209,72 +211,114 @@ pub(crate) async fn state_fetch_root<DB: Blockstore + Clone + Sync + Send + 'sta
...
-                failures += 1;
-                log::debug!("Request failed: {msg}");
+
+            while let Some(cid) = to_be_fetched.pop() {
+                if task_set.len() == MAX_CONCURRENT_REQUESTS {
+                    if let Some(ret) = task_set.join_next().await {
+                        handle_worker(&mut fetched, &mut failures, ret?)
+                    }
+                }
+                task_set.spawn_blocking({
```
elmattic, line 280: What's the reason for moving to `spawn_blocking` here?

lemmih: Pushing a value to the `dfs` mutex may block.

elmattic: But in that case the bitswap request is blocking as well now, no? Just wondering if it would not be better to have an async task calling `network_send.send_async` then pushing to the `dfs` mutex.

lemmih: Both ways are possible. Probably won't be any measurable difference.

elmattic (approved):

PR #2983, opened 2023-06-14
feat: nv17 state migrations

hanabi1224:
## Summary of changes

Changes introduced in this pull request:

- Port nv17 state migration from Go code
- Port unit tests from Go code
- Regression test for nv17 state migration on calibnet

## Reference issue to close (if applicable)

Closes https://github.com/ChainSafe/forest/issues/2801

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

PR #3000, opened 2023-06-16
feat: Refactor Manifest

hanabi1224:
## Summary of changes

- Refactor `Manifest` to be compatible with actor versions less than 10. (By not requiring actors like `EAM_ACTOR` that are only available with actor version 10 onward)
- Remove `ManifestV2` re-export
- Rename `ManifestV3` to `Manifest` since the actual version number is 1 for both `fvm2::Manifest` and `fvm3::Manifest`
- Add `fn code_by_name` which is needed for retrieving actors from mocked manifests in unit tests, see  https://github.com/ChainSafe/forest/pull/2983/files#diff-c81300115f64e99c31163d90a1a118cc70b0ca332946cbc643f4d75d7d342e98R655

Closes

## Change checklist


lemmih (approved):

src/state_migration/nv18/migration.rs
```diff
@@ -34,28 +35,24 @@ impl<BS: Blockstore + Clone + Send + Sync> StateMigration<BS> {
...
-            .get_cbor(new_manifest)?
-            .ok_or_else(|| anyhow!("new manifest not found"))?;
-        let new_manifest = ManifestNew::load(&store, &new_manifest_data, version)?;
+        let new_manifest = Manifest::load(&store, new_manifest)?;
 
-        current_manifest.builtin_actor_codes().for_each(|code| {
-            let id = current_manifest.id_by_code(code);
-            let new_code = new_manifest.code_by_id(id).unwrap();
+        current_manifest.builtin_actors().for_each(|(name, code)| {
+            let new_code = new_manifest.code_by_name(name).unwrap();
```
LesnyRumcajs, line 44: Are you sure we want to panic in such a case? If so, let's at least provide a helpful message with `expect`.

LesnyRumcajs (approved):

src/state_migration/nv18/migration.rs
hanabi1224: The original logic uses unwrap and the error message in code_by_name should be clear enough

hanabi1224:
```rust
pub fn code_by_name(&self, name: &str) -> anyhow::Result<&Cid> {
        self.by_name
            .get(name)
            .ok_or_else(|| anyhow!("Failed to retrieve actor code by name {name}"))
    }
```

PR #3003, opened 2023-06-16
[Enhancement] Introduce a dry-run param for fetch-params for easier testing.

### LINKED ISSUE #2956: Avoid network access when testing `param-fetch`
**Issue summary**

Two tests in Forest verify the download location for proof-parameter files. Unfortunately, they work by actually downloading the proof files, which can be slow. An alternative would be to implement a `--dry-run` flag that prints out where the files would end up if they were downloaded.

Acceptance criteria:
- [x] The tests still verify that the download folder is correctly computed.
- [x] The tests do not require internet access.

Slow param-fetch tests:
- `test_download_location_of_proof_parameter_files_default`
- `test_download_location_of_proof_parameter_files_env`

ruseinov:
## Summary of changes

- Introduced a `dry-run` parameter for `forest-cli fetch-params`.
- Modified tests to read the output instead of checking the directory.

Closes #2956

## Change checklist


lemmih (approved):
This PR makes `cargo nextest run` several times faster for me.

aatifsyed:
Mixed feelings about this one - adding product options _for testing_ kind of defeats the principles of integration testing. That said, it's not like we're particularly principled already.

jdjaustin (approved):

ruseinov:
> Mixed feelings about this one - adding product options _for testing_ kind of defeats the principles of integration testing. That said, it's not like we're particularly principled already.

Won’t disagree, though a dry-run is a nice option. I guess I’d rather unit-test the private function that decides the path, but it’s also important to verify this.

ruseinov:
A good way to fix it is to convert it all to structs and use a mock for testing.

LesnyRumcajs:
> Won’t disagree, though a dry-run is a nice option. I guess I’d rather unit-test the private function that decides the path, but it’s also important to verify this.
> A good way to fix it is to convert it all to structs and use a mock for testing.

@ruseinov How about we create an issue with this? I do agree with @aatifsyed; having a test that checks that the test flag works won't catch bugs in actual logic.

Alternatively, to speed up things without a `dry-run` flag, we could add a `--force-recreate` (or whatever else) flag that would download the proof params if and only if they are not already downloaded. It should be as fast as a dry-run on a local machine (which I believe is the point of introducing it first) and catch regressions in the CI, given the environment there is always clean.

lemmih:
The tests target newly created temporary directories - skipping already-downloaded files won't be any faster as the test directories are always empty.

LesnyRumcajs:
They could be if they are cached.

ruseinov:
As far as I'm concerned the tests in question were originally intended for the download path calculation, which is what they are doing now as well, verifying exactly the same function. At least that's what one can gather from the test names. 
However, I do agree that we perhaps need a separate set of tests that would verify this thing e2e.

Definitely not a fan of having cached files for testing, that will also potentially give way to other kinds of regressions and the first run will always be slow. 

The one thing we could do is have extra tests that do the actual downloads and have them enabled via some sort of flag.  I would investigate how it's done in other projects/libraries to see if we can learn from other people's mistakes.

PR #3010, opened 2023-06-19
chore: avoid the risk of collisions in `CidHashSet`

lemmih:
## Summary of changes

Forest is dealing with such a small number of CIDs (tens of millions) that they could be stored as 64 bit numbers with only 1 expected collision every 1000 years or so. But let's not risk it. Storing the full hash is only a few bytes more.

- Store the full 32-byte hash for the most commonly used CID variant.

All benchmarks were run on storage-optimized droplets with 16GiB of RAM and 300 GiB of SSD. The average of at least three runs were used:

|   | Time (seconds)  | Peak RSS (GiB)  |
|---|---|---|
| Export snapshot (baseline) |  4978 (x1.00) | 4.81 (x1.00) |
| Export snapshot (BTreeSet)  | 5500 (x1.10)  | 6.69 (x1.39)  |
| Export snapshot (HashSet)  | 5331 (x1.07)  |  6.07 (x1.26) |

Closes

## Change checklist


lemmih:
> How abt making fallback a variant of CidVariant?

I meant for the `CidVariant` enum to be the set of known CID types that we want to treat differently. This way, it always consists of a tag and an array of bytes, and it is smaller in size than a `Cid`.

LesnyRumcajs:
This seems like a significant drop in performance. Are you positive we are fine with this tradeoff? What would be the result of a collision, in practice?

lemmih:
> This seems like a significant drop in performance. Are you positive we are fine with this tradeoff? What would be the result of a collision, in practice?

A collision during snapshot exporting would corrupt the snapshot. A collision during garbage collection would crash Forest. The consequences are pretty drastic. We've only gotten away with it because collisions are so rare.

+25% peak memory usage is rough but still significantly lower than the memory required to import a snapshot (8-9 GiB).

src/ipld/cid_hashset.rs
```diff
@@ -2,27 +2,84 @@
...
+
+// Nearly all Filecoin CIDs are V1, DagCbor encoded, and hashed with Blake2b265. If other types of
+// CID become popular, they can be added to the CidVariant structure.
+enum CidVariant {
+    V1DagCborBlake2b([u8; 32]),
+}
+
+impl TryFrom<Cid> for CidVariant {
+    type Error = ();
+    fn try_from(cid: Cid) -> Result<Self, Self::Error> {
```
LesnyRumcajs, line 19: Could we have a unit test for this?

src/ipld/cid_hashset.rs
```diff
@@ -2,27 +2,84 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use ahash::HashSet;
-use cid::Cid;
+use cid::{Cid, Version};
+use fvm_ipld_encoding::DAG_CBOR;
+use std::collections::BTreeSet;
 
+const BLAKE2B256: u64 = 0xb220;
```
LesnyRumcajs, line 9: Can we import this constant from somewhere?

src/ipld/cid_hashset.rs
```diff
@@ -2,27 +2,84 @@
...
+
+// Nearly all Filecoin CIDs are V1, DagCbor encoded, and hashed with Blake2b265. If other types of
+// CID become popular, they can be added to the CidVariant structure.
+enum CidVariant {
+    V1DagCborBlake2b([u8; 32]),
```
LesnyRumcajs, line 14: `32` could be a named constant given that it appears a few times. Alternatively, the entire `[u8; 32]` could be aliased.

LesnyRumcajs (approved):
> > This seems like a significant drop in performance. Are you positive we are fine with this tradeoff? What would be the result of a collision, in practice?
> 
> A collision during snapshot exporting would corrupt the snapshot. A collision during garbage collection would crash Forest. The consequences are pretty drastic. We've only gotten away with it because collisions are so rare.
> 
> +25% peak memory usage is rough but still significantly lower than the memory required to import a snapshot (8-9 GiB).

The consequences are rather convincing.

src/ipld/cid_hashset.rs
lemmih: It appears to be hidden.

src/ipld/cid_hashset.rs
lemmih: Hoisted out the size of the blake2b hash.

PR #3011, opened 2023-06-19
fix: remove #[allow(unused, clippy::upper_case_acronyms, clippy::enum_variant_names)]

aatifsyed:
## Summary of changes
- ~don't have modules like `daemon::daemon` - either rename the child, or fold it into the parent~ Broken out to #3037 
- Remove stuff that's never called
- Make stuff that's only called in tests `#[cfg(test)]`
- Fixup clippy naming stuff
- libp2p changes are in #3038 

Note that merging this is blocked on other MRs
- #3026 so that the state diff code isn't dead
- #3012 so that all our modules aren't public


lemmih (approved):
@hanabi1224 Any insight into the libp2p changes?

hanabi1224:
@lemmih I agree that the change is simple and straightforward, however the root cause of last regression is still unknown, I would suggest making the libp2p change in a small PR that is easy to revert in case any regression happens.  Does that make sense?

src/beacon/mock_beacon.rs
```diff
@@ -11,23 +9,18 @@ use byteorder::{BigEndian, ByteOrder};
 use crate::beacon::{Beacon, BeaconEntry};
 
 /// Mock beacon used for testing. Deterministic based on an interval.
-pub struct MockBeacon {
-    interval: Duration,
-}
+pub struct MockBeacon {}
```
LesnyRumcajs, line 12: Let's just derive `Default`, if possible.

src/shim/machine/manifest.rs
```diff
@@ -3,32 +3,42 @@
...
-pub const SYSTEM_ACTOR_NAME: &str = "system";
-pub const VERIFREG_ACTOR_NAME: &str = "verifiedregistry";
-// actor version >= 9
-pub const DATACAP_ACTOR_NAME: &str = "datacap";
-// actor version >= 10
-pub const EVM_ACTOR_NAME: &str = "evm";
-pub const EAM_ACTOR_NAME: &str = "eam";
-pub const PLACEHOLDER_ACTOR_NAME: &str = "placeholder";
-pub const ETH_ACCOUNT_ACTOR_NAME: &str = "ethaccount";
+macro_rules! name {
```
LesnyRumcajs, line 14: This should most likely be replaced by, e.g., https://github.com/ChainSafe/fil-actor-states/blob/main/fil_actors_shared/src/v11/runtime/builtins.rs

It can be a separate PR, though.

LesnyRumcajs (approved):

src/shim/machine/manifest.rs
aatifsyed: #3076

Issue #2928, opened 2023-06-05
Remove home-rolled `prompt_confirm` function


aatifsyed:
- [x] remove the `atty` crate in favour of [`IsTerminal`](https://doc.rust-lang.org/std/io/trait.IsTerminal.html)
- [x] standardize on the indicatif/dialoguer UI crates
  - [ ] remove home-rolled `prompt_confirm` function
    - > Also, as far as I remember, we re-implemented it here for this reason: https://github.com/console-rs/dialoguer/issues/170
      > But it would be nice not to use such workarounds.

      _Originally posted by @LesnyRumcajs in https://github.com/ChainSafe/forest/pull/2893#discussion_r1217710558_
  - [x] rationalize our progress bars and color story
- [ ] ~~log to stderr, not stdout~~ The logs are the primary out of `forestd` and should be sent to stdout. The other tools (`forest-wallet`, `forest-tool`, `forest-cli`) already log to stderr.

aatifsyed:
It's likely that a lot of our `atty` checks have subtle/edge bugs. E.g https://github.com/ChainSafe/forest/blob/1af4fde3cb1ef3e24a3ff9d7082d171d4c8b1892/forest/daemon/src/daemon.rs#L505
Checks that stdin is a tty, but the confirm is actually rendered on stdout

PR #3022, opened 2023-06-20
[Fix] Progress-bar: use zstd uncompressed size when available

### LINKED ISSUE #3005: Incorrect progress bar when importing compressed snapshots
**Describe the bug**

**To Reproduce** Steps to reproduce the behavior:

1. Remove any `mainnet` data: `forest-cli db clean`
2. Import a snapshot: `forest --encrypt-keystore false --halt-after-import --auto-download-snapshot`
3. Notice how `forest` appears to freeze after the progress bar hits 100%.

(Note, the same can be observed using a calibnet snapshot, but the effect is less noticeable.)

**Expected behaviour**

The progress bar is configured such that 100% is the size of the compressed snapshot but, when running, it counts the **decompressed** bytes. Since our compression factor is roughly 2x, when the progress bar hits 100%, only about half of the compressed snapshot has been consumed.

The progress bar should show how much of the snapshot file we've consumed.

- OS: Linux
- Rust version(e.g. `rustc --version`): rustc 1.70.0 (90c541806 2023-05-31)
- Branch/commit: `main`/3e06cd98709ff2ef6ec11d9e943482acc686edf8

ruseinov:
## Summary of changes

- Refactoring of the current approach to streaming files and progress tracking.
- Getting rid of unnecessary abstractions.
- Making sure that the progress is tracked correctly for the archived files, based on the archive size.

Closes #3005 

To make this testable it'd require quite some refactoring, perhaps it's worth considering going forward. 
For this to be useful we'll need to complete https://github.com/ChainSafe/forest/issues/3023 as well.

## Change checklist

- [x] Refactor this to have `Box`'d readers instead of `Either`


ruseinov:
This totally works, the only pitfall is that we don't store uncompressed size in forest when exporting the snapshot. This would require another issue.

`zstd whatever` and further `forest --encrypt-keystore false --halt-after-import --import-snapshot=whatever.zst --chain calibnet` is enough to test the fact that the size is being read correctly, otherwise we fall back to the archive size.

hanabi1224 (approved):

lemmih:
We can import snapshots in two ways: From a file or a URL. When downloading from a URL, the progress indicator shows how much of the compressed archive has been consumed. If we change the logic when importing from a file, we must also change it for URLs.

But all things considered, I don't see why the progress bar should measure uncompressed bytes. The user-visible files are always compressed, so measuring progress in compressed bytes seems more intuitive.

lemmih (changes requested):
We need to measure the same thing regardless of whether we import from a file or a URL. Compressed bytes or uncompressed bytes is up for debate, but we have to be consistent.

ruseinov:
> We can import snapshots in two ways: From a file or a URL. When downloading from a URL, the progress indicator shows how much of the compressed archive has been consumed. If we change the logic when importing from a file, we must also change it for URLs.
> 
> But all things considered, I don't see why the progress bar should measure uncompressed bytes. The user-visible files are always compressed, so measuring progress in compressed bytes seems more intuitive.

The issue here is that we are using BufReader 

> We need to measure the same thing regardless of whether we import from a file or a URL. Compressed bytes or uncompressed bytes is up for debate, but we have to be consistent.

You are right!

Yep, I've just figured out that the url import does not do a download to the directory. It's somewhat confusing as we have yet another option that allows us to download a snap if required, and that actually downloads first and then imports second.

Perhaps we should revisit the way these are currently done to have more logic in common between all the import methods.

lemmih:
Yeah, we basically have three ways of importing a snapshot:
- Streaming import from a file.
- Streaming import from a URL.
- Downloading a snapshot to disk and then importing the file. This option is faster than a streaming download most of the time.

src/utils/net/download.rs
```diff
@@ -54,149 +53,125 @@ impl<R: AsyncRead + Unpin> AsyncRead for FetchProgress<R> {
...
+    /// Auto-detects whether or not the file has to be streamed from a remote location, and
+    /// whether or not it's zstd compressed.
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - Snapshot location, could be either remote or local
+    pub async fn read(
+        path: &str,
+    ) -> anyhow::Result<Either<ZstdDecoder<FetchProgress>, FetchProgress>> {
+        let is_remote_file: bool = path.starts_with("http://") || path.starts_with("https://");
```
lemmih, line 85: Would this work?
`let is_remote_file = Url::parse(path).is_ok()`

ruseinov: Yeah, I think it'd be better. I kept this in order to be consistent with the initial implementation, but there's no reason why we couldn't accept more protocols.

ruseinov:
There is an alternative solution to this, that could be done via Box'ing readers. That shouldn't impact the performance much, if at all, but then we could get rid of `Either` and all that nesting.

lemmih:
> There is an alternative solution to this, that could be done via Box'ing readers. That shouldn't impact the performance much, if at all, but then we could get rid of `Either` and all that nesting.

I like that idea.

ruseinov:
> > There is an alternative solution to this, that could be done via Box'ing readers. That shouldn't impact the performance much, if at all, but then we could get rid of `Either` and all that nesting.
> 
> I like that idea.

I will implement that. I wanted to refactor this first using more or less the same abstractions, I think it's way more DRY, but still unnecessary complicated and a bit ugly.

aatifsyed (dismissed):
I don't think we should write the code this way - feel free to make broader changes to make the code simpler and more idiomatic.

Specifically, I'd like you to consider the following changes
- Use standard APIs like `Stream` and `AsyncRead`/`AsyncWrite` where possible
- Take advantage of the flexibility that they offer to write more composable business logic
  - Code that is agnostic over files and URLs
  - Code that is agnostic over compression
- Do not manually implement `Async*` traits, they're normally a code smell. Here, we've gone for wrapping rather than composing in `FileReader`. Just remove this module. Use or write flexible UI APIs like `indicatif::ProgressBar::wrap_async_stream` instead.
- Use tools from the ecosystem instead of e.g writing our own `Either` (you were right that this smelled bad, but instead of using `Box<dyn>` you can just use someone else's :)

I've added an example in https://github.com/ChainSafe/forest/compare/aatifsyed/demonstrate-streams - it ends up being about 100 lines of code, handles compressed/uncompressed and url/file perfectly. Check it out if you need hints, want a co-author :)

src/utils/misc/either.rs
```diff
@@ -16,6 +16,22 @@ pub enum Either<L, R> {
...
+    }
+
+    fn consume(self: Pin<&mut Self>, amt: usize) {
+        match Pin::into_inner(self) {
+            Self::Left(left) => Pin::new(left).consume(amt),
+            Self::Right(right) => Pin::new(right).consume(amt),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
ruseinov: As discussed before - I aimed at a small refactor keeping the abstractions, but that was not satisfactory. Totally on the same page here, but I we don’t need Either at all.

src/utils/net/download.rs
```diff
@@ -9,38 +9,37 @@ use std::{
...
-    /// Holds a Reader, tracks read progress and draw a progress bar.
-    pub struct FetchProgress<R> {
+    /// Holds a Reader, tracks read progress and draws a progress bar.
+    pub struct FetchProgress {
         #[pin]
-        pub inner: R,
+        pub inner: Either<DownloadStream, BufReader<async_fs::File>>,
         pub progress_bar: ProgressBar,
     }
 }
```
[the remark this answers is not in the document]
ruseinov: Yep, that’s pretty ugly

ruseinov:
@aatifsyed the review looks good, thanks for this! the idea was to refactor it a bit keeping most of the abstractions and fix the issue, then follow up. But it is indeed so ugly - the rewrite will happen now instead. Never postpone anything for tomorrow :)

I have a slightly different plan of attack than you suggested, will have to evaluate what is cleaner/better - and proceed.

ruseinov:
I've taken a look at the streams solution @aatifsyed posted and it looks good. It could be further enhanced by Box`ing the readers/streams to avoid the use of `Either` altogether, but it remains to be seen wither or not this is beneficial.
@lemmih I'm wondering what is the reasoning to have several different ProgressBar libraries within the project. I guess we could clean it up too? 

I guess the path would be as follows: migrate everything that's currently using `download.rs` to the new paradigm. And then create a follow-up issue to get rid of all sorts of ProgressBar/file streaming usages scattered around the codebase in favour of one approach and one library.

aatifsyed:
If you prefer, but avoiding `dyn` in this case is basically free - we already have it, and it saves us on vtable indirection. 

```
Left(foo)
Box::new(foo) as Box<fun AsyncRead>
```

I know which I prefer

aatifsyed:
(I'm the reason we have several progress bar libraries - see ISSUE)

we should just remove them all

lemmih:
> I've taken a look at the streams solution @aatifsyed posted and it looks good. It could be further enhanced by `Box`ing the readers/streams to avoid the use of `Either` altogether, but it remains to be seen wither or not this is beneficial. @lemmih I'm wondering what is the reasoning to have several different ProgressBar libraries within the project. I guess we could clean it up too?
> 
> I guess the path would be as follows: migrate everything that's currently using `download.rs` to the new paradigm. And then create a follow-up issue to get rid of all sorts of ProgressBar/file streaming usages scattered around the codebase in favour of one approach and one library.

I'd like to move away from progress bars that use terminal codes and replace them with regular logging. Work on this is going on here: https://github.com/ChainSafe/forest/issues/2955. Ideally the new log-based progress indicators would be a drop-in replacement.

ruseinov:
> If you prefer, but avoiding `dyn` in this case is basically free - we already have it, and it saves us on vtable indirection.
> 
> ```
> Left(foo)
> Box::new(foo) as Box<fun AsyncRead>
> ```
> 
> I know which I prefer

I'm going to need to implement the actual code to see what's cleaner in the context. What I like about not having to rely on `Either` are cleaner interfaces that don't have 3rd-party library return types. However it remains to be seen what looks/works better. I'd say that performance gain here is negligible, I'd choose an easier-to-use and cleaner interface over a tiny performance gain every day of the week. 
That also goes for dependencies, but we already have tokio, so that's a bit less of concern, unless we decide to split the project into different crates again at some point in the distant future.

ruseinov:
> (I'm the reason we have several progress bar libraries - see #2928)
> 
> we should just remove them all, we're not UI designers.

I like this idea a lot!

ruseinov:
@aatifsyed I have basically utilised your suggested technique here with some modifications.
I have initially opted for fixing some of the below issues in this same PR, but now that I look at it - it's best to split this into several focused PRs.

1. Not a big fan of returning `Either`, tokio or not. It feels better to return async `futures` primitives as this will eventually make it into `std`. It's best to avoid depending on specific implementations in public interfaces.
2. I have moved the progress bar style into the new `download.rs` for now, as it's not used anywhere else at the moment. If we decide to remove progress bars completely - we probably won't need to revisit this at all. Perhaps could be dealt with as part of https://github.com/ChainSafe/forest/issues/2928
3. Now that we have enabled both `tokio` and `futures-io` features for `async-compression` - I think  would be good to have a follow-up PR that opts just for `tokio` feature and fixes all the errors that are popping up in `chain/store/chain_store.rs`. 
4. Follow-up PR that completely removes our local implementation of `Either` from the repo.

src/utils/net/download.rs
```diff
@@ -1,202 +1,91 @@
...
-        let inner = ZstdDecoder::new(BufReader::new(file));
 
-        Ok(FetchProgress {
-            progress_bar: pb,
-            inner,
-        })
+    // This method checks the header in order to see whether or not we are operating on a zstd
+    // archive. The zstd header has a maximum size of 18 bytes:
+    // https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#zstandard-frames.
+    async fn is_zstd(buf: &[u8]) -> anyhow::Result<bool> {
```
lemmih, line 66: This function looks to be neither async nor fallible.

```suggestion
    fn is_zstd(buf: &[u8]) -> bool {
```

src/utils/net/download.rs
```diff
@@ -1,202 +1,91 @@
...
+        Ok(matches!(zstd_safe::get_frame_content_size(buf), Ok(_)))
```
lemmih, line 67: Alternative:
```suggestion
        zstd_safe::get_frame_content_size(buf).is_ok()
```

lemmih (approved):
Looks good. The PR description should be updated to reflect the new changes.

src/utils/net/download.rs
ruseinov: Yes, I've made changes to it, but forgot to change the signature. Good catch, thanks!

ruseinov:
> Looks good. The PR description should be updated to reflect the new changes.

done.

src/utils/net/download.rs
```diff
@@ -1,202 +1,91 @@
...
-        let r = Pin::new(&mut self.inner).poll_read(cx, buf);
-        if let Poll::Ready(Ok(size)) = r {
-            self.progress_bar.add(size as u64);
+pub struct StreamedContentReader {}
+
+impl StreamedContentReader {
+    /// This method is parsing a file-path/URL passed to it and attempts to open a stream.
+    /// Additionally, it detects whether or not the resulting stream is a zstd archive and treats
+    /// it accordingly.
+    pub async fn read(path: &str) -> anyhow::Result<Box<dyn futures::AsyncRead + Send + Unpin>> {
```
aatifsyed, line 22: no blocking: this should just be a function

aatifsyed (approved):

src/utils/net/download.rs
ruseinov: This has to return a `Future` one way or another, not entirely sure what this comment means. Care to elaborate?

ruseinov: I think auto-merge took care of this already, but if you have any ideas for changes here - a follow-up is not going to be a problem.

ruseinov: I guess you are referring to returning a `Box`d Future here.

PR #3028, opened 2023-06-20
fix: turn off incremental builds

aatifsyed:
## Summary of changes
- Multiple team members are hitting ICEs due to incremental builds
- This disables that for our repo
- Note you can still enable this on your machine if it's important for you, by adding config in a parent directory: https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure


.github/workflows/forest.yml
ruseinov, line 1: curious: Was it also a problem in CI?

lemmih: Nope. The CI already disables incremental builds.

lemmih (approved):

LesnyRumcajs (commented):
Is it actually disabled? To my understanding, if you omit this entry, the build will default to the profile value, which is `false` only for release builds.

What is the compilation time impact of this change?

elmattic (approved):

elmattic:
> Is it actually disabled? To my understanding, if you omit this entry, the build will default to the profile value, which is `false` only for release builds. That is to say, incremental compilation will still be there for the dev builds.

We can disable it with:
```
[profile.dev]
incremental = false
```

That said I would prefer to leave this choice to the developer. Safe goes for the compilation backend in debug (i.e. cranelift etc).

ruseinov:
> That said I would prefer to leave this choice to the developer. Same goes for the compilation backend in debug (i.e. cranelift etc).

I tend to agree. Why don't we leave the flag be and just make it `false` by default? It's way easier to work with than having to re-introduce the whole section.

aatifsyed:
> One other solution to all these problems is to remove config.toml from the repository altogether and git ignore .cargo dir. That would be more idiomatic.

I like this

aatifsyed:
https://github.com/ChainSafe/forest/issues/3146

PR #3038, opened 2023-06-21
fix: remove #[allow(deprecated)]

aatifsyed:
Broken out of #3011 to be easy to revert in case the networking is broken


lemmih (approved):
Fingers crossed. :)

ruseinov (approved):

PR #3042, opened 2023-06-21
[Fix] Cargo.lock package version

ruseinov:
Fixes the `release_checklist.md` and includes `Cargo.lock` package version update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2759,7 +2759,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "forest-filecoin"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


aatifsyed (changes requested):
I don't think this is the right approach - change CI to fail by using `cargo build --locked`

ruseinov:
> I don't think this is the right approach - change CI to fail by using `cargo build --locked`

We are already doing `cargo install --locked` and that obviously does not help.

aatifsyed:
I think we want this in CI:
```
cargo generate-lockfile --locked
```

ruseinov:
> cargo generate-lockfile --locked

We definitely don't want that here, it fails with the latest `Cargo.lock`. Let's fix it properly outside of the scope of this PR.

lemmih (approved):

aatifsyed (approved):

PR #3045, opened 2023-06-21
fix dockerhub rate limiting

LesnyRumcajs:
## Summary of changes

- mirror the https://github.com/tonistiigi/xx to GHCR which should not count towards the limit,
- use the same image for the builder as for the end image (this should lower the number of layers we download),

Closes

## Change checklist


LesnyRumcajs:
@lemmih As an alternative (for using the same image for the builder and the release), we could get nice images, like [rust](https://gallery.ecr.aws/docker/library/rust), with the public AWS ECR - 500G per month anonymous (it may be too little for the CI) or 5T free tier. The usage is [straightforward](https://github.com/docker/login-action#aws-public-elastic-container-registry-ecr). This would get rid of our rate-limiting issues for good.

Pricing is [here](https://aws.amazon.com/ecr/pricing/), in our case under the *Data transferred from public repositories*.

elmattic (approved):

aatifsyed (approved):

PR #3047, opened 2023-06-21
fix: remove deleg_cns

aatifsyed:
## Summary of changes

It's dead code :shrug:


aakoshh:
@LesnyRumcajs thank you very much for asking :pray: , I reckon you can remove it.  cc @adlrocha

I was going to follow up with https://github.com/aakoshh/forest/pull/6 but we ran into such an impedance mismatch that I gave up on trying to add different consensus types to Forest (or Lotus, about which there is an ongoing conversation with team).

LesnyRumcajs:
@aakoshh Thanks for the swift reply! Please let us know if you ever need this back. :)

lemmih (approved):
Looks good. We can probably also get rid of `RewardCalc` but that can be done in a separate PR.

LesnyRumcajs (approved):
It might be worth adding an entry in the changelog for completeness.

PR #3048, opened 2023-06-21
fix: remove rocksdb

aatifsyed:
One db to rule them all


LesnyRumcajs:
Should this comment be removed? https://github.com/ChainSafe/forest/blob/aatifsyed/no-rocksdb/src/chain_sync/tipset_syncer.rs#L1200-L1205

Should this section be removed? https://github.com/chainsafe/forest/blob/aatifsyed/no-rocksdb/documentation/src/trouble_shooting.md#file-descriptor-limits

Also, the changelog needs to be updated with a breaking change.

lemmih (changes requested):
Looks good. Let's remove the last vestiges of rocksdb that Hubert pointed out.

elmattic:
`Flush` method of `Store` trait can be removed as well.

lemmih (approved):

PR #3049, opened 2023-06-21
feat: main functions accept arguments

aatifsyed:
## Summary of changes

This is useful for benchmarking invocations using `criterion` (introduced in a later PR)


ruseinov:
Not related to this PR, somehow having one-liners in `src/bin` feels wrong. I am assuming these are leftovers from before everything was one crate. 
Can't help but wonder if we should move `forest_main` and `forestd_main` logic into the actual `bin/blah.rs` files.

lemmih:
> Not related to this PR, somehow having one-liners in `src/bin` feels wrong. I am assuming these are leftovers from before everything was one crate. Can't help but wonder if we should move `forest_main` and `forestd_main` logic into the actual `bin/blah.rs` files.
> 
> EDIT: If this sounds reasonable - perhaps that could be done as part of this PR, then you won't need to do the argument plumbing at all.

We don't have an easy way of doing this. It would require redesigning a significant chunk of our code.

lemmih (approved):

sudo-shashank (approved):

ruseinov (approved):

PR #3055, opened 2023-06-22
fix: run yarn install

aatifsyed:
I don't know anything about javascript, but I can't run prettier properly on our codebase without dirtying our lockfile.
This is the result of running `yarn install`, with some fixing up of stuff that didn't look right to commit.


aatifsyed:
>  In general, why is it a big deal? 
I can't just run prettier, if I try and run in from my directory, I have to
```
```

LesnyRumcajs:
@aatifsyed For the moment, you could run `docker run --rm -v $(pwd):/work tmknom/prettier --write **/*.md`. I know it is not ideal, but hopefully, we won't have to resort to such workarounds in the future.

ruseinov:
Interesting, I don't have such issues. But I run commands through yarn run $whatever_package_json_target . Perhaps that fixes it for you @aatifsyed ?

hanabi1224:
What's the version of your local `yarn`?  Note that forest uses yarn v2 which is incompatible with yarn v3

PR #3057, opened 2023-06-23
chore: use upstream arbitrary implementation for ActorState

lemmih:
## Summary of changes

- Replace our custom Arbitrary implementation with the upstream version.

Part of #2143

## Change checklist


Cargo.toml
```diff
@@ -58,7 +58,7 @@ fs_extra = "1.2"
 futures = "0.3"
 futures-util = "0.3"
 fvm = { version = "~2.4", default-features = false }
-fvm3 = { package = "fvm", default-features = false, version = "~3.4" }
+fvm3 = { package = "fvm", default-features = false, version = "~3.4", features = ["arb"] }
```
LesnyRumcajs, line 61: Could we potentially use the `arb` feature only in `dev-dependencies` like here? https://github.com/filecoin-project/ref-fvm/blob/master/shared/Cargo.toml#L45

lemmih: Ah, right. Now that we're a single crate, we don't have to depend on `quickcheck` outside of our dev-dependencies. I'll punt that change to another PR, okay?

LesnyRumcajs: Of course.

LesnyRumcajs (approved):

elmattic (approved):

Issue #3034, opened 2023-06-21
Add Forest Devnet setup to the CI


LesnyRumcajs:
**Issue summary**
To ensure there are no regressions, we should add the devnet setup to the CI. This may require some *tricks* given that the initial setup may be demanding for a regular GH worker.

Some bottlenecks and potential workarounds:
- Param fetching seems slow even on a decent machine; we could store them on DO and download them manually,
- Forest build is slow; we could re-use the Docker image created in the docker workflow. Alternatively, we could grab the binary and put it into a brand new Docker image - we already store those in integration tests.
- Compiling Lotus may be slow/wasteful - we could upload the specific Lotus 1.23 2k image to GHCR.

After the devnet is set up, we could consider the following checks:
- [ ] Forest is syncing (we may re-use the health check script from the `sync check`,
- [ ] Check some basic operations, like listing peers, checking the sync status or sending FIL.

**Other information and links**

PR #3062, opened 2023-06-23
Run devnet in CI

LesnyRumcajs:
## Summary of changes

- Forest binaries are re-used and used in building the Forest CI image. This adds minimal overhead.
- pre-built the Lotus image to https://github.com/orgs/ChainSafe/packages/container/package/lotus-devnet. Building it in the CI adds 15 minutes to the job duration, which is not worth it for the sake of having everything built in the CI at every commit (or even push to `main`). I moved it into a triggerable workflow.
- check - check if Forest can get past the genesis block.

We could think of manually grabbing param files, but at the moment, it seems reasonable enough.
image

Closes https://github.com/ChainSafe/forest/issues/3034

## Change checklist


lemmih (approved):

hanabi1224 (approved):
It will be great if we could run devnet check in `gen_coverage_report.sh` to increase codecov of forest networking layer

LesnyRumcajs:
> It will be great if we could run devnet check in `gen_coverage_report.sh` to increase codecov of forest networking layer

It'd require quite a bit of tinkering for no big gain - I believe we'd only additionally catch the additional devnet parameters setting.

Issue #3074, opened 2023-06-26
Database overlay backed by CAR file


lemmih:
**Issue summary**

A CAR file is an unordered stream of `(CID, Ipld)` pairs representing a (possibly incomplete) DAG. We often have to query data in a CAR file, and we currently do it by loading each key-value pair into a database. This is relatively slow, though, and it temporarily doubles the required storage space.

It should be possible to memory map a CAR file, scan through each key-value pair, build a mapping from key to position in the mapped file, and query it directly without using a database. Keeping the entire index in memory is feasible since each key is only 4 bytes, and there are roughly 55 million pairs in a mainnet snapshot file.

Doing the same with a zstd compressed CAR file is more work but still doable.

Use cases:

- `forest-cli snapshot validate`. Could validate the integrity without requiring any extra storage space.
- `forest --import-snapshot ...`. Could connect to a network without a lengthy DB import.
- In the future, queries on historical data could be completed much quicker if CAR files were directly used.

**Other information and links**

lemmih:
See [`tokio_util::codec`](https://docs.rs/tokio-util/latest/tokio_util/codec/index.html) for frame reading infrastructure. It should be possible to decode a stream of bytes (compressed CAR file) into a stream of key-value pairs annotated with the offset of the zstd frame and the byte offset of the value inside the frame. This would allow building an index for directly querying a compressed CAR file.

aatifsyed:
Frames tend to be large (archives are often one frame), and must be decompressed linearly, so I don't think we should support compressed archives for this feature

lemmih:
> Zstd frames tend to be large (archives are often one frame), and must be decompressed linearly, so I don't think we should support compressed archives for this feature, at least not for MVP.
> 
> That is, seeking to offset 1024 involves seeking to 0, decompressing til 1024, then continuing. We'll absolutely demolish the disk.

Forest generates the compressed CAR files and we can set the frame count to suit our needs. That said, starting without support for compressed archives would still be a great first step.

> I can do some initial investigation into this this week

Sounds great!

Issue #2676, opened 2023-03-16
Simplify trait constraints on DB


creativcoder:
**Issue summary**

Currently we have:  `DB: Blockstore + Store + Clone + Send + Sync + 'static,` in impl blocks in many places in forest codebase. We have a PR to implement `Blockstore for Arc<BS>`: https://github.com/filecoin-project/ref-fvm/pull/1715 upstream.

Once that is merged we can use the new release of fvm_ipld_blockstore to simplify trait constraints on DB.

**Other information and links**

hanabi1224:
@lemmih @creativcoder After https://github.com/ChainSafe/forest/pull/3328 being merged, could we close this one?  I assume `Send + Sync` cannot be simplified in most places

hanabi1224:
Yes

lemmih:
Completed by @hanabi1224

PR #3085, opened 2023-06-27
feat: CAR-backed Blockstore

aatifsyed:
## Summary of changes

Work for https://github.com/ChainSafe/forest/issues/3074.

I've not wrapped a `Blockstore`, opting to keep a `write_cache` instead for simplicity.
Is this appropriate for the intended usecase?

If not, I will probably create an inner `ParityDb`, backed by an anonymous file/directory, as that will be more appropriate for this usecase (users shouldn't have to configure a database, or pick a directory, it's conceptually private to the `CarBackedBlockstore`.

## Change checklist

  - N/A


src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,272 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
```
ruseinov, line 2: nit: needs an empty line.

src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,272 @@
...
+//! ││├──────────┤││ |
+//! │││data      │││ |
+//! ├┴┴──────────┴┴┤ ▼ length
+//! │  frame...    │
+//! ```
+//!
+//! # Future work
+//! - [`fadvise`](https://linux.die.net/man/2/posix_fadvise)-based APIs to pre-fetch parts of the file, to improve random access performance.
+//! - Use an inner [`Blockstore`] for writes.
+//! - Support compressed snapshots.
```
ruseinov, line 55: If we had a modest frame size for zstd archives - we could have the same mapping to an archive as we have to raw `.car` file, but with an extra computation step + having to decompress a particular frame. Curious what the performance overhead is for an archive that consists of small frames. It seems that the storage overhead is negligible, even if we include a full header with each frame.

src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,272 @@
...
+            Cid::read_bytes(&mut *reader),
+        ) {
+            (Ok(length), Ok(offset), Ok(cid)) => {
+                let next_block_offset = offset + u64::try_from(length).unwrap();
+                if let Err(e) = reader.seek(SeekFrom::Start(next_block_offset)) {
+                    return Some(Err(cid::Error::Io(e)));
+                }
+                Some(Ok((cid, BlockLocation { offset, length })))
+            }
+            (Err(e), _, _) | (_, Err(e), _) => Some(Err(cid::Error::Io(e))),
```
ruseinov, line 222: nit: Perhaps some simple comments for the error cases here could also be beneficial as it gets a little convoluted with all the different types of errors and nesting.

ruseinov (approved):
Looks good, left some thoughts and nits.

src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,272 @@
...
+//! **Note that all operations on this store are blocking**.
+//!
+//! On creation, [`CarBackedBlockstore`] builds an in-memory index of the [`Cid`]s in the file,
+//! and their offsets into that file.
+//!
+//! When a block is requested [`CarBackedBlockstore`] scrolls to that offset, and reads the block, on-demand.
+//!
+//! Writes are currently cached in-memory.
+//!
+//! Performance is pathological with random access.
```
LesnyRumcajs, line 15: What do you mean by pathological? Let's clarify it with some numbers.

src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,272 @@
...
+            }
+            (Some(BlockLocation { offset, length }), Vacant(_)) => {
+                debug!("fetching from disk");
+                let mut reader = self.reader.lock();
+                reader.seek(SeekFrom::Start(*offset))?;
+                let cid = Cid::read_bytes(&mut *reader)?;
+                assert_eq!(cid, *k);
+                let cid_len = reader.stream_position()? - *offset;
+                let data_len = *length - usize::try_from(cid_len).unwrap();
+                let mut data = vec![0; data_len];
```
LesnyRumcajs, line 153: Why do we need to initialize it with zeroes? Given that this chunk of memory will get populated in the next line, this may be redundant.

src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,272 @@
...
+    match unsigned_varint::io::read_usize(reader) {
+        Ok(u) => Ok(u),
+        Err(Io(e)) => Err(e),
+        Err(Decode(e)) => Err(io::Error::new(InvalidData, e)),
+        Err(other) => Err(io::Error::new(Other, other)),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
LesnyRumcajs, line 240: For the sake of coverage, we could add cases for empty and corrupted CAR files.

lemmih (changes requested):
Love it. I changed `forest-cli snapshot validate` to use this code, and here are the numbers when validating a 36GiB snapshot:
```
    CarDB:    total: 44s,  indexing: 29s, walking: 15s
    ParityDB: total: 171s, loading: 147s, walking: 24s
```

Results:

- No wasted disk space. This blockstore doesn't require any disk space (as long as the CAR files are uncompressed).
- Significantly faster load times: 29s vs 147s.
- Significantly faster access times: 15s vs 24s to visit every node in the graph.

Future improvements:

- The `CarBackedBlockstore` takes a buffered reader. This might not be the best option, especially when we know exactly how much to read after seeking.
- The CIDs are not validated. The `CarReader` validates that the CIDs are correct. We should do the same (but in parallel).


This blockstore is a huge improvement for `forest-cli snapshot validate`. Let's use it immediately.

src/car_backed_blockstore.rs
aatifsyed: > Why do we need to initialize it with zeroes

In short, because rust is a memory-safe language.

You're not allowed to observe uninitialised memory (though the compiler is allowed to optimise memory that you never observe away).

We could jump through hoops, with 
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.set_len or https://doc.rust-lang.org/std/mem/union.MaybeUninit.html

But I don't know if it's worth it yet

src/car_backed_blockstore.rs
ruseinov: Medical background shows! This is also a valid mathematical term, I would, however, clarify it for posterity's sake.

src/car_backed_blockstore.rs
ruseinov: Looks good as is to me, some context: 
https://github.com/rust-lang/rust/pull/40409
https://users.rust-lang.org/t/is-set-len-the-way-to-make-an-unintialized-vec-as-a-buffer/38248/4

hanabi1224:
How feasible is it to support multiple `car` files? I can think of a scenario that uses a subset `car` file of a snapshot and the manifest bundle `car` file (2 car files in total) to unit test state migration(s)

lemmih:
> How feasible is it to support multiple `car` files? I can think of a scenario that uses a subset `car` file of a snapshot and the manifest bundle `car` file (2 car files in total) to unit test state migration(s)

The easiest way to handle that would be to concatenate the two CAR files.

aatifsyed:
> The easiest way to handle that would be to concatenate the two CAR files.

From a code read, I'm not sure our current code handles, that, neither would this PR, but this could be a future enhancement.

src/car_backed_blockstore.rs
LesnyRumcajs: I'm not convinced. `calloc` is more expensive than `malloc`, right? I mean, if there is no clean way to do it nicely in Rust then I'm good with this, but requesting zero-initialized memory is wasteful from what I see here.

aatifsyed: There is no alternative in safe rust :shrug:

src/car_backed_blockstore.rs
```diff
@@ -139,100 +149,126 @@ where
...
-                }
-                Some(Ok((cid, BlockLocation { offset, length })))
-            }
-            (Err(e), _, _) | (_, Err(e), _) => Some(Err(cid::Error::Io(e))),
-            (_, _, Err(e)) => Some(Err(e)),
+/// The following functions may hit EOF, or return data.
+/// We represent EOF as [`None`].
+///
+/// This macro helps us compose these functions
+macro_rules! ok {
```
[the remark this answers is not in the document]
aatifsyed: This was my take on https://github.com/ChainSafe/forest/pull/3085#discussion_r1243842981 :shrug:

I could change to `Result<Option<_>>` instead?

src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,333 @@
...
+#[derive(Debug)]
+struct BlockDataLocation {
+    offset: u64,
+    length: u32,
+}
+
+/// See [module documentation](mod@self) for more.
+pub struct CarBackedBlockstore<ReaderT> {
+    // https://github.com/ChainSafe/forest/issues/3096
+    inner: Mutex<CarBackedBlockstoreInner<ReaderT>>,
```
aatifsyed, line 96: @lemmih did you have to do something similar, or am I missing something?

aatifsyed: See also #2676

lemmih: Yeah, I had to do something similar.

src/car_backed_blockstore.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub fn roots(&self) -> Vec<Cid> {
+        self.inner.lock().roots.clone()
+    }
+
+    /// To be correct:
+    /// - `reader` must read immutable data. e.g if it is a file, it should be [`flock`](https://linux.die.net/man/2/flock)ed.
+    ///   [`Blockstore`] API calls may panic if this is not upheld.
+    /// - `reader`'s buffer should have enough room for the [`CarHeader`] and any [`Cid`]s.
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub fn new(mut reader: ReaderT) -> cid::Result<Self> {
```
lemmih, line 112: Would wrapping `reader` in a `BufReader` improve performance during the indexing?

aatifsyed: I've tested a few buffer sizes, on a 36G snapshot and a 2.6G snapshot

```console
$ du -h ...
36G     /home/aatif/chainsafe/snapshots/filecoin_full_calibnet_2023-04-07_450000.car
2.6G    /home/aatif/chainsafe/snapshots/forest_snapshot_calibnet_2023-06-29_height_690463.car
```

# Large snapshot
Note there's a low confidence here - n=1

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./target/release/examples/benchmark --mode buffer8k /home/aatif/chainsafe/snapshots/filecoin_full_calibnet_2023-04-07_450000.car` | 226.545 | 226.545 | 226.545 | 1.05 |
| `./target/release/examples/benchmark --mode buffer1k /home/aatif/chainsafe/snapshots/filecoin_full_calibnet_2023-04-07_450000.car` | 237.063 | 237.063 | 237.063 | 1.09 |
| `./target/release/examples/benchmark --mode buffer100 /home/aatif/chainsafe/snapshots/filecoin_full_calibnet_2023-04-07_450000.car` | 216.535 | 216.535 | 216.535 | 1.00 |
| `./target/release/examples/benchmark --mode unbuffered /home/aatif/chainsafe/snapshots/filecoin_full_calibnet_2023-04-07_450000.car` | 234.794 | 234.794 | 234.794 | 1.08 |

# Small snapshot
Probably not worth a lot, because the OS will probably keep a hot cache of the file on disk

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./target/release/examples/benchmark --mode buffer8k /home/aatif/chainsafe/snapshots/forest_snapshot_calibnet_2023-06-29_height_690463.car` | 3.229 ± 0.248 | 2.946 | 3.660 | 1.32 ± 0.13 |
| `./target/release/examples/benchmark --mode buffer1k /home/aatif/chainsafe/snapshots/forest_snapshot_calibnet_2023-06-29_height_690463.car` | 2.443 ± 0.148 | 2.276 | 2.668 | 1.00 |
| `./target/release/examples/benchmark --mode buffer100 /home/aatif/chainsafe/snapshots/forest_snapshot_calibnet_2023-06-29_height_690463.car` | 2.562 ± 0.305 | 2.149 | 3.043 | 1.05 ± 0.14 |
| `./target/release/examples/benchmark --mode unbuffered /home/aatif/chainsafe/snapshots/forest_snapshot_calibnet_2023-06-29_height_690463.car` | 5.788 ± 0.192 | 5.573 | 6.105 | 2.37 ± 0.16 |


Note that `std::io::Seek::seek`-ing on a `std::io::BufReader` always discards the buffer.

So _maybe_ a small buffer is good for our use-case here? I'd say it's not clear, and I don't want to spend too much time benchmarking. Gonna have a 100 byte buffer and call it at that, please do double check my benchmarks if you've got time: here's the code https://github.com/ChainSafe/forest/commit/7a20c4c9e4b7b583ec7d7593c823e0688e89e374

lemmih (approved):
Lovely.

src/car_backed_blockstore.rs
lemmih: Unabridged results over 10 runs:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./target/release/examples/benchmark --mode buffer8k ../lotus/filecoin_full_calibnet_2023-04-07_450000.car` | 52.861 ± 0.587 | 52.156 | 54.317 | 1.02 ± 0.02 |
| `./target/release/examples/benchmark --mode buffer1k ../lotus/filecoin_full_calibnet_2023-04-07_450000.car` | 52.028 ± 0.687 | 51.455 | 53.880 | 1.00 |
| `./target/release/examples/benchmark --mode buffer100 ../lotus/filecoin_full_calibnet_2023-04-07_450000.car` | 53.525 ± 0.638 | 52.877 | 55.187 | 1.03 ± 0.02 |
| `./target/release/examples/benchmark --mode unbuffered ../lotus/filecoin_full_calibnet_2023-04-07_450000.car` | 85.451 ± 1.011 | 84.166 | 87.108 | 1.64 ± 0.03 |

PR #3087, opened 2023-06-28
feat: print state diff on state mismatch

hanabi1224:
## Summary of changes

- Print detailed state diff on state mismatch during tipset validation to make troubleshooting easier(for state migration etc.).

Testing steps: 
- `git checkout hm/nv17-migrations`
- `git merge hm/print_state_diff_on_state_missmatch`
- Change `BytesKey(client_id.payload_bytes())` to `BytesKey(client_id.to_bytes())` [here](https://github.com/ChainSafe/forest/pull/2983/files#diff-63d98fd869120791415854c513c2d17779e40c974eb6e97db20af0b269f6691dR101)
- `cargo run --bin forest --release -- --chain calibnet --encrypt-keystore false --import-snapshot https://forest-snapshots.fra1.digitaloceanspaces.com/debug/filecoin_calibnet_height_16900.car.zst --no-gc --halt-after-import --height=-200`

Output:
```console
Address t06 changed:
 {
  "code": {
    "/": "bafk2bzacebh7dj6j7yi5vadh7lgqjtq42qi2uq4n6zy2g5vjeathacwn2tscu"
  }

  "sequence": 0

  "balance": "0.0"

  "state": [
    {
      "/": {
        "bytes": "mAFA"
      }
    }

    {
      "/": "bafy2bzacebgeizyq72bgriabprs2h6enteryca7axtmymdpwxbvkdvk3vsk5s"
    }

    {
      "/": "bafy2bzaceamp42wmmgr2g2ymg46euououzfyck7szknvfacqscohrvaikwfay"
    }
-
    {
      "/": "bafy2bzacedtfehthxk6tzjxt7shn3j5n5fqu4grncflhyczxm5yt3bige53vu"
    }
+
    {
      "/": "bafy2bzaced4ex6qplbu67qhrxvxnjxk4vjpuyi3obso5idcmwd3giyxvetqu2"
    }

    {
      "/": {
        "int": "39"
      }
    }

    {
      "/": "bafy2bzaceamp42wmmgr2g2ymg46euououzfyck7szknvfacqscohrvaikwfay"
    }
  ]
}
```

Closes

## Change checklist


src/state_manager/mod.rs
```diff
@@ -1297,6 +1298,15 @@ where
                             ?actual_receipt,
                             "state mismatch"
                         );
+                        // TODO: Download expected state from network when not available in local DB
```
lemmih, line 1301: This function requires the state to be present in the database. Validating tipsets received from the network (as opposed to a snapshot file) happens elsewhere. We can remove this TODO.

lemmih (approved):
I'm okay with merging this as-is. Going forward, the code for validating tipsets in a snapshot should move from the daemon into the cli. And it would be nice if printing out a state-diff was guarded behind a flag.

hanabi1224:
> I'm okay with merging this as-is. Going forward, the code for validating tipsets in a snapshot should move from the daemon into the cli. And it would be nice if printing out a state-diff was guarded behind a flag.

How about using env var `FOREST_VERBOSE_STATE_DIFF`?

lemmih:
> > I'm okay with merging this as-is. Going forward, the code for validating tipsets in a snapshot should move from the daemon into the cli. And it would be nice if printing out a state-diff was guarded behind a flag.
> 
> How about using env var `FOREST_VERBOSE_STATE_DIFF`?

I'm not a fan of environment variables. They're much less discoverable than command-line flags.

src/state_manager/mod.rs
hanabi1224: Done.

src/state_manager/mod.rs
```diff
@@ -1297,6 +1298,22 @@ where
                             ?actual_receipt,
                             "state mismatch"
                         );
+                        // The flag is `false` by default, to enable it, set env var FOREST_VERBOSE_STATE_DIFF=1 or FOREST_VERBOSE_STATE_DIFF=true
+                        let should_print_state_diff =
+                            match std::env::var("FOREST_VERBOSE_STATE_DIFF") {
+                                Ok(value) if value == "1" || value.to_lowercase() == "true" => true,
+                                _ => false,
+                            };
```
lemmih, line 1306: I greatly prefer the PR without this code. It's unclear to me why you added this.

hanabi1224: Removed

LesnyRumcajs (approved):

PR #3104, opened 2023-07-03
Add progress log type

elmattic:
## Summary of changes

- Remove usage of progress bars in daemon to use progress logs.

Closes

## Change checklist


aatifsyed:
@elmattic in our chat we had a look at the [indicatif crate](https://docs.rs/indicatif/latest/indicatif/), and how their design focusses on wrapping iterators etc.

You asked _why_ this might be desirable, so here's a brief exploration.

The code we want basically looks like this:
```rust
struct ProgressLog {
    completed_items: u64,
    total_items: u64,
}

impl ProgressLog {
    fn new(total_items: u64) -> Self {
        let completed_items = 0;
        println!("{}/{}", completed_items, total_items);
        Self {
            completed_items,
            total_items,
        }
    }

    fn increment_and_emit_log(&mut self) {
        self.completed_items += 1;
        println!("{}/{}", self.completed_items, self.total_items);
    }
}

fn main() {
    let v = vec!["hello", "world"];
    let mut progress_log = ProgressLog::new(u64::try_from(v.len()).unwrap()); // User has to do setup...
    
    for item in v {
        progress_log.increment_and_emit_log(); // And maintainance
        do_work(item);
    }
}

fn do_work<T>(_: T) {}
```
[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a67d063e69d7f3f6cbbe7e450c272ab1)

This is a (relatively) large and error-prone library - the user has to do some setup, has to remember to call a method whenever they do work. We can do better :)

Let's think about what the user _actually wants_:
- They want to do some work on items in an iterator
- They want someone else to take care of the logging (that's us)

There's a foundational API being used here - the [`Iterator`](https://doc.rust-lang.org/std/iter/trait.Iterator.html) API - that provides a place for us to wrap, letting the user _transparently focus on the work they wanted to do_, while we handle presentation.

Consider the following implementation:
```rust
struct WithProgress<Inner> {
    inner: Inner,
    completed_items: u64,
}

impl<Inner> WithProgress<Inner>
where
    Inner: Iterator,
{
    fn new_iter(into_iter: impl IntoIterator<IntoIter = Inner>) -> Self {
        let inner = into_iter.into_iter();
        let completed_items = 0;
        match inner.size_hint() {
            (0, _) => todo!("already completed"),
            (nonzero, _) => println!("{}/{}", 0, nonzero),
        }
        Self {
            inner,
            completed_items,
        }
    }
}

impl<Inner> Iterator for WithProgress<Inner>
where
    Inner: Iterator,
{
    type Item = Inner::Item;

    fn next(&mut self) -> Option<Self::Item> {
        match self.inner.next() {
            Some(item) => {
                self.completed_items += 1;
                let (remaining, _) = self.inner.size_hint();
                println!(
                    "{}/{}",
                    self.completed_items,
                    u64::try_from(remaining).unwrap() + self.completed_items
                );
                Some(item)
            }
            None => {
                // TODO handle fusing
                println!("finished {} items", self.completed_items);
                None
            }
        }
    }
}

fn main() {
    let v = vec!["hello", "world"];

    for item in WithProgress::new_iter(v) {
        do_work(item);
    }
}

fn do_work<T>(_: T) {}
```
[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f41d2bc85723789a0a7e4be7f164b5b0)

The user now _doesn't need to worry_ about instantiating their logger, and making calls to it - the _callsite_ basically looks the same.

What's more, this allows our WithProgress wrapper to extend to `Stream`s, `Read/Write` types etc

Note there's work to be done here, e.g what if the user has a better estimate of the length than we do using [`Iterator::size_hint`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint), or handling [_fusing_](https://doc.rust-lang.org/std/iter/trait.FusedIterator.html)

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,117 @@
...
+            it: stream,
+            progress: self.clone(),
+        }
+    }
+}
+
+pub struct ProgressLogBuilder {
+    frequency: Option<Duration>,
+    message: Option<String>,
+}
```
aatifsyed, line 68: Do we _need_ a builder? They're often used for complex configuration of private fields, but our only consumer is internal, and we don't have that much config :shrug:

src/chain/store/index.rs
```diff
@@ -143,16 +143,15 @@ impl<BS: Blockstore> ChainIndex<BS> {
             return self.walk_back(from, to);
         }
         let total_size = from.epoch() - to;
-        let pb = ProgressBar::new(total_size as u64);
-        pb.message("Scanning blockchain ");
-        pb.set_max_refresh_rate(Some(std::time::Duration::from_millis(500)));
+        // let pb = ProgressBar::new(total_size as u64);
+        // pb.message("Scanning blockchain ");
+        // pb.set_max_refresh_rate(Some(std::time::Duration::from_millis(500)));
+        let mut wi = wrap_iter("Scanning blockchain", 0..total_size);
```
lemmih, line 149: Can't use an iterator here, unfortunately.

src/chain_sync/tipset_syncer.rs
```diff
@@ -862,16 +862,14 @@ async fn sync_headers_in_reverse<DB: Blockstore + Clone + Sync + Send + 'static,
     tracker.write().set_epoch(current_head.epoch());
 
     let total_size = proposed_head.epoch() - current_head.epoch();
-    let pb = ProgressBar::new(total_size as u64);
-    pb.message("Downloading headers ");
-    pb.set_max_refresh_rate(Some(std::time::Duration::from_millis(500)));
+    let mut wi = wrap_iter("Downloading headers", 0..total_size);
```
lemmih, line 865: Can't use an iterator here, unfortunately. Epochs are not guaranteed to be consecutive.

src/ipld/util.rs
```diff
@@ -117,10 +117,11 @@ where
...
     let estimated_total_records = estimated_total_records.unwrap_or_default();
-    let bar = ProgressBar::new(estimated_total_records);
-    bar.message(progress_bar_message.unwrap_or("Walking snapshot "));
-    bar.set_units(progress_bar::Units::Default);
-    bar.set_max_refresh_rate(Some(Duration::from_millis(500)));
+    let message = progress_bar_message.unwrap_or("Walking snapshot");
+    let bar = Arc::new(Mutex::new(ProgressLog::new(
+        message,
+        estimated_total_records,
+    )));
```
lemmih, line 124: To mimic the old progress bars, we could put the Arc+Mutex inside `ProgressLog`.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,234 @@
...
+                println!("finished {} items", self.progress.completed_items);
+                None
+            }
+        }
+    }
+}
+
+pub fn wrap_stream<S: futures_core::Stream>(message: &str, stream: S) -> WithProgressStream<S> {
+    let (lower_bound, upper_bound) = stream.size_hint();
+    let total_items = upper_bound.unwrap_or(lower_bound) as u64;
```
lemmih, line 65: I'm not sure this will work well. We might have to specify the upper bound separately.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,234 @@
...
+            );
+            self.last_logged = now;
+        }
+    }
+
+    fn finish(&mut self) {}
+}
+
+#[derive(Debug, Clone)]
+pub struct ProgressLog {
```
lemmih, line 208: How is `ProgressLog` different from `WithProgress`?

src/chain_sync/tipset_syncer.rs
elmattic: Ah yes, ok will go back to previous design.

src/ipld/util.rs
elmattic: Yes.

src/utils/io/progress_log.rs
elmattic: For now, it's not different. But since I will add `Arc+Mutex` inside it will be! The wrappers like `async_read` don't need it.

src/utils/io/progress_log.rs
lemmih: No, I don't think we should use the size hint at all.

ruseinov: > No, I don't think we should use the size hint at all.

Yeah, I think using inaccurate values for progress tracking is only going to confuse people and also won't reflect the actual state of things.

src/chain/store/index.rs
```diff
@@ -143,16 +143,12 @@ impl<BS: Blockstore> ChainIndex<BS> {
             return self.walk_back(from, to);
         }
         let total_size = from.epoch() - to;
-        let pb = ProgressBar::new(total_size as u64);
-        pb.message("Scanning blockchain ");
-        pb.set_max_refresh_rate(Some(std::time::Duration::from_millis(500)));
+        let pl = ProgressLog::new("Scanning blockchain", total_size as u64);
```
ruseinov, line 146: Now that I'm thinking about this, why not do something like:

```
ProgressLog::new().with_message("Scanning blockchain").with_total_size(total_size)
```

Makes it more flexible and explicit.

elmattic: So like some builder pattern? I did one but with @aatifsyed we agree it was superfluous.

src/utils/io/progress_bar.rs
```diff
@@ -81,6 +81,7 @@ impl ProgressBar {
         }
     }
 
+    #[allow(dead_code)]
```
lemmih, line 84: Use or delete.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,235 @@
...
+#[allow(dead_code)]
+pub fn wrap_iter<Inner>(
+    message: &str,
+    into_iter: impl IntoIterator<IntoIter = Inner>,
+) -> WithProgressIter<Inner>
+where
+    Inner: Iterator,
+{
+    let inner = into_iter.into_iter();
+    let (lower_bound, upper_bound) = inner.size_hint();
```
ruseinov, line 27: I'm guessing we want to get rid of this as well in favour of something explicit, like `with_total_size`. Unless I'm missing something - size hint is not a precise indication, so the progress tracking is going to be very confusing.

src/utils/net.rs
```diff
@@ -44,19 +43,15 @@ pub async fn reader(location: &str) -> anyhow::Result<impl AsyncRead> {
...
     };
 
-    if content_length > 0 {
-        read_progress.set_length(content_length);
-        read_progress.set_style(progress_style())
-    }
-
-    let mut reader = tokio::io::BufReader::new(read_progress.wrap_async_read(stream));
+    let mut reader =
+        tokio::io::BufReader::new(wrap_async_read("Reading ...", stream, content_length));
```
lemmih, line 54: We aren't just reading here. We're progressing a stream: Download/read -> uncompress -> write into DB. Maybe "Loading" would be a better description.

lemmih (changes requested):

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,235 @@
...
+
+#[derive(Debug, Clone)]
+pub struct ProgressLog {
+    progress: Arc<Mutex<WithProgress>>,
+}
+
+impl ProgressLog {
+    pub fn new(message: &str, total_items: u64) -> Self {
+        ProgressLog {
+            progress: Arc::new(Mutex::new(WithProgress::new(message, total_items))),
```
ruseinov, line 215: I would opt for atomic instead of `Mutex`. It's more efficient as atomic ops are supported by the CPU. That way we avoid locking altogether as well as context switching in case we hit a lock.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,235 @@
...
+    completed_items: u64,
+    frequency: Duration,
+    start: Instant,
+    last_logged: Instant,
+    message: String,
+    total_items: u64,
+}
+
+impl WithProgress {
+    fn new(message: &str, total_items: u64) -> Self {
```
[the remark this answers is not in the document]
ruseinov: Or perhaps this is better done in `ProgressLog`, since this is a private constructor anyway.

src/utils/io/progress_bar.rs
elmattic: Exactly, let's remove all of the untested code.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,235 @@
...
+        progress: WithProgress::new(message, total_items),
+    }
+}
+
+/// Wraps an iterator to display its progress.
+#[derive(Debug)]
+pub struct WithProgressStream<S> {
+    pub(crate) stream: S,
+    progress: WithProgress,
+}
```
lemmih, line 88:
```suggestion
use pin_project_lite::pin_project;
pin_project! {
pub struct WithProgressStream<S> {
   #[pin]
    pub(crate) stream: S,
    progress: WithProgress,
}
}
```

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,235 @@
...
+        match &item {
+            std::task::Poll::Ready(Some(_)) => {
+                this.progress.inc(1);
+            }
+            std::task::Poll::Ready(None) => this.progress.finish(),
+            std::task::Poll::Pending => {}
+        }
+        item
+    }
+}
```
lemmih, line 141:
```suggestion
impl<S: futures_core::Stream> futures_core::Stream for WithProgressStream<S> {
    type Item = S::Item;

    fn poll_next(
        self: std::pin::Pin<&mut Self>,
        cx: &mut std::task::Context<'_>,
    ) -> std::task::Poll<Option<Self::Item>> {
        let this = self.project();
        let item = this.stream.poll_next(cx);
        match &item {
            std::task::Poll::Ready(Some(_)) => {
                this.progress.inc(1);
            }
            std::task::Poll::Ready(None) => this.progress.finish(),
            std::task::Poll::Pending => {}
        }
        item
    }
}
```

lemmih (changes requested):
Added some comments about how to remove `Unpin`.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+        result
+    }
+
+    fn consume(mut self: Pin<&mut Self>, amt: usize) {
+        Pin::new(&mut self.stream).consume(amt);
+    }
+}
+
+impl<S: futures_core::Stream + Unpin> futures_core::Stream for WithProgressStream<S> {
```
lemmih, line 123: You want to implement `size_hint` here as well.

src/chain/store/index.rs
ruseinov: Well, this won't require the actual builder. No strong opinion here, however I don't see a reason to require `total_size` at construction, it does not always work. What that means is we'd have to explicitly pass 0 there, which makes the interface confusing and a bit dirty.

lemmih: We always have a `total_size`, don't we?

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,238 @@
...
+            let elapsed_secs = (now - self.start).as_secs_f64();
+            let elapsed_duration = format_duration(Duration::from_secs(elapsed_secs as u64));
+
+            let throughput = self.completed_items as f64 / elapsed_secs;
+
+            let eta_secs =
+                (self.total_items.saturating_sub(self.completed_items)) as f64 / throughput;
+            let eta_duration = format_duration(Duration::from_secs(eta_secs as u64));
+
+            info!(
```
lemmih, line 199: Let's use a different target. Right now, the info messages are tagged with `forest_filecoin::utils::io::progress_log` which is a bit long. Maybe just `forest::progress` would be enough.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,147 @@
...
+        stream: read,
+        progress: WithProgress::new(message, total_items),
+    }
+}
+
+pin_project! {
+    /// Wraps an iterator to display its progress.
+    pub struct WithProgressStream<S> {
+        #[pin]
+        pub(crate) stream: S,
```
lemmih, line 34: Does this field have to be public?

src/utils/io/progress_log.rs
elmattic: I removed this type since it was dead code.

src/chain/store/index.rs
elmattic: We do always have it.

src/utils/io/progress_log.rs
elmattic: This code was removed since it was sitting unused.

lemmih (approved):

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+
+    pub fn set(&self, value: u64) {
+        self.progress.lock().set(value);
+    }
+
+    pub fn set_total(&self, value: u64) {
+        self.progress.lock().set_total(value);
+    }
```
aatifsyed, line 126: I don't think this method should exist - the caller should either know how many items there are ahead of time, or not have a stable number of items. We shouldn't need to care about this last case.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
...
+impl ProgressLog {
+    pub fn new(message: &str, total_items: u64) -> Self {
+        ProgressLog {
+            progress: Arc::new(Mutex::new(WithProgress::new(message, total_items))),
+        }
+    }
+
+    pub fn set(&self, value: u64) {
+        self.progress.lock().set(value);
+    }
```
aatifsyed, line 122: I don't think this method should exist, but I'm okay with you merging it is as `#[deprecated]`, with suitable `#[allow(deprecated)] // TRACKING ISSUE: ...` comments.

Users should only be manually updating lengths if they're not on the golden path - using an `Iterator` or `Read`er or `Stream`. We should be updating callsites so they follow this pattern (but that may be nontrivial, so a separate PR may be required.

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
...
+            );
+            self.last_logged = now;
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ProgressLog {
+    progress: Arc<Mutex<WithProgress>>,
+}
```
aatifsyed, line 111: I think I understand why you've added this - you've been forced to by callsites.

But I think this should be `#[deprecated]`, and callsites should be directed to using `Iterator` or `Stream` APIs where appropriate, which is non-trivial.

We should then be able to remove any kind of synchronisation - we should NOT need any for this feature, or if we do, it should be the caller's responsibility to manage

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
```
aatifsyed, line 3: I think this needs a module comment, explaining
- what this is / how it behaves
- _why_ it exists
- when it should be used - _long_ running operations, like reading a large file
- usage principles - this _wraps_ APIs, and you shouldn't be making 'inner' calls

src/utils/io/progress_log.rs
aatifsyed: A closer look at the callsites:
- walk_snapshot
  - garbage collection _definitely wrong_
  - snapshot export _can't we just report progress on the written file?_
- sync_headers_in_reverse _is this really a long running operation?_
- get_tipset_by_height _is this really a long running operation?_

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
...
+const UPDATE_FREQUENCY: Duration = Duration::from_millis(5000);
+
+pin_project! {
+    /// Wraps an iterator to display its progress.
+    pub struct WithProgressStream<S> {
+        #[pin]
+        stream: S,
+        progress: WithProgress,
+    }
+}
```
aatifsyed, line 26: I can see why we need a separate struct - it allows us to have a `ProgressLog` a `WithProgressStream`, with the implementation detail of `WithProgress` common.

But I think we should have only one struct, if not in this PR, then in a clean up for the issues below.


Here's the API I think we should be going for
```rust
pub struct WithProgress<Inner> {
    inner: Inner,
    known_total: Option<u64>,
    seen: u64,
    last_logged: Instant,
    message: String,
}

impl<Inner> WithProgress<Inner> {
    pub fn new(inner: Inner, message: impl Display);
    pub fn new_with_known_total(inner: Inner, message: impl Display, known_total: u64);
    fn update_and_log_if_required(&mut self, newly_seen: u64);
}

// as required by user code
impl<Inner: Iterator> Iterator for WithProgress<Inner> {}
impl<Inner: AsyncRead> AsyncRead for WithProgress<Inner> {}
impl<Inner: AsyncWrite> AsyncWrite for WithProgress<Inner> {}
impl<Inner: Stream> Stream for WithProgress<Inner> {}

```

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
...
+            Poll::Pending
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct WithProgress {
+    completed_items: u64,
+    total_items: u64,
+    frequency: Duration,
```
aatifsyed, line 49:
```suggestion
```
nonblocking

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
...
+        let now = Instant::now();
+        if (now - self.last_logged) > self.frequency {
+            let elapsed_secs = (now - self.start).as_secs_f64();
+            let elapsed_duration = format_duration(Duration::from_secs(elapsed_secs as u64));
+
+            let throughput = self.completed_items as f64 / elapsed_secs;
+
+            let eta_secs =
+                (self.total_items.saturating_sub(self.completed_items)) as f64 / throughput;
+            let eta_duration = format_duration(Duration::from_secs(eta_secs as u64));
```
aatifsyed, line 96: nonblocking: I think our usecase is server logs, right? in which case is throughput and ETA appropriate?

src/utils/net.rs
```diff
@@ -44,19 +43,18 @@ pub async fn reader(location: &str) -> anyhow::Result<impl AsyncRead> {
...
-        read_progress.set_length(content_length);
-        read_progress.set_style(progress_style())
-    }
-
-    let mut reader = tokio::io::BufReader::new(read_progress.wrap_async_read(stream));
+    let mut reader = tokio::io::BufReader::new(ProgressLog::wrap_async_read(
+        "Loading",
+        stream,
+        content_length,
+    ));
```
aatifsyed, line 57: I don't think this is the right place for this - _callers_ of `reader` should decide if their call is worth logging or not

(I'm aware I touched this code :D)

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,138 @@
...
+pin_project! {
+    /// Wraps an iterator to display its progress.
+    pub struct WithProgressStream<S> {
+        #[pin]
+        stream: S,
+        progress: WithProgress,
+    }
+}
+
+impl<R: tokio::io::AsyncRead> tokio::io::AsyncRead for WithProgressStream<R> {
```
aatifsyed, line 28: This is a misnomer: `WithProgressStream` is too specific a name - note it's implementing `AsyncRead`, not `Stream`, which is suprising. See previous comment above also.

aatifsyed (changes requested):
I'm excited about this! It takes us to a cleaner UI.

I think there are some important changes that should be made - some of them to call sites, but the biggest ones are to the API surface of the new code.

If you're keen to get this in, I'm happy to punt the changes to a new issue after you've created one.

src/utils/io/progress_log.rs
ruseinov: I'm curious why won't we do something like: 
```
impl<Inner> WithProgress<Inner> {
    pub fn new(inner: Inner) -> Self;
    pub fn with_message(&self, message) -> Self;
    pub fn with_total(&self, total);
    fn update_and_log_if_required(&mut self, newly_seen: u64);
}
```

It's a much nicer api than verbose constructors and much more flexible too.

src/utils/io/progress_log.rs
lemmih: > walk_snapshot

When walking the graph, we only have an estimate of how many nodes will be visited.

> sync_headers_in_reverse is this really a long running operation?

Yes, it can take hours (or days) if you're syncing from genesis.

> get_tipset_by_height

It used to be around 10-20 minutes. Since we added checkpoints, it should be no more than a few seconds, though. It should be fine to remove the progress indicator in this case.

src/utils/io/progress_log.rs
elmattic: Agree! Yes this should be done in a follow up PR, improvements to the logger, API and updating callsites to use maybe a less procedural approach (and favor streams, iterator).

src/utils/io/progress_log.rs
aatifsyed: That also works!

ruseinov: Yeah, message should probably be required, unless we want this to be flexible for some reason. No strong opinion there, would just prefer to have 1 constructor instead of several. And for optional params use this pattern. I'm still not sure if there's ever a time when we don't know the `total_size` at init.

src/utils/io/progress_log.rs
elmattic: Good question. I consider that right now we are probably the first users and spending more time looking at logs that anyone else.
And I find throughput and ETA still valuable. But at some point we could make them optional.

src/utils/io/progress_log.rs
elmattic: Any suggestion for a better name?

src/utils/net.rs
elmattic: Hmmm. With the new progress logger design, if the call is not worth logging (< 5s) you won't see anything in the console.

src/utils/io/progress_log.rs
elmattic: Agree. Removed logs for `get_tipset_by_height`.

elmattic: Not sure, but `#[deprecated]` will probably make Clippy unhappy, no?

src/utils/io/progress_log.rs
ruseinov: `WithProgressAsync` ?

src/utils/io/progress_log.rs
ruseinov: Sure, not blocking.

src/utils/io/progress_log.rs
aatifsyed: I think we should bin the estimate in that case - else we've just got two counts increasing at different rates until they are the same. Let's just keep the one?

src/utils/io/progress_log.rs
aatifsyed: I think we should just have a top-level `WithProgress`. Implementations of sync traits shouldn't care about the pinned inner. 

If we can't do that right now, then we should keep this as `WithProgress`, and rename `ProgressLog` to something like `WithProgressRaw`, with a view to hiding it.

It's an iterator|reader|writer|stream _with progress_ logs

src/utils/io/progress_log.rs
```diff
@@ -0,0 +1,176 @@
...
+            this.progress.inc(buf.filled().len() as u64 - prev_len);
+            Poll::Ready(e)
+        } else {
+            Poll::Pending
+        }
+    }
+}
+
+impl<S> WithProgress<S> {
+    pub fn wrap_async_read(message: &str, read: S, total_items: u64) -> WithProgress<S> {
```
aatifsyed, line 84:
```suggestion
    pub fn new(message: &str, read: S, total_items: u64) -> WithProgress<S> {
```

aatifsyed (approved):
Great work!

PR #3107, opened 2023-07-03
[Enhancement] Introduce 'head' parameter for snapshot validation.

ruseinov:
## Summary of changes

- Introduce `HEAD` param to make sure we can specify ranges when importing and validating a snapshot.

Closes

## Change checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,9 +43,10 @@
 - [#3086](https://github.com/ChainSafe/forest/pull/3085):
   `forest-cli snapshot validate` is faster and uses less disk space, operating
   directly on the snapshot rather than loading through a database.
-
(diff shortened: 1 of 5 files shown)


lemmih (approved):

lemmih:
Using `i64` (or `ChainEpoch`) might be easier.

ruseinov:
> Using `i64` (or `ChainEpoch`) might be easier.
I agree that it's not the most eloquent of solutions.

It makes no sense to accept `i64` though as we don't want to allow negative values for this one. Hence the `u64`. 
In an ideal world we should have validation for this and all the other CLI params.  Perhaps a good issue to create for when we'd want to re-vamp and move those out of the daemon?

elmattic (approved):

Issue #3119, opened 2023-07-04
Adapative request window size


LesnyRumcajs:
**Issue summary**

### Problem
The request window size of 32, introduced in https://github.com/ChainSafe/forest/pull/2890, appears to increase the risk of a lookup error, at least on mainnet.
```
2023-07-04T07:03:02.137812Z ERROR forest_filecoin::chain_sync::chain_muxer: Bootstrapping failed, re-evaluating the network head to retry the bootstrap. Error = TipsetRangeSyncer(NetworkMessageQueryFailed("ChainExchange request failed for all top peers. 9 network failures, 23 lookup failures, request:\nChainExchangeRequest { start: [Cid(bafy2bzaceabapzbxar7gh5jpoh2jpcd2gzdb36x4mocg5v3owjq3ki6twanvw), Cid(bafy2bzaceaqy553mahvlt2zfclnzud4iizvydyzggpwuavoil6l34jmt7grfc), Cid(bafy2bzacedtjnls5v5dsbsuoy7gjskid52jb2lwonxwhkuvjqjgx4tj36w4da), Cid(bafy2bzaceccy3czx4ooc3ovvrycl65s5k65i7sxdjx2cnp4n47tay5rwirada), Cid(bafy2bzacec6ybx6nla55w4ljn5c2f5dd3s7vouft3emcd22cl65ld77egwjus), Cid(bafy2bzacecu7go4q5bi736ean6brjkqeltq7duroq5bpwc2pbjqwksxeqguj6), Cid(bafy2bzacea7fa35jym3727z62xztdp57jnb6bov3it66qurhv5f45wqa77p6y), Cid(bafy2bzacebyyhr3l44nb2i3hto6zredfd77phgtrce4ll7b4zcbigah7idmx4)], request_len: 32, options: 2 }"))
```
Nonetheless, a higher value also ensures faster sync time (from the same PR)
> Time taken for head sync (calibnet):
> | Request window size  | Time to sync |
> | ------------- | ------------- |
> | 8  | 10 minutes  |
> | 16  | 7 minutes  |
> | 32  | 5 minutes  |
> | 64  | 5 minutes  |

For reference, the default for Lotus is 8.

### Solution

To be both performant and resilient, we could introduce a concept of an adaptive request window size. We would default to a window of 32 (or whatever the override is), but in case of a lookup issue, we would go down to 16, and then if that fails, down to 8 and so on.
```
if lookup_error
  request_size = max(request_size / 2, 1)
end
```

[...]

PR #3129, opened 2023-07-05
feat: adaptive window size for tipset syncer

hanabi1224:
## Summary of changes

- [x] Implement `AdaptiveValueProvider`
- [x] Tests
- [x] Manually test on mainnet with https://forest-snapshots.fra1.cdn.digitaloceanspaces.com/debug/forest_snapshot_mainnet_2023-06-28_height_2988801.car.zst

Closes https://github.com/ChainSafe/forest/issues/3119

## Change checklist


src/utils/misc/value_provider.rs
```diff
@@ -0,0 +1,226 @@
...
+                        Some(i)
+                    }
+                })
+                .expect("Infallible");
+        }
+    }
+
+    /// Tracks failure, a value downgrade will be triggered when consecutive failures is greater than `config.downgrade_threshold`
+    pub fn track_failure(&self) {
+        let old = self.consective_failure_counter.fetch_add(1, ORDERING);
```
hanabi1224, line 79: AtomicUsize does not panic abt overflow so no special handling is needed

lemmih:
I'm not convinced this is the right approach. I don't understand why peers reject requests of 32 blocks, and I'm not sure if we're being a bad network participant by requesting that many blocks. Also, we don't need to request more than the default amount of blocks to fully saturate our pipeline if we use proper streaming logic. The current code is quite awkward and won't request more than 1 batch at a time.

lemmih:
Let's hold off on merging this until we can benchmark against a streaming solution.

lemmih (changes requested):
Blocking until we can benchmark against a streaming solution.

PR #3137, opened 2023-07-06
[Enhancement] Bail on database lookup errors.

### LINKED ISSUE #3126: [Enhancement] Figure out a way to throw a lookup error when there aren't enough state-roots.
**Issue summary**

The current problem is that we get a state mismatch on both lotus and forest if we don't have enough state-roots to compute something, instead of throwing a lookup error.

How to reproduce: 
1. load a `snapshot_2986560_2989450_1688016756.car.zst` into Forest
2. `cargo run --release --bin forest --  --encrypt-keystore false --import-snapshot snapshot_2986560_2989450_1688016756.car.zst --height=2987370 --head=2987371 --skip-load=true`
3. State mismatch!

How to have proper state:
1. load a `snapshot_2983680_2986570_1687930358.car.zst` into Forest, before following the aforementioned steps.
..
3. No state mismatch!

ruseinov:
## Summary of changes

- Propagate database lookup errors properly to avoid mismatches caused by missing state roots. 

Closes #3126 

## Change checklist

### DIFF
--- a/src/interpreter/fvm.rs
+++ b/src/interpreter/fvm.rs
@@ -1,9 +1,11 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::{cell::Ref, sync::Arc};
 
 use crate::blocks::BlockHeader;
+use crate::interpreter::errors::Error;
 use crate::networks::ChainConfig;
 use crate::shim::{
     gas::{price_list_by_network_version, Gas, GasTracker},
@@ -23,6 +25,7 @@ use fvm_shared::{
     clock::ChainEpoch,
     consensus::{ConsensusFault, ConsensusFaultType},
 };
+use tracing::{error, info};
 
 use crate::interpreter::resolve_to_key_addr;
 
@@ -33,6 +36,7 @@ pub struct ForestExternsV2<DB> {
     lookback: Box<dyn Fn(ChainEpoch) -> anyhow::Result<Cid>>,
     db: DB,
     chain_config: Arc<ChainConfig>,
+    bail: AtomicBool,
 }
 
 impl<DB: Blockstore> ForestExternsV2<DB> {
@@ -51,6 +55,7 @@ impl<DB: Blockstore> ForestExternsV2<DB> {
             lookback,
             db,
             chain_config,
+            bail: AtomicBool::new(false),
         }
     }
 
@@ -90,13 +95,17 @@ impl<DB: Blockstore> ForestExternsV2<DB> {
         Ok((addr.into(), gas_used.round_up() as i64))
     }
 
-    fn verify_block_signature(&self, bh: &BlockHeader) -> anyhow::Result<i64> {
+    fn verify_block_signature(&self, bh: &BlockHeader) -> anyhow::Result<i64, Error> {
         let (worker_addr, gas_used) =
             self.worker_key_at_lookback(&bh.miner_address().into(), bh.epoch())?;
 
         bh.check_block_signature(&worker_addr.into())?;
 
         Ok(gas_used)
+    }
+
+    pub fn bail(&self) -> bool {
+        self.bail.load(Ordering::Relaxed)
     }
 }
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,9 @@
 
 ### Changed
 
+- [#3126](https://github.com/ChainSafe/forest/issues/3126): Bail on database
+  lookup errors.
```
LesnyRumcajs, line 35:
```suggestion
  lookup errors instead of silently ignoring them.
```

src/interpreter/fvm.rs
```diff
@@ -218,24 +227,47 @@ impl<DB: Blockstore> Consensus for ForestExternsV2<DB> {
...
-                    total_gas += gas_used;
-                    if let Ok(gas_used) = self.verify_block_signature(&bh_2) {
+                let res = self.verify_block_signature(&bh_1);
+
+                let match_error = |err: Error, total_gas: i64| {
+                    match err {
+                        // When a lookup error occurs we should just bail terminating all the
+                        // computations.
+                        Error::Lookup(_) => {
+                            error!("database lookup error: {}", err);
```
LesnyRumcajs, line 237:
```suggestion
                            error!("database lookup error: {err}");
```

src/interpreter/fvm.rs
```diff
@@ -218,24 +227,47 @@ impl<DB: Blockstore> Consensus for ForestExternsV2<DB> {
...
+                            Err(err)
+                        }
+                        // invalid consensus fault: cannot verify block header signature
+                        Error::Signature(_) => Ok((None, total_gas)),
```
LesnyRumcajs, line 241: Do we know why is it okay to discard this error? I get that this logic was there before, I'm just curious.

src/interpreter/fvm3.rs
```diff
@@ -225,24 +238,47 @@ impl<DB: Blockstore> Consensus for ForestExterns<DB> {
...
-                    total_gas += gas_used;
-                    if let Ok(gas_used) = self.verify_block_signature(&bh_2) {
+                let res = self.verify_block_signature(&bh_1);
+
+                let match_error = |err: Error, total_gas: i64| {
+                    match err {
+                        // When a lookup error occurs we should just bail terminating all the
+                        // computations.
+                        Error::Lookup(_) => {
+                            error!("database lookup error: {}", err);
```
LesnyRumcajs, line 248:
```suggestion
                            error!("database lookup error: {err}");
```

src/interpreter/fvm3.rs
```diff
@@ -91,14 +96,22 @@ impl<DB: Blockstore> ForestExterns<DB> {
...
+    fn verify_block_signature(&self, bh: &BlockHeader) -> anyhow::Result<i64, Error> {
         let (worker_addr, gas_used) =
             self.worker_key_at_lookback(&bh.miner_address().into(), bh.epoch())?;
 
         bh.check_block_signature(&worker_addr.into())?;
 
         Ok(gas_used)
     }
+
+    /// Signifies where or not we have to bail due to database lookup problems.
```
LesnyRumcajs, line 108:
```suggestion
    /// Signifies whether or not we have to bail due to database lookup problems.
```

LesnyRumcajs (approved):

src/interpreter/fvm.rs
ruseinov: Yes, AFAIK when the signature is not correct - we ignore the consensus fault, but we need to report the total gas used for the lookup.

src/interpreter/fvm.rs
```diff
@@ -218,24 +227,47 @@ impl<DB: Blockstore> Consensus for ForestExternsV2<DB> {
                 // check blocks are properly signed by their respective miner
                 // note we do not need to check extra's: it is a parent to block b
                 // which itself is signed, so it was willingly included by the miner
-                if let Ok(gas_used) = self.verify_block_signature(&bh_1) {
-                    total_gas += gas_used;
-                    if let Ok(gas_used) = self.verify_block_signature(&bh_2) {
+                let res = self.verify_block_signature(&bh_1);
```
lemmih, line 230: `verify_block_signature` has three possible outputs: fatal error, non-fatal error, ok(gas_used). We can handle these three branches without any nesting. Not a big deal, but it would make the code easier to read for me.

lemmih (approved):

src/interpreter/fvm.rs
ruseinov: I was juggling with possible options to make it more readable, the trick here is that a non-fatal error still requires the gas to be passed around so the eloquence suffers there. Let me see what I can do here.

ruseinov: Fixed.

PR #3149, opened 2023-07-07
feat: use zstd compressed cars as blockstores

aatifsyed:
## Summary of changes

- new `forest snapshot compress ...` compresses snapshots with many zstd frames.
- `forest snapshot validate ...` now supports compressed snapshots prepared by the above.

See also
#3085 
#3074 

## Change checklist


src/cli/subcommands/snapshot_cmd.rs
```diff
@@ -142,20 +158,92 @@ impl SnapshotCommands {
...
+                destination,
+                compression_level,
+                frame_size,
+            } => {
+                // We've got a binary blob, and we're not exactly sure if it's compressed, and we can't just peek the header:
+                // For example, the zstsd magic bytes are a valid varint frame prefix:
+                assert_eq!(
+                    unsigned_varint::io::read_usize(&[0xFD, 0x2F, 0xB5, 0x28][..]).unwrap(),
+                    6141,
+                );
```
LesnyRumcajs, line 206: Does it belong in the runtime? It seems like something we could check at compile time, though this `read_usize` probably is not `const`. Perhaps it would belong more in tests if anywhere?

src/cli/subcommands/snapshot_cmd.rs
LesnyRumcajs, line 1: It would be great to add some integration tests for this subcommand.

LesnyRumcajs (approved):
LGTM

.config/forest.dic
```diff
@@ -8,9 +8,11 @@ bitfield
 bitswap
 BLAKE2b
 blockchain/M
+blockstore/SM
 BLS
 callee
 CAR/SM
+CARv1/SM
```
hanabi1224, line 15: nit: I would just backquote terms like `CARv1` to bypass spell checker

src/cli/subcommands/snapshot_cmd.rs
lemmih: The purpose of the assertion is to illustrate why we can't easily tell whether an archive is compressed or not.

lemmih (approved):
Excellent PoC. Love that compressed files are _faster_ than uncompressed.

Issue #3093, opened 2023-06-28
Implement `forest-cli db dump`


LesnyRumcajs:
**Issue summary**
> @lemmih: Finding all reachable nodes from a given tipset is quite slow. It's much faster to iterate over all the keys in the database and output them unordered. The resulting CAR file will be unordered and may contain unreachable nodes.

The goal here is to implement a snapshot generation that should just iterate over entries in the DB and dump them into a CAR file.

There are some caveats:
- ParityDB does not directly store the key (Cid) value - we must regenerate it. We don't have enough information. The codec is usually `DAG_CBOR`, and the hash function is `blake2b` - but not in all cases. For this reason, the plan is to have two columns in the DB - one "optimized" that assumes all entries comply with the aforementioned characteristics and another one that contains the necessary metadata.
- Given that the result will be unordered, the `HEAD` must be encoded in the CAR header frame.

**Other information and links**

LesnyRumcajs:
It turned out the proposed change doesn't provide strong benefits to defend itself, apart from decreased memory usage. We may revisit it in the future.

PR #3160, opened 2023-07-10
Implement `forest-cli db dump` and re-organize the database

LesnyRumcajs:
## Summary of changes

- implemented `forest-cli db dump` subcommand. While the idea is promising, at the moment it is not more performant than a regular `snapshot export`.
- re-organized the database to use three columns instead of one:
  - Simplified column for DAG-CBOR Blake2b256 entries which is > 99% percent of what goes into the database,
  - Full column for other Cid-keyed entries. This column is BTree indexed, allowing key retrieval at the cost of performance. Given that there are not many entries of such type, it is okay.
  - Another column, for writing arbitrary key-values.

Closes https://github.com/ChainSafe/forest/issues/3093

## Change checklist

### DIFF
--- a/src/chain/store/chain_store.rs
+++ b/src/chain/store/chain_store.rs
@@ -123,6 +123,57 @@ where
     }
 }
 
+#[async_trait::async_trait]
+pub trait DBDump {
+    /// Writes the entire database to the given writer as a CAR file. The CAR header
+    /// is initialized with the given tipset as the root.
+    async fn dump_db_as_car<W>(
+        &self,
+        writer: W,
+        tipset: &Tipset,
+        compression: bool,
+    ) -> anyhow::Result<()>
+    where
+        W: futures::AsyncWrite + Send + Unpin + 'static;
+}
+
+#[async_trait::async_trait]
+impl<DB> DBDump for Arc<ChainStore<DB>>
+where
+    DB: crate::db::Dump + Send + Sync,
+{
+    async fn dump_db_as_car<W>(
+        &self,
+        writer: W,
+        tipset: &Tipset,
+        compression: bool,
+    ) -> anyhow::Result<()>
+    where
+        W: futures::AsyncWrite + Send + Unpin + 'static,
+    {
+        let global_pre_time = SystemTime::now();
+        info!("DB dump started");
+
+        let writer = if compression {
+            Either::Left(ZstdEncoder::new(BufWriter::new(writer)))
+        } else {
+            Either::Right(BufWriter::new(writer))
+        };
+
+        self.db.write_exportable(writer, tipset).await?;
+
+        info!(
+            "DB dump finished, took {} seconds",
+            global_pre_time
+                .elapsed()
+                .expect("time cannot go backwards")
+                .as_secs()
+        );
+
+        Ok(())
+    }
+}
+
 impl<DB> ChainStore<DB>
 where
     DB: Blockstore + Send + Sync,
--- a/src/cli/subcommands/db_cmd.rs
+++ b/src/cli/subcommands/db_cmd.rs
@@ -1,18 +1,24 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use std::path::PathBuf;
 use std::sync::Arc;
 
+use crate::blocks::tipset_keys_json::TipsetKeysJson;
+use crate::cli_shared::snapshot;
 use crate::cli_shared::{chain_path, cli::Config};
 use crate::db::db_engine::db_root;
+use crate::rpc_api::db_api::DBDumpParams;
 use crate::rpc_api::progress_api::GetProgressType;
-use crate::rpc_client::{db_ops::db_gc, progress_ops::get_progress};
+use crate::rpc_client::chain_get_name;
+use crate::rpc_client::db_ops::db_dump;
+use crate::rpc_client::{chain_head, db_ops::db_gc, progress_ops::get_progress};
 use crate::utils::io::ProgressBar;
 use chrono::Utc;
 use clap::Subcommand;
 use tracing::error;
 
-use crate::cli::subcommands::{handle_rpc_err, prompt_confirm};
+use crate::cli::subcommands::{cli_error_and_die, handle_rpc_err, prompt_confirm};
 
 #[derive(Debug, Subcommand)]
 pub enum DBCommands {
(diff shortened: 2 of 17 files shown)


aatifsyed (changes requested):
I think I need more background to understand this PR

- What problem are we trying to solve?
- What does it mean to transition to a multi-column database? Why do we need it? What are the tradeoffs? How should we balance that in-code (at the interface level)

Can we think of some other ways to solve the problems we're trying to?
e.g
```rust
pub trait List: Blockstore {
    /// Return all Cids contained in this blockstore
    fn cids(&self) -> Vec<Cid>;
}

// This would compose nicely with e.g progress
// But what are e.g the performance tradeoffs?
pub fn dump(blockstore: impl List, destination: impl Write) {...}
```

src/chain/store/chain_store.rs
```diff
@@ -123,6 +123,57 @@ where
...
+        W: futures::AsyncWrite + Send + Unpin + 'static,
+    {
+        let global_pre_time = SystemTime::now();
+        info!("DB dump started");
+
+        let writer = if compression {
+            Either::Left(ZstdEncoder::new(BufWriter::new(writer)))
+        } else {
+            Either::Right(BufWriter::new(writer))
+        };
```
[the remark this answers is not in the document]
ruseinov: It's a good one. Now that I look at this - it's starting to look troublesome that we have this code snipper all over the place. 

Perhaps we should create a follow-up issue to clean this up, something like this:
1. `fn maybe_compressed(compression: bool)`
2. Make it easy to apply at the [call site](https://github.com/ChainSafe/forest/pull/3160/files#diff-52be9447b2b05d5c8e80758cd43b035f24fb0f1c7cd509aaa37f9912c434cdc4R56) as opposed to embedding it into a variety of methods
3. Clean up all of our internals and move this decision to top-level methods.

src/chain/store/chain_store.rs
```diff
@@ -123,6 +123,57 @@ where
     }
 }
 
+#[async_trait::async_trait]
+pub trait DBDump {
+    /// Writes the entire database to the given writer as a CAR file. The CAR header
+    /// is initialized with the given tipset as the root.
+    async fn dump_db_as_car<W>(
```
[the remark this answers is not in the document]
ruseinov: After looking at ParityDB's iterators it seems that the iteration order is unspecified. I'm not sure what happens when reading an unordered `.car` file. I would at least test this on a somewhat big database to see how the performance is and whether or not there are any errors. 
2 values in the database might not be enough to surface any problems.

src/cli/subcommands/db_cmd.rs
```diff
@@ -106,6 +122,67 @@ impl DBCommands {
                     }
                 }
             }
+            DBCommands::Dump {
+                output_path,
+                no_compression,
+            } => {
+                println!("Dumping database to CAR file");
```
ruseinov, line 129: I would refactor this into it's own method for readability. And some documentation would be much appreciated.

src/db/rolling/impls.rs
```diff
@@ -157,6 +157,24 @@ impl FileBackedObject for DbIndex {
...
+        Dump::total_exportable_entries(&self.current())
+    }
+
+    async fn write_exportable<W>(
+        &self,
+        writer: W,
+        tipset: &crate::blocks::Tipset,
+    ) -> anyhow::Result<()>
+    where
+        W: futures::AsyncWrite + Send + Unpin + 'static,
```
ruseinov, line 172: Does this need to be `'static`?

src/db/parity_db.rs
```diff
@@ -189,12 +305,248 @@ impl DBStatistics for ParityDb {
...
+
+    // This test will:
+    // * write important entries to the database,
+    // * dump the database to a CAR file,
+    // * read the CAR file and verify that the entries are present
+    #[tokio::test]
+    async fn write_exportable_test() -> anyhow::Result<()> {
+        let mut db = TempParityDB::new();
+        let data = [
+            b"h'nglui mglw'nafh Cthulhu R'lyeh".to_vec(),
```
ruseinov, line 504: I would probably invest some time looking into dumping a bigger database and then trying to read the car file / validate it comparing to what `export` does.

Just to see if there are any performance issues related to the ordering.

LesnyRumcajs:
> I think I need more background to understand this PR
> 
> * What problem are we trying to solve?
This is an experiment suggested by @lemmih. The idea was that iterating over the entries in the DB and writing them to a CAR file would be faster than walking the entire tree. It is not. (though it uses less memory).

> * What does it mean to transition to a multi-column database? Why do we need it? What are the tradeoffs? How should we balance that in-code (at the interface level). If we're transitioning to a columnar database, does it even make sense to continue with Parity?

I need help understanding the problem here. ParityDb has multi-column support out of the box. Why would we need to switch to something else when introducing more columns? 

The reasons are already commented on in the code. In short, the key (in our case, CIDs) is not retrievable with the current settings (it is not using a BTree storage). Globally changing this incurs a performance penalty, so I introduced another column for entries that cannot be easily reconstructed from their values. An alternative is to embed the codec and hasher into the values themselves which is, subjectively, a worse approach and likely less performant.

If you look at the changes in the code, the public interface to the DB did not change. Only the ParityDb module should know about the existence of many columns and handle them.

> 
> Can we think of some other ways to solve the problems we're trying to? e.g
> 
> ```rust
> pub trait List: Blockstore {
>     /// Return all Cids contained in this blockstore
>     fn cids(&self) -> Vec<Cid>;
> }
> 
> // This would compose nicely with e.g progress
> // But what are e.g the performance tradeoffs?
> pub fn dump(blockstore: impl List, destination: impl Write) {...}
> ```

I'd rather not put on the heap billions of Cids. Memory-wise this would be even more resource-consuming than walking the tree.

lemmih:
> All in all, given that it is a failed experiment, I'm happy to close this, especially given that it's unlikely anyone will use this at the current stage. I ran it to completion as suggested by @lemmih.

I can live without the `forest-cli db dump` command, but we still want the column re-organization. Iterating over the CIDs in our blockstore is required by the not-yet-implemented garbage collector.

On a side note, I still think this command can be significantly faster than a regular snapshot export. It turned out that the car writer was the bottleneck, but we can totally fix that.

LesnyRumcajs:
> On a side note, I still think this command can be significantly faster than a regular snapshot export. It turned out that the car writer was the bottleneck, but we can totally fix that.

It wasn't that much of a bottleneck. Even with a dummy sink (and not re-creating the Cids) the dump time was "only" 50%  faster than a regular export. But yeah, we can revisit this in the future.

LesnyRumcajs:
> I can live without the forest-cli db dump command, but we still want the column re-organization. Iterating over the CIDs in our blockstore is required by the not-yet-implemented garbage collector.

Closing. I'll push the DB-reorg with some additional changes in another PR.

PR #3161, opened 2023-07-10
feat: forest-cli snapshot compute-state

### LINKED ISSUE #3124: Implement `forest-cli snapshot compute-state`
**Issue summary**

For debugging, it would help if we had something similar to `lotus state compute-state`. Since this computation doesn't rely on connecting to the p2p network, let's implement it directly operating on snapshot files.

Actions:
- [x] Set up the VM using a snapshot file.
- [x] Implement the equivalent of the `--vm-height` and `--json` flags.
- [x] Verify that the JSON output is roughly comparable with Lotus'.

https://github.com/filecoin-project/lotus/blob/4240990c6c4158651b954ca770535b1d7ba5874b/cli/state.go#L1027

elmattic:
## Summary of changes

- Add a `snapshot compute-state` subcommand to forest-cli

Example:

`./target/release/forest-cli snapshot compute-state --snapshot forest_snapshot_calibnet_2023-06-20_height_664544.car.zst --vm-height 664532 --json`

### Caveats
- Only fvm3 execution traces are supported
- Expect a few discrepancies with Lotus traces in values like timings:`"tt"`, `"Duration"`, or empty list (`[]` instead of `null`) otherwise we should have a 100% match.
- ~~Lotus seems to fail at serialising big values: "GasLimit": 4611686018427387903, (forest) "GasLimit": 4611686018427388000, (lotus)~~ - see https://github.com/ChainSafe/forest/issues/3442
- `Message`s don't have a `cid` field - see https://github.com/ChainSafe/forest/issues/3389

Closes #3124

## Change checklist


src/cli/subcommands/compute_state_cmd.rs
```diff
@@ -0,0 +1,90 @@
...
+    json: bool,
+}
+
+impl ComputeStateCommand {
+    pub async fn run(self, config: Config) -> anyhow::Result<()> {
+        match self {
+            _ => {
+                println!("Computing state @{}", self.vm_height);
+
+                let temp = TempParityDB::new();
```
lemmih, line 40: Use the `CarBackedBlockstore`. It's fast and convenient.

elmattic:
I've added some missing tests.

src/interpreter/vm.rs
```diff
@@ -63,6 +64,97 @@ pub struct BlockMessages {
...
+            msg_receipt: ret.msg_receipt(),
+            gas_cost: MessageGasCost::new(msg, ret),
+            execution_trace: trace,
+            error: ret.failure_info().unwrap_or_default(),
+            duration: 0,
+        }
+    }
+}
+
+fn build_exec_trace(exec_trace: Vec<ExecutionEvent_v3>) -> Option<Trace> {
```
lemmih, line 121: What does this function do? Maybe give some examples.

src/shim/executor.rs
```diff
@@ -57,6 +67,34 @@ impl ApplyRet {
...
+    pub fn over_estimation_burn(&self) -> TokenAmount {
+        match self {
+            ApplyRet::V2(v2) => (&v2.over_estimation_burn).into(),
+            ApplyRet::V3(v3) => (&v3.over_estimation_burn).into(),
+        }
+    }
+
+    pub fn exec_trace(&self) -> Vec<ExecutionEvent_v3> {
+        match self {
+            ApplyRet::V2(_v2) => vec![], // We don't support fvm2 execution trace at the moment
```
lemmih, line 94: Let's open an issue for this.

src/interpreter/vm.rs
```diff
@@ -63,6 +67,112 @@ pub struct BlockMessages {
...
+    pub fn from_implicit(msg_cid: Cid, msg: &Message, ret: &ApplyRet) -> Self {
+        Self {
+            gas_cost: MessageGasCost::from_implicit(msg, ret),
+            ..Self::new(msg_cid, msg, ret)
+        }
+    }
+}
+
+fn build_exec_trace(exec_trace: Vec<ExecutionEvent_v3>) -> Option<Trace> {
+    let exec_trace: Option<Trace> = if !exec_trace.is_empty() {
```
lemmih, line 140: No need to check if `exec_trace` is empty.

src/interpreter/vm.rs
```diff
@@ -63,6 +67,112 @@ pub struct BlockMessages {
...
+        let mut trace_iter = exec_trace.into_iter();
+        let mut initial_gas_charges = Vec::new();
+        loop {
+            match trace_iter.next() {
```
lemmih, line 144: How about `while let Some(..) = trace_iter.next()` ?

src/json/invoc_result.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    impl From<InvocResult> for InvocResultJson {
+        fn from(ir: InvocResult) -> Self {
+            InvocResultJson(ir)
+        }
+    }
+
+    #[derive(Serialize, Deserialize)]
+    #[serde(rename_all = "PascalCase")]
+    struct JsonHelper {
```
lemmih, line 39: We own `InvocResult`. There's no need for this helper; we can directly control how to serializer `InvocResult`.

elmattic: Agree, we can do that for all the `TraceXXX` types as well and remove all boilerplate code (json helpers and quickcheck tests).

lemmih (dismissed):
Blocking the PR until Aatif can update it.

src/interpreter/vm.rs
elmattic: I don't think it's possible because you can only break with a value inside a `loop` construct.

src/cli/subcommands/snapshot_cmd.rs
```diff
@@ -496,3 +516,59 @@ fn validation_spinner(prefix: &'static str) -> indicatif::ProgressBar {
...
+    let roots = std::mem::take(&mut block_stream.header.roots);
+
+    // Initialize Blockstore
+    let store = Arc::new(AnyCar::try_from(snapshot)?);
+
+    // Prepare call to apply_block_messages
+    let ts = Tipset::load_required(&store, &TipsetKeys::new(roots.into()))?;
+
+    let genesis = ts.genesis(&store)?;
+    let network = if genesis.cid() == &*calibnet::GENESIS_CID {
```
ruseinov, line 537: Can't help but wonder if we should actually introduce a global helper for this, I suspect this won't be the only place we would want to distinguish between the three.

src/interpreter/vm.rs
ruseinov: Definitely needs some sort of doc.

aatifsyed: @elmattic feel like adding this? I still don't really know what this function does - I just refactored the json stuff

src/state_manager/mod.rs
```diff
@@ -606,18 +626,21 @@ where
...
-    ) -> Result<CidPair, Error>
+        enable_tracing: bool,
+    ) -> Result<(CidPair, ComputeStateOutput), Error>
     where
         CB: FnMut(&Cid, &ChainMessage, &ApplyRet) -> Result<(), anyhow::Error> + Send,
     {
         let this = Arc::clone(self);
-        tokio::task::spawn_blocking(move || this.compute_tipset_state_blocking(tipset, callback))
-            .await?
+        tokio::task::spawn_blocking(move || {
```
ruseinov, line 640: nit: why the extra curly braces? does not make it any more readable.

aatifsyed: this is rustfmt's choice because the line is longer

src/cli/subcommands/snapshot_cmd.rs
aatifsyed: #3394

src/interpreter/vm.rs
```diff
@@ -282,9 +406,11 @@ where
         mut callback: Option<
             impl FnMut(&Cid, &ChainMessage, &ApplyRet) -> Result<(), anyhow::Error>,
         >,
-    ) -> Result<Vec<Receipt>, anyhow::Error> {
+        enable_tracing: bool,
+    ) -> Result<(Vec<Receipt>, Vec<InvocResult>), anyhow::Error> {
         let mut receipts = Vec::new();
         let mut processed = HashSet::<Cid>::default();
+        let mut invoc_results = Vec::new();
```
LesnyRumcajs, line 413: Can we set the expected capacity to reduce re-allocations?

src/interpreter/vm.rs
```diff
@@ -335,16 +466,33 @@ where
                         ret.msg_receipt().exit_code()
                     );
                 }
+
+                if enable_tracing {
```
LesnyRumcajs, line 470: I wonder if we could reduce the number of these branches in the code, e.g, have a dynamic object.

LesnyRumcajs (commented):
I know I'm both annoying and boring with my mantra, but do we have tests for this feature?

src/interpreter/vm.rs
elmattic: The function code has been extracted from this one https://github.com/filecoin-project/filecoin-ffi/blob/v1.23.0/rust/src/fvm/machine.rs#L207.
We're only interested in the part that constructs the trace.

The trace is basically a function call graph and for each function we log all the the operations burning gas. The `build_exec_trace` is converting the `ExecutionEvent` vector to this graph structure, nothing more.

elmattic:
> I know I'm both annoying and boring with my mantra, but do we have tests for this feature?

What about adding similar tests that unit test the `build_lotus_trace` like in the [ffi repo](https://github.com/filecoin-project/filecoin-ffi/blob/v1.23.0/rust/src/fvm/machine.rs#L485)?

src/cli/subcommands/snapshot_cmd.rs
```diff
@@ -94,6 +98,18 @@ pub enum SnapshotCommands {
         #[arg(long, default_value_t = false)]
         force: bool,
     },
+    /// Compute a state transition.
+    ComputeState {
+        /// Path to a snapshot CAR, which may be zstd compressed
+        #[arg(long)]
+        snapshot: PathBuf,
```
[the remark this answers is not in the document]
elmattic: Good idea. Could we add this subcommand to `forest-tool snapshot` so we don't have to move all the code later?

src/json/invoc_result.rs
aatifsyed: done

src/interpreter/vm.rs
aatifsyed: I think n is expected to be small here, so not worth the optimisation

src/interpreter/vm.rs
aatifsyed: #3405

src/cli/subcommands/snapshot_cmd.rs
```diff
@@ -94,6 +99,33 @@ pub enum SnapshotCommands {
...
+    /// It does this by (essentially) hashing the state of the world.
+    ///
+    /// The world can change when new blocks are mined and transmitted.
+    /// A block may contain a message to e.g transfer FIL between two parties.
+    /// Blocks are ordered by "epoch", which can be thought of as a timestamp.
+    ///
+    /// Snapshots contain (among other things) these messages.
+    ///
+    /// The command calculates the state of the world at EPOCH-1, applies all
+    /// the messages at EPOCH, and prints the resulting hash of the state of the world.
```
lemmih, line 116: This implied that _two_ states were computed, rather than just one.

```suggestion
    /// The command takes the state of the world at EPOCH-1, applies all
    /// the messages at EPOCH, and prints the resulting hash of the state of the world.
```

src/interpreter/vm.rs
```diff
@@ -449,3 +494,345 @@ where
...
+        fn default() -> Self {
+            Self {
+                exit_code: ExitCode::OK,
+                return_data: vec![],
+                return_codec: 0,
+            }
+        }
+    }
+
+    fn build_lotus_trace(
```
lemmih, line 741: This function needs to be documented and refactored. What it does is fairly simple: It takes a flat array of `enter`, `return`, and `charge` events. Then, it arranges those events in a tree. For example: `enter X, charge 10, return, enter Y, charge 20, return` becomes `[Enter X [Charge 10], Enter Y [Charge 20]]`.

This function should _not_ use an iterator for `trace_iter`. It can be written as fairly simple and idiomatic Rust.

lemmih (changes requested):
Looks good. Only a few things left to do:

- Slight refactoring here and there.
- Property testing to make sure functions like `build_lotus_trace` do not panic.
- Check that we correctly compute the state hash for calibnet genesis and mainnet genesis. These state hashes should be computable without downloading any additional data.

src/cli/subcommands/snapshot_cmd.rs
aatifsyed: Surely two states are computed? we need the state root for n-1 to calculate the state at n? or am I mistaken

src/cli/subcommands/snapshot_cmd.rs
```diff
@@ -496,3 +527,51 @@ fn validation_spinner(prefix: &'static str) -> indicatif::ProgressBar {
...
+    }
+    let beacon = Arc::new(chain_config.get_beacon_schedule(timestamp));
+    let tipset = chain_index
+        .tipset_by_height(epoch, Arc::new(ts), ResolveNullTipset::TakeOlder)
+        .context(format!("couldn't get a tipset at height {}", epoch))?;
+
+    let ((st, _), output) = apply_block_messages(
+        timestamp,
+        Arc::clone(&Arc::new(chain_index)),
+        Arc::clone(&Arc::new(chain_config)),
```
lemmih, line 559:
```suggestion
        Arc::new(chain_index),
        Arc::new(chain_config),
```

lemmih (changes requested):
`build_lotus_trace` and `build_exec_trace` need to be rewritten. See the comments I left a few weeks ago.

aatifsyed:
Leaving this PR for education, and development for this feature is now in #3430

PR #3165, opened 2023-07-11
chore: remove vestigial code

lemmih:
## Summary of changes

- `frc46_token` is unused, remove it.
- Re-exported signatures are unused, remove them.
- Remove `Inner` trait.

Closes

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,6 @@ fil_actor_verifreg_state = "6"
 fil_actors_shared = "6"
 filecoin-proofs-api = { version = "14.0", default-features = false }
 flume = "0.10"
-frc46_token = "7"
 fs_extra = "1.2"
 futures = "0.3"
 futures-util = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ fvm_ipld_blockstore = "0.2"
 fvm_ipld_car = "0.7"
 fvm_ipld_encoding = "0.4"
 fvm_ipld_hamt = "0.7"
-fvm_shared2 = { package = "fvm_shared", version = "~2.5", default-features = false, features = ["testing"] }
+fvm_shared2 = { package = "fvm_shared", version = "~2.5", default-features = false, features = [] }
 fvm_shared3 = { package = "fvm_shared", version = "~3.4", default-features = false, features = ["testing", "proofs"] }
 gethostname = "0.4"
 git-version = "0.3"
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -59,7 +59,6 @@ fil_actor_verifreg_state = "6"
 fil_actors_shared = "6"
 filecoin-proofs-api = { version = "14.0", default-features = false }
 flume = "0.10"
-frc46_token = "7"
```
aatifsyed, line 62: How did you find this? Do we need a better unused deps story?

Cargo.toml
```diff
@@ -71,7 +70,7 @@ fvm_ipld_blockstore = "0.2"
 fvm_ipld_car = "0.7"
 fvm_ipld_encoding = "0.4"
 fvm_ipld_hamt = "0.7"
-fvm_shared2 = { package = "fvm_shared", version = "~2.5", default-features = false, features = ["testing"] }
+fvm_shared2 = { package = "fvm_shared", version = "~2.5", default-features = false, features = [] }
```
aatifsyed, line 73: How do you feel about removing the `default-features` and `features` keys? There are no default features, and there's no point having an empty features array

aatifsyed (approved):

src/cli/subcommands/fetch_params_cmd.rs
```diff
@@ -51,14 +51,14 @@ fn ram_to_int(size: &str) -> anyhow::Result<SectorSize> {
...
-    type SectorSize = <crate::shim::sector::SectorSize as Inner>::FVM;
+    type SectorSize = crate::shim::sector::SectorSize;
 
     match trimmed {
-        "2048" | "2Ki" | "2ki" => Ok(SectorSize::_2KiB.into()),
-        "8388608" | "8Mi" | "8mi" => Ok(SectorSize::_8MiB.into()),
-        "536870912" | "512Mi" | "512mi" => Ok(SectorSize::_512MiB.into()),
-        "34359738368" | "32Gi" | "32gi" => Ok(SectorSize::_32GiB.into()),
-        "68719476736" | "64Gi" | "64gi" => Ok(SectorSize::_64GiB.into()),
+        "2048" | "2Ki" | "2ki" => Ok(SectorSize::_2KiB),
```
ruseinov, line 57: Can't help but wonder if we need to support 3 representations of each sector size. At the very least we could ignore the case and have it `to_lower`.

ruseinov (approved):
looks good to me apart from one minor nit and some deps comments that @aatifsyed has already made. We should definitely remove all the unused flags in Cargo.toml here or in a follow-up. And look into a tool that would check for unused dependencies too?

Cargo.toml
lemmih: I want to use `cargo udeps` again.

Issue #3183, opened 2023-07-13
Temporarily disable peer banning


LesnyRumcajs:
**Issue summary**
It might (or not) be the culprit for https://github.com/ChainSafe/forest/issues/3155. It isn't easy to reproduce, though, and it is our best guess at this point. Other nodes don't seem to implement peer banning.

**Other information and links**

LesnyRumcajs:
Peer banning was relaxed in https://github.com/ChainSafe/forest/commit/13071737400f37af86c0c75993f5260677b25016

PR #3184, opened 2023-07-13
Disable peer banning

LesnyRumcajs:
## Summary of changes

-

Closes https://github.com/ChainSafe/forest/issues/3183

## Change checklist


src/chain_sync/network_context.rs
```diff
@@ -363,7 +363,9 @@ where
                     RequestResponseError::ConnectionClosed
                     | RequestResponseError::DialFailure
                     | RequestResponseError::UnsupportedProtocols => {
-                        peer_manager.mark_peer_bad(peer_id).await;
```
hanabi1224, line 366: Hmm this function does not ban peers in `libp2p::Swarm` but remove them from a custom list, we should consider re-writing the logic

hanabi1224 (approved):

src/chain_sync/network_context.rs
hanabi1224: We have another `PeerManager::ban_peer` which bans peers in `libp2p::Swarm`. While `mark_peer_bad` demotes the chances of peers being selected as network query targets as I understand. I have not read thru the logic but it seems that it removes peers without a duration, we might want to re-write the logic to make it soft-banning at least

```rust
/// Removes a peer from the set and returns true if the value was present
    /// previously
    pub async fn mark_peer_bad(&self, peer_id: PeerId) -> bool {
        let mut peers = self.peers.write().await;
        let removed = remove_peer(&mut peers, &peer_id);
        if removed {
            metrics::FULL_PEERS.dec();
        }

        // Add peer to bad peer set
        debug!("marked peer {} bad", peer_id);
        if peers.bad_peers.insert(peer_id) {
            metrics::BAD_PEERS.inc();
        }

        removed
    }

    /// Remove peer from managed set, does not mark as bad
    pub async fn remove_peer(&self, peer_id: &PeerId) -> bool {
        let mut peers = self.peers.write().await;
        debug!("removed peer {}", peer_id);
        let removed = remove_peer(&mut peers, peer_id);
        if removed {
            metrics::FULL_PEERS.dec();
        }
        removed
    }
```

LesnyRumcajs: Ah, I see it now, thanks. Shall we temporarily remove the `ban_peer`? If not, what kind of `soft-banning` do you have in mind?

hanabi1224: Basically, the PeerManager maintains a peer list by itself, to optimize network query performance by using peer latency etc. to prioritize the list. IMHO I don't think it's worth the complexity and I would remove the entire logic, as we have seen we can get big perf improvement by just simply increasing the request window size.  

I'm not sure if the root cause of the issue is that, all peers are banned in libp2p::Swarm or in PeerManager, if this PR fixes the issue, it means PeerManager is buggy and we should maybe just remove it.

LesnyRumcajs: I'll run it for a while on a droplet. In general, in the issues logs, we don't ban that many peers, but we do mark quite a few as bad.

LesnyRumcajs:
Superseded by https://github.com/ChainSafe/forest/pull/3267

PR #3187, opened 2023-07-13
feat: stream messages into the tipset validator

lemmih:
## Summary of changes

- request a stream of messages and pipe them directly into the validator
- syncing from 2 day old snapshot goes from 10m 30s to 3m 50s on my machine. Almost entirely CPU bound.
- no performance difference when syncing to mainnet - we're already CPU bound.
- a buffer size of 64 is fairly aggressive but it only uses a bounded amount of memory.

Closes

## Change checklist

### DIFF
--- a/src/blocks/tipset.rs
+++ b/src/blocks/tipset.rs
@@ -181,7 +181,7 @@ impl Tipset {
     }
 
     /// Constructs and returns a full tipset if messages from storage exists
-    pub fn fill_tipset(&self, store: impl Blockstore) -> Option<FullTipset> {
+    pub fn fill_from_blockstore(&self, store: impl Blockstore) -> Option<FullTipset> {
         // Collect all messages before moving tipset.
         let messages = self
             .blocks()
--- a/src/blocks/tipset.rs
+++ b/src/blocks/tipset.rs
@@ -183,7 +183,7 @@ impl Tipset {
     /// Constructs and returns a full tipset if messages from storage exists
     pub fn fill_tipset(&self, store: impl Blockstore) -> Option<FullTipset> {
         // Collect all messages before moving tipset.
-        let messages: Vec<(Vec<_>, Vec<_>)> = self
+        let messages = self
             .blocks()
             .iter()
             .map(|h| crate::chain::store::block_messages(&store, h))
--- a/src/blocks/tipset.rs
+++ b/src/blocks/tipset.rs
@@ -180,15 +180,14 @@ impl Tipset {
             .transpose()?)
     }
 
-    /// Constructs and returns a full tipset if messages from storage exists -
-    /// non self version
+    /// Constructs and returns a full tipset if messages from storage exists
     pub fn fill_tipset(&self, store: impl Blockstore) -> Option<FullTipset> {
         // Collect all messages before moving tipset.
         let messages: Vec<(Vec<_>, Vec<_>)> = self
             .blocks()
             .iter()
             .map(|h| crate::chain::store::block_messages(&store, h))
-            .collect::<Result<Vec<_>,_>>()
+            .collect::<Result<Vec<_>, _>>()
             .ok()?;
 
         // Zip messages with blocks
@@ -205,7 +204,9 @@ impl Tipset {
             .collect();
 
         // the given tipset has already been verified, so this cannot fail
-        Some(FullTipset::new(blocks).unwrap())
+        Some(FullTipset::new(blocks).expect(
+            "block headers have already been verified when the Tipset so this check cannot fail",
+        ))
     }
 
     /// Returns epoch of the tipset.
--- a/src/blocks/tipset.rs
+++ b/src/blocks/tipset.rs
@@ -182,26 +182,21 @@ impl Tipset {
 
     /// Constructs and returns a full tipset if messages from storage exists
     pub fn fill_from_blockstore(&self, store: impl Blockstore) -> Option<FullTipset> {
-        // Collect all messages before moving tipset.
-        let messages = self
-            .blocks()
-            .iter()
-            .map(|h| crate::chain::store::block_messages(&store, h))
-            .collect::<Result<Vec<_>, _>>()
-            .ok()?;
-
-        // Zip messages with blocks
+        // Find tipset messages. If any are mnissing, return `None`.
         let blocks = self
             .blocks()
             .iter()
             .cloned()
-            .zip(messages)
-            .map(|(header, (bls_messages, secp_messages))| Block {
-                header,
-                bls_messages,
-                secp_messages,
+            .map(|header| {
+                let (bls_messages, secp_messages) =
+                    crate::chain::store::block_messages(&store, &header).ok()?;
+                Some(Block {
+                    header,
+                    bls_messages,
+                    secp_messages,
+                })
             })
-            .collect();
+            .collect::<Option<Vec<_>>>()?;
 
         // the given tipset has already been verified, so this cannot fail
         Some(
--- a/src/blocks/tipset.rs
+++ b/src/blocks/tipset.rs
@@ -182,7 +182,7 @@ impl Tipset {
 
     /// Constructs and returns a full tipset if messages from storage exists
     pub fn fill_from_blockstore(&self, store: impl Blockstore) -> Option<FullTipset> {
-        // Find tipset messages. If any are mnissing, return `None`.
+        // Find tipset messages. If any are missing, return `None`.
         let blocks = self
             .blocks()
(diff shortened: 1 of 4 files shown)


src/blocks/tipset.rs
```diff
@@ -180,6 +180,36 @@ impl Tipset {
             .transpose()?)
     }
 
+    /// Constructs and returns a full tipset if messages from storage exists
+    pub fn fill_tipset(&self, store: impl Blockstore) -> Option<FullTipset> {
```
aatifsyed, line 184:
```suggestion
    pub fn fill_from_blockstore(&self, store: impl Blockstore) -> Option<FullTipset> {
```
nonblocking

src/libp2p/chain_exchange/message.rs
```diff
@@ -212,6 +213,7 @@ impl TryFrom<&TipsetBundle> for FullTipset {
 
 /// Constructs a [`FullTipset`] from headers and compacted messages from a
 /// bundle.
+// TODO(aatifsyed): refactor
```
aatifsyed, line 216:
```suggestion
```
Maybe create an issue for this

src/libp2p/chain_exchange/message.rs
```diff
@@ -151,6 +151,7 @@ pub struct CompactedMessages {
     /// Unsigned BLS messages.
     pub bls_msgs: Vec<Message>,
     /// Describes which block each message belongs to.
+    /// if `bls_msg_includes[2] = vec![5]` then `block[2]` contains `bls_msgs[5]`
```
aatifsyed, line 154:
```suggestion
    /// if `bls_msg_includes[2] = vec![5]` then `TipsetBundle.blocks[2]` contains `bls_msgs[5]`
```

src/chain_sync/tipset_syncer.rs
```diff
@@ -1072,61 +1085,40 @@ async fn sync_messages_check_state<DB: Blockstore + Clone + Send + Sync + 'stati
...
-            if batch.is_empty() {
-                if let Some(full_tipset) = task_chainstore.fill_tipset(&tipset) {
-                    s.send_async(full_tipset).await?;
-                    continue;
-                }
+    let db = chainstore.blockstore();
+
+    // Stream through the tipsets from lowest epoch to highest epoch
+    stream::iter(tipsets.into_iter().rev())
+        // Chunk tipsets in batches (default batch size is 8)
```
aatifsyed, line 1093: Tell me more about what this means?

We ask peers about N tipsets. This is a batch...
If N is too big, the peer will hang up...

src/chain_sync/tipset_syncer.rs
```diff
@@ -1013,12 +1014,20 @@ fn sync_tipset<DB: Blockstore + Clone + Sync + Send + 'static, C: Consensus>(
...
+    db: &DB,
+) -> Result<Vec<FullTipset>, TipsetRangeSyncerError<C>> {
+    // If all of the tipsets can be populated with messages from the local
+    // database, do so. If any of them cannot, move forward with the full
+    // network request. It's quite rare that we already have the messages. It
+    // only happens when the local database has been seeded with the
+    // information.
+    if let Some(cached) = batch.iter().map(|tipset| tipset.fill_tipset(db)).collect() {
+        return Ok(cached);
+    }
```
aatifsyed, line 1029:
```suggestion
    if let Some(already_in_blockstore) = batch.iter().map(|tipset| tipset.fill_tipset(db)).collect() {
        // user has already seeded the database with this information (or we're recovering from e.g a crash)
        return Ok(already_in_blockstore);
    }
```
```

File: src/chain_sync/tipset_syncer.rs [hunk 6]
```diff
@@ -1013,12 +1014,20 @@ fn sync_tipset<DB: Blockstore + Clone + Sync + Send + 'static, C: Consensus>(
     })
 }
 
-async fn fetch_batch<DB: Blockstore + Clone + Send + Sync + 'static, C: Consensus>(
-    batch: &[Arc<Tipset>],
+async fn fetch_batch<DB: Blockstore, C: Consensus>(
```
aatifsyed, line 1017:
```suggestion
/// Ask peers for the [`Message`]s that these [`Tipset`]s should contain
async fn fetch_batch<DB: Blockstore, C: Consensus>(
```

Maybe this is where the preamble about batch sizes should go

src/chain_sync/tipset_syncer.rs
```diff
@@ -1032,29 +1041,33 @@ async fn fetch_batch<DB: Blockstore + Clone + Send + Sync + 'static, C: Consensu
...
-            if let Some(m) = bundle.messages {
-                crate::chain::persist_objects(chainstore.blockstore(), &m.bls_msgs)?;
-                crate::chain::persist_objects(chainstore.blockstore(), &m.secp_msgs)?;
-            } else {
-                warn!("ChainExchange request for messages returned null messages");
-            }
-
-            s.send_async(full_tipset).await?;
-        }
+        compacted_messages
```
aatifsyed, line 1044:
```suggestion
        // inflate our tipsets with the messages from the wire format
        compacted_messages
```

aatifsyed (changes requested):

src/libp2p/chain_exchange/message.rs
lemmih: Too lazy. Threw it in the bin. :)

src/chain_sync/tipset_syncer.rs
lemmih: Elaborated on batch sizes in `fetch_batch`.

src/blocks/tipset.rs
```diff
@@ -180,6 +180,36 @@ impl Tipset {
...
+                header,
+                bls_messages,
+                secp_messages,
+            })
+            .collect();
+
+        // the given tipset has already been verified, so this cannot fail
+        Some(
+            FullTipset::new(blocks)
+                .expect("block headers have already been verified so this check cannot fail"),
```
ruseinov, line 209: nit: what about returning a `Result` here instead? Then you don't have to make an assumption you make in `expect`, because this is basically a glorified `unwrap`. 

The way this has been done in some other projects I've seen is: if you want to mark a certain code path as `defensive` - just use some sort of macro for it, that would error out in dev mode and log an error in production. Or something similar. 

I think having `expect` in production code is not the best approach to be honest.

lemmih: Logically it cannot fail. Returning a `Result` from a non-fallible function is worse than a (well-described) `expect`, imo.

ruseinov: Typically `expect/panic` usages are either somewhere top-level or within tests. It's not obvious from the code above that it cannot fail either, if you want to make it obvious - do this:
```
if Ok(..) {
  return Some()
}
return None
```

That's a much better solution, than an expect to save a couple of lines of code.

lemmih: The intent is not to save a few lines of code. This function is infallible. I can prove this to another human, but I cannot (easily) prove it to rustc.

Swallowing the panic would be even worse. If this condition does not hold up then there's a bug in Forest, and we should halt.

aatifsyed: I think it's correct to panic here

src/blocks/tipset.rs
```diff
@@ -180,6 +180,31 @@ impl Tipset {
             .transpose()?)
     }
 
+    /// Constructs and returns a full tipset if messages from storage exists
+    pub fn fill_from_blockstore(&self, store: impl Blockstore) -> Option<FullTipset> {
+        // Find tipset messages. If any are mnissing, return `None`.
```
aatifsyed, line 185:
```suggestion
        // Find tipset messages. If any are missing, return `None`.
```

aatifsyed (approved):

src/blocks/tipset.rs
ruseinov: > This function is infallible.

Let's then make it truly infallible by skipping the block validation? Unless there is a good reason not to.

lemmih: > > This function is infallible.
> 
> Let's then make it truly infallible by skipping the block validation? Unless there is a good reason not to.

That is a possibility. But then we would lose out on an assertion. Let's keep it for now and find a good way of proving the correctness of the function at a later time.

ruseinov (approved):

src/blocks/tipset.rs
ruseinov: let's do. lgtm

Issue #3189, opened 2023-07-14
Re-organize the DB


LesnyRumcajs:
**Issue summary**

At the moment, the backend DB consists of a single column where key-retrieval is not possible (keys are hashed). This introduces challenges when:
- trying to list the Cids in the blockstore; while >99% of entries would be `DAG-CBOR` encoded with Blake2b256 as hasher for the Cid, there are a few entries that slip these constraints,
- Outside of the blockstore, storing settings - currently, we do it semi-manually with files.

We could enable key retrieval by making the ParityDb use BTree storage, but it would introduce a significant performance and disk space penalty.

Another approach is to encode the values on our own, e.g., reserve the first few bytes of a value for some metadata. This is not ideal, as it would increase disk space usage for no reason (over 99% of entries are DAG-CBOR with Blake2b256 hasher) and bug-prone.

The chosen approach (a byproduct of a failed experiment https://github.com/ChainSafe/forest/issues/3093) is to use multiple columns.

```rust
enum DbColumn {
    /// Column for storing IPLD data with `Blake2b256` hash and `DAG_CBOR` codec.
    /// Most entries in the `blockstore` will be stored in this column.
    GraphDagCborBlake2b256,
    /// Column for storing other IPLD data (different codec or hash function).
    /// It allows key retrieval at the cost of degraded performance. Given that
    /// there will be a small number of entries in this column, the performance
    /// degradation is negligible.
    GraphFull,
    /// Column for storing anything non-IPLD data. This column is not exportable.
    Other,
}
```

This change should not affect the public API (significantly). Clients of the DB must be completely oblivious of the underlying DB structure.

**Other information and links**

PR #3193, opened 2023-07-14
re-organize db and repurpose Store trait

LesnyRumcajs:
## Summary of changes

- re-organized the database to use three columns instead of one:
  - Simplified column for DAG-CBOR Blake2b256 entries which is > 99% percent of what goes into the database,
  - Full column for other Cid-keyed entries. This column is BTree indexed, allowing key retrieval at the cost of performance. Given that there are not many entries of such type, it is okay.
  - Another column, for writing arbitrary key-values.

Closes https://github.com/ChainSafe/forest/issues/3189

## Change checklist


src/db/parity_db.rs
```diff
@@ -3,15 +3,69 @@
...
+    /// Column for storing Forest-specific settings.
+    Settings,
+}
+
+impl DbColumn {
+    fn create_column_options(compression: CompressionType) -> Vec<parity_db::ColumnOptions> {
+        DbColumn::iter()
+            .map(|col| {
+                match col {
+                    DbColumn::GraphDagCborBlake2b256 => parity_db::ColumnOptions {
```
creativcoder, line 47: ` DbColumn::GraphDagCborBlake2b256 | DbColumn::Settings`

LesnyRumcajs: I like it. Thanks.

src/db/memory.rs
```diff
@@ -9,55 +9,65 @@ use anyhow::Result;
...
-/// A thread-safe `HashMap` wrapper.
+type MemoryDbInner = Arc<RwLock<HashMap<Vec<u8>, Vec<u8>>>>;
+
+/// A thread-safe `HashMap` wrapper, acting as a memory-backed blockstore.
 #[derive(Debug, Default, Clone)]
-pub struct MemoryDB {
-    db: Arc<RwLock<HashMap<Vec<u8>, Vec<u8>>>>,
+pub struct MemoryDB(MemoryDbInner);
+
+impl Deref for MemoryDB {
```
creativcoder, line 22: I guess, `Deref` is usually for smart pointer types. We can instead add an `into_inner()` which is standard or just do `.0`

LesnyRumcajs: It's a leftover. Removed.

src/db/parity_db.rs
```diff
@@ -140,16 +190,36 @@ impl Blockstore for ParityDb {
         D: AsRef<[u8]>,
         I: IntoIterator<Item = (Cid, D)>,
     {
-        let values = blocks
+        let values = blocks.into_iter().map(|(k, v)| {
+            let column = Self::choose_column(&k);
+            match column {
+                DbColumn::GraphDagCborBlake2b256 | DbColumn::Settings | DbColumn::GraphFull => {
```
lemmih, line 196: Is this defensive coding in case new columns are added? Can't we rely on `Self::choose_column` always picking the right column?

src/db/parity_db.rs
```diff
@@ -140,16 +190,36 @@ impl Blockstore for ParityDb {
...
+        self.db
+            .commit_changes(tx)
+            .map_err(|e| anyhow!("error bulk writing: {e}"))
     }
 }
 
 impl BitswapStoreRead for ParityDb {
     fn contains(&self, cid: &Cid) -> anyhow::Result<bool> {
-        Ok(self.exists(cid.to_bytes())?)
+        for column in [DbColumn::GraphDagCborBlake2b256, DbColumn::GraphFull] {
```
lemmih, line 212: Interesting optimization. Maybe add a note saying that this is faster because 99.99% of all values are in the `GraphDagCborBlake2b256` column. If that assumption changes in the future, this code will become slower.

src/db/tests/subtests/mod.rs
```diff
@@ -1,56 +1,56 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use crate::db::Store;
+use crate::db::SettingsStore;
 
-pub fn write<DB>(db: &DB)
+pub fn write_bin<DB>(db: &DB)
```
lemmih, line 6: This has got to be one of the strangest tests I've ever seen.

lemmih (approved):

src/db/parity_db.rs
```diff
@@ -3,15 +3,69 @@
...
+    Settings,
+}
+
+impl DbColumn {
+    fn create_column_options(compression: CompressionType) -> Vec<parity_db::ColumnOptions> {
+        DbColumn::iter()
+            .map(|col| {
+                match col {
+                    DbColumn::GraphDagCborBlake2b256 => parity_db::ColumnOptions {
+                        preimage: true,
```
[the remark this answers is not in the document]
LesnyRumcajs: Later or now? I'd rather we do it in one go to avoid yet another DB drop.

src/db/parity_db.rs
LesnyRumcajs: I believe it's kind of a leftover. I'll change it.

src/db/mod.rs
```diff
@@ -1,44 +1,53 @@
...
 pub mod rolling;
 
-/// Store interface used as a KV store implementation
-pub trait Store {
-    /// Read single value from data store and return `None` if key doesn't
-    /// exist.
-    fn read<K>(&self, key: K) -> Result<Option<Vec<u8>>, Error>
+/// Interface used to store and retrieve settings from the database.
+/// To store IPLD blocks, use the `BlockStore` trait.
+pub trait SettingsStore {
```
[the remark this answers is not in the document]
LesnyRumcajs: I'm not convinced about this. It's a constraint that doesn't bring anything (how is,e.g., base64 of garbage more readable than garbage?) but may prove problematic in the future when we will actually need it.

src/db/tests/subtests/mod.rs
lemmih: It has nothing to do with this PR.

These kinds of tests are close to being useless. It checks that writing a value into an empty DB doesn't panic. Fuzzing for panics would require the same number of LoCs and be genuinely helpful.

src/db/parity_db.rs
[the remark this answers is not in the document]
lemmih: You can do it now if you feel like it.

PR #3115, opened 2023-07-04
chore: upgrade libp2p to 0.52.1

hanabi1224:
## Summary of changes

Changes introduced in this pull request:

- upgrade libp2p to 0.52.1

## Reference issue to close (if applicable)

Closes https://github.com/ChainSafe/forest/issues/2762

## Other information and links

https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-v0.52.0

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

Issue #2762, opened 2023-04-14
Upgrade libp2p from 0.51.1


LesnyRumcajs:
**Issue summary**
The recent upgrade https://github.com/ChainSafe/forest/pull/2732 most likely introduced a regression https://github.com/ChainSafe/forest/issues/2755

Reverting this upgrade seems to mitigate the issue (all my mainnet sync trials passed for Forest pinned to `0.51.1`, whereas none passed with the updated libp2p). That was around ten trials. 

The revert https://github.com/ChainSafe/forest/pull/2759 is a temporary workaround, the library should be updated, and the `deprecated` suppressions should be removed. These can be found by gripping this ticket link in the codebase.

:warning: when updating libp2p, make sure it syncs to mainnet.

If it turns out that the issue is back again without the libp2p update, this one should be closed and the original bug ticket should get re-opened.

**Other information and links**

LesnyRumcajs:
https://github.com/ChainSafe/forest/pull/3219

thomaseizinger:
Let us (@mxinden & myself) know if we can help you with the update! :)

LesnyRumcajs:
@thomaseizinger Thanks! I really appreciate it, I will reach out once I contrive a minimal reproducible example of the issue.

hanabi1224:
Update: the issue is likely to be related to https://github.com/libp2p/rust-libp2p/pull/4250  @LesnyRumcajs has forest running well by raising the connection limit

LesnyRumcajs:
There were some other changes that may have helped, e.g., not banning peers when they abruptly close the connection.

LesnyRumcajs:
https://github.com/ChainSafe/forest/pull/3267

PR #3219, opened 2023-07-18
Revert "chore: upgrade libp2p to 0.52.1 (#3115)"

LesnyRumcajs:
This reverts commit 4469ff43b02011244220c8a82dd4beafd3a0a437.

## Summary of changes

- it seems the upgrade produced same issues as described in https://github.com/ChainSafe/forest/issues/2762

Closes

## Change checklist


aatifsyed:
Does this need a regression test?

LesnyRumcajs:
> Does this need a regression test?

How would you achieve this? For context, it's not that the chain exchange failure happens all the time with the new version. It happens sometimes, even after 24 hours. Unfortunately, we do not know currently what exact conditions are triggering this (apart from the fact that it doesn't seem to happen with the older libp2p release)

lemmih:
If we revert the PR, how will we debug the problem?

LesnyRumcajs:
> If we revert the PR, how will we debug the problem?

We can debug it on a branch. It's essential not to restart our infrastructure services every time this issue appears.

lemmih (approved):

hanabi1224 (approved):

PR #3216, opened 2023-07-17
refactor: separate `ChainStore` and `ChainIndex`

lemmih:
## Summary of changes

Changes introduced in this pull request:

- Move `tipset_from_keys` from `ChainStore` into `ChainIndex`. This is where it belongs, conceptually.
- I'll rewrite all of `ChainIndex` in a follow-up PR.

## Reference issue to close (if applicable)

Closes

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

PR #3220, opened 2023-07-18
refactor: simplify `ChainIndex`, hoist out checkpoints

lemmih:
## Summary of changes

- Get rid of the skip-link code. It was used to query really old epochs but we don't need it.
- Move `tipset_by_height` from `ChainStore` to `ChainIndex`.
- Refactor the known blocks (aka checkpoints) to be simpler. They no longer need to be verified in the daemon (can be handled offline.)
- Add new `forest-cli archive checkpoints` command for generating new checkpoints.

Closes

## Change checklist


lemmih:
This needs to be reviewed first: https://github.com/ChainSafe/forest/pull/3216

build/known_blocks.yaml
```diff
@@ -0,0 +1,52 @@
+# This file maps epochs to block headers for calibnet and mainnet. Forest use
+# this mapping to quickly identify the origin network of a tipset.
+#
+# Block headers can be inspected on filfox:
+# - https://filfox.info/en/block/bafy2bzacebnfm6dvxo7sm5thcxnv3kttoamb53uxycnvtdxgk5mh7d73qlly2
+# - https://calibration.filfox.info/en/block/bafy2bzacedhkkz76zdekpexha55b42eop42e24ajmajm26wws4nbvtq7louvu
+#
+# This file was generated with `forest-cli archive checkpoints`
```
aatifsyed, line 8: new issue:

I'd like to remove the serialization for this, and turn it into rust code, probably using something like [databake](https://docs.rs/databake/latest/databake/trait.Bake.html), but that's a little heavyweight - we could then just
```rust
struct KnownBlocks<'a> {
    calibnet: &'a [(i64, Cid)],
    mainnet: &'a [(i64, Cid)],
}
const KNOWN_BLOCKS: KnownBlocks<'static> = todo!();
```

Unfortunately, multihash and cid aren't there yet:
https://github.com/multiformats/rust-cid/issues/138
https://github.com/multiformats/rust-multihash/issues/330

src/blocks/tipset.rs
```diff
@@ -265,6 +266,49 @@ impl Tipset {
...
+            mainnet: HashMap<ChainEpoch, String>,
+        }
+
+        static KNOWN_HEADERS: OnceLock<KnownHeaders> = OnceLock::new();
+        let headers = KNOWN_HEADERS.get_or_init(|| {
+            serde_yaml::from_str(include_str!("../../build/known_blocks.yaml")).unwrap()
+        });
+
+        let calibnet_cid = Cid::from_str(calibnet::GENESIS_CID).unwrap();
+        let mainnet_cid = Cid::from_str(mainnet::GENESIS_CID).unwrap();
```
aatifsyed, line 288: wibni: `GENESIS_CID: Lazy<Cid>` rather than a string

src/chain/store/index.rs
```diff
@@ -7,129 +7,40 @@ use crate::blocks::{Tipset, TipsetKeys};
...
 
     /// `Blockstore` pointer needed to load tipsets from cold storage.
     db: Arc<DB>,
 }
 
+#[derive(Debug, Clone, Copy)]
+// Methods for resolving fetches of null tipsets.
+// Imagine epoch 10 is null but epoch 9 and 11 exist. If epoch we request epoch
+// 10, should 9 or 11 be returned?
+pub enum ResolveNullTipset {
```
aatifsyed, line 35:
```suggestion
/// Methods for resolving fetches of null tipsets.
/// Imagine epoch 10 is null but epoch 9 and 11 exist. If epoch we request epoch
/// 10, should 9 or 11 be returned?
pub enum ResolveNullTipset {
```

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
     ) -> Result<Arc<Tipset>, Error> {
-        if from.epoch() - to <= SKIP_LENGTH {
-            return self.walk_back(from, to);
+        if to == 0 {
+            return Ok(Arc::new(Tipset::from(from.genesis(&self.db)?)));
+        }
+        if to > from.epoch() {
+            return Err(Error::Other(
+                "Looking for tipset with height greater than start point".to_string(),
+            ));
```
aatifsyed, line 115: > The `from` tipset's epoch must not be smaller than `to`.

Function contract is unclear about the consequences. Should we panic in this case? I think so. Otherwise document this as a trivial failure

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
+    ///           └───▲───┘          └───┬───┘  └───────┘
+    ///               │                  │
+    ///               └──────────────────┘
+    /// ```
+    /// If the requested epoch points to a null tipset, there are two options:
+    /// Pick the nearest older tipset or pick the nearest younger tipset.
+    /// Requesting epoch 2 with `ResolveNullTipset::TakeYounger` will return
+    /// epoch 3. Requesting with `ResolveNullTipset::TakeOlder` will return
+    /// epoch 1.
+    pub fn tipset_by_height(
```
aatifsyed, line 103: Outstanding!

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
-                    checkpoint_tipsets::genesis_from_checkpoint_tipset(lbe.tipset.key())
-                {
-                    let tipset = self.load_tipset(&genesis_tipset_keys)?;
-                    info!(
-                        "Resolving genesis using checkpoint tipset at height: {}",
-                        lbe.tipset.epoch()
-                    );
-                    return Ok(tipset);
-                }
+        for (child, parent) in self.chain(from).tuple_windows() {
```
aatifsyed, line 118: Nice! new iterator APIs are really reaping dividends :))

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
+                    ResolveNullTipset::TakeOlder => return Ok(parent),
+                    ResolveNullTipset::TakeYounger => return Ok(child),
+                }
             }
-
-            cur = lbe.target.clone();
         }
+        Err(Error::Other(
+            "Tipset with epoch={to} does not exist".to_string(),
+        ))
```
aatifsyed, line 132: wibni: error types are totally wacky

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
-    ) -> Result<Arc<Tipset>, Error> {
-        self.walk_back(from, to)
+    /// Iterate from the given tipset to genesis. Missing tipsets cut the chain short.
+    pub fn chain(&self, from: Arc<Tipset>) -> impl Iterator<Item = Arc<Tipset>> + '_ {
+        itertools::unfold(Some(from), move |tipset| {
+            tipset.take().map(|child| {
+                *tipset = self.load_tipset(child.parents()).ok();
+                child
+            })
+        })
```
aatifsyed, line 142: deja vu: https://github.com/ChainSafe/forest/blob/a6ea63fb2ca4caab2e9d57a50d67af4289d24286/src/blocks/tipset.rs#L261-L268

Are these the same? If so, I think we have a problem

On one hand, this is a fairly trivial function, but on the other, I don't want to live in a codebase where there are multiple subtly different ways of doing the same thing. I think this needs
- consolidation in your PR
- some thinking about how to address the wider issue - what are your thoughts? What documentation could have led you to find the duplicate?

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
-        self.walk_back(ts, target)
+    fn tipset_child(parent: &Tipset, epoch: ChainEpoch) -> Tipset {
+        // Use a static counter to give all tipsets a unique timestamp
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let n = COUNTER.fetch_add(1, Ordering::Relaxed);
+        Tipset::from(
+            BlockHeader::builder()
+                .parents(parent.key().clone())
+                .epoch(epoch)
+                .timestamp(n)
```
aatifsyed, line 173: why do the timestamps need to be unique? why not e.g cast the epoch?

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
+                .build()
+                .unwrap(),
```
aatifsyed, line 175: new issue: this builder shouldn't exist. Use something like tidy-builder instead

src/chain/store/index.rs
```diff
@@ -151,139 +62,186 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
+            &epoch3
+        );
+    }
 
-            if to > pts.epoch() {
-                // Pts is lower than to epoch, return the tipset above that height
-                return Ok(ts);
-            }
+    #[test]
+    fn get_different_branches() {
```
aatifsyed, line 211: new issue:

wibni declarative API for building tipsets. I can implement or mentor on this, I've done similar stuff before, and it's _super important_ to make testing easy.
```rust
tipsets!(
    genesis --> epoch1a(timestamp = 1) --> epoch2a,
    genesis --> epoch1b --> epoch2b
)
```

aatifsyed (approved):
Generally happy with these changes, only serious comment is <TODO>

build/known_blocks.yaml
aatifsyed: #3227

src/chain/store/index.rs
aatifsyed: #3228

src/chain/store/index.rs
aatifsyed: #3229

src/chain/store/index.rs
lemmih: Same semantics, but different performance. `Tipset::chain` is semantically the same as `ChainIndex::chain`, and `Tipset::load` is semantically the same as `ChainIndex::load_tipset`. The only difference is caching. We need both versions.

I'm open to suggestions for documentation and/or name changes.

src/chain/store/index.rs
lemmih: I want tipsets with the same epoch to be unique. Otherwise, tipset-epoch-2a will be the same as tipset-epoch-2b in my tests and examples.

lemmih: The identity of a tipset is entirely determined by its fields. If two tipsets have the same fields, they _are_ the same tipset. Giving each tipset a unique timestamp is the easiest way of providing each tipset a unique identity.

build/known_blocks.yaml
LesnyRumcajs: Why should it be Rust code?

src/blocks/tipset.rs
```diff
@@ -265,6 +266,49 @@ impl Tipset {
...
+        // to have a known genesis.
+        #[derive(Serialize, Deserialize)]
+        struct KnownHeaders {
+            calibnet: HashMap<ChainEpoch, String>,
+            mainnet: HashMap<ChainEpoch, String>,
+        }
+
+        static KNOWN_HEADERS: OnceLock<KnownHeaders> = OnceLock::new();
+        let headers = KNOWN_HEADERS.get_or_init(|| {
+            serde_yaml::from_str(include_str!("../../build/known_blocks.yaml")).unwrap()
```
LesnyRumcajs, line 284: Could we map those strings to cids here? That is to say, have a typed value as soon as possible and avoid `to_string` later.

src/blocks/tipset.rs
```diff
@@ -265,6 +266,49 @@ impl Tipset {
...
+        });
+
+        let calibnet_cid = Cid::from_str(calibnet::GENESIS_CID).unwrap();
+        let mainnet_cid = Cid::from_str(mainnet::GENESIS_CID).unwrap();
+
+        for tipset in self.clone().chain(&store) {
+            // Search for known calibnet and mainnet blocks
+            for (genesis_cid, known_blocks) in [
```
LesnyRumcajs, line 292: Maybe we could play some golf with `find` but it's not a big deal

src/cli/subcommands/archive_cmd.rs
```diff
@@ -79,6 +79,11 @@ pub enum ArchiveCommands {
         #[arg(short)]
         depth: Option<ChainEpochDelta>,
     },
+    /// Print block headers at 30 day interval for a snapshot file
+    Checkpoints {
```
LesnyRumcajs, line 83: Could we have tests for this command?

src/cli/subcommands/archive_cmd.rs
```diff
@@ -79,6 +79,11 @@ pub enum ArchiveCommands {
         #[arg(short)]
         depth: Option<ChainEpochDelta>,
     },
+    /// Print block headers at 30 day interval for a snapshot file
```
LesnyRumcajs, line 82: Does it make sense to parametrize the interval?

src/chain/store/index.rs
```diff
@@ -151,139 +63,189 @@ impl<DB: Blockstore> ChainIndex<DB> {
...
+    /// ```text
+    ///           ┌───────┐          ┌───────┐  ┌───────┐
+    /// Genesis◄──┤Epoch 1│   Null   │Epoch 3◄──┤Epoch 4│
+    ///           └───▲───┘          └───┬───┘  └───────┘
+    ///               │                  │
+    ///               └──────────────────┘
+    /// ```
+    /// If the requested epoch points to a null tipset, there are two options:
+    /// Pick the nearest older tipset or pick the nearest younger tipset.
+    /// Requesting epoch 2 with `ResolveNullTipset::TakeYounger` will return
```
LesnyRumcajs, line 102: We could use fancy doc links instead of just marking this as code.

src/blocks/tipset.rs
lemmih: I'll leave that as a future improvement. :)

src/cli/subcommands/archive_cmd.rs
lemmih: The exact interval matters a little. A week or a quarter would also work fine. Probably not worth it to parameterize.

src/chain/store/index.rs
lemmih: You've got it, mr fancy pants. :)

scripts/tests/forest_cli_check.sh
```diff
@@ -55,6 +55,12 @@ pushd "$(mktemp --directory)"
     if "$FOREST_CLI_PATH" snapshot validate --check-network mainnet "$validate_me"; then
         exit 1
     fi
+
+    : : check that it contains at least one expected checkpoint
+    # If calibnet is reset or the checkpoint interval is changed, this check has to be updated
+    if "$FOREST_CLI_PATH" archive checkpoints "$validate_me" | grep bafy2bzaceatx7tlwdhez6vyias5qlhaxa54vjftigbuqzfsmdqduc6jdiclzc; then
```
LesnyRumcajs, line 61: I'm not sure we even need the `if` condition, given the `-e` flag to the script. It's not a big deal, though.

LesnyRumcajs (approved):
rock-solid

src/chain/store/index.rs
lemmih: totally

build/known_blocks.yaml
aatifsyed: It's less work, and makes an error state impossible.
Consider this:

```rust
use std::net::Ipv4Addr;

const MEH: &str = "127.0.0.1";
const NICE: Ipv4Addr = Ipv4Addr::new(127, 0, 0, 1);

fn meh() -> Ipv4Addr {
    MEH.parse().unwrap()
}

fn nice() -> Ipv4Addr {
    NICE
}
```

Why should `meh()` need to `unwrap`, when it can be `NICE`?
`NICE` also facilitates inlining, and all that other good stuff :)

PR #3223, opened 2023-07-18
feat: read-only hashtable

lemmih:
## Summary of changes

- read-only linear probing hashtable.
- keys are `u64`, values are `u64`.

Closes

## Change checklist


src/utils/db/cid_index.rs
```diff
@@ -0,0 +1,79 @@
...
+use std::hash::Hasher;
+
+fn hash_cid(cid: Cid) -> usize {
+    usize::from_le_bytes(cid.hash().digest()[0..8].try_into().unwrap())
+}
+
+#[derive(Debug)]
+pub struct ProbingHashtableBuilder {
+    table: Vec<Option<(usize, u64)>>,
+}
```
aatifsyed, line 14: This looks a bit in-depth, and probably very difficult to review - why not something like
```rust
#[repr(transparent)]
struct CidHash(Cid);
impl std::hash::Hash for CidHash {
    fn hash<H>(&self, state: &mut H)
       where H: Hasher {
       state.write(todo!("get the bytes of randomness you want"));
    }
}
```

aatifsyed: To be clear, I strongly discourage writing a custom hashtable implementation

aatifsyed: I wouldn't be comfortable reviewing a custom hashtable implementation - can you think of other ways to achieve your goals of low memory usage for big indexes?

Could a custom `Hash` implementation help?
```
#[repr(transparent)]
struct CidHash(Cid);
impl std::hash::Hash for CidHash {
    fn hash<H>(&self, state: &mut H) where H: Hasher {
       state.write(todo!("get the bytes of randomness you want"));
    }
}
```

Or how about doing your own probing like this?
```rust
type QuickCidHashMap = HashMap<u64, Vec<(Cid, Location)>>;
```

And so your wire format is still pretty small?

PR #3236, opened 2023-07-19
refactor: paritydb uses lz4

aatifsyed:


lemmih (approved):

creativcoder (approved):
sweet!

lemmih:
@aatifsyed Yay or nay? I f like this change. The compression should not be configurable. LZ4 is the best option; let's force it on everyone.

PR #3237, opened 2023-07-19
refactor: remove dead dbstatistics trait

aatifsyed:
stats is hardcoded to false, so this code is all dead

acceptable alternative is making this configurable via an env var


aatifsyed:
my assumptions about this being dead were wrong

see also #3246

Issue #3248, opened 2023-07-20
Unable to decode saved libp2p keypair, PeerId constantly changing


LesnyRumcajs:
**Describe the bug**

Unable to decode saved libp2p keypair

**To Reproduce** Steps to reproduce the behavior:

1. Run Forest. The keypair will be saved to `$HOME/.local/share/forest/libp2p/keypair`. 
2. Run Forest. The keypair will be overwritten by a new one. 
**Log output**

```
❯ forest --chain calibnet --encrypt-keystore false --save-token token.jwt
2023-07-20T11:52:25.648668Z  INFO forest_filecoin::daemon::main: Using default calibnet config
2023-07-20T11:52:25.649774Z  INFO forest_filecoin::daemon: Starting Forest daemon, version 0.11.1+git.2825539e4
2023-07-20T11:52:25.649801Z  INFO forest_filecoin::libp2p::service: Could not decode networking keystore!
2023-07-20T11:52:25.649982Z  INFO forest_filecoin::daemon: PeerId: 12D3KooWQ7W8XQNAakP9uXqB4PTd7BGcP8EhDwnSYTVWVkCkYSiv
2023-07-20T11:52:25.649989Z  WARN forest_filecoin::daemon: Forest has encryption disabled

❯ forest --chain calibnet --encrypt-keystore false --save-token token.jwt
2023-07-20T11:52:34.903981Z  INFO forest_filecoin::daemon::main: Using default calibnet config
2023-07-20T11:52:34.905552Z  INFO forest_filecoin::daemon: Starting Forest daemon, version 0.11.1+git.2825539e4
2023-07-20T11:52:34.905572Z  INFO forest_filecoin::libp2p::service: Could not decode networking keystore!
2023-07-20T11:52:34.905633Z  INFO forest_filecoin::utils::io: Permissions set to 0600 on File { fd: 9, path: "/home/rumcajs/.local/share/forest/libp2p/keypair", read: false, write: true }
2023-07-20T11:52:34.905650Z  INFO forest_filecoin::daemon: PeerId: 12D3KooWJsFkxNScojVutnLfTTUzyp6BPjzsveYxUgWQ6LvKmYZs
```

**Expected behaviour**

Forest correctly decodes the saved keypair. The PeerId is stable.

**Screenshots**

**Environment (please complete the following information):**

- OS:
- Rust version(e.g. `rustc --version`)
- Branch/commit

**Other information and links**

PR #3250, opened 2023-07-20
fix re-creating libp2p keypair

LesnyRumcajs:
## Summary of changes

- Forest is now able to re-create the keypair from file and thanks to this, keep the same PeerId,
- added tests so that the regression in this is caught,
- corrupted keypairs are not immediately overwritten but backed up.

Closes https://github.com/ChainSafe/forest/issues/3248

## Change checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,10 @@
 
 ### Fixed
 
+- [#3248](https://github.com/ChainSafe/forest/issues/3248): Fixed Forest being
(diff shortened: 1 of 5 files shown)


LesnyRumcajs:
> Is there any chance to make a regression test for this? Say comparing file hash across multiple daemon runs

I believe the unit test should be sufficient. It covers 100% usage of the key pair in the daemon.

lemmih (approved):

PR #3257, opened 2023-07-21
chore: replace `unsigned_varint` with custom implementation

lemmih:
## Summary of changes

- The `unsigned_varint` crate appears to be unmaintained. Let's roll our own varint encoder and decoder.
- `libp2p_bitswap` still uses `unsigned_varint` and I don't want to modify the code.

Closes

## Change checklist

### DIFF
--- a/src/utils/encoding/mod.rs
+++ b/src/utils/encoding/mod.rs
@@ -3,13 +3,11 @@
 
 use crate::shim::address::Address;
 use blake2b_simd::Params;
-use bytes::{Buf, BufMut, Bytes, BytesMut};
 use filecoin_proofs_api::ProverId;
 use fvm_ipld_encoding::strict_bytes::{Deserialize, Serialize};
-use integer_encoding::VarInt;
 pub use serde::{de, ser, Deserializer, Serializer};
-use std::io;
-use tokio_util::codec::{Decoder, Encoder};
+
+pub mod uvibytes;
 
 /// `serde_bytes` with max length check
 pub mod serde_byte_array {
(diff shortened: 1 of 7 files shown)


src/utils/encoding/mod.rs
```diff
@@ -185,3 +189,70 @@ mod tests {
         Ok(())
     }
 }
+
+// Unsigned VarInt (Uvi) Bytes
```
LesnyRumcajs, line 193: I find the lack of tests for a byte juggling features disturbing.

hanabi1224:
> libp2p_bitswap still uses unsigned_varint and I don't want to modify the code.

As a side note, libp2p_core uses `unsigned_varint` so it's not possible to remove it from the dependency tree yet

https://docs.rs/libp2p-core/latest/src/libp2p_core/upgrade/transfer.rs.html#50

src/utils/encoding/mod.rs
lemmih: Added tests (and fixed a bug)

src/utils/encoding/mod.rs
```diff
@@ -79,15 +83,123 @@ pub fn prover_id_from_u64(id: u64) -> ProverId {
     prover_id
 }
 
+// Unsigned VarInt (Uvi) Bytes
+pub struct UviBytes {
```
LesnyRumcajs, line 87: I'd put it into a different file but it's not a big deal. In general, I like when mod is just a collection of exports.

src/utils/encoding/mod.rs
```diff
@@ -79,15 +83,123 @@ pub fn prover_id_from_u64(id: u64) -> ProverId {
...
+        UviBytes::new_with_limit(128 * 1024 * 1024)
+    }
+}
+
+impl Decoder for UviBytes {
+    type Item = BytesMut;
+    type Error = io::Error;
+
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if self.len.is_none() {
```
LesnyRumcajs, line 111: Why not having this as an `else` branch to the next condition statement?

src/utils/encoding/mod.rs
```diff
@@ -79,15 +83,123 @@ pub fn prover_id_from_u64(id: u64) -> ProverId {
...
+        if item.remaining() > self.limit {
+            return Err(io::Error::new(
+                io::ErrorKind::PermissionDenied,
+                format!("uvi frame size limit exceeded, encode {}", item.remaining()),
+            ));
+        }
+        dst.reserve(item.remaining() + item.remaining().required_space());
+        let mut buffer = [0; 16];
+        let len = item.remaining().encode_var(&mut buffer);
+        dst.extend_from_slice(&buffer[0..len]);
```
LesnyRumcajs, line 152: Could we avoid this? That is to say, write to `dst` directly?

LesnyRumcajs (approved):

src/utils/encoding/mod.rs
lemmih: Nope. :(

src/utils/encoding/mod.rs
lemmih: Nope. :(

LesnyRumcajs: :crab:

src/utils/encoding/mod.rs
lemmih: Moved.

src/utils/encoding/mod.rs
lemmih: Because it has to deal with partial decoding. There are three states: Insufficient data to parse the frame length, invalid frame length, got header length (and doing incremental reads of the frame body).

sudo-shashank (approved):

PR #3283, opened 2023-07-27
feat: generate .forest.car.zst files by default

lemmih:
## Summary of changes

- `forest-cli snapshot export` generates `.forest.car.zst` output,
- `forest-cli archive export` generates `.forest.car.zst` output.
- bug fix: use tipset timestamp (rather than `now()`) when exporting.

It's more work to generate `.forest.car.zst` files but the new encoder is much faster than the old one. All in all, exporting is now a lot faster and the generated compressed files can be loaded directly into Forest.

Closes

## Change checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@
   lookup errors instead of silently ignoring them.
 - [#2999](https://github.com/ChainSafe/forest/issues/2999): Restored `--tipset`
   flag to `forest-cli snapshot export` to allow export at a specific tipset.
+- [#3283](https://github.com/ChainSafe/forest/pull/3283): All generated car
+  files use the new forest.car.zst format.
 
 ### Removed
 
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -3,58 +3,43 @@
 pub mod store;
 mod weight;
 use crate::blocks::Tipset;
+use crate::db::car::forest;
 use crate::ipld::stream_chain;
 use crate::utils::io::{AsyncWriterWithChecksum, Checksum};
 use anyhow::{Context, Result};
-use async_compression::futures::write::ZstdEncoder;
 use digest::Digest;
-use futures::future::Either;
-use futures::{io::BufWriter, AsyncWrite, AsyncWriteExt};
 use fvm_ipld_blockstore::Blockstore;
-use fvm_ipld_car::CarHeader;
-use std::sync::Arc;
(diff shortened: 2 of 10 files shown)


src/chain/mod.rs
```diff
@@ -3,58 +3,43 @@
...
 
-pub async fn export<W, D>(
-    db: impl Blockstore + Send + Sync,
+pub async fn export<D: Digest>(
+    db: impl Blockstore,
     tipset: &Tipset,
     lookup_depth: ChainEpochDelta,
-    writer: W,
-    compressed: bool,
+    writer: impl AsyncWrite + Unpin,
```
hanabi1224, line 20: How about `writer: impl AsyncBufWrite + Unpin` to enforce using BufWriter here

hanabi1224 (approved):

src/chain/mod.rs
lemmih: Do we want to enforce that?

hanabi1224: I see most of the `export` calls are passing File instead of BufWriter, should it be beneficial to use BufWriter given these exported snapshots are pretty large? if it is, should we enforce using BufWriter?

lemmih: `BufWriter` isn't automatically better, and it only improves performance when you have a lot of small writes. We do not have a lot of small writes.

sudo-shashank (approved):

CHANGELOG.md
```diff
@@ -54,6 +54,8 @@
   lookup errors instead of silently ignoring them.
 - [#2999](https://github.com/ChainSafe/forest/issues/2999): Restored `--tipset`
   flag to `forest-cli snapshot export` to allow export at a specific tipset.
+- [#3283](https://github.com/ChainSafe/forest/pull/3283): All generated car
```
LesnyRumcajs, line 57: It's a breaking change, right?

LesnyRumcajs: In other words, should any services (e.g., upload snapshot) be updated to this logic?

LesnyRumcajs:
It'd be nice to update tests to enforce this behaviour.
https://github.com/ChainSafe/forest/blob/main/scripts/tests/calibnet_export_check.sh#L18-L19

CHANGELOG.md
LesnyRumcajs: Are we also good with extensions and many regular expressions to match files?

lemmih: Oh,yeah, we'll need to update some regexes.

PR #3318, opened 2023-08-03
feat: implement `forest-tool benchmark export`

lemmih:
## Summary of changes

- Add an export benchmark command for easily measuring improvements.

Closes

## Change checklist

### DIFF
--- a/src/tool/subcommands/benchmark_cmd.rs
+++ b/src/tool/subcommands/benchmark_cmd.rs
@@ -1,14 +1,21 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use crate::chain::{
+    index::{ChainIndex, ResolveNullTipset},
+    ChainEpochDelta,
+};
 use crate::db::car::ManyCar;
 use crate::ipld::{stream_chain, stream_graph};
+use crate::shim::clock::ChainEpoch;
 use crate::utils::db::car_stream::CarStream;
 use anyhow::{Context as _, Result};
 use clap::Subcommand;
 use futures::{StreamExt, TryStreamExt};
 use indicatif::{ProgressBar, ProgressStyle};
+use std::ops::Deref;
 use std::path::PathBuf;
+use std::sync::Arc;
 use tokio::{
     fs::File,
     io::{AsyncWrite, AsyncWriteExt, BufReader},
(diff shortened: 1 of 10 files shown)


src/tool/subcommands/benchmark_cmd.rs
```diff
@@ -112,6 +128,36 @@ async fn benchmark_forest_encoding(
     Ok(())
 }
 
+// Exporting combines a graph traversal with ForestCAR.zst encoding. Ideally, it
+// should be no lower than `min(benchmark_graph_traversal,
+// benchmark_forest_encoding)`.
+async fn benchmark_exporting(
```
ruseinov, line 134: non-blocking: could be useful to have this bench support export without compression, to be able to detect bottlenecks easier.
In future it would also be nice to have benchmarks for the actual code we use, e.g. https://github.com/ChainSafe/forest/blob/25be5be685edbe49d92dd46ccbd7044dc1d64e23/src/chain/mod.rs#L16

Then we'd be able to compare the barebones benchmark to the full version and see if we have any significant deviations.

ruseinov (approved):
This looks good, some non-blocking comments.

src/tool/subcommands/benchmark_cmd.rs
lemmih: Isn't exporting without encoding a ForestCAR.zst file the same as a graph traversal?

src/tool/subcommands/benchmark_cmd.rs
```diff
@@ -112,6 +128,36 @@ async fn benchmark_forest_encoding(
...
+    input: Vec<PathBuf>,
+    compression_level: u16,
+    frame_size: usize,
+) -> Result<()> {
+    let store = open_store(input)?;
+    let heaviest = store.heaviest_tipset()?;
+    let stateroot_lookup_limit = heaviest.epoch() - 2000;
```
LesnyRumcajs, line 141: It might be worth commenting on where the 2000 value comes from.

LesnyRumcajs (approved):

src/tool/subcommands/benchmark_cmd.rs
lemmih: Oh yeah, this should be a command-line parameter.

src/tool/subcommands/benchmark_cmd.rs
ruseinov: yes, you are right.

hanabi1224 (approved):

PR #3319, opened 2023-08-03
fix: allow streaming of CAR.zst files with multiple frames

lemmih:
## Summary of changes

- multiple frames are disallowed by default but we need them for the ForestCAR.zst format.

Closes

## Change checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@
 
 ### Fixed
 
+- [#3319](https://github.com/ChainSafe/forest/pull/3319): Fix bug triggered by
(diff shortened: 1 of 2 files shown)


ruseinov (approved):
Looks good. Do we know if at affects anything rather than accepting multiple frames?

lemmih:
> Looks good. Do we know if at affects anything rather than accepting multiple frames?

It shouldn't. It just toggles whether the decoder ends after a single frame. The `async-compression` library enables it by default, so we didn't notice the problem before now.

LesnyRumcajs (approved):

Issue #3231, opened 2023-07-19
Move file-backed entries back into the database


LesnyRumcajs:
**Issue summary**

Currently, we store a few entries, like `GENESIS`, `HEAD` and `meta.yaml` outside of the database. After the DB re-organization, we have a dedicated column for such data.
```
share/forest/mainnet
❯ ls
GENESIS  HEAD  meta.yaml  paritydb/
```

It'd be great to move them into the DB and potentially get rid of the File-backed wrapper. To reduce the `SettingsStore` trait proliferation, it may make sense to pass to, e.g., `ChainStore` "another" DB in the form of a trait object. There may be other solutions.

**Other information and links**

PR #3320, opened 2023-08-03
Migrate settings to db

LesnyRumcajs:
## Summary of changes

- moved most of the file-backed entries to the database. There is one left, `DbIndex`, but having the DB govern itself doesn't seem right.
- This is a breaking change. We could have a 'migration' in theory. Happy to discuss if we want it.

Closes https://github.com/ChainSafe/forest/issues/3231

## Change checklist

### DIFF
--- a/src/chain/store/chain_store.rs
+++ b/src/chain/store/chain_store.rs
@@ -1,7 +1,7 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use std::{path::Path, sync::Arc};
+use std::sync::Arc;
 
 use crate::blocks::{BlockHeader, Tipset, TipsetKeys, TxMeta};
 use crate::interpreter::BlockMessages;
@@ -13,10 +13,7 @@ use crate::shim::{
     address::Address, econ::TokenAmount, executor::Receipt, message::Message,
     state_tree::StateTree, version::NetworkVersion,
 };
-use crate::utils::db::{
-    file_backed_obj::{ChainMeta, FileBacked},
-    BlockstoreExt, CborStoreExt,
-};
+use crate::utils::db::{BlockstoreExt, CborStoreExt};
 use ahash::{HashMap, HashMapExt, HashSet};
 use anyhow::Result;
 use cid::Cid;
@@ -34,6 +31,8 @@ use super::{
     Error,
 };
 use crate::chain::Scale;
+use crate::db::setting_keys::{ESTIMATED_RECORDS_KEY, HEAD_KEY};
+use crate::db::{SettingsStore, SettingsStoreExt};
 
 // A cap on the size of the future_sink
 const SINK_CAP: usize = 200;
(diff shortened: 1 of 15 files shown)


LesnyRumcajs:
> Shashank is away today and on Monday. If we want a quicker review, then we should assign someone else.

I know a volunteer when I see one!

lemmih:
> > Shashank is away today and on Monday. If we want a quicker review, then we should assign someone else.
> 
> I know a volunteer when I see one!

You dare use my own spell against me, Potter?

src/chain/store/chain_store.rs
```diff
@@ -107,57 +103,49 @@ where
...
-                chain_data_root.join("HEAD"),
-                || TipsetKeys::new(vec![*genesis_block_header.cid()]),
-                None,
-            )?;
-            let is_valid = chain_index.load_tipset(head_store.inner()).is_ok();
-            if !is_valid {
-                // If the stored HEAD is invalid, reset it to the genesis tipset.
-                head_store.set_inner(TipsetKeys::new(vec![*genesis_block_header.cid()]))?;
+
+        match settings.read_obj::<TipsetKeys>(HEAD_KEY)? {
```
lemmih, line 113: Optional golfing:

```rust
if !settings.read_obj::<TipsetKeys>(HEAD_KEY)?.is_some_and(|tipset_keys| chain_index.load_tipset(&tipset_keys).is_ok()) { 
  write genesis block as HEAD
}
```

src/chain/store/chain_store.rs
```diff
@@ -200,8 +193,14 @@ where
...
-        self.tipset_from_keys(self.file_backed_heaviest_tipset_keys.lock().inner())
-            .expect("Failed to load heaviest tipset")
+        self.tipset_from_keys(
+            &self
+                .settings
+                .read_obj::<TipsetKeys>(HEAD_KEY)
+                .expect("failed to load heaviest tipset")
+                .expect("failed to load heaviest tipset"),
+        )
+        .expect("failed to load heaviest tipset")
```
lemmih, line 203: Hmm, is there a way to reduce the number of duplicate error messages? Maybe we need a `require_obj` which returns `Err` instead of `None` when a key is missing.

src/db/parity_db.rs
```diff
@@ -44,20 +44,25 @@ impl DbColumn {
...
                         preimage: true,
                         // This is needed for key retrieval.
                         btree_index: true,
                         compression,
                         ..Default::default()
                     },
+                    DbColumn::Settings => parity_db::ColumnOptions {
+                        // explicitly disable preimage for settings column
+                        // othewise we are not able to overwrite entries
+                        preimage: false,
```
lemmih, line 62: Nice. Important change. :)

lemmih (approved):
Good. Once the GC is replaced, we can get rid of `FileBacked` altogether.

src/chain/store/chain_store.rs
LesnyRumcajs: I didn't know `is_some_and`. Good stuff!

src/chain/store/chain_store.rs
LesnyRumcajs: Splendid idea!

jdjaustin (approved):

PR #3326, opened 2023-08-04
ban tempfile default c-tor

LesnyRumcajs:
## Summary of changes

- banned the `NamedTempFile::new` method. While it is valid in contexts where we don't persist the data, it is difficult to enforce the usage in contexts where we actually need it. The issue surfaced a few times already (https://github.com/ChainSafe/forest/pull/2893#issuecomment-1576755240, https://github.com/ChainSafe/forest/pull/2693) and wasn't caught by the CI as the runners have a single fs.

Closes

https://docs.rs/tempfile/latest/tempfile/struct.TempPath.html#method.persist
>  Note: Temporary files cannot be persisted across filesystems. Also neither the file contents nor the containing directory are synchronized, so the update may not yet have reached the disk when persist returns.

## Change checklist

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -19,3 +19,7 @@ use parking_lot::RwLock instead to silently ignore panics."""
 path = "std::sync::Mutex"
 reason = """the standard library synchronization primitives are poisoned when aquiring threads panic.
 use parking_lot::Mutex instead to silently ignore panics."""
+
+[[disallowed-methods]]
+path = "tempfile::NamedTempFile::new"
+reason = """ The temporary files created by this method are not persistable if the temporary directory lives on a different filesystem than the target directory. While it is valid in other contexts (if not persisting files), it was misused many times and so we are banning it. Consider using `tempfile::NamedTempFile::new_in` or `tempfile::NamedTempFile::Builder"""
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -22,4 +22,4 @@ use parking_lot::Mutex instead to silently ignore panics."""
 
 [[disallowed-methods]]
 path = "tempfile::NamedTempFile::new"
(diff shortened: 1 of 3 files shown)


hanabi1224 (approved):

src/state_migration/tests/mod.rs
```diff
@@ -72,9 +72,10 @@ async fn test_state_migration(
     expected_new_state: Cid,
 ) -> Result<()> {
     // Car files are cached under data folder for Go test to pick up without network access
-    let car_path = format!("./src/state_migration/tests/data/{old_state}.car");
+    let car_path = PathBuf::from(format!("./src/state_migration/tests/data/{old_state}.car"));
     if !Path::new(&car_path).is_file() {
```
hanabi1224, line 76:
```suggestion
    if !car_path.is_file() {
```

.clippy.toml
```diff
@@ -19,3 +19,7 @@ use parking_lot::RwLock instead to silently ignore panics."""
 path = "std::sync::Mutex"
 reason = """the standard library synchronization primitives are poisoned when aquiring threads panic.
 use parking_lot::Mutex instead to silently ignore panics."""
+
+[[disallowed-methods]]
+path = "tempfile::NamedTempFile::new"
+reason = """ The temporary files created by this method are not persistable if the temporary directory lives on a different filesystem than the target directory. While it is valid in other contexts (if not persisting files), it was misused many times and so we are banning it. Consider using `tempfile::NamedTempFile::new_in` or `tempfile::NamedTempFile::Builder"""
```
aatifsyed, line 25:
```suggestion
reason = """The temporary files created by this method are not persistable if the temporary directory lives on a different filesystem than the target directory. While it is valid in other contexts (if not persisting files), it was misused many times and so we are banning it. Consider using `tempfile::NamedTempFile::new_in` or `tempfile::NamedTempFile::Builder"""
```

aatifsyed (approved):

PR #3327, opened 2023-08-04
stale: lotus json

aatifsyed:
See here for background: https://github.com/ChainSafe/forest/blob/8d038ca64cbe31ce1e947713bac3d57da396431f/src/lotus_json/mod.rs

The way forest currently handles the distinction between CBOR and JSON is to have a single struct represent
- the domain object
- the in memory representation
- the lotus cbor via `#[derive(Serialize, Deserialize)]`

This is largely the right decision - the [Serialize] and [Deserialize]
implementations of crates we depend on model the lotus cbor only.

However, the way we handle json is inconsistent:
- [Typically we create a `json` module for the domain object, for use with `serde(with = ...)`](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/token_amount.rs)
- [Sometimes we create a `FooJson` wrapper struct to wrap the deserialization](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/message_receipt.rs#L17)
- [Sometimes we create a `JsonHelper` struct for serde](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/signature.rs#L20-L25)
- Sometimes we create [different](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/blocks/header/json.rs#L37-L66) [structs](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/blocks/header/json.rs#L95-L124) for each serialization direction,
  where one typically wraps a reference.
- Typically we create additional [vec](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/cid.rs#L45-L78) and [opt](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/cid.rs#L80-L99) modules for domain objects which may be wrapped as `Vec<T>` and `Option<T>`

This PR introduces a `HasLotusJson` trait, which defines the one way to handle this.


aatifsyed:
#3339

PR #3339, opened 2023-08-08
feat: lotus_json for json interop handling

aatifsyed:
See here for background: https://github.com/ChainSafe/forest/blob/8d038ca64cbe31ce1e947713bac3d57da396431f/src/lotus_json/mod.rs

The way forest currently handles the distinction between CBOR and JSON is to have a single struct represent
- the domain object
- the in memory representation
- the lotus cbor via `#[derive(Serialize, Deserialize)]`

This is largely the right decision - the [Serialize] and [Deserialize]
implementations of crates we depend on model the lotus cbor only.

However, the way we handle json is inconsistent:
- [Typically we create a `json` module for the domain object, for use with `serde(with = ...)`](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/token_amount.rs)
- [Sometimes we create a `FooJson` wrapper struct to wrap the deserialization](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/message_receipt.rs#L17)
- [Sometimes we create a `JsonHelper` struct for serde](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/signature.rs#L20-L25)
- Sometimes we create [different](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/blocks/header/json.rs#L37-L66) [structs](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/blocks/header/json.rs#L95-L124) for each serialization direction,
  where one typically wraps a reference.
- Typically we create additional [vec](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/cid.rs#L45-L78) and [opt](https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/json/cid.rs#L80-L99) modules for domain objects which may be wrapped as `Vec<T>` and `Option<T>`

This PR introduces a `HasLotusJson` trait, which defines the one way to handle this.
It is so far only used in the `serialization-vectors` tests - adding new types for RPC calls etc is future work

### DIFF
--- a/src/blocks/header/lotus_json.rs
+++ b/src/blocks/header/lotus_json.rs
@@ -1,0 +1,162 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+use crate::lotus_json::*;
+use once_cell::sync::OnceCell;
+use serde::{Deserialize, Serialize};
+
+use super::BlockHeader;
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct BlockHeaderLotusJson {
+    miner: AddressLotusJson,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    ticket: Option<TicketLotusJson>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    election_proof: Option<ElectionProofLotusJson>,
+    beacon_entries: VecLotusJson<BeaconEntryLotusJson>,
+    win_po_st_proof: VecLotusJson<PoStProofLotusJson>,
+    parents: TipsetKeysLotusJson,
+    parent_weight: BigIntLotusJson,
+    height: i64,
+    parent_state_root: CidLotusJson,
+    parent_message_receipts: CidLotusJson,
+    messages: CidLotusJson,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    b_l_s_aggregate: Option<SignatureLotusJson>,
+    timestamp: u64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    block_sig: Option<SignatureLotusJson>,
+    fork_signaling: u64,
+    parent_base_fee: TokenAmountLotusJson,
+}
+
+impl HasLotusJson for BlockHeader {
+    type LotusJson = BlockHeaderLotusJson;
+
+    fn snapshots() -> Vec<(serde_json::Value, Self)> {
+        use serde_json::json;
+
+        vec![(
+            json!({
+                "BeaconEntries": null,
+                "Miner": "f00",
+                "Parents": null,
+                "ParentWeight": "0",
+                "Height": 0,
+                "ParentStateRoot": {
+                    "/": "baeaaaaa"
+                },
+                "ParentMessageReceipts": {
+                    "/": "baeaaaaa"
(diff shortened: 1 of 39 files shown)


src/blocks/header/lotus_json.rs
```diff
@@ -0,0 +1,162 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    block_sig: Option<SignatureLotusJson>,
+    fork_signaling: u64,
+    parent_base_fee: TokenAmountLotusJson,
+}
+
+impl HasLotusJson for BlockHeader {
+    type LotusJson = BlockHeaderLotusJson;
+
+    fn snapshots() -> Vec<(serde_json::Value, Self)> {
```
LesnyRumcajs, line 37: What is the purpose of this method?

src/blocks/header/lotus_json.rs
```diff
@@ -0,0 +1,162 @@
...
+use crate::lotus_json::*;
+use once_cell::sync::OnceCell;
+use serde::{Deserialize, Serialize};
+
+use super::BlockHeader;
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct BlockHeaderLotusJson {
+    miner: AddressLotusJson,
```
LesnyRumcajs, line 12: In general, why do you put `LotusJson` everywhere? Is it so distinct from a regular `Json`?

src/blocks/header/lotus_json.rs
aatifsyed: It's for testing our de/serialization: https://jestjs.io/docs/snapshot-testing
This makes sure every struct has a test

src/blocks/header/lotus_json.rs
aatifsyed: This is a piecewise upgrade of our codebase - I don't want there to be two different `ActorStateJson`s.
I could rename to `FilJson` if this is a blocker, but going to leave for now 🤷

src/blocks/header/lotus_json.rs
LesnyRumcajs: Shouldn't it be only visible under tests?

aatifsyed: We _could_ do that, but it would require a second trait, or a `#[cfg(test)]` under the trait definition, both of which require more verbosity for the implementor - I want the fast path to be a quick as possible for users - one trait to implement, and they get tested etc for free

LesnyRumcajs: For me, it's a bit confusing if I'd see something like this in method suggestions. Let's mark it somehow that it shouldn't be used outside tests.

aatifsyed: I played around with adding `#[cfg(test)]` for this, but it kills e.g autocomplete - traits are not expected to have `#[cfg(...)]` members.

I'm keeping this as is for ease of use, but I've improved the documentation for this method.

PR #3343, opened 2023-08-08
database migration plan

LesnyRumcajs:
## Summary of changes

- added a proposed database migration plan. I'm open to discussing it.

Closes

## Change checklist

### DIFF
--- a/documentation/src/SUMMARY.md
+++ b/documentation/src/SUMMARY.md
@@ -17,6 +17,7 @@
 - [Developer documentation](./developer_documentation/introduction.md)
   - [Application architecture](./developer_documentation/application_architecture.md)
   - [Contributing](./developer_documentation/contributing.md)
+  - [Database migrations](./developer_documentation/database_migrations.md)
   - [Local GH Actions](./developer_documentation/local_actions.md)
   - [Mainnet compatibility](./developer_documentation/mainnet_compatibility.md)
   - [Memory analysis](./developer_documentation/memory-analysis.md)
--- a/documentation/src/developer_documentation/database_migrations.md
+++ b/documentation/src/developer_documentation/database_migrations.md
@@ -1,0 +1,185 @@
+# Database migrations
+
+## Problem
+
+Storing data is one of the prime purposes of Forest, and we often play around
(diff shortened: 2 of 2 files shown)


lemmih (approved):
I like it. Notes:

- I'm okay with ignoring databases that are newer than the current version of Forest. But I guess if we can open the database and the checks all pass, then it's probably fine.
- Creating a new database if something goes wrong sounds good. Starting from a clean snapshot is almost always better than failing.
- The development mode looks good.

hanabi1224:
It's a bit weird to me to use the forest version as DB version since forest has more features than its DB and I would expect most of the version bumps are not DB related.  How about maintaining another DB version?  The CI check in the proposal could still catch when breaking changes are made to the DB but the version is not updated. For local development, one could freely change the DB until they bump the DB version and add a migration script

lemmih:
I'm in favor of _not_ having a separate database version. It's easy to have useless version information. For example, actor and network versions are separate, but there's no benefit to this; it just causes more complexity. I would much rather use Forest versions for the database and have explicit null migrations between versions that do not change the DB layout. The sequence of migrations to get from version A to version B is still linear and unambiguous.

LesnyRumcajs:
> Also a dedicated DB version makes it clear that every DB version bump should have a corresponding migration script, migrating a super old DB from v2 to v10 means running migration v3 to v10 in sequence. It's not very clear to me how to achieve this in the current plan

I tried to portray this in the [Checking if migration exists](https://github.com/ChainSafe/forest/blob/5d9ffaef3042f72096650b76ce5a3fe35aa4f6f4/documentation/src/developer_documentation/database_migrations.md#checking-if-migration-exists) section. While we may need a sequential upgrade (which could take a long time), we could have "shortcut" migrations in our migrations table. So a possible route would be `1--2--5--9--10`. A simple graph traversal algorithm would suffice.

hanabi1224 (approved):
@LesnyRumcajs Sounds good

LesnyRumcajs:
I see no other objections; I'll move with the outlined plan to the implementation phase.

PR #3345, opened 2023-08-08
feat: switch to lotus_json for some types

aatifsyed:
### Changes to `HasLotusJson`
- Make all `ActorStateLotusJson` etc types private.
- Introduce a single `LotusJson<T>` type, which covers lotus (de) serialization.
- Change pattern for `HasLotusJson::LotusJson` types: just transform fields to `LotusJson<T>` - this could be `#[derive(...)]`-ed in the future
- Change `HasLotusJson` bounds so that we have more control over types (rather than requiring `From` and `Into`)

### Migrations to using `HasLotusJson`
- `actor_state`
- `signed_message`
- `gossip_block`
- `sync_state`
- `tipset_json`
- `tipset_keys_json`
- ...


src/lotus_json/receipt.rs
```diff
@@ -0,0 +1,66 @@
...
+        })
+    }
+}
+
+#[test]
+fn shapshots() {
+    assert_all_snapshots::<Receipt>()
+}
+
+// #[cfg(test)]
```
elmattic, line 63: Maybe uncommenting test?

aatifsyed: https://github.com/ChainSafe/forest/pull/3345/commits/1c97af97416c4d095d836e6a22c5d3aadb661547

elmattic (approved):

src/lotus_json/mod.rs
```diff
@@ -289,12 +305,70 @@ pub mod base64_standard {
     }
 }
 
+/// MUST NOT be used in any `LotusJson` structs
```
LesnyRumcajs, line 308: I guess without reflection we won't be able to ensure this?

src/lotus_json/receipt.rs
```diff
@@ -0,0 +1,104 @@
...
+    assert_all_snapshots::<Receipt>()
+}
+
+/// [Receipt] knows if it is `V2` or `V3`, but there's no way for
+/// the serialized representation to retain that information,
+/// so [`assert_unchanged_via_json`] tests with arbitrary input will fail.
+///
+/// This can only be fixed by rewriting [Receipt].
+#[test]
+#[should_panic]
```
LesnyRumcajs, line 70: Can we avoid this? I mean, not unwrap the results but check if they are errors? Otherwise this test will pass on any panic (even if it happens before any assertions)

src/rpc_api/mod.rs
```diff
@@ -1,12 +1,12 @@
...
-/// implementation of Filecoin). Current progress is tracked in
-/// `ARCHITECTURE.md`.
-///
-/// Follow the pattern set below, and don't forget to add an entry to the
-/// [`ACCESS_MAP`] with the relevant permissions (consult the go implementation,
-/// looking for a comment like `// perm: admin`)
+//! In general, `forest` wants to support the same RPC messages as `lotus` (go
+//! implementation of Filecoin). Current progress is tracked in
+//! `ARCHITECTURE.md`.
+//!
```
LesnyRumcajs, line 6: this didn't age well
```suggestion
```

LesnyRumcajs (approved):

src/lotus_json/message.rs
```diff
@@ -43,65 +45,60 @@ impl HasLotusJson for Message {
...
             value,
-            method_num,
-            params,
             gas_limit,
             gas_fee_cap,
             gas_premium,
-        } = value;
+            method,
+            params,
+            cid: _ignored, // BUG?(aatifsyed)
```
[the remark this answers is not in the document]
aatifsyed: https://github.com/ChainSafe/forest/issues/3389

src/lotus_json/mod.rs
aatifsyed: Yep, or a custom linter

PR #3352, opened 2023-08-10
fix: transfer settings store data during GC

hanabi1224:
## Summary of changes

@LesnyRumcajs let me know if I'm wrong but I did not find the logic that transfers settings to the new database during garbage collection in https://github.com/ChainSafe/forest/pull/3320. This PR tries to fix the issue

- transfer settings store data during GC

Closes

## Change checklist

### DIFF
--- a/src/db/rolling/impls.rs
+++ b/src/db/rolling/impls.rs
@@ -171,7 +171,10 @@ impl RollingDB {
         db_index_inner_mut.current = new_db_name;
         db_index_inner_mut.current_creation_epoch = current_epoch;
         db_index.sync()?;
+
         delete_db(&old_db_path);
+
+        self.transfer_settings()?;
 
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


lemmih (changes requested):
Keeping a static list of all settings keys is not a robust solution. Let's sunset the GC instead.

hanabi1224:
> Keeping a static list of all settings keys is not a robust solution. Let's sunset the GC instead.

@lemmih  As discussed on Slack, changed to enumerate DB keys instead of keeping a static list

LesnyRumcajs (approved):

LesnyRumcajs:
It might be temporary, but let's have some tests for this.

src/db/rolling/impls.rs
```diff
@@ -238,14 +264,39 @@ fn delete_db(db_path: &Path) {
...
     use rand::Rng;
     use tempfile::TempDir;
 
     use super::*;
+    use crate::libp2p_bitswap::BitswapStoreRead;
+
+    #[quickcheck]
+    fn ensure_settings_are_transferred(keys: Vec<String>) -> Result<()> {
+        let db_root = TempDir::new()?;
+        println!("Creating rolling db under {}", db_root.path().display());
```
lemmih, line 281: This debug println can be removed.

lemmih (changes requested):
Updating `Cargo.lock` doesn't appear to be necessary. Let's leave those changes out.

src/db/rolling/impls.rs
hanabi1224: Removed

hanabi1224:
> Updating `Cargo.lock` doesn't appear to be necessary. Let's leave those changes out.

@lemmih removed

lemmih (approved):

PR #3353, opened 2023-08-10
feat: impl Blockstore::has for ForestCar and AnyCar

hanabi1224:
## Summary of changes

`Blockstore::has` implementation can be more efficient than the default impl (`Blockstore::get().is_some()`) by leveraging the index
 
- impl Blockstore::has for ForestCar and AnyCar

Closes

## Change checklist

### DIFF
--- a/src/db/car/any.rs
+++ b/src/db/car/any.rs
@@ -108,6 +108,14 @@ impl<ReaderT> Blockstore for AnyCar<ReaderT>
 where
     ReaderT: ReadAt,
 {
+    fn has(&self, k: &Cid) -> anyhow::Result<bool> {
+        match self {
+            AnyCar::Forest(forest) => forest.has(k),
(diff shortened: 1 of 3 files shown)


lemmih (changes requested):
I like the idea, but there's a bug in the code. Looking up a key in `self.indexed` will return an array of _possible_ matches. Those matches still have to be checked against the CID in the CAR block.

hanabi1224:
> I like the idea, but there's a bug in the code. Looking up a key in `self.indexed` will return an array of _possible_ matches. Those matches still have to be checked against the CID in the CAR block.

@lemmih Updated to match the logic in `get` as much as possible.  One question: may I assume `None` branch means `true` as I see below comment in `get`
```rust
// This lookup only fails in case of a hash collision
if let Some(value) = get_result {
    return Ok(Some(value));
}
```

PR #3362, opened 2023-08-11
`forest-cli wallet` subcommands migration

elmattic:
## Summary of changes

- Moved the following `forest-cli wallet` subcommands to `forest-wallet`

Closes

## Change checklist


src/wallet/main.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::shim::address::{CurrentNetwork, Network};
+use clap::Parser;
+use std::str::FromStr;
+
+pub fn main<ArgT>(args: impl IntoIterator<Item = ArgT>) -> anyhow::Result<()>
+where
+    ArgT: Into<OsString> + Clone,
+{
+    // Capture Cli inputs
+    let Cli { opts, cmd } = Cli::parse_from(args);
```
lemmih, line 18: The command-line options are not appropriate for the wallet tool.

lemmih: Oh, I see, this `Cli` is different from the other `Cli` structures. All good, then.

jdjaustin (approved):

src/cli_shared/cli/mod.rs
```diff
@@ -231,6 +231,18 @@ impl CliOpts {
     }
 }
 
+/// CLI RPC options
+#[derive(Default, Debug, Parser)]
+pub struct CliRpcOpts {
+    /// Choose network chain to connect to
+    #[arg(long)]
+    pub chain: Option<NetworkChain>,
```
lemmih, line 239: This `chain` field is not needed. In the short-term, the wallet commands can query the network from the forest-node. In the long-term (when `forest-wallet` starts managing the keys separately from the node), we need to tag the keys with their testnet status.

lemmih (changes requested):

src/cli_shared/cli/mod.rs
elmattic: Ok, i was hesitant at removing it.

src/wallet/main.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()?
+        .block_on(async {
+            let chain = if let Ok(name) = chain_get_name((), &opts.token).await {
```
lemmih, line 24: Let's fail if we cannot get the chain name from the node.

lemmih (changes requested):

lemmih (approved):

LesnyRumcajs:
I don't quite understand. What benefit does moving `forest-cli wallet` to `forest-wallet` bring? I mean, the keys are still governed by the Forest node. To my understanding, the goal was to have better key security by decoupling it completely from the Forest node. See https://github.com/ChainSafe/forest/discussions/2274

elmattic:
> I don't quite understand. What benefit does moving `forest-cli wallet` to `forest-wallet` bring? I mean, the keys are still governed by the Forest node. To my understanding, the goal was to have better key security by decoupling it completely from the Forest node. See #2274

I think it's the end goal. And moving commands to a new standalone binary is a first step.

Issue #3334, opened 2023-08-07
Store ForestCAR.zst files directly in the database


lemmih:
**Issue summary**

When importing a CAR, CAR.zst, or ForestCAR.zst file, we should convert it to ForestCAR.zst and place it directly in the database folder. ForestCAR.zst files are smaller and faster than ParityDB. They're read-only and dead information cannot be garbage collected, but that matters little compared to the benefits of using compressed data.

Acceptance criteria:
- [x] Convert imported CAR and CAR.zst files into ForestCAR.zst. ForestCAR.zst files obviously do not have to be re-converted. Move the file to the database folder.
- [x] Opening a database should give a block store that is the union of ParityDB and all available ForestCAR.zst files.
- [ ] Add flag `--consume-snapshot`. It should do the same as `--import-snapshot`, but it'll move or delete the snapshot.

**Other information and links**

PR #3365, opened 2023-08-14
feat: serve forest CAR file as Blockstore

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/3334

-

Closes

## Change checklist

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -13,12 +13,12 @@ use crate::cli_shared::{
     cli::{CliOpts, Config},
     snapshot,
 };
-use crate::db::car::ManyCar;
+use crate::db::car::{AnyCar, ManyCar};
 use crate::db::{
     db_engine::{db_root, open_proxy_db},
     rolling::DbGarbageCollector,
 };
-use crate::genesis::{get_network_name_from_genesis, import_chain, read_genesis_header};
+use crate::genesis::{get_network_name_from_genesis, read_genesis_header};
 use crate::key_management::{
     KeyStore, KeyStoreConfig, ENCRYPTED_KEYSTORE_NAME, FOREST_KEYSTORE_PHRASE_ENV,
 };
@@ -32,6 +32,9 @@ use crate::shim::{
     version::NetworkVersion,
 };
 use crate::state_manager::StateManager;
+use crate::utils::db::car_stream::CarStream;
+use crate::utils::io::random_access::RandomAccessFile;
+use crate::utils::net;
 use crate::utils::{
     monitoring::MemStatsTracker, proofs_api::paramfetch::ensure_params_downloaded, retry,
     version::FOREST_VERSION_STRING, RetryArgs,
@@ -39,10 +42,11 @@ use crate::utils::{
 use anyhow::{bail, Context};
 use bundle::load_actor_bundles;
 use dialoguer::{console::Term, theme::ColorfulTheme};
-use futures::{select, Future, FutureExt};
+use futures::{select, Future, FutureExt, TryStreamExt};
 use lazy_static::lazy_static;
 use raw_sync::events::{Event, EventInit as _, EventState};
 use shared_memory::ShmemConf;
+use std::fs;
 use std::{
     cell::RefCell,
     net::TcpListener,
@@ -52,6 +56,7 @@ use std::{
     time::Duration,
 };
 use tempfile::{Builder, TempPath};
+use tokio::io::AsyncWriteExt;
 use tokio::{
     signal::{
         ctrl_c,
@@ -174,10 +179,32 @@ pub(super) async fn start(
     let keystore = Arc::new(RwLock::new(keystore));
 
     let chain_data_path = chain_path(&config);
-    let db = Arc::new(ManyCar::new(Arc::new(open_proxy_db(
-        db_root(&chain_data_path),
-        config.db_config().clone(),
-    )?)));
+    let (db, heaviest_tipset_from_imported_snapshot) = {
+        let mut heaviest_tipset = None;
+        let db_root_dir = db_root(&chain_data_path);
+        if !db_root_dir.is_dir() {
+            fs::create_dir_all(&db_root_dir)?;
+        }
+        let forest_car_db_path = db_root_dir.join("snapshot.forest.car.zst");
+
+        if let Some(path) = &config.client.snapshot_path {
+            heaviest_tipset = Some(
+                import_chain_as_forest_car(&path.display().to_string(), &forest_car_db_path)
+                    .await?,
+            );
+        }
+
+        let mut store = ManyCar::new(Arc::new(open_proxy_db(
+            db_root_dir,
+            config.db_config().clone(),
+        )?));
+
+        if forest_car_db_path.is_file() {
+            store.read_only_files(std::iter::once(forest_car_db_path))?;
+        }
+
+        (Arc::new(store), heaviest_tipset)
+    };
 
     let mut services = JoinSet::new();
 
@@ -224,6 +251,10 @@ pub(super) async fn start(
         config.chain.clone(),
         genesis_header.clone(),
     )?);
+
+    if let Some(ts) = heaviest_tipset_from_imported_snapshot {
+        chain_store.set_heaviest_tipset(ts.into())?;
+    }
 
     let db_garbage_collector = {
         let db = db.clone();
@@ -401,20 +432,6 @@ pub(super) async fn start(
 
     let mut config = config;
     fetch_snapshot_if_required(&mut config, epoch, opts.auto_download_snapshot).await?;
-
-    if let Some(path) = &config.client.snapshot_path {
-        let stopwatch = time::Instant::now();
-        import_chain(
-            &state_manager,
-            &path.display().to_string(),
-            config.client.skip_load,
-            config.client.chunk_size,
-            config.client.buffer_size,
-        )
-        .await
-        .context("Failed miserably while importing chain from snapshot")?;
-        info!("Imported snapshot in: {}s", stopwatch.elapsed().as_secs());
-    }
 
     if let (true, Some(validate_from)) = (config.client.snapshot, config.client.snapshot_height) {
         // We've been provided a snapshot and asked to validate it
@@ -707,6 +724,57 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
             "Error: the passwords do not match. Try again or press Ctrl+C to abort.",
         )
         .interact_on(&term)
+}
+
+async fn import_chain_as_forest_car(
+    from_path: &str,
+    forest_car_path: &Path,
+) -> anyhow::Result<Tipset> {
+    info!("Importing chain from snapshot at: {from_path}");
+
+    let stopwatch = time::Instant::now();
+    let mut reader = net::reader(from_path).await?;
+    let forest_car_db_temp_path = tempfile::NamedTempFile::new_in(
+        forest_car_path
+            .parent()
+            .context("`forest_car_path` should have a parent directory.")?,
+    )?
+    .into_temp_path();
+    let mut tmp_snapshot_writer =
+        tokio::io::BufWriter::new(tokio::fs::File::create(&forest_car_db_temp_path).await?);
+    tokio::io::copy(&mut reader, &mut tmp_snapshot_writer).await?;
+
+    let (is_forest_car, ts) = {
+        let car = AnyCar::new(RandomAccessFile::open(&forest_car_db_temp_path)?)?;
+        let ts = car.heaviest_tipset()?;
+        (
+            match car {
+                AnyCar::Forest(_) => true,
+                _ => false,
+            },
+            ts,
+        )
+    };
+    if is_forest_car {
+        forest_car_db_temp_path.persist(forest_car_path)?;
+    } else {
+        let car_stream = CarStream::new(tokio::io::BufReader::new(
+            tokio::fs::File::open(&forest_car_db_temp_path).await?,
+        ))
+        .await?;
+        let roots = car_stream.header.roots.clone();
+        let mut writer = tokio::io::BufWriter::new(tokio::fs::File::create(forest_car_path).await?);
+        let frames = crate::db::car::forest::Encoder::compress_stream(
+            8000usize.next_power_of_two(),
+            zstd::DEFAULT_COMPRESSION_LEVEL as _,
+            car_stream.map_err(anyhow::Error::from),
+        );
+        crate::db::car::forest::Encoder::write(&mut writer, roots, frames).await?;
+        writer.flush().await?;
+    }
+
+    info!("Imported snapshot in: {}s", stopwatch.elapsed().as_secs());
+    Ok(ts)
 }
 
 #[cfg(test)]


src/db/car/forest.rs
```diff
@@ -72,6 +72,8 @@ use std::{
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use tokio_util::codec::{Decoder, Encoder as _};
 
+pub const FOREST_CAR_FILE_EXTENSION: &str = ".forest.car.zst";
```
LesnyRumcajs, line 75: Could we use this constant where we have other occurrences of this, e.g., `snapshot_cmd`?

src/daemon/mod.rs
```diff
@@ -709,14 +708,168 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
         .interact_on(&term)
 }
 
+pub async fn open_forest_car_union_db(
```
LesnyRumcajs, line 711: I'd try to not bloat this file any further, it's already quite big. Maybe let's have a dedicated one?

LesnyRumcajs, line 711: Is this method unit-tested?

src/daemon/mod.rs
```diff
@@ -171,11 +180,12 @@ pub(super) async fn start(
...
     let keystore = Arc::new(RwLock::new(keystore));
 
-    let chain_data_path = chain_path(&config);
-    let db = Arc::new(ManyCar::new(Arc::new(open_proxy_db(
-        db_root(&chain_data_path),
-        config.db_config().clone(),
-    )?)));
+    if opts.detach {
+        unblock_parent_process()?;
+    }
```
lemmih, line 185: Detaching before the RPC is available looks wrong. Why did you change this?

src/daemon/mod.rs
```diff
@@ -506,9 +505,9 @@ async fn fetch_snapshot_if_required(
             })
             .await;
             if !have_permission {
-                bail!("Forest requires a snapshot to sync with the network, but automatic fetching is disabled.")
+                std::process::exit(0);
```
lemmih, line 508: Why?

src/daemon/mod.rs
```diff
@@ -695,14 +694,170 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
         .interact_on(&term)
 }
 
+pub async fn open_forest_car_union_db(
```
lemmih, line 697: I don't understand what this function is supposed to do. It appears that it does several unrelated things (like opening the current database AND handling snapshot importing). Could you write some documentation for it?

src/daemon/mod.rs
```diff
@@ -695,14 +694,170 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
...
+                    .file_name()
+                    .and_then(OsStr::to_str)
+                    .context("Infallible getting file name")?,
+            )
+            .await?;
+        }
+    }
+
+    let (is_forest_car, ts) = {
+        let car = AnyCar::new(RandomAccessFile::open(&downloaded_car_temp_path)?)?;
```
lemmih, line 818: This is very expensive and can take hours and hundreds of GiBs of memory.

src/daemon/mod.rs
```diff
@@ -695,14 +694,170 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
...
+        let ts = car.heaviest_tipset()?;
+        (matches!(car, AnyCar::Forest(_)), ts)
+    };
+
+    let forest_car_db_path =
+        forest_car_db_dir.join(format!("{}{FOREST_CAR_FILE_EXTENSION}", ts.epoch()));
```
lemmih, line 824: This could clash with other files.

src/daemon/mod.rs
```diff
@@ -695,14 +694,170 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
...
+        match AnyCar::new(RandomAccessFile::open(&file)?) {
+            Ok(car) => {
+                if matches!(car, AnyCar::Forest(_)) {
+                    store.read_only(car);
+                    info!("Loaded car DB at {}", file.display());
+                } else {
+                    warn!(
+                        "Skip loading car DB at {}: invalid .forest.car.zst format",
+                        file.display()
+                    );
```
lemmih, line 737: A corrupted database is an error. Do note that the ForestCAR.zst format is not fixed, and re-encoding may be required.

src/daemon/mod.rs
```diff
@@ -695,14 +694,170 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
...
+        };
+    }
+
+    // TODO: use `--consume-snapshot` CLI option once it's implemented
+    let mut consume_snapshot_file = false;
+    if config.client.snapshot_path.is_none() {
+        let epoch = {
+            if store.read_only_len() == 0 {
+                0
+            } else if let Ok(Some(ts)) = Tipset::load_heaviest(&store, store.writer().as_ref()) {
```
lemmih, line 750:
```suggestion
            } else if let Ok(Some(ts)) = Tipset::load_heaviest(&store, &store) {
```

src/daemon/mod.rs
```diff
@@ -695,14 +694,170 @@ fn create_password(prompt: &str) -> std::io::Result<String> {
...
+            heaviest_tipset = Some(ts);
+            store.read_only_files(std::iter::once(car_db_path.clone()))?;
+            info!("Loaded car DB at {}", car_db_path.display());
+        }
+    }
+
+    Ok((Arc::new(store), heaviest_tipset))
+}
+
+async fn import_chain_as_forest_car(
```
lemmih, line 774: This function needs to be documented and the code could benefit from significant refactoring.

lemmih (commented):
This PR contains a lot of changes that I need help to understand. Let's hold off merging this until the code is well-documented and intuitive.

src/daemon/mod.rs
hanabi1224: Moved to `db_util.rs`

src/daemon/mod.rs
hanabi1224: Fixed

src/daemon/mod.rs
hanabi1224: Reverted. This was a temp change to fix daemon e2e tests, moving `exit_after_init` check to before snapshot import instead

src/daemon/mod.rs
hanabi1224: In the current implementation, to avoid creating a wrapper type of `Mutex<ManyCar>` that does not require `&mut self` for APIs like `fn read_only`, snapshot import is moved to before DB initialization, thus any functions that require a mutable `ManyCar` is moved into this function(snapshot import, auto download snapshot).

src/daemon/mod.rs
hanabi1224: If it's the same snapshot, I think it should not matter. If it's a diff car file, I wanted to name it `{from_epoch}-{to_epcoh}.forest.car.zst`, and I'd like to discuss about injecting the epoch range info to the diff car header. Basically 

```rust
let cid = store.put((TipsetKeys, TipsetKeys, GenesisTipsetKeys))
output_car.roots = vec![cid]
```

src/daemon/mod.rs
hanabi1224: `SettingsStore` is not implemented for `ManyCar`, do you think we should implement it?

lemmih: Yes.

src/db/car/forest.rs
hanabi1224: Yes, I will do this in a follow-up PR to not bloat this one

src/daemon/mod.rs
hanabi1224: Fixed

src/daemon/mod.rs
hanabi1224: Moved TCP endpoint binding logic to before calling `unblock_parent_process`

src/daemon/mod.rs
hanabi1224: Changed to bail when there are errors

src/daemon/mod.rs
hanabi1224: Changed filename to {unix_timestap}.forest.car.zst

src/daemon/mod.rs
hanabi1224: Doc added. Could you elaborate on the refactoring part

src/daemon/mod.rs
hanabi1224: Doc added

src/daemon/db_util.rs
```diff
@@ -0,0 +1,343 @@
...
+use tokio::io::AsyncWriteExt;
+use tracing::{error, info, warn};
+use url::Url;
+use walkdir::WalkDir;
+
+/// This function tries to open the forest database directory as [`ManyCar<Arc<RollingDB>>`], it
+/// 1. loads `parity-db`
+/// 2. loads all existing CAR files in `.forest.car.zst` format
+/// 3. asks to fetch the latest snapshot when it's required to run forest.
+/// 4. imports the snapshot(from either CLI options or step 3) and stores it to the database in `.forest.car.zst` format
```
lemmih, line 39: I don't see why all of these steps have to be done in a single function. Loading CAR files from disk seems to be completely separate from ParityDB, user prompting, and file downloading. Let's separate it out.

src/daemon/db_util.rs
```diff
@@ -0,0 +1,343 @@
...
+                info!("Loaded car DB at {}", file.display());
+            }
+            Err(err) => {
+                has_errors = true;
+                error!("Error loading car DB at {}: {err}", file.display())
+            }
+        };
+    }
+
+    // Let it check all DB files altogether instead of failing 1 by 1.
```
lemmih, line 85: I'd prefer bailing early. Corrupted files aren't a common occurance.

src/daemon/db_util.rs
```diff
@@ -0,0 +1,343 @@
...
+            info!("Loaded car DB at {}", car_db_path.display());
+        }
+    }
+
+    Ok((Arc::new(store), heaviest_tipset))
+}
+
+/// This function validates and stores the CAR binary from `from_path`(either local path or URL) into the `{DB_ROOT}/car_db/`
+/// (automatically trans-code into `.forest.car.zst` format when needed), and returns its final file path and the heaviest tipset.
+async fn import_chain_as_forest_car(
```
lemmih, line 122: Let's refactor this function.

- If `from_path` is a URL, download it to `downloaded_car_temp_path`.
- Add a utility function that copies if renaming fails. This should reduce lines 131-163 to two or three lines.
- Hoist out the code that converts from CAR to ForestCAR.zst.

src/daemon/db_util.rs
```diff
@@ -0,0 +1,343 @@
...
+    // So we need a snapshot (which will be from a recent network version)
+    let require_a_snapshot = network_version_is_small;
+    let have_a_snapshot = config.client.snapshot_path.is_some();
+
+    match (require_a_snapshot, have_a_snapshot, auto_download_snapshot) {
+        (false, _, _) => Ok(()),   // noop - don't need a snapshot
+        (true, true, _) => Ok(()), // noop - we need a snapshot, and we have one
+        (true, false, true) => {
+            // we need a snapshot, don't have one, and have permission to download one, so do that
+            let max_retries = 3;
```
lemmih, line 244: We retry three times here but don't do any retries in the other branch? Let's refactor to remove the duplicate code.

src/daemon/mod.rs
```diff
@@ -171,11 +161,12 @@ pub(super) async fn start(
...
     let keystore = Arc::new(RwLock::new(keystore));
 
-    let chain_data_path = chain_path(&config);
-    let db = Arc::new(ManyCar::new(Arc::new(open_proxy_db(
-        db_root(&chain_data_path),
-        config.db_config().clone(),
-    )?)));
+    if opts.exit_after_init {
+        return Ok(());
+    }
```
lemmih, line 166: Why did you move this?

src/daemon/mod.rs
```diff
@@ -437,89 +410,6 @@ pub(super) async fn start(
         .map(|_| {})
 }
 
-/// If our current chain is below a supported height, we need a snapshot to bring it up
```
lemmih, line 440: Why did you move this into `db_util.rs`?

src/daemon/db_util.rs
```diff
@@ -0,0 +1,343 @@
...
+        bail!("Error loading DB file(s).");
+    }
+
+    // TODO: use `--consume-snapshot` CLI option once it's implemented
+    let mut consume_snapshot_file = false;
+    if config.client.snapshot_path.is_none() {
+        let epoch = {
+            if store.read_only_len() == 0 {
+                0
+            } else if let Ok(Some(ts)) = Tipset::load_heaviest(&store, &store) {
```
lemmih, line 96: This looks wrong. Can you explain this code to me?

lemmih (changes requested):

src/daemon/mod.rs
hanabi1224: All `daemon` integration tests rely on this flag to not fail because they don't provide a snapshot. I have to either move this early or do `process::exit(0)` when there's no permission to download a snapshot, see https://github.com/ChainSafe/forest/pull/3365#discussion_r1299220428

lemmih: Oh, I see. You had to move this check earlier because you moved the snapshot downloading into the DB opening code. Hmm. We can clean things up once we've refactored the DB opening code.

src/daemon/db_util.rs
hanabi1224: The idea is that when there are no CAR files loaded(manually deleted or whatever), the union DB should be treated as corrupted. Otherwise, we load the heaviest tipset from the store as normal

hanabi1224: Alternatively, I can inspect all tipsets back to genesis but that takes longer, is there a better way of achieving this?

```rust
async fn verify_tipsets_integrity(
    store: &impl Blockstore,
    from: Tipset,
    config: &Config,
) -> anyhow::Result<bool> {
    let start = Utc::now();
    info!("Verifying the database integrity...");
    let mut current = from;
    while current.epoch() > 0 {
        if let Some(parent) = Tipset::load(store, current.parents())? {
            current = parent;
        } else {
            break;
        }
    }

    let result = if current.epoch() > 0 {
        false
    } else {
        let genesis_header = read_genesis_header(
            config.client.genesis_file.as_ref(),
            config.chain.genesis_bytes(),
            store,
        )
        .await?;
        if let Some(genesis_ts) = Tipset::load(
            store,
            &TipsetKeys::new(FrozenCids::from_iter([*genesis_header.cid()])),
        )? {
            current == genesis_ts
        } else {
            false
        }
    };

    info!(
        "Done verifying the database integrity, took {}s",
        (Utc::now() - start).num_seconds()
    );

    Ok(result)
}
```

src/daemon/mod.rs
hanabi1224: It is only used for DB initialization. I can move it back to `mod.rs` or a separate `snapshot_util.rs` to not bloat `mod.rs`, let me know which is preferrable

src/daemon/db_util.rs
hanabi1224: I can refactor this with a separate PR, as the code is not part of this PR (only moved this function)

src/daemon/db_util.rs
hanabi1224: Fixed

src/daemon/db_util.rs
lemmih: Okay, let's move it back to the original location, then. If you don't touch the code, you don't have to fix it. If you touch the code, you fix it. :)

hanabi1224: Moved back, the retry lines are no longer part of diff

src/daemon/mod.rs
hanabi1224: Moved back, see https://github.com/ChainSafe/forest/pull/3365#discussion_r1303987435

src/daemon/db_util.rs
hanabi1224: The goal of this function is to hide all these steps to `fn start` in `mod.rs`, which could DB with 1-liner

hanabi1224: Are you suggesting making all these steps a separate function and use them all in `mod.rs` or still having a wrapper fn here in `db_util.rs`?

lemmih: Either should be fine to start with. Once the unrelated tasks have been separated into distinct functions then we can do more refactoring.

src/daemon/db_util.rs
hanabi1224: Fixed

src/daemon/db_util.rs
hanabi1224: Fixed

src/daemon/mod.rs
hanabi1224: `test_prepare_and_open_forest_car_union_db` added

src/daemon/db_util.rs
```diff
@@ -0,0 +1,190 @@
...
+
+fn move_or_copy_file(from: &Path, to: &Path, prefer_move: bool) -> anyhow::Result<()> {
+    if prefer_move && fs::rename(from, to).is_ok() {
+        return Ok(());
+    } else {
+        fs::copy(from, to)?;
+    }
+
+    Ok(())
+}
```
lemmih, line 126:
```suggestion
fn move_or_copy_file(from: &Path, to: &Path, consume: bool) -> io::Result<()> {
    if consume && fs::rename(from, to).is_ok() {
        Ok(())
    } else {
        fs::copy(from, to)?;
        if consume { fs::remove_file(from)? }
    }
}
```

src/daemon/db_util.rs
```diff
@@ -0,0 +1,190 @@
...
+                if let Some(filename) = entry.file_name().to_str() {
+                    if filename.ends_with(FOREST_CAR_FILE_EXTENSION) {
+                        return Some(entry.into_path());
+                    }
+                }
+            }
+            None
+        })
+    {
+        match ForestCar::new(Mmap::map_path(&file)?) {
```
lemmih, line 44: The use of mmap should be governed by the environment variable.

lemmih: Also, you can just use `.context()` here.

src/db/car/many.rs
```diff
@@ -18,19 +18,19 @@ use cid::Cid;
...
+    read_only: Mutex<Vec<AnyCar<Box<dyn super::RandomAccessFileReader>>>>,
     writer: WriterT,
 }
 
 impl<WriterT> ManyCar<WriterT> {
     pub fn new(writer: WriterT) -> Self {
         ManyCar {
             shared_cache: Arc::new(Mutex::new(ZstdFrameCache::default())),
-            read_only: Vec::new(),
+            read_only: Mutex::new(Vec::new()),
```
lemmih, line 33: What's most appropriate here, `Mutex` or `RwLock`?

src/daemon/db_util.rs
```diff
@@ -0,0 +1,190 @@
...
+        downloaded_car_temp_path.persist(&forest_car_db_path)?;
+    } else {
+        // Use another temp file to make sure all final `.forest.car.zst` files are complete and valid.
+        let forest_car_db_temp_path =
+            tempfile::NamedTempFile::new_in(forest_car_db_dir)?.into_temp_path();
+        transcode_into_forest_car(&downloaded_car_temp_path, &forest_car_db_temp_path).await?;
+        forest_car_db_temp_path.persist(&forest_car_db_path)?;
+    }
+
+    let ts = ForestCar::new(Mmap::map_path(&forest_car_db_path)?)?.heaviest_tipset()?;
```
lemmih, line 92: The use of mmap should be governed by the environment variable.

src/daemon/mod.rs
```diff
@@ -278,8 +277,6 @@ pub(super) async fn start(
         return Ok(());
     }
 
-    let epoch = chain_store.heaviest_tipset().epoch();
-
```
lemmih, line 282: Why did you remove this?

lemmih (changes requested):
Getting closer. There's still a lot of logic that was changed and I cannot see why you changed it.

src/daemon/mod.rs
hanabi1224: And then removed, as this function no longer exists

src/db/car/many.rs
hanabi1224: I'm using `Mutex` to align with `shared_cache`. What's your recommendation?

src/daemon/db_util.rs
hanabi1224: Fixed

src/daemon/db_util.rs
hanabi1224: Fixed

src/daemon/mod.rs
hanabi1224: Moved back

src/daemon/mod.rs
hanabi1224: Because logically it's insufficient to use `epoch` to make the decision in `fetch_snapshot_if_required`, ideally it should validate the integrity of the union DB (the change is not included in this PR as suggested)

lemmih: I don't understand why it's logically insufficient. If it _is_ insufficient, document why in the code. If it _isn't_, let's revert the code.

hanabi1224: Code doc added [here](https://github.com/ChainSafe/forest/pull/3365/files#diff-bf435d9dfa9a63896065cbbd3082d65b24a937afbab2638ae820385108446a59R479)
```rust
// TODO: Add integrity check of the union DB to decide whether or not a snapshot is required.
// Checking the heaviest tipset that is only loaded from parity-db is insufficient
```

lemmih: You only wrote that it is insufficient, but you haven't written why it is insufficient. I do not believe it is insufficient. Write a comment that convinces me otherwise.

hanabi1224: @lemmih  The first part of the comment is the reason. We require a snapshot when forest cannot run with the current union DB, successfully reading epoch info from parity-db does not mean DB is complete (e.g. CAR DB files are manually removed). Does that make sense?

src/db/car/many.rs
hanabi1224: Switched to `RwLock`

src/daemon/mod.rs
hanabi1224: Reverted as suggested

src/daemon/db_util.rs
```diff
@@ -0,0 +1,190 @@
...
+    time,
+};
+use tokio::io::AsyncWriteExt;
+use tracing::info;
+use url::Url;
+use walkdir::WalkDir;
+
+pub fn load_all_forest_cars<T>(
+    store: &ManyCar<T>,
+    forest_car_db_dir: impl AsRef<Path>,
```
lemmih, line 26: nit: `&Path` is simpler.

src/daemon/db_util.rs
```diff
@@ -0,0 +1,190 @@
...
+                    }
+                }
+            }
+            None
+        })
+    {
+        let car = ForestCar::try_from(file.as_path())
+            .with_context(|| format!("Error loading car DB at {}", file.display()))?;
+        store.read_only(car.into());
+        info!("Loaded car DB at {}", file.display());
```
lemmih, line 48: Let's remove this or demote it to a DEBUG message. It's exposing an implementation detail that users should not concern themselves with.

src/daemon/db_util.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    Ok((forest_car_db_path, ts))
+}
+
+async fn download_to(url: Url, destination: &Path) -> anyhow::Result<()> {
+    snapshot::download_file(
+        url,
+        destination
+            .parent()
+            .context("Infallible getting the directory")?,
```
lemmih, line 105: I don't think these are infallible. Let's update the error message. (Or refactor the code.)

src/daemon/mod.rs
```diff
@@ -453,9 +463,16 @@ async fn fetch_snapshot_if_required(
     config: &mut Config,
```
lemmih, line 463: Let's radically simplify this function. There's no longer any point in downloading the snapshot here. Instead, just set the snapshot URL in `config.client.snapshot_path` and return.

src/utils/io/mmap.rs
```diff
@@ -36,6 +40,49 @@ impl Size for Mmap {
...
+    fn size(&self) -> io::Result<Option<u64>> {
+        use EitherMmapOrRandomAccessFile::*;
+        match self {
+            Mmap(mmap) => mmap.size(),
+            RandomAccessFile(file) => file.size(),
+        }
+    }
+}
+
+fn prefer_file_io_over_mmap() -> bool {
```
lemmih, line 78: A preference is optional. This is not optional. It's a requirement. If the environment variables tells us not to use mmap then we must not use mmap.

src/daemon/mod.rs
```diff
@@ -514,9 +531,9 @@ async fn fetch_snapshot_if_required(
             })
             .await;
             if !have_permission {
-                bail!("Forest requires a snapshot to sync with the network, but automatic fetching is disabled.")
+                bail!("Forest requires a snapshot to sync with the network, but automatic fetching is disabled.");
```
lemmih, line 534:
```suggestion
                bail!("Forest requires a snapshot to sync with the network, but automatic fetching is disabled.")
```

lemmih (changes requested):

src/utils/io/mmap.rs
hanabi1224: I don't quite get it, could you elaborate?

lemmih: The name `prefer_file_io_over_mmap` implies that it's still optional whether we use mmap or file-io. But it is not optional. If we're told to use file-io then we must use file-io.

src/daemon/mod.rs
hanabi1224: Fixed

src/daemon/mod.rs
hanabi1224: Fixed

src/daemon/db_util.rs
hanabi1224: Fixed

src/daemon/db_util.rs
hanabi1224: Fixed

src/daemon/db_util.rs
```diff
@@ -0,0 +1,190 @@
...
+
+pub fn load_all_forest_cars<T>(
+    store: &ManyCar<T>,
+    forest_car_db_dir: impl AsRef<Path>,
+) -> anyhow::Result<()> {
+    let forest_car_db_dir = forest_car_db_dir.as_ref();
+    if !forest_car_db_dir.is_dir() {
+        fs::create_dir_all(forest_car_db_dir)?;
+    }
+    for file in WalkDir::new(forest_car_db_dir)
```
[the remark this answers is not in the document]
hanabi1224: Added `.max_depth(1)`.  Was thinking about putting diff cars in a subdirectory but nvm

src/utils/io/mmap.rs
hanabi1224: How about renaming it to `should_use_file_io`?

src/daemon/mod.rs
```diff
@@ -393,21 +391,32 @@ pub(super) async fn start(
...
+            &mut config,
+            epoch,
+            opts.auto_download_snapshot,
+            &db_root_dir,
         )
-        .await
-        .context("Failed miserably while importing chain from snapshot")?;
-        info!("Imported snapshot in: {}s", stopwatch.elapsed().as_secs());
+        .await?;
+        consume_snapshot_file = true;
```
lemmih, line 406: No need to set this flag as true. All downloaded files are always consumed.

lemmih (approved):
`db_util.rs` can be refactored away, but let's leave that for another PR.

src/daemon/mod.rs
hanabi1224: Removed

src/db/car/forest.rs
hanabi1224: https://github.com/ChainSafe/forest/pull/3427

src/cli_shared/snapshot.rs
```diff
@@ -72,11 +76,31 @@ pub async fn fetch(
...
+pub async fn download_file_with_retry(
+    url: Url,
+    directory: &Path,
+    filename: &str,
+) -> anyhow::Result<PathBuf> {
+    Ok(retry(
+        RetryArgs {
+            timeout: None,
+            max_retries: Some(3),
+            delay: Some(Duration::from_secs(60)),
```
LesnyRumcajs, line 91: Do we want to have it somehow parametrized?

src/cli_shared/snapshot.rs
```diff
@@ -72,11 +76,31 @@ pub async fn fetch(
         .date_and_height_and_forest();
     let filename = filename(vendor, chain, date, height, forest_format);
 
-    match download_aria2c(&url, directory, &filename).await {
+    download_file_with_retry(url, directory, &filename).await
+}
+
+pub async fn download_file_with_retry(
+    url: Url,
```
LesnyRumcajs, line 83: Does it need to be owned?

src/daemon/mod.rs
```diff
@@ -393,21 +391,30 @@ pub(super) async fn start(
...
-        .await
-        .context("Failed miserably while importing chain from snapshot")?;
-        info!("Imported snapshot in: {}s", stopwatch.elapsed().as_secs());
+        .await?;
+    }
+
+    // Import chain if needed
+    if !opts.skip_load.unwrap_or_default() {
+        if let Some(path) = &config.client.snapshot_path {
+            // TODO: respect `--consume-snapshot` CLI option once it's implemented
```
LesnyRumcajs, line 409: do we have a tracking issue for this?

LesnyRumcajs (commented):
Do we have good coverage for this feature? E.g., scenarios with multiple CAR files?

src/cli_shared/snapshot.rs
hanabi1224: I feel it's better to not ask the function caller to decide these parameters, how about changing it to 
```rust
RetryArgs {
    timeout: None,
    ..Default::default()
}
```

src/daemon/mod.rs
hanabi1224: Yes it's part of https://github.com/ChainSafe/forest/issues/3334

src/cli_shared/snapshot.rs
hanabi1224: Not here, fixed

hanabi1224:
> Do we have good coverage for this feature? E.g., scenarios with multiple CAR files?

@LesnyRumcajs  I think CI tests should have covered most of the important scenarios.  As for multiple CAR, it should have been covered by `ManyCar` tests but I can extend the CI check to load multiple CARs from both vendors we support if that's preferrable, what do you think?

src/cli_shared/snapshot.rs
LesnyRumcajs: Sounds good

src/daemon/mod.rs
LesnyRumcajs: let's have the link in the comment

LesnyRumcajs:
> > Do we have good coverage for this feature? E.g., scenarios with multiple CAR files?
> 
> @LesnyRumcajs I think CI tests should have covered most of the important scenarios. As for multiple CAR, it should have been covered by `ManyCar` tests but I can extend the CI check to load multiple CARs from both vendors we support if that's preferrable, what do you think?

It wouldn't hurt to have such integration test.

Another thing that may be interesting to test (if I understand the properties of this feature correctly) is to add an artificially crafted CAR with a single entry (that would normally not exist in the current blockchain) and see if we can retrieve it via e.g., `forest-cli chain read-obj`.

hanabi1224:
> Another thing that may be interesting to test (if I understand the properties of this feature correctly) is to add an artificially crafted CAR with a single entry (that would normally not exist in the current blockchain) and see if we can retrieve it via e.g., `forest-cli chain read-obj`.

@LesnyRumcajs How about loading `test-snapshots/chain4.car`? Its root CID does not exist in a calibnet DB and that would test both `multi-CAR` and `read-obj`

src/daemon/mod.rs
hanabi1224: Done

hanabi1224:
@LesnyRumcajs `forest-cli chain read-obj` test added

LesnyRumcajs (approved):

PR #3379, opened 2023-08-15
[Enhancement] Zero-copy DAG_CBOR into Vec<Cid> parser.

ruseinov:
## Summary of changes

- Introduced a deserializer to speed up CBOR > Vec<Cid> parsing. 

Closes

## Change checklist


src/utils/encoding/cid_de_cbor.rs
```diff
@@ -0,0 +1,300 @@
...
+            {
+                Ok(())
+            }
+
+            #[inline]
+            fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                self.visit_byte_buf(v.to_owned())
```
lemmih, line 64:
```suggestion
                Ok(())
```

ruseinov:
> I'm confused - the title says zero-copy serialization of `Vec<Cid>`, but I'd expect to see a lifetime in that case, deserializing into a &`'serial [Cid]` or something.
> 
> Plus, the in-memory representation of a CID is different to the CBOR serialization, so I'd expect to see a new struct, but I don't - what am I missing?

The title does not say anything about serialization whatsoever. It's deserialization of a `CBOR`-encoded blob into a `Vec<Cid>`. 

1. We are recursively traversing the `DAB_CBOR`-encoded blob without parsing everything into a recursive `Ipld` struct. 
2. All the results are saved into a `Vec<Cid>`. 
3. That yields 3x performance boost over normal parsing.

ruseinov:
> > I'm confused - the title says zero-copy serialization of `Vec<Cid>`, but I'd expect to see a lifetime in that case, deserializing into a &`'serial [Cid]` or something.
> > Plus, the in-memory representation of a CID is different to the CBOR serialization, so I'd expect to see a new struct, but I don't - what am I missing?
> 
> The CIDs are copied (which can be done without any heap allocations.) Other data, such as bytes and strings, are not copied.

Yeah, zero-copy here is a bit misleading, I'll make sure this is documented properly.

lemmih:
> > > I'm confused - the title says zero-copy serialization of `Vec<Cid>`, but I'd expect to see a lifetime in that case, deserializing into a &`'serial [Cid]` or something.
> > > Plus, the in-memory representation of a CID is different to the CBOR serialization, so I'd expect to see a new struct, but I don't - what am I missing?
> > 
> > 
> > The CIDs are copied (which can be done without any heap allocations.) Other data, such as bytes and strings, are not copied.
> 
> Yeah, zero-copy here is a bit misleading, I'll make sure this is documented properly.

I don't think it's inaccurate to call this a zero-copy parser. The PR is doing two things: a) parsing an ipld structure, b) extracting CIDs from `Link` branches. The parsing part is definitely zero-copy. It's not a generalized zero-copy `Ipld` parser, though.

aatifsyed:
Ahh I understand now!

> The CIDs are copied (which can be done without any heap allocations.) Other data, such as bytes and strings, are not copied.
> We are recursively traversing the DAG_CBOR-encoded blob without parsing everything into a recursive Ipld struct.
> The PR is doing two things: a) parsing an ipld structure, b) extracting CIDs from Link branches. The parsing part is definitely zero-copy. It's not a generalized zero-copy Ipld parser, though.

Zero-copy in my head means holding references into buffers. Copying CIDs _out_ of it is not zero copy ;) 

I think `(pre?) filtered deserialization` could be a better term here?

src/utils/encoding/cid_de_cbor.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        impl<'de, 'a> Visitor<'de> for CollectCidVisitor<'a> {
+            type Value = ();
+
+            fn expecting(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+                fmt.write_str("any valid IPLD kind")
+            }
+
+            #[inline]
+            fn visit_str<E>(self, _value: &str) -> Result<Self::Value, E>
```
aatifsyed, line 51: nonblocking: can you macro these away with a comment? Or at least lift the most relevant fns to the top?
"the default visitor errors when visiting these, but we want to skip over them, so return `Ok(())`"
Out of curiosity why have you only done that for _some_ of the visited types?

src/utils/encoding/cid_de_cbor.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Default)]
+pub struct CidVec(Vec<Cid>);
+
+impl Deref for CidVec {
+    type Target = Vec<Cid>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
aatifsyed, line 22: > Implementing Deref for smart pointers makes accessing the data behind them convenient, which is why they implement Deref. On the other hand, the rules regarding Deref and [DerefMut](https://doc.rust-lang.org/std/ops/trait.DerefMut.html) were designed specifically to accommodate smart pointers. Because of this, **Deref should only be implemented for smart pointers** to avoid confusion.

- https://doc.rust-lang.org/std/ops/trait.Deref.html (emphasis not mine)

Not a smart pointer

src/utils/encoding/cid_de_cbor.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl DerefMut for CidVec {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+// [`CollectCid`] struct allows for recursive traversal of [`Ipld`] in order to retrieve all the
+// CIDs it encounters, using a preallocated vector.
+struct CollectCid<'a>(&'a mut Vec<Cid>);
```
aatifsyed, line 32:
```suggestion
/// [`FilterCids`] traverses and [`Ipld`] tree, appending [`CID`]s (and only CIDs) to a single vector.
/// This is much faster than constructing an [`Ipld`] tree and then performing the filtering.
struct FilterCids<'a>(&'a mut Vec<Cid>);
```

src/utils/encoding/cid_de_cbor.rs
ruseinov: We only care about list, map and newtype_struct, because a list and a map could contain more of those and the only type that converts to a CID is the newtype_struct.

lemmih:
> Ahh I understand now!
> 
> > The CIDs are copied (which can be done without any heap allocations.) Other data, such as bytes and strings, are not copied.
> > We are recursively traversing the DAG_CBOR-encoded blob without parsing everything into a recursive Ipld struct.
> > The PR is doing two things: a) parsing an ipld structure, b) extracting CIDs from Link branches. The parsing part is definitely zero-copy. It's not a generalized zero-copy Ipld parser, though.
> 
> Zero-copy in my head means holding references into buffers. Copying CIDs _out_ of it is not zero copy ;)

Imagine a zero-copy parser that returns an `Ipld` structure with references instead of owned data. Then imagine traversing that structure to extract the CIDs. This PR does both at once, with the intermediate structure deforested out.

src/utils/encoding/cid_de_cbor.rs
ruseinov: I don't think that introducing a macro here is going to help anything, it's just going to obfuscate the obvious. I'll re-arrange the order, however.

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@
 
 ### Added
 
+- [#3379](https://github.com/ChainSafe/forest/pull/3379): Zero-copy DAG_CBOR
+  into Vec<Cid> parser.
```
lemmih, line 38: This is not user-visible addition. I don't think it should be in the CHANGELOG. If we want a CHANGELOG entry, it should be under `Changed`, and it should say the performance was improved when walking the state graph.

src/utils/encoding/cid_de_cbor.rs
lemmih: I would have gone with `into_inner` here.

CHANGELOG.md
ruseinov: Sounds good.

src/utils/encoding/cid_de_cbor.rs
ruseinov: Alright, `into_inner` it is then.

aatifsyed: > not using helpful abstractions in this case seems wasteful.

It's the wrong abstraction, and a commonly misused one at that, which is why it's called out in the docs

ruseinov: > It's the wrong abstraction, and a commonly misused one at that, which is why it's called out in the docs

Sure is, though I'm still of the opinion that thin wrappers could benefit a lot from this. I would go further and say that the need for such wrappers to override certain behaviours calls for something magic/transparent to make the wrapper adhere to the same interface as whatever is wrapped. I wish there was a more eloquent way of implementing custom ser/de for types, e.g. the ability to have an implementation for a type alias.

Personal opinions aside - this has been fixed.

src/utils/encoding/cid_de_cbor.rs
```diff
@@ -0,0 +1,177 @@
...
+use cid::serde::BytesToCidVisitor;
+use cid::Cid;
+use core::fmt;
+use serde::de::{self, DeserializeSeed, SeqAccess, Visitor};
+use serde::Deserializer;
+
+/// [`CidVec`] allows for efficient zero-copy de-serialization of `DAG_CBOR`-encoded nodes into a
+/// vector of [`Cid`].
+#[derive(Default)]
+pub struct CidVec(Vec<Cid>);
```
aatifsyed, line 13: I don't really like this name -it's too general - it was the original name for what is now `FrozenCids`, for example.

Can you think of a better name, or maybe different API? `pub fn filter_cids(_: ???)`

ruseinov: Yeah, it's a good point. My idea behind this name is exactly that - this represents a `Vec<Cid>`. If we had a better way of dealing with ser/de without a wrapper - this wouldn't be needed. 

Perhaps a good middleground is to do something like: `pub fn extract_cids(cbor_blob: &bytes): Vec<Cid>`.

ruseinov: It's done.

LesnyRumcajs (commented):
Is this feature covered by unit tests?

ruseinov:
> Is this feature covered by unit tests?

Good point, it actually isn't.

src/utils/encoding/cid_de_cbor.rs
```diff
@@ -0,0 +1,185 @@
...
+/// [`CidVec`] allows for efficient zero-copy de-serialization of `DAG_CBOR`-encoded nodes into a
+/// vector of [`Cid`].
+#[derive(Default)]
+struct CidVec(Vec<Cid>);
+
+impl CidVec {
+    pub fn into_inner(self) -> Vec<Cid> {
+        self.0
+    }
+}
```
aatifsyed, line 27:
```suggestion
/// Find and extract all the [`Cid`]s from a `DAG_CBOR`-encoded blob without employing any
/// intermediate allocations.
pub fn extract_cids(cbor_blob: &[u8]) -> anyhow::Result<Vec<Cid>> {
    let CidVec(v) = from_slice_with_fallback(cbor_blob)?;
    Ok(v)
}

struct CidVec(Vec<Cid>);
```

ruseinov: This won't compile. I'll clean it up manually.

src/utils/encoding/cid_de_cbor.rs
```diff
@@ -0,0 +1,247 @@
...
+                                u8::arbitrary(g),
+                            ]),
+                        )
+                    }
+                    Ipld::Map(map) => map.values_mut().for_each(|val| cleanup_ipld(val, g)),
+                    Ipld::List(vec) => vec.iter_mut().for_each(|val| cleanup_ipld(val, g)),
+                    // Cleaning up Integer and Float in order to avoid parser mistakes that result
+                    // in tag detection and a subsequent Cid decoding failure.
+                    // See https://github.com/ipld/serde_ipld_dagcbor/blob/master/src/de.rs#L178 and
+                    // https://github.com/ipld/serde_ipld_dagcbor/blob/master/src/de.rs#L119 .
```
LesnyRumcajs, line 221: This could use more explanation, for someone fairly fresh to this subject it's a bit shady to replace all numeric occurrences with zeroes.

Also, let's use permalinks, otherwise those links will get quickly outdated and point to random pieces of code (or 404).

ruseinov: Sounds good.

ruseinov: done

PR #3387, opened 2023-08-16
Add wallet delete RPC endpoint

### LINKED ISSUE #2939: Add support for wallet key deletion
**Issue summary**

In `forest-cli` we don't have an easy way for deleting an address from a wallet. The only way to do it right now would be to export, then reimport only the keys we would like to keep.

We should have an easy and safe way to perform this basic operation.

**Task summary**

- [ ] Implement `Filecoin.WalletDelete` endpoint
- [ ] Implement `forest-cli wallet delete` subcommand
- [ ] Implement corresponding unit and smoke tests

https://lotus.filecoin.io/lotus/manage/lotus-cli/#lotus-wallet-delete

creativcoder:
## Summary of changes

- Adds `wallet delete` RPC endpoint to forest-cli

Closes #2939

## Change checklist


jdjaustin (changes requested):
How about adding a short blurb in `documentation/src/cli.md` to document this change?

LesnyRumcajs (changes requested):
I'm not entirely convinced we should develop wallet commands at the moment, given we intend to change wallet handling. That said, I'm okay with merging this but this feature should be tested.

creativcoder:
> How about adding a short blurb in `documentation/src/cli.md` to document this change?

Added: https://github.com/ChainSafe/forest/pull/3387/files#diff-868cc172255261be1eab7fcf569b320a523dd70231b31401cf291774b2b98c4fR129

creativcoder:
> I'm not entirely convinced we should develop wallet commands at the moment, given we intend to change wallet handling. That said, I'm okay with merging this but this feature should be tested.

Added tests

jdjaustin (approved):

scripts/tests/calibnet_wallet_check.sh
```diff
@@ -71,6 +71,17 @@ done
...
+# wallet delete tests
+ADDR_DEL=$(forest-cli wallet new)
+
+forest-cli wallet delete -a "$ADDR_DEL"
+
+WALLET_OUT=$(forest-cli wallet list)
+
+if [[ ${WALLET_OUT} == *"$ADDR_DEL"* ]];then
+    echo "Error: the created wallet address $ADDR_DEL was not deleted"; exit 1;
+fi
```
LesnyRumcajs, line 83:
```suggestion
# Validate that the wallet no longer exists.
forest-cli wallet list | grep --null-data --invert-match "${$ADDR_DEL}"
```

src/wallet/subcommands/wallet_cmd.rs
```diff
@@ -93,6 +93,12 @@ pub enum WalletCommands {
         #[arg(short)]
         signature: String,
     },
+    /// Deletes the wallet associated with the given address.
+    Delete {
+        /// The address of the wallet to delete
+        #[arg(short)]
+        address: String,
```
LesnyRumcajs, line 100: Let's just make it a positional argument.

src/rpc/wallet_api.rs
```diff
@@ -222,3 +222,36 @@ where
...
+    crate::key_management::remove_key(&addr, &mut keystore)?;
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{shim::crypto::SignatureType, KeyStore};
+
+    #[tokio::test]
```
LesnyRumcajs, line 246: Let's make them more exhaustive. I'd like to see an attempt to delete a non-existing key (both because the keystore is empty and because it was not inserted) and some logic to validate what happens to the default key after removal.

LesnyRumcajs (changes requested):

scripts/tests/calibnet_wallet_check.sh
creativcoder: I get a
```
zsh: bad substitution
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', library/std/src/io/stdio.rs:1019:9                                                            
[1]    1736333 exit 101   forest-wallet list
``` 

on my local

and on ci: `./scripts/tests/calibnet_wallet_check.sh:80:59: error: To expand via indirection, use arrays, ${!name} or (for sh only) eval. [SC2082]` - not sure what's the fix in this case.

LesnyRumcajs: Sorry, it's `"${ADDR_DEL}"` at the end.

LesnyRumcajs (approved):

documentation/src/cli.md
```diff
@@ -126,6 +126,11 @@ Verify the message's integrity with an address and signature. Outputs `true` if
 signature verifies message integrity, otherwise `false`. Usage:
 `forest-wallet verify -m <hex message> -a <address> -s <signature>`
 
+### Delete:
+
+Deletes a wallet given its address. Usage:
+`forest-cli wallet delete -a <address>`
```
LesnyRumcajs, line 132: It's no longer a flag but a positional argument, no?

PR #3392, opened 2023-08-17
fix: make it a compile time error when the actor bundle size is invalid

hanabi1224:
## Summary of changes

- Make it a compile time error when the actor bundle size is invalid
- Add a `cargo publish --dry-run` CI job to capture errors like assets are invalid, assets are not properly included, and temporary crate patch, etc. that blocks cargo publishing.

Closes

## Change checklist

### DIFF
--- a/.github/workflows/forest.yml
+++ b/.github/workflows/forest.yml
@@ -77,12 +77,6 @@ jobs:
   cargo-publish-dry-run:
     runs-on: ubuntu-latest
     steps:
-      # To help investigate transient test failures
-      - run: lscpu
-      # find the nearest S3 space for storing cache files
-      - name: Show IP
-        run: curl ifconfig.me
-        continue-on-error: true
       - name: Checkout Sources
         uses: actions/checkout@v3
       - run: ./assets/ci_download.sh
--- a/.github/workflows/forest.yml
+++ b/.github/workflows/forest.yml
@@ -75,7 +75,6 @@ jobs:
           if-no-files-found: error
 
   cargo-publish-dry-run:
-    name: cargo publish --dry-run
     runs-on: ubuntu-latest
     steps:
       # To help investigate transient test failures
--- a/.github/workflows/forest.yml
+++ b/.github/workflows/forest.yml
@@ -74,6 +74,28 @@ jobs:
             ~/.cargo/bin/forest*
           if-no-files-found: error
 
+  cargo-publish-dry-run:
+    name: cargo publish --dry-run
+    runs-on: ubuntu-latest
+    steps:
+      # To help investigate transient test failures
+      - run: lscpu
+      # find the nearest S3 space for storing cache files
+      - name: Show IP
+        run: curl ifconfig.me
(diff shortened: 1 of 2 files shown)


aatifsyed (approved):

.github/workflows/forest.yml
```diff
@@ -74,6 +74,28 @@ jobs:
             ~/.cargo/bin/forest*
           if-no-files-found: error
 
+  cargo-publish-dry-run:
+    name: cargo publish --dry-run
```
LesnyRumcajs, line 78: The additional name isn't really more helpful so we might as well drop it.
```suggestion
```

.github/workflows/forest.yml
```diff
@@ -74,6 +74,28 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      # To help investigate transient test failures
+      - run: lscpu
+      # find the nearest S3 space for storing cache files
+      - name: Show IP
+        run: curl ifconfig.me
+        continue-on-error: true
```
LesnyRumcajs, line 86: Why do we need it in this job exactly? If there isn't anything special about it, let's either have it everywhere (ideally via a composite action to reduce code duplication) or only in critical places.

src/daemon/bundle.rs
```diff
@@ -5,11 +5,15 @@ use anyhow::Context;
...
+use static_assertions::const_assert_eq;
 
 pub async fn load_actor_bundles(db: &impl Blockstore) -> anyhow::Result<Vec<Cid>> {
     const ERROR_MESSAGE: &str = "Actor bundles assets are not properly downloaded, make sure git-lfs is installed and run `git lfs pull` again. See <https://github.com/git-lfs/git-lfs/blob/main/INSTALLING.md>";
+
     const ACTOR_BUNDLES_CAR_ZST: &[u8] = include_bytes!("../../assets/actor_bundles.car.zst");
-    assert!(ACTOR_BUNDLES_CAR_ZST.len() > 1024 * 1024, "{ERROR_MESSAGE}");
+    // Check bundle size at compile time, see `ERROR_MESSAGE` for details.
+    // Note: make sure the bundle size is updated with the bundle file.
+    const_assert_eq!(ACTOR_BUNDLES_CAR_ZST.len(), 2438387);
```
LesnyRumcajs, line 16: For the sake of completeness, it'd be helpful to provide instructions on how to regenerate the bundle. This way, a potential network upgrade implementor would be less frustrated by this sudden obstacle.

.github/workflows/forest.yml
hanabi1224: Removed

src/daemon/bundle.rs
hanabi1224: I'm working on a `forest-tool` subcommand that regenerates the bundle, and will update instructions in STATE_MIGRATION.md altogether

LesnyRumcajs (approved):

Issue #3203, opened 2023-07-17
version the database and automatically migrate data


lemmih:
**Issue summary**

Storing data is one of the prime purposes of Forest, and we often play around with different layouts. For example, the copying garbage collector requires a specific layout. The layout will be completely changed by the (not yet finished) non-moving garbage collector. Hubert's work on re-organizing the DB also involves layout changes: #3189

Up until now, we've been okay with deleting the database and starting from a fresh snapshot. This was annoying but acceptable since we never had more than 100GiB of data. Going forward, we'll be setting up nodes with ~14 TiB of data, and we want them to not require any maintenance when new versions of Forest are released.

I propose we put our database in separate folders for each Forest version. If `forest-0.1` has created a database in `%DATA_DIR/0.1/` and we try to load it with `forest-0.2`, these steps should happen:
- `0.1/` is opened and migrated to `pending/`.
- Validity checks are run against `pending/`. We can check that at least 900 tipsets are present, that the network identity (mainnet/calibnet) hasn't changed, and maybe even re-execute a few tipsets.
- If all checks pass (assuming they also passed using the pre-migrated database), atomically rename `pending/` to `0.2` and delete `0.1`.

Note: not all Forest versions require migrations. When no migration is required, the database folder can simply be renamed.

Note: Migrating from version A to version B should apply migrations for all intermediate versions.

Acceptance criteria:
- [X] Per-network versioned database.
- [x] Set up migration system with automated checks for the _pre_ and _post-migration_ databases. If pre-migration checks fail then the post-migration checks can be skipped.
- [ ] Documentation for how to write migrations.
- [X] CI test that goes through all migrations, one by one. This can be done by syncing to calibnet with an old version of Forest. Then sync to calibnet using Forest releases in chronological order.
- [X] CI test that goes through all migrations, from oldest straight to newest. This can be done by first syncing to calibnet with an old version of Forest and then syncing with the newest version.

**Other information and links**

lemmih:
Oh, and let's treat git versions of Forest differently. It'd be nice to allow for arbitrary changes to the DB format in development without having to re-initialize the database. Possible solution:

- Have a `dev` folder next to the regular versioned database folder.
- Use the `dev` folder if it exists. Otherwise, migrate from the versioned folder non-destructively (that is, migrate from the latest stable version without deleting the data).
- If opening the `dev` folder fails, delete it and try migrating again.

It'll double the space required to do dev work, but that shouldn't be a problem.

[...]

PR #3411, opened 2023-08-22
feat: automatically migrate database folder, ensuring eternal backwards compatibility

LesnyRumcajs:
## Summary of changes

- Introduced a database migration framework: this should ensure we no longer have to manually re-create Forest databases going forward.
- More post-migration checks would be great to implement - I can either do this here or in a follow-up PR,
- the integration test should ensure we catch breaking changes early and we always have a migration at hand (and that the person that breaks it has the golden opportunity to implement the migration!)
- implemented a PoC migration from `0.12.1`. This migration won't be performed outside of tests (unless someone manually changes the `paritydb` directory to `0.12.1`)
- adhered (mostly) to the [agreed design](https://github.com/ChainSafe/forest/blob/main/documentation/src/developer_documentation/database_migrations.md)
- Introduced a DB mode environmental variable, more explained in the Rust documentation and README.

Closes https://github.com/ChainSafe/forest/issues/3203

## Change checklist

### DIFF
--- a/documentation/src/environment_variables.md
+++ b/documentation/src/environment_variables.md
@@ -4,7 +4,23 @@ Besides CLI options and the configuration values in the configuration file,
 there are some environment variables that control the behaviour of a `forest`
 process.
 
-| Environment variable       | Value     | Default | Description                                              |
-| -------------------------- | --------- | ------- | -------------------------------------------------------- |
-| FOREST_KEYSTORE_PHRASE_ENV | any text  | empty   | The passphrase for the encrypted keystore                |
-| FOREST_CAR_LOADER_FILE_IO  | 1 or true | false   | Load CAR files with `RandomAccessFile` instead of `Mmap` |
+| Environment variable       | Value                            | Default | Description                                              |
+| -------------------------- | -------------------------------- | ------- | -------------------------------------------------------- |
+| FOREST_KEYSTORE_PHRASE_ENV | any text                         | empty   | The passphrase for the encrypted keystore                |
+| FOREST_CAR_LOADER_FILE_IO  | 1 or true                        | false   | Load CAR files with `RandomAccessFile` instead of `Mmap` |
+| FOREST_DB_DEV_MODE         | [see here](#-forest_db_dev_mode) | current | The database to use in development mode                  |
+
+### FOREST_DB_DEV_MODE
+
+By default, Forest will create a database of its current version or try to
+migrate to it. This can be overridden with the `FOREST_DB_DEV_MODE`
+environmental variable.
+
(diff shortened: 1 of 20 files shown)


documentation/src/environment_variables.md
```diff
@@ -4,7 +4,23 @@ Besides CLI options and the configuration values in the configuration file,
...
+| FOREST_CAR_LOADER_FILE_IO  | 1 or true                        | false   | Load CAR files with `RandomAccessFile` instead of `Mmap` |
+| FOREST_DB_DEV_MODE         | [see here](#-forest_db_dev_mode) | current | The database to use in development mode                  |
+
+### FOREST_DB_DEV_MODE
+
+By default, Forest will create a database of its current version or try to
+migrate to it. This can be overridden with the `FOREST_DB_DEV_MODE`
+environmental variable.
+
+| Value                          | Description                                                                                                                                      |
```
jdjaustin, line 19: For some reason, this section does not display correctly in the rich diff preview.

LesnyRumcajs: It should be fine now.

jdjaustin (approved):

hanabi1224:
A question about the scope of the DB migration, does it cover only parity-db or also the Forest CAR DB? See https://github.com/ChainSafe/forest/pull/3365#discussion_r1299225452

> Do note that the ForestCAR.zst format is not fixed, and re-encoding may be required.

lemmih:
> A question about the scope of the DB migration, does it cover only parity-db or also the Forest CAR DB? See [#3365 (comment)](https://github.com/ChainSafe/forest/pull/3365#discussion_r1299225452)
> 
> > Do note that the ForestCAR.zst format is not fixed, and re-encoding may be required.
> 
> @lemmih correct me if I'm wrong, as I understand, we need to migrate forest CAR files when a breaking change has been made to ForestCAR.zst format

CAR files are as much part of the database as ParityDB is.

LesnyRumcajs:
> A question about the scope of the DB migration, does it cover only parity-db or also the Forest CAR DB? See [#3365 (comment)](https://github.com/ChainSafe/forest/pull/3365#discussion_r1299225452)
> 
> > Do note that the ForestCAR.zst format is not fixed, and re-encoding may be required.
> 
> @lemmih correct me if I'm wrong, as I understand, we need to migrate forest CAR files when a breaking change has been made to ForestCAR.zst format

The DB migration should encompass all migrations required to have Forest transition from one version to another. At the moment we care only about the ParityDb (as it's the only one that breaks on every other update). That said, in the future we should also track things like keystore (probably should be moved to the DB? Not 100% sure, we'd need to assess security implications of this).

PR #3412, opened 2023-08-23
Update cid_de_cbor.rs

ruseinov:
## Summary of changes

- More precise comments for cid parser tests. 

Closes

## Change checklist


lemmih (approved):
We do a similar clean up when serializing IPLD to JSON. I really think we should write a mutable iterator.

jdjaustin (approved):

ruseinov:
> We do a similar clean up when serializing IPLD to JSON. I really think we should write a mutable iterator.

It would have to be an opinionated iterator then, because we'd have to `flat_map` lists and maps.

PR #3417, opened 2023-08-24
fix: relax diff snapshot tests to allow null tipsets

lemmih:
## Summary of changes

- The presence of null tipsets means that it's difficult to specify exact tests when splitting and merging snapshots. This PR relaxes the epoch limits for the diff-snapshot tests and should make the tests succeed even when there are lots of null tipsets.

Closes

## Change checklist


lemmih:
Ah, no, I'm just being silly.

Issue #3410, opened 2023-08-22
Tracking issue for RUSTSEC-2023-0052 and RUSTSEC-2023-0053


jdjaustin:
**Issue summary**

Update `rustls-webpki` and `webpki` to address RUSTSEC-2023-0052 and RUSTSEC-2023-0053.
```
Crate:     rustls-webpki
Version:   0.100.1
Title:     rustls-webpki: CPU denial of service in certificate path building
Date:      2023-08-22
ID:        RUSTSEC-2023-0053
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0053
Severity:  7.5 (high)
Solution:  Upgrade to >=0.100.2, <0.101.0 OR >=0.101.4
Dependency tree:
rustls-webpki 0.100.1
└── webpki-roots 0.23.1
    └── libp2p-websocket 0.42.0
        └── libp2p 0.52.2
            └── forest-filecoin 0.12.1

Crate:     rustls-webpki
Version:   0.101.3
Title:     rustls-webpki: CPU denial of service in certificate path building
Date:      2023-08-22
ID:        RUSTSEC-2023-0053
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0053
Severity:  7.5 (high)
Solution:  Upgrade to >=0.100.2, <0.101.0 OR >=0.101.4
Dependency tree:
rustls-webpki 0.101.3
└── rustls 0.21.6
    ├── tokio-rustls 0.24.1
    │   ├── reqwest 0.11.18
    │   │   ├── tracing-loki 0.2.4
    │   │   │   └── forest-filecoin 0.12.1
    │   │   └── forest-filecoin 0.12.1
    │   └── hyper-rustls 0.24.1
    │       └── reqwest 0.11.18
    ├── reqwest 0.11.18
    └── hyper-rustls 0.24.1

Crate:     webpki
Version:   0.22.0
Title:     webpki: CPU denial of service in certificate path building
Date:      2023-08-22
ID:        RUSTSEC-2023-0052
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0052
Severity:  7.5 (high)
Solution:  No fixed upgrade is available!
Dependency tree:
webpki 0.22.0
├── webpki-roots 0.22.6
│   └── reqwest 0.11.18
│       ├── tracing-loki 0.2.4
│       │   └── forest-filecoin 0.12.1
│       └── forest-filecoin 0.12.1
├── rustls 0.20.8
[...]

PR #3418, opened 2023-08-24
fix: RUSTSEC-2023-0052

hanabi1224:
## Summary of changes

Remove `webpki` crate from `Cargo.lock` by
- cargo update -p reqwest
- Remove `websocket` feature from `libp2p` as it does not seem to be useful at the moment

Closes https://github.com/ChainSafe/forest/issues/3410

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,7 +271,7 @@ dependencies = [
  "futures-core",
  "futures-io",
  "memchr",
- "pin-project-lite 0.2.12",
+ "pin-project-lite",
  "tokio",
  "zstd",
  "zstd-safe",
@@ -337,7 +337,7 @@ checksum = "cd56dd203fef61ac097dd65721a419ddccb106b2d2b70ba60a6b529f03961a51"
 dependencies = [
  "async-stream-impl",
  "futures-core",
- "pin-project-lite 0.2.12",
+ "pin-project-lite",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -8470,33 +8371,11 @@ dependencies = [
  "wasm-bindgen",
 ]
 
-[[package]]
-name = "webpki"
```
hanabi1224, line 8474: Problematic crate removed

src/libp2p/service.rs
```diff
@@ -826,8 +826,7 @@ async fn emit_event(sender: &Sender<NetworkEvent>, event: NetworkEvent) {
 pub fn build_transport(local_key: Keypair) -> anyhow::Result<Boxed<(PeerId, StreamMuxerBox)>> {
     let build_tcp = || libp2p::tcp::tokio::Transport::new(libp2p::tcp::Config::new().nodelay(true));
     let build_dns_tcp = || libp2p::dns::TokioDnsConfig::system(build_tcp());
-    let transport =
-        libp2p::websocket::WsConfig::new(build_dns_tcp()?).or_transport(build_dns_tcp()?);
+    let transport = build_dns_tcp()?;
```
LesnyRumcajs, line 829: So, we are not using WS anymore? Do you think this is fine for communicating with other peers?

hanabi1224: WS is only useful when TCP is disabled or not available (e.g. WASM nodes that run in browsers), which is not the case (I don't think lotus offers an option to disable TCP) and none of the bootstrap nodes use WS address

LesnyRumcajs: Alright, in case it turns out we need it, we can re-enable it again.

LesnyRumcajs (approved):

lemmih (approved):

PR #3426, opened 2023-08-30
fix: include all blocks when calculating diff sets

lemmih:
## Summary of changes

- all blocks going back to genesis should be included when calculating diff sets
- this fix reduces the file size of diff files

Closes

## Change checklist


LesnyRumcajs (commented):
Do we have a way of testing/proving the stated behaviour that would fail before this fix?

lemmih:
> Do we have a way of testing/proving the stated behaviour that would fail before this fix?

No, we don't have any tools that can test that yet. We need a way to state that the intersection of a base snapshot and a diff snapshot is null.

LesnyRumcajs (approved):

aatifsyed (approved):

Issue #3361, opened 2023-08-11
Using hundreds of CAR files as a single block store is slow


lemmih:
**Issue summary**

Multiple CAR files can be opened and joined into a single block store. This is convenient when using diff snapshots containing only the unique data generated over a range of epochs.

However, looking up a value in a `ManyCar` block store queries each CAR file. So if there are hundreds of CAR files, one lookup may query hundreds of files, one by one.

Steps to reproduce:
1. `cd calibnet_diff`
2. `aws --endpoint https://2238a825c5aca59233eab1f221f7aefb.r2.cloudflarestorage.com/ s3 cp "s3://forest-archive/calibnet/lite/forest_snapshot_calibnet_2022-11-01_height_0.forest.car.zst" .`
3. `aws --endpoint https://2238a825c5aca59233eab1f221f7aefb.r2.cloudflarestorage.com/ s3 cp "s3://forest-archive/calibnet/diff/" . --recursive`
4. `forest-cli snapshot validate --check-links 780000 --check-stateroots 780000 forest_snapshot_calibnet_2022-11-01_height_0.forest.car.zst forest_diff_calibnet_height_*`

One possible solution would be to order the CAR files according to when they were last used. Getting a successful hit from a CAR file should move it to the front of the list, ensuring active CAR files are queried first. This will improve performance if nearby epochs are accessed together rather than randomly.

Acceptance criteria:
- [ ] Traversing 2000 epochs with 260 CAR files should be, at most, 10% slower than traversing with 10 CAR files.

**Other information and links**

ruseinov:
Just throwing ideas here.

I remember us talking about this before, when we used a parallel iterator to deal with parallel queries. 

The proposed solution sounds good, however it only optimises for one scenario. Which means that the performance is likely to suffer even more in case of random access. It could be interesting to have several scenarios covered with the ability to choose one or the other using an argument, with a sane default.

Perhaps we could try and re-introduce the parallel iterator for those cases when we have lots of CAR files, with an argument that optionally enables it and proper docs that explain performance implications. 

We could also have some other interesting ways of optimising this, for example relying on a certain filename format, where we know which epochs each of the CAR files represent. That could perhaps help us put them in an order that optimises for certain types of traversal. Not sure if I'm making sense here, this is likely to be very similar to your `LRU`-style approach, except more work and constraints.

hanabi1224:
[...]

PR #3433, opened 2023-08-31
feat: serve actor bundles with forest CAR file instead of parity-db

hanabi1224:
## Summary of changes

- Save actor bundles as forest car file instead of loading into parity-db

Closes

## Change checklist

### DIFF
--- a/src/daemon/bundle.rs
+++ b/src/daemon/bundle.rs
@@ -1,16 +1,27 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use std::path::Path;
+
 use anyhow::Context;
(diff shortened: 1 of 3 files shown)


lemmih:
I'm not sure this is worth it. Each additional CAR file slows down the database.

hanabi1224:
@lemmih So this is blocked by https://github.com/ChainSafe/forest/issues/3361  
Have you had a chance to try the performance after mmap change was merged? And should we benchmark https://github.com/ChainSafe/forest/pull/3317 again to see if mmap makes difference?

lemmih:
> @lemmih So this is blocked by #3361 Have you had a chance to try the performance after mmap change was merged? And should we benchmark #3317 again to see if mmap makes difference?

The mmap change should reduce the overhead. But the overhead is still there. Why not insert the data into ParityDB?

Issue #3309, opened 2023-08-02
snapshot export: missing blocks


LesnyRumcajs:
**Describe the bug**

Exported snapshots are missing blocks, including (but not only) the genesis block. This does not happen every time. 

**To Reproduce** Steps to reproduce the behaviour:

Can't reproduce every time, but it happens with a limited amount of `snapshot export` and `snapshot validate`.

**Log output**

```
./forest_snapshot_calibnet_2023-08-02_height_787358.forest.car.zst: 1.58GB
Export completed.
Testing snapshot validity
./forest_snapshot_calibnet_2023-08-02_height_787358.forest.car.zst: 2473970486 bytes 
Verifying snapshot checksum
forest_snapshot_calibnet_2023-08-02_height_787358.forest.car.zst: OK
Validating CAR files
./forest_snapshot_calibnet_2023-08-02_height_787358.forest.car.zst: 2473970486 bytes 
Validating CAR file forest_snapshot_calibnet_2023-08-02_height_787358.forest.car
Error: Snapshot does not contain a genesis block
```
from https://github.com/ChainSafe/forest/actions/runs/5734304765/job/15540585277?pr=3296#step:6:843

**Expected behaviour**

All exported snapshots are valid.

**Screenshots**

We exported quite a few such snapshots. This can be noticed by the lower size.
![image](https://github.com/ChainSafe/forest/assets/9642092/d421c684-25de-4f66-9d50-8e86120b2202)

**Environment (please complete the following information):**

- Environment: daily snapshot service
- Branch/commit `forest-filecoin 0.12.0+git.b03ca5d61`

**Other information and links**

Some of the corrupted snapshots moved to https://cloud.digitalocean.com/spaces/forest-snapshots?i=88c522&path=debug%2F%233309%2F

LesnyRumcajs:
Resolved via https://github.com/ChainSafe/forest-iac/pull/201

LesnyRumcajs:
bad bot

LesnyRumcajs:
Sadly, it's not resolved.
[no-debug.txt](https://github.com/ChainSafe/forest/files/12319493/no-debug.txt)

LesnyRumcajs:
Note - I wasn't able to reproduce this issue with over 200 calibnet exports on my machine (I used [this](https://gist.github.com/LesnyRumcajs/d313f190a7271c5bc97f5e40118b2536) script)

LesnyRumcajs:
[...]

PR #3440, opened 2023-09-01
Try fix snapshot export

elmattic:
## Summary of changes

As I can't reproduce it on a DO droplet, this PR might not fix [our issue](https://github.com/ChainSafe/forest/issues/3309). But this should not worsen the situation.

- Use `writer` `shutdown` after flushing
- Use `file` `sync_all` before dropping

Closes

https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.sync_all

## Change checklist


lemmih (changes requested):
I don't think this code should be merged without a solid reason.

src/chain/mod.rs
```diff
@@ -54,6 +54,7 @@ pub async fn export<D: Digest>(
 
     // Flush to ensure everything has been successfully written
     writer.flush().await.context("failed to flush")?;
+    writer.shutdown().await.context("failed to shutdown")?;
```
lemmih, line 57: This code literally does nothing. `shutdown` is only defined on sockets.

lemmih:
I'm fairly certain the errors we've been seeing have nothing to do with the file being truncated. The magic number we add at the end of the CAR.forest.zst file is still there.

src/chain/mod.rs
elmattic: Reading the code I was thinking this will should call internally `poll_flush` method from `AsyncWrite` trait in `BufWriter`.

elmattic:
> I don't think this code should be merged without a solid reason.

Putting the PR back to draft then.

elmattic:
Closing it as this is probably not the right fix.

PR #3447, opened 2023-09-01
chore: implement arbitrary for `Config`

### LINKED ISSUE #2965: ConfigPartial shouldn't exist
> If it is wrong now (as the documentation tends to lag behind the code changes) then please modify it to reflect the current state. A wild TODO is, in general, not very helpful.

_Originally posted by @LesnyRumcajs in https://github.com/ChainSafe/forest/pull/2893#discussion_r1226458059_

lemmih:
## Summary of changes

- lower `u64` and `usize` to `u32` such that they can be encoded as TOML.
- implement `Arbitrary` for `Config`.
- remove `ConfigPartial`.

Closes #2965

## Change checklist

### DIFF
--- a/src/networks/mod.rs
+++ b/src/networks/mod.rs
@@ -35,6 +35,7 @@ const DEFAULT_REQUEST_WINDOW: usize = 8;
 /// Forest builtin `filecoin` network chains. In general only `mainnet` and its
 /// chain information should be considered stable.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(test, derive(derive_quickcheck_arbitrary::Arbitrary))]
 #[serde(tag = "type", content = "name", rename_all = "lowercase")]
 pub enum NetworkChain {
     Mainnet,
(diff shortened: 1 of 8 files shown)


LesnyRumcajs (approved):

jdjaustin (approved):

src/networks/mod.rs
```diff
@@ -172,6 +173,21 @@ pub struct ChainConfig {
...
+impl quickcheck::Arbitrary for ChainConfig {
+    fn arbitrary(g: &mut quickcheck::Gen) -> ChainConfig {
+        use quickcheck::Arbitrary;
+        // Note: Not all `u64` values can be stored as TOML. Maybe we should use
+        // `u32` instead. Or maybe these values shouldn't be configurable at all.
+        ChainConfig {
+            network: Arbitrary::arbitrary(g),
+            genesis_cid: Arbitrary::arbitrary(g),
+            recent_state_roots: Arbitrary::arbitrary(g),
+            ..ChainConfig::default()
```
aatifsyed, line 186: Why?

lemmih: The config has to be serializable as TOML. But TOML does not support `u64`. Generating the `u64` fields will, correctly, cause our tests to fail. The note above talks about possible ways of fixing this.

aatifsyed: I think you should fix this by doing `#[arbitrary(gen(|g| u32::arbitrary(g) as u64))]`

lemmih: I'm not sure that's the best way forward. Ideally, bad states should be unrepresentable. Another option is to delete those fields entirely. I don't think we'll ever support arbitrary networks (ie, not calibnet, mainnet, or devnet).

aatifsyed (approved):

Issue #2945, opened 2023-06-06
Forest integration and unit tests


aatifsyed:
**Issue summary**
Our tests right now are a bit messy, they're split across
- shell scripts
  - coverage
  - tests
- rust test
  - unit tests
  - integration tests

Some tests call out to the web, or are otherwise undesirable to run locally - these tend to be put in shell scripts.

~When I _do_ run `cargo test`, it takes 5 minutes, which is too long for a nice dev-test cycle.~

Additionally, whenever I'm debugging an integration test, I have to write a bunch of boilerplate to actually get logs out of `forest` via the `Assert`.

I'd like the following
- [ ] Add a test harness for integration tests.
  - [x] Making forest a single workspace will make this much easier
  - [ ] Support running `forest` and `forest-cli` interactions
  - See also #2907 
- [ ] Replace all shell script tests with rust integration tests, with ci-aware segregation. We would always e.g fetch snapshots in CI, but not locally (unless the user wanted to)
  - [ ] Use `cargo nextest` everywhere, giving us `cargo nextest <include-ci-tests>` or something like that
- [ ] Remove coverage scripts, they should just be `cov cargo nextext <all>`
- [x] #3013

**Other information and links**

LesnyRumcajs:
Are you not able to run the tests locally?

aatifsyed:
Not our integration tests - they fetch gigabytes from the web, and take ages, hence wanting segregated tests with a single test runner

LesnyRumcajs:
Could you please elaborate on how those would those segregated tests look like? Unless you're on cellular, they shouldn't take long, even locally.
![image](https://github.com/ChainSafe/forest/assets/9642092/07d504e3-8a79-468c-af99-f5a81443d02a)

aatifsyed:
Updated the description :)

aatifsyed:
Maybe I should simply get a better PC

```console
        PASS [  10.530s] forest_key_management keystore::test::test_generate_key
        PASS [  10.707s] forest_key_management keystore::test::test_read_write_encrypted_keystore
        PASS [  36.846s] forest_utils io::writer_checksum::test::file_writer_fs_buf_writer
        LEAK [  53.346s] forest_libp2p_bitswap::go_compat tests::bitswap_go_compat_test
        SLOW [> 60.000s] forest-cli::config test_download_location_of_proof_parameter_files_default
        SLOW [> 60.000s] forest-cli::config test_download_location_of_proof_parameter_files_env
[...]

PR #3448, opened 2023-09-01
fix: return early for out-of-bounds mmap reads

### LINKED ISSUE #3446: Handle some corner case in `archive export`
**Describe the bug**

Not a big deal but giving an empty file to `forest-cli archive export` will result in a panic.

**To Reproduce** Steps to reproduce the behaviour:

1. Run `touch dummy.car`
2. Run `forest-cli archive export dummy.car`
3. See error

**Log output**

<details>
  <summary>Log Output</summary>
  
  ```
$ forest-cli archive export dummy.car
thread 'main' panicked at 'attempt to add with overflow', src/utils/io/mmap.rs:25:19
  ```
</details>

**Expected behaviour**

We should handle this situation without crashing.

- OS: macOS
- Branch: main

lemmih:
## Summary of changes

- return early if mmap read request is out-of-bounds.

Closes #3446

## Change checklist


aatifsyed (changes requested):
I don't think this should be a shell script

scripts/tests/forest_tool_check.sh
```diff
@@ -5,6 +5,12 @@
...
-FOREST_TOOL_PATH="forest-tool"
+source "$(dirname "$0")/harness.sh"
 
 "$FOREST_TOOL_PATH" state-migration actor-bundle
+
+# Exporting with an empty car should fail but not panic
+touch empty.car
+ERROR=$($FOREST_CLI_PATH archive export empty.car 2>&1 || true)
+assert_eq "$ERROR" "Error: input not recognized as any kind of CAR data (.car, .car.zst, .forest.car)"
+rm empty.car
```
aatifsyed, line 16: Can you write something like https://github.com/ChainSafe/forest/blob/62df2d8d608f2629664ad35e3e82fe3f930ee038/tests/import_snapshot_tests.rs?

```rust
touch("empty.car");
forest_cli().args(["archive export empty.car").assert().failure().stdout().contains("Error: input not recognized...")
```

Bonus points if you port the existing test and remove this file.

See also https://github.com/ChainSafe/forest/issues/2945

src/utils/io/mmap.rs
```diff
@@ -22,6 +22,9 @@ impl Mmap {
 impl ReadAt for Mmap {
     fn read_at(&self, pos: u64, buf: &mut [u8]) -> io::Result<usize> {
         let start = pos as usize;
+        if start >= self.0.len() {
+            return Ok(0);
```
aatifsyed, line 26:
```suggestion
            // This matches the behaviour for seeking past the end of a file
            return Ok(0);
```

aatifsyed (approved):
Fantastic!

lemmih:
Hmm, turns out that `state-migration actor-bundle` was in a separate script to bypass the 30s timeout.

aatifsyed:
> Hmm, turns out that `state-migration actor-bundle` was in a separate script to bypass the 30s timeout.

I'd advocate for keeping it in rust, and segregating/configuring it via nextest

.config/nextest.toml
```diff
@@ -1,2 +1,6 @@
 [profile.default]
 slow-timeout = { period = "10s", terminate-after = 3 }
+
+[[profile.default.overrides]]
+filter = 'test(state_migration_actor_bundle_runs)'
+slow-timeout = { period = "120s", terminate-after = 3 }
```
LesnyRumcajs, line 6: Why is it so slow? Can we make it faster? I'd very much prefer a regular cargo test to be speedy.

aatifsyed: Or alternatively, have it run in ci only using profiles: https://nexte.st/book/configuration.html#profiles

sudo-shashank (approved):

elmattic (approved):

.config/nextest.toml
lemmih: I'll keep in the regular tests for now. Otherwise, it would be out of sight and never fixed. :)

PR #3461, opened 2023-09-05
Add CarWriter sink

### LINKED ISSUE #3192: Consider `struct CarWriter: futures::Sink`
We have two ways to read CARs:
- [`CarStream`](https://github.com/ChainSafe/forest/blob/7e8d98141051d18996200ec3617a9b4e60c11620/src/utils/db/car_stream.rs#L96) for when we want a stream of IPLD blocks. This is used a couple of times in `forest-tool`
- [`ManyCar`](https://github.com/ChainSafe/forest/blob/7e8d98141051d18996200ec3617a9b4e60c11620/src/db/car/many.rs)
  - [ ] Make sure that only `ManyCar` is used - people shouldn't be manually creating `AnyCar` etc

We currently have two distinct ways to write CARs:
- For `.forest.car.zst` we use [`forest::Encoder::write`](https://github.com/ChainSafe/forest/blob/7e8d98141051d18996200ec3617a9b4e60c11620/src/db/car/forest.rs#L248).
- For `.car` we use [`fvm_ipld_car::CarHeader::write_stream_async`](https://docs.rs/fvm_ipld_car/latest/fvm_ipld_car/struct.CarHeader.html#method.write_stream_async).

It would afford us more flexiblity in output if there was just one writer for the different formats:
```rust
let filtered_blocks: impl TryStream<Ok=(Cid, ...)>;

// --output-format=carv1
filtered_blocks.forward(
    CarWriter::car_v1(File::open(...)),
);

// --output-format=forest-car-zst
filtered_blocks.forward(
    CarWriter::forest_car_zst(...)
)
```

# Open questions
- This makes writing CAR files in e.g `forest-cli` `forest-tool` more flexible - think `snapshot export`
- Writing the `Sink` will be pretty thorny - is it worth the flexibility?

elmattic:
## Summary of changes

-

Closes #3192

## Change checklist


src/utils/db/car_stream.rs
```diff
@@ -6,7 +6,9 @@ use cid::{
     multihash::{Code, MultihashDigest},
     Cid,
 };
-use futures::{Stream, StreamExt};
+use flume::r#async::RecvStream;
+use futures::{sink::Sink, Stream, StreamExt};
+use fvm_ipld_car::CarHeader as FvmCarHeader;
```
lemmih, line 11: I know this is a draft, but note that you cannot use `fvm_ipld_car`. The entire point is to _not_ use that library.

src/utils/db/car_stream.rs
```diff
@@ -137,6 +140,68 @@ impl<ReaderT: AsyncBufRead> Stream for CarStream<ReaderT> {
     }
 }
 
+pin_project! {
+    pub struct CarWriter<W> {
+        #[pin]
+        inner: W,
+        buffer: Vec<u8>,
```
lemmih, line 147: We want something like `Bytes` or `BytesMut` instead of `Vec<u8>`

src/utils/db/car_stream.rs
```diff
@@ -137,6 +140,68 @@ impl<ReaderT: AsyncBufRead> Stream for CarStream<ReaderT> {
...
+
+    fn poll_ready(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        let this = self.project();
+        if !this.buffer.is_empty() {
+            let bytes_written = ready!(this.inner.poll_write(cx, &this.buffer))?;
+            *this.buffer = this.buffer[bytes_written..].to_vec();
+            if !this.buffer.is_empty() {
+                return Poll::Pending;
+            }
+        }
```
lemmih, line 176:
```suggestion
        while !this.buffer.is_empty() {
            let bytes_written = ready!(this.inner.poll_write(cx, &this.buffer))?;
            *this.buffer = this.buffer[bytes_written..].to_vec();
        }
```

src/utils/db/car_stream.rs
```diff
@@ -137,6 +140,68 @@ impl<ReaderT: AsyncBufRead> Stream for CarStream<ReaderT> {
...
+        Poll::Ready(Ok(()))
+    }
+    fn start_send(self: Pin<&mut Self>, item: (Cid, Vec<u8>)) -> Result<(), Self::Error> {
+        let this = self.project();
+
+        let payload = [item.0.to_bytes(), item.1].concat();
+        let mut var_bytes = [0; 10];
+        let len = payload.len().encode_var(&mut var_bytes);
```
lemmih, line 184: We already have code for encoding a block.

```rust
let block = ...;
let buffer = ...;
block.write(&mut buffer)?;
```

elmattic: Is there some support for `impl std::io::Write` for `BytesMut`?

elmattic:
I had a failing `calibnet` check on the CI before. Not sure if related to those changes.

https://github.com/ChainSafe/forest/actions/runs/6121819227/job/16616724044

src/tool/subcommands/state_migration_cmd.rs
```diff
@@ -68,25 +68,22 @@ async fn generate_actor_bundle() -> Result<()> {
...
+
+    let stream = merge_car_streams(car_streams).map(|b| {
+        let b = b.expect("There should be no invalid blocks");
+        (b.cid, b.data)
+    });
+
+    let result = stream
+        .map(Ok)
+        .forward(CarWriter::new_carv1(all_roots, zstd_encoder)?)
+        .await;
```
lemmih, line 84:
```suggestion
    stream
        .map(Ok)
        .forward(CarWriter::new_carv1(all_roots, zstd_encoder)?)
        .await?;
    Ok(())
```

src/utils/db/car_stream.rs
lemmih: Hmm, maybe not. Then write to an intermediate vector.

src/utils/db/car_stream.rs
```diff
@@ -137,6 +140,68 @@ impl<ReaderT: AsyncBufRead> Stream for CarStream<ReaderT> {
...
+
+        this.buffer.extend_from_slice(&var_bytes[0..len]);
+        this.buffer.extend_from_slice(&payload);
+
+        Ok(())
+    }
+    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // TODO: find out if we really need to call poll_ready here?
```
lemmih, line 192: I think we have to call `poll_ready` here. Why do you think otherwise?

src/utils/db/car_stream.rs
```diff
@@ -137,6 +140,68 @@ impl<ReaderT: AsyncBufRead> Stream for CarStream<ReaderT> {
...
+        ready!(self.as_mut().poll_ready(cx))?;
+        self.project().inner.poll_flush(cx)
+    }
+    fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        ready!(self.as_mut().poll_ready(cx))?;
+
+        let this = self.project();
+        assert!(this.buffer.is_empty());
```
lemmih, line 200: This assertion can be removed. It's fairly simple to prove that it is always true.

src/utils/db/car_stream.rs
elmattic: Maybe `bytes::buf::Writer` can do the trick.

lemmih (changes requested):
Overall okay. Some tweaks are required.

Did you figure out the non-determinism issue? Try coming up with a minimal reproducible case.

elmattic:
> Did you figure out the non-determinism issue? Try coming up with a minimal reproducible case.

Not yet, reproducing it is easy though.

src/tool/subcommands/state_migration_cmd.rs
```diff
@@ -68,24 +68,18 @@ async fn generate_actor_bundle() -> Result<()> {
...
-            &mut std::pin::pin!(merge_car_streams(car_streams).map(|b| {
-                let b = b.expect("There should be no invalid blocks");
-                (b.cid, b.data)
-            })),
-        )
-        .await?;
+    );
 
+    let stream =
+        merge_car_streams(car_streams).map(|b| b.expect("There should be no invalid blocks"));
```
lemmih, line 77: Panic is not appropriate here.

lemmih (changes requested):

src/utils/db/car_stream.rs
elmattic: Quoting doc for `poll_flush` method:

> Returns Poll::Ready(Ok(())) when no buffered items remain. If this value is returned then it is guaranteed that all previous values **sent** via start_send have been flushed.

My understanding is that an item is **sent** in a two-step way. First `start_send` is called, then encoded item is written via `poll_write`.
So consider the sequence of calls: `start_send item1`, `poll_ready item1 -> ready`, `start_send item2`, `poll_flush`, we know we can return `Poll::Ready(Ok(()))` for flush because `item1` has been fully sent (which is not the case for `item2` yet).

But since I'm not sure about this reasoning I will keep the call to `poll_ready`.

src/tool/subcommands/state_migration_cmd.rs
```diff
@@ -68,24 +68,16 @@ async fn generate_actor_bundle() -> Result<()> {
...
-            })),
-        )
-        .await?;
+    );
 
+    let stream = merge_car_streams(car_streams);
+
+    stream
+        .forward(CarWriter::new_carv1(all_roots, zstd_encoder)?)
+        .await?;
```
lemmih, line 80:
```suggestion
    merge_car_streams(car_streams)
        .forward(CarWriter::new_carv1(all_roots, zstd_encoder)?)
        .await?;
```

lemmih (approved):
Looks good. We need to delete all references to `fvm_ipld_car` but that can be done in a follow-up PR.

elmattic:
I found the (one?) source of the non-determinism. Looks like if I remove the call to `poll_ready` in `poll_flush` method, I have a solid file size of `2432050` bytes (run the command around 20 times).

This is still puzzling me because the previous file created using `write_stream_async` method is giving me `2433114` bytes.

However I find that this file doesn't deflate well:
```
$ zstd -d actor_bundles.car.zst
ctor_bundles.car.zst : 0 B...     ctor_bundles.car.zst : Read error (39) : premature end 
$ zstd --test actor_bundles.car.zst 
ctor_bundles.car.zst : 0 B...     ctor_bundles.car.zst : Read error (39) : premature end 
```

This is not the case for the new one, and I have same size once deflated and checksum:
```
$ zstd -d actor_bundles.car.zst
actor_bundles.car.zst: 63329900 bytes                                          
$ sha256sum actor_bundles.car
2c5b55a53ab84bbb602cfe4005ad0a4ab38c84ca6298f93ec0a35794f7d4ffa8  actor_bundles.car
```

lemmih:
Could you write a QuickCheck test that passes data through CarWriter and CarStream, making sure the data isn't changed?

elmattic:
I've did some experiments and managed to match the `2433114` file size consistently (so the size we get using the `write_stream_async` method) by forcing a flush after `poll_write` so returning `self.project().inner.poll_flush(cx)` instead of `Poll::Ready(Ok(()))`. Had to remove `poll_shutdown` in close method though otherwise it panics (flush after shutdown error).

But then trying to deflate or `--test` the compressed file I get back to same error: `Read error (39) : premature end`.

aatifsyed:
I'm not sure about this implementation approach - I'll get back to you in a moment

aatifsyed (changes requested):
https://github.com/ChainSafe/forest/commit/4f8ce1ed47da8e828b029900e962d2bf43815d64

Can you have a look at this, and then we'll chat

lemmih:
> I'm not sure about this implementation approach - I'll get back to you in a moment

I think the approach is fine. Offloading frame encoding to `unsigned_varint/tokio-utils` is obviously preferable, but it's not a different approach.

aatifsyed (approved):
Let's refactor this once https://github.com/Nullus157/async-compression/issues/246 is fixed

Issue #3314, opened 2023-08-02
Walking the graph is slow


lemmih:
**Issue summary**

Exporting a snapshot is enough to be a problem. On my machine, I can export at ~25MiB/s. I believe the primary bottleneck is the graph traversal code.

Exporting a snapshot involves two parts:
- A depth-first search of the graph producing a list of blocks.
- Incrementally encoding the blocks into many z-frames.

The second part can be benchmarked by running `forest-cli snapshot compress`. On my machine, it runs at ~74 MiB/s. Benchmarking the traversal code is a bit more difficult, but it clocks in at ~65 MiB/s for me.

Let's try and make everything faster and more parallel.

Acceptance criteria:
- [x] DFS graph walking should produce at least 100 MiB/s of data.
- [ ] The ForestCAR.zst encoder should produce at least 100 MiB/s of data (if given sufficient input).
- [x] Unordered parallel graph traversal implementation. 

**Other information and links**

Useful commands:
- Download mainnet snapshot: `forest-cli snapshot fetch`
- Export snapshot `forest-cli archive export {INPUT.forest.car.zst} --output-path out.forest.car.zst`
- Compressing a snapshot `forest-cli snapshot compress {INPUT.forest.car.zst} --output out.forest.car.zst`
- Benchmark snapshot compression `cargo forest-tool benchmark forest-encoding {INPUT.forest.car.zst}`

lemmih:
See also `forest-tool benchmark`.

ruseinov:
We have the relevant benchmarks here: https://github.com/ChainSafe/forest/pull/3318 . Could be fixed as part of that PR.

lemmih:
Mostly done. No more low-hanging fruit.

PR #3474, opened 2023-09-08
[Feature] Unordered parallel graph walk V3

ruseinov:
## Summary of changes

- Introduced unordered parallel graph traversal to speed up operations that don't care about the block order.

Work on #3314 

## Change checklist


ruseinov:
```
unordered without "has"            traversed 87.43 GiB at 368.23 MiB/s in 00:04:03
unordered with "has"               traversed 87.43 GiB at 387.18 MiB/s in 00:03:51
unordered without parallel part    traversed 13.70 GiB at 252.89 MiB/s in 00:00:55
ordered                            traversed 87.43 GiB at 214.33 MiB/s in 00:06:57
ordered only sequential part       traversed 13.70 GiB at 257.50 MiB/s in 00:00:54
```

src/ipld/util.rs
```diff
@@ -286,14 +290,11 @@ impl<DB, T> ChainStream<DB, T> {
...
     }
-
-    pub fn into_seen(self) -> CidHashSet {
-        self.seen
-    }
 }
 
-/// Stream all blocks that are reachable before the `stateroot_limit` epoch. After this limit, only
-/// block headers are streamed. Any dead links are reported as errors.
+/// Stream all blocks that are reachable before the `stateroot_limit` epoch  in a depth-first
```
jdjaustin, line 295:
```suggestion
/// Stream all blocks that are reachable before the `stateroot_limit` epoch in a depth-first
```

src/ipld/util.rs
```diff
@@ -437,3 +438,259 @@ impl<DB: Blockstore, T: Iterator<Item = Tipset> + Unpin> Stream for ChainStream<
...
+
+/// Stream all blocks that are reachable before the `stateroot_limit` epoch in an unordered fashion.
+/// After this limit, only block headers are streamed. Any dead links are reported as errors.
+///
+/// # Arguments
+///
+/// * `db` - A database that implements [`Blockstore`] interface.
+/// * `tipset_iter` - An iterator of [`Tipset`], descending order `$child -> $parent`.
+/// * `stateroot_limit` - An epoch that signifies how far back we need to inspect tipsets.
+/// in-depth. This has to be pre-calculated using this formula: `$cur_epoch - $depth`, where
```
jdjaustin, line 477: I don't see this in this block. Is this an artifact from a previous iteration of this PR?

ruseinov: Not sure what the question is. The calculation happens on the client side, not anywhere in this implementation. Then we check epochs to avoid traversing blocks that we did not request. Does that make sense?

jdjaustin: Is there a misplaced period after `tipsets` and before the newline starting with `in-depth`?

ruseinov: yes!

jdjaustin (approved):

src/ipld/util.rs
```diff
@@ -437,3 +438,259 @@ impl<DB: Blockstore, T: Iterator<Item = Tipset> + Unpin> Stream for ChainStream<
         }
     }
 }
+
+enum UnorderedTask {
```
elmattic, line 442: Any reason to use an `enum` here?

src/ipld/util.rs
```diff
@@ -437,3 +438,259 @@ impl<DB: Blockstore, T: Iterator<Item = Tipset> + Unpin> Stream for ChainStream<
...
+#[allow(dead_code)]
+pub fn unordered_stream_chain<
+    DB: Blockstore + Sync + Send + 'static,
+    T: Iterator<Item = Tipset> + Unpin + Send + 'static,
+>(
+    db: Arc<DB>,
+    tipset_iter: T,
+    stateroot_limit: ChainEpoch,
+) -> UnorderedChainStream<DB, T> {
+    let (sender, receiver) = kanal::bounded(2048);
```
elmattic, line 488: Wondering how did you pick this `2048` value? Does the value has any impact on performances or memory usage?

src/ipld/util.rs
```diff
@@ -437,3 +438,259 @@ impl<DB: Blockstore, T: Iterator<Item = Tipset> + Unpin> Stream for ChainStream<
...
+        db: Arc<DB>,
+        block_sender: Sender<anyhow::Result<Block>>,
+        extract_receiver: Receiver<Cid>,
+        seen: Arc<Mutex<CidHashSet>>,
+        fail_on_dead_links: bool,
+    ) -> JoinHandle<anyhow::Result<()>> {
+        task::spawn(async move {
+            let mut handles = JoinSet::new();
+
+            for _ in 0..num_cpus::get() {
```
elmattic, line 561: Should I observe some speedup/slowdown if I increase/decrease the number of cores here? For example testing on a `calibnet` snapshot, I don't see much performance. Or should I run the benchmark command on a `mainnet` snapshot maybe?

src/ipld/util.rs
```diff
@@ -437,3 +438,259 @@ impl<DB: Blockstore, T: Iterator<Item = Tipset> + Unpin> Stream for ChainStream<
...
+                        this.queue.push(UnorderedTask::Cid(*block.cid()));
+
+                        if block.epoch() == 0 {
+                            // The genesis block has some kind of dummy parent that needs to be emitted.
+                            for p in &block.parents().cids {
+                                this.queue.push(UnorderedTask::Cid(p));
+                            }
+                        }
+
+                        // // Process block messages.
```
elmattic, line 654: Can remove comment here.

elmattic (approved):

src/ipld/util.rs
ruseinov: It's somewhat arbitrary, 2048 worked well on my machine and memory usage for 2048 items is negligible. I'll introduce a constant to be more transparent.

src/ipld/util.rs
ruseinov: Calibnet is too small of a snapshot, yeah. I have tested this and this setup worked better than others.

src/ipld/util.rs
ruseinov: It's a leftover, will remove it for good. Thanks for spotting this.

Issue #3498, opened 2023-09-18
`SmallCid` internals for collection types


aatifsyed:
`CidHashMap` was changed from this:
```rust
struct CidHashMap<V> {
    small: HashMap<[u8;32], V>,
    big: HashMap<Cid, V>
}
```
to this:
```
struct CidHashMap<V>(HashMap<SmallCid, V>)
```

in #3388 

That's wrong - we still care about saving overhead per-cid.

we need to restructure those changes, which looks like this I think:
- SmallCid stays, for use in FrozenCid. It's the right tradeoff here
- Have a public CompactCid, which is the implementation detail for SmallCid , for use in collection types

aatifsyed:
> We’re already bound by memory when manipulating Forest.CAR files. The size of the files we can handle is entirely determined by the available memory. Using 25% more memory means we can only generate Forest.CAR files that are 25% smaller.
- @lemmih, in Slack

So we care about this

aatifsyed:
# Using two `HashMap`s
![image](https://github.com/ChainSafe/forest/assets/38045910/5dd510b5-c70e-447e-a760-f58058fa8853)

# Using `HashMap<SmallCid, V>`
![image](https://github.com/ChainSafe/forest/assets/38045910/76d2f812-a92b-4d3a-b95e-dfda4ef50aff)

PR #3388, opened 2023-08-16
`CidVariant` refactor

jdjaustin:
## Summary of changes

Changes introduced in this pull request:

- Renamed `CidVariant` to `SmallCid` and modified type structure to hide enum details from users.

## Reference issue to close (if applicable)

Closes #3367 

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

PR #3299, opened 2023-07-31
Use `CidHashMap` in the plain CAR-file blockstore

jdjaustin:
## Summary of changes

Changes introduced in this pull request:

Replace maps used for write cache and index in `src/db/car/plain.rs` with `CidHashMap`.

- [x] Add missing methods to `CidHashMap`
- [x] Replace `HashMap<Cid, V>` with `CidHashMap<V>` in `car/plain.rs`
- [x] Measure change in peak memory use when validating an uncompressed calibnet and mainnet snapshot. See `/usr/bin/time -v forest-cli snapshot validate {snapshot}.car`

## Reference issue to close (if applicable)

Closes #3286 

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

PR #3217, opened 2023-07-17
`CidHashMap` implementation

jdjaustin:
## Summary of changes

Changes introduced in this pull request:

- [x] `CidHashMap` implementation with just enough methods to replace all occurrences of `HashMap<Cid, _>`.
- [x] Replace all occurrences of `HashMap<Cid, _>` with `CidHashMap<_>`.
- [x] Redefine `CidHashSet` to be `CidHashMap<Cid, ()>`.
- [x] Quickcheck properties to ensure that `CidHashMap<_>` has identical semantics to `HashMap<Cid,_>`.

## Reference issue to close (if applicable)

Closes associated task in #3202 

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

PR #3502, opened 2023-09-18
refactor: cid_collections

aatifsyed:
- https://github.com/ChainSafe/forest/pull/3217 Added `CidHashMap`
  ```rust
  struct CidHashMap<V> {
    small: HashMap<[u8; 32], V>,
    fallback: HashMap<Cid, V>
  }
  ```
- https://github.com/ChainSafe/forest/pull/3388 Introduced `SmallCid`
  ```rust
  enum SmallCid {
    Small([u8; 32]),
    Big(Box<Cid>),
  }
  ```
  And redefined `CidHashMap`
  ```rust
  struct CidHashMap<V>(HashMap<Cid, V>);
  ```
- https://github.com/ChainSafe/forest/pull/3299 Added the `Entry` API to `CidHashMap`

The refactor in #3388 cost us about ~20% overhead in map memory usage: https://github.com/ChainSafe/forest/issues/3498.

This refactor adds:
- a `cid_collections` module, containing `MaybeCompactedCid`, the core primitive, with tighter controls on canonicalisation and construction
- `SmallCid` is now an implementation detail of `FrozenCidVec`, NOT a general purpose util
- Reimplementations of `CidHashMap` and `CidHashSet`, with extended collections APIs


src/cid_collections/frozen_vec.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.inner.contains(&SmallCid::from(cid))
+    }
+}
+
+/// A [`MaybeCompactedCid`], with indirection to save space.
+///
+/// This is NOT a general purpose type - other collections should use the variants
+/// of [`MaybeCompactedCid`], so that the discriminant is not repeated.
+///
+/// (This saves ~20% in the [typical case](https://github.com/ChainSafe/forest/issues/3498#issuecomment-1723528854))
```
lemmih, line 35: This is either incorrect or misleading. There are a few different things at play here. Firstly, there's a gain in memory efficiency when going from plain `Cid` to `SmallCid`. Cids take up 96 bytes of memory, SmallCid takes 40 bytes. But using `CidHashSet` rather than `HashSet<SmallCid>` _also_ yields better efficiency at 32 bytes per entry rather than 40. This is the gain you measured in the link above.

Also, saying it saves ~20% unnecessarily complicates matters. The saving is exactly 25% for each CID. You measured ~24% because there are other allocated structures than just CIDs.

src/cid_collections/hash_map.rs
```diff
@@ -0,0 +1,321 @@
...
+};
+
+/// A space-optimised hashmap of [`Cid`]s.
+///
+/// We accept the implementation complexity of per-compaction-method HashMaps for
+/// the space savings.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CidHashMap<V> {
+    compact: ahash::HashMap<CidV1DagCborBlake2b256, V>,
+    uncompact: ahash::HashMap<Uncompactable, V>,
```
lemmih, line 18: This naming convention will be a bit awkward once we add support for more kinds of CIDs. But we can burn that bridge when we get to it.

src/cid_collections/hash_map.rs
```diff
@@ -0,0 +1,321 @@
...
+use cid::Cid;
+use std::collections::hash_map::{
+    Entry as StdEntry, IntoIter as StdIntoIter, OccupiedEntry as StdOccupiedEntry,
+    VacantEntry as StdVacantEntry,
+};
+
+/// A space-optimised hashmap of [`Cid`]s.
+///
+/// We accept the implementation complexity of per-compaction-method HashMaps for
+/// the space savings.
```
lemmih, line 14: This documentation is quite spartan. The previous docs elaborated more on our assumptions and expected space savings. Let's add documentation that is at least as detailed as it used to be.

src/cid_collections/hash_map.rs
```diff
@@ -0,0 +1,321 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CidHashMap<V> {
+    compact: ahash::HashMap<CidV1DagCborBlake2b256, V>,
+    uncompact: ahash::HashMap<Uncompactable, V>,
+}
+
+impl<V> CidHashMap<V> {
```
lemmih, line 21: All of these methods require documentation.

lemmih (changes requested):

src/cid_collections/hash_map.rs
aatifsyed: I think the documentation _is_ as detailed as it used to be, it's just moved to a common place.
It used to be repeated between e.g `CidHashMap`
https://github.com/ChainSafe/forest/blob/b0a2c2ca17edbff236810cacf3139d1123cf33ff/src/ipld/cid_hashmap.rs#L9-L19
and `SmallCid`
https://github.com/ChainSafe/forest/blob/b0a2c2ca17edbff236810cacf3139d1123cf33ff/src/utils/cid/mod.rs#L34-L46

Now there's _one_ place, and one struct with space - saving responsibility, and that's `MaybeCompactedCid`

I can add some design notes

src/cid_collections/hash_map.rs
aatifsyed: I would document these by forwarding to the standard library (I don't want to copy-paste the docs) - is that definitely what you want?

src/cid_collections/hash_map.rs
lemmih: But the space savings for `CidHashMap` are quite different than the space savings for `SmallCid`. There are space savings for using `Vec<SmallCid>` vs. `Vec<Cid>,` and there are space savings for using `CidHashSet` vs. `HashSet<SmallCid>`. You are accidentally mixing these up and writing factually incorrect documentation (such as the ~20% space savings for `SmallCid`).

lemmih (approved):

jdjaustin (approved):

PR #3504, opened 2023-09-19
[Feature] Mark-and-sweep GC. 

### LINKED ISSUE #3072: Implement a mark-and-sweep garbage collector
**Issue summary**

Context: We previously investigated the feasibility of using a mark-and-sweep garbage collector, but it was prohibitively expensive to scan values with RocksDB. New data shows that ParityDB can efficiently scan through all values. This opens the door for a more space-efficient mark-and-sweep GC.

Important caveats: We require state-roots going back `2 x chain_finality = 1800` epochs. Any data that is reachable within 1800 epochs may not be garbage collected.

The Forest database contains a persistent (i.e. immutable), directed acyclic graph. This makes implementing a mark-and-sweep garbage collector reasonably straightforward. The algorithm could look like this:

1. Scan through all IPLD values in the database and mark their hash value in a hash set. Collisions don't affect correctness, so the hash size matters little.
2. Wait for at least 1800 epochs after the scan.
3. Traverse the reachable graph from HEAD (including at least the last 1800 epochs) and remove reachable nodes from the hash set.
4. Scan the database again and remove all elements still in the hash set.

Notes:

- If there is a hash collision, some data may not be garbage collected. Hash collisions will never cause reachable data to be garbage collected.
- ParityDB does not store keys. This is fine in Forest since our keys are always the hash of the IPLD value.
- Make sure we no longer use [buffered writes](https://github.com/ChainSafe/forest/blob/4f77cf1c95a2ae4be08b984dbe6488ad3245f6bc/src/db/rolling/gc.rs#L226) and use [this approach](https://github.com/ChainSafe/forest/pull/3090) instead if applicable.

Some knowledge in regards to CID construction:
1. For the `GraphDagCborBlake2b256` parityDB column we currently use [this](https://github.com/ChainSafe/forest/blob/3bc34a63496b3f7f94e285f0910c61dd7af263d6/src/db/parity_db.rs#L404C39-L404C92) approach (V1, `DAG_CBOR` codec, Blake2b256 hash). We need to construct the CID manually, because this column does not have a btree_index therefore we have no way to iterate the keys. That's done due to performance reasons as `DAG_CBOR` is the most common type of key.
2. For the `GraphFull` column we have the btree-index so we can iterate the keys. Otherwise there is no way to know how to get the CID hash.

ruseinov:
## Summary of changes

- Introduces a mark-and-sweep garbage collector.

Closes #3072 

This will require a `0.13.1` release due to a database migration needed to deprecate RollingDB. 
Perhaps a good idea is to incorporate the `CHANGELOG.md` changes within this PR as well to simplify things. 

## Change checklist


ruseinov:
@lemmih This design is not taking into account the `AnyCar` addition. I'm assuming we do want to traverse the reachable graph including the car storage, correct?

lemmih:
> @lemmih This design is not taking into account the `AnyCar` addition. I'm assuming we do want to traverse the reachable graph including the car storage, correct?

The CAR files contain read-only data. There's no need for the GC to traverse them. In the future, we do want to find unused CAR files but you can leave that for another PR.

ruseinov:
> The CAR files contain read-only data. There's no need for the GC to traverse them. In the future, we do want to find unused CAR files but you can leave that for another PR.

Right, that makes sense.

ruseinov:
Currently blocked on some unused code. I will create a separate PR for this.

ruseinov:
The current state of GC seems fine, but it needs a data migration to work properly when using a database from a previous version. The reason is that we've moved away from RollingDB, so we need to merge the two databases it used under the hood.

I'm contemplating dropping estimated reachable records altogether, but we might still need that for our pseudo-gc-progressbar.

ruseinov:
Note that `progress api` has been removed due to the fact that it's no longer used.

src/utils/db/mod.rs
```diff
@@ -23,6 +22,9 @@ use tracing::info;
 /// for simplicity. The actual value for other underlying DB might be slightly
 /// different but that is negligible for calculating the total reachable data
 /// size
+/// NOTE: For some reason clippy detects this as unused, even though it is used
+/// below. Perhaps async-trait has something to do with it.
+#[allow(unused)]
```
lemmih, line 27: This should be a developer comment rather than a doc comment.

.config/forest.dic
```diff
@@ -98,3 +98,5 @@ VRF
 WebAssembly
 WebSocket
 zstd
+clippy
```
lemmih, line 101: Clippy shouldn't be part of our documentation if we can avoid it.
```suggestion
```

src/db/gc/mod.rs
```diff
@@ -0,0 +1,166 @@
+// Copyright 2019-2023 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
```
lemmih, line 2: Module-level documentation is required here. Should include:

- What problem are we trying to solve? (I.e., why is there garbage? What _is_ garbage?)
- Overview of the solution space. What other solutions exist, and why did we choose this particular solution?
- List the special constraints and domain knowledge. What kind of data are we dealing with? What does it mean that the graph is directed? What does it mean that the data is persistent? What does it mean that there can be multiple, concurrent branches in the blockchain? How does `chain finality` affect the GC?
- Performance characteristics. Include both algorithm complexities and empirical measurements. I.e., how much space is the GC expected to use? How much time does it take to do a GC run with a lite node where all the data is in ParityDB?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+use futures::StreamExt;
+use fvm_ipld_blockstore::Blockstore;
+use std::mem;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time;
+use tracing::info;
+
+// The average block time is set to a slightly bigger value on purpose, to save some cycles.
+const AVERAGE_BLOCK_TIME: Duration = Duration::from_secs(35);
```
lemmih, line 20: This is chain-specific. It should be taken from the chain configuration.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    async fn gc_workflow(&mut self, manual: bool) -> anyhow::Result<()> {
+        let depth = self.depth;
+        let tipset = self.chain_store.heaviest_tipset();
+        let current_epoch = tipset.epoch();
+        // Don't run the GC if there aren't enough state-roots yet. Manual GC overrides that.
+        if !manual || depth > current_epoch {
+            time::sleep(AVERAGE_BLOCK_TIME * (depth - current_epoch) as u32).await;
+            return anyhow::Ok(());
+        }
```
lemmih, line 132: I don't understand why manual GC overrides this check. If we aren't going to garbage collect anything, why run the garbage collector?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        // Make sure we don't GC if we haven't advanced at least `depth` number of epochs since
+        // the last sweep. Manual GC overrides that.
+        let epochs_since_gc = current_epoch - self.epoch_sweeped;
+        if self.marked.is_empty() && !manual && epochs_since_gc < depth {
```
lemmih, line 137: This check is too complicated for me to understand. Could you break it down into smaller pieces?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+            return anyhow::Ok(());
+        } else {
+            info!("populate keys for GC");
+            self.populate()?;
+        }
+
+        // Nothing to do.
+        if self.marked.is_empty() {
+            return anyhow::Ok(());
+        }
```
lemmih, line 148: This short circuit doesn't seem to serve a purpose. Let's remove it or add comments explaining why it is here.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        // Don't filter and sweep before we advance at least `depth`. Manual GC overrides that.
```
lemmih, line 150: Again, why does manually triggering the GC override these checks? It's a hard constraint that the GC may not remove any data generated within the last `chain_finality` epochs. This constraint must be followed regardless of how the GC is triggered.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        anyhow::Ok(())
+    }
+
+    // Remove marked keys from the database.
+    fn sweep(&mut self) -> anyhow::Result<()> {
+        let mut marked = HashSet::new();
+        mem::swap(&mut marked, &mut self.marked);
+        self.db.remove_keys(marked)
```
lemmih, line 91:
```suggestion
        let marked = mem::take(&mut self.marked);
        self.db.remove_keys(marked)
```

lemmih (changes requested):

src/db/gc/mod.rs
ruseinov: This can be removed.

src/db/gc/mod.rs
ruseinov: The current manual GC CLI interface expects the thing to run immediately. I reckon we should either be able to run the GC immediately assuming the user knows what they are doing or remove it for good.

src/db/gc/mod.rs
ruseinov: > Forks in the blockchain may occur in the last `chain_finality` epochs. The GC has to be aware of this. Write the module-level documentation, and then we can see if our mental models align.

Thank makes sense. As stated above, I think we have to get rid of the manual gc for good then. My initial implementation has taken this into account and the UX for such GC triggers is not amazing. We either have to change the GC cli to be `fire and forget` or get rid of it altogether.

src/db/gc/mod.rs
ruseinov: Since we are getting rid of the manual check one way or another - that'll make the whole thing fairly straighforward?

src/db/gc/mod.rs
ruseinov: As discussed in a different thread - we should either change the UX of the manual gc command or remove it for good. 
I didn't take forks into account, so we indeed need to employ the same rules for manual GC as we do for scheduled one.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+
+//!
+//! The current implementation of the garbage collector is concurrent mark-and-sweep.
+//!
+//! ## Design goals
+//! A correct GC algorithm that is simple and efficient for forest scenarios.
+//!
+//! ## GC Algorithm
+//! The `mark-and-sweep` algorithm was chosen due to it's simplicity, efficiency and low memory
+//! footprint. We do have to generate keys from values for some of the iterated columns due to a
```
lemmih, line 12: How ParityDB is iterated has nothing to do with the algorithm. It was mentioned in the previous GC because it directly affected the choice of algorithm.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//! each database key and storing those in a [`HashSet`].
+//! 2. Wait at least `chain finality` blocks.
+//! 3. Traverse reachable blocks starting from the current heaviest tipset and remove those from the
+//! marked `HashSet`, leaving only unreachable entries that are older than `chain finality` to avoid
+//! removing something that could later become reachable as a result of a fork.
+//! 4. Sweep, removing all the remaining marked entries from the database.
+//!
+//! ## Correctness
+//! This algorithm is traversing the reachable graph using the same tooling as the snapshot export.
+//! Therefore it ensures that we eliminate all the reachable from the set scheduled for removal.
```
lemmih, line 28: > Therefore it ensures that we eliminate all the reachable from the set scheduled for removal.

I can't parse this sentence. I think you're missing some words.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//! Additionally, it waits at least `chain finality` before filtering out and sweeping marked
```
lemmih, line 29: If you want to refer to `chain finality`, you must define it first.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//! 1. Mark: traverse all the relevant database columns, generating integer hash representations for
+//! each database key and storing those in a [`HashSet`].
+//! 2. Wait at least `chain finality` blocks.
+//! 3. Traverse reachable blocks starting from the current heaviest tipset and remove those from the
+//! marked `HashSet`, leaving only unreachable entries that are older than `chain finality` to avoid
+//! removing something that could later become reachable as a result of a fork.
+//! 4. Sweep, removing all the remaining marked entries from the database.
+//!
+//! ## Correctness
+//! This algorithm is traversing the reachable graph using the same tooling as the snapshot export.
```
lemmih, line 27: I feel you're trying to derive the GC algorithm's correctness from the snapshot-export code. This is doable, but you have to be much more deliberate in what you're trying to argue.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//! ## GC Algorithm
+//! The `mark-and-sweep` algorithm was chosen due to it's simplicity, efficiency and low memory
+//! footprint. We do have to generate keys from values for some of the iterated columns due to a
+//! ParityDB limitation. See <https://github.com/paritytech/parity-db/issues/187>.
+//! Previously the `semi-space` algorithm was used resulting in data duplication and up to 100%
+//! extra disk usage.
+//!
+//! ## GC Workflow
+//! 1. Mark: traverse all the relevant database columns, generating integer hash representations for
+//! each database key and storing those in a [`HashSet`].
```
lemmih, line 19: That they're stored in a `HashSet` is an implementation detail and doesn't have to be mentioned here.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//! This algorithm is traversing the reachable graph using the same tooling as the snapshot export.
+//! Therefore it ensures that we eliminate all the reachable from the set scheduled for removal.
+//! Additionally, it waits at least `chain finality` before filtering out and sweeping marked
+//! records, making sure nothing that could have become reachable in the meantime gets removed.
+//! A snapshot can be used to bootstrap the node from scratch, therefore the algorithm is considered
+//! correct when a valid snapshot can be exported using records available in the database after
+//! sweeping.
+//!
+//! ## Disk usage
+//! There's no additional disk space required to run this algorithm.
```
lemmih, line 36: Well, there's more to be said here. Will 10GiB of live data take up more than 10GiB of disk space? Are all unreachable keys guaranteed to be garbage collected?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//!
+//! ## Performance
+//! The GC Performance is calculated by benchmarking the three steps that have to be performed. The
+//! `filter` steps consists of two actions: walking the graph and filtering the `marked` set.
+//! 1. Traversing all the relevant database records and creating a set of keys. *To be benchmarked*
+//! 2. Walking the graph, this has already been benchmarked.
+//! `forest-tool benchmark graph-traversal`.
+//! WIP: fix and covert this to `unordered-graph-traversal`.
+//! 3. Filtering out the records found as a result of `step 2`.
+//! 4. Removing all the remaining `marked` records from the database *To be benchmarked*
```
lemmih, line 57: I don't know what you're trying to say with this section. It appears to have a different purpose than the `Performance` section of the previous GC.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//!
+//! ### Look up performance
+//! There should not be any noticeable look up penalty.
```
lemmih, line 60: The look-up overhead is provably zero. I don't think you should have this section at all.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//!
+//! ### Write performance
+//! The only thing that could affect write performance is DB re-index, which should not be affected
+//! much by the GC.
```
lemmih, line 64: Again, I don't see the purpose of this section.

lemmih (changes requested):

src/db/gc/mod.rs
ruseinov: Yeah, I kind of inherited the style a bit from the previous GC docs, because it kinda makes sense, but it definitely needs a better framework.

src/db/gc/mod.rs
ruseinov: `set` is a good enough then.

ruseinov:
I have removed the manual GC option along with `ProgressBar` and all it's mentions.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+//! footprint. Previously the `semi-space` algorithm was used resulting in data duplication and up
+//! to a 100% extra disk usage.
+//!
+//! ## GC Workflow
+//! 1. Mark: traverse all the relevant database columns, generating integer hash representations for
+//! each database key and storing those in a set.
+//! 2. Wait at least `chain finality` blocks.
+//! 3. Traverse reachable blocks starting at the current heaviest tipset and remove those from the
+//! marked set, leaving only unreachable entries that are older than `chain finality` to avoid
+//! removing something that could later become reachable as a result of a fork.
```
lemmih, line 25: > leaving only unreachable entries that are older than `chain finality` to avoid removing something that could later become reachable as a result of a fork.

This is important information. I feel this should go into the design goals as part of the problem description.

src/utils/db/mod.rs
```diff
@@ -23,6 +22,9 @@ use tracing::info;
 /// for simplicity. The actual value for other underlying DB might be slightly
 /// different but that is negligible for calculating the total reachable data
 /// size
+// NOTE: For some reason linter detects this as unused, even though it is used
+// below. Perhaps async-trait has something to do with it.
+#[allow(unused)]
```
lemmih, line 27: I want some clarity on this matter before we merge this PR.

src/utils/io/progress_log.rs
```diff
@@ -117,6 +117,7 @@ impl Progress {
         self.emit_log_if_required();
     }
 
+    #[allow(unused)]
```
lemmih, line 120: Delete it if it is unused.

lemmih (changes requested):

src/cli/subcommands/db_cmd.rs
```diff
@@ -1,22 +1,14 @@
...
+
 use clap::Subcommand;
 
-use crate::cli::subcommands::handle_rpc_err;
 use crate::utils::bail_moved_cmd;
 
 #[derive(Debug, Subcommand)]
 pub enum DBCommands {
-    /// Run DB garbage collection
-    GC,
```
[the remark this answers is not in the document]
ruseinov: Is there anything keeping us from doing a major release?

scripts/tests/harness.sh
```diff
@@ -44,7 +44,7 @@ function forest_query_format {
 
 function forest_run_node_detached {
   echo "Running forest in detached mode"
-  $FOREST_PATH --chain calibnet --encrypt-keystore false --log-dir "$LOG_DIRECTORY" --detach --save-token ./admin_token --track-peak-rss
+  $FOREST_PATH --chain calibnet --encrypt-keystore false --log-dir "$LOG_DIRECTORY" --no-gc --detach --save-token ./admin_token --track-peak-rss
```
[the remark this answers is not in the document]
ruseinov: This was done for manual GC testing, but since we don't have a manual GC anymore - it can be reverted.

src/cli/subcommands/db_cmd.rs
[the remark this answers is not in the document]
ruseinov: reinstated this for now.

src/utils/db/mod.rs
ruseinov: It turns out that neither this constant nor the function was used. I've removed both.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+//! 4. Sweep, removing all the remaining marked entries from the database.
+//!
+//! ## Correctness
+//! This algorithm considers all the blocks that are visited during the `snapshot export` task
+//! reachable, making sure they are kept in the database after the run. This is facilitated by
+//! waiting at least `chain finality` after the `mark` step before traversing the reachable graph
+//! and eliminating all the found blocks from the set marked for removal. A snapshot can be used to
+//! bootstrap the node from scratch, thus the algorithm is considered correct when a valid snapshot
+//! can be exported using records available in the database after sweeping.
```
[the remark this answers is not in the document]
ruseinov: Waiting at least `chain finality` is keeping all the unreachable blocks for `chain_finality` epochs. I agree, this is a bit long and confusing. I'll rephrase it.

src/cli/subcommands/db_cmd.rs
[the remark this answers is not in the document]
lemmih: > Is there anything keeping us from doing a major release?

Doing a major release would be fine.

LesnyRumcajs:
@ruseinov Please note that [v0.14.0 was released](https://github.com/ChainSafe/forest/pull/3594) so you will need to bump your version to 0.14.1. I have introduced some changes to the Migrator trait to make things easier for the future, and you may need to adapt your migration for that (if it proves difficult, please let me know).

ruseinov:
> @ruseinov Please note that [v0.14.0 was released](https://github.com/ChainSafe/forest/pull/3594) so you will need to bump your version to 0.14.1. I have introduced some changes to the Migrator trait to make things easier for the future, and you may need to adapt your migration for that (if it proves difficult, please let me know).

Roger that. Yeah, the release version will depend on whether or not we decide to fully deprecate the GC command as opposed to stubbing it out.

CHANGELOG.md
```diff
@@ -48,8 +52,6 @@
 - [#3422](https://github.com/ChainSafe/forest/issues/3422) Add NV21 (Watermelon)
   support for calibration network.
 
-### Changed
-
```
lemmih, line 52: Let's either remove all the empty sections or keep all of them.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
+version = "0.14.1"
```
lemmih, line 2: This field doesn't have to move.

src/cli/subcommands/db_cmd.rs
```diff
@@ -25,54 +17,16 @@ pub enum DBCommands {
         #[arg(long)]
         force: bool,
     },
+    // This is a noop as the manual GC is no longer available.
+    GC,
 }
 
 impl DBCommands {
-    pub async fn run(self, config: &Config) -> anyhow::Result<()> {
+    pub async fn run(self, _config: &Config) -> anyhow::Result<()> {
```
lemmih, line 25: Let's remove the `_config` parameter.

src/cli/subcommands/db_cmd.rs
```diff
@@ -25,54 +17,16 @@ pub enum DBCommands {
         #[arg(long)]
         force: bool,
     },
+    // This is a noop as the manual GC is no longer available.
+    GC,
```
lemmih, line 21: We can hide this command.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+//! `chain finality` - a number of epochs after which it becomes impossible to add or remove a block
+//! previously appended to the blockchain.
+//!
+//! ## Design goals
+//! A correct GC algorithm that is simple and efficient for forest scenarios.
+//!
+//! ## GC Algorithm
+//! The `mark-and-sweep` algorithm was chosen due to it's simplicity, efficiency and low memory
+//! footprint. Previously the `semi-space` algorithm was used resulting in data duplication and up
+//! to a 100% extra disk usage.
```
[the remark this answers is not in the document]
lemmih: My comment still stands. I see no reason to mention other implementations here.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+//! that could later become reachable as a result of a fork.
+//!
+//! ## GC Algorithm
+//! The `mark-and-sweep` algorithm was chosen due to it's simplicity, efficiency and low memory
+//! footprint. Previously the `semi-space` algorithm was used resulting in data duplication and up
+//! to a 100% extra disk usage.
+//!
+//! ## GC Workflow
+//! 1. Mark: traverse all the relevant database columns, generating integer hash representations for
+//! each database key and storing those in a set.
```
lemmih, line 23: I'd rather you didn't mention database columns here. Database columns are an implementation detail of the ParityDB backend. They do not exist in other backends, and the GC doesn't know about them.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+//! unreachable blocks from the database takes at least `chain finality`. The GC speed depends on
+//! the reachable graph size. Disk usage is also affected by the GC run interval. Additionally,
+//! since a truncated 4-byte hash is used - there's a slight possibility of a collision, which might
+//! result in an unreachable block being retained in the database. Still, the impact on the total
+//! disk size is negligible.
+//!
+//! ## Memory usage
+//! During the `mark` and up to the `sweep` stage, the algorithm requires `4 bytes` of memory for
+//! each database record. Additionally, the seen cache while traversing the reachable graph
+//! executing the `filter` stage requires at least `32 bytes` of memory for each reachable block.
```
lemmih, line 48: Let's put some real-world numbers on this. What is the expected memory usage with a 100GiB database (roughly the size of a mainnet snapshot in ParityDB)?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+//! 1. GC is triggered automatically and there have to be at least `chain finality` epochs stored
+//! for the `mark` step.
+//! 2. The `filter` step is triggered after at least `chain finality` has passed since the `mark`
+//! step.
+//! 3. Then, the `sweep` step happens.
+//! 4. Finally, the algorithm waits for a configured amount of time to initiate the next run.
+//!
+//! ## Performance
+//! Note: This needs to be measured and potentially defined in terms of `snapshot export` or any
+//! other visible and comparable metric.
```
lemmih, line 60: What can we say about the performance without doing any measurements? What are your expectations? If a node runs for a month, how many CPU seconds do you estimate the GC will use?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+
+/// [`MarkAndSweep`] is a simple garbage collector implementation that traverses all the database
+/// keys writing them to a [`HashSet`], then filters out those that need to be kept and schedules
+/// the rest for removal.
+///
+/// Note: The GC does not know anything about the hybrid CAR-backed and ParityDB approach, only
+/// taking care of the latter.
+pub struct MarkAndSweep<DB, GT> {
+    db: Arc<DB>,
+    get_heaviest_tipset: GT,
```
lemmih, line 94: How about `dyn Fn() -> Arc<Tipset>`?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+//! considered correct when a valid snapshot can be exported using records available in the database
+//! after the run.
+//!
+//! ## Disk usage
+//! There's no additional disk space required to run this algorithm. However, removing the
+//! unreachable blocks from the database takes at least `chain finality`. The GC speed depends on
+//! the reachable graph size. Disk usage is also affected by the GC run interval. Additionally,
+//! since a truncated 4-byte hash is used - there's a slight possibility of a collision, which might
+//! result in an unreachable block being retained in the database. Still, the impact on the total
+//! disk size is negligible.
```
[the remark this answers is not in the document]
ruseinov: Looks good to me, very precise. I would, however, rephrase `3` to be:
```
The blockstore backend may be fragmented, therefore not relinquishing the disk space back to the OS.
```

src/db/gc/mod.rs
ruseinov: I really don't want to box it and see no reason to. Why?

lemmih: It would be less code. Less is more. :)

src/db/gc/mod.rs
ruseinov: I wouldn't be able to speculate about this without actually measuring this. Also, it depends on the node specs.

lemmih: We've reasoned about the disk use and the memory usage. Why can't we reason about the CPU use? Is there nothing you can say about the time complexity of this algorithm?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+                ControlFlow::Continue => continue,
+                ControlFlow::Finished => {
+                    // Make sure we don't run the GC too often.
+                    time::sleep(interval).await;
+                }
+            }
+        }
+    }
+
+    async fn gc_workflow(&mut self) -> anyhow::Result<ControlFlow> {
```
lemmih, line 177: I find the `ControlFlow` part a bit confusing. It appears you're only using it for testing. I don't understand why, though. Let's discuss this in Slack tomorrow.

ruseinov: It's not for testing, it's used in order to be able to know whether or not we have to wait `interval` before restarting the cycle.

src/db/gc/mod.rs
ruseinov: In order to reason about CPU use we have to define that for graph walk algorithm first and that would depend on whether it's parallel or sequential. The rest is O(n).

src/db/gc/mod.rs
lemmih: Why not have `gc_workflow` always complete one iteration, then?

ruseinov: > Why not have `gc_workflow` always complete one iteration, then?

It felt easier at the time to yield to the main loop. But with removal of `manual gc` this might not be
needed anymore.

src/db/gc/mod.rs
ruseinov: Let me see what comes out of it.

src/db/gc/mod.rs
ruseinov: I'll contrive something.

src/ipld/util.rs
```diff
@@ -468,6 +372,11 @@ impl<DB, T> UnorderedChainStream<DB, T> {
             Err(v) => v.lock().clone(),
         }
     }
+
+    #[allow(dead_code)]
+    pub async fn join_workers(self) -> anyhow::Result<()> {
+        self.worker_handle.await?
+    }
```
lemmih, line 379: Why do we need dead code?

ruseinov: This is needed for the parallel graph traversal. That is being debugged at the moment.

src/db/gc/mod.rs
ruseinov: This actually needs to be measured.

src/ipld/util.rs
ruseinov: fixed

ruseinov:
@lemmih all of the PR comments have been addressed.

src/db/gc/mod.rs
ruseinov: After removing `ControlFlow` this seems easier to read. I still stand by the current approach that yields to the main loop as it's much cleaner than having to write inner loops and introducing code repetition and mutable variables.

src/db/gc/mod.rs
ruseinov: Done

src/db/gc/mod.rs
lemmih: > In order to reason about CPU use we have to define that for graph walk algorithm first and that would depend on whether it's parallel or sequential. The rest is O(n).

What's the time complexity of a depth-first graph traversal? And what's the time complexity of a parallel depth-first graph traversal?

ruseinov: Both are O(V+E), where v is the number of vertices and e is the number of edges.

.config/forest.dic
```diff
@@ -99,3 +99,6 @@ VRF
 WebAssembly
 WebSocket
 zstd
+ParityDB
+benchmark/GD
++
```
lemmih, line 104: We need a plus sign in the dictionary?

```suggestion
```

src/cli_shared/cli/client.rs
```diff
@@ -72,8 +71,6 @@ pub struct Client {
         |g| Duration::milliseconds(i64::arbitrary(g))
     )))]
     pub token_exp: Duration,
-    /// Display progress bars mode. Auto will display if TTY.
-    pub show_progress_bars: ProgressBarVisibility,
```
lemmih, line 76: This change is not backward compatible. Let's bump the major version number.

src/daemon/mod.rs
```diff
@@ -229,28 +231,26 @@ pub(super) async fn start(
         genesis_header.clone(),
     )?);
 
-    let db_garbage_collector = {
-        let db = db.clone();
+    let mut db_garbage_collector = {
```
lemmih, line 234: Can this be moved into the `if !opts.no_gc {` block?

src/daemon/mod.rs
```diff
@@ -343,14 +343,14 @@ pub(super) async fn start(
...
 
-        let gc_event_tx = db_garbage_collector.get_tx();
         services.spawn(async move {
             info!("JSON-RPC endpoint started at {}", config.client.rpc_address);
             let beacon = Arc::new(
                 rpc_state_manager
                     .chain_config()
                     .get_beacon_schedule(chain_store.genesis().timestamp()),
             );
+            info!("JSON-RPC endpoint started at {}", config.client.rpc_address);
```
lemmih, line 353: Duplicate.

```suggestion
```

src/db/gc/mod.rs
lemmih: Great!

src/db/gc/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///
+    /// * `depth` - Specifies how far back the full history should be maintained. Cannot be less
+    /// than chain finality.
+    /// * `interval` - GC Interval to avoid constantly consuming node's resources.
+    ///
+    /// NOTE: This currently does not take into account the fact that we might be starting the node
+    /// using CAR-backed storage with a snapshot, for implementation simplicity.
+    pub async fn gc_loop(&mut self, interval: Duration) -> anyhow::Result<()> {
+        loop {
+            self.gc_workflow(interval).await?
```
lemmih, line 160: This code will do a GC run every time Forest is started. Let's make sure no more than 1 GC run happens in `interval` amount of time. You can do this by sleeping _before_ calling `gc_workflow` rather than _after_`.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    // This function yields to the main GC loop if the conditions are not met for execution of the
+    // next step.
+    async fn gc_workflow(&mut self, interval: Duration) -> anyhow::Result<()> {
+        let depth = self.depth;
+        let tipset = (self.get_heaviest_tipset)();
+        let current_epoch = tipset.epoch();
+        // Don't run the GC if there aren't enough state-roots yet. Sleep and yield to the main loop
+        // in order to refresh the heaviest tipset value.
+        if depth > current_epoch {
+            time::sleep(self.block_time * (depth - current_epoch) as u32).await;
```
lemmih, line 173: If `depth > current_epoch`, you can just sleep for `interval` amount of time. No need to be any more precise than that.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    // Filter out the initial set, leaving only the entries that need to be removed.
+    // NOTE: One concern here is that this is going to consume a lot of CPU.
+    async fn filter(&mut self, tipset: Arc<Tipset>, depth: ChainEpochDelta) -> anyhow::Result<()> {
+        // NOTE: We want to keep all the block headers from genesis to heaviest tipset epoch.
+        let mut stream = stream_graph(
+            self.db.clone(),
+            (*tipset).clone().chain(self.db.clone()),
+            depth,
+        );
```
lemmih, line 132: We should warn on dead links. I'll add a separate issue for that.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+            db.put_cbor_default(&block2).unwrap();
+
+            let tipset = Arc::new(Tipset::from(&block2));
+            cs.set_heaviest_tipset(tipset).unwrap();
+            heaviest_tipset = cs.heaviest_tipset();
+        }
+    }
+
+    #[tokio::test]
+    async fn test_populate() {
```
lemmih, line 241: Could you document this test? I'm unsure what it is testing.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        // test marked
+        run_to_epoch(db, cs, depth);
+        gc.gc_workflow(interval).await.unwrap();
+        assert_eq!(gc.marked.len(), 2);
+        assert_eq!(gc.epoch_marked, 1);
+    }
+
+    #[tokio::test]
+    async fn test_filter_and_sweep() {
```
lemmih, line 268: This is a rather long test. Can it be broken down into smaller tests?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    use crate::networks::ChainConfig;
+
+    use crate::utils::db::CborStoreExt;
+
+    use core::time::Duration;
+
+    use crate::shim::clock::ChainEpoch;
+    use fvm_ipld_blockstore::Blockstore;
+    use std::sync::Arc;
+
```
lemmih, line 219: How about a quickcheck test that GCs a blockchain with N blocks and a depth of D?

lemmih (changes requested):
Let's pair on the module-level documentation. Still a few loose ends.

.config/forest.dic
ruseinov: Yes, this is needed for e.g. `O(V+E)` comments.

src/db/gc/mod.rs
ruseinov: It's a long test, because we need to `mark` before `filter` and `sweep`. It can be refactored into smaller things, but it's still going to have to run them all.

src/db/gc/mod.rs
ruseinov: Sure.

.config/forest.dic
lemmih: If you use back ticks (like you just did in your comment) then the spellchecker will leave it alone.

src/db/gc/mod.rs
ruseinov: What do you mean by that? As in we let quickcheck generate N and D for us?

ruseinov: It would have to be within a certain range, but sure, we can do it.

src/db/gc/mod.rs
ruseinov: I'm sleeping within the `gc_workflow`, but it does happen before each GC run now.

ruseinov:
> Let's pair on the module-level documentation. Still a few loose ends.

Sure, any suggestions? 

If it's possible to merge this and do this as a follow-up - that would also work, as the amount of code this touches results in many merge conflicts when the new stuff hits "main".

src/db/gc/mod.rs
ruseinov: I have looked into that, the issue is that quickcheck does not currently support async functions. There is a `quckcheck_async` library, but I have failed to make it work so far.

ruseinov: Done.

.config/forest.dic
ruseinov: Thanks, I did not know that.

src/cli_shared/cli/client.rs
ruseinov: done

src/db/gc/mod.rs
lemmih: It's hard for me to see exactly what you're testing here. If this test started failing, I wouldn't know what was wrong. If it can't be broken down into smaller pieces, could you describe exactly what you're testing and why you believe the assertions must hold true?

src/db/gc/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+        assert_eq!(gc.marked.len(), 0);
+        assert_eq!(db.get_keys().unwrap().len(), reachable_cnt as usize);
+
+        // try another run
+        gc.gc_workflow(interval).await.unwrap();
+        assert_eq!(gc.marked.len(), db.get_keys().unwrap().len());
+    }
+
+    #[quickcheck_async::tokio]
+    async fn test_workflow(depth: u8, current_epoch: u8, unreachable_cnt: u8) {
```
lemmih, line 319: This test is too long for me to understand. What properties are you testing? Are you testing more than one property? If so, can they be tested separately?

Ideally, quickcheck tests would map almost directly to properties or statements that we believe to be true. For example, we believe that no state trees within `depth` epochs will be garbage collected. We believe none of the block headers reachable from HEAD will _ever_ be garbage collected. We believe that block headers NOT reachable from HEAD will be garbage collected _after_ `chain_finality` epochs. These are invariants that should be true and it would be great to test them separately. Note, testing isn't primarily to find bugs in your code. The primary purpose of testing is to convince developers (others and yourself) that your believes about your code are correct.

lemmih (changes requested):

src/db/gc/mod.rs
ruseinov: Right, that makes sense. Let's test them separately. Yeah, the test is a bit long, but I have an idea on how to simplify it.

lemmih (approved):
I feel the quickcheck tests are still more complicated than I can easily wrap my head around. And some tests don't appear to cover a uniform set of scenarios. For example, the `test_populate` makes sure the GC works as expected at epoch 0 and at epoch `depth`. But it doesn't check anything in between. It wouldn't catch an off-by-one error in the GC.

Most of the properties are easy to write in English in a single line and I wish the code wouldn't have to be much longer than that. For example, we have the property `garbage collecting must never remove the spine of the blockchain`. Ideally this could be written as:
```rust
heaviest_tipset.chain(&db_before_gc) == heaviest_tipset.chain(&db_after_gc)
```

Let's merge this PR as-is and see if we get the bandwidth to improve the tests later.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+        assert_eq!(gc.epoch_marked, depth);
+    }
+
+    struct GCTester {
+        db: Arc<MemoryDB>,
+        store: Arc<ChainStore<MemoryDB>>,
+    }
+
+    #[quickcheck_async::tokio]
+    async fn no_reachable_data_cleanups(depth: u8, current_epoch: u8) {
```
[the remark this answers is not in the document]
ruseinov: That we don't gc reachable data. I'll rephrase

ruseinov:
> Let's merge this PR as-is and see if we get the bandwidth to improve the tests later.

Sounds good!

src/db/gc/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+}
+
+impl<DB: Blockstore + GarbageCollectable> MarkAndSweep<DB> {
+    /// Creates a new mark-and-sweep garbage collector.
+    ///
+    /// # Arguments
+    ///
+    /// * `db` - A reference to the database instance.
+    /// * `get_heaviest_tipset` - A function that facilitates heaviest tipset retrieval.
+    /// * `depth` - The number of state-roots to retain.
```
elmattic, line 109: What should `depth` be initialized to? In the associated issue we're speaking of a number of `2 x chain_finality`, here we're taking max of `chain_finality` and `recent_state_roots`.

src/db/gc/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+    fn sweep(&mut self) -> anyhow::Result<()> {
+        let marked = mem::take(&mut self.marked);
+        self.db.remove_keys(marked)
+    }
+
+    /// Starts the Garbage Collection loop.
+    ///
+    /// # Arguments
+    ///
+    /// * `depth` - Specifies how far back the full history should be maintained. Cannot be less
```
elmattic, line 156: We only have an `interval` argument here, but you are mentioning a `depth` argument as well.

elmattic:
Should this new implementation of GC fix our long-standing send issue? (https://github.com/ChainSafe/forest/issues/3645)
I would really like to run my script, but it would not be possible without a way to trigger your GC.

Eventually, let's try to uncomment the test in `calibnet_wallet_check.sh` and see what happens.

ruseinov:
> Should this new implementation of GC fix our long-standing send issue? (#3645) I would really like to run my script, but it would not be possible without a way to trigger your GC.
> 
> Eventually, let's try to uncomment the test in `calibnet_wallet_check.sh` and see what happens.

Manual GC is deprecated, it makes no sense with the current approach.

src/db/gc/mod.rs
ruseinov: yep, outdated docs. Good spot!

src/db/gc/mod.rs
ruseinov: It's not garbage collector's concern, the decision is made by the caller. But it's a good catch, and I will comment and fix it to be 2x chain_finality.

elmattic (approved):

ruseinov:
> @ruseinov You probably have to update the branch before the CI checks will run.

so it seems!

lemmih (approved):

PR #3505, opened 2023-09-19
chore: remove stream seek constraint

lemmih:
## Summary of changes

- Remove `AsyncSeek` constraint for `CarStream`. The seek constraint prevents streaming data from, say, an HTTP server.

Closes

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -184,6 +184,7 @@ url = { version = "2.3", features = ["serde"] }
 uuid = { version = "1.3", features = ['v4'] }
 walkdir = "2"
 zstd = "0.12.3"
+zstd-safe = "6.0"
 
 [target.'cfg(unix)'.dependencies]
 termios = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -184,7 +184,6 @@ url = { version = "2.3", features = ["serde"] }
 uuid = { version = "1.3", features = ['v4'] }
 walkdir = "2"
 zstd = "0.12.3"
-zstd-safe = "6.0"
 
 [target.'cfg(unix)'.dependencies]
 termios = "0.3"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -184,6 +184,7 @@ url = { version = "2.3", features = ["serde"] }
 uuid = { version = "1.3", features = ['v4'] }
 walkdir = "2"
 zstd = "0.12.3"
+zstd-safe = "6.0"
```
hanabi1224, line 187: `zstd-safe` has been re-exported from `zstd`

https://docs.rs/zstd-safe/6.0.6/zstd_safe/index.html

lemmih:
It's technically possible to craft an uncompressed CAR file which will be identified as a zstd compressed file. However, no valid _Filecoin_ CAR file can be misidentified in this way.

src/utils/db/car_stream.rs
```diff
@@ -12,10 +12,10 @@ use fvm_ipld_encoding::to_vec;
 use integer_encoding::VarInt;
 use pin_project_lite::pin_project;
 use serde::{Deserialize, Serialize};
-use std::io::{self, Cursor, SeekFrom};
+use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
-use tokio::io::{AsyncBufRead, AsyncRead, AsyncSeek, AsyncSeekExt, AsyncWrite};
```
hanabi1224, line 18: Do we still need `RandomAccessFile` after removing `AsyncSeek` constraint?

hanabi1224: Nvm it does not seem to be related

hanabi1224 (approved):

elmattic (approved):
Do we have a test that exercice the non-compressed path?

lemmih:
> Do we have a test that exercice the non-compressed path?

Added a test using the uncompressed genesis CAR file.

Issue #2884, opened 2023-05-16
[CI] Enable sccache in docker build and publish ci job


hanabi1224:
**Issue summary**

Since we have switched `sccache` backend to DO space, we can enable `sccache` in `build` stage with below changes

- [ ] Implement a build arg flag (e.g. SCCACHE_ENABLED=true) for enabling `sccache`
- [ ] Set `RUSTC_WRAPPER=sccache` when the flag above is on
- [ ] Download and install proper prebuilt `sccache` binaries for `amd64-linux` and `arm64-linux` respectively
- [ ] Pass `sccache` `s3` env args via build arg

To convert build arg to env var
```Dockerfile
ARG SCCACHE_ENDPOINT
ENV SCCACHE_ENDPOINT=$SCCACHE_ENDPOINT
```

To disable/enable sccache
```Dockerfile
ARG SCCACHE_ENABLED
# https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
ENV RUSTC_WRAPPER=${SCCACHE_ENABLED:+sccache}
```

**Other information and links**

LesnyRumcajs:
@hanabi1224 Is it the same as https://github.com/ChainSafe/forest/issues/2519? If so, let's close one of them.

lemmih:
@hanabi1224 This is done, right?

LesnyRumcajs:
@hanabi1224 bump

hanabi1224:
@lemmih @LesnyRumcajs No I don't think we have enabled `sccache` in docker build. I can prioritize this task

LesnyRumcajs:
Do we need it? I mean, is there a compelling reason to have this? What time gains are we talking about?

hanabi1224:
@LesnyRumcajs  I think the major benefit is that, we could save some $ by switching from buildjet to GH hosted runners

hanabi1224:
@LesnyRumcajs I believe this has been implemented in https://github.com/ChainSafe/forest/pull/3510 but superseded by the `buildjet` to `fuzzy` switch.

LesnyRumcajs:
We're not running this on `fuzzy` anymore, given it introduced some nasty race conditions.

lemmih:
If this task had been easy, it would already have been done. Closing as our time is worth more than what we spend on buildjet.

PR #3510, opened 2023-09-21
feat: enable sccache for docker build

hanabi1224:
## Summary of changes

- Add optional `sccache` support do `Dockerfile` to speed up CI build
- Switch to GH-hosted runner on CI (Or do we want to stay with buildjet? @lemmih @LesnyRumcajs )

The CI job takes ~16min now on GH-hosted runner, see https://github.com/ChainSafe/forest/actions/runs/6260502935

Closes https://github.com/ChainSafe/forest/issues/2884

## Change checklist


lemmih:
Is this still WIP?

hanabi1224:
> Is this still WIP?

No, title updated

Dockerfile
```diff
@@ -33,6 +33,28 @@ RUN update-ca-certificates
...
+ENV SCCACHE_REGION=$SCCACHE_REGION
+ARG AWS_ACCESS_KEY_ID
+ENV AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID
+ARG AWS_SECRET_ACCESS_KEY
+ENV AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
+
+ENV SCCACHE_VERSION=0.5.4
+RUN curl -sOL "https://github.com/mozilla/sccache/releases/download/v${SCCACHE_VERSION}/sccache-v${SCCACHE_VERSION}-$(uname -m)-unknown-linux-musl.tar.gz" && \
+    tar xzf "sccache-v${SCCACHE_VERSION}-$(uname -m)-unknown-linux-musl.tar.gz" && \
+    cp "sccache-v${SCCACHE_VERSION}-$(uname -m)-unknown-linux-musl/sccache" /usr/local/bin/sccache
```
LesnyRumcajs, line 51: if `SCCACHE_ENABLED` is not set, we should not download it.

.github/workflows/docker.yml
```diff
@@ -22,15 +22,15 @@ jobs:
     # This is done to limit the runner cost.
     if: github.event.pull_request.draft == false
     name: Build images and push to GHCR
-    runs-on: buildjet-8vcpu-ubuntu-2204
-    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
```
LesnyRumcajs, line 26: Are you positive that a clean build on `ubuntu-latest` will not timeout on this? E.g., when the toolchain is updated?

LesnyRumcajs: Note that cross-compilation for ARM takes _a lot_ of time, did you verify the build with it?

Dockerfile
LesnyRumcajs, line 51: it's not available in apt repositories?

.github/workflows/docker.yml
hanabi1224: Hmm, I'm not sure actually. But a slow clean build seems to be a deal breaker, should we stick on buildjet?

Dockerfile
hanabi1224: I don't think so

hanabi1224: Fixed

.github/workflows/docker.yml
hanabi1224: > Note that cross-compilation for ARM takes a lot of time, did you verify the build with it?

@LesnyRumcajs Yes the CI job for PR is targeting arm64, which takes ~18min on GH-hosted runner with cache.

```yaml
# Compile Docker image only for ARM64 for a regular PR to save some CI time.
platforms: ${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')) && 'linux/amd64,linux/arm64' || 'linux/arm64' }}
```

LesnyRumcajs: Could you please show me a build where an `arm64` image was built from scratch?

LesnyRumcajs: If we are not using buildjet then it's fine to have a generous timeout - we don't pay for it (though we occupy the workers which are limited)

hanabi1224: > Could you please show me a build where an arm64 image was built from scratch?

@LesnyRumcajs I don't have one yet, and I don't think it will be fast enough for a job that is enabled for every PR, otherwise we'd already use GH-hosted runner : )  And since this is a valid case, should we switch back to buildjet as sccache could still save some less $ for buildjet . Does that make sense?

Dockerfile
LesnyRumcajs: Indeed, it seems to be supported only from ubuntu 23 onward

.github/workflows/docker.yml
LesnyRumcajs: Let's first get some numbers:
1. Clean build for `amd64` on `ubuntu-latest`. Reasoning: we want to know the worst-case scenario for a regular build on a branch.
2. Consecutive build for `amd64`.
3. Clean build for `arm64` on `ubuntu-latest`. 
4. Consecutive builds for `arm64`.

Then let's compare it with the current state-of-the-art.

A possible solution, to use something from both worlds, is to use the GH worker for PRs and buildjet for builds involving cross-compilation.

LesnyRumcajs: Does it make sense?

lemmih:
What's the status of this PR?

hanabi1224:
@lemmih The PR works with `fuzzy` but I'm not sure if we still need `sccache` since the local cache now works with the single runner.

```Dockerfile
# Install Forest. Move it out of the cache for the prod image.
RUN --mount=type=cache,sharing=private,target=/root/.cargo/registry \
    --mount=type=cache,sharing=private,target=/root/.rustup \
    --mount=type=cache,sharing=private,target=/forest/target \
    make install-xx && \
    mkdir /forest_out && \
    cp /root/.cargo/bin/forest* /forest_out
```

lemmih:
> @lemmih The PR works with `fuzzy` but I'm not sure if we still need `sccache` since the local cache now works with the single runner. (And self-hosted runner does not clean local FS as per [doc](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security))
> 
> ```dockerfile
> # Install Forest. Move it out of the cache for the prod image.
> RUN --mount=type=cache,sharing=private,target=/root/.cargo/registry \
>     --mount=type=cache,sharing=private,target=/root/.rustup \
>     --mount=type=cache,sharing=private,target=/forest/target \
>     make install-xx && \
>     mkdir /forest_out && \
>     cp /root/.cargo/bin/forest* /forest_out
> ```
> 
> > This is not an issue with GitHub-hosted runners because each GitHub-hosted runner is always a clean isolated virtual machine, and it is destroyed at the end of the job execution.

So, should we close this PR?

PR #3538, opened 2023-10-03
[Fix] Pluck out kanal in favour of flume.

ruseinov:
## Summary of changes

- Get rid of `kanal` dependency in favour of `flume`.

Closes

Unfortunately `kanal` is not mature enough to be used in production or otherwise. It gets infinitely stuck if awaiting on a channel that's already been closed. 

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2865,7 +2865,6 @@ dependencies = [
  "itertools 0.11.0",
  "jsonrpc-v2",
  "jsonwebtoken",
- "kanal",
  "libc",
  "libipld",
  "libipld-core",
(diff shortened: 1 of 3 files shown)


ruseinov:
I think we can probably drop this for now. `kanal` does give a slight performance boost for parallel graph walk over `flume` and the current solution takes care of the problem I have encountered.

ruseinov:
> I think we can probably drop this for now. `kanal` does give a slight performance boost for parallel graph walk over `flume` and the current solution takes care of the problem I have encountered.

That said, I would refrain from using `kanal` on other occasions for now. It seems to have a bunch of open issues that need to be resolved first.

ruseinov:
I'm going to close this for now.

PR #3548, opened 2023-10-04
Simplify error shim

### LINKED ISSUE #3526: Refactor `ErrorNumber` to simplify the code
**Issue summary**

> `ErrorNumber` can be radically simplified, but that's for another time.  

https://github.com/ChainSafe/forest/pull/3520/#issuecomment-1737893185

elmattic:
## Summary of changes

- Store simply the error number in error shim

Closes #3526

## Change checklist


lemmih (changes requested):
We still want an enum, but `UnknownErrorNumber` can be simplified.

Ok so keeping the `error_number!`. Simplified by using an `u32` instead?

elmattic:
> We still want an enum, but `UnknownErrorNumber` can be simplified.

Ok so keeping the `error_number!`. Simplified by using an `u32` instead?

lemmih:
> > We still want an enum, but `UnknownErrorNumber` can be simplified.
> 
> Ok so keeping the `error_number!`. Simplified by using an `u32` instead?

Yes. We don't care which FVM version the errors came from.

aatifsyed (commented):
The old design preserves the name and message of the underlying error, which is protected by `#[non_exhaustive]`. Bit of an edge, case, but it's more correct, not?

Also the assertions on N4 need updating

lemmih:
> The old design preserves the name and message of the underlying error, which is protected by `#[non_exhaustive]`. Bit of an edge, case, but it's more correct, not?
> 
> Also the assertions on N4 need updating

It's a judgment call. Preserving the names of unknown errors is not worth the complexity.

src/shim/kernel.rs
```diff
@@ -96,7 +70,10 @@ error_number! {
 
 impl fmt::Display for ErrorNumber {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.as_unshimmed().fmt(f)
+        match self {
+            Self::Unknown(u) => write!(f, "{}", u),
+            e => write!(f, "{}", e),
+        }
```
lemmih, line 76: Are you sure this works? Looks like an infinite loop to me.

elmattic: Testing it.

elmattic: Fixing it.

src/shim/kernel.rs
```diff
@@ -96,7 +71,21 @@ error_number! {
...
+            Self::AssertionFailed => f.write_str("filecoin assertion failed"),
+            Self::InsufficientFunds => f.write_str("insufficient funds"),
+            Self::NotFound => f.write_str("resource not found"),
+            Self::InvalidHandle => f.write_str("invalid ipld block handle"),
+            Self::IllegalCid => f.write_str("illegal cid specification"),
+            Self::IllegalCodec => f.write_str("illegal ipld codec"),
+            Self::Serialization => f.write_str("serialization error"),
+            Self::Forbidden => f.write_str("operation forbidden"),
+            Self::BufferTooSmall => f.write_str("buffer too small"),
+            Self::Unknown(u) => write!(f, "{}", u),
```
aatifsyed, line 87: I don't like this - is it a pasted block from fvm?
It's not our responsibility to define these strings - can you think of a way to defer definitions to fvm?

elmattic: Yes, will try to find something better.

elmattic: @aatifsyed Is my new approach better?

src/shim/kernel.rs
```diff
@@ -8,77 +9,52 @@ use fvm4::kernel::SyscallError as E4;
...
 use std::fmt;
 
 macro_rules! error_number {
     ($($variant:ident),* $(,)?) => {
         #[derive(Debug, Clone)]
         #[non_exhaustive]
         pub enum ErrorNumber {
-            $($variant,)*
-            Unknown(UnknownErrorNumber),
+            $($variant(u32),)*
```
lemmih, line 20: The enum variants should not take a `u32`. They statically map to a single number.

lemmih (changes requested):

src/shim/kernel.rs
elmattic: I removed them. Where I am still unsure of is the need of the `from` function.

src/shim/kernel.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
+
```
lemmih, line 3:
```suggestion
```

src/shim/kernel.rs
```diff
@@ -136,3 +121,25 @@ impl From<E4> for EShim {
...
+        assert_eq!(
+            format!("{}", shim),
+            "syscall error: cthulhu (exit_code=illegal argument)"
+        );
+    }
+
+    #[test]
+    fn test_unknown_error_fmt() {
+        let shim: EShim = E3::new(N3::ReadOnly, "dagon").into();
```
lemmih, line 141: `ReadOnly` should be added to our list of errors.

src/shim/kernel.rs
```diff
@@ -136,3 +121,25 @@ impl From<E4> for EShim {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_error_fmt() {
```
lemmih, line 130: Why not directly test `ErrorNumber` rather than `SyscallError`?

lemmih (changes requested):

src/shim/kernel.rs
```diff
@@ -17,7 +17,7 @@ macro_rules! error_number {
         #[derive(Debug, Clone)]
         #[non_exhaustive]
         pub enum ErrorNumber {
```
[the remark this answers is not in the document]
aatifsyed: I don't think it should be - enum layout is only useful for:
- casting to primitives. This can never be cast since it's got a non-primitive variant, the unknown one
- FFI layout

It would also be incorrect - the fvm `#[repr]`s correspond to the actual error code, but if anyone read the discriminant, it would be incorrect in the unknown case.

ruseinov:
Having taken a look at the macro, I like the fact that it goes through all the available variants and validates that there is a variant in all of the supported versions.

I'm also thinking that it's a better idea to get rid of `#[non_exhaustive]` and `UnknownError` altogether. Why not validate them all and fail if there isn't a matching variant?

We could also develop a macro attribute, which would exclude `fvm` versions that don't support a certain error, which would most likely be the older versions.

That way we'll be able to maintain the whole list of errors, unless any of the future `fvm` versions start using old error codes for new errors, which I find unlikely.

src/shim/kernel.rs
```diff
@@ -8,95 +8,50 @@ use fvm4::kernel::SyscallError as E4;
...
-                match self {
-                    $(Self::$variant => Either::Left(N3::$variant),)*
-                    Self::Unknown(u) => u.as_unshimmed(),
-                }
-            }
-        }
-    };
+impl From<N2> for ErrorNumber {
+    fn from(value: N2) -> Self {
+        FromPrimitive::from_u32(value as u32).expect("conversion from N2 must work")
```
lemmih, line 17: You're not allowed to panic. You may `expect` when you can prove correctness to a human but not to rustc. However, as a human, `conversion from N2 must work` is not convincing to me. Either convince me with a proof of correctness or don't use `expect`.

elmattic:
> Having taken a look at the macro, I like the fact that it goes through all the available variants and validates that there is a variant in all of the supported versions.

Unfortunately, this is not always the case, as a new FVM version can introduce new error variants (e.g. `N3::ReadOnly`).

src/shim/kernel.rs
elmattic: I've added static assertions.

lemmih: I still see `expect`.

lemmih (approved):

src/shim/kernel.rs
```diff
@@ -8,74 +8,41 @@ use fvm4::kernel::SyscallError as E4;
...
+use std::fmt::Debug;
 
 macro_rules! error_number {
     ($($variant:ident),* $(,)?) => {
         #[derive(Debug, Clone)]
-        #[non_exhaustive]
         pub enum ErrorNumber {
             $($variant,)*
-            Unknown(UnknownErrorNumber),
+            Unknown(u32),
```
aatifsyed, line 21: There is a hole in the current design. It's not illegal for me to do this:
```rust
let e = ErrorNumber::Unknown(KnownErrorNumber::ReadOnly as u32);
```
but it would lead to incorrect error messages.

This should at least be documented.
```suggestion
            /// This is to catch `#[non_exhaustive]` upstream errors, and MUST NOT for be constructed
            /// outside of this module.
            Unknown(u32),
```

But you may choose to fix.

src/shim/kernel.rs
```diff
@@ -8,74 +8,41 @@ use fvm4::kernel::SyscallError as E4;
...
+use num_traits::FromPrimitive;
 use std::fmt;
+use std::fmt::Debug;
 
 macro_rules! error_number {
     ($($variant:ident),* $(,)?) => {
         #[derive(Debug, Clone)]
-        #[non_exhaustive]
         pub enum ErrorNumber {
             $($variant,)*
```
aatifsyed, line 20: You have two different enums with duplicated variants between them - can you think of a design that gets rid of the duplication while still meeting your design goals?

src/shim/kernel.rs
elmattic: Agree. Could this kind of construct be forbidden using our linting framework?

aatifsyed: No, such a lint would require type information, but this kind of thing should be achieved with a struct instead

aatifsyed: To be clear, I think this is niche enough to just document, but if you wanted to fix it, you'd have to introduce an intermediate :)

src/shim/kernel.rs
elmattic: Yes.

elmattic: @aatifsyed Removed the duplicated enum. We also now preserve upstream error messages (goal 2.)

src/shim/kernel.rs
```diff
@@ -1,5 +1,18 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
+//! We have three goals for our error shims:
+//! - preserve upstream error _numbers_.
+//! - preserve upstream error messages.
+//! - allow _matching_ on specific errors.
+//!
+//! There are a couple of things that make this difficult:
+//! - `fvm_shared*::error::ErrorNumber` is `#[non_exhaustive]`
+//! - ...
```
aatifsyed, line 10: Remove this or fill it in - it was an example message :)

aatifsyed (approved):

src/shim/kernel.rs
aatifsyed: I meant remove the `...` on line 10, not the entire docs, which I think you should keep :)

elmattic: Fixed.

aatifsyed (approved):

PR #3552, opened 2023-10-09
feat: custom lints

aatifsyed:
https://github.com/ChainSafe/forest/blob/9a7a0a98d747bd32bbb35719d884309de400ad55/tests/lint.rs#L3-L23


LesnyRumcajs (changes requested):
Let's have some tests. Otherwise, given the complexity of the code, it's impossible to make any modifications here with confidence that nothing got broken on the way.

LesnyRumcajs (approved):

sudo-shashank (approved):

Issue #3556, opened 2023-10-09
Stateless bootstrap node


lemmih:
**Issue summary**

Bootstrap nodes serve as the entry point to the Filecoin p2p swarm. Currently, bootstrap nodes are running Lotus and must sync to the network. This makes them slightly expensive (>$100/month). However, bootstrap nodes should be able to run without any state, making them much cheaper.

- [ ] Add a stateless mode to Forest. Forest still connects to the p2p swarm in this mode but does not sync to HEAD.
- [ ] For `Hello` requests, respond with the genesis tipset as the heaviest seen tipset. We don't want to claim to have validated any other tipsets than this.
- [ ] Cache the `N` most recent blocks.
- [ ] If a block is requested that we do not have, forward the request to other peers and cache the result.
- [ ] Respond to `ChainRequest`s with `PartialResponse`. We never want to respond with more than 1 block.

If we do not respond to `ChainRequest`s, other peers will eventually disconnect from us. Not a show-stopper, but we'd like for them to stay connected to us for as long as possible.

The stateless bootstrap node will run on droplets with roughly 1GiB of memory and 20GiB of disk. There should be wiggle room for caching a fair number of blocks.

**Other information and links**

LesnyRumcajs:
@lemmih what are we missing right now to have a bootstrap node (in terms of Forest implementation)? Or are we feature-complete and it's just a matter of deploying it and handling the logistics behind it?

LesnyRumcajs:
@lemmih Are we good closing this? The stateless nodes are operated by the infrastructure team.

lemmih:
Done.

Issue #3509, opened 2023-09-21
High-cardinality Prometheus metrics


LesnyRumcajs:
**Issue summary**
Right now, Forest produces lots of high-cardinality metrics. A long-running Forest instance can generate 176 KB of total [metrics](https://github.com/ChainSafe/forest/files/12648981/metrics.txt). This may not seem significant, but it's in fact huge given how often metrics are scraped and the need for them to be retained.

The largest culprit is the `peer_tipset_epoch`, `cns_block_validation_tasks_time_bucket` and `block_validation_tasks_time_bucket`
![image](https://github.com/ChainSafe/forest/assets/9642092/e1723080-0571-4542-9678-66f402e3a024)

The reason is the high cardinality of these metrics, that is to say, the amount of unique label combinations we attach to the metrics, e.g. a peer. Each peer will create a distinct row in the metrics database and will, most likely, keep reporting the same value because we are no longer connected to it:

```
peer_tipset_epoch{PEER="12D3KooWSGrHWDzyNVVT14aacpUwZAog5LczHeJvE1CENMHKtehy"} -1
peer_tipset_epoch{PEER="12D3KooWSGw3fdnYjSbqYky7F67TraGwhTj2K9Dmu16iQgB9XcyF"} -1
peer_tipset_epoch{PEER="12D3KooWSH3VF7GNFBRPny3ngRdjGzaRJzMobUqtCpbKdLiuJ3Xy"} -1
peer_tipset_epoch{PEER="12D3KooWSHG9vVStHMi9vhfgD1XaR221Ur6RjWcVsiY7WfVaX4QL"} -1
peer_tipset_epoch{PEER="12D3KooWSHRWoF1isZgRpjK4METrdaa6b5eHQwEooyKoGJfMm22m"} 3224674
peer_tipset_epoch{PEER="12D3KooWSHWJgPc9WG5d6qCteNrd8tH37NTAEPKqdWfgBRAmtNSJ"} -1
peer_tipset_epoch{PEER="12D3KooWSHcKeh15WakVSPRiDRKe6PxBsMKQekVhcL4sj4zgRBYV"} -1
peer_tipset_epoch{PEER="12D3KooWSJhokpUV38JSKvvepBkRDDHNFqbpUQXLYRcFicPd9f81"} -1
```

for `cns_block_validation_tasks_time_bucket`, the key from labels is the bucket and task type.
```
cns_block_validation_tasks_time_bucket{type="validate_winner_election",le="5"} 158432
[...]

PR #3593, opened 2023-10-16
feat: add `--stateless` flag to `forest`

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/3556

- [x] Add a stateless mode to Forest. Forest still connects to the p2p swarm in this mode but does not sync to HEAD.
- [x] For Hello requests, respond with the heaviest tipset in the database(genesis by default).
- [x] ~If a block is requested that we do not have, forward the request to other peers and cache the result.~ We don't forward the request as discussed [here](https://chainsafeio.slack.com/archives/C0144LHN732/p1698231600242419?thread_ts=1698230295.532219&cid=C0144LHN732)
- [x] Respond to ChainRequests with PartialResponse. We never want to respond with more than 1 block. Retrieve from the network when it's not available locally.
- [x] CI test

Closes

## Change checklist


src/chain_sync/chain_muxer.rs
```diff
@@ -850,6 +859,21 @@ where
...
+                        self.state = ChainMuxerState::Stateless(Either::Right(Box::pin(
+                            tokio::time::sleep(Duration::from_secs(60)),
+                        )));
+                    }
+                    Either::Right(ref mut sleep) => {
+                        std::task::ready!(sleep.as_mut().poll(cx));
+                        self.state =
+                            ChainMuxerState::Stateless(Either::Left(self.evaluate_network_head()));
+                    }
+                },
```
lemmih, line 876: Why do you call `evaluate_network_head`?

hanabi1224: We need to handle or discard the inbound p2p messages. `evaluate_network_head` does the basic handling

hanabi1224: Done.

src/chain_sync/chain_muxer.rs
```diff
@@ -850,6 +901,22 @@ where
...
+                        self.state = ChainMuxerState::Stateless(Either::Right(Box::pin(
+                            tokio::time::sleep(Duration::from_secs(60)),
+                        )));
+                    }
+                    Either::Right(ref mut sleep) => {
+                        std::task::ready!(sleep.as_mut().poll(cx));
+                        self.state = ChainMuxerState::Stateless(Either::Left(
+                            self.handle_network_messages(),
+                        ));
+                    }
```
lemmih, line 918: `handle_network_messages` loops forever. So there's no need for `sleep`. Or is there something I'm missing?

hanabi1224: Yes, the loop breaks only when `p2p_messages` channel is dropped, changed to returning `Poll::Ready` in that case

src/chain_sync/chain_muxer.rs
```diff
@@ -512,6 +515,48 @@ where
         Ok(Some((tipset, source)))
     }
 
+    fn handle_network_messages(&self) -> ChainMuxerFuture<(), ChainMuxerError> {
```
lemmih, line 518: `handle_network_messages` is fairly generic. This is the main event loop for the stateless mode, right?

```suggestion
    fn stateless_node(&self) -> ChainMuxerFuture<(), ChainMuxerError> {
```

src/chain_sync/network_context.rs
```diff
@@ -214,16 +246,14 @@ where
...
+        network_send: flume::Sender<NetworkMessage>,
         peer_id: Option<PeerId>,
         tsk: &TipsetKeys,
         request_len: u64,
         options: u64,
-    ) -> Result<Vec<T>, String>
-    where
-        T: TryFrom<TipsetBundle, Error = String> + Send + Sync + 'static,
-    {
+    ) -> Result<Vec<TipsetBundle>, String> {
```
lemmih, line 256: Looks like you're changing the semantics here. What happens if we are sent responses that can't be converted into full tipsets? Was the previous behavior wrong?

hanabi1224: The conversion now happens on the caller side, the logic should be same, see

https://github.com/ChainSafe/forest/pull/3593/files#diff-1686eeeaa959ad579a0b8ef0af0a7907816520a1c203b1077b12cda20444492bR174

https://github.com/ChainSafe/forest/pull/3593/files#diff-1686eeeaa959ad579a0b8ef0af0a7907816520a1c203b1077b12cda20444492bR196

lemmih: You've changed the semantics, though. With `fts` being a vector of results, the core logic has changed.

hanabi1224: I see what you mean, changed to `try_collect`.

(although the logic guarantees `len` of the Vec is 0 or 1, `Result<Vec<_>>` and `Vec<Result<_>>` makes no difference here.)

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,35 @@
...
+
+source "$(dirname "$0")/harness.sh"
+
+forest_init_stateless
+
+echo "Verifying the heaviest tipset to be the genesis"
+MSG=$($FOREST_CLI_PATH chain head)
+assert_eq "$MSG" $'[\n  "bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4"\n]'
+
+wait_util_p2p_is_ready
```
LesnyRumcajs, line 14: until?

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# This script tests the stateless mode of a forest node
+
+source "$(dirname "$0")/harness.sh"
+
+forest_init_stateless
+
+echo "Verifying the heaviest tipset to be the genesis"
+MSG=$($FOREST_CLI_PATH chain head)
+assert_eq "$MSG" $'[\n  "bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4"\n]'
```
LesnyRumcajs, line 12: Perhaps using `sed` would help avoiding issues with whitespace.

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,35 @@
...
+
+wait_util_p2p_is_ready
+ADDRESS=$($FOREST_CLI_PATH net listen | tail -n 1)
+echo "Stateless node address: $ADDRESS"
+PEER_ID=$(echo "$ADDRESS" | cut -d "/" -f7 -z)
```
LesnyRumcajs, line 17: If possible, let's use long flag names. I had to look-up what is `-z`.

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,35 @@
...
+{
+    echo "[network]"
+    echo "listening_multiaddrs = [\"/ip4/127.0.0.1/tcp/0\"]"
+    echo "bootstrap_peers = [\"$ADDRESS\"]"
+} > $CONFIG_PATH
+
+NODE_LOG_DIRECTORY=$(mktemp --directory)
+RUST_LOG="info,forest_filecoin::chain_sync::network_context=debug" $FOREST_PATH --chain calibnet --encrypt-keystore false --auto-download-snapshot --config "$CONFIG_PATH" --no-metrics --rpc false --log-dir "$NODE_LOG_DIRECTORY" &
+PID=$!
+# Verify that the stateless node can respond to chain exchange requests
```
LesnyRumcajs, line 31: I think a timeout would help here. I can imagine the expression not matching because someone changed one letter. Then, these two lines would wait until killed by GH.

LesnyRumcajs: It's also a bit flaky to rely on logs. Maybe we could rely on metrics instead?

scripts/tests/harness.sh
```diff
@@ -47,6 +47,21 @@ function forest_run_node_detached {
...
+  CONFIG_PATH="./stateless_forest_config.toml"
+  echo "${CONFIG_PATH}"
+  echo "Running forest in stateless and detached mode"
+  {
+    echo "[client]"
+    echo "data_dir = \"/tmp/stateless_forest_data\""
+    echo ""
+    echo "[network]"
+    echo "listening_multiaddrs = [\"/ip4/127.0.0.1/tcp/0\"]"
+  } > $CONFIG_PATH
```
LesnyRumcajs, line 60: Do you think HEREDOC would be more readable and easier to modify later?

scripts/tests/harness.sh
```diff
@@ -47,6 +47,21 @@ function forest_run_node_detached {
...
+  
+  $FOREST_PATH --detach --chain calibnet --encrypt-keystore false --config "$CONFIG_PATH" --log-dir "$LOG_DIRECTORY" --save-token ./stateless_admin_token --no-gc --stateless
```
LesnyRumcajs, line 62: You used `--no-gc` and `--stateless`. Should this be implied in the code?

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,35 @@
...
+until grep -r -m 1 "non-empty ChainExchange response from $PEER_ID" "$NODE_LOG_DIRECTORY"; do
+    sleep 1s;
+done
+kill -9 $PID
```
LesnyRumcajs, line 35: it's slightly more understandable
```suggestion
kill -KILL $PID
```

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,35 @@
...
+CONFIG_PATH="./forest_config.toml"
+{
+    echo "[network]"
+    echo "listening_multiaddrs = [\"/ip4/127.0.0.1/tcp/0\"]"
+    echo "bootstrap_peers = [\"$ADDRESS\"]"
+} > $CONFIG_PATH
+
+NODE_LOG_DIRECTORY=$(mktemp --directory)
+RUST_LOG="info,forest_filecoin::chain_sync::network_context=debug" $FOREST_PATH --chain calibnet --encrypt-keystore false --auto-download-snapshot --config "$CONFIG_PATH" --no-metrics --rpc false --log-dir "$NODE_LOG_DIRECTORY" &
+PID=$!
```
LesnyRumcajs, line 30: `PID` of what? Let's name it along the lines of `FOREST_NODE_PID`

src/chain_sync/chain_muxer.rs
```diff
@@ -512,6 +515,48 @@ where
...
+        let bad_block_cache = self.bad_blocks.clone();
+        let mem_pool = self.mpool.clone();
+        let block_delay = self.state_manager.chain_config().block_delay_secs as u64;
+
+        let future = async move {
+            loop {
+                let event = match p2p_messages.recv_async().await {
+                    Ok(event) => event,
+                    Err(why) => {
+                        debug!("Receiving event from p2p event stream failed: {}", why);
```
LesnyRumcajs, line 532:
```suggestion
                        debug!("Receiving event from p2p event stream failed: {why}");
```

scripts/tests/harness.sh
hanabi1224: I'm not sure if we should run `gc` by default to keep the disk usage of a stateless node bounded.  @lemmih what's your suggestion?

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: Fixed.

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: Good catch! Fixed.

scripts/tests/harness.sh
lemmih: Conceptually, a stateless node should not use the database at all. A stateless node must not use an unbounded amount of disk space. Only N blocks should be cached (ideally in memory); all other blocks should be forgotten.

hanabi1224: > Conceptually, a stateless node should not use the database at all. A stateless node must not use an unbounded amount of disk space. Only N blocks should be cached (ideally in memory); all other blocks should be forgotten.

@lemmih Should in-memory store change be included in this PR? I feel it might have quite some conflicts with the ongoing GC refactoring PR, or should I do that after GC PR is merged?

lemmih: Sure, you can solve #3556 with multiple PRs.

scripts/tests/harness.sh
hanabi1224: That's good idea, fixed.

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: We don't have such chain exchange metrics for each peer id. Could you elaborate?

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: Could you elaborate with a example script line?

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: Fixed

scripts/tests/calibnet_stateless_mode_check.sh
LesnyRumcajs: Something like `MSG=$($FOREST_CLI_PATH chain head | jq -r '.[0]'` should output `bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4`

scripts/tests/calibnet_stateless_mode_check.sh
LesnyRumcajs: > Add 10min timeout to the workflow step.

The problem with the workflow timeout is that it won't tell you precisely what was sleeping for so long, and it won't timeout on a local test.

LesnyRumcajs: > We don't have such chain exchange metrics for each peer id. Could you elaborate?

We could add some. We'd need to be tactical to not [bloat](https://github.com/ChainSafe/forest/issues/3509) the metric reports. We could have a metric of, e.g., chain exchanges per peer in the last 5 minutes.

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: Fixed.

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: @LesnyRumcajs How about adding per-peer chain exchange metrics after https://github.com/ChainSafe/forest/issues/3509 is tackled? For now, it will bloat the metrics report just like `peer_tipset_epoch`. Or would you suggest bloat it first then make the fix later together with https://github.com/ChainSafe/forest/issues/3509

LesnyRumcajs: I suggest not bloating it. I imagine metrics with labels and a timed cache of the most recent peers that did the chain exchange. Periodically, the metrics that expired would be deleted with something like https://docs.rs/prometheus/latest/prometheus/core/struct.MetricVec.html#method.remove_label_values, and so the `/metrics` page would not contain those peers.

It doesn't touch #3509 (though it'd potentially introduce something that could be used there).

If you believe implementing it would be a lot of work, we can do that in a separate PR. What's your pick?

hanabi1224: @LesnyRumcajs Fixed.

LesnyRumcajs: Beautiful!

src/libp2p/chain_exchange/message.rs
```diff
@@ -258,15 +280,127 @@ fn fts_from_bundle_parts(
...
-
     #[quickcheck]
     fn chain_exchange_response_status_roundtrip(status: ChainExchangeResponseStatus) {
         let serialized = serde_json::to_string(&status).unwrap();
         let parsed = serde_json::from_str(&serialized).unwrap();
         assert_eq!(status, parsed);
     }
+
+    #[test]
+    fn tipset_bundle_save_load_roundtrip() {
```
lemmih, line 305: What are you testing here?

src/libp2p/service.rs
```diff
@@ -674,54 +678,89 @@ async fn handle_chain_exchange_event<DB>(
     ce_event: request_response::Event<ChainExchangeRequest, ChainExchangeResponse>,
```
lemmih, line 678: The changes to this function are for fetching block headers when in `stateless` mode? If so, let's reconsider. For keeping the N most recent block headers, we can just get them from the gossip events.

src/libp2p/tests/decode_test.rs
```diff
@@ -15,37 +13,6 @@ use crate::shim::{
 };
 use num::BigInt;
 
-#[test]
-fn convert_single_tipset_bundle() {
```
lemmih, line 19: Why delete this test?

lemmih (changes requested):

src/libp2p/service.rs
hanabi1224: It's for `If a block is requested that we do not have, forward the request to other peers and cache the result.`

> If we do not respond to ChainRequests, other peers will eventually disconnect from us. Not a show-stopper, but we'd like for them to stay connected to us for as long as possible.

src/libp2p/chain_exchange/message.rs
hanabi1224: Removed with the no-chain-exchain-forwarding change

src/libp2p/service.rs
hanabi1224: Remove the logic that forwards chain exchange requests, as discussed

src/libp2p/tests/decode_test.rs
hanabi1224: Now it's back

CHANGELOG.md
```diff
@@ -65,6 +65,8 @@
 
 - [#3422](https://github.com/ChainSafe/forest/issues/3422) Add NV21 (Watermelon)
   support for calibration network.
+- [#3593](https://github.com/ChainSafe/forest/pull/3593): Add `--stateless` flag
```
LesnyRumcajs, line 68: Is it worth elaborating a bit on what this flag exactly does?

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,34 @@
...
+
+forest_init_stateless
+
+echo "Verifying the heaviest tipset to be the genesis"
+HEAD_CID=$($FOREST_CLI_PATH chain head | jq -r '.[0]')
+assert_eq "$HEAD_CID" "bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4"
+
+STATELESS_NODE_ADDRESS=$($FOREST_CLI_PATH net listen | tail -n 1)
+echo "Stateless node address: $STATELESS_NODE_ADDRESS"
+STATELESS_NODE_PEER_ID=$(echo "$STATELESS_NODE_ADDRESS" | cut --delimiter="/" --fields=7 --zero-terminated)
```
LesnyRumcajs, line 16: It might be worth commenting on an expected format with an example multiaddress.

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,34 @@
...
+# Run a normal forest node that only connects to the stateless node
+CONFIG_PATH="./forest_config.toml"
+cat <<- EOF > $CONFIG_PATH
+	[network]
+	listening_multiaddrs = ["/ip4/127.0.0.1/tcp/0"]
+	bootstrap_peers = ["$STATELESS_NODE_ADDRESS"]
+EOF
+
+# Disable discovery to not connect to more nodes
+$FOREST_PATH --chain calibnet --encrypt-keystore false --auto-download-snapshot --config "$CONFIG_PATH" --rpc false --mdns false --kademlia false --metrics-address 127.0.0.1:6117 &
```
LesnyRumcajs, line 28: Could we disable `mdns` and `kademlia` in config? Unless you think passing it as flags makes more sense.

scripts/tests/calibnet_stateless_mode_check.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+set -euxo pipefail
+
+# This script tests the stateless mode of a forest node
+
+source "$(dirname "$0")/harness.sh"
+
+forest_init_stateless
```
LesnyRumcajs, line 8: Will this node be killed in case of an error somewhere along the way?

LesnyRumcajs (approved):

lemmih (approved):
Looks good.

I'm not sure it'll actually work but there's no way of testing this without just deploying it and seeing if anything breaks.

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: Yes the detached node will be killed by `trap forest_cleanup EXIT` in `harness.sh`
`

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: I've no preference, fixed as suggested

scripts/tests/calibnet_stateless_mode_check.sh
hanabi1224: Done

CHANGELOG.md
hanabi1224: Done

LesnyRumcajs (approved):

PR #3595, opened 2023-10-16
docs: the blockchain, state tree, and snapshots

aatifsyed:
[rendered](https://github.com/ChainSafe/forest/files/12927509/rendered.pdf)

A ground-up education on:
- Actors
- The Filecoin blockchain
- The Filecoin state tree
- Snapshots

I didn't really understand how these all fit together until the past couple weeks.

# Reviewers
- Check for correctness please - there's a to-do in there for you also
- I thought account balances etc would be a more motivated example of state,
  but I can't find where those are in the graph.
  The [`fil_actor_account_state`](https://docs.rs/fil_actor_account_state/6.1.0/fil_actor_account_state/v11/struct.State.html)
  doesn't look how I expect at all.

# Location
I strongly believe (this) documentation should live in-tree, in `/src`, not `/documentation`.
It will:
- be able to link to the code and ecosystem.
- be able to link to other in-tree documentation.
- be more seen.


aatifsyed:
[rendered](https://github.com/ChainSafe/forest/files/12927509/rendered.pdf)

src/lib.rs
```diff
@@ -1,8 +1,230 @@
...
+//!
+//! An example of an actor is the [`cron`](fil_actors_shared::v11::runtime::builtins::Type::Cron)
+//! actor.
+//! Its [internal state](fil_actor_cron_state::v11::State) is a to-do list of
+//! other actors to invoke every epoch.
+//!
+//! See [the Filecoin docs](https://docs.filecoin.io/basics/the-blockchain/actors)
+//! for more information about actors.
+//!
+//! [^1]: The number of [built-in actors](https://docs.filecoin.io/basics/the-blockchain/actors#built-in-actors)
```
jdjaustin, line 18:
```suggestion
//! [^1]: The number of [built-in actors](https://github.com/filecoin-project/builtin-actors)
```
As you already linked to the Filecoin docs in the lines directly above, I think the link to the repo makes more sense here and (as an added bonus) is less likely to suffer from link rot.

src/lib.rs
```diff
@@ -1,8 +1,230 @@
...
+//! TODO(aatifsyed): where is my balance?
+//!
+//! # The Filecoin blockchain
+//!
+//! Filecoin consists of a blockchain of messages.
+//! Listed below are the core objects for the blockchain.
+//! Each one can be addressed by a [`Cid`](cid::Cid).
+//!
+//! - [`Message`](shim::message::Message)s are statements of messages between
+//!   a the actors.
```
jdjaustin, line 30:
```suggestion
//!   the actors.
```

src/lib.rs
```diff
@@ -1,8 +1,230 @@
...
+//!   They describe and (equivalently) represent a change in _the blockchain state_.
+//!   See [`apply_block_messages`](state_manager::apply_block_messages) to learn
+//!   more.
+//!   Messages may be [signed](message::SignedMessage).
+//!   TODO(aatifsyed): by whom? what does that mean?
```
jdjaustin, line 35: From https://spec.filecoin.io/systems/filecoin_vm/message/ :

"A semantically valid `SignedMessage` must carry a signature that verifies the payload as having been signed with the public key of the account actor identified by the `From` address. Note that when the `From` address is an ID-address, the public key must be looked up in the state of the sending account actor in the parent state identified by the block."

src/lib.rs
aatifsyed: You're right - the redundant link isn't worth it, I'll just remove the link altogether though

lemmih (approved):

hanabi1224 (approved):

PR #3598, opened 2023-10-17
fix: make_chain_exchange_response should return BlockNotFound properly

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/3556

- Fix `make_chain_exchange_response` to return `BlockNotFound` instead of `InternalError` when start block is missing from the database
- Fix `make_chain_exchange_response` to return `BadRequest` when `options` is invalid
- Refactor the function body to use `ChainIndex::chain` and `try_fold` instead of a loop

```rust
/// Status codes of a `chain_exchange` response.
#[derive(Clone, Debug, PartialEq, Eq, Copy)]
#[cfg_attr(test, derive(derive_quickcheck_arbitrary::Arbitrary))]
pub enum ChainExchangeResponseStatus {
    /// All is well.
    Success,
    /// We could not fetch all blocks requested (but at least we returned
    /// the `Head` requested). Not considered an error.
    PartialResponse,
    /// Request.Start not found.
    BlockNotFound,
    /// Requester is making too many requests.
    GoAway,
    /// Internal error occurred.
    InternalError,
    /// Request was bad.
    BadRequest,
    /// Other undefined response code.
    Other(#[cfg_attr(test, arbitrary(gen(|_| 1)))] i32),
}
```

Closes

## Change checklist


lemmih (changes requested):
Let's rewrite this using `cs.chain()` rather than an explicit loop.

Something like: `let response_chain = cs.chain_index.chain(root).take(len).collect();`

If `response_chain` is empty then return `BlockNotFound`. If `response_chain` doesn't have enough epochs, return `PartialResponse`. Otherwise return `Success`.

hanabi1224:
> Let's rewrite this using `cs.chain_index.chain()` rather than an explicit loop.

@lemmih Done

src/libp2p/chain_exchange/provider.rs
```diff
@@ -21,68 +21,81 @@ pub fn make_chain_exchange_response<DB>(
...
-            // Cloning blocks isn't ideal, this can maybe be switched to serialize this
-            // data in the function. This may not be possible without overriding rpc in
-            // libp2p
-            tipset_bundle.blocks = tipset.blocks().to_vec();
-        }
+    match cs
+        .chain_index
+        .chain(root)
+        .take(request.request_len as _)
+        .try_fold(
```
lemmih, line 56: Why fold instead of map?

hanabi1224: `try_fold` allows to terminate early on internal errors. 

> try_fold() takes two arguments: an initial value, and a closure with two arguments: an 'accumulator', and an element. The closure either returns successfully, with the value that the accumulator should have for the next iteration, or it returns failure, with an error value that is propagated back to the caller immediately (short-circuiting).

lemmih: Instead of manually dealing with errors, how about we split the function into two parts: An inner part that returns `Result<Vec<FullTipset>>` (or something similar) and an outer part that maps the result to `Success`, `PartialResponse`, or `InternalError`. That way, we don't have to return `InternalError` in multiple locations and we can use `?` as we normally would.

hanabi1224: I'm not sure if the outer part is a `map` function, as the function actually folds `Iterator<Item=Tipset>` into a single `ChainExchangeResponse`

> we don't have to return InternalError in multiple locations

Currently `InternalError` is returned in 2 places, 1 is to prepare the input of  `ChainIndex::chain`, the other is inside `try_fold`, I'm not sure I understand how to omit occurrence 1 with `map`

 could you elaborate on both points?

lemmih (approved):

hanabi1224:
@LesnyRumcajs @sudo-shashank kindly ping

src/libp2p/chain_exchange/message.rs
```diff
@@ -38,6 +38,11 @@ impl ChainExchangeRequest {
     pub fn include_messages(&self) -> bool {
         self.options & MESSAGES > 0
     }
+
+    /// If the value of `options` is valid
```
LesnyRumcajs, line 42: This comment is not very helpful. Could elaborate here?

LesnyRumcajs (approved):

src/libp2p/chain_exchange/message.rs
hanabi1224: How about `/// If either the [HEADERS] bit or the [MESSAGES] bit is set.`?

hanabi1224: @LesnyRumcajs please re-approve if the change looks OK as it's required by the new branch rule.

PR #3600, opened 2023-10-17
[Chore] Move temp db format into a helper.

ruseinov:
## Summary of changes

- Moved temporary_db_name format into a helper. 

Closes

## Change checklist


LesnyRumcajs (changes requested):
There's one more entry. https://github.com/ChainSafe/forest/blob/main/src/db/migration/migration_map.rs#L445

src/db/migration/db_migration.rs
```diff
@@ -79,6 +80,10 @@ impl DbMigration {
     }
 }
 
+pub(crate) fn db_name(from: &Version, to: &Version) -> String {
```
LesnyRumcajs, line 83: Plus, I think it belongs more in the `migration_map.rs` file.

ruseinov: Since it's used to implement `db_migration` trait's method - I don't see a reason to move it out of the file. But if needed - we can do it. No strong opinion here.

LesnyRumcajs: It's used to implement `MigrationOperation` trait methods defined in `migration_map.rs`, no?

src/db/migration/migration_map.rs
```diff
@@ -470,3 +471,7 @@ mod tests {
         assert!(migration.post_checks(temp_dir.path()).is_ok());
     }
 }
+
+pub(crate) fn db_name(from: &Version, to: &Version) -> String {
```
LesnyRumcajs, line 475: Do we still need it in the trait? Let's simplify things and just have a free `temporary_db_name` function. I guess it _may_ be overridden in the future, but probably YAGNI.

ruseinov: It is not in the trait.

LesnyRumcajs: Let me rephrase. `temporary_db_name` _is_ in the trait, and it's just a proxy for `db_name`. We could use the `db_name` directly (though renamed to be explicit for the purpose). Less code, less bugs. What do you think?

ruseinov: Yeah, that makes sense. I'll incorporate that.

ruseinov: I just realised that `temporary_db_name` is never used outside of any given migration, so there is no reason to have it as a trait member in the first place. This has been addressed.

LesnyRumcajs (approved):

lemmih (approved):

PR #3605, opened 2023-10-18
fix: update `fn tipset_from_keys` to return `Result<Option<Arc<Tipset>>, Error>`

hanabi1224:
## Summary of changes

As a follow-up PR of https://github.com/ChainSafe/forest/pull/3598 . More discussion on [slack](https://chainsafeio.slack.com/archives/C0144LHN732/p1697621293022769?thread_ts=1697620719.493709&cid=C0144LHN732)

Changes introduced in this pull request:

- update `fn tipset_from_keys` to return `Result<Option<Arc<Tipset>>, Error>`

## Reference issue to close (if applicable)

Closes

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

PR #3606, opened 2023-10-18
fix: reduce allocations in `ok_or` by switching to `ok_or_else`

hanabi1224:
## Summary of changes

Encounter the issue when implementing https://github.com/ChainSafe/forest/pull/3605

- reduce allocations in `Option::ok_or` by switching to `Option::ok_or_else`

Closes

## Change checklist

### DIFF
--- a/src/db/car/forest.rs
+++ b/src/db/car/forest.rs
@@ -116,15 +116,18 @@ impl<ReaderT: super::RandomAccessFileReader> ForestCar<ReaderT> {
         let mut footer_buffer = [0; ForestCarFooter::SIZE];
         cursor.read_exact(&mut footer_buffer)?;
 
-        let footer = ForestCarFooter::try_from_le_bytes(footer_buffer).ok_or(invalid_data(
-            format!("not recognizable as a `{}` file", FOREST_CAR_FILE_EXTENSION),
-        ))?;
+        let footer = ForestCarFooter::try_from_le_bytes(footer_buffer).ok_or_else(|| {
+            invalid_data(format!(
+                "not recognizable as a `{}` file",
+                FOREST_CAR_FILE_EXTENSION
+            ))
+        })?;
 
         let cursor = Cursor::new_pos(&reader, 0);
         let mut header_zstd_frame = decode_zstd_single_frame(cursor)?;
         let block_frame = UviBytes::<Bytes>::default()
             .decode(&mut header_zstd_frame)?
-            .ok_or(invalid_data("malformed uvibytes"))?;
+            .ok_or_else(|| invalid_data("malformed uvibytes"))?;
         let header = from_slice_with_fallback::<CarHeader>(&block_frame)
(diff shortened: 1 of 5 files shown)


src/db/car/plain.rs
```diff
@@ -306,8 +306,8 @@ fn cid_error_to_io_error(cid_error: cid::Error) -> io::Error {
 /// ```
 #[tracing::instrument(level = "trace", skip_all, ret)]
 fn read_header(mut reader: impl Read) -> io::Result<CarHeader> {
-    let header_len =
-        read_varint_body_length_or_eof(&mut reader)?.ok_or(io::Error::from(UnexpectedEof))?;
+    let header_len = read_varint_body_length_or_eof(&mut reader)?
+        .ok_or_else(|| io::Error::from(UnexpectedEof))?;
```
hanabi1224, line 310: This is no allocation but the underlying logic is more than constructing a struct

```rust
#[inline]
    pub(super) fn new_simple(kind: ErrorKind) -> Self {
        let utagged = ((kind as usize) << 32) | TAG_SIMPLE;
        // Safety: `TAG_SIMPLE` is not zero, so the result of the `|` is not 0.
        let res = Self(unsafe { NonNull::new_unchecked(ptr::invalid_mut(utagged)) }, PhantomData);
        // quickly smoke-check we encoded the right thing (This generally will
        // only run in std's tests, unless the user uses -Zbuild-std)
        debug_assert!(
            matches!(res.data(), ErrorData::Simple(k) if k == kind),
            "repr(simple) encoding failed {:?}",
            kind,
        );
        res
    }
```

lemmih (approved):

src/db/car/forest.rs
```diff
@@ -116,15 +116,18 @@ impl<ReaderT: super::RandomAccessFileReader> ForestCar<ReaderT> {
...
+                "not recognizable as a `{}` file",
+                FOREST_CAR_FILE_EXTENSION
+            ))
+        })?;
 
         let cursor = Cursor::new_pos(&reader, 0);
         let mut header_zstd_frame = decode_zstd_single_frame(cursor)?;
         let block_frame = UviBytes::<Bytes>::default()
             .decode(&mut header_zstd_frame)?
-            .ok_or(invalid_data("malformed uvibytes"))?;
```
hanabi1224, line 127:
```rust
#[stable(feature = "rust1", since = "1.0.0")]
    pub fn new<E>(kind: ErrorKind, error: E) -> Error
    where
        E: Into<Box<dyn error::Error + Send + Sync>>,
    {
        Self::_new(kind, error.into())
    }

fn _new(kind: ErrorKind, error: Box<dyn error::Error + Send + Sync>) -> Error {
        Error { repr: Repr::new_custom(Box::new(Custom { kind, error })) }
    }
```

aatifsyed (approved):

PR #3608, opened 2023-10-18
fix: remove `chain` field from `struct Config`

hanabi1224:
## Summary of changes

~Currently, in forest config file there're duplicate `bootstrap_peers` fields under both `network` and `chain` section. This PR tries to keep `network.bootstrap_peers` only~
As the follow-up cleanup of https://github.com/ChainSafe/forest/pull/3615

- remove `chain: Arc<ChainConfig>` field from `struct Config`
- add `chain: NetworkChain` field as suggested [here](https://github.com/ChainSafe/forest/pull/3615#pullrequestreview-1689604664)

Closes

## Change checklist


src/cli_shared/cli/mod.rs
```diff
@@ -162,13 +162,21 @@ impl CliOpts {
...
 
-        if let Some(chain) = &self.chain {
-            // override the chain configuration
-            cfg.chain = Arc::new(ChainConfig::from_chain(chain));
-        } else {
-            // override any custom changes to the chain configuration based on the used
-            // network.
-            cfg.chain = Arc::new(ChainConfig::from_chain(&cfg.chain.network));
+        // override chain specific configurations
+        let chain = self.chain.clone().unwrap_or(cfg.chain.network.clone());
```
lemmih, line 166: Maybe `unwrap_or_else` like in your other PR. :)

hanabi1224: Fixed

src/networks/mod.rs
```diff
@@ -182,14 +182,6 @@ struct DrandPoint<'a> {
...
     pub network: NetworkChain,
     pub genesis_cid: Option<String>,
-    #[cfg_attr(test, arbitrary(gen(
-        |g: &mut quickcheck::Gen| {
-            let addr = std::net::Ipv4Addr::arbitrary(&mut *g);
-            let n = u8::arbitrary(g) as usize;
-            vec![addr.into(); n]
-        }
-    )))]
-    pub bootstrap_peers: Vec<Multiaddr>,
```
lemmih, line 192: I think we should keep `bootstrap_peers` as a field, and exclude `ChainConfig` in its entirety from the configuration file.

lemmih (changes requested):

src/networks/mod.rs
hanabi1224: Done.

ruseinov (approved):
looks good to me. could be a smaller PR, but given the small amount of changes it seems easy to grasp.

src/cli_shared/cli/config.rs
```diff
@@ -45,7 +44,6 @@ pub struct Config {
     pub parity_db: crate::db::parity_db_config::ParityDbConfig,
     pub network: Libp2pConfig,
     pub sync: SyncConfig,
-    pub chain: Arc<ChainConfig>,
```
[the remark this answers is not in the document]
hanabi1224: Made the suggested change in https://github.com/ChainSafe/forest/pull/3615 and repurpose this PR to be the follow-up cleanup

src/cli_shared/mod.rs
```diff
@@ -11,9 +11,11 @@ pub use mimalloc;
 #[cfg(feature = "jemalloc")]
 pub use tikv_jemallocator;
 
+use crate::networks::NetworkChain;
+
 /// Gets chain data directory
-pub fn chain_path(config: &crate::cli_shared::cli::Config) -> PathBuf {
-    PathBuf::from(&config.client.data_dir).join(config.chain.network.to_string())
+pub fn chain_path(network: &NetworkChain, config: &crate::cli_shared::cli::Config) -> PathBuf {
+    PathBuf::from(&config.client.data_dir).join(network.to_string())
```
lemmih, line 18: How is `network` different from `config.chain`?

lemmih (changes requested):
- `recent_state_roots` and `request_window` should be configurable and are not determined by the blockchain.
- You're sometimes passing around both a `Config` and separate `NetworkChain`. But the `Config` already contains a `NetworkChain`. Let's not pass around two separate `NetworkChain`s.
- If the `ChainConfig` is entirely derived from the `NetworkChain`, you don't need to pass it around. Passing around the `NetworkChain` is sufficient.

src/cli_shared/mod.rs
hanabi1224: CLI `--chain` args has higher priority than `config.chain` and their value may disagree, I guess I can override `config.chain` at the very beginning instead.

hanabi1224: Fixed

hanabi1224:
@lemmih

> * `recent_state_roots` and `request_window` should be configurable and are not determined by the blockchain.

moved both to `SyncConfig` and they're now configurable

> * You're sometimes passing around both a `Config` and separate `NetworkChain`. But the `Config` already contains a `NetworkChain`. Let's not pass around two separate `NetworkChain`s.

Fixed.
> * If the `ChainConfig` is entirely derived from the `NetworkChain`, you don't need to pass it around. Passing around the `NetworkChain` is sufficient.

Fixed.

src/cli_shared/cli/mod.rs
```diff
@@ -93,11 +92,11 @@ pub struct CliOpts {
     /// pre-loaded database
     #[arg(long)]
     pub skip_load: Option<bool>,
-    /// Number of tipsets requested over chain exchange (default is 200)
+    /// Number of tipsets requested over one chain exchange (default is 8)
     #[arg(long)]
-    pub req_window: Option<i64>,
+    pub req_window: Option<usize>,
```
hanabi1224, line 97: I believe this option was not working because `config.sync.req_window` was never used.

src/cli_shared/cli/mod.rs
```diff
@@ -244,8 +238,8 @@ impl CliOpts {
         // (where to find these flags, should be easy to do with structops)
 
         // check and set syncing configurations
-        if let Some(req_window) = &self.req_window {
-            cfg.sync.req_window = req_window.to_owned();
+        if let Some(req_window) = self.req_window {
+            cfg.sync.request_window = req_window;
```
hanabi1224, line 242: `cfg.sync.request_window` was a duplicate of `ChainConfig::request_window` and never used.

lemmih (approved):
In the past, `SyncConfig::req_window` and `ChainConfig::request_window` referred to two different types of requests. The former was used in `sync_headers_in_reverse` and the latter was used in `sync_messages_check_state`. However, things got mixed up and `sync_headers_in_reverse` stopped using the parameter.

Let's merge this as-is and deal with `sync_headers_in_reverse` later. The current limit of requesting 100 headers per round-trip is definitely wrong (Lotus requests 800 per round-trip which is much faster) but we can deal with it later.

PR #3609, opened 2023-10-18
feat: add `--metrics [flag]` to `forest`

hanabi1224:
## Summary of changes

Adding an option to disable metrics endpoint (just like for RPC endpoint) to facilitate writing simple test for https://github.com/ChainSafe/forest/pull/3593

- add `--metrics [flag]` to `forest` to control the availability of the Prometheus server

Closes

## Change checklist

### DIFF
--- a/src/cli_shared/cli/client.rs
+++ b/src/cli_shared/cli/client.rs
@@ -42,6 +42,7 @@ pub struct Client {
     pub data_dir: PathBuf,
     pub genesis_file: Option<String>,
     pub enable_rpc: bool,
+    pub enable_metrics_endpoint: bool,
     pub rpc_token: Option<String>,
     /// If this is true, then we do not validate the imported snapshot.
     /// Otherwise, we validate and compute the states.
@@ -82,6 +83,7 @@ impl Default for Client {
             data_dir: dir.data_dir().to_path_buf(),
             genesis_file: None,
             enable_rpc: true,
+            enable_metrics_endpoint: true,
             rpc_token: None,
             snapshot_path: None,
             snapshot: false,
--- a/src/cli_shared/cli/mod.rs
+++ b/src/cli_shared/cli/mod.rs
@@ -49,6 +49,9 @@ pub struct CliOpts {
     /// Allow RPC to be active or not (default: true)
     #[arg(short, long)]
     pub rpc: Option<bool>,
+    /// Allow Metrics endpoint to be active or not (default: true)
+    #[arg(short, long)]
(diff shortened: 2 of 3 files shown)


jdjaustin (approved):

LesnyRumcajs (commented):
The usage now would be `--metrics false` and `--metrics true`, while a possibility, would never be used. How about `--disable-metrics` boolean?

hanabi1224:
> The usage now would be `--metrics false` and `--metrics true`, while a possibility, would never be used. How about `--disable-metrics` boolean? Or `--no-metrics`.

@LesnyRumcajs Sure, using `--metrics [true|false]` is to align with the existing `--rpc [true|false]` option. Do you suggest change only this one or both?

LesnyRumcajs:
We should change both for consistency, but in separate PRs.

hanabi1224:
> We should change both for consistency, but in separate PRs.

@LesnyRumcajs  Changed to `--no-metrics`

src/cli_shared/cli/client.rs
```diff
@@ -82,6 +83,7 @@ impl Default for Client {
             data_dir: dir.data_dir().to_path_buf(),
             genesis_file: None,
             enable_rpc: true,
+            enable_metrics_endpoint: true,
```
LesnyRumcajs, line 86: why exactly do you need this filed in the `client` section?

hanabi1224: To match the logic of `enable_rpc: true,`

LesnyRumcajs: I see, I thought (dreamed) that the config is structured somehow differently.

src/cli_shared/cli/mod.rs
```diff
@@ -186,9 +189,16 @@ impl CliOpts {
...
-            cfg.client.metrics_address = metrics_address;
+
+        if self.no_metrics {
+            cfg.client.enable_metrics_endpoint = false;
+        } else {
+            cfg.client.enable_metrics_endpoint = true;
+            if let Some(metrics_address) = self.metrics_address {
+                cfg.client.metrics_address = metrics_address;
+            }
         }
```
LesnyRumcajs, line 200: A bit of code-gold if you wish
```suggestion
        cfg.client.enable_metrics_endpoint = !self.no_metrics;
        cfg.client.metrics_address = self.metrics_address.unwrap_or(cfg.client.metrics_address);

```

LesnyRumcajs (approved):

PR #3616, opened 2023-10-20
feat: `forest-tool` command for measuring RPC completeness and compatibility

### LINKED ISSUE #3584: RPC compatiblity testing
**Issue summary**

Notes and thoughts.

We want to verify that Forest is syntactically compatible with a known subset of the Lotus RPC calls. This means our requests/response pairs follow the same JSON schema. We also want to verify that our responses are sensible. Exactly what this means depends on the RPC call in question, unfortunately.

We can assert that blockchain-related RPC calls return exactly the same as their Lotus counterpart (for example, `ChainGetTipSetByHeight` must return the same answer in both Forest and Lotus). Non-blockchain RPC calls (such as `MpoolPending`) are more difficult to test, and we may settle for merely testing the JSON schema.

Using glif.io for compatibility testing isn't ideal since they do not give access to several important RPC methods.

lemmih:
## Summary of changes

- Add `forest-tool api compare` command for testing API compatibility.
- Document how to add new RPC methods (and how to debug them).
- Client-side support for `ChainGetMessagesInTipset`, `StateMinerPower`, `Discover` and `Session`.

Closes #3584

## Change checklist


src/lotus_json/mod.rs
```diff
@@ -355,10 +355,11 @@ impl<T> Stringify<T> {
     }
 }
 
+#[macro_export]
```
aatifsyed, line 358: I don't think you should do this, it exposes the macro as part of our public api:
https://doc.rust-lang.org/beta/reference/macros-by-example.html#scoping-exporting-and-importing

add `pub(crate) use lotus_json_with_self` and use it like a normal item instead

lemmih:
This PR is quite minimal. Further improvements are tracked here: https://github.com/ChainSafe/forest/issues/3641

jdjaustin (approved):

aatifsyed (approved):

PR #3622, opened 2023-10-23
runtime download for actor bundles with alt url

LesnyRumcajs:
## Summary of changes

-

Closes

## Change checklist


src/daemon/bundle.rs
```diff
@@ -1,33 +1,84 @@
...
+/// Tries to load the actor bundle from the specified location to the blockstore. If the bundle is
+/// not present, it will be generated.
+pub async fn load_actor_bundles(db: &impl Blockstore) -> anyhow::Result<()> {
+    if !are_bundles_present(db, &crate::networks::ACTOR_BUNDLES) {
+        info!("Generating actor bundle");
+        let bundle = tempfile::Builder::new()
+            .prefix("forest-actor-bundle")
+            .suffix(".car.zst")
+            .tempfile()?;
+        generate_actor_bundle(bundle.path()).await?;
```
hanabi1224, line 21: Why do we still need bundling? How about downloading and loading only the missing ones?

LesnyRumcajs: It was to avoid some code repetition, but I agree that it's wasteful. Let me make it right.

LesnyRumcajs: @hanabi1224 fixed!

hanabi1224: Thanks!

hanabi1224 (approved):

src/daemon/bundle.rs
```diff
@@ -1,33 +1,45 @@
...
 use fvm_ipld_blockstore::Blockstore;
+use std::io::Cursor;
+use tracing::warn;
 
-pub async fn load_actor_bundles(db: &impl Blockstore) -> anyhow::Result<CarHeader> {
-    const ERROR_MESSAGE: &str = "Actor bundles assets are not properly downloaded, make sure git-lfs is installed and run `git lfs pull` again. See <https://github.com/git-lfs/git-lfs/blob/main/INSTALLING.md>";
+/// Tries to load the missing actor bundles to the blockstore. If the bundle is
+/// not present, it will be downloaded.
+pub async fn load_actor_bundles(db: &impl Blockstore) -> anyhow::Result<()> {
+    stream::iter(
```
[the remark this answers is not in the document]
LesnyRumcajs: Yes!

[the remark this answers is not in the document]
LesnyRumcajs: Done. I also took the liberty of parallelizing `state-migrate` bundle download which makes it ~30-40% faster. https://github.com/ChainSafe/forest/pull/3622/commits/45b3ffa24d1226f8421fc54b9ab7b1adbd172500

src/daemon/bundle.rs
```diff
@@ -1,33 +1,45 @@
...
+                     manifest: root,
+                     url,
+                     alt_url,
+                 }| async move {
+                    let response = if let Ok(response) = http_get(url).await {
+                        response
+                    } else {
+                        warn!("failed to download bundle from primary URL, trying alternative URL");
+                        http_get(alt_url).await?
+                    };
```
lemmih, line 32: I'm kinda worried that we'll forget to mirror the files. How about we download the files from both sources and print a warning if one mirror is down (and an error if the files are different)?

src/daemon/bundle.rs
```diff
@@ -1,33 +1,45 @@
...
+                    let bytes = response.bytes().await?;
+                    let header = load_car(db, Cursor::new(bytes)).await?;
+                    ensure!(header.roots.len() == 1);
+                    ensure!(&header.roots[0] == root);
```
elmattic, line 36: nit: we could use also `ensure!(header.roots.first() == Some(root));` instead of `ensure!(&header.roots[0] == root);`. This would avoid us the need of testing for length before (and testing length before **or** after would work).

src/cli_shared/cli/mod.rs
```diff
@@ -150,6 +150,9 @@ pub struct CliOpts {
     /// Check your command-line options and configuration file if one is used
     #[arg(long)]
     pub dry_run: bool,
+    /// Skip loading actors from the actors bundle.
+    #[arg(long)]
+    pub skip_load_actors: bool,
```
elmattic, line 155: In what situations would we need to skip loading of actors bundle?

elmattic (approved):
Glad to see removal of `git-lfs`!

src/cli_shared/cli/mod.rs
LesnyRumcajs: If we don't intend to do any migrations and downloading the bundle is wasteful, then skipping the bundle makes sense. Also, it facilitates testing.

src/daemon/bundle.rs
LesnyRumcajs: https://github.com/ChainSafe/forest/pull/3622/commits/2b27412e2e10c4edd15387962b7a280d1962011b

@lemmih Does it look okay for you?

src/daemon/bundle.rs
LesnyRumcajs: I took from another piece of code. This is not exactly the same, as it would fail if there are more roots, thus ensuring the bundles are _exactly_ what we expect them to be.

elmattic: I was not speaking of removing the header roots length test, since we need both yes. Just replacing array access with `.first()`, but no big deal for sure.

lemmih (approved):

PR #3629, opened 2023-10-24
Add more contextual information to wallet script

elmattic:
## Summary of changes

This PR brings some improvement to help with the debugging of #2726.

- Log date and time just before sending the message.
- Log Forest mpool to verify that our message was correctly submitted. This also shows the message payload.
- Query the Glif API Node and check the presence of the message as well.

Closes

Sometimes, the cURL command produces `Broken pipe` errors. I don't know if we could get rid of those.

## Change checklist


aatifsyed (changes requested):
At this level of script complexity, I think it's worth moving this to `tests/`

See https://github.com/ChainSafe/forest/pull/3448

What infrastructure do you need to be able to migrate this test case?

LesnyRumcajs (commented):
I'm unsure about relying on `glif.io` in our tests. It's a single point of failure and can be down or unreachable for any reason, blocking our CI as a result. 

Can we think of other approaches?

elmattic:
> I'm unsure about relying on `glif.io` in our tests. It's a single point of failure and can be down or unreachable for any reason, blocking our CI as a result.

Could this `glif.io` check produce a failure that would block our CI? I'm asking this because this wasn't meant to be (in other words, what flags or shell command should be used for cURL to mitigate failures).
Basically, I would like to:
  - Use `glif` for troubleshooting the send issue
  - Once this goal is achieved, remove the `glif` check

I'm open to other approaches as well.

> 
> Can we think of other approaches?

LesnyRumcajs:
> > I'm unsure about relying on `glif.io` in our tests. It's a single point of failure and can be down or unreachable for any reason, blocking our CI as a result.
> 
> Could this `glif.io` check produce a failure that would block our CI? I'm asking this because this wasn't meant to be (in other words, what flags or shell command should be used for cURL to mitigate failures). Basically, I would like to:
> 
> * Use `glif` for troubleshooting the send issue
> * Once this goal is achieved, remove the `glif` check
> 
> I'm open to other approaches as well.
> 
> > Can we think of other approaches?

At the very least, we could skip the check if `glif.io` or the RPC call is returning anything other than `200`.

elmattic:
> > > I'm unsure about relying on `glif.io` in our tests. It's a single point of failure and can be down or unreachable for any reason, blocking our CI as a result.
> > 
> > 
> > Could this `glif.io` check produce a failure that would block our CI? I'm asking this because this wasn't meant to be (in other words, what flags or shell command should be used for cURL to mitigate failures). Basically, I would like to:
> > 
> > * Use `glif` for troubleshooting the send issue
> > * Once this goal is achieved, remove the `glif` check
> > 
> > I'm open to other approaches as well.
> > > Can we think of other approaches?
> 
> At the very least, we could skip the check if `glif.io` or the RPC call is returning anything other than `200`.

Sounds good.  But after giving some thought, we don't need to merge this code; just re-running the workflow until the error happens will be enough for me.

LesnyRumcajs:
> Sounds good. But after giving some thought, we don't need to merge this code; just re-running the workflow until the error happens will be enough for me.

In that case, I'd recommend either running it in a loop in the script itself or make the entire job run multiple times (either copy-paste or perhaps use `strategy`) so you don't have to manually trigger it. Or do the same on a droplet to not exhaust GH workers.

elmattic:
Closing it as I found out the message does not leave Forest node in case of failures.

PR #3642, opened 2023-10-30
add GET timeout to mirror tests

LesnyRumcajs:
## Summary of changes

- Addresses a timeout [failure](https://github.com/ChainSafe/forest/actions/runs/6692088982/job/18180570367?pr=3593#step:7:1095) in the CI. Now, each GET will take no more than 30s. Given many futures are spawned in parallel, this should not compound. Note that this may be too little on a horrible internet connection, but on a regular one, it should be fine. The test will fail if both sources are completely down (this will mean we don't have enough redundancy)

Closes

## Change checklist

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -9,7 +9,7 @@ slow-timeout = { period = "120s", terminate-after = 3 }
 # It is only run on CI, so we can afford to be more patient.
 [[profile.default.overrides]]
 filter = 'test(networks::actors_bundle::tests::check_bundles_are_mirrored)'
-slow-timeout = { period = "60s", terminate-after = 3 }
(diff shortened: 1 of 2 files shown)


LesnyRumcajs:
This timeout is getting a bit too high. I'm inclined to leave only the check for existence (that is to say, no 404), and allow handle proper replication in https://github.com/ChainSafe/forest-iac/issues/323. What do you think @lemmih?

LesnyRumcajs:
The reason for the timeouts is most likely bad internet connection for the `fuzzy` worker (~3 Mbps).

lemmih (approved):

hanabi1224 (approved):

PR #3650, opened 2023-11-01
refactor: embedded index for `forest.car.zst`

aatifsyed:
# Changes
- Simpler objects like `Slot`.
- Simpler de/serialization API.
- Improved `Builder/Writer` paradigm.
- Reorganised so the index isn't squirreled away in another module
- Less plumbing for `distance` etc
- Improved separation of concerns for `Reader`
- Only one `write_into` method
- Improve peak memory usage

master:
image

feature:
image


src/db/car/forest.rs
```diff
@@ -268,24 +274,22 @@ impl Encoder {
...
 
         // Create index
-        let index_offset = offset as u64 + 8;
-        let builder = CarIndexBuilder::new(cid_mapping.into_iter());
-        write_skip_frame_header_async(sink, builder.encoded_len()).await?;
-        builder.write_async(sink).await?;
+        let mut index = vec![];
+        index::write(cid_to_frame_offset, &mut index).expect("Vec has infallible IO");
+        write_skip_frame_header_async(sink, index.len().try_into().unwrap()).await?;
+        sink.write_all(&index).await?;
```
lemmih, line 288: Streaming the index to disk would save us several GiBs of memory. My gut feeling is that this optimization belongs in this PR.

lemmih (changes requested):
Looks good. A few things we need to align on:

- We're not using a custom hash function. We've talked about this before, and it's vital that we're on the same page about this. Letting the caller decide whether to use `blake2b`, `sha256` or `md5` isn't the same as having a custom hash function. And it should certainly not be surprising that, say, `blake2b` has few collisions in practice.
- A lot of the testing has been removed. Everything that was tested before should be tested in this PR as well.
- Exporting should be a bit faster now. Let's benchmark and include the results in the PR description.
- The index-by-sorting method lends itself well to streaming, which would reduce peak memory requirements significantly. You might argue that such a change should be in a separate PR, but I don't think that is the case. We can discuss in Slack.

src/utils/db/car_index/tests.rs
```diff
@@ -1,92 +0,0 @@
...
-    let table = mk_table(&entries);
-    for (&hash, value_set) in map.iter() {
-        assert_eq!(&AHashSet::from_iter(query(&table, hash)), value_set);
-    }
-}
-
-// Identical to HashMap<Hash, HashSet<FrameOffset>> with few hash collisions
-// but all hash values map to optimal_position 0..10
-#[quickcheck]
-fn lookup_clash_many(mut entries: Vec<(Hash, FrameOffset)>) {
```
aatifsyed, line 80: ported

src/utils/db/car_index/tests.rs
```diff
@@ -1,92 +0,0 @@
...
-#[quickcheck]
-fn lookup_singleton(key: Hash, value: FrameOffset) {
-    let table = mk_table(&[(key, value)]);
-    assert_eq!(query(&table, key), vec![value]);
-    assert_eq!(query(&table, !key), Vec::<FrameOffset>::new());
-}
-
-// Identical to HashMap<Hash, HashSet<FrameOffset>> with almost no collision
-#[quickcheck]
-fn lookup_wide(entries: Vec<(Hash, FrameOffset)>) {
```
aatifsyed, line 40: I think this is equal the `hashmap_of_cids` test

src/utils/db/car_index/tests.rs
```diff
@@ -1,92 +0,0 @@
...
-            .and_modify(|set: &mut AHashSet<FrameOffset>| {
-                set.insert(position);
-            })
-            .or_insert(AHashSet::from([position]));
-    }
-    map
-}
-
-#[quickcheck]
-fn lookup_singleton(key: Hash, value: FrameOffset) {
```
aatifsyed, line 32: I think this is covered by the `hashmap_of_cids` test

src/utils/db/car_index/tests.rs
```diff
@@ -1,92 +0,0 @@
...
-    let map = mk_map(&entries);
-    let table = mk_table(&entries);
-    for (&hash, value_set) in map.iter() {
-        assert_eq!(&AHashSet::from_iter(query(&table, hash)), value_set);
-    }
-}
-
-// Identical to HashMap<Hash, HashSet<FrameOffset>> with many collision
-#[quickcheck]
-fn lookup_narrow(mut entries: Vec<(Hash, FrameOffset)>) {
```
aatifsyed, line 50: I don't really understand this test

src/db/car/forest.rs
aatifsyed: I've used `SyncIoBridge`, but that requires `#[tokio::test(flavor = "multi_thread")]`, which feels a bit ugly

aatifsyed (commented):
Annotated which tests I've ported and which I haven't

src/utils/db/car_index/tests.rs
lemmih: This is your `everything_maps_to_first_10_slots` test.

lemmih: Ah, this is not `everything_maps_to_first_10_slots`.

This tests what happens when there are hash collisions. As far as I can tell, you don't have any tests for hash collisions.

lemmih (changes requested):
Looks good. Two things:

1. You still keep two copies of the index in memory. We can save a few GiBs by only keeping one copy of the index in memory before it is written to disk.
2. I don't see any tests with hash collisions.

lemmih:
> Looks good. Two things:
> 
>     1. You still keep two copies of the index in memory. We can save a few GiBs by only keeping one copy of the index in memory before it is written to disk.
> 
>     2. I don't see any tests with hash collisions.

Let's include results from `forest-tool benchmark export`. Ideally, runtime should be slightly shorter and peak memory usage should be halved.

lemmih (approved):
Looking good. So excited to see benchmark results.

src/db/car/forest.rs
```diff
@@ -84,21 +89,21 @@ pub struct ForestCar<ReaderT> {
...
     roots: Vec<Cid>,
 }
 
 impl<ReaderT: super::RandomAccessFileReader> ForestCar<ReaderT> {
-    pub fn new(reader: ReaderT) -> io::Result<Self> {
+    pub fn new(reader: ReaderT) -> io::Result<ForestCar<ReaderT>> {
         let (header, footer) = Self::validate_car(&reader)?;
 
-        let index = CarIndex::open(reader, footer.index)?;
+        let indexed = index::Reader::new(positioned_io::Slice::new(reader, footer.index, None))?;
```
hanabi1224, line 102: The `reader` here can be of `EitherMmapOrRandomAccessFile` type, have you verifyied that `positioned_io::Slice` works fine with `mmap2` with no performance panity?

hanabi1224 (approved):

src/db/car/forest.rs
```diff
@@ -268,24 +274,21 @@ impl Encoder {
         offset += header_len;
 
         // Write seekable zstd and collect a mapping of CIDs to frame_offset+data_offset.
-        let mut cid_mapping = Vec::new();
+        let mut builder = index::Builder::new();
```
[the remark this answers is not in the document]
aatifsyed: It is backwards compatible :)

src/db/car/forest.rs
aatifsyed: It should, positioned_io::Slice simply does an addition on each read_at call

PR #3658, opened 2023-11-02
daemon: download only bundles for the given network

LesnyRumcajs:
## Summary of changes

- there is no point in downloading bundles for calibnet and devnet when running mainnet-only. This change makes it so that only actor bundles for the network at hand are downloaded.

Closes

## Change checklist

### DIFF
--- a/src/daemon/bundle.rs
+++ b/src/daemon/bundle.rs
@@ -22,8 +22,10 @@ pub async fn load_actor_bundles(
         ACTOR_BUNDLES
             .iter()
             .filter(|bundle| {
-                !db.has(&bundle.manifest).unwrap_or(false)
-                    && discriminant(network) == discriminant(&bundle.network)
+                !db.has(&bundle.manifest).unwrap_or(false) &&
+                // Comparing only the discriminant is enough. All devnets share the same
+                // actor bundle.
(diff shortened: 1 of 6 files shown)


hanabi1224 (approved):

src/daemon/bundle.rs
```diff
@@ -2,27 +2,35 @@
...
+    db: &impl Blockstore,
+    network: &NetworkChain,
+) -> anyhow::Result<()> {
     FuturesUnordered::from_iter(
         ACTOR_BUNDLES
             .iter()
-            .filter(|bundle| !db.has(&bundle.manifest).unwrap_or(false))
+            .filter(|bundle| {
+                !db.has(&bundle.manifest).unwrap_or(false)
+                    && discriminant(network) == discriminant(&bundle.network)
```
aatifsyed, line 26: could you comment why you're using this instead of simply `==`?

aatifsyed: i.e is there a devnet bundle that's shared amongst all of them?

LesnyRumcajs: Yeah. All devnets share the same set of bundles. I added a comment for clarity.

ruseinov (approved):

PR #3669, opened 2023-11-04
rfc: generate our API from go code

aatifsyed:
1. Take the go code from lotus: https://github.com/ChainSafe/forest/blob/5a05dd184a6e8f9eaeb21c8f122b3f9a0b89b974/codegen/api/api_full.go#L322
2. Define analogues for go types: https://github.com/ChainSafe/forest/blob/5a05dd184a6e8f9eaeb21c8f122b3f9a0b89b974/codegen/src/main.rs#L257-L264
3. Generate a `trait Api` for everything we can: https://github.com/ChainSafe/forest/blob/5a05dd184a6e8f9eaeb21c8f122b3f9a0b89b974/src/rpc2.rs#L1-L4

You get a printout of the types you need for the remaining endpoints:
```
AuthNew depends on the following unmapped types:
- []auth.Permission
```

I think the core mapping can be the basis of our API definition. We can go via OpenAPI or OpenRPC as we prefer.
Traits, signatures etc TBD


aatifsyed:
Notes:
- Not everything will be `HasLotusJson`, but it _does_ need to impl `JsonSchema`/`Serialize` etc
- Where will the rate handlers and auth etc go?

PR #3689, opened 2023-11-08
[Feature] Mark-and-sweep GC.

ruseinov:
## Summary of changes

- Introduces Mark and Sweep GC
- Fixes the migration for it

Closes https://github.com/ChainSafe/forest/issues/3072

## Change checklist


src/db/migration/v0_14_0.rs
```diff
@@ -63,17 +63,21 @@ impl MigrationOperation for Migration0_15_2_0_16_0 {
...
+        if old_car_db_path.is_dir() {
+            info!(
+                "copying snapshot from {source_db} to {temp_db_path}",
+                source_db = old_car_db_path.display(),
+                temp_db_path = new_car_db_path.display()
+            );
+
+            fs_extra::fs_extra::copy_items(
+                &[old_car_db_path.as_path()],
+                new_car_db_path.clone(),
```
LesnyRumcajs, line 77: does it need to be cloned?

LesnyRumcajs (approved):

src/db/migration/v0_14_0.rs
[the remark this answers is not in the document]
ruseinov: Yep, I'm aware. Didn't want any extra changes here, but might as well do it.

src/db/migration/v0_14_0.rs
ruseinov: It's an exact copy from `MigrationVoid`, I honestly didn't check the signatures. Looks like it does not. In which case we should fix this in all the migrations. I'll create an issue for that.

LesnyRumcajs (approved):

Issue #3639, opened 2023-10-30
RPC method v0 tracking issue


lemmih:
**Issue summary**

Version 0 methods:

* General
  * [Closing](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Closing)
  * [Discover](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Discover)
  * [Session](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Session)
  * [Shutdown](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Shutdown)
  * [Version](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Version)
* [Auth](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Auth)
  * [AuthNew](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#AuthNew)
  * [AuthVerify](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#AuthVerify)
* [Beacon](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Beacon)
  * [BeaconGetEntry](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#BeaconGetEntry)
* [Chain](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#Chain)
  * [ChainDeleteObj](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#ChainDeleteObj)
  * [ChainExport](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#ChainExport)
  * [ChainGetBlock](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#ChainGetBlock)
  * [ChainGetBlockMessages](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#ChainGetBlockMessages)
  * [ChainGetGenesis](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md#ChainGetGenesis)
[...]

PR #3705, opened 2023-11-09
feat: implement `StateMinerActiveSectors` rpc method

lemmih:
## Summary of changes

- Implement `StateMinerActiveSectors` RPC method, including compliance tests.

Part of #3639

## Change checklist


src/rpc/state_api.rs
```diff
@@ -196,9 +196,9 @@ pub(in crate::rpc) async fn state_miner_active_sectors<DB: Blockstore>(
 
     // Collect active sectors from each partition in each deadline.
     let mut active_sectors = vec![];
-    miner_state.for_each_deadline(policy, bs, |_dlidx, dl| {
-        dl.for_each(bs, |_partidx, part| {
-            active_sectors.push(part.active_sectors());
+    miner_state.for_each_deadline(policy, bs, |_dlidx, deadline| {
```
ruseinov, line 199: We are basically repeating ourselves here, `Filecoin.StateMinerFaults` and perhaps another couple of RPC methods. In Lotus they are using callbacks to deal with that, not sure if we want to replicate this.

We could do something like:

`fn callback(enum: SectorType) -> FnMut.. {
   match SectorType {
      Active => sectors.push(partition.active_sectors());
...
    }
}` 

It's not a whole lot of code, but having a helper for this could be more comprehensive and less repetitive. I'm happy to take this on as a follow-up if it makes sense.

LesnyRumcajs (approved):
changelog entry?

lemmih:
> changelog entry?

I'm leaning towards leaving out entries for individual RPC methods. Once we enforce compliance with subsets of the RPC API, I'll include it in the changelog.

sudo-shashank (approved):

src/rpc/state_api.rs
lemmih: Yeah, let's look at refactoring this soon.

PR #3712, opened 2023-11-14
fix(libp2p): reduce connection limits

hanabi1224:
## Summary of changes

As a follow-up of https://github.com/ChainSafe/forest/pull/3713
Currently, `target_peer_count` is only used to decide whether or not to perform proactive kad queries, while it does not prevent forest from accepting inbound connections. When total connections reach 2k+(test on @0.16.0, @0.15.2 and @0.15.0), timeout rate of `hello` and `bitswap` become high. This PR tries to setup `connection_limits` behaviour with `target_peer_count` to limit the total active connections

Manually tested on mainnet, smaller connection limits could also mitigate timeouts in `bitswap` and `hello` 

- Calculate the connection limits from `target_peer_count`, instead of using static values.

Manually tested on mainnet, forest is able to sync with normal performance.

Closes

## Change checklist

### DIFF
--- a/src/libp2p/behaviour.rs
+++ b/src/libp2p/behaviour.rs
@@ -106,13 +106,30 @@ impl ForestBehaviour {
             .target_peer_count(config.target_peer_count as u64)
             .finish()?;
 
+        const MAX_ESTABLISHED_PER_PEER: u32 = 4;
         let connection_limits = connection_limits::Behaviour::new(
             connection_limits::ConnectionLimits::default()
-                .with_max_pending_incoming(Some(4096))
-                .with_max_pending_outgoing(Some(8192))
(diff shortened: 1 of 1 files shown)


ruseinov (approved):
Looks good, could be great to have a comment that explains why we are multiplying by `MAX_ESTABLISHED_PER_PEER`. It seems obvious, but might save the reader some time understanding the code.

LesnyRumcajs (approved):
Hopefully, we won't end up with bad peers and Forest dying with them, which was the case before.

hanabi1224:
> Hopefully, we won't end up with bad peers and Forest dying with them, which was the case before.

@LesnyRumcajs I will try to follow up with some extra peer-banning logic to eliminate bad peers

PR #3714, opened 2023-11-14
fix(chain muxer): do not assemble gossipsub blocks into full tipsets during bootstrap

hanabi1224:
## Summary of changes

As a follow-up of https://github.com/ChainSafe/forest/pull/3713
It should be unnecessary to make bitswap requests during BOOTSTRAP. This PR tries to avoid that and improve BOOTSTRAP performance 

- Do not assemble gossipsub blocks into full tipsets during bootstrap, to avoid making unnecessary underlying bitswap requests
- Do not assemble gossipsub blocks into full tipsets when they are older than `head_epoch - chain_finality`, to reduce unnecessary underlying bitswap requests
- Do not assemble hello request into full tipsets when they are older than `head_epoch - chain_finality`, to reduce unnecessary underlying chain exchange requests

Manually tested on mainnet, forest is able to sync with normal performance.

Closes

## Change checklist


src/chain_sync/chain_muxer.rs
```diff
@@ -365,28 +370,48 @@ where
...
-                    Ok(tipset) => tipset,
-                    Err(why) => {
-                        debug!("Querying full tipset failed: {}", why);
-                        return Err(why);
+                // Retrieve tipset from store based on passed in TipsetKeys
+                if let Ok(Some(tipset)) = chain_store.load_tipset(&tipset_keys) {
+                    // Skip old tipsets
+                    if chain_store.is_epoch_finalized(tipset.epoch()) {
+                        return Ok(None);
+                    } else {
```
ruseinov, line 395: nit: else is redundant here.

src/chain_sync/chain_muxer.rs
```diff
@@ -365,28 +370,48 @@ where
...
+                        )
+                        .await
+                        {
+                            Ok(tipset) => tipset,
+                            Err(why) => {
+                                debug!("Querying full tipset failed: {}", why);
+                                return Err(why);
+                            }
+                        };
+                        (full_tipset, source)
```
ruseinov, line 410: nit: if you return here - you can get rid of the `else` block below and just do a return there too.

src/chain_sync/chain_muxer.rs
```diff
@@ -436,9 +461,24 @@ where
                         .with_label_values(&[metrics::values::PUBSUB_BLOCK])
                         .inc();
                     // Assemble full tipset from block
-                    let tipset =
-                        Self::gossipsub_block_to_full_tipset(b, source, network.clone()).await?;
-                    (tipset, source)
+                    if let GossipBlockProcessingStrategy::GetFullTipset =
```
ruseinov, line 464: nit: same here, there could be less nesting if we get rid of the `else` blocks.

src/chain_sync/chain_muxer.rs
```diff
@@ -517,7 +550,15 @@ where
             .with_label_values(&[source.to_string().as_str()])
             .set(tipset.epoch());
 
-        Ok(Some((tipset, source)))
+        if should_skip {
```
ruseinov, line 553: nit: unnecessary `else` block.

ruseinov:
Some minor readability nits. It would also be nice to see some tests for this if at all possible.

src/chain_sync/chain_muxer.rs
hanabi1224: Fixed

src/chain_sync/chain_muxer.rs
hanabi1224: Fixed.

src/chain_sync/chain_muxer.rs
hanabi1224: The `full_tipset` is assigned to a variable and processed later, early return here breaks the logic.

src/chain_sync/chain_muxer.rs
hanabi1224: Fixed.

hanabi1224:
> Some minor readability nits. It would also be nice to see some tests for this if at all possible.

@ruseinov This should have been covered by the integration test scripts, and I've also manually tested on mainnet

hanabi1224:
Kindly ping.  @lemmih any feedback on this one?

lemmih:
> Kindly ping. @lemmih any feedback on this one?

I'm unsure what this PR is attempting to do. Is it purely a performance optimization? If so, how much does it improve performance? Does it justify the extra code complexity?

hanabi1224:
> I'm unsure what this PR is attempting to do. Is it purely a performance optimization? If so, how much does it improve performance? Does it justify the extra code complexity?

@lemmih Besides performance optimization, I think it makes the logic of chain muxer more correct and clean. When a forest node is bootstrapped with a snapshot that is a few thousand epochs behind the network, it takes hours to sync(which is the BOOTSTRAP stage). During such a long BOOTSTRAP, a forest node makes 10k-100k bitswap requests(from my observation) to assemble the gossip blocks(passively received from the p2p network) into full tipsets and store the messages into its database. Neither the network nor the database activities seem to be necessary. Disabling the gossip-block-to-full-tipset assembly saves network bandwidth and disk space. Does that make sense?

LesnyRumcajs:
@hanabi1224 @lemmih what do we do with this PR?

lemmih:
> @hanabi1224 @lemmih what do we do with this PR?

I'm leaning toward closing this PR.

PR #3719, opened 2023-11-16
Implemented Filecoin.Filecoin.StateReadState API

sudo-shashank:
## Summary of changes

-

Closes

## Change checklist

### DIFF
--- a/src/rpc/state_api.rs
+++ b/src/rpc/state_api.rs
@@ -7,7 +7,7 @@ use crate::cid_collections::CidHashSet;
 use crate::ipld::json::IpldJson;
 use crate::libp2p::NetworkMessage;
 use crate::lotus_json::LotusJson;
-use crate::rpc_api::data_types::{MarketDeal, MessageLookup, RPCState};
+use crate::rpc_api::data_types::{ApiActorState, MarketDeal, MessageLookup, RPCState};
 use crate::shim::{
     address::Address, clock::ChainEpoch, executor::Receipt, message::Message,
     state_tree::ActorState, version::NetworkVersion,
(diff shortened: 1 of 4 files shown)


src/rpc/state_api.rs
```diff
@@ -377,3 +377,27 @@ pub(in crate::rpc) async fn state_get_randomness_from_beacon<
...
+    let actor = data
+        .state_manager
+        .get_actor(&addr, *ts.parent_state())?
+        .ok_or("Actor address could not be resolved")?;
+    let blk = data
+        .state_manager
+        .blockstore()
+        .get(&actor.state)?
+        .ok_or("Failed to get block from blockstore")?;
+    let state = fvm_ipld_encoding::from_slice::<Vec<Cid>>(&blk)?[0];
```
ruseinov, line 396: nit: All of this could be a method in `state_manager`, similar to other RPC methods

elmattic: @ruseinov I think the best place for `<category>_<endpoint>` is in `src/rpc/<category>.rs`. Do you have something else in mind?

elmattic, line 396: What happens if the vector of `Cids` is empty? Is it even possible?

lemmih: Good question. Let's add a test for that.

sudo-shashank: added

src/rpc_api/data_types.rs
```diff
@@ -299,19 +299,38 @@ impl HasLotusJson for ActorState {
...
     balance: TokenAmount,
     #[serde(with = "crate::lotus_json")]
     code: Cid,
     #[serde(with = "crate::lotus_json")]
+    state: ApiState,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Clone)]
+#[serde(rename_all = "PascalCase")]
+struct ApiState {
```
lemmih, line 315: Why hoist this out as a separate type? Why not directly use `Ipld`?

src/rpc/state_api.rs
ruseinov: And then that method is used the way @elmattic suggests:

https://github.com/ChainSafe/forest/blob/cc652738faa0c8e7d626d474e8e3838912ce0dcc/src/rpc/state_api.rs#L100-L111

src/rpc_api/data_types.rs
sudo-shashank: this is to match lotus repsonse 
```
Object {"Balance": String("0"), "Code": Object {"/": String("bafk2bzacecioupndtcnyw6iq2hbrxag3aufvczlv5nobnfbkbywqzcyfaa376")}, "State": Object {"BuiltinActors": Object {"/": String("bafy2bzacecfgf3634uaq7lynhauelvpz77ebgouejdx6eqnokajdknylv4m6c")}}}
```

lemmih: Hmm, quite surprising that it has a `BuiltinActors` field... I think the Lotus RPC endpoint is buggy. We'll just have to copy those bugs.

lemmih (approved):

ruseinov:
@sudo-shashank please read this thread: https://github.com/ChainSafe/forest/pull/3719#discussion_r1395345666

It's not a blocker, but it would make the codebase more uniform. Basically stick most of the logic into `state_manager` and just call a couple of methods in the `controller`. 

Approving this anyway.

ruseinov (approved):
looks good, would still like to see most of the code moved from `state_read_state` into `read_state` method in `state_manager` and just call it from `state_read_state`. 

Not a huge deal though.

PR #3735, opened 2023-11-22
feat: implement Filecoin.StateAccountKey RPC API

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/3639

- implement Filecoin.StateAccountKey RPC API

(Manually changed the code to compare the last 500 tipsets)
```
     Running `target/release/forest-tool api compare /home/me/fr/snapshots/calibnet/forest_snapshot_calibnet_2023-11-21_height_1108092.forest.car.zst --filter StateAccount`
| RPC Method                     | Forest | Lotus |
|--------------------------------|--------|-------|
| Filecoin.StateAccountKey (476) | Valid  | Valid |
```

Lotus code:
```go
func (m *StateModule) StateAccountKey(ctx context.Context, addr address.Address, tsk types.TipSetKey) (address.Address, error) {
	ts, err := m.Chain.GetTipSetFromKey(ctx, tsk)
	if err != nil {
		return address.Undef, xerrors.Errorf("loading tipset %s: %w", tsk, err)
	}

	return m.StateManager.ResolveToDeterministicAddress(ctx, addr, ts)
}

// ResolveToDeterministicAddress is similar to `vm.ResolveToDeterministicAddr` but does not allow `Actor` type of addresses.
// Uses the `TipSet` `ts` to generate the VM state.
func (sm *StateManager) ResolveToDeterministicAddress(ctx context.Context, addr address.Address, ts *types.TipSet) (address.Address, error) {
	switch addr.Protocol() {
	case address.BLS, address.SECP256K1, address.Delegated:
		return addr, nil
	case address.Actor:
		return address.Undef, xerrors.New("cannot resolve actor address to key address")
	default:
	}

	if ts == nil {
		ts = sm.cs.GetHeaviestTipSet()
	}

	cst := cbor.NewCborStore(sm.cs.StateBlockstore())

	// First try to resolve the actor in the parent state, so we don't have to compute anything.
	tree, err := state.LoadStateTree(cst, ts.ParentState())
	if err != nil {
		return address.Undef, xerrors.Errorf("failed to load parent state tree at tipset %s: %w", ts.Parents(), err)
	}

	resolved, err := vm.ResolveToDeterministicAddr(tree, cst, addr)
	if err == nil {
		return resolved, nil
	}

	// If that fails, compute the tip-set and try again.
	st, _, err := sm.TipSetState(ctx, ts)
	if err != nil {
		return address.Undef, xerrors.Errorf("resolve address failed to get tipset %s state: %w", ts, err)
	}

	tree, err = state.LoadStateTree(cst, st)
	if err != nil {
		return address.Undef, xerrors.Errorf("failed to load state tree at tipset %s: %w", ts, err)
	}

	return vm.ResolveToDeterministicAddr(tree, cst, addr)
}

// ResolveToDeterministicAddr returns the public key type of address
// (`BLS`/`SECP256K1`) of an actor identified by `addr`, or its
// delegated address.
func ResolveToDeterministicAddr(state types.StateTree, cst cbor.IpldStore, addr address.Address) (address.Address, error) {
	if addr.Protocol() == address.BLS || addr.Protocol() == address.SECP256K1 || addr.Protocol() == address.Delegated {
		return addr, nil
	}

	act, err := state.GetActor(addr)
	if err != nil {
		return address.Undef, xerrors.Errorf("failed to find actor: %s", addr)
	}

	if state.Version() >= types.StateTreeVersion5 {
		if act.Address != nil {
			// If there _is_ an f4 address, return it as "key" address
			return *act.Address, nil
		}
	}

	aast, err := account.Load(adt.WrapStore(context.TODO(), cst), act)
	if err != nil {
		return address.Undef, xerrors.Errorf("failed to get account actor state for %s: %w", addr, err)
	}
	return aast.PubkeyAddress()

}
```

Closes

## Change checklist

### DIFF
--- a/src/shim/state_tree.rs
+++ b/src/shim/state_tree.rs
@@ -292,7 +292,7 @@ where
         }
     }
 
-    /// ResolveToDeterministicAddr returns the public key type of
+    /// Returns the public key type of
     /// address(`BLS`/`SECP256K1`) of an actor identified by `addr`,
     /// or its delegated address.
     pub fn resolve_to_deterministic_addr(
--- a/src/shim/state_tree.rs
+++ b/src/shim/state_tree.rs
@@ -308,7 +308,8 @@ where
                     .get_actor(&addr)?
                     .with_context(|| format!("failed to find actor: {addr}"))?;
 
-                // if state.Version() >= types.StateTreeVersion5
+                // A workaround to implement `if state.Version() >= types.StateTreeVersion5`
+                // When state tree version is not available in rust APIs
                 if !matches!(self, Self::FvmV2(_) | Self::V0(_)) {
                     if let Some(address) = actor.delegated_address {
                         return Ok(address.into());
--- a/src/shim/state_tree.rs
+++ b/src/shim/state_tree.rs
@@ -291,6 +291,36 @@ where
             StateTree::V0(_) => bail!("StateTree::set_actor not supported on old state trees"),
         }
     }
+
+    /// ResolveToDeterministicAddr returns the public key type of
+    /// address(`BLS`/`SECP256K1`) of an actor identified by `addr`,
+    /// or its delegated address.
+    pub fn resolve_to_deterministic_addr(
+        &self,
+        store: &impl Blockstore,
+        addr: Address,
+    ) -> anyhow::Result<Address> {
+        use crate::shim::address::Protocol::*;
+        match addr.protocol() {
+            BLS | Secp256k1 | Delegated => Ok(addr),
+            _ => {
+                let actor = self
+                    .get_actor(&addr)?
+                    .with_context(|| format!("failed to find actor: {addr}"))?;
+
+                // if state.Version() >= types.StateTreeVersion5
+                if matches!(self, Self::FvmV4(_) | Self::FvmV3(_)) {
+                    if let Some(address) = actor.delegated_address {
+                        return Ok(address.into());
+                    }
+                }
+
+                let account_state =
+                    fil_actor_interface::account::State::load(store, actor.code, actor.state)?;
+                Ok(account_state.pubkey_address().into())
+            }
+        }
+    }
 }
 
 /// `Newtype` to wrap different versions of `fvm::state_tree::ActorState`
--- a/src/shim/state_tree.rs
+++ b/src/shim/state_tree.rs
@@ -309,7 +309,7 @@ where
                     .with_context(|| format!("failed to find actor: {addr}"))?;
 
                 // if state.Version() >= types.StateTreeVersion5
-                if matches!(self, Self::FvmV4(_) | Self::FvmV3(_)) {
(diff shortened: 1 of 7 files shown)


lemmih (approved):

src/shim/state_tree.rs
```diff
@@ -291,6 +291,36 @@ where
...
+    ) -> anyhow::Result<Address> {
+        use crate::shim::address::Protocol::*;
+        match addr.protocol() {
+            BLS | Secp256k1 | Delegated => Ok(addr),
+            _ => {
+                let actor = self
+                    .get_actor(&addr)?
+                    .with_context(|| format!("failed to find actor: {addr}"))?;
+
+                // if state.Version() >= types.StateTreeVersion5
```
LesnyRumcajs, line 311:
```suggestion
```

src/shim/state_tree.rs
```diff
@@ -291,6 +291,36 @@ where
...
+                if matches!(self, Self::FvmV4(_) | Self::FvmV3(_)) {
+                    if let Some(address) = actor.delegated_address {
+                        return Ok(address.into());
+                    }
```
LesnyRumcajs, line 315: Do we even need to check if it's FVM4 or FVM3? What if (when) we get FVM5?

src/shim/state_tree.rs
hanabi1224: This is the Go code. Rust api does not expose state tree version so I have to use some workaround here

src/shim/state_tree.rs
hanabi1224: That's good point, it's better to check older versions

src/shim/state_tree.rs
LesnyRumcajs: So let's better have a comment than a commented-out Go code. :)

src/shim/state_tree.rs
hanabi1224: Fixed

LesnyRumcajs: But do we need to have this check at all? I mean, paranoid checks are fine by me, but I believe the `delegated_address` should be `None` for older versions.

LesnyRumcajs (approved):

src/shim/state_tree.rs
hanabi1224: > But do we need to have this check at all? I mean, paranoid checks are fine by me, but I believe the `delegated_address` should be `None` for older versions.

I think you are right, but I'm inclined to keep the logic the same with Go, what do you think?

```go
if state.Version() >= types.StateTreeVersion5 {
		if act.Address != nil {
			// If there _is_ an f4 address, return it as "key" address
			return *act.Address, nil
		}
	}
```

LesnyRumcajs: Sure, it's a cheap operation anyway.

Issue #3640, opened 2023-10-30
RPC method v1 tracking issue


lemmih:
**Issue summary**

Version 1 methods:
* General
  * [Closing](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Closing)
  * [Discover](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Discover)
  * [Session](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Session)
  * [Shutdown](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Shutdown)
  * [Version](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Version)
* [Auth](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Auth)
  * [AuthNew](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#AuthNew)
  * [AuthVerify](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#AuthVerify)
* [Chain](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#Chain)
  * [ChainBlockstoreInfo](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#ChainBlockstoreInfo)
  * [ChainCheckBlockstore](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#ChainCheckBlockstore)
  * [ChainDeleteObj](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#ChainDeleteObj)
  * [ChainExport](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#ChainExport)
  * [ChainExportRangeInternal](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#ChainExportRangeInternal)
  * [ChainGetBlock](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#ChainGetBlock)
[...]

PR #3794, opened 2023-12-05
feat: implement `EthAccounts` and validate against Lotus

jdjaustin:
## Summary of changes

- Implement `EthAccounts` and test against Lotus

Result:

| RPC Method           | Forest | Lotus |
|----------------------|--------|-------|
| Filecoin.EthAccounts | Valid  | Valid |

Closes associated task in #3640 

## Change checklist


lemmih (approved):

src/rpc/eth_api.rs
```diff
@@ -6,6 +6,11 @@ use crate::rpc_api::data_types::RPCState;
 use fvm_ipld_blockstore::Blockstore;
 use jsonrpc_v2::{Data, Error as JsonRpcError};
 
+// EthAccounts will always return [] since we don't expect Forest to manage private keys
```
LesnyRumcajs, line 9: Should this be a `///`?

jdjaustin: Good question, I see both non-doc comments and doc comments for our RPC methods. In this case it's consistent with the Lotus codebase, but if you think this should be changed to a doc comment I'm happy to make the change.

LesnyRumcajs: IMO if it's above the method signature, it should be a doc comment. If it's inside the method body, it should be an ordinary comment. I'm good with either approach. With the current phrasing, probably the latter.

LesnyRumcajs (approved):

PR #3817, opened 2023-12-12
[Feature] Implement StateVerifiedClientStatus Lotus-compatible RPC.

ruseinov:
## Summary of changes

- Implement StateVerifiedClientStatus Lotus-compatible RPC.

work on https://github.com/ChainSafe/forest/issues/3639

Tested with: `cargo run --bin=forest-tool -- api compare --filter StateVerifiedClientStatus --lotus /dns/127.0.0.1/tcp/1234/http snapshot-latest`.

I have found a way to verify at least one of the two Lotus codepaths, so that's something. 

## Change checklist


sudo-shashank:
@ruseinov you can verify with https://github.com/filecoin-project/builtin-actors/tree/v12.0.0

src/tool/subcommands/api_cmd.rs
```diff
@@ -465,6 +465,17 @@ fn snapshot_tests(store: &ManyCar, n_tipsets: usize) -> anyhow::Result<Vec<RpcTe
     tests.extend(state_tests(&shared_tipset));
     tests.extend(eth_tests_with_tipset(&shared_tipset));
 
+    // Those proven hard to verify as most of the time both lotus and forest return 'null'. So it's
+    // hardcoded.
```
aatifsyed, line 469: this doesn't quite make sense to me - consider rewording?

aatifsyed (approved):

src/tool/subcommands/api_cmd.rs
ruseinov: This note implies understanding of snapshot tests, which takes `n` last tipsets and tries to verify the results for both Lotus and Forest using the addresses found. Since last `n` tipsets data doesn't return anything but `null` for both - the addresses are hardcoded.

src/state_manager/mod.rs
```diff
@@ -1313,6 +1314,35 @@ where
...
+    ) -> anyhow::Result<Option<DataCap>> {
+        let id = self.lookup_id(addr, ts)?.expect("actor not found");
+        let network_version = self.get_network_version(ts.epoch());
+
+        // This is a copy of Lotus code, we need to treat all the actors below version 9
+        // differently. Which maps to network below version 17.
+        if (u32::from(network_version.0)) < 17 {
+            let act = self
+                .get_actor(&Address::VERIFIED_REGISTRY_ACTOR, *ts.parent_state())
+                .map_err(|e| Error::State(e.to_string()))?
```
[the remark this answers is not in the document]
ruseinov: I don't understand what `wibni` means. 

It's a nice refactoring, I'll add this as an issue for another PR, because we have quite a bit of occurrences using `Error::State(blah.to_string())` pattern and I'd rather not have this change in this PR.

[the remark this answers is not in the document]
aatifsyed: would be nice if :)

aatifsyed (approved):

ruseinov:
> Have you tested if this also works with the new `--ws` flag?

I have not, should I? Not sure if it's at all relevant as long as the format is the same. If you didn't have to introduce any changes to other endpoints for said flag - I don't expect this one needing any special treatment. Otherwise let me know what needs to be done.

elmattic:
> > Have you tested if this also works with the new `--ws` flag?
> 
> I have not, should I?

No, it was just by curiosity. The test of this flag should be done on the CI anyway.

> Not sure if it's at all relevant as long as the format is the same. If you didn't have to introduce any changes to other endpoints for said flag - I don't expect this one needing any special treatment. Otherwise let me know what needs to be done.

FYI, you can test it in your case with: `cargo run --bin=forest-tool -- api compare --ws --filter StateVerifiedClientStatus --lotus /dns/127.0.0.1/tcp/1234/ws snapshot-latest`.

src/state_manager/mod.rs
```diff
@@ -1313,6 +1314,36 @@ where
         )
     }
 
+    pub fn verified_client_status(
+        self: &Arc<Self>,
+        addr: &Address,
+        ts: &Arc<Tipset>,
+    ) -> anyhow::Result<Option<DataCap>> {
+        let id = self.lookup_id(addr, ts)?.expect("actor not found");
```
LesnyRumcajs, line 1322: Shouldn't we rather propagate the error? We wouldn't like to crash the node in case in case of a massing actor, no?

ruseinov: Yeah, you are right. We should, I will just convert the `None` value to an error.

LesnyRumcajs (approved):

PR #3827, opened 2023-12-17
feat: implement `StateListMessages` rpc

qiweiii:
## Summary of changes

- add StateListMessages rpc method
- test against lotus
  ```sh
  # run lotus
  lotus daemon --import-snapshot {snapshot}

  # run forest
  forest --chain calibnet \
  --rpc true \
  --no-gc --encrypt-keystore false \
  --import-snapshot {snapshot}
 
  # compare api
  forest-tool api compare --forest /ip4/127.0.0.1/tcp/2345 --lotus /ip4/127.0.0.1/tcp/1234 {snapshot} --filter StateListMessages
  ```

| RPC Method                      | Forest | Lotus |
|---------------------------------|--------|-------|
| Filecoin.StateListMessages (59) | Valid  | Valid |

#3640

Closes

## Change checklist


lemmih:
I'll be updating our CI to work with external PRs.

qiweiii:
@lemmih is the ci now working for external pr?

src/rpc/state_api.rs
```diff
@@ -804,6 +804,60 @@ pub(in crate::rpc) async fn state_vm_circulating_supply_internal<
...
+/// Looks back and returns all messages with a matching to or from address, stopping at the given height.
+pub(in crate::rpc) async fn state_list_messages<DB: Blockstore + Send + Sync + 'static>(
+    data: Data<RPCState<DB>>,
+    Params(LotusJson((from_to, tsk, max_height))): Params<
+        LotusJson<(MessageFilter, TipsetKeys, i64)>,
+    >,
+) -> Result<LotusJson<Vec<Cid>>, JsonRpcError> {
+    let ts = data.chain_store.load_required_tipset(&tsk)?;
+
+    if from_to.from == None && from_to.to == None {
```
ruseinov, line 816: nit: this could be a method in `MessageFilter`, e.g. `fn is_empty`.

src/rpc/state_api.rs
```diff
@@ -804,6 +804,60 @@ pub(in crate::rpc) async fn state_vm_circulating_supply_internal<
...
+        return Err("must specify at least To or From in message filter".into());
+    } else if let Some(to) = from_to.to {
```
ruseinov, line 818: what happens if we have both `to` and `from` specified?

src/rpc/state_api.rs
```diff
@@ -804,6 +804,60 @@ pub(in crate::rpc) async fn state_vm_circulating_supply_internal<
...
+        data.state_manager
```
ruseinov, line 819: Is this call used only for validation? It clearly does not return anything, just propagates an error.

src/rpc/state_api.rs
ruseinov: if we need to verify both, the correct way of doing that would be getting rid of the `else if` blocks and just doing:

```if let Some(to) ..```

```if let Some(from)..```

File: src/tool/subcommands/api_cmd.rs [hunk 4]
```diff
@@ -555,6 +574,14 @@ fn snapshot_tests(store: &ManyCar, n_tipsets: usize) -> anyhow::Result<Vec<RpcTe
                         ApiInfo::state_search_msg_limited_req(msg.cid()?, 800),
                     ));
                     tests.push(RpcTest::basic(ApiInfo::mpool_get_nonce_req(msg.from())));
+                    tests.push(RpcTest::identity(ApiInfo::state_list_messages_req(
+                        MessageFilter {
+                            from: None,
```
ruseinov, line 579: Since you have decided to test all the options, you'd want to also test 
```
from: None,
to: None,
```

src/rpc/state_api.rs
qiweiii: Yes, it's only for validation

src/rpc/state_api.rs
qiweiii: I was following lotus logic since the results need to mathc lotus result:

https://github.com/filecoin-project/lotus/blob/9b4a3ba23dd4e8c8255754e48ab3f134ec8860ef/node/impl/full/state.go#L1175-L1176

but I agree your suggestion make more sense.

ruseinov: Since this is following Lotus logic, which looks buggy to me, please keep following it.
But please mention this in a comment and provide this link to justify.

ruseinov:
Please add the command you've been testing this with and it's output to the PR description.

qiweiii:
> Please add the command you've been testing this with and it's output to the PR description.

Updated

lemmih (approved):

PR #3846, opened 2024-01-04
Fix: RPC parallel test runs

sudo-shashank:
## Summary of changes

- Added `max_concurrent_requests` to limit max concurrent test runs which prevents overloading the rpc servers.
This issue was observed for `lotus` only and not for `forest`, when approx 1000+ RPC tests were executed simultaneously.

Closes

## Change checklist

### DIFF
--- a/src/tool/subcommands/api_cmd.rs
+++ b/src/tool/subcommands/api_cmd.rs
@@ -58,7 +58,7 @@ pub enum ApiCommands {
 }
 
 /// For more information about each flag, refer to the Forest documentation at:
-/// https://docs.forest.chainsafe.io/rustdoc/forest_filecoin/tool/subcommands/api_cmd/enum.ApiCommands.html
+/// <https://docs.forest.chainsafe.io/rustdoc/forest_filecoin/tool/subcommands/api_cmd/enum.ApiCommands.html>
 struct ApiTestFlags {
     filter: String,
     fail_fast: bool,
--- a/src/tool/subcommands/api_cmd.rs
+++ b/src/tool/subcommands/api_cmd.rs
@@ -57,6 +57,8 @@ pub enum ApiCommands {
     },
 }
 
+/// For more information about each flag, refer to the Forest documentation at:
+/// https://docs.forest.chainsafe.io/rustdoc/forest_filecoin/tool/subcommands/api_cmd/enum.ApiCommands.html
 struct ApiTestFlags {
     filter: String,
     fail_fast: bool,
--- a/src/tool/subcommands/api_cmd.rs
+++ b/src/tool/subcommands/api_cmd.rs
@@ -56,6 +56,14 @@ pub enum ApiCommands {
     },
 }
 
+struct ApiTestFlags {
+    filter: String,
+    fail_fast: bool,
+    n_tipsets: usize,
+    run_ignored: RunIgnored,
+    max_concurrent_requests: usize,
+}
+
 impl ApiCommands {
     pub async fn run(self) -> anyhow::Result<()> {
         match self {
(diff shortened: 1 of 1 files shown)


LesnyRumcajs:
> Added max_concurrent_requests to limit max concurrent test runs which prevents overloading the rpc servers

Could you please elaborate on this? Which conditions made the RPC servers overload, at what QPS? Was it Lotus or Forest? It's useful to know these limitations.

sudo-shashank:
> > Added max_concurrent_requests to limit max concurrent test runs which prevents overloading the rpc servers
> 
> Could you please elaborate on this? Which conditions made the RPC servers overload, at what QPS? Was it Lotus or Forest? It's useful to know these limitations.

updated PR description

LesnyRumcajs:
Thanks, this is interesting. I believe we should benchmark Forest and Lotus for the Top N queries, we might need it for hardware/replicas count recommendations.

src/tool/subcommands/api_cmd.rs
```diff
@@ -50,9 +51,20 @@ pub enum ApiCommands {
         #[arg(long, value_enum, default_value_t = RunIgnored::Default)]
         /// Behavior for tests marked as `ignored`.
         run_ignored: RunIgnored,
+        /// Maximum number of concurrent requests
+        #[arg(long, default_value = "8")]
+        max_concurrent_requests: usize,
     },
 }
 
+struct ApiTestFlags {
```
LesnyRumcajs, line 60: Let's document this. I have no clue what, for example, a `String` `filter` is and what does it expect.

elmattic:
Is this a problem of API calls for Lotus being rate-limited by default?

We could run Lotus without limitation if this is the case.

src/tool/subcommands/api_cmd.rs
```diff
@@ -632,41 +641,43 @@ async fn compare_apis(
...
 async fn run_tests(
     tests: Vec<RpcTest>,
     forest: &ApiInfo,
     lotus: &ApiInfo,
-    filter: String,
-    fail_fast: bool,
-    run_ignored: RunIgnored,
+    config: &ApiTestFlags,
 ) -> anyhow::Result<()> {
+    let semaphore = Arc::new(Semaphore::new(config.max_concurrent_requests));
```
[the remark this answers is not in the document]
elmattic: We have some semaphores in the `RaceBatch`  😉

src/tool/subcommands/api_cmd.rs
```diff
@@ -50,9 +51,20 @@ pub enum ApiCommands {
         #[arg(long, value_enum, default_value_t = RunIgnored::Default)]
         /// Behavior for tests marked as `ignored`.
         run_ignored: RunIgnored,
+        /// Maximum number of concurrent requests
+        #[arg(long, default_value = "8")]
+        max_concurrent_requests: usize,
```
elmattic, line 56: Supporting a way to turn off concurrent requests limitation could be nice (i.e., using the exceptional value 0). But for now I think this is ok.

src/tool/subcommands/api_cmd.rs
sudo-shashank: we already have a short description for these flags under `ApiCommands`
https://github.com/ChainSafe/forest/blob/ced0f7570ccc9e52b01db3ddf56dc0439288da91/src/tool/subcommands/api_cmd.rs#L41-L52

LesnyRumcajs: I wasn't aware of that. Perhaps adding a Rust doc link to the `ApiCommands` struct would help future readers. What do you think?

LesnyRumcajs (approved):

src/tool/subcommands/api_cmd.rs
sudo-shashank: yes, i'll add

elmattic:
> This issue was observed for lotus only and not for forest

Looking at trace logs, I can see lots of errors for Forest as well if I use "unlimited" mode (i.e. `--max-concurrent-requests 10000`):

`error sending request for URL (http://127.0.0.1:2345/rpc/v0): connection error: Connection reset by peer (os error 54)`

`2345` is the port used by Forest.

Btw, I fixed some because I had to low FDs (so using `ulimit -n 8192`).

elmattic (approved):

LesnyRumcajs (approved):

Issue #2610, opened 2023-02-27
Switch to official `prometheus-client` crate


hanabi1224:
**Issue summary**

[`prometheus-client`](https://crates.io/crates/prometheus-client) has been used by `libp2p-metrics` and it's maintained by the official, we should consider migrating from the [`prometheus`](https://crates.io/crates/prometheus) crate.

**Other information and links**

https://github.com/tikv/rust-prometheus/issues/392#issuecomment-968072946

eserilev:
hi, i'd like to pick this one up

LesnyRumcajs:
@eserilev Sure, feel free to check this out.

The goal would be to completely eliminate the `prometheus` dependency (at least directly). An important check that it works would be running Forest for a moment and confirming that all the metrics that there were before are still present.

To get those, start `forest`, answer all the prompts and let it run for a few minutes. Then, in another terminal, grab the metrics.
```
# start forest, answer all the prompts and let it run for a few minutes
forest --chain calibnet
# grab the metrics
wget localhost:6116/metrics
```

hxalid:
👋  is this issue being worked on or you still need someone to pick this up?

lemmih:
> 👋 is this issue being worked on or you still need someone to pick this up?

You can pick it up.

PR #3850, opened 2024-01-05
chore: migrate from prometheus to prometheus_client (part2)

hanabi1224:
## Summary of changes

- Migrate all forest metrics to `prometheus_client`
- Use `kubert-prometheus-process` as the collector for process metrics
- Remove `prometheus` from `Cargo.toml`

Closes https://github.com/ChainSafe/forest/issues/2610

## Change checklist


src/metrics/mod.rs
```diff
@@ -52,11 +42,17 @@ pub async fn init_prometheus<DB>(
 where
     DB: DBStatistics + Send + Sync + 'static,
 {
-    let registry = prometheus::default_registry();
+    // Add the process collector to the registry
+    if let Err(err) = kubert_prometheus_process::register(
+        DEFAULT_REGISTRY.write().sub_registry_with_prefix("process"),
```
ruseinov, line 47: Perhaps a better idea is to expose `DEFAULT_REGISTRY` through a helper method. That way if we ever want to move/refactor the way it's used - we could keep the method, but change the underlying implementation/variable.

hanabi1224: Done

LesnyRumcajs (approved):

Issue #3863, opened 2024-01-11
GC fixes and improvements.


ruseinov:
**Issue summary**

This issue is a placeholder for a collection of things that needs to be checked and improved in regards to GC.

1. Seems to be stuck on graph traversal, both ordered (single-threaded) and unordered. This might have something to do with running GC in the background while running a node. CPU spikes at times hinting at the fact that it works, but it takes days and never gets to the cleanup step.
2. Needs to be properly profiled on a running node. Could be done locally, the important thing here is to see what is happening when this runs in the background. 

**Other information and links**

LesnyRumcajs:
Being an important part of a long-running node, we should ensure it is properly monitored. Our own infrastructure node should, therefore, employ a check (optimally, based on a metric) or a set of checks, e.g., at least 1 garbage collection performed in 2 days (to be safe), at least N records cleaned or something along those lines.

LesnyRumcajs:
This needs to be verified that it works correctly and fixed if it doesn't, or closed if it does. https://github.com/ChainSafe/forest/issues/3645

ruseinov:
> This needs to be verified that it works correctly and fixed if it doesn't, or closed if it does. #3645

This issue is irrelevant to the new GC, there also isn't a reliable nor easy way to verify it as the GC process takes a long time between various steps and ultimately removes unreachable data that is older than `2*chain_finality`. 

There is a different issue though that has to do with the fact that the GC does not work properly at the moment. I would keep this issue open until we have a working `Mark'n'sweep` GC.

LesnyRumcajs:
So, should we close #3645?

ruseinov:
> So, should we close #3645?

Let's keep it open for now and close it when the new GC works.

ruseinov:
The current status is:
1. The GC works on calibnet on a 2* vCPU node with 16GB RAM.
2. The GC seems to break the node syncing on mainnet with the above setup:
```
2024-06-24T14:50:10.282997Z ERROR forest_filecoin::chain_sync::tipset_syncer: Sync messages check state failed for tipset range
[...]

PR #3876, opened 2024-01-15
[Fix] Introduce GC logging on failure, don't break the loop.

ruseinov:
## Summary of changes

- GC error log.
- No more breaking GC loop on error.

Work on https://github.com/ChainSafe/forest/issues/3863

## Change checklist


src/db/gc/mod.rs
```diff
@@ -163,7 +163,10 @@ impl<DB: Blockstore + GarbageCollectable + Sync + Send + 'static> MarkAndSweep<D
     /// using CAR-backed storage with a snapshot, for implementation simplicity.
```
hanabi1224, line 163: Is this note still valid, if not please update it.
```rust
/// NOTE: This currently does not take into account the fact that we might be starting the node
/// using CAR-backed storage with a snapshot, for implementation simplicity.
```

src/db/gc/mod.rs
```diff
@@ -163,7 +163,10 @@ impl<DB: Blockstore + GarbageCollectable + Sync + Send + 'static> MarkAndSweep<D
...
     pub async fn gc_loop(&mut self, interval: Duration) -> anyhow::Result<()> {
```
hanabi1224, line 164:
```suggestion
    pub async fn gc_loop(&mut self, interval: Duration) {
```

src/db/gc/mod.rs
```diff
@@ -185,6 +188,9 @@ impl<DB: Blockstore + GarbageCollectable + Sync + Send + 'static> MarkAndSweep<D
             // Make sure we don't run the GC too often.
             time::sleep(interval).await;
```
hanabi1224, line 189: Is there a way to not sleep for 10h here and make it testable with CI calibnet checks

src/db/gc/mod.rs
ruseinov: This not will always be valid, it states that it does not care about car-backed storage blocks.

src/db/gc/mod.rs
ruseinov: This won't work as the `services.spawn` expects a `Result`.

src/db/gc/mod.rs
ruseinov: We can make it configurable. This `sleep` was introduced at David's request to avoid running the GC too often.

src/db/gc/mod.rs
hanabi1224: In this case, I don't quite understand why `fn filter` works when it's takes the parity-db only. Shouldn't `fn unordered_stream_graph` take the entire DB (parity-db + car-db)?

ruseinov: It should work just fine. We only care about cleaning up unreachable nodes. Since all the data that's stored in ParityDB is in the future for any data that's in car-db - we don't really care about it. Unless I'm missing some weird edge-case. I'm assuming that snapshot we export are finalized, in which case there should not be any issues.

The one enhancement I think we should do based on your comments and my revisiting the code is specifying the first epoch that's stored in the database as opposed to the car file. Otherwise
```rust
        if depth > current_epoch {
            time::sleep(interval).await;
            return anyhow::Ok(());
        }

````
is basically useless. Or we can get rid of it altogether and just rely on initial sleep that we do later for simplicity.

File: src/db/gc/mod.rs (hunk 3 above)
Comment (hanabi1224) [reply src/db/gc/mod.rs:189]: Why is it necessary to wait `finality` epochs as apposed to keeping `2*finality` state roots like snapshot export?

File: src/db/gc/mod.rs (hunk 1 above)
Comment (hanabi1224) [reply src/db/gc/mod.rs:163]: > I'm assuming that snapshot we export are finalized

I'm not positive about this, as I understand this can only be true when we export snapshot from `latest - finality`, @lemmih could you weight in on this?

> just rely on initial sleep 

I suggest storing the last GC epoch in parity-db metadata comlumn, instead of waiting for 10h on every start. In an edge case, if we have a node that restarts (maybe for getting updates) every 9h then GC would never get a chance to run.

Comment (ruseinov) [reply src/db/gc/mod.rs:163]: @lemmih what do you think about the above? Seems reasonable to me. 
We could also initialize the last GC epoch as the heaviest tipset we see in car-backed storage when the database is empty.

File: src/db/gc/mod.rs (hunk 3 above)
Comment (hanabi1224) [reply src/db/gc/mod.rs:189]: > It is necessary to wait finality epochs

Why is it necessary? Isn't it sufficient to 
1. get all parity-db keys
2. get all keys that is required for exporting an snapshot
3. remove keys in 1 but not in 2

As I understand, mark and sweep GC is similar to below steps
1. exporting a snapshot
4. purge DB
5. import the exported snapshot.

Comment (ruseinov) [reply src/db/gc/mod.rs:189]: This is not exactly how it works. We should not be deleting potentially unfinalized data, because it might come up in future blocks, which is when we're going to have an inconsistent database. So after marking all the DB keys we should wait `finality` to make sure everything that's unreachable is finalized.

Comment (hanabi1224) [reply src/db/gc/mod.rs:189]: > We should not be deleting potentially unfinalized data

I believe snapshot export logic keeps 2*finality state roots, could you elaborate on how it could still be deleting potentially unfinalized data?

Comment (ruseinov) [reply src/db/gc/mod.rs:189]: Ah, yes, of course it's `2 * finality`. It does not change the logic though, the idea here is we do have to wait `interval` that in this case equals at least `2 * finality`, before we can filter and remove `marked` set.

We were talking about why we have to wait `2 * finality` or in our case `depth`, which is 
```rust
            let depth = cmp::max(
                chain_config.policy.chain_finality * 2,
                config.sync.recent_state_roots,
            );
```
to run the `filter` and `sweep` step.

We need to:
1. Make sure we can always export a valid snapshot.
2. Make sure that we never remove anything that could potentially become reachable as a result of forks.

hanabi1224: > Make sure that we never remove anything that could potentially become reachable as a result of forks.

When all keys that are required for snapshot export are unmarked (to keep), why would the marked keys become reachable later? Since the marked ones can only be before latest - 2 * finality

ruseinov: The original idea was that we don't want to cleanup unreachable stuff until it's `2 * finality` old. There's no way of telling that it's that old, except marking it for deletion and waiting for said amount of time, then checking if it's still unreachable.

Note that snapshot export algorithm on deals in reachable nodes, hence it's only useful for the `filter` step, to make sure we don't touch the reachable graph.

The case we are taking care of here is this one:
1. Imagine we have unreachable nodes, we mark them for deletion.
2. While the GC is running some of those nodes become reachable.
3. We end up with inconsistent DB state if we remove any of the nodes that are now reachable.

hanabi1224: > There's no way of telling that it's that old, except marking it for deletion and waiting for said amount of time

I still don't get why the steps I describe above would not mark stuff that is 2 * finality old.  step 1 returns all keys `A`, step 2 returns all keys that are younger than 2 * finality, `B`, (A - B) should be 2 * finality old and save to delete, no?

ruseinov: We have no idea how old the unreachable keys are.

ruseinov: Step 2 does not return unreachable keys at all.

hanabi1224: > Step 2 does not return unreachable keys at all.

Why does that mater? We have all unfinalized stuff in step 2, that said, those unreachable keys are finalized, should it not be safe to delete?

ruseinov: Again, we have literally no idea how old the unreachable keys are. And the only way of telling that at the moment is waiting `2 * finality` after they have been marked for deletion. We have no way of telling whether unreachable keys are finalized or not for exactly the same reason.

Exported snapshots don't have this problem, because we export them at tipset `X`, where the state is correct, ignoring all the tipsets that were syncing after the export has started. We don't have this luxury when cleaning up the database of a running node. 

Our problem here is that once we start deleting unreachable nodes that we have no idea how old they are - it could turn out that some of those keys become reachable as a result of further syncing the chain. And this is how we eventually get to the problem of trying to retrieve a key that does not exist.

hanabi1224: > We don't have this luxury when cleaning up the database of a running node.

Note that step 1 happens before step 2, data that is stored after step 2 is started will not be marked or deleted. So I'm not convinced `we have literally no idea how old the unreachable keys(A-B) are`

ruseinov: The GC is supposed to run in the background, that's the whole point of it, so let's not be assuming that we'll be pausing the tipset syncer. Note that you would have to pause the syncer before each GC run and for the duration of it, which obviously is going to take a long time.

If you are not convinced that we have no idea about the age of unreachable blocks without actually waiting an `interval` after those have been marked - you are very welcome to prove that it's indeed possible. I'd like you to explain how you would differentiate random unreachable nodes in terms of their age without being able to associate them with a certain epoch.

hanabi1224: I was about to prove that : ) if you agree that the stateroots in (A-B) is finalized and outdated when tipset syncer is paused, it won't become reachable again when tipset syncer is running because it's finalized.  (It might still contain the tipset headers and messages that are post GC epoch but that can be checked again before deletion)

> we have no idea about the age of unreachable blocks without actually waiting an interval after those have been marked

If this is true, I'd like to understand what's the benefit of switching from the previous GC approach that is testable and benchmarkable

ruseinov: The only way to prove that an unreachable node will not become reachable is explicitly waiting for a given interval of time as far as I know and that's the reason the algorithm was designed the way it is.

Some of the benefits are described in package-level comments. 
1. It's an external implementation as opposed to a DB wrapper. Which makes it way more straightforward to implement and decouples the database layout from GC needs.
2. Simple algorithm, simple abstractions. The implementation is trivial.
3. More space-efficient. 
4. Perfect for running in the background without downtime.

We can make it artificially testable by introducing more abstractions in order to be able to mock certain parts of the algorithm.

hanabi1224: > We can make it artificially testable by introducing more abstractions in order to be able to mock certain parts of the algorithm.

If I'm not wrong, this still does not allow effective manual e2e testing or benchmarking on calibnet or mainnet, sounds like a deal breaker to me.

Note that the old GC has solved the exact problem (running in the background without downtime) without waiting

ruseinov: It's not ideal, whether or not it's a deal breaker, because it's not manually testable - I'm not really sure. We can always benchmark every part of the three steps, there's no need to benchmark all three steps at the same time.

@lemmih could you please chip in with your thoughts on the matter? I kind of agree that this is hard to test and impossible to test manually, but it does offer a lot simpler codebase and algorithm.

lemmih: Waiting for chain finality is required for correctness. Due to possible forks, we may not delete data unless (1) it is unreachable from the current HEAD, and (2) it was written to the database more than `chain_finality` epochs ago. This _does_ make end-to-end testing more difficult: How do we quickly test the GC when it is not allowed to garbage collect anything in the first 7.5 hours?

Lowering (or removing) the wait time is not an option. The wait is required for correctness and cannot be any lower than 7.5 hours on mainnet or calibnet.

I posit that end-to-end testing is not important. We can verify with quickcheck that the GC behaves as expected (ie. deletes what it should delete, doesn't delete what it shouldn't, etc). And we can statically guarantee that the GC loop will never exit (except for critical failures, which should terminate Forest). Furthermore, the nodes we run in production will show whether there are any GC leaks over time.

In summary: Keep the 7.5 hour wait (this is not optional). Keep the quickcheck tests. Don't do end-to-end tests in the CI (as the GC doesn't do anything interesting in this timespan). Prove to the reader that the GC loop will run as long as Forest itself.

LesnyRumcajs: As noted in the [issue comment](https://github.com/ChainSafe/forest/issues/3863#issuecomment-1887148833), one way out of this is providing GC alarms in our long-running node. This is not ideal, given our current release process, but it's better than pure faith. :pray:

ruseinov: I'm currently figuring out why the GC gets stuck. It could be due to the fact that we bootstrap the node with a snapshot (car-backed storage) and the parallel graph walk somehow get stuck due to not having access to data all the way to genesis. Once that is debugged and resolved - we'll be able to monitor GC via checks and metrics indeed.

LesnyRumcajs:
@ruseinov Can this be closed? Or should it be merged?

ruseinov:
> @ruseinov Can this be closed? Or should it be merged?

yep, about time.

LesnyRumcajs (approved):

hanabi1224 (approved):

lemmih (approved):

PR #3895, opened 2024-01-23
fix(ci): cache fil proof parameter files

hanabi1224:
## Summary of changes

We've encountered quite a few CI timeouts due to slow downloads of fil proof parameter files(The slowness issue has been tracked by https://github.com/ChainSafe/forest/issues/3826). This PR tries to use build cache to avoid downloading these files to improve CI stability and reduce timecost.

Changes introduced in this pull request:

- cache proof param files in the jobs that need them.
- set `FIL_PROOFS_PARAMETER_CACHE` to `/var/tmp/filecoin-proof-parameters` to share the folder between forest and lotus
- ~~change devnet docker-compose config to mount local folder instead of volume for proof params~~
- remove `update-schedules.dff` and set `LOTUS_WATERMELON_HEIGHT=10` instead, see https://github.com/filecoin-project/lotus/blob/8f73f157933435f5020d7b8f23bee9e4ab71cb1c/build/params_2k.go#L108
- print lotus versions in `docker-compose.yml`
- upgrade lotus image tag to `2024-01-23-90ff9b2`(generated by https://github.com/ChainSafe/forest/actions/runs/7622411436)

To verify the change, search `paramfetch` in CI job logs, below lines should be gone.

```
# forest
2024-01-22T15:15:01.863850Z  INFO forest_filecoin::utils::proofs_api::paramfetch: Fetching param file "/home/runner/.local/share/forest/filecoin-proof-parameters/v28-proof-of-spacetime-fallback-merkletree-poseidon_hasher-8-8-0-0377ded656c6f524f1618760bffe4e0a1c51d5a70c4509eedae8a27555733edc.vk" from https://proofs.filecoin.io/ipfs/
2024-01-22T15:15:01.863873Z  INFO forest_filecoin::utils::proofs_api::paramfetch: Fetching param file "/home/runner/.local/share/forest/filecoin-proof-parameters/v28-proof-of-spacetime-fallback-merkletree-poseidon_hasher-8-8-0-559e581f022bb4e4ec6e719e563bf0e026ad6de42e56c18714a2c692b1b88d7e.vk" from https://proofs.filecoin.io/ipfs/
...

# lotus
+ lotus fetch-params 2048
...
2024-01-22T07:07:15.572Z	INFO	paramfetch	go-paramfetch@v0.0.4/paramfetch.go:262	GET https://proofs.filecoin.io/ipfs/QmYM6Hg7mjmvA3ZHTsqkss1fkdyDju5dDmLiBZGJ5pz9y9
  
[...]

PR #3896, opened 2024-01-23
feat: allow overriding devnet upgrade heights with env vars

hanabi1224:
## Summary of changes

Lotus allows overriding upgrade heights with `LOTUS_XX_HEIGHT` environment variables on 2k devnet, this PR tries to implement the same feature for forest devnet, using `FOREST_XX_HEIGHT` environment variables.  The feature is covered by `devnet checks` CI job.

Note: ~It's based on https://github.com/ChainSafe/forest/pull/3895 and should let https://github.com/ChainSafe/forest/pull/3895 go first~

- allow customing devnet upgrade heights with `FOREST_XX_HEIGHT` environment variables

Closes

## Change checklist


src/networks/devnet/mod.rs
```diff
@@ -6,132 +6,134 @@ use once_cell::sync::Lazy;
...
         HeightInfo {
             height: Height::Thunder,
-            epoch: -1,
+            epoch: get_upgrade_height("FOREST_THUNDER_HEIGHT", -1),
             bundle: None,
         },
         HeightInfo {
             height: Height::Watermelon,
-            epoch: 10,
+            epoch: get_upgrade_height("FOREST_WATERMELON_HEIGHT", 200),
```
LesnyRumcajs, line 136: Perhaps it'd be more idiomatic to have it as `Result` so the usage would be `get_upgrade_height("FOREST_WATERMELON_HEIGHT").unwrap_or(200)`. Not a big deal, though.

LesnyRumcajs (approved):

src/networks/devnet/mod.rs
hanabi1224: I feel `get_upgrade_height("FOREST_WATERMELON_HEIGHT").unwrap_or(200)` suggests `FOREST_WATERMELON_HEIGHT` is defined in normal cases, so I'd prefer keeping it this way, just like lotus.

```go
getUpgradeHeight := func(ev string, def abi.ChainEpoch) abi.ChainEpoch {
		hs, found := os.LookupEnv(ev)
		if found {
			h, err := strconv.Atoi(hs)
			if err != nil {
				log.Panicf("failed to parse %s env var", ev)
			}

			return abi.ChainEpoch(h)
		}

		return def
	}
```

LesnyRumcajs: Lotus has clunky Go error handling and C-like notion of optional values. How about an `Option<ChainEpoch>`?

hanabi1224: As for the function name `get_upgrade_height` I'd prefer `get_upgrade_height(&str, i64) -> i64`, or I can rename it to `get_upgrade_height_from_env_var(&str) -> Option<i64>`, what do you think?

LesnyRumcajs: IMO the 2nd one is more expressive and leaves less room for guessing. I'd omit the `_var` in the name.

hanabi1224: @LesnyRumcajs fixed

sudo-shashank (approved):

PR #3902, opened 2024-01-25
feat: generate fat docker image that contains proof parameters

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/3892

Changes introduced in this pull request:

- generate fat docker images that contain proof parameter files (choosing `-fat` suffix as it's shorter for typing)
- update CI workflows accordingly
- actor bundle inclusion will be done in a separate PR, together with readme updates

## Reference issue to close (if applicable)

Closes

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

Issue #3892, opened 2024-01-22
Provide a `full` Forest docker image


LesnyRumcajs:
**Issue summary**

Certain locations have issues when downloading artefacts required for Forest to run properly. These include actor bundles and proof parameters.

One option to make life easier for the users from such locations is to provide a Docker image with all those artefacts already included. Such image should be based on our regular one and be tagged with, e.g., `-fat` or `-full` or `-bundled`. We'd end up with images like `v0.16.4-bundled`.

**Other information and links**

PR #3912, opened 2024-01-29
fix(ci): build and push fat docker image before the slim one

hanabi1224:
## Summary of changes

As part of https://github.com/ChainSafe/forest/issues/3892 and a follow-up of https://github.com/ChainSafe/forest/pull/3902

The fat image is built on top of the slim one with extra steps fetching data from the internet. There's a chance that the slim image is successfully built and pushed but the fat image fails. After reversing the order, it should either both fail(fail fast) or both succeed as the build-slim-image step should have been cached when the build-fat-image step is successful.

- build and push fat docker image before the slim one

Closes

## Change checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -53,6 +53,7 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


LesnyRumcajs (approved):
It might be worth adding a comment, so in the future it's not refactored the other way around.

hanabi1224:
> It might be worth adding a comment, so in the future it's not refactored the other way around.

@LesnyRumcajs Done.

lemmih (approved):

PR #3943, opened 2024-02-06
Migrate Forest to `jsonrpsee` crate

elmattic:
## Summary of changes

- Migrate to `jsonrpsee` server

```console
$ ./forest-tool api compare --lotus /ip4/127.0.0.1/tcp/1234/http --forest /ip4/127.0.0.1/tcp/2345/http forest_snapshot_calibnet_2024-02-05_height_1326491.forest.car.zst

| RPC Method                                     | Forest              | Lotus               |
|------------------------------------------------|---------------------|---------------------|
| Filecoin.BeaconGetEntry                        | Valid               | Valid               |
| Filecoin.ChainGetBlock                         | Valid               | Valid               |
| Filecoin.ChainGetBlockMessages (61)            | Valid               | Valid               |
| Filecoin.ChainGetGenesis                       | Valid               | Valid               |
| Filecoin.ChainGetMessage (2)                   | Valid               | Valid               |
| Filecoin.ChainGetMessagesInTipset (20)         | Valid               | Valid               |
| Filecoin.ChainGetParentMessages (61)           | Valid               | Valid               |
| Filecoin.ChainGetParentReceipts (61)           | Valid               | Valid               |
| Filecoin.ChainGetTipSet                        | Valid               | Valid               |
| Filecoin.ChainGetTipSetByHeight                | Valid               | Valid               |
| Filecoin.ChainHasObj                           | Valid               | Valid               |
| Filecoin.ChainHead                             | Valid               | Valid               |
| Filecoin.ChainReadObj                          | Valid               | Valid               |
| Filecoin.EthAccounts                           | Valid               | Valid               |
| Filecoin.EthBlockNumber                        | Valid               | Valid               |
| Filecoin.EthChainId                            | Valid               | Valid               |
| Filecoin.EthGasPrice                           | Valid               | Valid               |
| Filecoin.EthGetBalance (4)                     | Valid               | Valid               |
| Filecoin.MinerGetBaseInfo (7)                  | InvalidResponse     | Valid               |
| Filecoin.MinerGetBaseInfo (54)                 | Valid               | Valid               |
| Filecoin.MpoolGetNonce (2)                     | Valid               | Valid               |
| Filecoin.MpoolPending                          | Valid               | Valid               |
| Filecoin.MsigGetAvailableBalance               | Valid               | Valid               |
| Filecoin.MsigGetPending                        | Valid               | Valid               |
| Filecoin.NetAddrsListen                        | Valid               | Valid               |
| Filecoin.NetPeers                              | Valid               | Valid               |
| Filecoin.Session                               | Valid               | Valid               |
| Filecoin.StartTime                             | Valid               | Valid               |
| Filecoin.StateAccountKey (4)                   | Valid               | Valid               |
| Filecoin.StateCall (7)                         | Valid               | Valid               |
| Filecoin.StateCirculatingSupply (20)           | Valid               | Valid               |
| Filecoin.StateGetActor                         | Valid               | Valid               |
| Filecoin.StateGetRandomnessFromBeacon          | Valid               | Valid               |
| Filecoin.StateGetRandomnessFromTickets         | Valid               | Valid               |
| Filecoin.StateListMessages (2)                 | InternalServerError | InternalServerError |
| Filecoin.StateListMessages (4)                 | Valid               | Valid               |
| Filecoin.StateListMiners                       | Valid               | Valid               |
| Filecoin.StateLookupID (4)                     | Valid               | Valid               |
| Filecoin.StateMinerActiveSectors (62)          | Valid               | Valid               |
| Filecoin.StateMinerDeadlines (61)              | Valid               | Valid               |
| Filecoin.StateMinerFaults (61)                 | Valid               | Valid               |
| Filecoin.StateMinerInfo (2)                    | InvalidResponse     | Valid               |
| Filecoin.StateMinerInfo (59)                   | Valid               | Valid               |
| Filecoin.StateMinerPower (61)                  | Valid               | Valid               |
| Filecoin.StateMinerProvingDeadline (61)        | Valid               | Valid               |
| Filecoin.StateMinerRecoveries (61)             | Valid               | Valid               |
| Filecoin.StateMinerSectorCount (61)            | Valid               | Valid               |
| Filecoin.StateNetworkName                      | Valid               | Valid               |
| Filecoin.StateNetworkVersion                   | Valid               | Valid               |
| Filecoin.StateReadState (2)                    | Valid               | Valid               |
| Filecoin.StateSearchMsg                        | InternalServerError | InternalServerError |
| Filecoin.StateSearchMsg                        | Valid               | InternalServerError |
| Filecoin.StateSearchMsgLimited                 | InternalServerError | Valid               |
| Filecoin.StateSearchMsgLimited                 | InvalidResponse     | Valid               |
| Filecoin.StateSectorGetInfo                    | Valid               | Valid               |
| Filecoin.StateVMCirculatingSupplyInternal (20) | Valid               | Valid               |
| Filecoin.StateVerifiedClientStatus (2)         | Valid               | Valid               |
| Filecoin.StateWaitMsg                          | Timeout             | Timeout             |
| Filecoin.StateWaitMsg                          | Valid               | Timeout             |
| Filecoin.Version                               | Valid               | Valid               |
| Filecoin.WalletBalance                         | Valid               | Valid               |
| Filecoin.WalletValidateAddress                 | Valid               | Valid               |
| Filecoin.WalletVerify                          | Valid               | Valid               |
```

Websocket transport flag should work as well:

```console
$ ./forest-tool api compare --ws --lotus /ip4/127.0.0.1/tcp/1234/ws --forest /ip4/127.0.0.1/tcp/2345/ws forest_snapshot_calibnet_2024-02-14_height_1352592.forest.car.zst
```

Closes

## Change checklist


aatifsyed:
Could you elaborate on the rationale? What does https://github.com/paritytech/jsonrpsee get us that https://github.com/paritytech/jsonrpc didn't?

elmattic:
> Could you elaborate on the rationale? What does https://github.com/paritytech/jsonrpsee get us that https://github.com/paritytech/jsonrpc didn't?

Sure. First, this is the crate we use today: https://crates.io/crates/jsonrpc-v2 (the one you mentioned is the old version of `jsonrpsee` from paritytech and is no longer maintained).
The primary rationale behind the move is that jsonrpsee supports WebSocket pubsub (following [eth specs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub)), and we can customize the library to handle our own pubsub specification.

I've tested it in this branch to support `Filecoin.ChainNotify` endpoint: https://github.com/ChainSafe/forest/pull/3865/

In the context of supporting the Ethereum JSON-RPC API, this crate should help us as well.

src/rpc/error.rs
```diff
@@ -0,0 +1,170 @@
...
+use crate::message_pool::Error as MessagePoolError;
+use crate::state_manager::Error as StateManagerError;
+
+use jsonrpsee::types::error::{ErrorObjectOwned, INTERNAL_ERROR_CODE};
+
+pub struct JsonRpcError {
+    error: ErrorObjectOwned,
+}
+
+impl From<anyhow::Error> for JsonRpcError {
```
aatifsyed, line 16: I don't think `INTERNAL_ERROR_CODE` is suitable for all of these - that's basically a HTTP 500

src/rpc/mod.rs
```diff
@@ -2,46 +2,133 @@
...
+    DB: Blockstore + Send + Sync + 'static,
+{
+    // `Arc` is needed because we will share the state between two modules
+    let state = Arc::new(state);
+    let keystore = state.keystore.clone();
+    let mut module = RpcModule::new(state.clone());
+
+    register_methods(
+        &mut module,
+        state.state_manager.chain_config().block_delay_secs as u64,
```
aatifsyed, line 75: nit: `u64::from(...)` so readers don't need to ask themselves if this will truncate

src/rpc/mod.rs
```diff
@@ -2,46 +2,133 @@
...
-    beacon_api::beacon_get_entry,
-    common_api::{session, shutdown, start_time, version},
-    rpc_http_handler::{rpc_http_handler, rpc_v0_http_handler},
-    rpc_ws_handler::{rpc_v0_ws_handler, rpc_ws_handler},
-    state_api::*,
-};
+const MAX_RESPONSE_BODY_SIZE: u32 = 16 * 1024 * 1024;
+
+#[derive(Clone)]
+struct PerConnection<RpcMiddleware, HttpMiddleware> {
```
aatifsyed, line 52: Great naming

src/rpc_api/data_types.rs
```diff
@@ -109,6 +106,27 @@ pub struct MarketDeal {
     pub state: DealState,
 }
 
+impl Clone for MarketDeal {
```
aatifsyed, line 109: Add an issue to fil-actor-interface, and a tracking issue in this repo to implement `Clone` for `DealProposal`

src/rpc/auth_api.rs
```diff
@@ -1,18 +1,24 @@
...
+use jsonrpsee::types::Params;
+
+use std::sync::Arc;
 
 /// RPC call to create a new JWT Token
 pub async fn auth_new<DB: Blockstore>(
-    data: Data<RPCState<DB>>,
-    Params(params): Params<AuthNewParams>,
+    params: Params<'_>,
+    data: Arc<Arc<RPCState<DB>>>,
```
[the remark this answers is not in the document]
elmattic: The first `Arc` is needed to share the state between two RPC modules.
https://github.com/ChainSafe/forest/blob/ce33815f8ffb9670a4f31bb327027b2ebbed3f3d/src/rpc/mod.rs#L69

We only have one module now, but to implement specific pubsub code, we will need to implement a dedicated `RpcModule`.
In the context of `async` methods, `jsonrpsee` creates a second `Arc` around the state.

src/rpc/auth_layer.rs
```diff
@@ -0,0 +1,108 @@
...
+        let keystore = self.keystore.clone();
+        let service = self.service.clone();
+
+        async move {
+            let auth_header = headers.get(AUTHORIZATION).cloned();
+            let res = check_permissions(keystore, auth_header, req.method_name()).await;
+
+            match res {
+                Ok(()) => service.call(req).await,
+                Err(code) => MethodResponse::error(Id::Null, ErrorObject::from(code)),
```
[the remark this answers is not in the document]
elmattic: Good catch!

src/rpc/mod.rs
elmattic: Yes, that's way better.

src/rpc/error.rs
elmattic: `-32603` is for an internal error no? According to the JSON-RPC spec here: 

https://www.jsonrpc.org/specification#error_object

ruseinov: If you have no way to differentiate between errors - internal error seems fine. Not sure what else can be done here. There's a `SERVER_ERROR`, but that's a range of codes. And an `INTERNAL_ERROR`. Since we don't really know what to do with each individual error - I guess the only thing left here is to group them all and treat as the same error code. I can't see any other error code suitable for this purpose.

ruseinov: Ah, I now understand Aatif's comment better. We have many different error types, but all of them are treated as `INTERNAL_ERROR`.

I'd say that perhaps `anyhow::Error` is actually suitable for this, but then again we have another bunch of types that we could at least differentiate. 

I guess we could use `Server error` instead with a bunch of custom-defined codes per error type in `From` definition.

src/rpc/auth_api.rs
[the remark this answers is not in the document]
elmattic: Yes, because all methods in a module need access to the state.

In our case, we will have to call `chain_api::chain_notify` to get the subscriber:

```
async fn chain_notify<DB: Blockstore>(
     data: Arc<RPCState<DB>>,
 ) -> Result<Subscriber<HeadChange>, JsonRpseeError> {
     let head_change = data.chain_store.publisher().subscribe();

     Ok(head_change)
 }
```

ruseinov:
Seems good apart from two nits already mentioned by @aatifsyed.

Perhaps the error codes could be improved, so that we at least have a way of knowing which error type was returned.
And the double `Arc` thing has to be either avoided or type aliased if unavoidable.

src/rpc/auth_api.rs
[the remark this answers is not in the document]
elmattic: Hum, there's only one `Arc` here.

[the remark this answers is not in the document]
elmattic: Ah I see why, ignore my comment.

elmattic:
> Perhaps the error codes could be improved, so that we at least have a way of knowing which error type was returned.

Yes, will do. Error codes could be much more precise.

And we don't always validate params; for example, in `eth_chain_id`; if we send incorrect params, Lotus answers with:
```
    JsonRpcError {
        code: 500,
        message: "{\"jsonrpc\":\"2.0\",\"id\":0,\"error\":{\"code\":-32602,\"message\":\"wrong param count (method 'Filecoin.EthChainId'): 2 != 0\"}}\n",
    },
```
While for Forest, it's valid. This can be addressed in a follow-up PR, though.

src/rpc/error.rs
aatifsyed: `INTERNAL_ERROR` is suitable for e.g handler panic - treat it like HTTP `500 Internal Server Error `.

Use other error codes (outside of the -32000..=-32099 range), here.

OTOH, if lotus always just always returns `INTERNAL_ERROR`, we might decide to copy that.
After this discussion, I'll leave this up to your judgement

src/rpc/auth_api.rs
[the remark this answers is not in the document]
aatifsyed: What's the resolution here? this is the only thing blocking an approval from me I think

src/rpc/error.rs
elmattic: I'm ok with improving error codes but in a subsequent PR. Let's finish this one to avoid dealing with more merge conflicts.

src/rpc/auth_api.rs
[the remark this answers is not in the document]
elmattic: I think the double `Arc<Arc<T>>` is ok for now (I've hidden the ugliness using the `Data` type alias). This should resolve at the `jsonrpee` crate level. Maybe with something like:

```
pub fn from_arc(ctx: Arc<Context>) -> Self {
  Self { ctx, methods: Default::default() }
}
```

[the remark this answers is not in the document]
aatifsyed: I'm grumpy about this, but if there's truly no other way then fine

aatifsyed (approved):

src/rpc/auth_api.rs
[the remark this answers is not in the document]
elmattic: The `from_arc` works and I can remove one Arc level: `pub type Data<T> = Arc<T>;`

This will require change in upstream repository though.

src/rpc/chain_api.rs
```diff
@@ -1,50 +1,62 @@
...
     data: Data<RPCState<DB>>,
-    Params(LotusJson((block_cid,))): Params<LotusJson<(Cid,)>>,
 ) -> Result<LotusJson<Vec<ApiMessage>>, JsonRpcError> {
+    let LotusJson((block_cid,)): LotusJson<(Cid,)> = params.parse()?;
+
     let store = data.state_manager.blockstore();
     let block_header: CachingBlockHeader = store
         .get_cbor(&block_cid)?
-        .ok_or_else(|| format!("can't find block header with cid {block_cid}"))?;
+        .context(format!("can't find block header with cid {block_cid}"))?;
```
lemmih, line 59:
```suggestion
        .with_context(|| format!("can't find block header with cid {block_cid}"))?;
```

src/rpc/chain_api.rs
```diff
@@ -1,50 +1,62 @@
...
-    Params(LotusJson((msg_cid,))): Params<LotusJson<(Cid,)>>,
 ) -> Result<LotusJson<Message>, JsonRpcError> {
+    let LotusJson((msg_cid,)): LotusJson<(Cid,)> = params.parse()?;
+
     let chain_message: ChainMessage = data
         .state_manager
         .blockstore()
         .get_cbor(&msg_cid)?
-        .ok_or_else(|| format!("can't find message with cid {msg_cid}"))?;
+        .context(format!("can't find message with cid {msg_cid}"))?;
```
lemmih, line 43:
```suggestion
        .with_context(|| format!("can't find message with cid {msg_cid}"))?;
```

src/rpc/chain_api.rs
```diff
@@ -55,13 +67,15 @@ pub async fn chain_get_parent_message<DB: Blockstore>(
...
     data: Data<RPCState<DB>>,
-    Params(LotusJson((block_cid,))): Params<LotusJson<(Cid,)>>,
 ) -> Result<LotusJson<Vec<ApiReceipt>>, JsonRpcError> {
+    let LotusJson((block_cid,)): LotusJson<(Cid,)> = params.parse()?;
+
     let store = data.state_manager.blockstore();
     let block_header: CachingBlockHeader = store
         .get_cbor(&block_cid)?
-        .ok_or_else(|| format!("can't find block header with cid {block_cid}"))?;
+        .context(format!("can't find block header with cid {block_cid}"))?;
```
lemmih, line 78:
```suggestion
        .with_context(|| format!("can't find block header with cid {block_cid}"))?;
```

lemmih (approved):

src/rpc/node_api.rs
```diff
@@ -1,12 +1,16 @@
 // Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 #![allow(clippy::unused_async)]
+#![allow(clippy::redundant_allocation)]
```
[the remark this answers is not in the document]
hanabi1224: I tried commenting out some of them locally and did not get clippy errors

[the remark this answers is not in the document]
elmattic: Nice, thanks! I will remove them in that case.

hanabi1224 (approved):

Issue #3856, opened 2024-01-08
[Forest] NV22 state migration


LesnyRumcajs:
**Issue summary**
Implement NV22 migration. ~~It seems it will only need code IDs and system actor changes (similar to some recent calibnet patches, but all actors, not only the Miner actor).~~

I was overly optimistic. There may be more changes; see @lemmih comment below.

**Other information and links**

LesnyRumcajs:
Accepted FIPs for NV22:
- [FIP-0063: Switching to new Drand mainnet network](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0063.md)
  * no changes to the state
- [FIP0074: Remove cron-based automatic deal settlement](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0074.md)
  * no changes to the state
- [FIP0076: Direct Data Onboarding](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0076.md)
  * non-trivial migration required, as outlined in the [migration section](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0076.md#migration) of the FIP. It might be as heavy as NV21 migration, given that it requires reading all the metadata from the miner actor to migrate the market actor.

For completeness, new built-in actor codes will also need to be deployed, as usual.

PR #3955, opened 2024-02-14
NV22 dragon state migration

LesnyRumcajs:
## Summary of changes

- support for NV22 + state migrations - calibration upgrade date defined, mainnet TBD with the final bundle update,
- tested:
  - :heavy_check_mark:  mainnet/calibnet produces the same state as Lotus v1.26.0-rc2,
  - :heavy_check_mark:  butterflynet interop tests with Lotus
  - :heavy_check_mark:  devnet checks in CI updated
  - :warning: mainnet migration takes quite a while - 150s on my machine, 500s on a droplet, requiring 64 GB of RAM.

Closes https://github.com/ChainSafe/forest/issues/3856

## Change checklist


ruseinov (approved):

src/state_migration/common/state_migration.rs
```diff
@@ -78,18 +80,29 @@ impl<BS: Blockstore + Send + Sync> StateMigration<BS> {
         }
 
         let cache = MigrationCache::new(NonZeroUsize::new(10_000).expect("infallible"));
+        let num_threads = std::env::var("FOREST_STATE_MIGRATION_THREADS")
```
ruseinov, line 83: nit: would validate that to avoid footgun.

src/state_migration/common/state_migration.rs
```diff
@@ -78,18 +80,29 @@ impl<BS: Blockstore + Send + Sync> StateMigration<BS> {
...
+
         let pool = rayon::ThreadPoolBuilder::new()
             .thread_name(|id| format!("state migration thread: {id}"))
-            .num_threads(3) // minimum needed, more doesn't increase performance in any way
+            .num_threads(num_threads)
             .build()?;
 
-        let (state_tx, state_rx) = crossbeam_channel::bounded(1);
-        let (job_tx, job_rx) = crossbeam_channel::bounded(1);
+        let (state_tx, state_rx) = crossbeam_channel::bounded(30);
```
ruseinov, line 95: why crossbeam if we're using flume elsewhere?

src/state_migration/common/state_migration.rs
```diff
@@ -78,18 +80,29 @@ impl<BS: Blockstore + Send + Sync> StateMigration<BS> {
...
+        let (job_tx, job_rx) = crossbeam_channel::bounded(30);
```
ruseinov, line 96: nit: why are the channels bound specifically to `30`?

LesnyRumcajs: Magic number, same as cache set to 10k. Setting it to higher value, just like with cache sizes, may increase the memory footprint.

src/state_migration/common/state_migration.rs
LesnyRumcajs: I forced it to at least 3.

lemmih (approved):

src/state_migration/common/state_migration.rs
ruseinov: as long as it works. I would probably make it a constant with a little comment instead, but not critical.

PR #3934, opened 2024-02-06
fix: refactor RPC implementations to use ApiTipsetKey(Option<TipsetKey>) to not assume TipsetKey to be empty

hanabi1224:
## Summary of changes

Part of https://github.com/ChainSafe/forest/issues/3861

`TipsetKey` should not be empty and can be refactored using `NonEmpty` collections. However, this is blocked by the RPC implementations that assume `TipsetKey` to be empty and fall back to the heaviest tipset. This PR tries to refactor RPC implementations to use `ApiTipsetKey(Option<TipsetKey>)` instead. The actual refactoring to the `TipsetKey` struct will come in a subsequent PR.

Changes introduced in this pull request:

- refactor RPC implementations to use `ApiTipsetKey(Option<TipsetKey>)`
- rename `ChainStore::load_required_tipset` to `ChainStore::load_required_tipset_with_fallback`
- fix non-RPC code that misuses the `_with_fallback` version of `load_*_tipset`

## Reference issue to close (if applicable)

Closes

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

Issue #3861, opened 2024-01-10
Add `#[deny(clippy::indexing_slicing)]`


aatifsyed:
@LesnyRumcajs had to do some manual auditing of our indexing code, one of his takeaways was to avoid using unchecked indexing like `my_vec[0]`

Clippy can do those checks for us, but before we add `#[deny(clippy::indexing_slicing)]`, and the equivalent `#[deny(clippy::get_unwrap)]`, there are a few cases that need to be refactored, removed or annotated with `#[allow(..)]`:
```console
$ cargo clippy -- --no-deps --deny clippy::indexing_slicing --deny clippy::get_unwrap
...
error: could not compile `forest-filecoin` (lib) due to 79 previous errors
```

This should definitely happen _after_ #3858 is merged, and should maybe be preceded by other introductions of `NonEmpty`

PR #3964, opened 2024-02-19
fix: refactor TipsetKey using NonEmpty collection

hanabi1224:
## Summary of changes

Part of https://github.com/ChainSafe/forest/issues/3861 and a follow-up of https://github.com/ChainSafe/forest/pull/3934

- refactor TipsetKey using NonEmpty collection

Closes

## Change checklist


src/cid_collections/small_cid_vec.rs
```diff
@@ -11,31 +12,31 @@ use crate::blocks::TipsetKey;
...
+    pub fn into_cids(self) -> NonEmpty<Cid> {
+        self.0.map(From::from)
+    }
+}
+
+impl Default for SmallCidNonEmptyVec {
+    fn default() -> Self {
+        Self(nonempty![Cid::default().into()])
     }
 }
```
lemmih, line 41: This does not seem right. Can you explain it to me?

lemmih (changes requested):
I'm a bit confused. I thought `TipsetKey` could be empty. Can you verify that the genesis block has a non-empty parent tipset key?

src/cid_collections/small_cid_vec.rs
ruseinov: Yeah, I guess the idea here is that `nonempty!` is not actually empty. So it should not have a `Default` trait implemented, otherwise what's the point of it?

hanabi1224: Dropped `Default` derives for non test targets

hanabi1224:
> I'm a bit confused. I thought `TipsetKey` could be empty. Can you verify that the genesis block has a non-empty parent tipset key?

@lemmih As discussed, this is guaranteed on the implementation side, we might want to propose a small update to the spec later.
(https://filecoinproject.slack.com/archives/C05P37R9KQD/p1709050261991189?thread_ts=1708351510.708049&cid=C05P37R9KQD)

src/blocks/chain4u.rs
```diff
@@ -16,6 +16,7 @@ use crate::{
         sector::PoStProof,
     },
 };
+use chain4u::header::GENESIS_BLOCK_PARENTS;
```
ruseinov, line 19: nit: guess this slipped through the cracks in the original PR, while I like the name, module name with digits looks kinda ugly. But concise :D

ruseinov (approved):
looks good

src/ipld/util.rs
```diff
@@ -316,7 +316,7 @@ impl<DB: Blockstore, T: Iterator<Item = Tipset> + Unpin> Stream for ChainStream<
 
                         if block.epoch == 0 {
                             // The genesis block has some kind of dummy parent that needs to be emitted.
-                            for p in block.parents.cids.clone() {
+                            for p in block.parents.to_cids() {
```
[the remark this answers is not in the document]
hanabi1224: https://github.com/ChainSafe/forest/issues/4024

lemmih (approved):

PR #3978, opened 2024-02-20
Add custom module for fil pubsub

elmattic:
## Summary of changes

- Add partial support for `Filecoin.ChainNotify` method

Note: To fully support the endpoint, the type `"revert"` needs to be implemented (we currently support `"current"` and `"apply"`).

Closes

## Change checklist


src/rpc/chain_api.rs
```diff
@@ -445,6 +447,33 @@ pub(crate) async fn chain_get_min_base_fee<DB: Blockstore>(
...
+pub(crate) fn chain_notify<DB: Blockstore>(
+    _params: Params<'_>,
+    data: Arc<RPCState<DB>>,
+) -> Subscriber<ApiHeadChange> {
+    let (sender, receiver) = broadcast::channel(100);
+
+    // As soon as the channel is created, send the current tipset
+    let current = data.chain_store.heaviest_tipset();
+    let (change, headers) = ("current".into(), current.block_headers().clone().into());
+    let _ = sender.send(ApiHeadChange { change, headers });
```
lemmih, line 459: Let's assert that this call doesn't fail.

src/rpc/channel.rs
```diff
@@ -0,0 +1,446 @@
...
+
+use jsonrpsee::core::server::helpers::{MethodResponse, MethodResponseResult, MethodSink};
+use jsonrpsee::server::{
+    IntoSubscriptionCloseResponse, MethodCallback, Methods, RegisterMethodError,
+    SubscriptionMessage, SubscriptionMessageInner,
+};
+use jsonrpsee::types::{error::ErrorCode, ErrorObjectOwned, Id, Params, ResponsePayload};
+
+use parking_lot::Mutex;
+use rustc_hash::FxHashMap;
```
lemmih, line 71: Let's use a normal hash map.

src/rpc/chain_api.rs
elmattic: Should the function return a `Result<Subscriber<ApiHeadChange>, JsonRpcError>`?
Or do you mean we should use a static assert?

aatifsyed (approved):

hanabi1224 (approved):

PR #3999, opened 2024-02-27
fix(p2p): periodically re-dial to disconnected bootstrap nodes

hanabi1224:
## Summary of changes

This PR allows a forest node to re-dial to disconnected bootstrap nodes when discovery is disabled or not functioning in some rare conditions

- periodically re-dial to disconnected bootstrap nodes

Closes

## Change checklist

### DIFF
--- a/src/libp2p/service.rs
+++ b/src/libp2p/service.rs
@@ -173,6 +173,7 @@ pub enum NetRPCMethods {
 /// The `Libp2pService` listens to events from the libp2p swarm.
 pub struct Libp2pService<DB> {
     swarm: Swarm<ForestBehaviour>,
+    bootstrap_peers: HashMap<PeerId, Multiaddr>,
     cs: Arc<ChainStore<DB>>,
     peer_manager: Arc<PeerManager>,
     network_receiver_in: flume::Receiver<NetworkMessage>,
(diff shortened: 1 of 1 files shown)


LesnyRumcajs (commented):
Under expected conditions, there may be no point in keeping connection to the bootstrap nodes - it should be only required for the initial connection.

How about dialing the bootstrap nodes only if the node is not connected to any peers at all?

hanabi1224:
> How about dialing the bootstrap nodes only if the node is not connected to any peers at all?

@LesnyRumcajs There's an edge case that, when all the connected nodes are stateless, a stateful node still cannot sync

LesnyRumcajs:
> > How about dialing the bootstrap nodes only if the node is not connected to any peers at all?
> 
> @LesnyRumcajs There's an edge case that, when all the connected nodes are stateless(e.g. crawler) or out of sync by thousands of epochs, a stateful node still cannot sync

Aren't stateless nodes able to point to relevant peers? To my understanding, please correct me if I'm wrong, the bootstrap nodes (Forest ones) are stateless anyway.

hanabi1224:
> Aren't stateless nodes able to point to relevant peers?

@LesnyRumcajs They are, but if discovery is disabled and p2p port is not publicly exported, connections to other nodes still cannot be established

LesnyRumcajs (approved):
I'm convinced.

lemmih (approved):

PR #4027, opened 2024-03-06
fix(p2p): validate chain exchange responses & bug fix

hanabi1224:
## Summary of changes

- Fix a bug in assembling chain exchange messages into full tipets when the response length is different from the request length
- Add some basic validation to `handle_chain_exchange_request` when request type is `MESSAGE`

Closes

## Change checklist

### DIFF
--- a/src/chain_sync/tipset_syncer.rs
+++ b/src/chain_sync/tipset_syncer.rs
@@ -998,10 +998,11 @@ async fn fetch_batch<DB: Blockstore>(
             .map_err(TipsetRangeSyncerError::NetworkMessageQueryFailed)?;
 
         // inflate our tipsets with the messages from the wire format
+        // Note: compacted_messages.len() can be not equal to batch.len()
         compacted_messages
             .into_iter()
+            .zip(batch.iter().rev())
             .rev()
-            .zip(batch.iter())
             .map(|(messages, tipset)| {
                 let bundle = TipsetBundle {
                     blocks: tipset.block_headers().iter().cloned().collect_vec(),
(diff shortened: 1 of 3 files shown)


src/chain_sync/tipset_syncer.rs
```diff
@@ -771,7 +772,7 @@ async fn sync_tipset_range<DB: Blockstore + Sync + Send + 'static>(
         &bad_block_cache,
         parent_tipsets,
         &genesis,
-        InvalidBlockStrategy::Strict,
+        InvalidBlockStrategy::Forgiving,
```
lemmih, line 775: What are the consequences of moving from `Strict` to `Forgiving`?

hanabi1224: Below is the only usage of `InvalidBlockStrategy::Strict` I found. It uses a bad block cache to bypass validating known bad blocks, but I keep getting `chain contained [CID]` during sync in Strict mode. To my understanding, using `forgiving` mdoe is disabling a perf trick instead of  changing the logic
```rust
// Only do bad block accounting if the function was called with
                // `is_strict` = true
                if let InvalidBlockStrategy::Strict = invalid_block_strategy {
                    match &why {
                        TipsetRangeSyncerError::TimeTravellingBlock(_, _)
                        | TipsetRangeSyncerError::TipsetParentNotFound(_) => (),
                        why => {
                            bad_block_cache.put(cid, why.to_string());
                        }
                    }
                }
```

lemmih (approved):

LesnyRumcajs (approved):

lemmih (approved):

PR #4029, opened 2024-03-06
private key from keypair tool

LesnyRumcajs:
## Summary of changes

- a simple tool to get the private key from the keypair so that it can be used in Lotus and vice-versa. Theoretically, we _could_ make the import/export more user-friendly, but then we'd be in the field of compatibility with Lotus internal structure, which is not guaranteed to be the same across versions.

Closes

## Change checklist


hanabi1224 (approved):

samuelarogbonlo (approved):
Nice

hanabi1224 (approved):

Issue #4032, opened 2024-03-06
rpc methods have OpenRPC definitions


aatifsyed:
Migrate chain_api etc to generate `OpenRPC` docs

aatifsyed:
Code snippets file for VSCode:

```jsonc
{
    "rpc method": {
        "prefix": "impl RpcMethod",
        "body": [
            "pub enum ${1:MethodName} {}",
            "impl RpcMethod<${2:ARITY}> for $1 {",
            "const NAME: &'static str = \"Filecoin.$1\";",
            "const PARAM_NAMES: [&'static str; $2] = [$8];",
            "const API_VERSION: ApiVersion = ApiVersion::${3:V0};",
            "type Params = ($5,);",
            "type Ok = $6;",
            "async fn handle(ctx: Ctx<impl Blockstore $9>, ($7,): Self::Params) -> Result<Self::Ok, JsonRpcError> {",
            "$4", // paste handler here
            "}", // fn handle
            "}", // impl RpcMethod
        ]
    },
    "imports": {
        "prefix": "use",
        "body": [
            "use crate::rpc::{reflect::SelfDescribingRpcModule, ApiVersion, Ctx, RPCState, RpcMethod, RpcMethodExt as _, JsonRpcError};",
            "use schemars::JsonSchema;"
        ]
    },
    "register_all": {
        "prefix": "fn register_all",
        "body": [
            "pub fn register_all(module: &mut SelfDescribingRpcModule<RPCState<impl Blockstore + Send + Sync + 'static>>) {",
            "$0",
            "}",
        ]
    }
}
```

PR #4033, opened 2024-03-06
feat: initial implementation of OpenRPC definitions for methods

aatifsyed:
## Summary of changes
- `crate::rpc::reflect::SelfDescribingRpcModule` creates OpenRPC definitions for methods
- `chain_get_path` uses that struct
- cleanup error types
- add TODOs for various issues

## Follow up
#4032

## Change checklist


LesnyRumcajs:
Let's have an issue with the context for this PR. What problem does it solve? Why is it needed? What decisions needed to be made?

src/rpc/error.rs
```diff
@@ -1,170 +1,81 @@
...
-
-impl From<std::io::Error> for JsonRpcError {
-    fn from(e: std::io::Error) -> Self {
-        Self {
-            error: ErrorObjectOwned::owned::<()>(INTERNAL_ERROR_CODE, e.to_string(), None),
-        }
-    }
+// TODO(forest): https://github.com/ChainSafe/forest/issues/3965
+//               Just mapping everything to an internal error is not appropriate
+from2internal! {
```
aatifsyed, line 52: @elmattic I used a macro here to clean up repetitive code

src/rpc/reflect/openrpc_types.rs
```diff
@@ -0,0 +1,219 @@
...
+                .collect::<Vec<_>>();
+
+            if !late_mandatory_params.is_empty() {
+                return Err(ParamListError(
+                    String::from("Mandatory parameters may not follow optional parameters,")
+                        + &format!(
+                            "but the optional parameter {} is followed by [{}]",
+                            first_opt_param.name,
+                            late_mandatory_params.join(", ")
+                        ),
```
LesnyRumcajs, line 162: Why do you concat here and not use a single `format!`?

LesnyRumcajs (commented):
1. In general, should the `reflect` code live in the Forest repository? Could it potentially live elsewhere and be published to crates.io as something others could reuse?
2. As noted elsewhere, panics are rough. Can they be avoided so that the error can be handled by the caller?

src/rpc/reflect/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+        where
+            F: Clone + Send + Sync + 'static + Fn(Arc<Ctx>, $($arg),*) -> Fut,
+            //                                    ^^^ Receives an Arc rather than a reference lest
+            //                                        we get the dreaded "one type is more general than the other"
+            Ctx: Send + Sync + 'static,
+            $(
+                $arg: for<'de> Deserialize<'de> + Clone,
+            )*
+            Fut: Future<Output = Result<R, JsonRpcError>> + Send + Sync + 'static,
+            R: Serialize,
```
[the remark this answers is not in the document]
aatifsyed: but... it's beautiful :)

src/rpc/reflect/openrpc_types.rs
aatifsyed: As a single call it is a very long line, so this makes it easier to read

src/rpc/reflect/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+            calling_convention,
+            methods: vec![],
+        }
+    }
+    /// Even if the calling convention is [`ParamStructure::ByPosition`], OpenRPC
+    /// requires each argument to have a unique name.
+    ///
+    /// # Panics
+    /// - if param names aren't unique
+    /// - if optional parameters follow mandatory params
```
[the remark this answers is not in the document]
aatifsyed: In general I prefer to:
- return `Result` for recoverable or incidental errors
- `panic` if the code is written incorrectly

This is a case of the latter - someone has incorrectly defined a method - it's not a user error, and the only way to fix it is to rewrite the code.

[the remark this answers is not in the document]
LesnyRumcajs: Rewriting the code when something is wrong is idealistic. The code we ship should also be a bit defensive - if there's a bug in a single, e.g., RPC call, it shouldn't crash the entire server. Yes, the application has a bug, but it should not lead to potential DoS.

src/rpc/reflect/openrpc_types.rs
LesnyRumcajs: What's wrong with breaking the line? This also avoids potential operations which the compiler may not optimize away.
```
format!("Mandatory parameters may not follow optional parameters,\
        but the optional parameter {} is followed by [{}]",
        first_opt_param.name,
        late_mandatory_params.join(", ")
)
```

aatifsyed: much better idea :)

src/rpc/reflect/mod.rs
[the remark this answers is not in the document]
aatifsyed: added snapshot tests as discussed :)

LesnyRumcajs (approved):
Approving. That said, I have my doubts regarding the design decision of having panics here. I hope it won't bite us in the future.

Cargo.toml
```diff
@@ -216,6 +217,7 @@ derive-quickcheck-arbitrary = "0.1.1"
 fvm3 = { package = "fvm", default-features = false, version = "~3.8", features = ["arb"] }
 fvm_shared3 = { package = "fvm_shared", version = "~3.6", default-features = false, features = ["arb"] }
 http-range-header = "0.4.0"
+insta = { version = "1.36.1", features = ["yaml"] }
 libp2p = { version = "0.53", features = ['tcp', 'noise', 'yamux', 'request-response', 'tokio'] }
```
hanabi1224, line 221: nit: `libp2p` could be cleaned up from dev-deps

hanabi1224: https://github.com/ChainSafe/forest/pull/4050

hanabi1224 (approved):

src/rpc/reflect/util.rs
```diff
@@ -0,0 +1,57 @@
+// Copyright 2019-2024 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+use std::fmt::Display;
+
+use serde::{de::Visitor, forward_to_deserialize_any, Deserialize, Deserializer};
+
+/// "Introspection" by tracing a [`Deserialize`] to see if it's [`Option`]-like.
+///
+/// This allows us to smoothly operate between rust functions that take an optional
```
ruseinov, line 9: nit: I'm unsure that `operate between` is a thing, if I'm understanding the intention correctly `switch between` would be one way to put it.

src/rpc/reflect/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+           + 'static
+           + Fn(jsonrpsee::types::Params<'static>, Arc<Ctx>) -> Self::Future;
+}
+
+/// Return type-specific information for a function signature so that a [`ContentDescriptor`] can be created.
+pub trait GenerateSchemas {
+    fn generate_schemas(gen: &mut SchemaGenerator) -> Vec<(Schema, bool)>;
+}
+
+macro_rules! do_impls {
```
ruseinov, line 132: please document this.

src/rpc/reflect/openrpc_types.rs
```diff
@@ -0,0 +1,219 @@
...
+                )));
+            }
+        };
+        Ok(Self { inner: params })
+    }
+}
+
+validated_vec!(Params => ContentDescriptor);
+
+macro_rules! validated_vec {
```
ruseinov, line 172: nit: why not a derive macro? Seems more logical. I understand that you need to supply a param, but that's easy enough using attributes.

ruseinov: I guess the need to define it in a different module is a pain.
Let's make sure that all the macros are documented then, it would be easier to maintain and review.

src/rpc/reflect/parser.rs
```diff
@@ -0,0 +1,359 @@
...
+                return Err(ParseError::MustBeNamed)
+            }
+            (Some(RequestParameters::ByName(_)), ParamStructure::ByPosition) => {
+                return Err(ParseError::MustBePositional)
+            }
+            // `parse` won't be called, so do additional checks here
+            (Some(RequestParameters::ByPosition(it)), _) if names.is_empty() && !it.is_empty() => {
+                return Err(ParseError::UnexpectedPositional(it.len()))
+            }
+            (Some(RequestParameters::ByName(it)), _) if names.is_empty() && !it.is_empty() => {
```
ruseinov, line 70: let's have all the patterns documented. Some of them are obvious, some of them take a little while to understand.

src/rpc/reflect/openrpc_types.rs
aatifsyed: derive macros would need to be a different crate. I'm not against writing one, but this is much simpler

LesnyRumcajs (approved):

hanabi1224 (approved):

PR #4036, opened 2024-03-07
fix(p2p): improve chain exchange client logic

hanabi1224:
## Summary of changes

Follow-up of https://github.com/ChainSafe/forest/pull/4027
https://github.com/ChainSafe/forest/pull/4027 fixed a bug in chain exchange logic when `request len != response len`, however, the incomplete response leads to the below error(which does not block the sync tho). This PR tries to improve `get_batch` logic to get complete response with multiple chain exchange requests, to get rid of the below error.

```
bafy2bzacecvrbndaxoxqwgdx4ydlauxakrqyhdhorpx75hsrnkv3sauihneda] in EPOCH = 1413731 failed: Validation error: Validation error: Could not compute base fee: Invalid tipset: no msg root with cid 
```

-

Closes

## Change checklist

### DIFF
--- a/src/chain_sync/tipset_syncer.rs
+++ b/src/chain_sync/tipset_syncer.rs
@@ -981,6 +981,43 @@ async fn fetch_batch<DB: Blockstore>(
     network: &SyncNetworkContext<DB>,
     db: &DB,
 ) -> Result<Vec<FullTipset>, TipsetRangeSyncerError> {
+    const MAX_RETRY_ON_ERROR: usize = 3;
+    let mut n_retry_left = MAX_RETRY_ON_ERROR;
+    let mut error = None;
+
+    let mut result = vec![];
+    let mut n_missing = batch.len();
+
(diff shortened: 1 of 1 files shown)


src/chain_sync/tipset_syncer.rs
```diff
@@ -980,6 +980,43 @@ async fn fetch_batch<DB: Blockstore>(
     batch: Vec<Arc<Tipset>>,
     network: &SyncNetworkContext<DB>,
     db: &DB,
+) -> Result<Vec<FullTipset>, TipsetRangeSyncerError> {
```
LesnyRumcajs, line 983: Let's document this logic.

LesnyRumcajs (approved):

src/chain_sync/tipset_syncer.rs
LesnyRumcajs: We can do it in a follow-up PR, added to queue.

PR #4128, opened 2024-04-01
feat: log inner error in additon to the context

hanabi1224:
## Summary of changes

`anyhow::Context` provides a convenient way of wrapping the error with some context text, and the inner text is present in stack trace when the final error is unwrapped. However, In a long-running service like forest, most of the errors are logged with `tracing` instead of being unwrapped. In this case, unless the debug formatting(`"{error:?}"`) is specified, the inner errors which are quite helpful for troubleshooting in many cases are not present in the log text. To fix the issue, we could 
1. specify debug formatting
2. make a simple wrapper of `.map_err`.

This PR tries to implement the 2nd solution above.

- Introduce `crate::error::Context2` as a drop-in replacement of `anyhow::Context`

Closes

## Change checklist

### DIFF
--- a/src/state_manager/mod.rs
+++ b/src/state_manager/mod.rs
@@ -37,8 +37,9 @@ use crate::shim::{
 };
 use crate::state_manager::chain_rand::draw_randomness;
 use crate::state_migration::run_state_migrations;
+use crate::utils::error::Context2 as _;
 use ahash::{HashMap, HashMapExt};
-use anyhow::{bail, Context as _};
+use anyhow::bail;
 use bls_signatures::{PublicKey as BlsPublicKey, Serialize as _};
 use chain_rand::ChainRand;
 use cid::Cid;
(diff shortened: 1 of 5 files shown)


aatifsyed (changes requested):
I don't think we should do this (or I don't understand the problem enough).

```rust
let e: anyhow::Result;
tracing::error(?e, "error doing foo");
```

This gives us the desired formatting:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b139670fd7096276d92c0da257608a86

If you really want to do it this way, I would have a shim around anyhow::Error that just changed the formatting params

LesnyRumcajs:
How it's done is a detail. The fact that we need to use the `Debug` to get the full trace is non-intuitive, making the current error messages difficult to reason about.

Either this change needs to be implemented, one way or another, or all error messages should use the `?e` or a variant of that. If we do the latter, it should be enforced in CI. As discussed with @hanabi1224 in Slack, our expectation is that the `Display` of an error is a sum of its errors, not just the top-level context, which is rarely useful on its own.

aatifsyed (approved):
Fine with this as an interim measure, but prefer #4143

hanabi1224:
Superseded by https://github.com/ChainSafe/forest/pull/4174

Issue #4113, opened 2024-03-26
Exporting a snapshot with depth=900 might not be sufficient to bootstrap a forest or lotus node


hanabi1224:
**Describe the bug**

**To Reproduce** Steps to reproduce the behaviour:

1. `forest --chain calibnet --encrypt-keystore false --no-gc --height=-900 --auto-download-snapshot`
2. `forest-cli snapshot export -d=900`
3. `forest --chain calibnet --encrypt-keystore false --import-snapshot [SNAPSHOT]` or `lotus --remove-existing-chain --import-snapshot [SNAPSHTO]` fail with below errors

```
7u5nd7v6ovgi4dyitvsc757ayemvbmdg bafy2bzacecxlqvpliee2eliwtrw2f4jg2tmsxgamsaiphhm6guxprljkv57jw]: collectChain failed: collectChain syncMessages: message processing failed: validating block bafy2bzacebd5w4eezqkhbilor4xu3gvteaoquc762v2ucllhjdqmwmhtpbh4c: 1 error occurred:
        * determining if miner has min power failed:
    github.com/filecoin-project/lotus/chain/consensus/filcns.(*FilecoinEC).ValidateBlock.func2
        /opt/filecoin/chain/consensus/filcns/filecoin.go:186
  - loading power actor state:
    github.com/filecoin-project/lotus/chain/stmgr.minerHasMinPower
        /opt/filecoin/chain/stmgr/actors.go:410
  - load state tree:
    github.com/filecoin-project/lotus/chain/stmgr.(*StateManager).ParentState
        /opt/filecoin/chain/stmgr/read.go:28
  - failed to load state tree bafy2bzaced4wmuzsqsbeap77zfypimgtpolgtuxgjutvrwwg2itt3fqe3yxua:
    github.com/filecoin-project/lotus/chain/state.LoadStateTree
        /opt/filecoin/chain/state/statetree.go:295
  - failed to load hamt node:
    github.com/filecoin-project/specs-actors/actors/util/adt.AsMap
        /go/pkg/mod/github.com/filecoin-project/specs-actors@v0.9.15/actors/util/adt/map.go:41
  - ipld: could not find bafy2bzaced4wmuzsqsbeap77zfypimgtpolgtuxgjutvrwwg2itt3fqe3yxua
``` 

Repo CI log: https://github.com/ChainSafe/forest/actions/runs/8423386587/job/23080675345

**Log output**

<details>
  <summary>Log Output</summary>
  
[...]

PR #4133, opened 2024-04-02
feat(test): snapshot parity test against lotus

hanabi1224:
## Summary of changes

Part of https://github.com/ChainSafe/forest/issues/4113
For detailed description and repo steps, see https://github.com/ChainSafe/forest/issues/4113#issuecomment-2049485719

- Add snapshot parity tests
- ~Fix an issue that forest cannot be bootstrapped with a d=900 snapshot while lotus can be bootstrapped with the same snapshot~

Closes

## Change checklist


lemmih (approved):

elmattic (approved):

lemmih (dismissed):
Let's double check that we're actually aligned with Lotus' behavior.

hanabi1224:
> Let's double check that we're actually aligned with Lotus' behavior.

@lemmih I used the below steps to validate a lotus snapshot with the updated `forest-tool,` and it was successful.

1. `lotus chain export --tipset bafy2bzacebt3drb4kqhhx3axbts665skytvjg3pxdzyf3qhbux2qyp6bp27qg,bafy2bzacecpedn6y3ccdadoe4xyktq22hkbuxyfez3v7fp2cpr7aztiojpqn2 --recent-stateroots 2000 --skip-old-msgs ~/fr/snapshots/calibn
et/lotus_snapshot_calibnet_2024-04-02_height_1490667.car`
2. forest-tool snapshot validate lotus_snapshot_calibnet_2024-04-02_height_1490667.car

ruseinov (approved):

ruseinov:
Actually, why isn't there a code change?

.github/workflows/forest.yml
```diff
@@ -406,3 +409,49 @@ jobs:
...
+      - name: Prepare cache folder for uploading
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          docker run --rm \
+            -v snapshot_parity_filecoin-proofs:/proofs \
+            -v $FIL_PROOFS_PARAMETER_CACHE:/cache \
+            $SHELL_IMAGE \
+            sh -c "cp /proofs/* /cache"
+          sudo chmod -R 755 $FIL_PROOFS_PARAMETER_CACHE
+
```
lemmih, line 457: nit:
```suggestion
```

src/chain_sync/tipset_syncer.rs
```diff
@@ -1094,20 +1083,25 @@ async fn sync_messages_check_state<DB: Blockstore + Send + Sync + 'static>(
...
+                        full_tipset.clone(),
+                        genesis,
+                        invalid_block_strategy,
+                    )
+                    .await?;
+                    drop(timer);
+                    chainstore.set_heaviest_tipset(Arc::new(full_tipset.into_tipset()))?;
+                    tracker.write().set_epoch(current_epoch);
+                    metrics::LAST_VALIDATED_TIPSET_EPOCH.set(current_epoch);
+                }
```
lemmih, line 1104: I don't understand this. Why do we fail if we validate HEAD?

hanabi1224: When we import a depth=900 snapshot at epoch 1900, it contains 900 state roots from epoch 1001 to 1900, which is just sufficient for validating epoch 1901, but if we re-validate epoch 1900, state root at epoch 1000 is required.

lemmih: Why do we re-validate epoch 1900?

lemmih: Okay, I think I see what is going on.

To validate tipset N, we execute tipset N-1 and check that the state-root matches the parent state-root of tipset N. So, when we import a snapshot with X epochs and receive tipset X+1, we'll execute the messages in tipset X. This causes Forest to fail because the tipsets in the snapshot must be assumed to be correct.

How does Lotus get around this?

lemmih: In any case, I'm fairly sure this constraint is incorrect: `current_epoch != head.epoch() || full_tipset.key() != head.key()`. We need a more principled solution.

hanabi1224: Another option could be populating the validation result cache for the snapshot head after import, what do you think?

lemmih: You're skipping the validation of the heaviest tipset from the snapshot. I don't think that is correct.

lemmih: > Another option could be populating the validation result cache for the snapshot head after import, what do you think?

We need the state-root hash to populate the validation result cache. We can't get it until we've received the next tipset.

hanabi1224: > You're skipping the validation of the heaviest tipset from the snapshot. I don't think that is correct.

That's how lotus works. When importing the same depth=900 snapshot, lotus succeeds only because it does not validate the snapshot head

lemmih: Sounds good.

hanabi1224: Fix removed. I will continue the convo in the issue thread and attach the relevant lotus code.

ruseinov (approved):

Issue #4136, opened 2024-04-03
Forest is slow in entering FOLLOW mode


hanabi1224:
**Describe the bug**

**To Reproduce** Steps to reproduce the behaviour:

1. Run `forest --chain calibnet --encrypt-keystore false --auto-download-snapshot`
2. See console output

**Log output**

<details>
  <summary>Log Output</summary>
  
```console
2024-04-03T06:27:01.015114Z  INFO forest_filecoin::chain_sync::chain_muxer: Bootstrap successfully completed, now evaluating the network head to ensure the node is in sync
2024-04-03T06:27:01.015137Z  INFO forest_filecoin::chain_sync::chain_muxer: Evaluating network head...
2024-04-03T06:28:30.886281Z  INFO forest_filecoin::chain_sync::chain_muxer: Local node is behind the network, starting BOOTSTRAP from LOCAL_HEAD = 1493308 -> NETWORK_HEAD = 1493311
2024-04-03T06:28:30.886630Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1493308, N blocks = 2
2024-04-03T06:28:30.892328Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1493309, N blocks = 3
2024-04-03T06:28:30.932514Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1493310, N blocks = 1
2024-04-03T06:28:30.946807Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1493311, N blocks = 1
2024-04-03T06:28:30.953669Z  INFO forest_filecoin::chain_sync::chain_muxer: Bootstrap successfully completed, now evaluating the network head to ensure the node is in sync
2024-04-03T06:28:30.953701Z  INFO forest_filecoin::chain_sync::chain_muxer: Evaluating network head...
2024-04-03T06:30:00.326906Z  INFO forest_filecoin::chain_sync::chain_muxer: Local node is behind the network, starting BOOTSTRAP from LOCAL_HEAD = 1493311 -> NETWORK_HEAD = 1493314
2024-04-03T06:30:00.327499Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1493311, N blocks = 1
[...]

PR #4137, opened 2024-04-03
fix: ChainMuxer sometimes does not enter FOLLOW mode when it should

hanabi1224:
## Summary of changes

This PR tries to address an issue that `forest-cli sync wait` sometimes takes very long to exit by letting the node enter FOLLOW mode sooner when it's appropriate.  CI jobs that run `forest-cli sync wait` should also benefit from this change.

- 

Closes https://github.com/ChainSafe/forest/issues/4136

## Change checklist

### DIFF
--- a/src/chain_sync/chain_muxer.rs
+++ b/src/chain_sync/chain_muxer.rs
@@ -576,7 +576,11 @@ where
             // Otherwise in some conditions, `forest-cli sync wait` takes very long to exit (only when the node enters FOLLOW mode)
             match (
                 chain_store.heaviest_tipset().epoch(),
-                get_now_epoch(genesis_timestamp, block_delay as i64),
+                get_now_epoch(
+                    chrono::Utc::now().timestamp(),
+                    genesis_timestamp,
+                    block_delay as i64,
+                ),
             ) {
                 (local_epoch, now_epoch) if local_epoch >= now_epoch => {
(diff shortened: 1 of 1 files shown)


src/chain_sync/chain_muxer.rs
```diff
@@ -1002,3 +1012,16 @@ where
...
     }
 }
+
+// The formula matches lotus
+// ```go
+// sinceGenesis := build.Clock.Now().Sub(genesisTime)
+// expectedHeight := int64(sinceGenesis.Seconds()) / int64(build.BlockDelaySecs)
+// ```
+// See <https://github.com/filecoin-project/lotus/blob/b27c861485695d3f5bb92bcb281abc95f4d90fb6/chain/sync.go#L180>
+fn get_now_epoch(genesis_timestamp: i64, block_delay: i64) -> i64 {
```
lemmih, line 1022: nit: If we pass in the timestamp, this function would be pure. Pure functions are nice.

lemmih (approved):
This is okay as a temporary fix. It'll be replaced once we re-do the state machine.

src/chain_sync/chain_muxer.rs
hanabi1224: Fixed.

LesnyRumcajs (approved):

PR #4138, opened 2024-04-03
feat: implement `Filecoin.EthGetBlockByNumber`

### LINKED ISSUE #4183: `Filecoin.EthGetBlockByNumber`
Lotus documentation: https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#ethgetblockbynumber

elmattic:
## Summary of changes

- Add `Filecoin.EthGetBlockByNumber` RPC method

```console
./forest-tool api compare --lotus /ip4/127.0.0.1/tcp/1234/http --forest /ip4/127.0.0.1/tcp/2345/http forest_snapshot_calibnet_2024-04-03_height_1494010.forest.car.zst --filter EthGetBlockByNumber
| RPC Method                       | Forest | Lotus |
|----------------------------------|--------|-------|
| Filecoin.EthGetBlockByNumber (2) | Valid  | Valid |
```

Closes #4183

## Change checklist


src/lotus_json/mod.rs
```diff
@@ -320,6 +320,40 @@ pub mod hexify_bytes {
...
+        serializer.serialize_str(&s)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        #[allow(clippy::indexing_slicing)]
+        if s.len() >= 2 && &s[..2] == "0x" {
```
LesnyRumcajs, line 345: You can avoid this warning with `.get(..2)`

src/lotus_json/mod.rs
```diff
@@ -320,6 +320,40 @@ pub mod hexify_bytes {
     }
 }
 
+pub mod hexify_vec_bytes {
```
LesnyRumcajs, line 323: Isn't https://docs.rs/hex/latest/hex/ doing it already?

src/rpc/methods/chain.rs
```diff
@@ -869,6 +869,58 @@ quickcheck::quickcheck! {
     }
 }
 
+pub async fn get_parent_receipts<DB: Blockstore + Send + Sync + 'static>(
```
lemmih, line 872: Can you use `Receipt::get_receipt` here? If not, let's explain why in the comments.

src/lotus_json/mod.rs
LesnyRumcajs: E.g., https://github.com/ChainSafe/forest/blob/ccfcf4dab725bbfce8fe7d1a373e307c1b9fa149/src/blocks/tests/serialization_vectors.rs#L84-L85

Or https://github.com/ChainSafe/forest/blob/18c5f2617369942c5157223dc0bee5a7331a61e9/src/lotus_json/mod.rs#L297-L321 (not sure if's not redundant as well)

src/rpc/methods/chain.rs
```diff
@@ -869,6 +869,58 @@ quickcheck::quickcheck! {
...
+    data: Ctx<DB>,
+    message_receipts: Cid,
+) -> Result<Vec<ApiReceipt>> {
+    let store = data.state_manager.blockstore();
+
+    let mut receipts = Vec::new();
+
+    // Try Receipt_v4 first. (Receipt_v4 and Receipt_v3 are identical, use v4 here)
+    if let Ok(amt) = Amt::<fvm_shared4::receipt::Receipt, _>::load(&message_receipts, store)
```
LesnyRumcajs, line 881: It'd make sense to have a `shim`, so that the usage for `fvmX` is not sprawled across the entire codebase. We utilize this pattern quite a bit there. What do you think?

src/rpc/methods/chain.rs
```diff
@@ -869,6 +869,58 @@ quickcheck::quickcheck! {
...
+        .map_err(|_| {
+            ErrorObjectOwned::owned::<()>(
+                1,
```
LesnyRumcajs, line 884: What is this magic number?

src/rpc/methods/chain.rs
```diff
@@ -869,6 +869,58 @@ quickcheck::quickcheck! {
...
+                format!("failed to root: ipld: could not find {}", message_receipts),
```
LesnyRumcajs, line 885:
```suggestion
                format!("failed to root: ipld: could not find {message_receipts}"),
```

src/rpc/methods/chain.rs
```diff
@@ -869,6 +869,58 @@ quickcheck::quickcheck! {
...
+            });
+            Ok(())
+        })?;
+    } else {
+        // Fallback to Receipt_v2.
+        let amt = Amt::<fvm_shared2::receipt::Receipt, _>::load(&message_receipts, store).map_err(
+            |_| {
+                ErrorObjectOwned::owned::<()>(
+                    1,
+                    format!("failed to root: ipld: could not find {}", message_receipts),
```
LesnyRumcajs, line 905:
```suggestion
                    format!("failed to root: ipld: could not find {message_receipts}"),
```

src/rpc/methods/eth.rs
```diff
@@ -34,11 +49,54 @@ pub const ETH_BLOCK_NUMBER: &str = "Filecoin.EthBlockNumber";
...
+
+const EVM_WORD_LENGTH: usize = 32;
+
+/// Keccak-256 of an RLP of an empty array
+const EMPTY_UNCLES: &str = "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347";
+
+const EIP_1559_TX_TYPE: u64 = 2;
+
+/// The address used in messages to actors that have since been deleted.
+const REVERTED_ETH_ADDRESS: &str = "0xff0000000000000000000000ffffffffffffffff";
```
LesnyRumcajs, line 74: There's a golden opportunity here for docs! I'd love to learn about more about empty uncles :rofl:

src/rpc/methods/eth.rs
```diff
@@ -72,9 +156,47 @@ impl Address {
         }
     }
 
+    pub fn from_filecoin_address(addr: &FilecoinAddress) -> Result<Self> {
```
LesnyRumcajs, line 159: Is this method unit-tested?

LesnyRumcajs: Could you link here the corresponding Go impl?

src/rpc/methods/eth.rs
```diff
@@ -72,9 +156,47 @@ impl Address {
...
     fn is_masked_id(&self) -> bool {
         self.0.as_bytes().starts_with(&MASKED_ID_PREFIX)
     }
+
+    #[allow(clippy::indexing_slicing)]
+    fn from_actor_id(id: u64) -> Self {
+        let mut payload = ethereum_types::H160::default();
+        payload.as_bytes_mut()[0] = 0xff;
+        payload.as_bytes_mut()[12..20].copy_from_slice(&id.to_be_bytes());
+
```
LesnyRumcajs, line 197: Could we use `get(X..Y)` instead?

LesnyRumcajs: Is it covered by unit tests?

src/rpc/methods/eth.rs
```diff
@@ -179,6 +325,196 @@ impl HasLotusJson for BlockNumberOrHash {
...
+            gas_limit: tx.gas.0,
+            input: tx.input.0,
+            v: tx.v,
+            r: tx.r,
+            s: tx.s,
+        }
+    }
+}
+
+#[allow(clippy::indexing_slicing)]
```
LesnyRumcajs, line 447: Could we use `.get`?

src/rpc/methods/eth.rs
```diff
@@ -179,6 +325,196 @@ impl HasLotusJson for BlockNumberOrHash {
...
+        let i = (value.leading_zeros() / 8) as usize;
+        let bytes = value.to_be_bytes();
+        bytes[i..].into()
+    } else {
+        // If all bytes are zero, return an empty slice
+        BytesMut::new()
+    }
+}
+
+#[allow(clippy::indexing_slicing)]
```
LesnyRumcajs, line 459: Could we use `.get`?

src/rpc/methods/eth.rs
```diff
@@ -179,6 +325,196 @@ impl HasLotusJson for BlockNumberOrHash {
...
+            input: tx.input.0,
+            v: tx.v,
+            r: tx.r,
+            s: tx.s,
+        }
+    }
+}
+
+#[allow(clippy::indexing_slicing)]
+fn format_u64(value: u64) -> BytesMut {
```
LesnyRumcajs, line 448: Is it unit tested?

src/rpc/methods/eth.rs
```diff
@@ -179,6 +325,196 @@ impl HasLotusJson for BlockNumberOrHash {
...
+        let bytes = value.to_be_bytes();
+        bytes[i..].into()
+    } else {
+        // If all bytes are zero, return an empty slice
+        BytesMut::new()
+    }
+}
+
+#[allow(clippy::indexing_slicing)]
+fn format_bigint(value: &BigInt) -> BytesMut {
```
LesnyRumcajs, line 460: Is it unit-tested?

src/rpc/methods/eth.rs
```diff
@@ -393,10 +729,449 @@ fn tipset_by_block_number_or_hash<DB: Blockstore>(
...
+fn eth_tx_args_from_unsigned_eth_message(msg: &Message) -> Result<TxArgs> {
+    let mut to = Some(Address::default());
+    let mut params = vec![];
+
+    if msg.version != 0 {
+        bail!("unsupported msg version: {}", msg.version);
+    }
+
+    if !msg.params().bytes().is_empty() {
+        // TODO: could we do better?
```
LesnyRumcajs, line 773: Could we?

LesnyRumcajs (commented):
I think there's a good opportunity here to:
* add lots of unit tests - I believe we control the inputs here pretty easily, so we can do as much assertions (and `quickcheck`s) as possible.
* Add extensive docs - the code is non-trivial, especially for those not accustomed with Ethereum internals,
* let's avoid index slicing suppression by using `.get`

src/rpc/methods/eth.rs
elmattic: Nope.

src/rpc/methods/eth.rs
elmattic: Link added.

src/rpc/methods/chain.rs
elmattic: I don't know 🤡

src/rpc/methods/eth.rs
elmattic: Uncles are blocks that do not participate in the main chain, but uncle miners are rewarded nevertheless because they burn some energy during Ethash PoW.
In our case, we pretend there are no uncles whatsoever. We have tipsets!

src/rpc/methods/chain.rs
elmattic: Sounds good.

src/lotus_json/mod.rs
elmattic: https://docs.rs/serde_with/latest/serde_with/hex/struct.Hex.html
Looks like they don't prepend with `0x` (if I'm looking at the right doc). Could we tweak that?

elmattic: `hexify_bytes` works only on fixed-size arrays, AFAIK.

elmattic: Maybe `with_prefix!` could help us for `0x`.

src/rpc/methods/chain.rs
elmattic: `Receipt::get_receipt` return some `Option<Receipt>` but we need to build a `Vec<ApiReceipt>`.

src/rpc/methods/eth.rs
elmattic: In that case we will need to return a `Result<BytesMut>`, no?

src/rpc/methods/eth.rs
```diff
@@ -393,10 +729,449 @@ fn tipset_by_block_number_or_hash<DB: Blockstore>(
...
+        .first()
+        .parent_base_fee
+        .clone()
+        .into();
+    block.gas_used = Uint64(gas_used);
+    block.transactions = if full_tx_info {
+        Transactions::Full(full_transactions)
+    } else {
+        Transactions::Hash(hash_transactions)
+    };
```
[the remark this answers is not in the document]
elmattic: Sure.

src/rpc/methods/eth.rs
LesnyRumcajs: If you are 100% sure this can't fail, you can `.expect` to keep the same return type.

src/lotus_json/mod.rs
LesnyRumcajs: This could help, perhaps. https://docs.rs/serde_with/latest/serde_with/macro.with_prefix.html

src/rpc/methods/eth.rs
elmattic: Done.

src/rpc/methods/eth.rs
elmattic: I've added a test. Turns out that `format_bigint` only works for unsigned bigints (but that was expected).
I will change the `eth::BigInt` constructor to reflect this.

elmattic: In go there's no `big.BigUInt` only `big.BigInt`. Why don't we use `BigUInt` on our side in `eth` module? Just a question.

src/rpc/methods/eth.rs
elmattic: I don't think so.

src/lotus_json/mod.rs
elmattic: This avoid slice indexing but `split_at_checked` in unstable:
```rust
fn hex_str_to_bytes(value: &str) -> Result<Vec<u8>, anyhow::Error> {
    if let Some((head, tail)) = value.split_at_checked(2) {
        if head == "0x" && value.len() % 2 == 0 {
            return tail
                .as_bytes()
                .chunks(2)
                .map(|b| {
                    if let Ok(s) = std::str::from_utf8(b) {
                        u8::from_str_radix(s, 16).context("invalid digit found")
                    } else {
                        bail!("non-utf8 string")
                    }
                })
                .collect();
        }
    }
    bail!("invalid hex string")
}
```

src/rpc/methods/eth.rs
LesnyRumcajs: Likely nobody thought of using `BigUint`. :) Not sure if there won't be some wild `-1` numbers every now and then, I'd leave it to `BigInt` which is everywhere in our codebase anyway.

src/lotus_json/mod.rs
LesnyRumcajs: I don't really understand, how is it related?

elmattic: Never mind, it was a test to avoid using `unwrap/except`.
I'll go with `.get(..2)`.

src/rpc/methods/eth.rs
elmattic: Exactly.

src/rpc/methods/chain.rs
elmattic: It's an implementation-defined server-error code for `JSONRPC`.

src/rpc/methods/chain.rs
elmattic: Done. And I will refactor some other functions in a subsequent PR.

src/shim/executor.rs
```diff
@@ -167,6 +167,30 @@ impl Receipt {
         let receipts = amt.get(i)?;
         Ok(receipts.cloned().map(Receipt::V2))
     }
+
+    pub fn get_parent_receipts(
```
lemmih, line 171: There's no `parent` here. This function should just be named `get_receipts`.

src/rpc/methods/eth.rs
```diff
@@ -393,10 +737,455 @@ fn tipset_by_block_number_or_hash<DB: Blockstore>(
     }
 }
 
+async fn execute_tipset<DB: Blockstore + Send + Sync + 'static>(
+    data: Ctx<DB>,
+    tipset: &Arc<Tipset>,
+) -> Result<(Cid, Vec<(ChainMessage, ApiReceipt)>)> {
```
lemmih, line 743: Why do you need `ApiReceipt` rather than just `Receipt`?

lemmih (changes requested):

src/rpc/methods/chain.rs
```diff
@@ -861,6 +862,33 @@ quickcheck::quickcheck! {
     }
 }
 
+pub fn get_parent_receipts<DB: Blockstore + Send + Sync + 'static>(
```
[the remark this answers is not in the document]
elmattic: Fixed.

src/shim/executor.rs
```diff
@@ -167,6 +167,27 @@ impl Receipt {
         let receipts = amt.get(i)?;
         Ok(receipts.cloned().map(Receipt::V2))
     }
+
+    pub fn get_receipts(db: &impl Blockstore, receipts: Cid) -> anyhow::Result<Vec<Receipt>> {
+        let mut parent_receipts = Vec::new();
```
lemmih, line 172:
```suggestion
    pub fn get_receipts(db: &impl Blockstore, receipts_cid: Cid) -> anyhow::Result<Vec<Receipt>> {
        let mut receipts = Vec::new();
```

lemmih (approved):
Lots of moving parts.

LesnyRumcajs:
No green checkmark, no review!

PR #4155, opened 2024-04-08
refactor: migrate `beacon_api` to use `trait RpcMethod`

aatifsyed:
## Summary of changes

- Add some supporting methods to [`RpcTest`]
- Title

Closes

## Change checklist

### DIFF
--- a/src/tool/subcommands/api_cmd.rs
+++ b/src/tool/subcommands/api_cmd.rs
@@ -213,7 +213,7 @@ struct RpcTest {
     ignore: Option<&'static str>,
 }
 
-/// Duplication between <method> and <method>_raw is a temporary measure, and
+/// Duplication between `<method>` and `<method>_raw` is a temporary measure, and
 /// should be removed when <https://github.com/ChainSafe/forest/issues/4032> is
 /// completed.
 impl RpcTest {


lemmih (approved):

src/rpc/beacon_api.rs
```diff
@@ -1,27 +1,38 @@
...
-    let (first,): (ChainEpoch,) = params.parse()?;
-
-    let (_, beacon) = data.beacon.beacon_for_epoch(first)?;
-    let rr =
-        beacon.max_beacon_round_for_epoch(data.state_manager.get_network_version(first), first);
-    let e = beacon.entry(rr).await?;
-    Ok(e.into())
+pub enum BeaconGetEntry {}
+impl RpcMethod<1> for BeaconGetEntry {
+    const NAME: &'static str = "Filecoin.BeaconGetEntry";
```
LesnyRumcajs, line 23: Side quest (likely low priority) - RpcMethod (or something else) should support aliases to make switching between Forest and Lotus easier. https://github.com/filecoin-project/lotus/blob/master/api/eth_aliases.go

Unless it is supported somewhere else already?

LesnyRumcajs (approved):

src/rpc/beacon_api.rs
aatifsyed: I think this should be doable in `RpcMethodExt::register` if we add the following member:
```rust
trait RpcMethod {
    const ALIASES: &[&'static str] = [];
}
```

I'll add this to #4032

PR #4158, opened 2024-04-08
refactor: migrate chain_api to use trait RpcMethod (part 1)

aatifsyed:
## Summary of changes

-

Closes

## Change checklist

### DIFF
--- a/src/rpc/chain_api.rs
+++ b/src/rpc/chain_api.rs
@@ -13,8 +13,8 @@ use crate::lotus_json::HasLotusJson;
 use crate::lotus_json::LotusJson;
 #[cfg(test)]
 use crate::lotus_json::{assert_all_snapshots, assert_unchanged_via_json};
-use crate::message::ChainMessage;
-use crate::rpc::types::{ApiHeadChange, ApiTipsetKey, BlockMessages};
+use crate::message::{ChainMessage, SignedMessage};
+use crate::rpc::types::{ApiHeadChange, ApiTipsetKey};
 use crate::rpc::{
     reflect::SelfDescribingRpcModule, ApiVersion, Ctx, JsonRpcError, RPCState, RpcMethod,
     RpcMethodExt as _,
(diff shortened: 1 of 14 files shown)


lemmih (approved):

src/rpc/chain_api.rs
```diff
@@ -41,198 +42,91 @@ use tokio::sync::{
...
-        let messages = load_api_messages_from_tipset(store, &parent_tipset)?;
-        Ok(LotusJson(messages))
+pub enum ChainGetParentMessages {}
+impl RpcMethod<1> for ChainGetParentMessages {
+    const NAME: &'static str = "Filecoin.ChainGetParentMessages";
+    const PARAM_NAMES: [&'static str; 1] = ["block_cid"];
+    const API_VERSION: ApiVersion = ApiVersion::V0;
+    type Params = (LotusJson<Cid>,);
+    type Ok = LotusJson<Vec<ApiMessage>>;
+    async fn handle(
```
LesnyRumcajs, line 89: Can we have a breather here and in other places to separate different logical pieces of the implementation?

```
    const NAME: &'static str = "Filecoin.ChainGetParentMessages";
    const PARAM_NAMES: [&'static str; 1] = ["block_cid"];
    const API_VERSION: ApiVersion = ApiVersion::V0;

    type Params = (LotusJson<Cid>,);
    type Ok = LotusJson<Vec<ApiMessage>>;

    async fn handle(
```

src/rpc/snapshots/.gitattributes
```diff
@@ -0,0 +1 @@
+*.snap linguist-generated=true
```
LesnyRumcajs, line 1: This is not a well-known feature; let's add a link/comment to help readers understand it.

src/rpc/chain_api.rs
LesnyRumcajs: While it's not a part of `rustfmt` (sadly), it's a standard found in most places, e.g., https://doc.rust-lang.org/std/str/trait.FromStr.html

LesnyRumcajs (approved):

Issue #4143, opened 2024-04-04
use `tracing`'s structured logging


aatifsyed:
Before:
https://github.com/ChainSafe/forest/blob/9e3ed095a32d14a293ce7ed7b52b48d74d3b67b3/src/chain/store/chain_store.rs#L221
After:
```rust
info!(epoch = %ts.epoch(), key = %ts.key(), "new heaviest tipset");
```

Before:
https://github.com/ChainSafe/forest/blob/9e3ed095a32d14a293ce7ed7b52b48d74d3b67b3/src/rpc/channel.rs#L360
After:
```rust
tracing::error!(?e, "Failed to serialize channel message");
```

# Plan
- refactor all `tracing` callsites
  - Could this be automated?
- educate team on
  - `std::error::Error` and `anyhow::Error`'s design
  - (structured) logging principles
- [hard] write CI

# Motivation
> Either... or all error messages should use the `?e` or a variant of that. If we do the latter, it should be enforced in CI. 

_Originally posted by @LesnyRumcajs in https://github.com/ChainSafe/forest/issues/4128#issuecomment-2036753097_

PR #4174, opened 2024-04-10
refactor: use tracing's structured logging

aatifsyed:
#4143

Semi-automated with  https://github.com/aatifsyed/synsert/blob/5f7d39c1ff1ae47d2b028b045378197036a71fa1/examples/tracing.rs

## Change checklist

### DIFF
--- a/src/blocks/tipset.rs
+++ b/src/blocks/tipset.rs
@@ -329,9 +329,9 @@ impl Tipset {
                 ticket.vrfproof < other_ticket.vrfproof
             });
         if broken {
-            info!("Weight tie broken in favour of {}", self.key());
+            info!(winner = %self.key(), "weight tie broken");
         } else {
-            info!("Weight tie left unbroken, default to {}", other.key());
+            info!(default = %other.key() , "weight tie defaulted");
         }
         broken
     }
--- a/src/chain_sync/chain_muxer.rs
+++ b/src/chain_sync/chain_muxer.rs
@@ -271,7 +271,7 @@ where
                 match network.hello_request(peer_id, request).await {
                     Ok(response) => response,
                     Err(e) => {
-                        debug!("Hello request failed: {}", e);
+                        debug!(error = %e, "hello request failed");
                         return;
                     }
                 };
@@ -301,17 +301,13 @@ where
         network: SyncNetworkContext<DB>,
     ) -> Result<FullTipset, ChainMuxerError> {
         debug!(
-            "Received block over GossipSub: {} height {} from {}",
-            block.header.cid(),
-            block.header.epoch,
-            source,
+            block.header.cid = %block.header.cid(), %block.header.epoch, %source, "received block over GossipSub"
         );
 
         let epoch = block.header.epoch;
 
         debug!(
-            "Getting messages of gossipblock, epoch: {epoch}, block: {}",
-            block.header.cid()
+            % epoch, block.header.cid = %block.header.cid(), "getting messages of gossipblock"
         );
         // Get bls_message in the store or over Bitswap
         let bls_messages: Vec<_> = block
@@ -344,8 +340,7 @@ where
     fn handle_pubsub_message(mem_pool: Arc<MessagePool<M>>, message: SignedMessage) {
         if let Err(why) = mem_pool.add(message) {
             debug!(
-                "GossipSub message could not be added to the mem pool: {}",
-                why
+                reason = %why, "GossipSub message could not be added to the mem pool"
             );
         }
     }
@@ -386,7 +381,7 @@ where
                 {
                     Ok(tipset) => tipset,
                     Err(why) => {
-                        debug!("Querying full tipset failed: {}", why);
+                        debug!(reason = %why, "querying full tipset failed");
                         return Err(why);
                     }
                 };
@@ -480,8 +475,7 @@ where
             < chain_store.heaviest_tipset().epoch()
         {
             debug!(
-                "Skip processing tipset at epoch {} from {source} that is too old",
-                tipset.epoch()
+                tipset.epoch = %tipset.epoch(), %source, "skip processing old tipset"
             );
             return Ok(None);
         }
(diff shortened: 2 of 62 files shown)


src/chain_sync/chain_muxer.rs
```diff
@@ -271,7 +271,7 @@ where
                 match network.hello_request(peer_id, request).await {
                     Ok(response) => response,
                     Err(e) => {
-                        debug!("Hello request failed: {}", e);
+                        debug!(error = %e, "hello request failed");
```
hanabi1224, line 274: Should we use `?` here to log the inner error?

```rust
Err(error) => {
                        debug!(?error, "hello request failed");
                        return;
                    }
```

src/chain_sync/chain_muxer.rs
```diff
@@ -344,8 +340,7 @@ where
     fn handle_pubsub_message(mem_pool: Arc<MessagePool<M>>, message: SignedMessage) {
         if let Err(why) = mem_pool.add(message) {
             debug!(
-                "GossipSub message could not be added to the mem pool: {}",
-                why
+                reason = %why, "GossipSub message could not be added to the mem pool"
```
hanabi1224, line 343:
```suggestion
                reason = ?why, "GossipSub message could not be added to the mem pool"
```

src/chain_sync/chain_muxer.rs
```diff
@@ -386,7 +381,7 @@ where
                 {
                     Ok(tipset) => tipset,
                     Err(why) => {
-                        debug!("Querying full tipset failed: {}", why);
+                        debug!(reason = %why, "querying full tipset failed");
```
hanabi1224, line 384:
```suggestion
                        debug!(reason = ?why, "querying full tipset failed");
```

src/chain_sync/chain_muxer.rs
```diff
@@ -495,8 +489,7 @@ where
         ) {
             metrics::INVALID_TIPSET_TOTAL.inc();
             warn!(
-                "Validating tipset received through GossipSub failed: {}",
-                why
+                reason = %why, "failed to validate tipset received through GossipSub"
```
hanabi1224, line 492:
```suggestion
                reason = ?why, "failed to validate tipset received through GossipSub"
```

src/chain_sync/chain_muxer.rs
```diff
@@ -531,7 +524,7 @@ where
                 let event = match p2p_messages.recv_async().await {
                     Ok(event) => event,
                     Err(why) => {
-                        debug!("Receiving event from p2p event stream failed: {why}");
+                        debug!(reason = %why, "receiving event from p2p event stream failed");
```
hanabi1224, line 527:
```suggestion
                        debug!(reason = ?why, "receiving event from p2p event stream failed");
```

hanabi1224 (approved):

src/chain_sync/network_context.rs
```diff
@@ -173,9 +171,7 @@ where
                         || header_len != msg.secp_msg_includes.len()
                     {
                         tracing::warn!(
-                            "header_len: {header_len}, msg.bls_msg_includes.len(): {}, msg.secp_msg_includes.len(): {}",
-                            msg.bls_msg_includes.len(),
-                            msg.secp_msg_includes.len()
+                            %header_len, msg.bls_msg_includes.len = %msg.bls_msg_includes.len(), msg.secp_msg_includes.len = %msg.secp_msg_includes.len()
```
ruseinov, line 174: nit: kinda nice, but confusing. I would never be able to deduce what `msg.bls_msg_includes.len = %msg.bls_msg_includes.len()` does without the diff.

aatifsyed: I think this just comes down to knowing `tracing::event`'s API, like knowing `format!` or `printf` 🤷

aatifsyed:
```console
$ git diff main | cat | rg '@@' | wc -l
182
```
So many call sites

LesnyRumcajs:
I'll look into `tracing` docs and get back to you here.

src/chain_sync/network_context.rs
ruseinov: Perhaps it would be easier to understand if you did something like `bls_msg_includes_len = ...`. It's not blocking, but this is not reading well nor should reading the code require one to know `tracing` API specifics. `format!` and `printf` are intuitive, while this just looks like you are assigning a value to a property of an object.

src/blocks/tipset.rs
```diff
@@ -329,9 +329,9 @@ impl Tipset {
                 ticket.vrfproof < other_ticket.vrfproof
             });
         if broken {
-            info!("Weight tie broken in favour of {}", self.key());
+            info!(winner = %self.key(), "weight tie broken");
         } else {
-            info!("Weight tie left unbroken, default to {}", other.key());
+            info!(default = %other.key() , "weight tie defaulted");
```
LesnyRumcajs, line 334: Let's not use the `default`, I believe it's a keyword to be used in the future.

src/chain_sync/chain_muxer.rs
LesnyRumcajs, line 384: https://github.com/ChainSafe/forest/pull/4174/files#r1559147574

src/chain_sync/network_context.rs
LesnyRumcajs: While I agree that the `tracing` API is not particularly obvious, it seems a standard in the Rust ecosystem https://blessed.rs/crates#section-common-subsection-logging. As such, we should learn how to utilize it and get familiar with it. I recommend tinkering a bit to get a hold of it, what results are produced etc. We can't stay in the `printf` world forever. :t-rex:

LesnyRumcajs (commented):
What do we do with the errors marked as `%` - Display? Shouldn't they all be `?` to include the the entire trace, not only the latest `context`?

LesnyRumcajs:
@aatifsyed what do we do with this draft?

Issue #3370, opened 2023-08-14
Consider using abscissa for cli and secrets management


aatifsyed:
We're doing a lot of thinking about
- Breaking up forest into multiple CLIs: https://github.com/ChainSafe/forest/issues/3172
- CLI/config file unification https://github.com/ChainSafe/forest/issues/3246

Can we use https://docs.rs/abscissa/latest/abscissa/ to just do all that thinking for us?

aatifsyed:
@creativcoder What is the acceptance criteria for the work you're thinking of doing?

I **don't** think you should be porting our CLI to abscissa until it's clear that it's the right thing to do, hence _consider_.

What are the rough edges of our current approach?
- Inconsistent arguments across the codebase e.g [#3407](https://github.com/ChainSafe/forest/pull/3407)
- Arguments appear that don't make sense for the current operation e.g [#2076](https://github.com/ChainSafe/forest/issues/2076)
- We don't have a good clap/config file unification story: [#3182](https://github.com/ChainSafe/forest/issues/3182)
- We're worried about secrets management: [#2274](https://github.com/ChainSafe/forest/discussions/2274)
- Any more?

So the next step on this issue is a write-up, maybe a PoC of a new approach with solves all or some of these issues.
Can you articulate why it would be better?

See also [#3246](https://github.com/ChainSafe/forest/issues/3246) for using other libraries to do this

PR #4177, opened 2024-04-10
proofs api refactor

LesnyRumcajs:
## Summary of changes

- refactored the proofs API module a bit, which includes:
  - separating the files, logically
  - adjusting visibility
  - adding docs (and adjusting current ones to reflect reality)
  - adding tests where the logic was more than trivial
  - code-golfing

Closes

## Change checklist


documentation/src/environment_variables.md
```diff
@@ -4,14 +4,16 @@ Besides CLI options and the configuration values in the configuration file,
...
-| FOREST_KEYSTORE_PHRASE          | any text                         | empty   | The passphrase for the encrypted keystore                                 |
-| FOREST_CAR_LOADER_FILE_IO       | 1 or true                        | false   | Load CAR files with `RandomAccessFile` instead of `Mmap`                  |
-| FOREST_DB_DEV_MODE              | [see here](#-forest_db_dev_mode) | current | The database to use in development mode                                   |
-| FOREST_ACTOR_BUNDLE_PATH        | file path                        | empty   | Path to the local actor bundle, download from remote servers when not set |
-| FIL_PROOFS_PARAMETER_CACHE      | dir path                         | empty   | Path to folder that caches fil proof parameter files                      |
-| FOREST_PROOFS_ONLY_IPFS_GATEWAY | 1 or true                        | false   | Use only IPFS gateway for proofs parameters download                      |
+| Environment variable            | Value                            | Default                          | Description                                                               |
```
aatifsyed, line 7: new issue: ban `std::env::var`, and have one `ForestVars` structure

- We should have one struct where forest environment variables are described (for end users) and accessed (in code)
- Remove `documentation/src/environment_variables.md`, and have it link to the rustdoc for the above

These means we have:
- better control over environment variables
- documentation that stays up-to-date

We could also have an approach like #3370 or using https://docs.rs/figment/latest/figment/index.html for config

Cargo did something similar last year: https://github.com/rust-lang/cargo/issues/11588

src/utils/proofs_api/parameters.rs
```diff
@@ -0,0 +1,156 @@
...
+use cid::Cid;
+use serde::{Deserialize, Serialize};
+use tracing::{debug, warn};
+
+use super::is_env_truthy;
+
+const PROOF_DIGEST_LEN: usize = 16;
+
+/// Environment variable that allows skipping checksum verification of the parameter files.
+const TRUST_PARAMS_ENV: &str = "TRUST_PARAMS";
```
aatifsyed, line 26: nonblocking: if this is dangerous, I'd prefer `FOREST_FORCE_TRUST_PARAMS` or `FOREST_UNSAFE_TRUST_PARAMS`

src/utils/proofs_api/parameters.rs
```diff
@@ -0,0 +1,156 @@
...
+/// the miners. These external crates require a specific set of parameter files
+/// to be located at in a specific folder. By default, it is
+/// `/var/tmp/filecoin-proof-parameters` but it can be overridden by the
+/// `FIL_PROOFS_PARAMETER_CACHE` environment variable. Forest will automatically
+/// download the parameter files from Cloudflare/IPFS and verify their validity. For
+/// consistency, Forest will prefer to download the files it's local data
+/// directory. To this end, the `FIL_PROOFS_PARAMETER_CACHE` environment
+/// variable is updated before the parameters are downloaded.
+///
+/// More information available here: <https://github.com/filecoin-project/rust-fil-proofs#parameter-file-location>
```
aatifsyed, line 107: nit: permalink please

aatifsyed (approved):

documentation/src/environment_variables.md
aatifsyed: #4182

aatifsyed: (nonblocking, obviously)

src/utils/proofs_api/paramfetch.rs
```diff
@@ -14,28 +18,28 @@ use crate::{
...
 
 /// Domain bound to the Cloudflare R2 bucket.
 const CLOUDFLARE_PROOF_PARAMETER_DOMAIN: &str = "filecoin-proof-parameters.chainsafe.dev";
 
 /// If set to 1, enforce using the IPFS gateway for fetching parameters.
 const PROOFS_ONLY_IPFS_GATEWAY_ENV: &str = "FOREST_PROOFS_ONLY_IPFS_GATEWAY";
-const DIR_ENV: &str = "FIL_PROOFS_PARAMETER_CACHE";
-const GATEWAY_ENV: &str = "IPFS_GATEWAY";
-const TRUST_PARAMS_ENV: &str = "TRUST_PARAMS";
+const IPFS_GATEWAY_ENV: &str = "IPFS_GATEWAY";
```
hanabi1224, line 42: Maybe worth mentioning in doc somewhere
https://lotus.filecoin.io/kb/nodes-in-china/
```
Running Lotus requires the download of chain’s proof parameters which are large files which by default are hosted outside of China and very slow to download there. To get around that, users should set the following environment variable when running either of lotus, lotus-miner and lotus-worker:

export IPFS_GATEWAY=https://proof-parameters.s3.cn-south-1.jdcloud-oss.com/ipfs/
```

hanabi1224: At first glance I feel we should prefix the environment variable with `FOREST_` but in this case maybe it's more convenient to re-use the lotus one. They don't prefix it with `LOTUS_` for some reason.

hanabi1224 (approved):

src/utils/proofs_api/parameters.rs
LesnyRumcajs: Same is in other comment; I was a bit hesitant to change the variable names to avoid breaking changes, but probably better do it now than later.

documentation/src/environment_variables.md
ruseinov: It would indeed be nice, this would also allow for specifying one prefix for all the env variables, e.g. `FOREST_`. 
Though instead of having to do the plumbing for this it'd be nice to have some sort of `singleton` that could be accessed anywhere in the code.

src/utils/proofs_api/paramfetch.rs
ruseinov: > At first glance I feel we should prefix the environment variable with `FOREST_` but in this case maybe it's more convenient to re-use the lotus one. They don't prefix it with `LOTUS_` for some reason.

Agreed on the prefix. As stated above - it would be nice to have some sort of tool that would allow us to specify the prefix for everything we load. Then we can load variables with and without the prefix.

LesnyRumcajs: @hanabi1224 Great point there, I was not aware of that. On top of that, I was able to find a bug (not sure how long it was there) where the `IPFS_GATEWAY` was not overridable - the environment variable only changed what was logged. :clown_face: See [a4c53c5](https://github.com/ChainSafe/forest/pull/4177/commits/a4c53c53ea7ff6cf0115c24539b2efecc6eb3661)

ruseinov (approved):

ruseinov (approved):

hanabi1224 (approved):

PR #4187, opened 2024-04-12
feat: fix and verify GasEstimateGasLimit

### LINKED ISSUE #901: RPC Gas Limit Estimation Under Estimating
** Acceptance Criteria ** 
- [ ] Gas API should accurately estimate the gas without hardcoding extra gas

**Describe the bug**

- Gas estimation is done when sending a transaction
- Gas estimation simulates the call into the VM and uses that as a gas limit
- Originally, ran into this when spamming our node with send calls (just hardcoded a workaround at the time)
- The Gas API's gas limit estimation keeps under reporting. Right now, we just add some extra gas to the gas limit, but this is a temporary fix for the sake of unblocking us in other areas. There is a TODO to fix this and should be looked into https://github.com/ChainSafe/forest/blob/main/node/rpc/src/gas_api.rs#L253

lemmih:
## Summary of changes

- Fix the gas limit computation by signing the incoming message (with a dummy signature).
- Verify exact compliance of `GasEstimateGasLimit` against Lotus.

Closes #901

## Change checklist

### DIFF
--- a/src/rpc/methods/gas.rs
+++ b/src/rpc/methods/gas.rs
@@ -220,7 +220,9 @@ where
         )),
         Protocol::Delegated => ChainMessage::Signed(SignedMessage::new_unchecked(
             msg,
-            Signature::new(SignatureType::Delegated, vec![0; 65]),
+            // In Lotus, delegated signatures have the same length as SECP256k1.
+            // This may or may not change in the future.
+            Signature::new(SignatureType::Delegated, vec![0; SECP_SIG_LEN]),
         )),
         _ => ChainMessage::Unsigned(msg),
     };
--- a/src/rpc/methods/gas.rs
+++ b/src/rpc/methods/gas.rs
@@ -213,7 +213,7 @@ where
     // Pretend that the message is signed. This has an influence on the gas
     // cost. We obviously can't generate a valid signature. Instead, we just
     // fill the signature with zeros. The validity is not checked.
-    let mut chain_msg = match from_a.protocol().into() {
+    let mut chain_msg = match from_a.protocol() {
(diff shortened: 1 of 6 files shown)


src/rpc/methods/gas.rs
```diff
@@ -195,19 +210,31 @@ where
         .unwrap_or_default();
 
     let ts = data.mpool.cur_tipset.lock().clone();
+    // Pretend that the message is signed. This has an influence on the gas
+    // cost. We obviously can't generate a valid signature. Instead, we just
+    // fill the signature with zeros. The validity is not checked.
+    let mut chain_msg = match from_a.protocol() {
+        Protocol::Secp256k1 => ChainMessage::Signed(SignedMessage::new_unchecked(
+            msg,
+            Signature::new_secp256k1(vec![0; 65]),
```
LesnyRumcajs, line 219: Let's use a constant from `ref-fvm`; `SECP_SIG_LEN`.

lemmih: Do you think I can use `SECP_SIG_LEN` for the delegated signature as well?

LesnyRumcajs: I don't think delegated addresses _must_ to use `secp256k1` and thus their length - https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0055.md#delegated-signature-type

That said, delegated signatures are used, to my understanding, for Ethereum messages, which also use `secp256k1`, so we should be safe assuming it's that as well. I'd  use the constant and add a comment.

LesnyRumcajs (approved):

aatifsyed (approved):

PR #4191, opened 2024-04-13
refactor: migrate wallet api to trait RpcEndpoint

aatifsyed:
## Summary of changes

-

Closes

## Change checklist


LesnyRumcajs (approved):

src/wallet/subcommands/wallet_cmd.rs
```diff
@@ -204,7 +221,7 @@ pub enum WalletCommands {
     /// Get account balance
     Balance {
         /// The address of the account to check
-        address: String,
+        address: Address,
```
lemmih, line 224: This is a regression. Using `Address` rather than `String + StrictAddress` silently ignores mainnet/testnet address mismatches. For the safe handling of funds, it's important to fail if the user, for example, tries to send FIL from a mainnet address to a testnet address.

lemmih: I'm not sure how to write a unit test to catch this.

lemmih (changes requested):
Looks good. Have to revert to using `String + StrictAddress`, though.

src/wallet/subcommands/wallet_cmd.rs
LesnyRumcajs: Could we have a test in https://github.com/ChainSafe/forest/blob/main/scripts/tests/calibnet_wallet_check.sh for this?

aatifsyed: I'll punt the address test to https://github.com/ChainSafe/forest/issues/2945

src/wallet/subcommands/wallet_cmd.rs
```diff
@@ -252,7 +269,7 @@ pub enum WalletCommands {
     /// Validates whether a given string can be decoded as a well-formed address
     ValidateAddress {
         /// The address to be validated
-        address: String,
+        address: Address,
```
lemmih, line 272: Using `Address` here moves the validation to the client side. The point of this command is to validate on the server side. I can't argue for this command's usefulness, but let's do everything client-side or server-side.

lemmih (changes requested):

src/wallet/subcommands/wallet_cmd.rs
aatifsyed: Gotcha, changed the implementation of `WalletValidateAddress` to match lotus' too

PR #4198, opened 2024-04-15
feat(rpc): implement Filecoin.StateGetBeaconEntry

hanabi1224:
## Summary of changes

> StateGetBeaconEntry returns the beacon entry for the given filecoin epoch. If the entry has not yet been produced, the call will block until the entry becomes available

- implement Filecoin.StateGetBeaconEntry and tests

```
| Filecoin.StateGetBeaconEntry (3)                | Valid  | Valid |
```

Closes https://github.com/ChainSafe/forest/issues/4195

https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#stategetbeaconentry

## Change checklist


lemmih (approved):

src/rpc/methods/state.rs
```diff
@@ -1148,3 +1157,39 @@ pub async fn state_deal_provider_collateral_bounds<DB: Blockstore + Send + Sync
...
+            let block_delay = ctx.state_manager.chain_config().block_delay_secs;
+            let now_epoch = crate::chain_sync::get_now_epoch(
+                chrono::Utc::now().timestamp(),
+                genesis_timestamp as _,
+                ctx.state_manager.chain_config().block_delay_secs as _,
+            );
+            if epoch <= now_epoch {
+                break;
+            }
+            tokio::time::sleep(Duration::from_secs(block_delay as _)).await;
```
LesnyRumcajs, line 1186: Can we deterministically calculate when we're at the right epoch? This would get rid of a loop, and allow getting the beacon almost immediately if it should be there in 1s (otherwise, we'd wait 29s for no good reason). What do you think?

lemmih: It might not be worth the extra logic. Requesting the beacon for future epochs is unlikely to happen often.

hanabi1224: I've made the change before seeing the discussion, plz take another look.

src/rpc/methods/state.rs
```diff
@@ -1176,3 +1185,39 @@ pub async fn state_deal_provider_collateral_bounds<DB: Blockstore + Send + Sync
...
+            let block_delay = ctx.state_manager.chain_config().block_delay_secs as i64;
+            // Give it a 1s clock drift buffer
+            let epoch_timestamp = genesis_timestamp + block_delay * epoch + 1;
+            let now_timestamp = chrono::Utc::now().timestamp();
+            match epoch_timestamp.saturating_sub(now_timestamp) {
+                diff if diff > 0 => {
+                    tokio::time::sleep(Duration::from_secs(diff as u64)).await;
+                }
+                _ => {}
+            };
```
LesnyRumcajs, line 1214: Might be simpler to just sleep for zero in case it's negative. Not a huge deal obviously.
```suggestion
            let diff = epoch_timestamp.saturating_sub(now_timestamp).max(0) as u64;
            tokio::time::sleep(Duration::from_secs(diff)).await;
```

LesnyRumcajs (approved):

PR #4203, opened 2024-04-15
refactor: migrate sync to trait RpcMethod

aatifsyed:
## Summary of changes

Blocked on #4200 

-

Closes

## Change checklist


src/rpc/methods/eth.rs
```diff
@@ -320,14 +318,15 @@ pub async fn eth_get_balance<DB: Blockstore>(
...
+pub async fn eth_syncing<DB: Blockstore + Sync + Send + 'static>(
     _params: Params<'_>,
     data: Ctx<DB>,
 ) -> Result<LotusJson<EthSyncingResult>, ServerError> {
-    let RPCSyncState { active_syncs } = sync_state(data).await?;
+    let crate::rpc::sync::RPCSyncState { active_syncs } =
+        crate::rpc::sync::SyncState::handle(data.clone(), ()).await?;
     match active_syncs
-        .iter()
-        .rev()
```
lemmih, line 330: Do we not need `.rev()` anymore?

src/cli/subcommands/attach_cmd.rs
```diff
@@ -326,11 +327,6 @@ impl AttachCommand {
         bind_request!(context, api,
                 // Node API
                 "node_status" => |()| ApiInfo::node_status_req(),
-
-                // Sync API
-                "sync_check_bad" => ApiInfo::sync_check_bad_req,
```
hanabi1224, line 331: Should we put `forest-cli attach` behind a feature if we disable most of the APIs? `boa-engine` takes quite long to compile  cc @elmattic

src/rpc/methods/eth.rs
aatifsyed: I think that would change the logic, no?

aatifsyed: good catch!

src/cli/subcommands/attach_cmd.rs
aatifsyed: Defining a pattern for adding these back is tracked in #4032, so there's a chance it would be feature gated only to be re-enabled. I don't mind - your argument is compelling

PR #4233, opened 2024-04-17
fix: avoid unnecessarily copies in decode_zstd_single_frame

hanabi1224:
## Summary of changes

- Resolve a comment `// This unnecessarily copies the zstd frame. :(` in `fn decode_zstd_single_frame`

Closes

## Change checklist

### DIFF
--- a/src/db/car/forest.rs
+++ b/src/db/car/forest.rs
@@ -242,12 +242,10 @@ where
 
 fn decode_zstd_single_frame<ReaderT: Read>(reader: ReaderT) -> io::Result<BytesMut> {
     let mut zstd_frame = vec![];
-
     zstd::Decoder::new(reader)?
         .single_frame()
         .read_to_end(&mut zstd_frame)?;
-    // This unnecessarily copies the zstd frame. :(
-    Ok(BytesMut::from(zstd_frame.as_slice()))
+    Ok(zstd_frame.into_iter().collect())
 }
 
 pub struct Encoder {}


src/db/car/forest.rs
```diff
@@ -242,12 +242,10 @@ where
 
 fn decode_zstd_single_frame<ReaderT: Read>(reader: ReaderT) -> io::Result<BytesMut> {
     let mut zstd_frame = vec![];
-
     zstd::Decoder::new(reader)?
         .single_frame()
         .read_to_end(&mut zstd_frame)?;
-    // This unnecessarily copies the zstd frame. :(
-    Ok(BytesMut::from(zstd_frame.as_slice()))
+    Ok(zstd_frame.into_iter().collect())
```
LesnyRumcajs, line 248: Won't `collect` do the copy on its own?

aatifsyed (changes requested):
This comment points out that the memory usage of this function is 2*len

It remains 2*len after this change.

src/db/car/forest.rs
hanabi1224: I think it should transfer ownership and avoid copy or clone. In below sample, `A` does not derive `Copy` or `Clone`
 ```rust
#[derive(Debug)]
struct A;

fn main() {
    let a = vec![A, A, A];
    let b: Vec<_> = a.into_iter().collect();
    println!("{b:?}");
}
```

**Comment by hanabi1224** [comment]
@aatifsyed with the below code, 
when I use `let b: BytesMut = a.as_slice().into();` line, I got 
```
dhat: Total:     2,048 bytes in 2 blocks
dhat: At t-gmax: 2,048 bytes in 2 blocks
dhat: At t-end:  0 bytes in 0 blocks
dhat: The data has been saved to dhat-heap.json, and is viewable with dhat/dh_view.html
```
when I use `let b: BytesMut = a.into_iter().collect();` line, I got
```
dhat: Total:     1,024 bytes in 1 blocks
dhat: At t-gmax: 1,024 bytes in 1 blocks
dhat: At t-end:  0 bytes in 0 blocks
dhat: The data has been saved to dhat-heap.json, and is viewable with dhat/dh_view.html
```

```rust
use bytes::BytesMut;

#[global_allocator]
static ALLOC: dhat::Alloc = dhat::Alloc;

fn main() {
    let _profiler = dhat::Profiler::new_heap();
    let a = vec![1; 1024];
    let b: BytesMut = a.as_slice().into();
    // let b: BytesMut = a.into_iter().collect();
}
```

src/db/car/forest.rs
hanabi1224: Also checkout https://github.com/ChainSafe/forest/pull/4233#issuecomment-2061553746

aatifsyed (approved):
I stand corrected!

Your proof works even in `--dev`, so I was wrong about the semantics.

For those following along at home, here's why:
- https://github.com/tokio-rs/bytes/blob/4e2c9c065a06bf9cb5d7dd46e3b29f62a1c20057/src/bytes_mut.rs#L1314
- https://github.com/rust-lang/rust/blob/6c6b3027ef62e911142cfc55589baef4e9f38ec8/library/alloc/src/vec/spec_from_iter.rs#L37-L64

src/db/car/forest.rs
LesnyRumcajs: I need to go back to Rust for Dummies.

LesnyRumcajs (approved):

PR #4235, opened 2024-04-17
fix: api compare test failures

hanabi1224:
## Summary of changes

This PR tries to fix [CI test failures](https://github.com/ChainSafe/forest/actions/runs/8725844010/job/23940006136?pr=4218)

```
Forest response: []
Lotus response: null
Diff: -[]

+null
```

The root cause is, the framework accepts any serializable Ok types whose raw JSON formats differ from their lotus JSON 
 formats. (Vec<T> in this case, vec![] should serialize into null instead of [])  Constrain `Ok: HasLotusJson` to avoid this kind of bug.

-

Closes

## Change checklist

### DIFF
--- a/src/rpc/methods/chain.rs
+++ b/src/rpc/methods/chain.rs
@@ -68,7 +68,7 @@ impl RpcMethod<1> for ChainGetMessage {
     const API_VERSION: ApiVersion = ApiVersion::V0;
 
     type Params = (LotusJson<Cid>,);
-    type Ok = LotusJson<Message>;
+    type Ok = Message;
 
     async fn handle(
         ctx: Ctx<impl Blockstore>,
@@ -79,10 +79,10 @@ impl RpcMethod<1> for ChainGetMessage {
             .blockstore()
             .get_cbor(&msg_cid)?
             .with_context(|| format!("can't find message with cid {msg_cid}"))?;
-        Ok(LotusJson(match chain_message {
+        Ok(match chain_message {
             ChainMessage::Signed(m) => m.into_message(),
             ChainMessage::Unsigned(m) => m,
-        }))
+        })
     }
 }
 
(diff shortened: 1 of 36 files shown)


src/rpc/methods/chain.rs
```diff
@@ -94,7 +93,7 @@ impl RpcMethod<1> for ChainGetParentMessages {
     const API_VERSION: ApiVersion = ApiVersion::V0;
 
     type Params = (LotusJson<Cid>,);
-    type Ok = Vec<ApiMessage>;
+    type Ok = LotusJson<Vec<ApiMessage>>;
```
hanabi1224, line 96: @aatifsyed Should we constrain `Ok: HasLotusJson` to avoid this?

lemmih (approved):
Looks okay to me. @aatifsyed Is this aligned with your vision?

LesnyRumcajs (approved):

aatifsyed (commented):
> @aatifsyed Is this aligned with your vision?

Not quite.

I want to separate the `LotusJson` and `RpcMethod` APIs for clarity - they're currently intertwined and I find them quite confusing.

The proximate cause for this is me changing methods to return `Vec<T>` instead of `LotusJson<Vec<T>>` somewhere migrating methods for #4032.

I did this because `LotusJson<Vec<T>>` currently translates to `Option<Vec<LotusJson<T::LotusJson>>` for de/serialization.

The root cause is my bad design decision when I implemented `HasLotusJson for Vec<T>`:
https://github.com/ChainSafe/forest/blob/52b31c0746b80a17197e49e86553225f77f29813/src/lotus_json/vec.rs#L9-L12

That's why this PR has had to add so many `lotus_json_with_self!` calls.

If we want to keep the APIs separate, RPC types should NOT have to implement `HasLotusJson`:
https://github.com/ChainSafe/forest/blob/c4fc0b2aa5ade0d776a107b44a118020ab83d977/src/rpc/mod.rs#L56-L93

So what's the "right" fix if we want to keep this "vision"?
- Change `impl HasLotusJson for Vec` to not be recursive.
- Change sites where there used to be `LotusJson<Vec<T>>` to `LotusJson<Vec<LotusJson<T>>`
- Change the `#[serde(with = .. )]` annotations. This is the hard part.

Alternative approaches:
- Keep the APIs entangled. That's this PR. We still get the RPCMethod stuff, but the code is a bit entangled IMO.
- Create a new `rpc::Vec` to handle this stuff. This also helps us with JsonSchema stuff, and I think we'll have to do something similar for `NonEmpty` too.

@hanabi1224 what do you think?

PR #4238, opened 2024-04-19
fix(rpc): Filecoin.StateListMessages

hanabi1224:
## Summary of changes

- Fix the RPC method and migrate to the new RPC framework
- Fix the tests

```
| Filecoin.StateListMessages (243)                | Valid  | Valid |
```

Closes

## Change checklist

### DIFF
--- a/src/tool/subcommands/api_cmd.rs
+++ b/src/tool/subcommands/api_cmd.rs
@@ -750,7 +750,7 @@ fn snapshot_tests(store: Arc<ManyCar>, n_tipsets: usize) -> anyhow::Result<Vec<R
             }
 
             let (bls_messages, secp_messages) = crate::chain::store::block_messages(&store, block)?;
-            for msg in bls_messages.into_iter().unique() {
+            for msg in bls_messages.into_iter().unique().take(5) {
                 tests.push(RpcTest::identity(ChainGetMessage::request((msg
                     .cid()?
                     .into(),))?));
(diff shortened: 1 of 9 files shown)


src/tool/subcommands/api_cmd.rs
```diff
@@ -820,30 +839,33 @@ fn snapshot_tests(store: Arc<ManyCar>, n_tipsets: usize) -> anyhow::Result<Vec<R
...
-                tests.push(RpcTest::identity(ApiInfo::state_list_messages_req(
+                    }
+                    .into(),
+                    LotusJson(tipset.key().into()),
+                    tipset.epoch().into(),
+                ))?));
+                tests.push(RpcTest::identity(StateListMessages::request((
                     MessageFilter {
                         from: None,
-                        to: None,
```
hanabi1224, line 842: Empty MessageFilter causes internal errors for both forest and lotus.  
If we want to keep this test case, we need to treat it a success when InternalErrors happen on both sides. @lemmih   @LesnyRumcajs what do you think?

LesnyRumcajs: I'm okay with the condition that it's explicit for this particular test (like `should_fail`). I wouldn't like the behaviour like before, where all matching internal server errors were considered a success.

lemmih (approved):

src/tool/subcommands/api_cmd.rs
hanabi1224: Sounds great! Opened https://github.com/ChainSafe/forest/issues/4241 to track this feature.

LesnyRumcajs (approved):

PR #4257, opened 2024-04-22
EthGetBlockByNumber: Fix incorrect initialization

elmattic:
## Summary of changes

- Fix incorrect initialization of `to` address
- Add unit test

Closes

## Change checklist

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -39,12 +39,6 @@ jobs:
         uses: mozilla-actions/sccache-action@v0.0.4
         timeout-minutes: ${{ fromJSON(env.CACHE_TIMEOUT_MINUTES) }}
         continue-on-error: true
-      - name: Apt Dependencies
-        uses: nick-fields/retry@v3
-        with:
-          timeout_minutes: 5
-          max_attempts: 3
-          command: sudo make install-deps
       - name: Link Checker (Repo Readme)
         uses: lycheeverse/lychee-action@v1.9.3
         with:
--- a/.github/workflows/forest.yml
+++ b/.github/workflows/forest.yml
@@ -68,9 +68,6 @@ jobs:
         uses: mozilla-actions/sccache-action@v0.0.4
         timeout-minutes: '${{ fromJSON(env.CACHE_TIMEOUT_MINUTES) }}'
         continue-on-error: true
-      - name: Install Apt Dependencies
-        run: |
-          sudo make install-deps
       - name: Cargo Install
         env:
           # To minimize compile times: https://nnethercote.github.io/perf-book/build-configuration.html#minimizing-compile-times
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,18 +19,6 @@ jobs:
     steps:
       - name: Checkout Sources
         uses: actions/checkout@v4
-      - name: Apt Dependencies
-        if: contains(matrix.os, 'ubuntu')
-        uses: nick-fields/retry@v3
-        with:
-          timeout_minutes: 5
-          max_attempts: 3
-          command: sudo make install-deps
-      - name: Homebrew Utils
-        if: contains(matrix.os, 'macos')
-        run: |
-          brew install --verbose coreutils protobuf
-          protoc --version
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
         timeout-minutes: 5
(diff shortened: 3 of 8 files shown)


src/rpc/methods/eth.rs
```diff
@@ -1387,4 +1387,42 @@ mod test {
...
...152600060208201526102748383845161027b565b9392505050565b604080518082019091526060815260006020820152825182111561029e57600080fd5b83515160006102ad848361041d565b905085602001518111156102ca576102ca86610197836002610430565b8551805183820160200191600091808511156102e4578482525b505050602086015b60208610610324578051825261030360208361041d565b915061031060208261041d565b905061031d602087610469565b95506102ec565b51815160001960208890036101000a0190811690199190911617905250849150509392505050565b60006020828403121561035e57600080fd5b813567ffffffffffffffff8116811461027457600080fd5b60005b83811015610391578181015183820152602001610379565b50506000910152565b602081526000825160208084015280518060408501526103c1816060860160208501610376565b601f01601f1916929092016060019392505050565b6001600160f81b03198316815281516000906103f9816001850160208701610376565b919091016001019392505050565b634e487b7160e01b600052601160045260246000fd5b808201808211156100a6576100a6610407565b80820281158282048414176100a6576100a6610407565b60008261046457634e487b7160e01b600052601260045260246000fd5b500690565b818103818111156100a6576100a661040756fea264697066735822122081b223bd9e056cbd628bc4f964ffff9dffe3fbccec93813ea82acb424d10b94364736f6c63430008130033").unwrap()),
```
LesnyRumcajs, line 1404: Could we have a human-readable version of this string, or is it not feasible?

src/rpc/methods/eth.rs
```diff
@@ -1387,4 +1387,42 @@ mod test {
...
+        use crate::shim::address::Address;
+        use fvm_ipld_encoding::RawBytes;
+
+        let msg = Message {
+            version: 0,
+            from: Address::from_str("f410fxcemznbd52stbbtwdm4ayh3edyyyksnkmjbadva").unwrap(),
+            to: Address::from_str("f010").unwrap(),
+            sequence: 4,
+            value: TokenAmount::default(),
+            method_num: 4,
```
LesnyRumcajs, line 1403: How are the parameters to this test generated? Are all of them random-ish or do they convey a particular meaning? For example, what happens if I change the `method_num` to 10, `sequence` to 42 and `version` to 3? Will the binary blob below change? It'd be helpful to add some comments here.

src/rpc/methods/eth.rs
elmattic: It's feasible but not worth it, and it would not bring tangible benefits. It's supposed to be bytes that can decode to something like a byte array (an RLP-encoded message).

src/rpc/methods/eth.rs
elmattic: It's an input payload that generates a `to` address that is `None`.

The payload was extracted using our tool `api compare` with `--n-tipsets 3000`, and I grabbed the ones that fit the bill.
That's a good point. I will add some comments. We only care about the `to` member being `None`, so an assertion of this property will suffice.

hanabi1224 (approved):

src/rpc/methods/eth.rs
LesnyRumcajs: Can we not set this at all? Looking briefly at `eth_tx_args_from_unsigned_eth_message`, it doesn't seem to bring a lot to a test that is supposed to check the `to` field. Same goes for other values - seems they could be pretty much default.

elmattic: @LesnyRumcajs I've covered the two codepath, so the `None` and the `Some` cases.

src/rpc/methods/eth.rs
```diff
@@ -1411,4 +1412,36 @@ mod test {
...
+            ..Message::default()
+        };
+
+        assert_eq!(
+            eth_tx_args_from_unsigned_eth_message(&msg)
+                .unwrap()
+                .to
+                .unwrap(),
+            Address(H160::from_str("0xa251031ed6b4779e2a0b913683e71043d88002a3").unwrap())
+        );
```
LesnyRumcajs, line 1445: Perhaps this could be split into a separate test? Not a big deal.

LesnyRumcajs (approved):
Beautiful!

src/rpc/methods/eth.rs
elmattic: Done.

hanabi1224 (approved):

hanabi1224 (approved):

PR #4267, opened 2024-04-24
Decrease `request_timeout` in builders to avoid panics

elmattic:
## Summary of changes

- Decrease `request_timeout` value in `WsClientBuilder` and `HttpClientBuilder` to avoid panics (see logs below)

Rational:

We don't need to handle timeouts at the `jsonrpsee` client level since we shim our own timeouts during the call, but setting it to some max value seems unsafe.

Setting a 10s timeout and simulating a slow endpoint in forest daemon:

```console
$ FOREST_RPC_DEFAULT_TIMEOUT=10 ./forest-tool api compare --lotus /ip4/127.0.0.1/tcp/1234/ws --forest /ip4/127.0.0.1/tcp/2345/ws forest_snapshot_calibnet_2024-04-19_height_1539740.forest.car.zst --filter EthChainId
Request dump: RpcRequest { method_name: "Filecoin.EthChainId", params: Null, result_type: PhantomData<serde_json::value::Value>, api_version: V1, timeout: 10s }
Request params JSON: null
Lotus response: "0x4cb2f"

| RPC Method          | Forest              | Lotus |
|---------------------|---------------------|-------|
| Filecoin.EthChainId | InternalServerError | Valid |
Error: Some tests failed
```

Closes

```
$ RUST_BACKTRACE=1 ./forest-tool api compare --lotus /ip4/127.0.0.1/tcp/1234/ws --forest /ip4/127.0.0.1/tcp/2345/ws forest_snapshot_calibnet_2024-04-19_height_1539740.forest.car.zst --filter EthChainId
thread 'tokio-runtime-worker' panicked at library/std/src/time.rs:417:33:
overflow when adding duration to instant
```

## Change checklist

### DIFF
--- a/src/rpc/client.rs
+++ b/src/rpc/client.rs
@@ -18,6 +18,7 @@ use jsonrpsee::core::client::ClientT as _;
 use jsonrpsee::core::params::{ArrayParams, ObjectParams};
 use jsonrpsee::core::ClientError;
 use serde::de::DeserializeOwned;
+use std::time::Duration;
 use tracing::{debug, Instrument, Level};
 use url::Url;
 
@@ -152,6 +153,7 @@ impl Debug for UrlClient {
 
 impl UrlClient {
     async fn new(url: Url, token: impl Into<Option<String>>) -> Result<Self, ClientError> {
+        const ONE_DAY: Duration = Duration::from_secs(24 * 3600); // we handle timeouts ourselves.
         let headers = match token.into() {
             Some(token) => HeaderMap::from_iter([(
                 header::AUTHORIZATION,
@@ -172,6 +174,7 @@ impl UrlClient {
                     .set_headers(headers)
                     .max_request_size(MAX_REQUEST_BODY_SIZE)
                     .max_response_size(MAX_RESPONSE_BODY_SIZE)
+                    .request_timeout(ONE_DAY)
                     .build(&url)
                     .await?,
             ),
@@ -180,6 +183,7 @@ impl UrlClient {
                     .set_headers(headers)
                     .max_request_size(MAX_REQUEST_BODY_SIZE)
                     .max_response_size(MAX_RESPONSE_BODY_SIZE)
+                    .request_timeout(ONE_DAY)
                     .build(&url)?,
             ),
             it => {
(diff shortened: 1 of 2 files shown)


elmattic:
> looks good.
> 
> nit: I'm still unsure what does `// we handle timeouts ourselves` comment mean, but it's perhaps out of scope anyway.

Check this code: https://github.com/ChainSafe/forest/blob/main/src/rpc/client.rs#L65-L97

src/rpc/client.rs
```diff
@@ -153,7 +153,7 @@ impl Debug for UrlClient {
 
 impl UrlClient {
     async fn new(url: Url, token: impl Into<Option<String>>) -> Result<Self, ClientError> {
-        let timeout = Duration::MAX; // we handle timeouts ourselves.
```
LesnyRumcajs, line 156: Will a subscription method like `ChainNotify` behave correctly with this change?

elmattic: Right now, we don't have a clear plan for testing such methods, so we ignore it.

LesnyRumcajs: Why do we ignore it? I understand we don't have an automated way of testing this yet. If you set locally the timeout to 10s, will this method still work as expected after 15s?

LesnyRumcajs: Also, why does the overflow occur? Why is a value added to the timeout we don't use?

elmattic: > Why do we ignore it?

Testing a method like `ChainNotify` and `EthSubscribe` requires new infrastructure code that we don't have at the moment.

elmattic: > If you set locally the timeout to 10s, will this method still work as expected after 15s?

Which method are you referring to?

hanabi1224:
> overflow when adding duration to instant

This seems like a bug in upstream lib, should we PR a fix as well?

src/rpc/client.rs
LesnyRumcajs: I'm referring to `ChainNotify`. We implemented it, so we had to test it somehow, right?

LesnyRumcajs (approved):

PR #4270, opened 2024-04-26
[Fix] GC issues on calibnet.

ruseinov:
## Summary of changes

- use sequential graph traversal for now
- store/retrieve last gc epoch and use that for further runs

work on #3863 

We definitely need to debug the parallel graph traversal to see what goes wrong. 

In the meantime I have several takeaways:
- parallel graph traversal needs proper testing and fixes
- calibnet GC works, because the system requirements for calibnet nodes aren't huge. it works on a 4GB RAM node with 3GB of swap
- problems with mainnet GC likely have something to do with the fact that we consume as much as possible memory for car file indexing and the snapshot is quite big. needs to be tested

## Change checklist

### DIFF
--- a/src/db/gc/mod.rs
+++ b/src/db/gc/mod.rs
@@ -72,8 +72,8 @@
 use crate::blocks::Tipset;
 use crate::chain::ChainEpochDelta;
 
-use crate::db::{truncated_hash, GarbageCollectable};
-use crate::ipld::unordered_stream_graph;
+use crate::db::{truncated_hash, GarbageCollectable, SettingsStore};
+use crate::ipld::stream_graph;
 use crate::shim::clock::ChainEpoch;
 use ahash::{HashSet, HashSetExt};
 use futures::StreamExt;
@@ -83,6 +83,8 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::time;
 use tracing::info;
+
+const SETTINGS_KEY: &str = "LAST_GC_RUN";
 
 /// [`MarkAndSweep`] is a simple garbage collector implementation that traverses all the database
 /// keys writing them to a [`HashSet`], then filters out those that need to be kept and schedules
@@ -99,7 +101,7 @@ pub struct MarkAndSweep<DB> {
     block_time: Duration,
 }
 
-impl<DB: Blockstore + GarbageCollectable + Sync + Send + 'static> MarkAndSweep<DB> {
+impl<DB: Blockstore + SettingsStore + GarbageCollectable + Sync + Send + 'static> MarkAndSweep<DB> {
     /// Creates a new mark-and-sweep garbage collector.
     ///
     /// # Arguments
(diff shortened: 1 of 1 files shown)


ruseinov:
cc @hanabi1224 this seems to fix calibnet GC, works on a weak 4GB + 3GB node.

src/db/gc/mod.rs
```diff
@@ -204,6 +223,8 @@ impl<DB: Blockstore + GarbageCollectable + Sync + Send + 'static> MarkAndSweep<D
         info!("GC sweep");
         self.sweep()?;
 
+        self.update_last_gc_run(current_epoch)?;
```
hanabi1224, line 226: I'm wondering why it has to be `current_epoch` instead of `current_epoch - 2 * CHAIN_FINALITY`. That way, we don't have to wait

ruseinov: The original idea was that we don’t want to continuously run this, because the graph traversal takes up significant amount of resources. And there is no real reason to do so, as we don’t get so much unreachable stuff on each run.

hanabi1224: This idea is to avoid waiting during a GC run. We can still set threshold between multiple runs to reduce frequency. Does that make sense or did I misunderstand something?

ruseinov: We discussed that before - there is no way for us to avoid waiting during a GC run, because the only way to know the age of any given CID is by knowing when we marked it, as we are iterating all the available keys, without knowing which epoch they belong to. The GC makes no assumptions about the marked CIDs. 

This PR fixes calibnet GC, feel free to propose design improvements via an issue, then we can review those and see what’s possible.

ruseinov: About the last run epoch - we record the epoch GC fully executed last. The way that it’s further used in calculations for the next run is whatever we decide it to be. It’s more logical to use the actual epoch when the GC finished as opposed to some calculated value.

hanabi1224: As I understand, the algorithm needs to wait for `depth` epochs, but logically, these epochs don't have to be from the future, and what I suggest is to wait for existing `depth` epochs in the same algorithm, to avoid waiting in time, does that make sense?

ruseinov: > As I understand, the algorithm needs to wait for `depth` epochs, but logically, these epochs don't have to be from the future, and what I suggest is to wait for `depth` epochs that already exist in the database with the same algorithm to avoid waiting in time (`EPOCH_DELAY * depth`), does that make sense?

It doesn't make sense, because we populate keys based on iterating everything that exists in the database (ParityDB, not CarDB). So we have absolutely no idea which epoch the CID belongs to. This is why we have to wait `2*chain_finality` after we have marked certain keys for deletion.

hanabi1224 (approved):

src/db/gc/mod.rs
```diff
@@ -167,15 +169,32 @@ impl<DB: Blockstore + GarbageCollectable + Sync + Send + 'static> MarkAndSweep<D
         }
     }
 
+    fn update_last_gc_run(&self, epoch: ChainEpoch) -> anyhow::Result<()> {
+        self.db
+            .write_bin(SETTINGS_KEY, epoch.to_string().as_bytes())
+    }
+
+    // Unfortunately there seems to be no good way of decoding a slice into i64 without array init
+    // and manipulation, therefore a string representation is used.
```
aatifsyed, line 178: Are you sure?
```
i64::from_le_bytes(self.db.read_bin(SETTINGS_KEY)?.unwrap_or_default().try_into()?)
```

aatifsyed (approved):

src/db/gc/mod.rs
ruseinov: Depends on what you consider a `good way`. I'm aware of the `try_into` option.

PR #4281, opened 2024-05-01
refactor: use nunny instead of nonempty

aatifsyed:
## Summary of changes

- `nunny` supports `arbitrary` and `schemars`, and saves us memory on some iterators

Closes

## Change checklist


ruseinov (approved):

src/tool/subcommands/benchmark_cmd.rs
```diff
@@ -197,7 +197,10 @@ async fn benchmark_forest_encoding(
     let file = tokio::io::BufReader::new(File::open(&input).await?);
 
     let mut block_stream = CarStream::new(file).await?;
-    let roots = std::mem::take(&mut block_stream.header.roots);
+    let roots = std::mem::replace(
```
hanabi1224, line 200: Why not `impl<T: Default> Default for nunny:Vec`?

src/tool/subcommands/car_cmd.rs
```diff
@@ -57,15 +56,15 @@ impl CarCommands {
...
                     .await?;
 
-                let all_roots = NonEmpty::from_vec(
+                let all_roots = NonEmpty::new(
                     car_streams
                         .iter()
                         .flat_map(|it| it.header.roots.iter())
                         .unique()
                         .cloned()
                         .collect_vec(),
```
hanabi1224, line 65: There seems to be an opportunity to save the `collect_vec` call if `NonEmpty` can be constructed from an iterator

aatifsyed: It can't be contracted from an iterator - the iterator might be empty

src/tool/subcommands/benchmark_cmd.rs
aatifsyed: I think a non-empty default isn't very natural

PR #4286, opened 2024-05-06
chore(test): bump lotus version in CI tests

hanabi1224:
## Summary of changes

- Bump lotus version to 1.27
- Fix `SectorOnChainInfo` struct
- ~Blacklist `EthGetBlockByNumber`~
- Fix `gasLimit` in `EthGetBlockByNumber`

Failures(gasLimit not match in EthGetBlockByNumber):
```
-  "gasLimit": "0x2540be400",

+  "gasLimit": "0x6fc23ac00",
```

```go
func NewEthBlock(hasTransactions bool, tipsetLen int) EthBlock {
	b := EthBlock{
		Sha3Uncles:       EmptyUncleHash, // Sha3Uncles set to a hardcoded value which is used by some clients to determine if has no uncles.
		StateRoot:        EmptyEthHash,
		TransactionsRoot: EmptyRootHash, // TransactionsRoot set to a hardcoded value which is used by some clients to determine if has no transactions.
		ReceiptsRoot:     EmptyEthHash,
		Difficulty:       EmptyEthInt,
		LogsBloom:        FullEthBloom[:],
		Extradata:        []byte{},
		MixHash:          EmptyEthHash,
		Nonce:            EmptyEthNonce,
		GasLimit:         EthUint64(build.BlockGasLimit * int64(tipsetLen)),
		Uncles:           []EthHash{},
		Transactions:     []interface{}{},
	}
	if hasTransactions {
		b.TransactionsRoot = EmptyEthHash
	}

	return b
}
```

Closes

## Change checklist


src/shim/actors/miner.rs
```diff
@@ -0,0 +1,20 @@
...
+
+mod r#impl;
+
+use fil_actor_interface::miner::State;
+use fil_actors_shared::fvm_ipld_bitfield::BitField;
+use fvm_ipld_blockstore::Blockstore;
+
+use crate::rpc::types::SectorOnChainInfo;
+
+pub trait MinerStateExt {
```
hanabi1224, line 12: I feel it makes things much easier to move `fil_actors_inteface` code into `forest`

LesnyRumcajs: Doesn't it make it inconsistent? I don't have a strong opinion on where this should be, but I'd rather have all this in one place and not scattered randomly.

src/rpc/types/mod.rs
```diff
@@ -260,7 +260,7 @@ pub struct ApiState {
 
 lotus_json_with_self!(ApiState);
 
-#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "PascalCase")]
 pub struct SectorOnChainInfo {
```
LesnyRumcajs, line 265: I assume there's no way of not re-defining this?

hanabi1224: I think so, this is the struct for LotusJson serde.

src/shim/actors/miner.rs
hanabi1224: We could have a vote maybe. I prefer the interface in forest to 
1. avoid publishing `fil_actor_*` crates on every change.
2. avoid the shim in the current `fil_actors_interface` which does not implement `LotusJson` or `JsonSchema` and we have to re-define a same struct in `forest`

LesnyRumcajs: Feel free to propose an initiative to change this. I only would like this to be in the same place to avoid wildly jumping between repos.

elmattic: @hanabi1224 I prefer the actor interface to be in forest as well.

elmattic (approved):

src/rpc/methods/eth.rs
```diff
@@ -407,9 +407,9 @@ pub struct Block {
 }
 
 impl Block {
-    pub fn new(has_transactions: bool) -> Self {
+    pub fn new(has_transactions: bool, tipset_len: usize) -> Self {
```
hanabi1224, line 410: https://github.com/filecoin-project/lotus/pull/11747

LesnyRumcajs (approved):

lemmih (approved):

PR #4291, opened 2024-05-07
fix: update `sync_headers_in_reverse` logic to match lotus

hanabi1224:
## Summary of changes

- Update  `sync_headers_in_reverse` logic to match lotus
- Avoid re-validating the current head when the local chain is not a fork
- Avoid some unnecessary clones

Log(importing a snapshot whose head is not a fork, it no longer re-validates the head):
```
2024-05-07T12:55:50.070707Z  INFO forest_filecoin::chain_sync::chain_muxer: Local node is behind the network, starting BOOTSTRAP from LOCAL_HEAD = 1591906 -> NETWORK_HEAD = 1592005
2024-05-07T12:55:52.146710Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1591907, N blocks = 2
```

lotus code(`func (syncer *Syncer) collectHeaders`): https://github.com/filecoin-project/lotus/blob/master/chain/sync.go#L684
Some highlights:
```go
...
// we want to sync all the blocks until the height above our
// best tipset so far
untilHeight := known.Height() + 1
...
loop:
	for blockSet[len(blockSet)-1].Height() > untilHeight {
             ...
        }
...
if base.IsChildOf(knownParent) {
		// common case: receiving a block that's potentially part of the same tipset as our best block
		return blockSet, nil
	}

	// We have now ascertained that this is *not* a 'fast forward'
	log.Warnf("(fork detected) synced header chain (%s - %d) does not link to our best block (%s - %d)", incoming.Cids(), incoming.Height(), known.Cids(), known.Height())
	fork, err := syncer.syncFork(ctx, base, known, ignoreCheckpoint)
```

Closes https://github.com/ChainSafe/forest/issues/4113

## Change checklist


src/blocks/tipset.rs
```diff
@@ -402,6 +402,16 @@ impl Tipset {
...
     }
+
+    /// Check if `self` is the child of `other`
+    pub fn is_child_of(&self, other: &Self) -> bool {
+        self.parents() == other.key()
+        // See <https://github.com/filecoin-project/lotus/blob/01ec22974942fb7328a1e665704c6cfd75d93372/chain/types/tipset.go#L258>
+        // FIXME: The height check might go beyond what is meant by
+		//  "parent", but many parts of the code rely on the tipset's
+		//  height for their processing logic at the moment to obviate it.
+        && self.epoch() > other.epoch()
```
ruseinov, line 413: q: what you want to say here is that there is just a sanity check, correct? Just an extra measure to make sure the relationship is somewhat correct?

hanabi1224: This is completely copied from lotus. See the link above : )

src/chain_sync/tipset_syncer.rs
```diff
@@ -844,73 +843,75 @@ async fn sync_headers_in_reverse<DB: Blockstore + Sync + Send + 'static>(
...
             tracker.write().set_epoch(tipset.epoch());
             parent_tipsets.push(tipset);
         }
     }
     drop(wp);
 
-    let oldest_tipset = parent_tipsets.last().clone();
-    // Determine if the local chain was a fork.
-    // If it was, then sync the fork tipset range by iteratively walking back
+    let oldest_tipset = parent_tipsets.last();
```
ruseinov, line 858: nit: perhaps `ordest_parent` is a better name, similar to https://github.com/ChainSafe/forest/pull/4291/files#diff-c2e102581e8811dd894188425099a003280f0c7e38b9c5a7ca75f2a0dd0bdaddR824

hanabi1224: Good point, actually `parent` is quite confusing here, renamed `parent_tipsets` to `pending_tipsets(_for_sync)`, and `oldest_tipset` to `oldest_pending_tipset`

src/blocks/tipset.rs
lemmih: Do we have any code that depends on this check? I think we reject blocks that do not have sensible epochs at an earlier stage. I think we can just remove this extra check.

lemmih (approved):
Logic appears to be sound (though still hard to follow).

src/blocks/tipset.rs
hanabi1224: > Do we have any code that depends on this check? I think we reject blocks that do not have sensible epochs at an earlier stage. I think we can just remove this extra check.

@lemmih No other code depends on this check. Removed.

src/blocks/tipset.rs
```diff
@@ -403,6 +403,13 @@ impl Tipset {
         }
         anyhow::bail!("Genesis block not found")
     }
+
+    /// Check if `self` is the child of `other`
+    pub fn is_child_of(&self, other: &Self) -> bool {
+        // Note: the extra `&& self.epoch() > other.epoch()` check in lotus is dropped
+        // See <https://github.com/filecoin-project/lotus/blob/01ec22974942fb7328a1e665704c6cfd75d93372/chain/types/tipset.go#L258>
```
aatifsyed, line 410: excellent comment

src/chain_sync/tipset_syncer.rs
```diff
@@ -807,111 +811,114 @@ async fn sync_headers_in_reverse<DB: Blockstore + Sync + Send + 'static>(
...
-            }
-            if &potential_common_ancestor == fork_tipset {
-                // Remove elements from the vector since the Drain
-                // iterator is immediately dropped
-                let mut fork_tipsets = fork_tipsets;
-                fork_tipsets.drain((i + 1)..);
-                parent_tipsets.extend(fork_tipsets);
-                break;
-            }
+    info!("Fork detected, searching for a common ancestor between the local chain and the network chain");
```
aatifsyed, line 872: nit: this log message is basically the same as the comment above, you can probably delete the comment

aatifsyed (approved):
AIUI this is transpilation of some go code.
If so, WIBNI you checked in a comment to the reference code, so readers can understand why it is the way it is.

src/chain_sync/tipset_syncer.rs
hanabi1224: Fixed.

hanabi1224:
> WIBNI you checked in a comment to the reference code, so readers can understand why it is the way it is.

@aatifsyed link added!

aatifsyed (approved):

PR #4312, opened 2024-05-10
refactor: move some HasLotusJson impls to the right place

aatifsyed:
## Summary of changes

- see title
- Note that the implementations are still out-of-pattern - we're implementing them on _external types_, not shim types, but wrapping `PendingBeneficiaryChange` should be another PR

Closes

## Change checklist


hanabi1224 (approved):

src/lotus_json/miner_power.rs
```diff
@@ -0,0 +1,47 @@
...
+            has_min_power: lotus_json.has_min_power,
+        }
+    }
+}
+
+// MinerPower: !Debug
+// #[test]
+// fn snapshots() {
+//     assert_all_snapshots::<MinerPower>();
+// }
```
LesnyRumcajs, line 47: What is this?

src/lotus_json/mod.rs
```diff
@@ -213,13 +214,17 @@ decl_and_test!(
     vrf_proof for crate::blocks::VRFProof,
 );
 
+mod beneficiary_term; // fil_actor_miner_state::v12::BeneficiaryTerm: !quickcheck::Arbitrary
```
LesnyRumcajs, line 217: Do you find these comments helpful? I feel they can get outdated quickly, e.g., with the next actors' version bump, given there are no compile time checks and the truth is only in the module itself.

ruseinov: Apart from the concern listed above I'm curious what actual purpose those serve. I'm assuming these are just comments carrying information about whether or not a certain type derives `Arbitrary`.

ruseinov:
@aatifsyed I'm pretty sure this needs another pass to address the minor nits.

src/lotus_json/miner_power.rs
aatifsyed: `fn assert_all_snapshots<T: Debug>()` but `MinerPower: !Debug!`, so it can't be tested even though it SHOULD.

so the comment documents the tech debt

src/lotus_json/mod.rs
aatifsyed: I've made the purpose of these comments clearer

aatifsyed: wibni to `static_assertions::assert_not_impl_any` as a road bump, and I can if you feel strongly about this, but there is additional tech debt in that we're using `fil_actor_miner_state` _at all_ - I think these should be shim types. But this PR is only moving the impls, not fixing them

ruseinov (approved):

src/lotus_json/miner_power.rs
ruseinov: Please add an issue for this.

src/lotus_json/mod.rs
ruseinov: We have done some wrapping on the fil-actor-states side for Claim and Allocation. Perhaps we could have something similar here, but not in this PR.

PR #4328, opened 2024-05-15
fix: set tipset sample size to 1

lemmih:
## Summary of changes

- Forest would sample tipset HEADs from 5 peers before fully joining a network. This obviously doesn't work if there are fewer than 5 peers available. This PR temporarily sets the sample size to 1. The chain muxer is scheduled to be completely re-written soon.

Closes

## Change checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,9 @@
 - [#4297](https://github.com/ChainSafe/forest/pull/4297) Fixed double decoding
   of message in the `Filecoin.WalletSign` RPC method.
 
+- [#4328](https://github.com/ChainSafe/forest/pull/4328) Fix issues when
+  connecting to a network with fewer than 5 peers.
+
 ## Forest 0.17.2 "Dovakhin"
 
 This is a **mandatory** release for all mainnet node operators. It changes the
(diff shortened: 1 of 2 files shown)


LesnyRumcajs (commented):
Does it make the node more likely to follow a bad fork or what are the downsides? Perhaps we could hide parameterize it with environmental variable to unblock miner integration but keep the 5 peers requirement?

lemmih:
> Does it make the node more likely to follow a bad fork or what are the downsides? Perhaps we could hide parameterize it with environmental variable to unblock miner integration but keep the 5 peers requirement?

It makes it more likely that we will follow a fork rather than a more common branch, but we should still eventually end up following the heaviest tipset.

IMO, it isn't worth making it configurable. This entire `chain_muxer` code is incredibly fragile, and it's being replaced as soon as possible. The new state machine will not have anything similar to the tipset sample parameter.

LesnyRumcajs (approved):
:pray:

ruseinov:
> IMO, it isn't worth making it configurable. This entire `chain_muxer` code is incredibly fragile, and it's being replaced as soon as possible. The new state machine will not have anything similar to the tipset sample parameter.

Is there a WIP branch with the state machine? :) Looking forward to this new bit of code, `chain_muxer` is quite hard to follow.

PR #4330, opened 2024-05-15
fix: handle optional parameters

aatifsyed:
`LotusJson<Option<T>>` is handled as an optional argument


ruseinov:
nit: Does this fix an encountered issue? Perhaps a simple test is in order to verify that `LotusJson<Option<T>>` encodes/decodes the way we expect. It's a couple of lines of code, but goes a long way, especially if this code is ever touched by somebody else.

ruseinov (approved):

lemmih (approved):

PR #4337, opened 2024-05-16
chore: stateless nodes don't need bundles

LesnyRumcajs:
## Summary of changes

- stateless nodes don't do state migration and don't do validations, and so they don't require actor bundles and proof param files downloaded at runtime.

Closes

## Change checklist


lemmih (approved):

hanabi1224 (approved):
I did not remove these because I was not sure whether some PRC calls would panic forest if the actor bundle and parameter files are not loaded.

PR #4340, opened 2024-05-17
feat(rpc): implement Filecoin.EthGetStorageAt

hanabi1224:
## Summary of changes

-

Closes

## Change checklist

### DIFF
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -31,7 +31,7 @@ use cid::{
     Cid,
 };
 use fvm_ipld_blockstore::Blockstore;
-use fvm_ipld_encoding::{CBOR, DAG_CBOR, IPLD_RAW};
+use fvm_ipld_encoding::{RawBytes, CBOR, DAG_CBOR, IPLD_RAW};
 use itertools::Itertools;
 use keccak_hash::keccak;
 use num_bigint::{self, Sign};
@@ -51,6 +51,7 @@ macro_rules! for_each_method {
         $callback!(crate::rpc::eth::EthBlockNumber);
         $callback!(crate::rpc::eth::EthChainId);
         $callback!(crate::rpc::eth::EthGetCode);
+        $callback!(crate::rpc::eth::EthGetStorageAt);
         $callback!(crate::rpc::eth::EthGasPrice);
         $callback!(crate::rpc::eth::EthGetBalance);
         $callback!(crate::rpc::eth::EthGetBlockByNumber);
@@ -159,15 +160,6 @@ pub struct Int64(
 );
 
 lotus_json_with_self!(Int64);
-
-#[derive(PartialEq, Debug, Deserialize, Serialize, Default, Clone, JsonSchema)]
-pub struct Bytes(
-    #[schemars(with = "String")]
-    #[serde(with = "crate::lotus_json::hexify_vec_bytes")]
-    pub Vec<u8>,
-);
-
-lotus_json_with_self!(Bytes);
 
 #[derive(PartialEq, Debug, Deserialize, Serialize, Default, Clone, JsonSchema)]
 pub struct Hash(#[schemars(with = "String")] pub ethereum_types::H256);
@@ -316,7 +308,7 @@ pub struct Block {
     pub gas_limit: Uint64,
     pub gas_used: Uint64,
     pub timestamp: Uint64,
-    pub extra_data: Bytes,
+    pub extra_data: EthBytes,
     pub mix_hash: Hash,
     pub nonce: Nonce,
     pub base_fee_per_gas: BigInt,
@@ -363,7 +355,7 @@ pub struct Tx {
     pub to: Option<EthAddress>,
     pub value: BigInt,
     pub r#type: Uint64,
-    pub input: Bytes,
+    pub input: EthBytes,
     pub gas: Uint64,
     pub max_fee_per_gas: BigInt,
     pub max_priority_fee_per_gas: BigInt,
@@ -897,7 +889,7 @@ fn eth_tx_from_signed_eth_message(smsg: &SignedMessage, chain_id: u32) -> Result
         v,
         r,
         s,
-        input: Bytes(tx_args.input),
+        input: EthBytes(tx_args.input),
         ..Tx::default()
     })
 }
@@ -949,10 +941,10 @@ fn encode_filecoin_params_as_abi(
     method: MethodNum,
     codec: u64,
     params: &fvm_ipld_encoding::RawBytes,
-) -> Result<Bytes> {
+) -> Result<EthBytes> {
     let mut buffer: Vec<u8> = vec![0x86, 0x8e, 0x10, 0xc4];
     buffer.append(&mut encode_filecoin_returns_as_abi(method, codec, params));
-    Ok(Bytes(buffer))
+    Ok(EthBytes(buffer))
 }
 
 fn encode_filecoin_returns_as_abi(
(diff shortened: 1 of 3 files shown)


sudo-shashank (approved):

src/rpc/methods/eth.rs
```diff
@@ -1338,13 +1330,84 @@ impl RpcMethod<2> for EthGetCode {
...
+        let params = RawBytes::new(GetStorageAtParams::new(position.0)?.serialize_params()?);
+        let message = Message {
+            from: FilecoinAddress::SYSTEM_ACTOR,
+            to: to_address,
+            method_num: METHOD_GET_STORAGE_AT,
+            gas_limit: BLOCK_GAS_LIMIT,
+            params,
+            ..Default::default()
+        };
+        let mut api_invoc_result = None;
```
ruseinov, line 1378: nit: could be done without`mut`.

```
let api_invoice_result = {
  for ts in ... {
     match {
     Ok(res) -> return Some(res)
  }
  None
}
```

src/rpc/methods/eth.rs
```diff
@@ -1338,13 +1330,84 @@ impl RpcMethod<2> for EthGetCode {
...
+        }
+
+        let mut ret = fvm_ipld_encoding::from_slice::<RawBytes>(msg_rct.return_data().as_slice())?
+            .bytes()
+            .to_vec();
+        if ret.len() < 32 {
+            let mut with_padding = vec![0; 32_usize.saturating_sub(ret.len())];
+            with_padding.append(&mut ret);
+            Ok(EthBytes(with_padding))
+        } else {
```
ruseinov, line 1405: nit: return early on `if`, then there is no need for `else`.

src/rpc/methods/eth.rs
```diff
@@ -1338,13 +1330,84 @@ impl RpcMethod<2> for EthGetCode {
...
+        for ts in ts.chain_arc(ctx.store()) {
+            match ctx.state_manager.call(&message, Some(ts)) {
+                Ok(res) => {
+                    api_invoc_result = Some(res);
+                    break;
+                }
+                Err(e) => tracing::warn!(%e),
+            }
+        }
+        let Some(api_invoc_result) = api_invoc_result else {
```
ruseinov, line 1388: Looking at this more I'm thinking: why do we need `Option` for `api_invoc_result`? It would be much easier to use `Result`. Then there is no need for `let Some..`

Something like:

```
  for ... {
    match {
       Ok(res) -> return Ok(res)
    } 
  }
  return Err(err)

  and then just do `api_invoc_result?`
  
```

src/rpc/methods/eth.rs
```diff
@@ -1338,13 +1330,84 @@ impl RpcMethod<2> for EthGetCode {
...
+            return Err(anyhow::anyhow!("no message receipt").into());
+        };
+        let Some(msg_rct) = api_invoc_result.msg_rct else {
+            return Err(anyhow::anyhow!("no message receipt").into());
+        };
+        if !api_invoc_result.error.is_empty() {
```
ruseinov, line 1394: It is actually possible for `!api_invoc_result.error.is_empty` when `msg_rct` is `Some`?

src/rpc/methods/eth.rs
hanabi1224: The `return` clause works for the function instead of the scope unless the scope is converted to a lambda. I was thinking about `break Some(res);` syntax just like in some other languages but it does not work in rust

src/rpc/methods/eth.rs
hanabi1224: Personally, I feel using early returns here is not Rustic given rust is expression-based

src/rpc/methods/eth.rs
hanabi1224: Logically yes, in practice, I'm not sure. I think it does not hurt to check both and make the logic look more robust

src/rpc/methods/eth.rs
hanabi1224: I don't think the `return` clause works this way. https://github.com/ChainSafe/forest/pull/4340#discussion_r1615529393

src/rpc/methods/eth.rs
elmattic: If you want to avoid `mut` here, you still could use labelled blocks, e.g.:

```
let api_invoc_result = 'block: {
    for ts in ts.chain_arc(ctx.store()) {
        match ctx.state_manager.call(&message, Some(ts)) {
            Ok(res) => {
                break 'block res;
            }
            Err(e) => tracing::warn!(%e),
        }
    }
    return Err(anyhow::anyhow!("Call failed").into());
};
```

src/rpc/methods/eth.rs
```diff
@@ -1338,13 +1330,84 @@ impl RpcMethod<2> for EthGetCode {
...
+            return Err(anyhow::anyhow!("GetBytecode failed: {}", api_invoc_result.error).into());
+        }
+
+        let mut ret = fvm_ipld_encoding::from_slice::<RawBytes>(msg_rct.return_data().as_slice())?
+            .bytes()
+            .to_vec();
+        if ret.len() < 32 {
```
elmattic, line 1388: If you mean to pad on an EVM word boundary, you can use the `EVM_WORD_LENGTH` constant.

src/rpc/methods/eth.rs
```diff
@@ -1338,13 +1330,84 @@ impl RpcMethod<2> for EthGetCode {
...
+    const PERMISSION: Permission = Permission::Read;
+
+    type Params = (EthAddress, EthBytes, BlockNumberOrHash);
+    type Ok = EthBytes;
+
+    async fn handle(
+        ctx: Ctx<impl Blockstore + Send + Sync + 'static>,
+        (eth_address, position, block_number_or_hash): Self::Params,
+    ) -> Result<Self::Ok, ServerError> {
+        let make_empty_result = || EthBytes(vec![0; 32]);
```
elmattic, line 1337: If you mean to return a zeroed EVM word, you could use the `EVM_WORD_LENGTH` constant.

src/rpc/methods/eth.rs
hanabi1224: @elmattic Didn't know about this. This is great, thanks!

src/rpc/methods/eth.rs
hanabi1224: Fixed.

src/rpc/methods/eth.rs
hanabi1224: Fixed.

src/rpc/methods/eth.rs
hanabi1224: Fixed.

src/rpc/methods/eth.rs
ruseinov: I guess it boils down to personal preference, however `if .. else` nesting here does not read very well.
Imagine if every if in this method would have an else. Non-blocking.

ruseinov (approved):

src/rpc/methods/eth.rs
```diff
@@ -1299,18 +1290,17 @@ impl RpcMethod<2> for EthGetCode {
...
-            match ctx.state_manager.call(&message, Some(ts)) {
-                Ok(res) => {
-                    api_invoc_result = Some(res);
-                    break;
+
+        let api_invoc_result = 'invoc: {
+            for ts in ts.chain_arc(ctx.store()) {
+                match ctx.state_manager.call(&message, Some(ts)) {
+                    Ok(res) => {
+                        break 'invoc res;
```
ruseinov, line 1298: The loop label here is unnecessary. Non-blocking.

ruseinov (approved):

src/rpc/methods/eth.rs
hanabi1224: Could you elaborate? I can compile the code without the loop label

```
error[E0571]: `break` with value from a `for` loop
    --> src/rpc/methods/eth.rs:1298:25
     |
1295 |             for ts in ts.chain_arc(ctx.store()) {
     |             ----------------------------------- you can't `break` with a value in a `for` loop
...
1298 |                         break res;
     |                         ^^^^^^^^^ can only break with a value inside `loop` or breakable block
     |
help: use `break` on its own without a value inside this `for` loop
     |
1298 |                         break;
     |                         ~~~~~
```

PR #4352, opened 2024-05-21
[Feature] StateGetClaim RPC endpoint.

### LINKED ISSUE #4336: StateGetClaim

ruseinov:
## Summary of changes

- Introduces `Filecoin.StateGetClaim` RPC endpoint

Closes #4336 

## Change checklist


src/rpc/types/mod.rs
```diff
@@ -164,6 +166,35 @@ pub struct ApiTipsetKey(pub Option<TipsetKey>);
...
+    #[schemars(with = "u64")]
+    pub term_min: ChainEpoch,
+    // The max period after term_start for which provider can earn QA-power for the data
+    #[schemars(with = "u64")]
+    pub term_max: ChainEpoch,
+    // The epoch at which the (first range of the) piece was committed.
+    #[schemars(with = "u64")]
+    pub term_start: ChainEpoch,
+    // ID of the provider's sector in which the data is committed.
+    #[schemars(with = "u64")]
```
LesnyRumcajs, line 195: Why is it required to put the type `#[schemars(with = "u64")]` everywhere? What's the behaviour, and reason, when the types don't match? For example, in `ChainEpoch` you put `u64` whereas it's `i64` underneath. What will happen for values like `-1`?

ruseinov: It needs to be fixed I’m pretty sure.

ruseinov: I've taken a look at this - it's not needed at all, except for enums with `repr(u64)`. I'll remove all the ocurrences.

ruseinov: done

ruseinov: Aha, it was needed for `u64` wrappers, like `PaddedPieceSize`.

ruseinov: Basically the tactic here is as follows - try compiling without any attributes, see what breaks, fix that.
This whole schemars thing is a bit confusing until one gets used to it.

src/lotus_json/mod.rs
```diff
@@ -215,14 +215,15 @@ decl_and_test!(
...
 mod hash_map;
 mod ipld; // NaN != NaN
 mod nonempty;
 mod opt; // can't make snapshots of generic type
+mod power_claim; // fil_actor_interface::power::Claim: !quickcheck::Arbitrary
 mod raw_bytes; // fvm_ipld_encoding::RawBytes: !quickcheck::Arbitrary
 mod receipt; // shim type roundtrip is wrong - see module
-mod vec; // can't make snapshots of generic type
+mod vec;
+mod verifreg_claim; // can't make snapshots of generic type
```
lemmih, line 226: The comment is for the `vec` module, right?
```suggestion
mod vec; // can't make snapshots of generic type
mod verifreg_claim;
```

src/tool/subcommands/api_cmd.rs
```diff
@@ -831,6 +831,11 @@ fn state_tests_with_tipset<DB: Blockstore>(
                 block.miner_address,
                 tipset.key().into(),
             ))?),
+            RpcTest::identity(StateGetClaim::request((
+                block.miner_address,
+                0,
```
lemmih, line 836: Is `0` usually a valid claim id or will this test always return `null`?

src/lotus_json/mod.rs
ruseinov: yes, I have already fixed that, but somehow it got back..

src/tool/subcommands/api_cmd.rs
ruseinov: We actually have to implement `StateGetClaims` to be able to test that one. I don't think there's another good way to supply a valid claim_id, as it depends on the miner. I will add a comment.

lemmih (approved):

LesnyRumcajs (approved):

Issue #4355, opened 2024-05-22
Investigate connection limiting and trimming in rust lib-p2p.


ruseinov:
## Issue summary

This came up as part of `NetProtectAdd` RPC endpoint implementation, which prevents certain peers from being disconnected, unless the process is low on memory. This is tightly coupled with go-libp2p connection manager implementation.

See https://github.com/filecoin-project/lotus/blob/2086b219d2828cc2bd4c943b0b792e61ab6321f9/node/impl/net/protect.go#L11-L17 for details.

lemmih:
We've looked into libp2p connection limits in the past and decided to hold off until we have more experience with running Forest.

PR #4356, opened 2024-05-22
[Feature] NetProtectAdd RPC.

### LINKED ISSUE #4353: NetProtectAdd

ruseinov:
## Summary of changes

- introduced NetProtectAdd RPC endpoint

Closes #4353 

Perhaps needs a follow-up #4355 

## Change checklist

### DIFF
--- a/src/rpc/methods/net.rs
+++ b/src/rpc/methods/net.rs
@@ -247,7 +247,7 @@ impl RpcMethod<1> for NetProtectAdd {
     type Ok = ();
 
     // This is a no-op due to the fact that `rust-libp2p` implementation is very different to that
-    // in go. However it would be nice to investigate connection limiting options in rust.
+    // in go. However it would be nice to investigate connection limiting options in Rust.
     // See: <https://github.com/ChainSafe/forest/issues/4355>.
     async fn handle(
         _: Ctx<impl Blockstore>,
--- a/src/rpc/methods/net.rs
+++ b/src/rpc/methods/net.rs
@@ -246,7 +246,7 @@ impl RpcMethod<1> for NetProtectAdd {
     type Params = (String,);
     type Ok = ();
 
-    // This is a no-op due to the fact that rust `lib-p2p` implementation is very different to that
+    // This is a no-op due to the fact that `rust-libp2p` implementation is very different to that
     // in go. However it would be nice to investigate connection limiting options in rust.
     // See: <https://github.com/ChainSafe/forest/issues/4355>.
     async fn handle(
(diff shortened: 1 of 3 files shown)


lemmih (approved):

src/rpc/methods/net.rs
```diff
@@ -236,6 +236,28 @@ impl RpcMethod<0> for NetVersion {
...
+impl RpcMethod<1> for NetProtectAdd {
+    const NAME: &'static str = "Filecoin.NetProtectAdd";
+    const PARAM_NAMES: [&'static str; 1] = ["acl"];
+    const API_VERSION: ApiVersion = ApiVersion::V1;
+    const PERMISSION: Permission = Permission::Admin;
+
+    type Params = (String,);
+    type Ok = ();
+
+    // This is a no-op due to the fact that rust `lib-p2p` implementation is very different to that
```
LesnyRumcajs, line 249:
```suggestion
    // This is a no-op due to the fact that `rust-libp2p` implementation is very different to that
```

src/rpc/methods/net.rs
```diff
@@ -236,6 +236,28 @@ impl RpcMethod<0> for NetVersion {
...
+    // in go. However it would be nice to investigate connection limiting options in rust.
```
LesnyRumcajs, line 250: Respect 🦀 
```suggestion
    // in go. However it would be nice to investigate connection limiting options in Rust.
```

src/rpc/methods/net.rs
```diff
@@ -236,6 +236,28 @@ impl RpcMethod<0> for NetVersion {
...
+    // See: <https://github.com/ChainSafe/forest/issues/4355>.
+    async fn handle(
+        _: Ctx<impl Blockstore>,
+        (peer_id,): Self::Params,
+    ) -> Result<Self::Ok, ServerError> {
+        let _ = PeerId::from_str(&peer_id)?;
```
LesnyRumcajs, line 256: Can we do nothing?

LesnyRumcajs (commented):
Can we add a test to cover this method in the report?

src/rpc/methods/net.rs
ruseinov: I don't understand. We need to make sure that `PeerId` is valid. And this is the same way it is done for other endpoints that parse `PeerId`.

LesnyRumcajs: I mean, this method is a no-op, so adding a check may bring incorrect assumptions that we do _something_ with this peer. It's not a blocker.

ruseinov: It has to line up with Lotus impl and that will throw an error when the peer_id does not parse.

LesnyRumcajs (approved):

PR #4367, opened 2024-05-27
fix(p2p): max concurrent streams in request_response protocol

hanabi1224:
## Summary of changes

This PR tries to fix the below errors when running `forest --chain mainnet --encrypt-keystore false --save-token /tmp/forest_token --stateless --target-peer-count=5000`

```
2024-05-27T07:46:02.738459Z  WARN libp2p_request_response::handler: Dropping outbound stream because we are at capacity
...
```

The error occurs when the target peer count is large thus the default `max_concurrent_streams`(100) become insufficient

- set `max_concurrent_streams` to `target_peer_count  * 10`

Closes

## Change checklist

### DIFF
--- a/src/libp2p/behaviour.rs
+++ b/src/libp2p/behaviour.rs
@@ -13,7 +13,7 @@ use libp2p::{
     identity::{Keypair, PeerId},
     kad::QueryId,
     metrics::{Metrics, Recorder},
-    ping,
+    ping, request_response,
     swarm::NetworkBehaviour,
     Multiaddr,
 };
@@ -66,6 +66,12 @@ impl ForestBehaviour {
         config: &Libp2pConfig,
         network_name: &str,
     ) -> anyhow::Result<Self> {
+        const MAX_ESTABLISHED_PER_PEER: u32 = 4;
+        const MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER: usize = 10;
+
+        let max_concurrent_request_response_streams = (config.target_peer_count as usize)
+            .saturating_mul(MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER);
+
         let mut gs_config_builder = gossipsub::ConfigBuilder::default();
         gs_config_builder.max_transmit_size(1 << 20);
         gs_config_builder.validation_mode(ValidationMode::Strict);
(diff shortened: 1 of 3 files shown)


LesnyRumcajs (approved):

LesnyRumcajs:
Would it make sense to make it more configurable via env variables or not really?

hanabi1224:
> Would it make sense to make it more configurable via env variables or not really?

@LesnyRumcajs Environment variable `FOREST_MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER` added.

src/libp2p/behaviour.rs
```diff
@@ -67,10 +68,23 @@ impl ForestBehaviour {
...
+        static MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER: Lazy<usize> = Lazy::new(|| {
+            std::env::var("FOREST_MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER")
+                .ok()
+                .and_then(|it| {
+                    if let Ok(n) = it.parse() {
+                        if n > 0 {
+                            return Some(n);
+                        }
+                    }
+                    tracing::warn!("Failed to parse the `FOREST_MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER` environment variable value, a positive integer expected, got {it}.");
```
LesnyRumcajs, line 80: I think we should just panic if it's not parsable. Warnings are easily overlooked, and lousy input should be punished.

hanabi1224: Fixed.

LesnyRumcajs: Do we need the extra logic? Parsing a negative number to unsigned should fail. So we'd need something along the lines of `std::env::var("FOREST_MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER").ok().map(|v| v.parse().expect("silly user")).unwrap_or(10)`. Am I missing something?

hanabi1224: 0 is invalid in this case

LesnyRumcajs: Use a `NonZero`, then! :)

```rust
    std::env::var("FOREST_MAX_CONCURRENT_REQUEST_RESPONSE_STREAMS_PER_PEER")
        .ok()
        .map(|v| {
            v.parse::<NonZeroUsize>()
                .expect("Must be a positive number")
                .get()
        })
        .unwrap_or(10);

```

hanabi1224: Fixed.

sudo-shashank (approved):

Issue #4346, opened 2024-05-20
Stateless node bad performance


LesnyRumcajs:
## Issue summary

We've been running a stateless node over the weekend (May 18th - May 20th). The results are not satisfactory, given the overall connectivity and resource usage.

1. Connectivity looks poor looking at the [bootstrapper dashboard.](https://probelab.io/bootstrappers/filecoin) - Forest node is `/dns4/bootstrap-mainnet-0.chainsafe-fil.io/tcp/34000/p2p/12D3KooWKKkCZbcigsWTEu1cgNetNbZJqeNtysRtFpq7DTqw3eqH`

![image](https://github.com/ChainSafe/forest/assets/9642092/8ad19e55-a1b7-49be-a8ce-f09c7b47f166)

2. Resource usage is high:
![image](https://github.com/ChainSafe/forest/assets/9642092/7d8b834c-36cf-4eb0-83a1-b75acedf97da)

We expected the resource usage for a stateless node to be around ~500 MiB at most, but it's higher. Looking at the trend, there may be a memory leak somewhere. The CPU usage is also high - higher than in full-node Lotus.

3. Lots of warnings in logs (which may contribute to resource usage, but also indicate a genuine issue on the p2p layer of Forest)

```
2024-05-20 10:16:04.127	
2024-05-20T08:16:04.127165Z  WARN forest_filecoin::libp2p_bitswap::request_manager: sending on a closed channel
2024-05-20 10:16:04.126	
2024-05-20T08:16:04.126292Z  WARN forest_filecoin::libp2p_bitswap::request_manager: sending on a closed channel
2024-05-20 10:15:40.039	
2024-05-20T08:15:40.039723Z  WARN libp2p_request_response::handler: Dropping inbound stream because we are at capacity
2024-05-20 10:15:35.344	
2024-05-20T08:15:35.344432Z  WARN forest_filecoin::libp2p::service: Ping timeout: 12D3KooWJsKmHDxvbH5h22Qy8k3hFVHdQ3hnodJsPWqocFnoNhMF
2024-05-20 10:15:34.770	
2024-05-20T08:15:34.770194Z  WARN forest_filecoin::libp2p::service: Ping timeout: 12D3KooWPrxMyMKyb9Zn6WwWjtwFucYJBsCMYHpvLuhutktjDfRK
2024-05-20 10:14:53.679	
[...]

PR #4372, opened 2024-05-28
fix(p2p): ban peers that don't support hello and chain exchange

hanabi1224:
## Summary of changes

(Part of https://github.com/ChainSafe/forest/issues/4346 investigation)

In the previous logic, we mark peers bad when they don't support `Hello` and `ChainExchange` protocols, but it does not make sense to hold connections to those peers and broadcast their addresses via `Kademlia` in the network.

- Ban peers that don't support `Hello` and `ChainExchange` protocols

Closes

## Change checklist

### DIFF
--- a/src/libp2p/peer_manager.rs
+++ b/src/libp2p/peer_manager.rs
@@ -212,7 +212,7 @@ impl PeerManager {
 
     /// Removes a peer from the set and returns true if the value was present
     /// previously
-    pub fn mark_peer_bad(&self, peer_id: PeerId) -> bool {
+    pub fn mark_peer_bad(&self, peer_id: PeerId, reason: impl Into<String>) -> bool {
         let mut peers = self.peers.write();
         let removed = remove_peer(&mut peers, &peer_id);
         if removed {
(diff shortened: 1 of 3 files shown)


src/libp2p/peer_manager.rs
```diff
@@ -212,15 +212,16 @@ impl PeerManager {
 
     /// Removes a peer from the set and returns true if the value was present
     /// previously
-    pub fn mark_peer_bad(&self, peer_id: PeerId) -> bool {
+    pub fn mark_peer_bad(&self, peer_id: PeerId, reason: impl Into<String>) -> bool {
```
LesnyRumcajs, line 215: An idea we can revisit (or not) later - would it make sense to add a label to the bad peers metric so that we see the reason they were banned? This would help with potential debugging when there are a lot of banned peers.

LesnyRumcajs (approved):

src/libp2p/peer_manager.rs
hanabi1224: That's a good point. I can implement this in a separate PR

PR #4039, opened 2024-03-07
feat(rpc): implement Filecoin.NetAutoNatStatus

hanabi1224:
## Summary of changes

Changes introduced in this pull request:

- implement `Filecoin.NetAutoNatStatus`

```console
     Running `target/debug/forest-tool api compare --filter NetAuto`
| RPC Method                | Forest | Lotus |
|---------------------------|--------|-------|
| Filecoin.NetAutoNatStatus | Valid  | Valid |
```

## Reference issue to close (if applicable)

Closes

## Change checklist

- [x] I have performed a self-review of my own code,
- [x] I have made corresponding changes to the documentation,
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md

PR #4376, opened 2024-05-29
fix(p2p): refine chain exchange peer selection logic

hanabi1224:
## Summary of changes

(Part of https://github.com/ChainSafe/forest/issues/4346 investigation)

-

Closes

## Change checklist

### DIFF
--- a/src/libp2p/metrics.rs
+++ b/src/libp2p/metrics.rs
@@ -1,8 +1,12 @@
 // Copyright 2019-2024 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use libp2p::PeerId;
 use once_cell::sync::Lazy;
-use prometheus_client::metrics::{counter::Counter, gauge::Gauge};
+use prometheus_client::{
+    encoding::{EncodeLabelKey, EncodeLabelSet, EncodeLabelValue, LabelSetEncoder},
+    metrics::{counter::Counter, family::Family, gauge::Gauge},
+};
 
 pub static PEER_FAILURE_TOTAL: Lazy<Counter> = Lazy::new(|| {
     let metric = Counter::default();
@@ -33,3 +37,33 @@ pub static BAD_PEERS: Lazy<Gauge> = Lazy::new(|| {
     );
     metric
 });
+
+pub static PEER_TIPSET_EPOCH: Lazy<Family<PeerLabel, Gauge>> = Lazy::new(|| {
+    let metric = Family::default();
+    crate::metrics::default_registry().register(
+        "peer_tipset_epoch",
+        "peer tipset epoch",
+        metric.clone(),
+    );
+    metric
+});
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct PeerLabel(PeerId);
+
+impl PeerLabel {
+    pub const fn new(peer: PeerId) -> Self {
+        Self(peer)
+    }
+}
+
+impl EncodeLabelSet for PeerLabel {
+    fn encode(&self, mut encoder: LabelSetEncoder) -> Result<(), std::fmt::Error> {
+        let mut label_encoder = encoder.encode_label();
+        let mut label_key_encoder = label_encoder.encode_label_key()?;
+        EncodeLabelKey::encode(&"PEER", &mut label_key_encoder)?;
+        let mut label_value_encoder = label_key_encoder.encode_label_value()?;
+        EncodeLabelValue::encode(&self.0.to_string(), &mut label_value_encoder)?;
+        label_value_encoder.finish()
+    }
+}
--- a/src/libp2p/peer_manager.rs
+++ b/src/libp2p/peer_manager.rs
@@ -7,7 +7,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use crate::blocks::Tipset;
+use crate::{blocks::Tipset, shim::clock::ChainEpoch};
 use ahash::{HashMap, HashSet};
 use flume::{Receiver, Sender};
 use parking_lot::RwLock;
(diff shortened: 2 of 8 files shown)


lemmih (approved):

aatifsyed (approved):

src/libp2p/metrics.rs
```diff
@@ -33,3 +37,33 @@ pub static BAD_PEERS: Lazy<Gauge> = Lazy::new(|| {
...
+    pub const fn new(peer: PeerId) -> Self {
+        Self(peer)
+    }
+}
+
+impl EncodeLabelSet for PeerLabel {
+    fn encode(&self, mut encoder: LabelSetEncoder) -> Result<(), std::fmt::Error> {
+        let mut label_encoder = encoder.encode_label();
+        let mut label_key_encoder = label_encoder.encode_label_key()?;
+        EncodeLabelKey::encode(&"PEER", &mut label_key_encoder)?;
```
ruseinov, line 64: can this not be:

``` "PEER".encode(&mut label_key_encoder)```?

File: src/libp2p/metrics.rs [hunk 2]
```diff
@@ -33,3 +37,33 @@ pub static BAD_PEERS: Lazy<Gauge> = Lazy::new(|| {
...
+        let mut label_value_encoder = label_key_encoder.encode_label_value()?;
+        EncodeLabelValue::encode(&self.0.to_string(), &mut label_value_encoder)?;
```
ruseinov, line 66: same as above, it's a bit confusing and there's totally a trait implementation for a variety of string flavours.

ruseinov: Ah, I can see now that this is super ambiguous. A pity. 

Still, you don't need to pass this by reference, this should be enough as there is an impl for "String":
```
EncodeLabelValue::encode(self.0.to_string(), &mut label_value_encoder)?;
```

src/libp2p/peer_manager.rs
```diff
@@ -127,22 +127,27 @@ impl PeerManager {
...
-                let cost = if (info.successes + info.failures) > 0 {
-                    // Calculate cost based on fail rate and latency
-                    let fail_rate = f64::from(info.failures) / f64::from(info.successes);
-                    info.average_time.as_secs_f64() + fail_rate * average_time.as_secs_f64()
+            .filter_map(|(p, info)| {
+                // Filter out nodes that are stateless (or far behind)
+                if info.head.epoch() > 0 {
+                    let cost = if (info.successes + info.failures) > 0 {
+                        // Calculate cost based on fail rate and latency
+                        let fail_rate = f64::from(info.failures) / f64::from(info.successes);
```
ruseinov, line 135: What happens if `info.successes` is `0.0`?

src/libp2p/metrics.rs
ruseinov: It can't, but it seems we can do `EncodeLabelKey::encode("PEER", &mut label_key_encoder)?;`

ruseinov:
@hanabi1224 I removed this from queue so that you get the time to read my comments. If you think it's good as is - go forward and merge it.

src/libp2p/metrics.rs
hanabi1224: It does not work.
```
EncodeLabelKey::encode("libp2p_message_kind", &mut label_key_encoder)?;
    |         ---------------------- ^^^^^^^^^^^^^^^^^^^^^ the trait `prometheus_client::encoding::EncodeLabelKey` is not implemented for `str`
    |         |
    |         required by a bound introduced by this call
```

src/libp2p/metrics.rs
hanabi1224: Same, this does not compile

```
error[E0308]: mismatched types
   --> src/libp2p/metrics.rs:66:34
    |
66  |         EncodeLabelValue::encode(self.0.to_string(), &mut label_value_encoder)?;
    |         ------------------------ ^^^^^^^^^^^^^^^^^^ expected `&_`, found `String`
    |         |
    |         arguments to this function are incorrect
    |
    = note: expected reference `&_`
```

src/libp2p/metrics.rs
hanabi1224: A side note, I don't think ` "PEER".encode(&mut label_key_encoder)` would improve readability when the trait is not well-known. The idea is similar to why we favor `Arc::clone(t)` over `t.clone` in the code base, see https://github.com/ChainSafe/forest/pull/4039#issuecomment-1983686248

> Using Arc::clone makes it explicit that you want to clone the Arc and not the underlying object. It makes the code easier to understand and harder to make mistakes.

src/libp2p/peer_manager.rs
hanabi1224: Good question, I only moved this piece of code around without inspecting it. I think `if info.success > 0` should be a proper fix

src/libp2p/metrics.rs
hanabi1224: Although the trait is implemented for `String`, the trait methods takes a borrowed self
fn [encode](https://docs.rs/prometheus-client/latest/prometheus_client/encoding/trait.EncodeLabelKey.html#tymethod.encode)(&self, encoder: &mut [LabelKeyEncoder](https://docs.rs/prometheus-client/latest/prometheus_client/encoding/struct.LabelKeyEncoder.html)<'_>) -> [Result](https://doc.rust-lang.org/nightly/core/result/enum.Result.html)<[()](https://doc.rust-lang.org/nightly/std/primitive.unit.html), [Error](https://doc.rust-lang.org/nightly/core/fmt/struct.Error.html)>;

ruseinov: Yeah, what's the point of those implementations is unclear then.

ruseinov (approved):

src/libp2p/metrics.rs
ruseinov: I can see now. This is pretty confusing looking at the code that's within those `encode` implementations. 
The way it's currently written is it wants to boil down to  `&str` impl, which is where it actually does ` encoder.write_str(self)?;`.

PR #4381, opened 2024-06-03
feat(rpc): implement Filecoin.StateSectorPartition

### LINKED ISSUE #4380: StateSectorPartition

hanabi1224:
## Summary of changes

- Implement RPC method `Filecoin.StateSectorPartition`
- API compare tests

```
     Running `target/quick/forest-tool api compare /home/me/fr/snapshots/calibnet/forest_snapshot_calibnet_2024-06-03_height_1668748.forest.car.zst --filter StateSectorPartition`
| RPC Method                          | Forest | Lotus |
|-------------------------------------|--------|-------|
| Filecoin.StateSectorPartition (145) | Valid  | Valid |
```

Closes #4380

## Change checklist


src/shim/actors/convert.rs
```diff
@@ -0,0 +1,276 @@
+// Copyright 2019-2024 ChainSafe Systems
```
LesnyRumcajs, line 1: Could you please explain why you copied https://github.com/ChainSafe/fil-actor-states/blob/main/fil_actor_interface/src/convert.rs here? If you intend to move it out of `fil-actor-states`, let's have a separate PR for this.

hanabi1224: These methods are not publicly exported.

src/shim/actors/miner/state.rs
```diff
@@ -114,4 +118,33 @@ impl MinerStateExt for State {
             }
         }
     }
+
+    fn find_sector<BS: Blockstore>(
```
ruseinov, line 122: I would expect this method to be a part of `fil-actor-states` just like we do it for other actors. Then we don't need all of those `from...*` method in `forest` either.

ruseinov: Here's a PR for inspiration: https://github.com/ChainSafe/fil-actor-states/pull/299

Basically we could do the same with `find_sector` and then just call that in `Forest` rpc code.

hanabi1224: I'd prefer getting rid of the process of updating and releasing `fil-actor-states` crates during the RPC development and we can backport all missing functions all at once after all RPC work is done. What do you think?

src/shim/actors/convert.rs
LesnyRumcajs: So right now we have two lengthy implementations doing basically the same thing. I'd rather we have only one.

src/shim/actors/miner/state.rs
hanabi1224: `find_sector` added in https://github.com/ChainSafe/fil-actor-states/pull/303

> I'd rather go through an ordeal of having to reference a working branch in fil-actor-states 

Referencing a working branch without publishing would block `cargo publish --dry-run` for forest.

> introducing these helpers and then moving them to another repo.

I think the ordeal is because with the current development process of `fil_actor_interface`, there's no way to make the crate carefully designed, when the requirement of what are needed and what are not needed is unclear, and the rest of the crates in the repo are mostly copy-pasted code. And I really doubt the usefulness of `fil_actor_interface` after spending so much maintenance effort when it's not (and not likely possible to be ) a carefully design library.

src/shim/actors/convert.rs
hanabi1224: Fixed.

LesnyRumcajs (approved):

src/shim/actors/convert.rs
LesnyRumcajs: Thank you!

LesnyRumcajs:
Let's add changelog entry.

ruseinov (approved):

PR #4389, opened 2024-06-05
fix(log): demote log levels on peer banning

hanabi1224:
## Summary of changes

As [discussed on slack](https://chainsafeio.slack.com/archives/C0144LHN732/p1717501294343469)

- demote log levels on peer banning from `warn` to `debug`
- demote verbose debug logs to trace to make output more readable when `RUST_LOG=forest_filecoin=debug`

Closes

## Change checklist

### DIFF
--- a/scripts/devnet/docker-compose.yml
+++ b/scripts/devnet/docker-compose.yml
@@ -176,7 +176,7 @@ services:
       - ./forest_config.toml.tpl:/forest/forest_config.toml.tpl
     environment:
       - FIL_PROOFS_PARAMETER_CACHE=${FIL_PROOFS_PARAMETER_CACHE}
-      - RUST_LOG=info,forest_filecoin::blocks::header=debug
+      - RUST_LOG=info,forest_filecoin::blocks::header=trace
       - FOREST_GENESIS_NETWORK_VERSION=${GENESIS_NETWORK_VERSION}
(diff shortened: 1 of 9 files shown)


ruseinov:
It would actually be great to address the root cause for `warn` noise as opposed to demoting the log level. Let's at least create an issue to investigate this.

ruseinov (approved):

hanabi1224:
> It would actually be great to address the root cause for `warn` noise as opposed to demoting the log level. Let's at least create an issue to investigate this.

@ruseinov Good idea, my take is that the warn noises are warnings that only make sense to developers instead of end users. I'm not sure that's 100% what you mean, so would you mind creating the issue and elaborate your idea?

ruseinov:
> > It would actually be great to address the root cause for `warn` noise as opposed to demoting the log level. Let's at least create an issue to investigate this.
> 
> @ruseinov Good idea, my take is that the warn noises are warnings that only make sense to developers instead of end users. I'm not sure that's 100% what you mean, so would you mind creating the issue and elaborate your idea?

I'm unsure if warns on startup are a side-effect of something else. I've seen @LesnyRumcajs mentioning that in Slack, perhaps he'll be best equipped to elaborate more as I'm not very familiar with the problem myself.

PR #4400, opened 2024-06-07
Add support for Eth mappings

elmattic:
## Summary of changes

- Introduce a new ParityDB namespace for storing Ethereum mappings such as `Hash` to `TipsetKey` and `Hash` to message `Cid`.
  This is required by methods like:
  - EthGetBlockByHash
  - EthGetBlockTransactionCountByHash
  - EthGetMessageCidByTransactionHash
- Increment Forest version to `0.19.0`

Closes

## Change checklist

### DIFF
--- a/src/daemon/db_util.rs
+++ b/src/daemon/db_util.rs
@@ -158,6 +158,8 @@ where
 {
     let mut delegated_messages = vec![];
 
+    info!("Populating column EthMappings");
+
     for ts in head_ts
         .clone()
         .chain(&state_manager.chain_store().blockstore())
--- a/src/daemon/db_util.rs
+++ b/src/daemon/db_util.rs
@@ -28,6 +28,12 @@ use tracing::{debug, info};
 use url::Url;
 use walkdir::WalkDir;
 
+#[cfg(doc)]
+use crate::rpc::eth::Hash;
+
+#[cfg(doc)]
+use crate::blocks::TipsetKey;
+
 pub fn load_all_forest_cars<T>(store: &ManyCar<T>, forest_car_db_dir: &Path) -> anyhow::Result<()> {
     if !forest_car_db_dir.is_dir() {
         fs::create_dir_all(forest_car_db_dir)?;
@@ -149,6 +155,12 @@ async fn transcode_into_forest_car(from: &Path, to: &Path) -> anyhow::Result<()>
     Ok(())
 }
 
+/// For the need for Ethereum RPC API, a new column in parity-db has been introduced to handle
+/// mapping of:
+/// - [`Hash`] to [`TipsetKey`].
+/// - [`Hash`] to delegated message [`Cid`].
+///
+/// This function traverses the chain store and populates the column.
 pub fn populate_eth_mappings<DB>(
     state_manager: &StateManager<DB>,
     head_ts: &Tipset,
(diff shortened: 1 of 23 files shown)


aatifsyed (approved):

src/daemon/db_util.rs
```diff
@@ -142,6 +149,96 @@ async fn transcode_into_forest_car(from: &Path, to: &Path) -> anyhow::Result<()>
     Ok(())
 }
 
+pub fn populate_eth_mappings<DB>(
```
lemmih, line 152: Could you describe what we're mapping and how many mappings we expect in a comment here?

src/daemon/db_util.rs
```diff
@@ -142,6 +149,96 @@ async fn transcode_into_forest_car(from: &Path, to: &Path) -> anyhow::Result<()>
...
+            }
+        }
+        state_manager.chain_store().put_tipset_key(ts.key())?;
+    }
+    let _ = process_signed_messages(state_manager, &delegated_messages)?;
+
+    Ok(())
+}
+
+fn process_signed_messages<DB>(
```
lemmih, line 188: How are the signed messages being processed? Describe in a comment.

src/db/migration/migration_map.rs
```diff
@@ -86,20 +89,26 @@ pub struct Migration {
...
-        let migrated_db = self.migrator.migrate(chain_data_path)?;
+        let migrated_db = self.migrator.migrate(chain_data_path, config)?;
         self.post_checks(chain_data_path)?;
 
         let new_db = chain_data_path.join(format!("{}", self.to));
+        info!(
+            "Renaming database {} to {}",
+            migrated_db.display(),
+            new_db.display()
+        );
```
lemmih, line 107: Is this important information?

src/db/parity_db.rs
```diff
@@ -69,6 +75,13 @@ impl DbColumn {
                         compression,
                         ..Default::default()
                     },
+                    DbColumn::EthMappings => parity_db::ColumnOptions {
+                        preimage: false,
+                        // This is needed for key retrieval.
+                        btree_index: true,
```
lemmih, line 81: Why is this needed for key retrieval?

lemmih (changes requested):

src/daemon/db_util.rs
elmattic: Sure.

src/db/parity_db.rs
elmattic: I'm not 100% sure. Hash column indexes should work at the moment (I'm testing them). But what if we need to implement GC on `Hash` to message the `Cid` later (like Lotus is doing, see https://github.com/filecoin-project/lotus/blob/master/node/impl/full/txhashmanager.go#L122).

elmattic: Since we want iteration for later implementation of GC on Eth mappings, let's keep it set to `true`.

ruseinov: Actually we don't need it for the GC. But we need it to retrieve data by CID when exposing a set of RPC endpoints.

ruseinov:
Some things to consider: https://github.com/ChainSafe/forest/blob/6f60fedc52a10d8575a6f06950fb278476dedbb0/src/db/parity_db.rs#L126 needs to account for the new column as well as all the places this method is being called.

src/db/parity_db.rs
elmattic: Hash indexes are enough for iterating on column values. But I'm unsure if we can retrieve the key for the value for the mapping `{Hash => message Cid}.`

elmattic: I was able to retrieve the transaction `Hash` from the mapping value (message `Cid` and timestamp pair).

lemmih (approved):

ruseinov (approved):

PR #4420, opened 2024-06-19
fix: refactor StateManager::replay to get rid of channels

hanabi1224:
## Summary of changes

To address the original comment
> // This isn't ideal to have, since the execution is synchronous, but this needs
        // to be the case because the state transition has to be in blocking
        // thread to avoid starving executor
        let (tx, rx) = flume::bounded(1);

- refactor `replay` into `replay_blocking` and `replay` to avoid using channels

```
     Running `target/quick/forest-tool api compare /home/me/fr/snapshots/calibnet/forest_snapshot_calibnet_2024-06-18_height_1711826.forest.car.zst --filter StateReplay`
| RPC Method                | Forest | Lotus |
|---------------------------|--------|-------|
| Filecoin.StateReplay (24) | Valid  | Valid |
```

Closes

## Change checklist

### DIFF
--- a/src/tool/subcommands/snapshot_cmd.rs
+++ b/src/tool/subcommands/snapshot_cmd.rs
@@ -432,15 +432,19 @@ fn print_computed_state(snapshot: PathBuf, epoch: ChainEpoch, json: bool) -> any
         beacon,
         &MultiEngine::default(),
         tipset,
-        Some(|ctx: &MessageCallbackCtx| {
-            message_calls.push((
-                ctx.message.clone(),
-                ctx.apply_ret.clone(),
-                ctx.at,
-                ctx.duration,
-            ));
-            anyhow::Ok(())
-        }),
+        if json {
+            Some(|ctx: &MessageCallbackCtx| {
+                message_calls.push((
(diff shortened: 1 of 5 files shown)


src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -432,15 +432,19 @@ fn print_computed_state(snapshot: PathBuf, epoch: ChainEpoch, json: bool) -> any
...
-        Some(|ctx: &MessageCallbackCtx| {
-            message_calls.push((
-                ctx.message.clone(),
-                ctx.apply_ret.clone(),
-                ctx.at,
-                ctx.duration,
-            ));
-            anyhow::Ok(())
-        }),
+        if json {
```
hanabi1224, line 435: `message_calls` is never used when `json == false`

```rust
if json {
        println!("{:#}", structured::json(state_root, message_calls)?);
    } else {
        println!("computed state cid: {}", state_root);
    }
```

aatifsyed (approved):

lemmih (approved):

ruseinov (approved):

PR #4425, opened 2024-06-19
[Fix] Sweep: delete after iteration.

ruseinov:
## Summary of changes

- GC: avoids ParityDB concurrency issues be scheduling items for deletion after iterating all the columns.

Closes

## Change checklist


src/db/gc/mod.rs
```diff
@@ -95,13 +96,15 @@ const SETTINGS_KEY: &str = "LAST_GC_RUN";
 pub struct MarkAndSweep<DB> {
     db: Arc<DB>,
     get_heaviest_tipset: Box<dyn Fn() -> Arc<Tipset> + Send>,
-    marked: HashSet<u32>,
+    marked: HashSet<Cid>,
```
hanabi1224, line 99: Let's use `SmallCid` to save memory usage

ruseinov: Once this has been tested - most definitely!

LesnyRumcajs (commented):
1. If it fixes the GC, it's worth adding a changelog entry.
2. Could you please describe the manual tests you performed to assess that GC is working?
3. Is it feasible to have an automatic test to reproduce the original issue and catch future regressions?

src/cli_shared/cli/mod.rs
```diff
@@ -147,7 +147,7 @@ pub struct CliOpts {
     pub track_peak_rss: bool,
     // GC is broken. Tracking issue: https://github.com/ChainSafe/forest/issues/3863
```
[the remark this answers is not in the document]
ruseinov: Needs to be removed.

ruseinov:
> 1. If it fixes the GC, it's worth adding a changelog entry.
> 2. Could you please describe the manual tests you performed to assess that GC is working?
> 3. Is it feasible to have an automatic test to reproduce the original issue and catch future regressions?

1. Yes, I wasn't sure about it.
2. Two types of tests: 
  a. Running the GC every 90 seconds and making sure it does not remove reachable records
  b. Running the GC on a weak node for several days on a normal schedule, same result
3. Unfortunately too complicated, the regressions happened due to `truncated_hash`, too many collisions. This does not even show immediately on a mainnet snap running the GC pretty often, so no way to reliably reproduce this, unless we tie the implementation to now-obsolete `truncated_hash`.

src/db/gc/mod.rs
```diff
@@ -196,7 +198,7 @@ impl<DB: Blockstore + SettingsStore + GarbageCollectable + Sync + Send + 'static
         }
 
         // This signifies a new run.
-        if self.marked.is_empty() {
+        if self.marked.len() == 0 {
```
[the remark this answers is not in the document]
ruseinov: Same, I was thinking about that, let's do it. I guess it's more a matter of taste at the moment, but if there are more occurrences - definitely annoying to have to `==0`.

src/db/gc/mod.rs
```diff
@@ -328,7 +330,7 @@ mod test {
 
         // test insufficient epochs
         gc.gc_workflow(ZERO_DURATION).await.unwrap();
-        assert!(gc.marked.is_empty());
+        assert_eq!(gc.marked.len(), 0);
```
LesnyRumcajs, line 333: We don't need this change anymore, right?
```suggestion
        assert!(gc.marked.is_empty());
```

LesnyRumcajs (approved):

hanabi1224 (approved):

Issue #4424, opened 2024-06-19
Various RPC discrepancies compared to Lotus


alt-ivan:
### 1) `StateSectorPreCommitInfo`
When sector precommit info is not found/non-existent, the RPC returns an error instead of `null` (as Lotus does).
Expected (Lotus response):
```
{"jsonrpc":"2.0","result":null,"id":1}"
```

Actual (Forest):
```
{"jsonrpc":"2.0","error":{"code":-32603,"message":"SectorPreCommitOnChainInfo not found"},"id":1}
```

### 2) 'Message` deserialization has required fields that are optional in Lotus and default to 0 there

The fields
```
version
nonce
gas_limit
gas_fee_cap
gas_premium
```

are all optional when using Lotus (and will default to 0), whereas in Forest they are required and deserialization fails without them.

### 3) `StateWaitMsg` missing parameters

In Forest, this RPC only accepts two parameters and two are missing. 
Currently accepted parameters: 
```
message_cid
confidence
```

Missing parameters:
```
limit: ChainEpoch
allow_replaced: bool
```

To summarize, Lotus accepts all 4 in that exact order:
```
message_cid
confidence
limit
allow_replaced
```

## To reproduce

Call the related RPCs

## Log output

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
  paste log output...
  ```
</details>

## Environment (please complete the following information):

- OS:
- Branch/commit
- Hardware

hanabi1224:
I will look into this

hanabi1224:
Hey @alt-ivan 
Could you post the lotus version (`lotus --version`) you were testing against? And it will be great to have the payload you were using as well.

Regarding `Filecoin.StateSectorPreCommitInfo`, I get the below response when the sector number is invalid

```
# payload
{"jsonrpc":"2.0","id":0,"method":"Filecoin.StateSectorPreCommitInfo","params":
["f04040",18446744073709551615,[{"/":"bafy2bzacednxz2tzqrneijbboehcrcbdt2357musif4dyetzyyjlvu6q6xnpq"},{"/":"bafy2bzaceblbbv22npiwjhqta3dmxxpzyedjrfxugsell3ouprocd3x333bqo"}]]
}
# response
{"error":{"code":1,"message":"precommit info does not exist"},"id":0,"jsonrpc":"2.0"}
```

alt-ivan:
Hi @hanabi1224 ,

[...]

PR #4427, opened 2024-06-19
fix(rpc): test coverage for StateSectorPreCommitInfo-not-found

hanabi1224:
## Summary of changes

Part of #4424

- cover invalid miner address case in API compare test for `Filecoin.StateSectorPreCommitInfo`
- cover invalid sector number case in API compare test for `Filecoin.StateSectorPreCommitInfo`

```
     Running `target/quick/forest-tool api compare /home/me/fr/snapshots/calibnet/forest_snapshot_calibnet_2024-06-19_height_1714793.forest.car.zst --filter StateSectorPreCommitInfo`
| RPC Method                             | Forest
         | Lotus                                                                                    |
|----------------------------------------|-----------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|
| Filecoin.StateSectorPreCommitInfo (10) | Rejected("Unknown miner actor code bafk2bzacedjnrb5glewazsxpcx6rwiuhl4kwrfcqolyprn6rrjtlzmthlhdq6") | Rejected("(get sset) failed to load miner actor state: actor code is not miner: system") |
| Filecoin.StateSectorPreCommitInfo (27) | Rejected("precommit info does not exist")
         | Rejected("precommit info does not exist")                                                |
| Filecoin.StateSectorPreCommitInfo (24) | Valid
         | Valid                                                                                    |
```

Closes

## Change checklist

### DIFF
--- a/src/tool/subcommands/api_cmd.rs
+++ b/src/tool/subcommands/api_cmd.rs
@@ -307,12 +307,18 @@ impl From<&RpcTest> for RpcTestHashable {
     }
 }
 
+enum PolicyOnRejected {
+    Fail,
+    Pass,
+    PassWithIdenticalError,
+}
+
 struct RpcTest {
     request: rpc::Request,
     check_syntax: Arc<dyn Fn(serde_json::Value) -> bool + Send + Sync>,
     check_semantics: Arc<dyn Fn(serde_json::Value, serde_json::Value) -> bool + Send + Sync>,
     ignore: Option<&'static str>,
-    pass_on_rejected: bool,
+    policy_on_rejected: PolicyOnRejected,
 }
 
 /// Duplication between `<method>` and `<method>_raw` is a temporary measure, and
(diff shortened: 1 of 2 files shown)


src/tool/subcommands/api_cmd.rs
```diff
@@ -340,7 +346,7 @@ impl RpcTest {
             }),
             check_semantics: Arc::new(|_, _| true),
             ignore: None,
-            pass_on_rejected: false,
+            policy_on_rejected: PolicyOnRejected::Fail,
```
hanabi1224, line 349: Should we make `PolicyOnRejected::PassWithIdenticalError` the default policy?

ruseinov: > Should we make `PolicyOnRejected::PassWithIdenticalError` the default policy?

Good question. 

If we define the endpoint wrong, sometimes both lotus and forest fail for all the test iterations. I don't think this should result in a green checkmark. 

An alternative approach here would be to make sure that we have at least 1 passing iteration, otherwise consider it a failure altogether.

aatifsyed (approved):

ruseinov (approved):

Issue #4436, opened 2024-06-20
remove rpc::eth::BigInt


aatifsyed:
https://github.com/ChainSafe/forest/blob/12a41ba986bdc101c841ddf3b04e16bebf074646/src/rpc/methods/eth.rs#L92C12-L92C18

use ethereum_types::U256 or use interact lotus_json::big_int better

PR #4437, opened 2024-06-20
fix: duplicate schema names

aatifsyed:
## Summary of changes

Fixes this in our generated OpenRPC
```jsonc
{ "components": { "schemas": { "Cid": ..., "Cid2": ... } } }
```

- add an `include` argument to `forest-tool shed openrpc`
- rename `eth::BigInt` to `eth::EthBigInt`, but see #4436 
- make LotusJson for Vec less special
- make LotusJson for Cid less special
- `<LotusJson as JsonSchema>::schema_id` (this fixes most of the bugs)
- tweak schemas for SignatureType,  Vec<u8>, 

## Change checklist

### DIFF
--- a/src/rpc/mod.rs
+++ b/src/rpc/mod.rs
@@ -465,7 +465,7 @@ where
     module
 }
 
-pub fn openrpc() -> openrpc_types::OpenRPC {
+pub fn openrpc(include: Option<&[&str]>) -> openrpc_types::OpenRPC {
     use schemars::gen::{SchemaGenerator, SchemaSettings};
     let mut methods = vec![];
     // spec says draft07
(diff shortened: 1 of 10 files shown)


src/rpc/mod.rs
```diff
@@ -465,7 +465,7 @@ where
     module
 }
 
-pub fn openrpc() -> openrpc_types::OpenRPC {
+pub fn openrpc(include: Option<&[&str]>) -> openrpc_types::OpenRPC {
```
ruseinov, line 468: nit: it would be nice to document this method, e.g. "by default exposes all the available endpoints, if `include` is specified - then only the selected ones".  Not having to read the actual code to understand what a method does goes a long way.

ruseinov:
Looks good to me apart from one minor nit.

ruseinov (approved):

src/tool/subcommands/shed_cmd.rs
```diff
@@ -116,10 +116,15 @@ impl ShedCommands {
                     println!("{}", BASE64_STANDARD.encode(keypair_data));
                 }
             }
-            ShedCommands::Openrpc => {
+            ShedCommands::Openrpc { include } => {
```
LesnyRumcajs, line 119: Could we have at least a simple component test that checks if the command succeeds (there are no panics) and that the output is a valid JSON? Along the lines of the ones here https://github.com/ChainSafe/forest/blob/main/tests/tool_tests.rs

aatifsyed: I think that would end up being just a more harnessed version of this ut, no? https://github.com/ChainSafe/forest/blob/main/src%2Frpc%2Fmod.rs#L517

LesnyRumcajs: Theoretically, yes. In practice, this part of code may grow/change (for whatever reason) and it would be helpful to immediately catch obvious regressions here.

aatifsyed: I'll do it for you 😘, but since [this subcommand is semver-exempt anyway](https://github.com/ChainSafe/forest/blob/6433e238da4e97225c8ae8c36b6dad7402213c09/src/tool/subcommands/mod.rs#L74), I don't think it's necessary.

LesnyRumcajs: Semver-exempt doesn't mean wild west - if a drunk developer puts less or more visible panic there, the CI should still fail.

LesnyRumcajs (approved):

Issue #3960, opened 2024-02-16
Support different `/rpc/vX` paths in `RpcMethod`


elmattic:
Previous code [here](https://github.com/ChainSafe/forest/blob/v0.16.6/src/rpc/mod.rs#L192-L197) and [here](https://github.com/ChainSafe/forest/blob/v0.16.6/src/rpc/rpc_util.rs#L44-L50).

`trait RpcMethod` assumes that each method will be exposed on one (and only one) endpoint:
https://github.com/ChainSafe/forest/blob/07576ded7933fa85ca623eb8ef84c3a4fb77eb22/src/rpc/reflect/mod.rs#L74-L75.

This was fine for a while because methods weren't different between endpoints, and [we ignore HTTP paths](https://github.com/ChainSafe/forest/blob/07576ded7933fa85ca623eb8ef84c3a4fb77eb22/src/rpc/mod.rs#L343) as a matter of course.

For #4424 and #4038 we need to start being path-aware again:
- (relatively) low-cost to support v0 methods.
- prepare our architecture for future API path changes
- this is required for us to emit OpenRPC documents per-path, which we want to support #4038

This issue covers the design and implementation of this work
- [x] `trait RpcMethod` allows multiple paths
  ```diff,rust
  - const API_VERSION: ApiVersion;
  + const API_PATHS: &[ApiPath];
  ```
  We can `const assert` for uniqueness and inhabitedness etc where we need (or use `nunny::slice![..]`).
- [ ] our [server code](https://github.com/ChainSafe/forest/blob/07576ded7933fa85ca623eb8ef84c3a4fb77eb22/src/rpc/mod.rs#L343) hoists the correct methods to the correct paths.
- [ ] `forest-tool api compare` tests all `API_PATHS` for each method.
- [x] our [openrpc code](https://github.com/ChainSafe/forest/blob/07576ded7933fa85ca623eb8ef84c3a4fb77eb22/src/rpc/reflect/mod.rs#L158) segregates by path, and we get rid of our [tagging logic](https://github.com/ChainSafe/forest/blob/07576ded7933fa85ca623eb8ef84c3a4fb77eb22/src/rpc/reflect/mod.rs#L41-L53)
  - [x] `forest-tool shed openrpc` accepts a path to dump out.

ansermino:
@lemmih Is this still relevant?

lemmih:
[...]

PR #4444, opened 2024-06-24
fix(rpc): update StateWaitMsg against v1 spec

hanabi1224:
## Summary of changes

(Part of https://github.com/ChainSafe/forest/issues/4424)
This PR fixes the `StateWaitMsg` RPC method against the [v1 spec](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v1-unstable-methods.md#StateWaitMsg) and uses it for both v0 and v1 endpoints until https://github.com/ChainSafe/forest/issues/3960 is resolved. (Keeping the v0 implementation with a different name for now)

- update `StateWaitMsg` against v1 spec

Closes

## Change checklist

### DIFF
--- a/src/tool/subcommands/api_cmd.rs
+++ b/src/tool/subcommands/api_cmd.rs
@@ -980,7 +980,12 @@ fn state_tests_with_tipset<DB: Blockstore>(
             tests.extend([
                 RpcTest::identity(StateReplay::request((tipset.key().into(), msg_cid))?),
                 validate_message_lookup(
-                    StateWaitMsg::request((msg_cid, 0))?.with_timeout(Duration::from_secs(30)),
+                    StateWaitMsg::request((msg_cid, 0, 10101, true))?
+                        .with_timeout(Duration::from_secs(15)),
+                ),
+                validate_message_lookup(
+                    StateWaitMsg::request((msg_cid, 0, 10101, false))?
+                        .with_timeout(Duration::from_secs(15)),
                 ),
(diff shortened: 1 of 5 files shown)


src/tool/subcommands/api_cmd.rs
```diff
@@ -980,7 +980,12 @@ fn state_tests_with_tipset<DB: Blockstore>(
             tests.extend([
                 RpcTest::identity(StateReplay::request((tipset.key().into(), msg_cid))?),
                 validate_message_lookup(
-                    StateWaitMsg::request((msg_cid, 0))?.with_timeout(Duration::from_secs(30)),
```
LesnyRumcajs, line 983: Can we test v0 as well, or is it not feasible?

hanabi1224: Not easily feasible, the actual V0 method name uses a `V0` suffix and does not exist in lotus

LesnyRumcajs: that's disappointing. It effectively makes the entire v0 path for this method untested and likely incorrect (now or in the future)

LesnyRumcajs (approved):

src/tool/subcommands/api_cmd.rs
lemmih: Right now, we can either test the v0 methods or the v1 methods but not both. Aatif is working on fixing this.

lemmih (approved):

Issue #4441, opened 2024-06-21
Forest stuck on downloading headers


LesnyRumcajs:
## Describe the bug

Forest sometimes gets stuck on downloading block headers.

## To reproduce

It happens from time to time, likely due to network conditions and peer selection (or not?). After a restart it usually recovers. No clear way to reproduce without further investigation.

## Log output

```
2024-06-20T14:49:09.455644Z  INFO forest_filecoin::daemon: Starting Forest daemon, version 0.19.0+git.a4fe792
2024-06-20T14:49:09.455784Z  WARN forest_filecoin::daemon: Forest has encryption disabled
2024-06-20T14:49:09.455877Z  INFO forest_filecoin::daemon: Admin token: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXSwiZXhwIjoxNzI0MDc4OTQ5fQ.2mMWKCjauicZBmxYUExdRkzBjm43IQQlHI_L_aFQhcA
2024-06-20T14:49:09.456279Z  INFO forest_filecoin::db::migration::db_migration: No database migration required
2024-06-20T14:49:09.622853Z  INFO forest_filecoin::daemon::bundle: Loading actor bundle from /var/tmp/forest_actor_bundle.car.zst set by FOREST_ACTOR_BUNDLE_PATH environment variable
2024-06-20T14:49:10.118055Z  INFO forest_filecoin::genesis: Initialized genesis: bafy2bzacecyaggy24wol5ruvs6qm73gjibs2l2iyhcqmvi7r7a4ph7zx3yqd4
2024-06-20T14:49:10.118127Z  INFO forest_filecoin::daemon: Prometheus server started at 0.0.0.0:6116
2024-06-20T14:49:10.118648Z  INFO forest_filecoin::daemon: Using network :: calibnet
2024-06-20T14:49:10.118987Z  INFO forest_filecoin::libp2p::behaviour: libp2p Forest version: 0.19.0+git.a4fe792
2024-06-20T14:49:10.119651Z  INFO libp2p_swarm: local_peer_id=12D3KooWAAc8fL1KRjwTEdWV2KJ9KarV51f4f3u7nQ1smJvy8XDT
[...]

PR #4454, opened 2024-06-26
fix(p2p): validate chain exchange responses for block headers

hanabi1224:
## Summary of changes

Part of #4441 investigation

-

Closes

## Change checklist

### DIFF
--- a/src/chain_sync/network_context.rs
+++ b/src/chain_sync/network_context.rs
@@ -141,8 +141,30 @@ where
         tsk: &TipsetKey,
         count: u64,
     ) -> Result<Vec<Arc<Tipset>>, String> {
-        self.handle_chain_exchange_request(peer_id, tsk, count, HEADERS, |_| true)
-            .await
+        self.handle_chain_exchange_request(
+            peer_id,
+            tsk,
+            count,
+            HEADERS,
+            |tipsets: &Vec<Arc<Tipset>>| {
+                if let Some(start) = tipsets.first() {
+                    if start.key() != tsk {
+                        tracing::warn!(epoch=%start.epoch(), expected=%tsk, actual=%start.key(), "start tipset key mismatch");
+                        return false;
+                    }
+                    for (ts, pts) in tipsets.iter().zip(tipsets.iter().skip(1)) {
+                        if ts.parents() != pts.key() {
+                            tracing::warn!(epoch=%ts.epoch(), expected_parent=%pts.key(), actual_parent=%ts.parents(), "invalid chain");
+                            return false;
+                        }
+                    }
+                    true
+                } else {
+                    false
+                }
+            },
+        )
+        .await
     }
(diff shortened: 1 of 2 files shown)


src/chain_sync/network_context.rs
```diff
@@ -139,10 +140,33 @@ where
...
+                            tracing::warn!(epoch=%ts.epoch(), expected_parent=%pts.key(), actual_parent=%ts.parents(), "invalid chain");
+                            return false;
+                        }
+                    }
+                    true
+                } else {
+                    tracing::warn!(%count, "invalid empty chain_exchange_headers response");
+                    false
+                }
+            },
```
LesnyRumcajs, line 167: Would it make sense to extract this to a named method a write a unit test to check and document the behaviour? It also might prevent future regressions.

LesnyRumcajs (approved):
I'm still waiting for a unit test but manage my expectations. No bonusly to incentivize one. :(

src/chain_sync/network_context.rs
hanabi1224: fixed

hanabi1224:
> I'm still waiting for a unit test but manage my expectations. No bonusly to incentivize one. :(

@LesnyRumcajs unit tests added!

src/chain_sync/network_context.rs
```diff
@@ -562,4 +597,37 @@ mod tests {
         assert_eq!(batch.get_ok().await, None);
         assert!(exceeded.load(Ordering::Relaxed));
     }
+
+    #[test]
+    #[allow(unused_variables)]
```
LesnyRumcajs, line 602: What variables are unused?

hanabi1224: `genesis_header`, `first_header` etc.

src/libp2p/peer_manager.rs
```diff
@@ -124,30 +124,49 @@ impl PeerManager {
...
-                    };
-                    Some((p, cost))
-                } else {
-                    None
+            .map(|(&p, info)| {
+                let is_stateful = info.head.epoch() > 0;
+                if is_stateful {
+                    n_stateful += 1;
                 }
+                let cost = if info.successes + info.failures > 0 {
```
hanabi1224, line 136: wibble: this was changed to `if info.successes > 0` to address https://github.com/ChainSafe/forest/pull/4376#discussion_r1636561639
However, this unexpectedly changes the logic to favor peers that never succeed.
The original code does not panic but returns `inf`

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=657686952703dcee856841d192b99774

```rust
fn main() {
    let x = 100.0 + f64::from(1) / f64::from(0);
    println!("{x}");
}

// Output: inf
```

LesnyRumcajs (approved):

lemmih (approved):

PR #4474, opened 2024-07-02
Move eth mapping creation to a background task

elmattic:
## Summary of changes

- Add Ethereum mapping populate task in background
- Add status to `/readyz` endpoint
- Add new subcommand `forest-cli healthcheck`.

```console
$ curl "http://127.0.0.1:2346/readyz?verbose"
[!] sync incomplete
[!] epoch outdated
[+] rpc server running
[!] no eth mapping
```

Closes

## Change checklist


sudo-shashank (approved):

scripts/tests/api_compare/docker-compose.yml
```diff
@@ -60,6 +61,43 @@ services:
       timeout: 10m
       retries: 3
       start_period: 10m
+  forest-rpc-ready:
```
hanabi1224, line 64: Why not just update `healthcheck` of `forest`?

src/cli_shared/mod.rs
```diff
@@ -19,6 +20,14 @@ pub fn chain_path(config: &Config) -> PathBuf {
     PathBuf::from(&config.client.data_dir).join(config.chain.to_string())
 }
 
+/// Gets car db path
+pub fn car_db_path(config: &Config) -> anyhow::Result<PathBuf> {
+    let chain_data_path = chain_path(config);
+    let db_root_dir = db_root(&chain_data_path)?;
+    let forest_car_db_dir = db_root_dir.join("car_db");
```
hanabi1224, line 27: I've seen many occurrences of `car_db` in the code base, can we make it a constant?

src/db/mod.rs
```diff
@@ -43,6 +46,16 @@ pub trait SettingsStore {
 
     /// Returns all setting keys.
     fn setting_keys(&self) -> anyhow::Result<Vec<String>>;
+
+    /// Sets the Ethereum mapping status to "up-to-date".
+    fn set_eth_mapping_up_to_date(&self) -> anyhow::Result<()> {
```
hanabi1224, line 51: I'm not sure if these `eth` specific methods belong to `SettingsStore` trait. Maybe use a different trait?

ruseinov: Yeah, I tend to agree. Might be confusing going forward. 
Perhaps we should introduce an actual trait like `Settings` that would have all those different methods for all sorts of settings i/o.

src/health/endpoints.rs
```diff
@@ -49,12 +50,13 @@ pub(crate) async fn readyz(
 ) -> Result<String, AppError> {
     let mut acc = MessageAccumulator::new_with_enabled(params.contains_key(VERBOSE_PARAM));
 
-    let mut healthy = true;
-    healthy &= check_sync_state_complete(&state, &mut acc);
-    healthy &= check_epoch_up_to_date(&state, &mut acc);
-    healthy &= check_rpc_server_running(&state, &mut acc).await;
+    let mut ready = true;
```
hanabi1224, line 53: How about 
```rust
let ready = [check_sync_state_complete(&state, &mut acc),
check_epoch_up_to_date(&state, &mut acc),
check_rpc_server_running(&state, &mut acc).await,
check_eth_mapping_up_to_date(&state, &mut acc),
].iter().fold(true, |acc, &v| acc && v);
```

LesnyRumcajs: What's the benefit of this approach? It seems more readable (severely less readable if you're not proficient in Rust) and, unless Rust does some optimizations under the hood, less performant (though it's negligible in this case).

scripts/tests/api_compare/docker-compose.yml
```diff
@@ -60,6 +61,43 @@ services:
...
+    networks:
+      - api-tests
+    entrypoint: [ "/bin/bash", "-c" ]
+    user: 0:0
+    command:
+      - |
+        set -uo pipefail
+        type curl >/dev/null 2>&1 || {
+          echo >&2 "curl is required but it's not installed. installing.";
+          apt-get update && apt-get install -y curl
```
LesnyRumcajs, line 80: This means `curl` needs to be installed on every `compose up`. It'd be nicer to embed it in an image.

src/health/endpoints.rs
elmattic: I agree with Hubert. I think we should keep it like this.

hanabi1224:
I believe `Calibnet eth mapping check` CI task needs to be re-enabled

scripts/tests/api_compare/docker-compose.yml
elmattic: Using now some healthcheck subcommand so we don't prereq `curl`.

src/db/mod.rs
elmattic: Done.

scripts/tests/api_compare/docker-compose.yml
elmattic: This could work, but I prefer to have a dedicated service for now.

For example, `forest-wallet-import` doesn't need to wait until Ethereum mapping is "up to date".

src/daemon/db_util.rs
```diff
@@ -168,15 +168,20 @@ where
 {
     let mut delegated_messages = vec![];
 
-    info!("Populating column EthMappings");
+    let hygge = state_manager.chain_config().epoch(Height::Hygge);
+    tracing::info!(
+        "Populating column EthMappings from range: [{}, {}]",
```
LesnyRumcajs, line 173: Would it make sense to have it reversed for readability?

LesnyRumcajs (approved):

src/daemon/db_util.rs
elmattic: Yeah, that might be clearer for the end user.

LesnyRumcajs (approved):

PR #4497, opened 2024-07-10
feat: forest-tool openrpc shed <path>

aatifsyed:
## Summary of changes

- Remove `ApiVersion` and replace it with `ApiPaths`, a closed set of URLs.
- Add plumbing for filtering at the `forest-tool openrpc shed` level
- note this does NOT change how we mount and serve RPC requests
- note this does NOT change how we test RPC endpoints

- part of #3960

## Change checklist

### DIFF
--- a/src/rpc/reflect/mod.rs
+++ b/src/rpc/reflect/mod.rs
@@ -26,31 +26,16 @@ use crate::lotus_json::HasLotusJson;
 use self::{jsonrpc_types::RequestParameters, util::Optional as _};
 use super::error::ServerError as Error;
 use fvm_ipld_blockstore::Blockstore;
+use itertools::{Either, Itertools as _};
 use jsonrpsee::RpcModule;
 use openrpc_types::{ContentDescriptor, Method, ParamStructure, ReferenceOr};
 use parser::Parser;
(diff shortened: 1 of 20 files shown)


src/rpc/reflect/mod.rs
```diff
@@ -103,37 +86,40 @@ pub enum Permission {
...
-pub enum ApiVersion {
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub enum ApiPaths {
+    /// Only expose this method on `/rpc/v0`
     V0,
+    /// Only expose this method on `/rpc/v1`
     V1,
+    /// Expose this method on both `/rpc/v0` and `/rpc/v1`
+    #[allow(dead_code)]
+    Both,
```
hanabi1224, line 100: `Both` seems less extensible when there's a future V2, how about using bit flags?

aatifsyed: I agree, but I would prefer to defer that to when the time comes - bitflags are slightly harder to iterate and we don't need that for now

hanabi1224 (approved):

elmattic (approved):

PR #4541, opened 2024-07-16
fix: reveal the proper logo for the waffle upgrade

hanabi1224:
## Summary of changes

There seems to be some off-by-1 error in the current logic that in CI devnet test it reveals the wrong logo on the waffle upgrade, see https://github.com/ChainSafe/forest/actions/runs/9958493368/job/27513328498#step:8:427

After the fix, see https://github.com/ChainSafe/forest/actions/runs/9965722823/job/27536633471?pr=4541#step:8:438

-

Closes

## Change checklist

### DIFF
--- a/src/state_migration/mod.rs
+++ b/src/state_migration/mod.rs
@@ -109,7 +109,7 @@ where
                 .map(|sr| format!("{}", sr.actors))
                 .unwrap_or_default();
             if new_state != *parent_state {
-                crate::utils::misc::reveal_upgrade_logo(chain_config.network_version(epoch));
+                crate::utils::misc::reveal_upgrade_logo(height.into());
                 tracing::info!("State migration at height {height}(epoch {epoch}) was successful, Previous state: {parent_state}, new state: {new_state}, new state actors: {new_state_actors}. Took: {elapsed}s.");
             } else {
                 anyhow:: bail!("State post migration at height {height} must not match. Previous state: {parent_state}, new state: {new_state}, new state actors: {new_state_actors}. Took {elapsed}s.");


LesnyRumcajs (approved):
complete fiasco

src/state_migration/mod.rs
```diff
@@ -109,7 +109,7 @@ where
                 .map(|sr| format!("{}", sr.actors))
                 .unwrap_or_default();
             if new_state != *parent_state {
-                crate::utils::misc::reveal_upgrade_logo(chain_config.network_version(epoch));
+                crate::utils::misc::reveal_upgrade_logo(height.into());
```
lemmih, line 112: I'm surprised that `chain_config.network_version(epoch) != height` since `epoch == chain_config.epoch(height)`. I thought those functions would be the inverse of each other.

lemmih (approved):

src/state_migration/mod.rs
hanabi1224: I checked the corresponding logic in lotus and we seem to match each other.  `fn network_version` is widely used in daemon and RPC implementations and I don't think we can easily change it to be the inverse of `chain_config.network_version`

```go
func (sm *StateManager) GetNetworkVersion(ctx context.Context, height abi.ChainEpoch) network.Version {
	// The epochs here are the _last_ epoch for every version, or -1 if the
	// version is disabled.
	for _, spec := range sm.networkVersions {
		if height <= spec.atOrBelow {
			return spec.networkVersion
		}
	}
	return sm.latestVersion
}
```

LesnyRumcajs: It's even more surprising given that I checked it on my machine (not devnet, though), and it worked as expected. 🤣

Issue #4550, opened 2024-07-17
Look into DRY'er actor version code.


ruseinov:
I wish we could make it less repetitious.

_Originally posted by @lemmih in https://github.com/ChainSafe/forest/pull/4545#pullrequestreview-2182672452_

PR #4551, opened 2024-07-17
[Enhancement] Macros for State.GetAll*

ruseinov:
## Summary of changes

- Macros for `VerifiedRegistryStateExt`

Work on #4550 

## Change checklist

### DIFF
--- a/src/shim/actors/verifreg/state.rs
+++ b/src/shim/actors/verifreg/state.rs
@@ -166,7 +166,7 @@ impl VerifiedRegistryStateExt for State {
 
 #[macro_export]
 macro_rules! list_all_inner_old {
-    ($state:ident, $store:ident, $version:ident, $method:ident, $map: ident) => {{
+    ($state:ident, $store:ident, $version:ident, $method:ident, $map:ident) => {{
         let mut entities = $state.$method($store)?;
         let mut actors = vec![];
         entities.for_each_outer(|k, _| {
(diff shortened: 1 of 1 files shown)


LesnyRumcajs (approved):

src/shim/actors/verifreg/state.rs
```diff
@@ -249,109 +153,55 @@ impl VerifiedRegistryStateExt for State {
...
+            State::V11(state) => list_all_inner_old!(state, store, v11, load_claims, result),
+            State::V12(state) => list_all_inner!(state, store, v12, load_claims, result),
+            State::V13(state) => list_all_inner!(state, store, v13, load_claims, result),
+            State::V14(state) => list_all_inner!(state, store, v14, load_claims, result),
         };
         Ok(result)
     }
 }
+
+#[macro_export]
```
hanabi1224, line 167: Do we have to export these macros?

LesnyRumcajs (approved):

src/shim/actors/verifreg/state.rs
ruseinov: If we don't - the macros have to be defined strictly above the code that uses them, which is inconvenient.

hanabi1224 (approved):

src/shim/actors/verifreg/state.rs
aatifsyed: this exports the macro as a public item at the root of the crate. Does `pub(crate) use list_all_inner` below the macro work? Else I think moving the definition to the top of the file is the right thing to do

aatifsyed (dismissed):

ruseinov:
>this exports the macro as a public item at the root of the crate. Does pub(crate) use list_all_inner below the macro work? Else I think moving the definition to the top of the file is the right thing to do

Gotta `export` to `use`. Alright, let's move this to the top then :/

LesnyRumcajs (approved):

PR #4573, opened 2024-07-22
fix(ci): set LOTUS_SYNC_BOOTSTRAP_PEERS=1 to let lotus daemon sync

hanabi1224:
## Summary of changes

This PR fixes an issue that CI jobs that run lotus daemon get stuck due to bootstrap peer connectivity issue

-

Closes

## Change checklist


LesnyRumcajs:
Note that this is a temporary workaround if we intend to have Forest as bootstrap nodes. The root issue was that the Glif node was unreachable for some, and then even more ChainSafe-hosted Lotus bootstrap nodes were not working properly. The only bootstrapper, for some regions, that was pingable was Forest, but it was not able to bootstrap other Lotus peers. They would connect to it, even exchange some blocks, but new peers would not get propagated.

LesnyRumcajs (approved):

scripts/devnet/.env
```diff
@@ -2,6 +2,7 @@ LOTUS_IMAGE=ghcr.io/chainsafe/lotus-devnet:2024-07-03-1f030c2
 FOREST_DATA_DIR=/forest_data
 LOTUS_DATA_DIR=/lotus_data
 FIL_PROOFS_PARAMETER_CACHE=/var/tmp/filecoin-proof-parameters
+LOTUS_SYNC_BOOTSTRAP_PEERS=1
```
LesnyRumcajs, line 5: there's no need to do it here, 2k networks have this value at `1` by default.

hanabi1224: Yes but it doesn't hurt right?

LesnyRumcajs: It looks like we are overriding a default behaviour (but we don't), it doesn't bring any value in this context, while only increasing the complexity if anyone is analysing it. That said, it's just a personal preference, not a huge deal.

lemmih (approved):

PR #4580, opened 2024-07-23
fix(p2p): make kademlia in forest work for lotus

### LINKED ISSUE #4576: Forest can't bootstrap a Lotus node
## Issue summary

It turns out a Forest node (`v0.19.0`) is not a able to bootstrap a Lotus node. It can, however, bootstrap another Forest node.

To reproduce on calibnet:
1. Start a Forest node, grab it's multiaddr with peer id.
2. Compile Lotus with the above as bootstrap node (make sure to set `LOTUS_SYNC_BOOTSTRAP_PEERS=1`)
3. With `lotus net peers` observe that the only peer it is connected to is the bootstrap node, whereas `forest-cli net peers` shows way more peers.

Similarly, do the same with another Forest and observe that the second instance has many other peers.

This behaviour _might_ stem from the fact that the peer exchange is not implemented in gossipsub in rust-libp2p ( [code mention](https://github.com/libp2p/rust-libp2p/blob/a749a41b95441a3dc87c61c472e7a9f00ca997fc/protocols/gossipsub/src/config.rs#L230-L234) and [issue](https://github.com/libp2p/rust-libp2p/issues/2398)) while [explicitly](https://github.com/filecoin-project/lotus/blob/ff0e889b587621a9bbb973a6e1ef0f145a565216/node/modules/lp2p/pubsub.go#L331-L344) enabled in Lotus. 

Does peer exchange over Kademlia and Identify not work in Lotus? If so, can this be mitigated? The alternative would be implementing peer exchange in gossipsub, which to my understanding, is not a trivial task.

To work on this, and ensure it doesn't break in the future, once fixed, we also need a CI job to check Forest-bootstrapper compatibility with Lotus (and another Forest).

hanabi1224:
## Summary of changes

Part of #4576

`go-libp2p-kad-dht` sends the remote peer id as query key during bootstrap, however, `rust-libp2p` returns empty peer list when the key it receives matches the local one. This PR configures Kademlia with a random peer ID as a workaround

- 

Closes 

## Change checklist

### DIFF
--- a/src/libp2p/discovery.rs
+++ b/src/libp2p/discovery.rs
@@ -139,7 +139,7 @@ impl<'a> DiscoveryConfig<'a> {
             // `go-libp2p-kad-dht` sends the remote peer id as query key during bootstrap,
             // however, `rust-libp2p` returns empty peer list when the key it receives
             // matches the local one. It's fine to use a random peer id as a workaround
-            // since Kadmelia is not used as value providers in filecoin p2p network.
+            // since Kademlia not used as value providers in filecoin p2p network.
             let kad_random_peer_id = PeerId::random();
             let store = MemoryStore::new(kad_random_peer_id);
             let kad_config = {
--- a/src/libp2p/discovery.rs
+++ b/src/libp2p/discovery.rs
@@ -134,18 +134,22 @@ impl<'a> DiscoveryConfig<'a> {
 
         let mut peers = HashSet::new();
 
-        // Kademlia config
-        let store = MemoryStore::new(local_peer_id);
-        let kad_config = {
-            let mut cfg = kad::Config::default();
-            cfg.set_protocol_names(vec![StreamProtocol::try_from_owned(format!(
-                "/fil/kad/{network_name}/kad/1.0.0"
-            ))?]);
-            cfg
-        };
-
         let kademlia_opt = if enable_kademlia {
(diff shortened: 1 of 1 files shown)


src/libp2p/discovery.rs
```diff
@@ -134,18 +134,22 @@ impl<'a> DiscoveryConfig<'a> {
...
-            cfg
-        };
-
         let kademlia_opt = if enable_kademlia {
-            let mut kademlia = kad::Behaviour::with_config(local_peer_id, store, kad_config);
+            // This is a workaround for <https://github.com/ChainSafe/forest/issues/4576>
+            // `go-libp2p-kad-dht` sends the remote peer id as query key during bootstrap,
+            // however, `rust-libp2p` returns empty peer list when the key it receives
+            // matches the local one. It's fine to use a random peer id as a workaround
+            // since Kadmelia is not used as value providers in filecoin p2p network.
```
LesnyRumcajs, line 142:
```suggestion
            // since Kademlia not used as value providers in filecoin p2p network.
```

hanabi1224:
> Just to state the current state - given it's a workaround, the actual fix should be where exactly? I mean, we can live with a workaround but _at least_ we should create an issue in a relevant project.

@LesnyRumcajs Then let's keep #4576 open. I'm not sure where the actual fix should go yet. Let me explore the code more and discuss with libp2p forks on slack

LesnyRumcajs (approved):

PR #4617, opened 2024-08-05
fix: fix incorrect `MinerGetBaseInfo`

LesnyRumcajs:
## Summary of changes

-

Closes

## Change checklist


src/state_manager/utils.rs
```diff
@@ -102,7 +102,12 @@ where
 
         let out = sectors
             .into_iter()
-            .map(|s_info| SectorInfo::new(*spt, s_info.sector_number, s_info.sealed_cid))
+            .map(|s_info| ExtendedSectorInfo {
```
hanabi1224, line 105: How about implementing `From<SectorInfo> for ExtendedSectorInfo`

hanabi1224 (approved):

src/state_manager/utils.rs
LesnyRumcajs: the `s_info` is not a `SectorInfo` but a `SectorOnChainInfo`. I'd leave it as is, given the conversion would be used only in a single place.

lemmih (approved):

Issue #3403, opened 2023-08-21
snapshot validate does not validate snapshots independently


LesnyRumcajs:
**Describe the bug**

The `forest-tool snapshot validate [FILES]...` does not seem to validate the snapshots independently. It may be by design, but it's a pretty unexpected behaviour and should be either mitigated functionally (by validating each snapshot independently and failing the command if one of them is corrupted) or by having a separate subcommand for validating a union of snapshots.

**To Reproduce** Steps to reproduce the behavior:

1. Download a snapshot that is missing a genesis block. You can find one [here](https://cloud.digitalocean.com/spaces/forest-snapshots?path=debug%2F%233403%2F)
2. Download any valid snapshot.
3. Run `forest-tool snapshot validate --check-network calibnet forest_snapshot_calibnet_2023-08-21_height_842289.forest.car.zst`. You should see:
```
  Checking IPLD integrity:       ✅ verified!
  Identifying genesis block:     ❌ No valid genesis block!                                                                                                                                                                                                                                                                     Error: Snapshot does not contain a genesis block
```
4. Run `forest-tool snapshot validate --check-network calibnet <another-snapshot>
```
❯ forest-tool snapshot validate --check-network calibnet forest_snapshot_calibnet_2023-08-21_height_842529.forest.car.zst
  Checking IPLD integrity:       ✅ verified!
  Identifying genesis block:     ✅ found!                                                                                                                                                                                                                                                                                        Verifying network identity:    ✅ verified!
  Running tipset transactions:   ✅ verified!                                                                                                                                                                                                                                                                                   Snapshot is valid
```
5. Run the same with both snapshots as positional arguments:
```
❯ forest-tool snapshot validate --check-network calibnet forest_snapshot_calibnet_2023-08-21_height_842289.forest.car.zst forest_snapshot_calibnet_2023-08-21_height_842529.forest.car.zst
  Checking IPLD integrity:       ✅ verified!
  Identifying genesis block:     ✅ found!                                                                                                                                                                                                                                                                                        Verifying network identity:    ✅ verified!
  Running tipset transactions:   ✅ verified!                                                                                                                                                                                                                                                                                   Snapshot is valid
```

**Expected behaviour**

This `snapshot validate <multiple files>...` should fail if any file is corrupt. If we want to validate a union, there should be a separate subcommand, e.g., `snapshot validate-union`.

**Screenshots**

<!-- If applicable, add screenshots to help explain your problem. -->

**Environment (please complete the following information):**

[...]

PR #4623, opened 2024-08-06
snapshot validate each file separately

qingyunha:
## Summary of changes

-  Rename the validate command to validate-diffs and add a new validate command that checks each file separately.

example:
```bash
$ ./target/debug/forest-tool snapshot validate  test-snapshots/chain4*
Validating test-snapshots/chain4.car
  Checking IPLD integrity:       ❌ Invalid IPLD data!
Error: missing key: bafy2bzacebpkfyohjp7thdmpjy44lkimctgphlxu3apnkgwtrkxy723vhbk5y
Validating test-snapshots/chain4.car.zst
  Checking IPLD integrity:       ❌ Invalid IPLD data!
Error: missing key: bafy2bzacebpkfyohjp7thdmpjy44lkimctgphlxu3apnkgwtrkxy723vhbk5y
Validating test-snapshots/chain4.forest.car.zst
  Checking IPLD integrity:       ❌ Invalid IPLD data!
Error: missing key: bafy2bzacebpkfyohjp7thdmpjy44lkimctgphlxu3apnkgwtrkxy723vhbk5y
Error: validate failed
```

Closes  https://github.com/ChainSafe/forest/issues/3403

## Change checklist

### DIFF
--- a/src/tool/subcommands/snapshot_cmd.rs
+++ b/src/tool/subcommands/snapshot_cmd.rs
@@ -183,11 +183,11 @@ impl SnapshotCommands {
                         Ok::<(), anyhow::Error>(())
                     }
                     .await;
-                    if result.is_err() {
+                    if let Err(e) = result {
+                        has_fail = true;
+                        eprintln!("Error: {:?}", e);
                         if fail_fast {
-                            bail!("validate failed");
-                        } else {
-                            has_fail = true;
+                            break;
                         }
                     }
                 }
--- a/src/tool/subcommands/snapshot_cmd.rs
+++ b/src/tool/subcommands/snapshot_cmd.rs
@@ -45,6 +45,23 @@ pub enum SnapshotCommands {
     },
 
     /// Validates the snapshot.
+    ValidateDiffs {
+        /// Number of recent epochs to scan for broken links
+        #[arg(long, default_value_t = 2000)]
+        check_links: u32,
+        /// Assert the snapshot belongs to this network. If left blank, the
+        /// network will be inferred before executing messages.
+        #[arg(long)]
+        check_network: Option<crate::networks::NetworkChain>,
+        /// Number of recent epochs to scan for bad messages/transactions
+        #[arg(long, default_value_t = 60)]
+        check_stateroots: u32,
+        /// Path to a snapshot CAR, which may be zstd compressed
+        #[arg(required = true)]
+        snapshot_files: Vec<PathBuf>,
+    },
+
+    /// Validates the snapshot.
     Validate {
         /// Number of recent epochs to scan for broken links
         #[arg(long, default_value_t = 2000)]
@@ -124,7 +141,7 @@ impl SnapshotCommands {
                 }
                 Err(e) => cli_error_and_die(format!("Failed fetching the snapshot: {e}"), 1),
             },
-            Self::Validate {
+            Self::ValidateDiffs {
                 check_links,
                 check_network,
                 check_stateroots,
@@ -139,6 +156,26 @@ impl SnapshotCommands {
                     check_stateroots,
                 )
                 .await
+            }
+            Self::Validate {
+                check_links,
+                check_network,
+                check_stateroots,
+                snapshot_files,
+            } => {
+                for file in snapshot_files {
+                    println!("Validating {}", file.display());
+                    let store = AnyCar::try_from(file.as_path())?;
+                    validate_with_blockstore(
+                        store.heaviest_tipset()?,
+                        Arc::new(store),
+                        check_links,
+                        check_network.clone(),
+                        check_stateroots,
+                    )
+                    .await?
+                }
+                Ok(())
             }
             Self::Compress {
                 source,
--- a/src/tool/subcommands/snapshot_cmd.rs
+++ b/src/tool/subcommands/snapshot_cmd.rs
@@ -44,7 +44,7 @@ pub enum SnapshotCommands {
         vendor: snapshot::TrustedVendor,
     },
 
-    /// Validates the snapshot.
+    /// Validate the provided snapshots as a whole.
     ValidateDiffs {
         /// Number of recent epochs to scan for broken links
         #[arg(long, default_value_t = 2000)]
@@ -61,7 +61,7 @@ pub enum SnapshotCommands {
         snapshot_files: Vec<PathBuf>,
     },
 
-    /// Validates these snapshots separately.
+    /// Validate the snapshots individually.
     Validate {
         /// Number of recent epochs to scan for broken links
         #[arg(long, default_value_t = 2000)]
--- a/src/tool/subcommands/snapshot_cmd.rs
+++ b/src/tool/subcommands/snapshot_cmd.rs
@@ -61,7 +61,7 @@ pub enum SnapshotCommands {
         snapshot_files: Vec<PathBuf>,
     },
 
-    /// Validates the snapshot.
+    /// Validates these snapshots separately.
     Validate {
         /// Number of recent epochs to scan for broken links
         #[arg(long, default_value_t = 2000)]
@@ -76,6 +76,9 @@ pub enum SnapshotCommands {
         /// Path to a snapshot CAR, which may be zstd compressed
         #[arg(required = true)]
         snapshot_files: Vec<PathBuf>,
+        /// Fail at the first invalid snapshot
+        #[arg(long)]
+        fail_fast: bool,
     },
 
     /// Make this snapshot suitable for use as a compressed car-backed blockstore.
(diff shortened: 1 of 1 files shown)


src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -140,6 +157,26 @@ impl SnapshotCommands {
...
                 .await
             }
+            Self::Validate {
+                check_links,
+                check_network,
+                check_stateroots,
+                snapshot_files,
+            } => {
+                for file in snapshot_files {
+                    println!("Validating {}", file.display());
```
hanabi1224, line 167: How about
```rust
print!("Validating {} ... ", file.display());
...
println!("success");
```

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -44,6 +44,23 @@ pub enum SnapshotCommands {
...
+        check_links: u32,
+        /// Assert the snapshot belongs to this network. If left blank, the
+        /// network will be inferred before executing messages.
+        #[arg(long)]
+        check_network: Option<crate::networks::NetworkChain>,
+        /// Number of recent epochs to scan for bad messages/transactions
+        #[arg(long, default_value_t = 60)]
+        check_stateroots: u32,
+        /// Path to a snapshot CAR, which may be zstd compressed
+        #[arg(required = true)]
```
hanabi1224, line 60:
```suggestion
        #[arg(num_args = 1.., required = true)]
```

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -44,6 +44,23 @@ pub enum SnapshotCommands {
...
+        snapshot_files: Vec<PathBuf>,
```
hanabi1224, line 61: Hopefully glob in bash should work for this `forest-tool snapshot validate-diffs *.car.zst`

qingyunha: Seems Vec argument already accept multiple values.

src/tool/subcommands/snapshot_cmd.rs
qingyunha: The result is displayed by ` validate_with_blockstore` like this:
  Checking IPLD integrity:       ✅ verified!
  Identifying genesis block:     ❌ No valid genesis block!  
and
  Checking IPLD integrity:       ✅ verified!
  Identifying genesis block:     ✅ found!                                                                                                                                                                                                                                                                                        
  Verifying network identity:    ✅ verified!
  Running tipset transactions:   ✅ verified!

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -44,6 +44,23 @@ pub enum SnapshotCommands {
         vendor: snapshot::TrustedVendor,
     },
 
+    /// Validates the snapshot.
```
LesnyRumcajs, line 47: Let's document how this subcommand is different from the other one.

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -140,6 +157,26 @@ impl SnapshotCommands {
                 )
                 .await
             }
+            Self::Validate {
+                check_links,
+                check_network,
+                check_stateroots,
+                snapshot_files,
+            } => {
+                for file in snapshot_files {
```
LesnyRumcajs, line 166: Idea: This will stop validation at the first invalid file. The subcommand would be more flexible if it'd go through all the provided snapshots (bonus points if in parallel, though I'd expect some stdout thread races) and then list valid and invalid snapshots, returning an error if any was corrupted. We could also add a flag `--fail-fast` to have the same behaviour as now.

What do you think?

LesnyRumcajs (commented):
Thanks for contributing! Can you think of a way of adding automatic tests for this feature?

qingyunha:
> Thanks for contributing! Can you think of a way of adding automatic tests for this feature?

I tried to add a test in `tess/tool_tests.rs`. And I need to add a valid snapshot file in the git repository. But the file size is very large (4G), and  I'm not sure if it belongs here.

src/tool/subcommands/snapshot_cmd.rs
qingyunha: @LesnyRumcajs PTAL. And I'm having trouble handling the stdout race in parallel, as you mentioned. 😄

LesnyRumcajs:
> > Thanks for contributing! Can you think of a way of adding automatic tests for this feature?
> 
> I tried to add a test in `tess/tool_tests.rs`. And I need to add a valid snapshot file in the git repository. But the file size is very large (4G), and I'm not sure if it belongs here.

How about using the very first snapshot that is already there?
```
❯ RUST_LOG=warn forest-tool snapshot validate src/networks/mainnet/genesis.car
  Checking IPLD integrity:       ✅ verified!
  Identifying genesis block:     ✅ found!
  Running tipset transactions:   ✅ verified!                                           
Snapshot is valid
```

src/tool/subcommands/snapshot_cmd.rs
LesnyRumcajs: You can get around this by changing `validate_with_blockstore` to accept a `Writer`. You'd pass it, e.g., a string buffer to be printed once the operation is done.

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -140,6 +160,42 @@ impl SnapshotCommands {
...
+                            Arc::new(store),
+                            check_links,
+                            check_network.clone(),
+                            check_stateroots,
+                        )
+                        .await?;
+                        Ok::<(), anyhow::Error>(())
+                    }
+                    .await;
+                    if result.is_err() {
```
LesnyRumcajs, line 186: It's usually useful to have the cause of the error, so I'd recommend printing it out as well.

src/tool/subcommands/snapshot_cmd.rs
qingyunha: Thank you. I have two questions here.
- Do you mean to provide a distinct string buffer for every snapshot?
- `Validate_with_blockstore` uses a ProgressBar, string buffer seems not an option.

LesnyRumcajs: 1. Yeah, distinct buffer. The entire text shouldn't be long anyway.
2. That does seem a bit challenging. We could either do some terminal splitting magic (hardcore-level, quite bug-prone) or just allow disabling the progress bar.

That said, given it's not trivial to implement, it will be more pragmatic to settle with the sequential validation and revisit the parallel one only when it's actually needed.

src/tool/subcommands/snapshot_cmd.rs
LesnyRumcajs, line 186: Let's use an `if let` to avoid unwrapping and have more idiomatic Rust in place.

LesnyRumcajs (commented):
Looking good! Let's resolve the outstanding unresolved comments and I think we'll be good to go.

src/tool/subcommands/snapshot_cmd.rs
qingyunha: done

src/tool/subcommands/snapshot_cmd.rs
qingyunha: I have changed the document of `validate` command.

src/tool/subcommands/snapshot_cmd.rs
qingyunha: Seems this is not necessary.

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -45,6 +45,23 @@ pub enum SnapshotCommands {
...
+        check_network: Option<crate::networks::NetworkChain>,
+        /// Number of recent epochs to scan for bad messages/transactions
+        #[arg(long, default_value_t = 60)]
+        check_stateroots: u32,
+        /// Path to a snapshot CAR, which may be zstd compressed
+        #[arg(required = true)]
+        snapshot_files: Vec<PathBuf>,
+    },
+
+    /// Validates these snapshots separately.
```
LesnyRumcajs, line 64:
```suggestion
    /// Validate the snapshots individually
```

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -45,6 +45,23 @@ pub enum SnapshotCommands {
     },
 
     /// Validates the snapshot.
+    ValidateDiffs {
```
LesnyRumcajs, line 48: Let's change the docs for this to `Validate the provided snapshots as a whole`

LesnyRumcajs (approved):
Looks good to me, there were no tests before so I'm not adamant on testing that now (though I'd love to see that!).

src/tool/subcommands/snapshot_cmd.rs
ruseinov: Let's prove this by providing a list of tested cli commands in the description of the PR.

e.g.
`forest-tool snapshot validate-diffs ....`

Once this is done I'm happy to consider this PR correct and approve it.

src/tool/subcommands/snapshot_cmd.rs
```diff
@@ -59,6 +76,9 @@ pub enum SnapshotCommands {
         /// Path to a snapshot CAR, which may be zstd compressed
         #[arg(required = true)]
         snapshot_files: Vec<PathBuf>,
+        /// Fail at the first invalid snapshot
+        #[arg(long)]
+        fail_fast: bool,
```
ruseinov, line 81: nit: I reckon this has to be a required argument, it's best not to assume what the use-case is, especially if it means a potentially very long runtime.

src/tool/subcommands/snapshot_cmd.rs
qingyunha: done.  https://github.com/ChainSafe/forest/pull/4623#issue-2449995066

src/tool/subcommands/snapshot_cmd.rs
qingyunha: Do you prefer this `#[arg(long, required = true, action = clap::ArgAction::Set)]`

ruseinov: Isn't `long, required = true` enough?

ruseinov (approved):

src/tool/subcommands/snapshot_cmd.rs
ruseinov: in fact - feel free to disregard this. Let's assume the default behaviour is to validate all the snapshots, like it is now.

Issue #4626, opened 2024-08-07
Consensus error when syncing from an old snapshot


LesnyRumcajs:
## Issue summary

Forest 0a85621ddcccda981c566e92e0321cc710639474 fails to sync from an older snapshot `forest_snapshot_calibnet_2024-06-25_height_1732055` [link](https://forest-snapshots.fra1.cdn.digitaloceanspaces.com/debug/#4446/forest_snapshot_calibnet_2024-06-25_height_1732055.forest.car.zst)

All goes well, until it reaches recent epochs and then it starts failing:

```
2024-08-07T10:55:22.021060Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1823977, N blocks = 2
2024-08-07T10:55:22.028158Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1823978, N blocks = 3
2024-08-07T10:55:22.036197Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1823979, N blocks = 3
2024-08-07T10:55:22.048367Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1823980, N blocks = 2
2024-08-07T10:55:22.058184Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1823981, N blocks = 3
2024-08-07T10:55:22.068021Z  INFO forest_filecoin::chain_sync::tipset_syncer: Validating tipset: EPOCH = 1823982, N blocks = 3
2024-08-07T10:55:22.093961Z  WARN forest_filecoin::chain_sync::tipset_syncer: Validating block [CID = bafy2bzaceappribw33awrgwuwdm3zvcvjd7a3mwur4f3ir3mlw7qi3so2pgu2] in EPOCH = 1823982 failed: Validation error: Validation error: Consensus error: Failed to verify winning PoSt: Unsupported mapping from Invalid(18) to PoSt-winning RegisteredProof
2024-08-07T10:55:22.094253Z ERROR forest_filecoin::chain_sync::tipset_syncer: Sync messages check state failed for tipset range
[...]

Issue #4610, opened 2024-08-02
Fix `Filecoin.MinerGetBaseInfo`


LesnyRumcajs:
## Issue summary

The `Filecoin.MinerGetBaseInfo` is not working correctly and is in the filter list.

```
api-compare-1  |      {
api-compare-1  |
api-compare-1  | -      "SealProof": 8,
api-compare-1  |
api-compare-1  | +      "SealProof": 13,
api-compare-1  |
api-compare-1  |        "SealedCID": {
api-compare-1  |
api-compare-1  |          "/": "bagboea4b5abcaadexcxa334w7diouerr3pnrczesftikwrcrdx3jk5v33hacbz23"
api-compare-1  |
api-compare-1  |        },
api-compare-1  |
api-compare-1  | +      "SectorKey": null,
api-compare-1  |
api-compare-1  |        "SectorNumber": 25784
api-compare-1  |
api-compare-1  |      }
```

LesnyRumcajs:
Re-opened via https://github.com/ChainSafe/forest/pull/4627

PR #4629, opened 2024-08-08
fix: fix winning proofs

LesnyRumcajs:
## Summary of changes

- fixed consensus issues caused by using obsolete proof enums,
- I had to revive a function deleted in https://github.com/filecoin-project/ref-fvm/pull/1903 with some modifications `registered_winning_post_proof`
- re-applied the fix for `MinerGetBaseInfo` where the issue surfaced. 

Closes https://github.com/ChainSafe/forest/issues/4626
Closes https://github.com/ChainSafe/forest/issues/4610

## Change checklist

### DIFF
--- a/src/lotus_json/registered_po_st_proof.rs
+++ b/src/lotus_json/registered_po_st_proof.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use crate::shim::sector::RegisteredPoStProof;
-use fvm_shared3::sector::RegisteredPoStProof as RegisteredPoStProofV3;
+use fvm_shared4::sector::RegisteredPoStProof as RegisteredPoStProofV4;
 
 use super::*;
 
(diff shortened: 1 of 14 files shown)


src/lotus_json/registered_po_st_proof.rs
```diff
@@ -13,15 +13,15 @@ impl HasLotusJson for RegisteredPoStProof {
     fn snapshots() -> Vec<(serde_json::Value, Self)> {
         vec![(
             json!(0),
-            RegisteredPoStProof::from(RegisteredPoStProofV3::StackedDRGWinning2KiBV1),
+            RegisteredPoStProof::from(RegisteredPoStProofV4::StackedDRGWinning2KiBV1),
```
hanabi1224, line 16: Not sure if it's a good idea to use fvm_shared_latest::sector::RegisteredPoStProof here so that we won't miss it with the next fvm upgrade

hanabi1224 (approved):

src/lotus_json/registered_po_st_proof.rs
LesnyRumcajs: It's not a real issue, given enum discriminants for the proofs remain the same. That said, it'd make sense to get rid of fvmX occurrences from outside of `shim` and `fvmX` modules. I'll address it in a follow-up.

lemmih (approved):

PR #4636, opened 2024-08-09
chore(test): run go interop tests in the same rust process

hanabi1224:
## Summary of changes

This PR uses `rust2go` to invoke the `go-kad` test app in the rust process, to demonstrate how FFI for the f3 sidecar works.

Note that `go1.21.x` is required to run `forest-interop-test`

- Move `go-kad` compact test to `forest-interop-test` and invoke with `rust2go`
- Move `go-bitswap` compact test to `forest-interop-test` and invoke with `rust2go`
- Update CI workflow and Makefile accordingly

Test result on CI: https://github.com/ChainSafe/forest/actions/runs/10355850383/job/28664406918?pr=4636#step:6:2191
```
        PASS [   1.077s] forest-interop-tests tests::bitswap_go_compat::bitswap_go_compat_test
        PASS [   2.859s] forest-filecoin::lint lint
        PASS [   1.131s] forest-interop-tests tests::kad_go_compat::kad_go_compat_test
```

Closes

## Change checklist


interop-tests/src/tests/go_ffi.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright 2019-2024 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+pub mod binding {
+    #![allow(warnings)]
+    rust2go::r2g_include_binding!();
+}
+
+#[derive(rust2go::R2G, Clone, Default)]
+pub struct EmptyReq {}
```
hanabi1224, line 10: This is required due to a bug https://github.com/ihciah/rust2go/pull/10

interop-tests/src/tests/go_app/gen.go
```diff
@@ -0,0 +1,256 @@
+package main
+
+/*
+// Generated by rust2go. Please DO NOT edit this C part manually.
```
LesnyRumcajs, line 4: Should the generated file be included in the version control?

interop-tests/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "forest-interop-tests"
+version = "0.1.0"
+authors = ["ChainSafe Systems <info@chainsafe.io>"]
```
LesnyRumcajs, line 4: It's likely a better point of contact.

```suggestion
authors = ["ChainSafe Systems <forest@chainsafe.io>"]
```

LesnyRumcajs (commented):
The end binary doesn't require `go`, right? That is to say, if we ship the `forest` binary, there are no extra requirements (except the ones that are already listed in the Dockerfile)?

interop-tests/src/tests/go_app/gen.go
hanabi1224: I'm neutral about it. It's not required but makes code readable for users who solely browse the repo. What's your recommendation?

LesnyRumcajs: Pros:
- any change to it must be reviewed,
- somehow searchable.

Cons:
- will the changes to this file be actually reviewed or just blindly approved, especially when it'll be some obscure Go with C code?
- it introduces some noise, e.g., simple `Cargo.toml` or lockfiles are easily approved,
- while it allows some searching, it will also make general grepping more difficult - in my workflow, I'd need to disable looking in `.go` files to avoid noise.

All in all, if it's trivial to generate, I'd generate it on the fly and not add it to the version control.

hanabi1224: Removed

LesnyRumcajs (approved):

lemmih (approved):

ihciah:
Hi! I'm the author of rust2go. Thank you for choosing rustgo! In out production environment we have validated its relibility. 

One thing to note is that, remember to add `GODEBUG=invalidptr=0,cgocheck=0` env when run it, otherwise there maybe panic `invalid pointer found on stack` in go side when it tries to grow the goroutine stack([related code](https://github.com/golang/go/blob/master/src/runtime/stack.go#L640)). This env can also be added with golang code on init: `os.Setenv("GODEBUG", ...)`.

For testing, it doesn't matter whether there is this environment variable or not. But if you use it to serve requests, remember to add it.

hanabi1224:
> Hi! I'm the author of rust2go. Thank you for choosing rustgo! In out production environment we have validated its relibility.
> 
> One thing to note is that, remember to add `GODEBUG=invalidptr=0,cgocheck=0` env when run it, otherwise there maybe panic `invalid pointer found on stack` in go side when it tries to grow the goroutine stack([related code](https://github.com/golang/go/blob/master/src/runtime/stack.go#L640)). This env can also be added with golang code on init: `os.Setenv("GODEBUG", ...)`.
> 
> For testing, it doesn't matter whether there is this environment variable or not. But if you use it to serve requests, remember to add it.

@ihciah Thank you for the hints, this has been addressed in https://github.com/ChainSafe/forest/pull/4682

Issue #4628, opened 2024-08-07
Add devnet support to `forest-tool api serve` and document the functionality


lemmih:
## Issue summary

TDB

LesnyRumcajs:
These need to be ported to `forest-tool api serve` https://github.com/ChainSafe/forest/blob/ded933f18c652e9c38d65553881f9692901b50fd/scripts/devnet/docker-compose.yml#L200-L206

LesnyRumcajs:
Linked to #4598, given one is hardly useful without the other.

PR #4638, opened 2024-08-12
feat: offline server devnet

LesnyRumcajs:
## Summary of changes

- added support for devnets in the offline Forest.

Closes https://github.com/ChainSafe/forest/issues/4628

## Change checklist


scripts/devnet/check.sh
```diff
@@ -38,3 +39,10 @@ while true; do
 
   sleep 1
 done
+
+# Check the offline RPC, which should be initialized at that point. It should be at the genesis height, so 0.
+height=$(get_sync_height ${FOREST_OFFLINE_RPC_PORT})
```
hanabi1224, line 44: There is chance offline server has not become healthy here. How about moving `check.sh` to a docker compose step with proper depends-on setups

ruseinov: There is a 10 minute timeout on the retries. In my book that's enough, so I leave it up to you whether or not this goes into a separate docker container.

LesnyRumcajs: It will be spun right after Lotus has initialised without any need for a sync, without waiting for connecting to Lotus, nor going through 5 different migrations over 20 epochs.

If it's not up and running at that point, something is definitely not right with the offline server itself.

hanabi1224 (approved):

LesnyRumcajs:
@ruseinov @lemmih

lemmih (approved):

Issue #4528, opened 2024-07-15
RPC methods Forest needs to cover the Filecoin Common Node API


aatifsyed:
- [x] #4750
  - stubbed in #4677
- [x] #4532 (no longer required)
- [x] #4751
  - stubbed in #4677
- [x] #4549
- [x] #4546
- [x] #4548
- [x] #4562
- [x] #4752
  - stubbed in #4658
- [x] #4753
  - stubbed in #4658
- [x] `Filecoin.StateGetAllAllocations` #4501
- [x] `Filecoin.StateGetAllClaims` #4500
- [x] `Filecoin.StateGetAllocationForPendingDeal` #4567 
- [x] `Filecoin.StateGetAllocationIdForPendingDeal` #4568 
- [x] `Filecoin.StateGetRandomnessDigestFromBeacon` #4559
- [x] `Filecoin.StateGetRandomnessDigestFromTickets` #4563 
- [x] `Filecoin.StateLookupRobustAddress` #4509
- [x] #4557
- [x] `Filecoin.WalletSignMessage` #4514

ansermino:
Lets consider stubbing all remaining methods so can produce a complete openrpc spec

PR #4658, opened 2024-08-21
feat: stub NetProtectList and NetProtectRemove

aatifsyed:
As per https://github.com/ChainSafe/forest/issues/4528#issuecomment-2284051504

### DIFF
--- a/src/rpc/error.rs
+++ b/src/rpc/error.rs
@@ -31,7 +31,9 @@ impl ServerError {
     pub fn known_code(&self) -> ErrorCode {
         self.inner.code().into()
     }
-    pub fn stubbed_4528() -> Self {
+    /// We are only including this method to get the JSON Schemas for our OpenRPC
+    /// machinery
+    pub fn stubbed_for_openrpc() -> Self {
         Self::new(
             4528,
             "unimplemented",
--- a/src/rpc/error.rs
+++ b/src/rpc/error.rs
@@ -31,6 +31,16 @@ impl ServerError {
     pub fn known_code(&self) -> ErrorCode {
         self.inner.code().into()
     }
+    pub fn stubbed_4528() -> Self {
+        Self::new(
+            4528,
+            "unimplemented",
+            Some(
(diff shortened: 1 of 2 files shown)


hanabi1224 (approved):

src/rpc/error.rs
```diff
@@ -31,6 +31,16 @@ impl ServerError {
     pub fn known_code(&self) -> ErrorCode {
         self.inner.code().into()
     }
+    pub fn stubbed_4528() -> Self {
```
LesnyRumcajs, line 34: I don't think adding issues' numbers helps in code readability.
```suggestion
    pub fn stubbed() -> Self {
```

LesnyRumcajs: Does it need to be issue specific? Can we have a common method for stubs (that may or may not accept optional issue for errors)? Otherwise, I wouldn't put in in a public API of the `ServerError`.

aatifsyed: > Can we have a common method for stubs

At the moment there's only one issue, if there are future methods which require stubbing and method tracking, I'm happy to have a common `stubbed` method.

> that may or may not accept optional issue for errors

Without the above, the callsite changes from `stubbed_4528` to `stubbed(4528)`, which I don't think addresses your initial comment

> I don't think adding issues' numbers helps in code readability.

> Otherwise, I wouldn't put in in a public API of the ServerError.

I don't know what you mean by this - I think it's helpful for us and users to have an issue number they can ping if they e.g want a specific method implemented.

---

How about I change this to `stubbed_for_openrpc`?

LesnyRumcajs: > Without the above, the callsite changes from stubbed_4528 to stubbed(4528), which I don't think addresses your initial comment

This would actually address my concern. As a developer, I don't want to see a `ServerError::stubbed_XYZ`. I want to see only methods that are usable for me.


> I don't know what you mean by this - I think it's helpful for us and users to have an issue number they can ping if they e.g want a specific method implemented.

I agree it's helpful - in the logs, return messages, but on in the public interface of the `ServerError`.

LesnyRumcajs (approved):

PR #4679, opened 2024-08-22
feat(rpc) Implement `Filecoin.MarketAddBalance`

### LINKED ISSUE #4612: `MarketAddBalance`

elmattic:
## Summary of changes

- implement RPC method `Filecoin.MarketAddBalance`
- calibnet script test

Closes #4612

## Change checklist


src/rpc/methods/market.rs
```diff
@@ -0,0 +1,44 @@
...
+use fvm_ipld_encoding::RawBytes;
+use num_bigint::BigInt;
+
+pub const METHOD_ADD_BALANCE: u64 = 2;
+
+pub enum MarketAddBalance {}
+impl RpcMethod<3> for MarketAddBalance {
+    const NAME: &'static str = "Filecoin.MarketAddBalance";
+    const PARAM_NAMES: [&'static str; 3] = ["wallet", "address", "amount"];
+    const API_PATHS: ApiPaths = ApiPaths::V0;
```
LesnyRumcajs, line 19: Should it be rather V1?

src/rpc/methods/market.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::rpc::error::ServerError;
+use crate::rpc::mpool::MpoolPushMessage;
+use crate::rpc::{ApiPaths, Ctx, Permission, RpcMethod};
+use crate::shim::{address::Address, message::Message};
+use cid::Cid;
+use fvm_ipld_blockstore::Blockstore;
+use fvm_ipld_encoding::RawBytes;
+use num_bigint::BigInt;
+
+pub const METHOD_ADD_BALANCE: u64 = 2;
```
LesnyRumcajs, line 13: Does it need to be public?

scripts/tests/calibnet_wallet_rpc_check.sh
```diff
@@ -0,0 +1,73 @@
...
+  exit 1
+fi
+
+MSG_CID=$(echo "$JSON" | jq -r '.result["/"]')
+echo "Message cid: $MSG_CID"
+
+# Try 10 times.
+for i in {1..10}
+do
+  sleep 30s
```
LesnyRumcajs, line 55: Should we rather have more retries and less sleeps? This would make the test pass faster.

elmattic: This duration was chosen because Filecoin has a block time of 30 seconds.

I will see what I can do to speed up the test.

src/rpc/methods/market.rs
elmattic: No.

scripts/tests/calibnet_wallet_rpc_check.sh
LesnyRumcajs: Usually, we won't end up at the block boundary, so it is possible that the message will be there in less than 30 seconds (or less than 60 seconds, e.g., after 35s). Is it correct?

elmattic: Yes, correct. I've moved to a 5s turnaround time.

elmattic: 10m 59s ➡️  9m 55s. Still most of the time is dominated by the setup.

LesnyRumcajs: Can we re-use the existing wallet check? That would help a lot with the setup and burn less resources in process. https://github.com/ChainSafe/forest/blob/main/scripts/tests/calibnet_wallet_check.sh

elmattic: Right.

LesnyRumcajs (approved):

hanabi1224 (approved):

PR #4729, opened 2024-09-04
chore(ci): upgrade runner images to ubuntu-24.04

hanabi1224:
## Summary of changes

- upgrade CI runner images to ubuntu:24.04 (except buildjet ones where 2404 is not yet available)

Closes

## Change checklist

### DIFF
--- a/scripts/devnet/forest_ci.dockerfile
+++ b/scripts/devnet/forest_ci.dockerfile
@@ -1,4 +1,5 @@
-FROM ubuntu:22.04
+# The version has to match Github CI runner images
+FROM ubuntu:24.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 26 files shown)


scripts/devnet/forest_ci.dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM ubuntu:22.04
+# The version has to match Github CI runner images
```
hanabi1224, line 1: We'd better use explict ubuntu version in [CI].yml otherwise CI might be broken someday when `ubuntu-latest` starts pointing to `ubuntu-24.04`

LesnyRumcajs (approved):
Would it make sense to have a follow-up that would bump the Forest Dockerfile's base image as well?

aatifsyed (approved):

Issue #4716, opened 2024-09-02
Setup Curio integration in the CI


LesnyRumcajs:
## Issue summary

Curio is a replacement for `lotus-miner`. Like `lotus-miner`, it is designed to use `lotus` to interact with the Filecoin blockchain. However, Curio uses a much smaller set of RPC methods, and Forest is nearly capable of serving as the blockchain gateway. Unfortunately, there are still several bugs that prevent mining with Forest. This issue tracks setting up automated testing of mining with Forest as the blockchain gateway. This issue does _not_ cover solving the remaining bugs and compatibility problems.

- [x] Merge Curio docker-compose with our devnet docker-compose.
- [ ] Automatically add sectors to be mined.
- [ ] Verify that Curio behaves correctly with Forest as the interface.
- [ ] Add additional stress testing.
- [ ] Find a way to reproduce the bugs encountered by the Curio team.

Clean-up and factoring:
- [ ] De-duplicate common files.
- [ ] Consider if having multiple `devnet` setups makes sense.

PR #4760, opened 2024-09-11
ci: devnet setup with curio

lemmih:
## Summary of changes

- Basic devnet with Curio + Forest.
- Mining does not yet work with Forest. This will be addressed in subsequent PRs.
- This `devnet` setup is not tested in CI since mining doesn't yet work.

Part of #4716

## Change checklist


scripts/devnet-curio/docker-compose.yml
```diff
@@ -0,0 +1,363 @@
...
+      - filecoin-proofs:/var/tmp/filecoin-proof-parameters:rw
+      - ./data/lotus:${LOTUS_DATA_DIR}
+      - ./data/forest:${FOREST_DATA_DIR}
+    env_file:
+      - curio.env
+    networks:
+      - devnet
+    entrypoint: [ "/bin/bash", "-c" ]
+    command:
+      - |
```
ruseinov, line 281: nit: this command here seems way too long and convoluted to be embedded into `docker-compose.yaml`. However I also understand that it might be nice to be able to read the logic directly as opposed to navigating to the corresponding bash script. Debugging is definitely easier via a separate script though.

I leave it up to you, I'd likely have gone with a dedicated `.sh` file.

scripts/devnet-curio/setup.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Wait for Forest to be ready. We can assume that it is ready when the
+# RPC server is up. This checks if Forest's RPC endpoint is up.
+ function call_forest_chain_head {
+    curl --silent -X POST -H "Content-Type: application/json" \
+         --data '{"jsonrpc":"2.0","id":2,"method":"Filecoin.NetPeers","params":"null"}' \
+         "http://127.0.0.1:${FOREST_RPC_PORT}/rpc/v0"
+ }
+ 
+ until call_forest_chain_head; do
```
ruseinov, line 33: nit: what happens if it's never ready? perhaps we should have some sort of a timeout here, even if large, that would shut the whole thing down as opposed to spinning forever.

scripts/devnet-curio/docker-compose.yml
lemmih: Why is debugging easier with a separate script?

What's a good way to run a script here?

scripts/devnet-curio/setup.sh
lemmih: Timeouts are to be handled outside the script.

scripts/devnet-curio/.env
```diff
@@ -0,0 +1,24 @@
...
+GENESIS_NETWORK_VERSION=18
+SHARK_HEIGHT=-10
+HYGGE_HEIGHT=-9
+LIGHTNING_HEIGHT=3
+THUNDER_HEIGHT=6
+WATERMELON_HEIGHT=9
+DRAGON_HEIGHT=12
+WAFFLE_HEIGHT=18
+DRAND_QUICKNET_HEIGHT=15
+TARGET_HEIGHT=24
```
LesnyRumcajs, line 24: For the sake of simplicity, it might make sense to start directly from NV23 and avoid migrations.

scripts/devnet-curio/.env
```diff
@@ -0,0 +1,24 @@
+CURIO_IMAGE=curio/curio-dev:dev
```
LesnyRumcajs, line 1: It would be safer to pin it to a concrete hash.

LesnyRumcajs: Or is it a local docker build?

lemmih: It's local. They do not upload their docker images. It's written in the README.md how to build the `curio/curio-dev` image.

scripts/devnet-curio/docker-compose.yml
ruseinov: If I had to debug this - I'd mount this script and let the entrypoint be just some unrelated long-running process to keep the container going. Then I would `docker ssh` and debug the script by running and modifying it as needed. 
Normally this should be faster than having to restart the container or the whole compose setup.

scripts/devnet-curio/.env
LesnyRumcajs: Would it make sense to make _our_ build of curio an upload it, along the lines of https://github.com/ChainSafe/forest/blob/main/.github/workflows/lotus-devnet-publish.yml?

This would allow embedding it in the CI. Otherwise, it's rather not reproducible.

scripts/devnet-curio/setup.sh
ruseinov: fair enough, non-blocking.

scripts/devnet-curio/.env
lemmih: Yes, this needs to be done eventually.

scripts/devnet-curio/README.md
```diff
@@ -0,0 +1,81 @@
...
+Stop the devnet with:
+
+```shell
+docker compose down
+```
+
+Remove the devnet:
+
+```shell
+docker compose rm
```
LesnyRumcajs, line 50: I'd usually do `❯ docker compose down --rmi all -v` to be thorough.

scripts/devnet-curio/check.sh
LesnyRumcajs, line 1: It might be worth noting that it's a copy of the same script from the `devnet` compose.

ruseinov (approved):
looks good to me!

some non-blocking nits

scripts/devnet-curio/docker-compose.yml
LesnyRumcajs: I agree that it would be beneficial to move it into a script.

1. Compose parts should ideally live on a similar abstraction level. It should be possible to take a look and assess what different containers are doing. This is not possible with a lengthy script exposing internal logic.
2. Shell scripts inside compose files are not linted by the shellcheck.

scripts/devnet-curio/forest_ci.dockerfile
LesnyRumcajs, line 1: Does it have to be copied? Can we point to the existing files in the `devnet/`?

LesnyRumcajs: Same applies for some of the other files, e.g., Lotus dockerfile or `check.sh`.

scripts/devnet-curio/docker-compose.yml
lemmih: How do you move it to a script? I also asked Roman and didn't get an answer. You Docker gods have to take pity on a Docker peasant like me. I failed to Google an answer, and I see no examples in our docker-compose files.

LesnyRumcajs: We do it in the sync check. https://github.com/ChainSafe/forest-iac/blob/093f9965ad0e148f3b5cc558e78720e9df3228e8/tf-managed/modules/sync-check/service/docker-compose.yml#L91-L97

Please let me know if you need any help with that.

lemmih: Done.

LesnyRumcajs: Oh look, linter errors! No green checkmark, no review!!!

scripts/devnet-curio/forest_ci.dockerfile
lemmih: I don't think files should be shared between the two devnet configurations.

scripts/devnet-curio/check.sh
lemmih: The Curio devnet will eventually replace the `lotus-miner` devnet, right? So, eventually, we'll only have a single copy of the file.

scripts/devnet-curio/forest_ci.dockerfile
LesnyRumcajs: That said, it might make more sense to put them in a common directory. We can do it in a follow-up. All in all, if possible, I'd like to avoid duplication.

LesnyRumcajs (approved):

scripts/devnet-curio/docker-compose.yml
ruseinov: > Why is debugging easier with a separate script?
> 
> What's a good way to run a script here?

I didn't understand the question, excuses. You just make sure the script is in the filesystem and then run it :)

PR #4780, opened 2024-09-19
feat(rpc) Implement `Filecoin.EthGetLogs`

elmattic:
## Summary of changes

- Partially implement RPC method `Filecoin.EthGetLogs`

```
./forest-tool api compare --lotus /ip4/127.0.0.1/tcp/1234/http --forest /ip4/127.0.0.1/tcp/2345/http forest_snapshot_calibnet_2024-10-03_height_2021383.forest.car.zst --filter EthGetLogs -n 800

| RPC Method                | Forest            | Lotus |
|---------------------------|-------------------|-------|
| Filecoin.EthGetLogs (272) | CustomCheckFailed | Valid |
| Filecoin.EthGetLogs (528) | Valid             | Valid |
```

This PR introduces a new flag/configuration setting (`enable_event_caching`) to retain events in the Tipset cache.

If neither the RPC API nor the methods related to events are used, a value of false can be passed.

Closes

## Change checklist

- [x] I have made corresponding changes to the documentation. All new code adheres to the team's [documentation standards](https://www.notion.so/chainsafe/Documentation-practices-1a7ec5e5486b4a56b26d9df6b1677a63),


src/rpc/methods/eth/filter/mod.rs
```diff
@@ -186,6 +195,178 @@ impl EthEventHandler {
...
+
+        let StateOutput { events, .. } = ctx.state_manager.tipset_state_plus(tipset).await?;
+
+        let events = if let Some(events) = events {
+            events
+        } else {
+            tracing::warn!("No events stored");
+            return Ok(());
+        };
+        for (i, (message, events)) in messages.iter().zip(events.into_iter()).enumerate() {
```
hanabi1224, line 231: Should we assert on `message.len() == events.len()`?

src/state_manager/mod.rs
```diff
@@ -1668,16 +1718,21 @@ where
 
     // FVM requires a stack size of 64MiB. The alternative is to use `ThreadedExecutor` from
     // FVM, but that introduces some constraints, and possible deadlocks.
-    stacker::grow(64 << 20, || -> anyhow::Result<(Cid, Cid)> {
+    stacker::grow(64 << 20, || -> anyhow::Result<StateOutput> {
```
hanabi1224, line 1721: This is much better than Cid tuple!

hanabi1224 (approved):

src/rpc/methods/eth/filter/mod.rs
elmattic: Great idea! This shouldn't happen in theory, but we can still handle it with an error, just to be on the safe side.

src/state_manager/mod.rs
```diff
@@ -1582,7 +1627,8 @@ pub fn apply_block_messages<DB>(
     tipset: Arc<Tipset>,
     mut callback: Option<impl FnMut(MessageCallbackCtx<'_>) -> anyhow::Result<()>>,
     enable_tracing: VMTrace,
-) -> Result<CidPair, anyhow::Error>
+    store_events: bool,
```
LesnyRumcajs, line 1630: I'd recommend not using a boolean here - the method has already plenty of parameters and it's cumbersome for the reader what is `foo(true, false)`. A nicer way is above, with an `enum` - it'd be more readable to have `apply_block_messages(..., StoreEvents::Enabled)`. Additionally, it's more extendible.

What do you think?

elmattic: I agree. We can either use another `enum`. But could we make use of `bitflags`? (and shave off on param numbers).

elmattic: Actually, an enum is the best fit here.

src/rpc/methods/eth/filter/mod.rs
```diff
@@ -186,6 +195,178 @@ impl EthEventHandler {
...
+                            "resolving address {} failed (EPOCH = {})",
+                            id_addr,
+                            tipset.epoch()
+                        )
+                    });
+                if result.is_err() {
+                    // Skip event
+                    continue;
+                };
+                let resolved = result.expect("Infallible");
```
LesnyRumcajs, line 252: nit: might be less scary and no need to check out the previous chunk of code to be sure it doesn't panic.

```suggestion
                let Ok(resolved) = result {
                    resolved
                } else {
                    continue;
                }
```

LesnyRumcajs:
Can we have unit tests for this implementation?

elmattic:
> Can we have unit tests for this implementation?

Sure.

ruseinov (approved):

elmattic:
@ruseinov Not too fast, buster!

elmattic:
@LesnyRumcajs added a few unit tests to cover the most intricated functions

src/shim/executor.rs
```diff
@@ -195,6 +211,144 @@ impl From<Receipt_v3> for Receipt {
...
+    pub fn entries(&self) -> Vec<Entry> {
+        match self {
+            Self::V3(v3) => v3.entries.clone().into_iter().map(Into::into).collect(),
+            Self::V4(v4) => v4.entries.clone().into_iter().map(Into::into).collect(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum StampedEvent {
```
LesnyRumcajs, line 307: Could we add some docs on the public items such as this `StampedEvent`?

src/shim/executor.rs
```diff
@@ -195,6 +211,144 @@ impl From<Receipt_v3> for Receipt {
...
+
+impl StampedEvent {
+    #[allow(dead_code)]
+    pub fn get_events(db: &impl Blockstore, events_cid: &Cid) -> anyhow::Result<Vec<StampedEvent>> {
+        let mut events = Vec::new();
+        if let Ok(amt) = Amt::<StampedEvent_v4, _>::load(events_cid, db) {
+            amt.for_each(|_, event| {
+                events.push(event.clone().into());
+                Ok(())
+            })?;
```
LesnyRumcajs, line 332: Isn't `for_each` deprecated in favour of `iter()`?

https://github.com/filecoin-project/ref-fvm/blob/488f704015a83de0db1737f5f47305939ea74f86/ipld/amt/src/amt.rs#L451-L462

src/state_manager/mod.rs
```diff
@@ -247,6 +266,11 @@ where
             beacon,
             chain_config,
             sync_config,
+            enable_event_caching: if enable_event_caching {
+                EventCache::Cached
+            } else {
+                EventCache::NotCached
+            },
```
LesnyRumcajs, line 273: Design idea - I wonder if we can make the code more straightforward by injecting an trait object of `EventCache`. It would be instantiated at the start of the program to either an implementation without cache or with a cache. This way, throughout the code, you wouldn't need to check if the caching is enabled or not - you'd write to it but whether it would actually happen would be completely transparent.

What do you think?

src/rpc/methods/eth.rs
```diff
@@ -2252,9 +2261,214 @@ impl RpcMethod<1> for EthGetTransactionReceipt {
...
+
+fn match_key(key: &str) -> Option<usize> {
+    match key.get(0..2) {
+        Some("t1") => Some(0),
+        Some("t2") => Some(1),
+        Some("t3") => Some(2),
+        Some("t4") => Some(3),
+        _ => None,
+    }
+}
```
LesnyRumcajs, line 2284: Will this work on mainnet?

src/rpc/methods/eth.rs
```diff
@@ -2381,4 +2595,302 @@ mod test {
...
+    fn test_eth_log_from_event() {
+        let entries = vec![
+            EventEntry {
+                flags: (Flags::FLAG_INDEXED_ALL).bits(),
+                key: "t1".into(),
+                codec: IPLD_RAW,
+                value: vec![
+                    226, 71, 32, 244, 92, 183, 79, 45, 85, 241, 222, 235, 182, 9, 143, 80, 241, 11,
+                    81, 29, 171, 138, 125, 71, 196, 129, 154, 8, 220, 208, 184, 149,
+                ]
```
LesnyRumcajs, line 2654: It'd be great to document where those test vectors come from.

src/shim/executor.rs
elmattic: Sure!

src/rpc/methods/eth.rs
elmattic: Yes. `"t1"`, `"t2"`, etc, all refer to topics. Those are hardcoded keys: https://github.com/filecoin-project/builtin-actors/blob/master/actors/evm/src/interpreter/instructions/log_event.rs#L15

src/state_manager/mod.rs
elmattic: Excellent idea! In the future we'll likely have:
- No cache (e.g., a node without RPC)
- Cache Events in tipset state cache
- Use sidecar indexer to retrieve Events

src/shim/executor.rs
elmattic: Good catch.

elmattic: Let's just remove this. It's deadcode anyway.

src/state_manager/mod.rs
elmattic: I've updated the code to store events in a dedicated cache (with a dedicated size), avoiding the need to store unnecessary `None` values in the SM tipset state cache when events are not in use. This also prevents polluting the SM cache and avoids potential performance slowdowns.

As for injecting an trait object, I think it would add some complexity compared to using an `enum` at this stage. Let's consider that for a future PR, or once we have an indexer in place?

src/rpc/methods/eth.rs
elmattic: These payloads correspond to existing topics on Calibnet, but they could just as easily be vectors filled with random bytes, as long as they are 32 bytes.

LesnyRumcajs: The comment that you added is also okay.

LesnyRumcajs (approved):

src/state_manager/mod.rs
elmattic: One more thing, we can remove the `--enable-event-caching` flag now.

PR #4886, opened 2024-10-12
[automated] Update Lotus version in API tests

LeshyBot:
### Changes
- Updates Lotus version in the JSON-RPC API tests to the latest release.


lemmih:
Lotus 1.30 is buggy. We'll wait for 1.31.

LesnyRumcajs:
@lemmih Leshy is not as smart as the dependabot, he'll open the same PR soon. Best is to keep it open, just mark it as do not merge.

PR #4895, opened 2024-10-15
fix: add the limit check when identify protocol received

ai-chen2050:
## Summary of changes

- Identify discovery protocol that will break up the limitations of the target_peer_count.
- So I add the limit checking at that point.

Closes

## Change checklist

### DIFF
--- a/src/libp2p/discovery.rs
+++ b/src/libp2p/discovery.rs
@@ -429,6 +429,10 @@ impl NetworkBehaviour for DiscoveryBehaviour {
                                     Some(info.clone());
                                 if let Some(kademlia) = self.discovery.kademlia.as_mut() {
                                     for address in &info.listen_addrs {
+                                        if self.n_node_connected >= self.target_peer_count {
+                                            // Already over discovery max, don't add new peers.
+                                            continue;
+                                        }
                                         kademlia.add_address(peer_id, address.clone());
                                     }
                                 }


src/libp2p/discovery.rs
```diff
@@ -429,6 +429,10 @@ impl NetworkBehaviour for DiscoveryBehaviour {
                                     Some(info.clone());
                                 if let Some(kademlia) = self.discovery.kademlia.as_mut() {
                                     for address in &info.listen_addrs {
+                                        if self.n_node_connected >= self.target_peer_count {
```
hanabi1224, line 432: This should not be inside the for loop

hanabi1224: Thanks for your contribution! However, with the current Kademlia config, addresses are automatically inserted into the routing table (see https://docs.rs/libp2p-kad/0.46.2/libp2p_kad/enum.BucketInserts.html). So I doubt if the change in this PR makes any difference. It'd be great to have unit tests for this scenario (we have not got a chance to prioritize the work yet)

> Whenever a connection to a peer is established as a result of a dialing attempt and that peer is not yet in the routing table, it is inserted as long as there is a free slot in the corresponding k-bucket. If the k-bucket is full but still has a free pending slot, it may be inserted into the routing table at a later time if an unresponsive disconnected peer is evicted from the bucket.

LesnyRumcajs:
Closing for now. Feel free to re-open once the concerns raised by @hanabi1224 are addressed!

PR #4908, opened 2024-10-16
[TEST] Testing fvm update perf regression

Stebalien:
Test PR to run CI against the FVM v2.8 -> v2.9 update to reproduce https://github.com/filecoin-project/ref-fvm/issues/2058.


Stebalien:
I've confirmed that importing _just_ wasmtime is causing the issue. I'm now trying to import `object` with the `write_core` feature enabled to see if feature unification is causing the issue. If it's not that, the next step is to check gimli.

Stebalien:
It doesn't appear to be `object`. I didn't test gimli because it's _only_ used by wasmtime v24 so it shouldn't affect feature unification.

Stebalien:
(testing on my own fork for now)

PR #4991, opened 2024-11-21
feat(rpc) Implement `Filecoin.EthTraceBlock`

### LINKED ISSUE #4708: EthTraceBlock

elmattic:
## Summary of changes

- implement RPC method `Filecoin.EthTraceBlock`
- API compare tests
- fix bug in `eth::decode_payload` and add unit tests

```
$ forest-tool api compare --lotus /ip4/127.0.0.1/tcp/1234/http --forest /ip4/127.0.0.1/tcp/2345/http forest_snapshot_calibnet_2024-12-16_height_2233655.forest.car.zst --filter EthTraceBlock -n 200
| RPC Method                   | Forest | Lotus |
|------------------------------|--------|-------|
| Filecoin.EthTraceBlock (200) | Valid  | Valid |
```

Closes #4708

## Change checklist


src/rpc/methods/eth/utils.rs
```diff
@@ -0,0 +1,165 @@
...
+
+use anyhow::{bail, Result};
+use cbor4ii::core::dec::Decode as _;
+use cbor4ii::core::Value;
+use fvm_ipld_blockstore::Blockstore;
+use fvm_ipld_encoding::{RawBytes, CBOR, DAG_CBOR, IPLD_RAW};
+use serde::de;
+
+/// The Identity multicodec code.
+pub const IDENTITY: u64 = 0x0;
```
LesnyRumcajs, line 17: It's not available anywhere to import?

src/shim/error.rs
```diff
@@ -22,11 +25,63 @@ use serde::{Deserialize, Serialize};
...
+            _ => None,
+        };
+        if let Some(name) = name {
+            write!(f, "{}({})", name, self.value())
+        } else {
+            match self.value() {
+                code if code > ExitCode_latest::SYS_MISSING_RETURN.value()
+                    && code < ExitCode_latest::FIRST_USER_EXIT_CODE =>
+                {
+                    write!(f, "SysErrReserved{}({})", code - 10, code)
```
LesnyRumcajs, line 70: Why subtract 10?

src/rpc/methods/eth/trace.rs
```diff
@@ -0,0 +1,624 @@
...
+use crate::rpc::methods::eth::lookup_eth_address;
+use crate::rpc::methods::state::ExecutionTrace;
+use crate::rpc::state::ActorTrace;
+use crate::shim::{actors::is_evm_actor, address::Address, error::ExitCode, state_tree::StateTree};
+use fil_actor_eam_state::v12 as eam12;
+use fil_actor_evm_state::v15 as evm12;
+use fil_actor_init_state::v12::ExecReturn;
+use fil_actor_init_state::v15::Method as InitMethod;
+use fvm_ipld_blockstore::Blockstore;
+use fvm_shared4::{error::ExitCode as ExitCodeV4, METHOD_CONSTRUCTOR};
```
LesnyRumcajs, line 22: Do we have to lock the actor version and the FVM version here? It would be great to avoid it and work through actor interface or shims.

src/rpc/methods/eth/utils.rs
elmattic: Unfortunately no. However, we could consider adding it to [fvm_ipld_encoding](https://github.com/filecoin-project/ref-fvm/blob/master/ipld/encoding/src/lib.rs).

src/shim/error.rs
elmattic: We want to align errors with Lotus display, So `code - 10` does that.

Maybe using `SYS_ASSERTION_FAILED` code instead? (it equals to 10).

src/rpc/methods/eth/trace.rs
elmattic: I can have a look.

src/rpc/methods/eth/utils.rs
LesnyRumcajs: What about https://github.com/filecoin-project/ref-fvm/blob/e9eef91d173d2cc2cb464c7f988fa2a6cc82577b/shared/src/lib.rs#L59-L60?

src/shim/error.rs
LesnyRumcajs: Do you know why Lotus subtracts 10? I'm okay with having it as well. Just please comment in the code the reason for this so the magic number is less magical.

src/rpc/methods/eth/utils.rs
elmattic: Thanks!

src/shim/error.rs
elmattic: Sure, I will explain in a comment the rational behind this.

elmattic: I think there is a better, cleaner and more correct solution to match Lotus exit codes.

We just need to create a PR in ref-fvm to uncomment those dead lines:
https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/error/mod.rs#L64
https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/error/mod.rs#L74
https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/error/mod.rs#L81-L84

And add those variant to our match expression in `forest/src/shim/error.rs`.

elmattic: Let's address that in a subsequent PR.

LesnyRumcajs (commented):
In general, does Lotus have a more extensive suite for this tracing? If so, could we migrate some of those tests?

elmattic:
> In general, does Lotus have a more extensive suite for this tracing? If so, could we migrate some of those tests?

Unfortunately, no. They have very few tests. I've already migrated the one that tests `decode_payload` free function.

LesnyRumcajs:
> > In general, does Lotus have a more extensive suite for this tracing? If so, could we migrate some of those tests?
> 
> Unfortunately, no. They have very few tests. I've already migrated the one that tests `decode_payload` free function.

Well, then, fortunately, I'm a man of great faith. :pray:

LesnyRumcajs (approved):

hanabi1224 (approved):

hanabi1224 (approved):

PR #5071, opened 2024-12-10
fix: fetch tipsets in a single request

lemmih:
## Summary of changes

- Tipset messages were being fetched individually and from the general swarm. This PR changes the logic to fetch the messages as a single request directed to the owner of the tipset. This fixes an issue where we couldn't fetch the messages fast enough, and Forest would slowly fall out of sync.

Closes

## Change checklist

### DIFF
--- a/src/chain_sync/chain_muxer.rs
+++ b/src/chain_sync/chain_muxer.rs
@@ -24,18 +24,14 @@ use crate::libp2p::{
 };
 use crate::message::SignedMessage;
 use crate::message_pool::{MessagePool, Provider};
-use crate::shim::{clock::SECONDS_IN_DAY, message::Message};
+use crate::shim::clock::SECONDS_IN_DAY;
 use crate::state_manager::StateManager;
 use crate::{
-    blocks::{Block, CreateTipsetError, FullTipset, GossipBlock, Tipset, TipsetKey},
+    blocks::{Block, CreateTipsetError, FullTipset, Tipset, TipsetKey},
     networks::calculate_expected_epoch,
 };
 use cid::Cid;
-use futures::{
-    future::{try_join_all, Future},
-    stream::FuturesUnordered,
-    try_join, StreamExt,
-};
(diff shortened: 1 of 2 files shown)


lemmih:
I will do a patch release once this is reviewed and merged.

LesnyRumcajs (approved):
I'd love if our leading P2P expert could pronounce if we're not missing anything here. @hanabi1224

src/chain_sync/chain_muxer.rs
```diff
@@ -454,8 +402,13 @@ where
                         return Ok(None);
                     }
                     // Assemble full tipset from block only in stateful mode
-                    let tipset =
-                        Self::gossipsub_block_to_full_tipset(b, source, network.clone()).await?;
+                    let tipset = Self::get_full_tipset(
+                        network.clone(),
+                        chain_store.clone(),
+                        source,
```
hanabi1224, line 408: In the previous logic, bitswap reqeusts are sent to all connected peers, should we fallback to non-source peers if it fails here?

lemmih: No, if a peer tells us about a new block without having it themselves, we should ignore it.

hanabi1224: Oh wait, as I understand, in gossipsub source could a proxy peer that does not actually own the block

hanabi1224: https://docs.rs/libp2p-gossipsub/latest/libp2p_gossipsub/enum.Event.html#variants

> propagation_source: [PeerId](https://docs.rs/libp2p-identity/0.2.9/x86_64-unknown-linux-gnu/libp2p_identity/peer_id/struct.PeerId.html)
The peer that forwarded us this message.

lemmih: "Fixed" by broadcasting the chain_exchange request. Eventually, all of this code needs to be reworked.

hanabi1224 (approved):

Issue #5108, opened 2025-01-06
Forest node on Calibnet gets stuck at epoch 2281645


hanabi1224:
## Describe the bug

and calibnet snapshot has not been published since Jan.1

```
2025-01-06T06:46:16.724863Z ERROR compute_tipset_state_blocking: forest_filecoin::interpreter::fvm4: database lookup error: cannot get worker key (current epoch: 2281645, height: 2280353, miner address: t0143103)
```

## To reproduce

2. Run '....'
3. See error

## Log output

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
  paste log output...
  ```
</details>

## Environment (please complete the following information):

- OS:
- Branch/commit
- Hardware

PR #5109, opened 2025-01-06
fix: block validation error at epoch 2281645

hanabi1224:
## Summary of changes

- fix logic to match Lotus
- bump Forest version

https://github.com/filecoin-project/lotus/blob/v1.32.0-rc2/chain/vm/fvm.go#L169
```go
	// (4) expensive final checks

	// check blocks are properly signed by their respective miner
	// note we do not need to check extra's: it is a parent to block b
	// which itself is signed, so it was willingly included by the miner
	gasA, sigErr := x.verifyBlockSig(ctx, &blockA)
	totalGas += gasA
	if sigErr != nil {
		log.Info("invalid consensus fault: cannot verify first block sig: %w", sigErr)
		return ret, totalGas
	}

	gas2, sigErr := x.verifyBlockSig(ctx, &blockB)
	totalGas += gas2
	if sigErr != nil {
		log.Info("invalid consensus fault: cannot verify second block sig: %w", sigErr)
		return ret, totalGas
	}
```

Closes https://github.com/ChainSafe/forest/issues/5108

## Change checklist

### DIFF
--- a/.config/lychee.toml
+++ b/.config/lychee.toml
@@ -4,17 +4,12 @@
 
 verbose = "debug"
 no_progress = true
-exclude_path = [
-    "./node_modules",
-    "./docs/node_modules",
-    "./documentation",
-    "./target"
-]
+exclude_path = ["./node_modules", "./docs/node_modules", "./documentation", "./target"]
 exclude = [
-    # Avoid Github rate limits
-    "github.com/ChainSafe/forest",
-    # Requires CAPTCHA verification
-    "faucet.calibnet.chainsafe-fil.io/funds.html"
+  # Avoid Github rate limits
+  "github.com/ChainSafe/forest",
+  # Requires CAPTCHA verification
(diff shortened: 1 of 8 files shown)


sudo-shashank (approved):

src/interpreter/fvm4.rs
LesnyRumcajs, line 1: Should this be reflected in FVM2 and FVM3 as well?

hanabi1224: I think so, fixed.

LesnyRumcajs (approved):

.config/lychee.toml
```diff
@@ -4,20 +4,15 @@
 
 verbose = "debug"
 no_progress = true
-exclude_path = [
```
hanabi1224, line 7: lint check job requires this format change

LesnyRumcajs (approved):

PR #5235, opened 2025-02-05
Fix block number or hash usage

### LINKED ISSUE #5125: Fix BlockNumberOrHash usage
## Describe the bug

`getTipsetByEthBlockNumberOrHash` is a function in **Lotus** used to retrieve a tipset. It accepts predefined strings such as "pending" or "latest". However, using "earliest" will result in an error.

This function is utilized in several Ethereum endpoints:

- EthGetTransactionCount
- EthGetBlockReceiptsLimited
- EthGetCode
- EthGetStorageAt
- EthGetBalance
- EthEstimateGas
- EthCall

In **Forest**, a similar function exists called `tipset_by_block_number_or_hash`. The key difference is that Forest also supports additional predefined strings such as "safe" and "finalized".

This function is used by:

- EthGetTransactionCount
- EthGetCode
- EthGetStorageAt
- EthGetBalance
- EthEstimateGas
- EthCall

And

- EthFeeHistory
- EthGetBlockByHash
- EthGetBlockByNumber
- EthGetTransactionByBlockNumberAndIndex
- EthTraceBlock

For the second list of functions, support for "safe" and "finalized" might be required. However, Forest is likely to diverge from Lotus in the first list.

To address these differences, we should:

1) **Use clear types**: Split the Predefined type into subtypes where necessary. For instance, functions like `EthGetBlockByNumber` should only accept the required subtype (e.g., a hexadecimal block number).

2) **Expand API coverage**: Ensure the API supports all possible values for predefined strings, including both accepted and rejected values, to improve robustness and compliance with the specification. This is especially important for defining a Common Node API and an OpenRPC specification.

- OS:
- Branch/commit
- Hardware

sudo-shashank:
## Summary of changes

- Introduced `ExtPredefined` which extends `Predefined` with `safe` and `finalized` predefined block.
- Changed input param to `EthHash` in `EthGetBlockByHash`
- Implemented missing case to handle `pending` in `EthGetTransactionCount`
- Added more test 

Closes #5125

## Change checklist


scripts/tests/api_compare/filter-list-offline
```diff
@@ -32,3 +32,6 @@
 !Filecoin.EthGetBlockByNumber
 !eth_getBlockByNumber
 !Filecoin.ChainSetHead
+!Filecoin.EthFeeHistory
```
hanabi1224, line 35: Could you add some comments on why these are filtered out?

sudo-shashank: these are now removed from filter list as i have changed test type to `basic` because to compare methods that take predefined block string(like `pending`, `latest`, `safe` or `finalized`) chain head for both the comparing nodes must be in sync if not you get different tipsets to compare which will fail eventually

elmattic (changes requested):

src/rpc/methods/eth.rs
```diff
@@ -256,6 +269,8 @@ pub struct BlockHash {
 pub enum BlockNumberOrHash {
     #[schemars(with = "String")]
     PredefinedBlock(Predefined),
+    #[schemars(with = "String")]
+    ExtPredefinedBlock(ExtPredefined),
```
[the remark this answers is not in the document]
sudo-shashank: fixed

src/rpc/methods/eth.rs
```diff
@@ -1917,18 +2053,22 @@ impl RpcMethod<2> for EthGetTransactionCount {
...
         if is_evm_actor(&actor.code) {
             let evm_state = evm::State::load(ctx.store(), actor.code, actor.state)?;
             if !evm_state.is_alive() {
                 return Ok(EthUint64(0));
             }
-
             Ok(EthUint64(evm_state.nonce()))
         } else {
-            Ok(EthUint64(ctx.mpool.get_sequence(&addr)?))
+            Ok(EthUint64(actor.sequence))
```
elmattic, line 2071: Is this a fix from previous code?

sudo-shashank: this is changed to match logic as per lotus code https://github.com/filecoin-project/lotus/blob/97dc1865cf1e205ed5b5f18402a4d9e8cfa29141/node/impl/eth/transaction.go#L381

src/tool/subcommands/api_cmd.rs
```diff
@@ -2043,7 +2223,7 @@ async fn run_tests(
                         true
                     }
                     PolicyOnRejected::PassWithQuasiIdenticalError
-                        if reason_lotus.starts_with(reason_forest) =>
+                        if reason_lotus.contains(reason_forest) =>
```
elmattic, line 2226: This may impact other tests that rely on an error string appearing at the beginning rather than elsewhere.

sudo-shashank: `contains` checks for the `reason_forest` to be present anywhere in `reason_lotus` should work even if `reason_lotus` starts with `reason_forest`

src/rpc/methods/eth.rs
```diff
@@ -715,7 +793,61 @@ fn tipset_by_block_number_or_hash<DB: Blockstore>(
...
+                if walk_ts != ts {
+                    bail!("tipset is not canonical");
+                }
+            }
+            Ok(ts)
+        }
+    }
+}
+
+fn tipset_by_ext_block_number_or_hash<DB: Blockstore>(
```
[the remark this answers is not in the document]
sudo-shashank: fixed

LesnyRumcajs:
@elmattic @lemmih @hanabi1224

lemmih (approved):

hanabi1224 (approved):

elmattic (approved):

Issue #5242, opened 2025-02-05
Wallet signature is not working properly


elmattic:
@elmattic So the issue is of two parts:
- The `forest-wallet sign` command was not printing the signature with it's type, like [lotus](https://github.com/filecoin-project/lotus/blob/6124799625f79279f950d3f680b6feb767d6e08c/cli/wallet.go#L471) does

- Ideally the signature will contains [type+data] the `forest-wallet verify` command was suppose to remove the first byte and consider rest of it as the signature data. instead it was passing the whole signature as it's data.

So after this change we will be able to verify the lotus wallet generated signature with forest-wallet.

_Originally posted by @akaladarshi in https://github.com/ChainSafe/forest/issues/5217#issuecomment-2636591316_

elmattic:
We should be able to verify signatures in Forest that are generated in Lotus, and vice versa.

e.g.:
```
./lotus wallet verify t410f2g4hvnfne4zluqvhrjpeonptlchqjzvxa6d5kla c0ffeebabe 03f7261df1d45a51c66023803f54ae700942edb37789150576e868f295af9cfe18002421ecf22b9fff8d8782bef38ce056e8a83761bccdf74163f710ce2206308100
valid
```

akaladarshi:
@elmattic just mentioning here for reminder, the `verify_delegated_sign` is able to verify the `lotus` generated signatures but while syncing it's failing (https://github.com/ChainSafe/forest/pull/5217/#issuecomment-2636607726)

elmattic:
> [@elmattic](https://github.com/elmattic) I am not able to raise a PR on top of my existing branch, until it's merged so I have raised it on my local fork.

FYI, your PR has been merged, so you can start a new branch from main or rebase.

PR #5255, opened 2025-02-08
feat: add signature verification for delegated accounts

akaladarshi:
## Summary of changes

- Add signature verification for delegated accounts

Closes https://github.com/ChainSafe/forest/issues/5242

- Current issue with the verification: [verify_delegated_sig is failing while syncing](https://github.com/ChainSafe/forest/issues/5242#issuecomment-2640052563)

## Change checklist


src/shim/crypto.rs
```diff
@@ -209,6 +209,52 @@ pub fn verify_bls_sig(
...
+        ));
+    }
+
+    if signature.len() != SECP_SIG_LEN {
+        return Err(format!(
+            "invalid delegated signature length. Was {}, must be {}",
+            signature.len(),
+            SECP_SIG_LEN
+        ));
+    }
```
LesnyRumcajs, line 238: nit: there's a nice macro `anyhow::ensure` that makes such validations more expressive.

src/shim/crypto.rs
```diff
@@ -209,6 +209,52 @@ pub fn verify_bls_sig(
     fvm_shared_latest::crypto::signature::ops::verify_bls_sig(signature, data, &addr.into())
 }
 
+/// Returns `String` error if a delegated signature is invalid.
+pub fn verify_delegated_sig(
+    signature: &[u8],
+    data: &[u8],
+    addr: &crate::shim::address::Address,
+) -> Result<(), String> {
```
LesnyRumcajs, line 217: Let's use `anyhow::Result<()>`. `String` itself should rarely be an error class.

src/shim/crypto.rs
```diff
@@ -209,6 +209,52 @@ pub fn verify_bls_sig(
...
+        EthAddress::eth_address_from_pub_key(&pub_key.serialize()).map_err(|e| e.to_string())?;
+
+    let rec_addr = eth_addr.to_filecoin_address().map_err(|e| e.to_string())?;
+
+    // check address against recovered address
+    if rec_addr == *addr {
+        Ok(())
+    } else {
+        Err("Delegated signature verification failed".to_owned())
+    }
```
LesnyRumcajs, line 255: `ensure!` could be used here as well.

LesnyRumcajs (commented):
Thanks. Some technical comments, otherwise LGTM.

src/shim/crypto.rs
LesnyRumcajs: Yes, it would need to be refactored, and it would be a good change. From briefly looking at the calls, it shouldn't be too complex.

The functions from `fvm_shared` should also be refactored (historical trivia: those were likely copied from Forest in 2021). However, given that it's a library's public interface, they should rather use a more constrained error type enumeration. Of course, refactoring `ref-fvm` methods is out of the scope of this issue, but it might be interesting to do so in the future.

LesnyRumcajs: Excellent, thanks @akaladarshi!

src/shim/crypto.rs
akaladarshi: Done.

src/shim/crypto.rs
akaladarshi: Done.

src/shim/crypto.rs
akaladarshi: Done.

LesnyRumcajs:
> @akaladarshi You can enable "format on save" in your favorite editor to prevent those pesky linter errors.

Either that or run `make lint` before submitting to ensure no lints fail during CI. You might need to install some additional tooling.

https://github.com/ChainSafe/forest/blob/ae5e424f3820018810598e3ca92c882d2d5b8df1/Makefile#L58

LesnyRumcajs:
@akaladarshi, did you try running the node locally with your changes? If you look at the logs for calibnet RPC checks for Forest, you will see:
>    2025-02-11T16:20:30.441807Z  WARN forest::chain_sync::tipset_syncer: Validating block [CID = bafy2bzacedyionjhebhdnqox3qd77vvdd3mvqrwhk27mqnvyl4tkr6qkzf6f2] in EPOCH = 2398815 failed: Validation error: Processing error: Error calculating weight: Can't create a valid state tree from the given root. This error may indicate unsupported version. state_root_cid=bafy2bzaceaepph3yqaewolhvxsgisqjctlse3qtled37khcofkgsny33uw7uu, state_root_version=unknown, Processing error: Could not update state: Can't create a valid state tree from the given root. This error may indicate unsupported version. state_root_cid=bafy2bzaceaepph3yqaewolhvxsgisqjctlse3qtled37khcofkgsny33uw7uu, state_root_version=unknown, Processing error: Failed to calculate state: Can't create a valid state tree from the given root. This error may indicate unsupported version. state_root_cid=bafy2bzaceaepph3yqaewolhvxsgisqjctlse3qtled37khcofkgsny33uw7uu, state_root_version=unknown, Validation error: Consensus error: StateManager error: Can't create a valid state tree from the given root. This error may indicate unsupported version. state_root_cid=bafy2bzaceaepph3yqaewolhvxsgisqjctlse3qtled37khcofkgsny33uw7uu, state_root_version=unknown, Consensus error: Power actor not found

akaladarshi:
@LesnyRumcajs Based on my investigation, I found out that there were few issues in how we were processing `Ethereum tx` -> `Filecoin message` vice versa i.e (method params was supposed to be encoded in CBOR). Also few functions were missing so I have updated the code accordingly. Please review the changes.

Meanwhile I will try to add few unit tests for the new changes.

LesnyRumcajs:
> @LesnyRumcajs Based on my investigation, I found out that there were few issues in how we were processing `Ethereum tx` -> `Filecoin message` vice versa i.e (method params was supposed to be encoded in CBOR). Also few functions were missing so I have updated the code accordingly. Please review the changes.
> 
> Meanwhile I will try to add few unit tests for the new changes.

@elmattic would you mind looking taking a look?

elmattic:
> > @LesnyRumcajs Based on my investigation, I found out that there were few issues in how we were processing `Ethereum tx` -> `Filecoin message` vice versa i.e (method params was supposed to be encoded in CBOR). Also few functions were missing so I have updated the code accordingly. Please review the changes.
> > Meanwhile I will try to add few unit tests for the new changes.
> 
> @elmattic would you mind looking taking a look?

Of course.

elmattic:
@akaladarshi Looks like `test_eip_1559` is failing.

akaladarshi:
> @akaladarshi Looks like `test_eip_1559` is failing.

I am looking into adding more unit tests and fixing/updating existing once, just wanted to get few comments on the changes.

src/shim/crypto.rs
```diff
@@ -132,15 +135,49 @@ impl Signature {
...
+    /// - BLS/SECP: Standard signature verification
+    pub fn authenticate_msg(
+        &self,
+        eth_chain_id: EthChainId,
+        msg: &SignedMessage,
+        addr: &crate::shim::address::Address,
+    ) -> anyhow::Result<()> {
+        let mut sig = self.clone();
+        let digest: Vec<u8>;
+        match self.sig_type {
```
elmattic, line 149: nit: you can also use the following form:
```
let digest = match self.sig_type {
};
```
which is a bit more idiomatic.

src/shim/crypto.rs
```diff
@@ -209,6 +246,47 @@ pub fn verify_bls_sig(
...
+    anyhow::ensure!(
+        signature.len() == SECP_SIG_LEN,
+        "invalid delegated signature length. Was {}, must be {}",
+        signature.len(),
+        SECP_SIG_LEN
+    );
+
+    let hash = keccak_256(data);
+    let mut sig = [0u8; SECP_SIG_LEN];
+    sig[..].copy_from_slice(signature);
```
elmattic, line 277: nit: remove second `ensure` and `mut` and replace them with some `try_into()`:

```
let sig: [u8; SECP_SIG_LEN] = signature.try_into().with_context(|| {
    format!(
        "invalid delegated signature length. Was {}, must be {}",
        signature.len(),
        SECP_SIG_LEN,
    )
})?;
```

elmattic (changes requested):
Great work, looks good. Just a few nits consider adding tests for functions like `verify_delegated_sig` and `authenticate_msg`.

elmattic:
There are also a few small typos to fix, it's `delegated`, not `delegate`. For example:

> fn delegate_sign() {

Search for occurrences of `delegate_` in the codebase.

src/eth/eip_1559_transaction.rs
```diff
@@ -139,9 +139,25 @@ impl EthEip1559TxArgs {
         from: Address,
         eth_chain_id: EthChainId,
     ) -> anyhow::Result<SignedMessage> {
-        ensure!(self.chain_id != eth_chain_id, "Invalid chain id");
```
akaladarshi, line 142: @elmattic What was the purpose of this check here (why are we using `!=`) ?.

akaladarshi:
@elmattic one more thing, I guess test is failing because perviously we were checking `self.chain_id != chain_id` [here](https://github.com/ChainSafe/forest/blob/410d7a382d316bb530f4c269db94fa2198086f27/src/eth/eip_1559_transaction.rs#L142) , and which was working because in tests we are providing different chain id's ( [ETH_CHAIN_ID](https://github.com/ChainSafe/forest/blob/410d7a382d316bb530f4c269db94fa2198086f27/src/eth/transaction.rs#L752) should have been set as `calibnet::ETH_CHAIN_ID` because that's what we are expecting [here](https://github.com/ChainSafe/forest/blob/410d7a382d316bb530f4c269db94fa2198086f27/src/eth/transaction.rs#L728))

src/eth/eip_1559_transaction.rs
elmattic: Seems to be a type of sanity check at first glance, but the `!=` is puzzling, indeed. @sudo-shashank can you elaborate on that?

elmattic: I believe the equivalent code in Lotus is:
```
if tx.ChainID != buildconstants.Eip155ChainId {
    return nil, fmt.Errorf("invalid chain id: %d", tx.ChainID)
}
```

akaladarshi: Then I guess it was mistake because `ensure!` and `if` condition is not the same thing. No worries I have updated it will push the changes.

elmattic: Yeah, it looks like a mistake on our side to me.

akaladarshi: Done.

src/shim/crypto.rs
akaladarshi: Done.

src/shim/crypto.rs
akaladarshi: Done.

elmattic:
@akaladarshi Great! Can you just fix the small lint error? Thanks! 🙏

elmattic (approved):

elmattic (approved):

src/key_management/wallet.rs
```diff
@@ -525,4 +525,25 @@ mod tests {
...
         assert!(sig.verify(&msg, &invalid_addr).is_err())
     }
+
+    #[test]
+    fn delegated_verify() {
+        let delegated_priv_key = generate(SignatureType::Delegated).unwrap();
+        let delegated_key_info = KeyInfo::new(SignatureType::Delegated, delegated_priv_key);
+        let delegated_key = Key::try_from(delegated_key_info).unwrap();
+        let addr = delegated_key.address;
+        println!("addr: {}", addr);
```
LesnyRumcajs, line 535:
```suggestion
```

leftover?

src/shim/crypto.rs
```diff
@@ -132,15 +135,48 @@ impl Signature {
...
+    /// Authenticates the message signature using protocol-specific validation:
+    /// - Delegated: Uses the Ethereum message with RLP encoding for signature verification, Verifies message roundtrip integrity
+    /// - BLS/SECP: Standard signature verification
+    pub fn authenticate_msg(
+        &self,
+        eth_chain_id: EthChainId,
+        msg: &SignedMessage,
+        addr: &crate::shim::address::Address,
+    ) -> anyhow::Result<()> {
+        let mut sig = self.clone();
```
LesnyRumcajs, line 147: Why do we need to clone the entire instance? Does it need to always be cloned?

akaladarshi: self is a `Signature` itself, cloning it, so we can update the signature bytes in `line 157`, without actually affecting the self.

src/key_management/wallet.rs
akaladarshi: Removed.

src/shim/crypto.rs
elmattic: You can use this syntax as well, which is a bit more idiomatic too: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax

LesnyRumcajs: Let me re-phrase. Do you think cloning (a relatively costly operation; you need to heap-allocate the entire vector) is necessary in case the signature type is not `Delegated` (which is the case most of the time)?

elmattic: > ```rust
>  match self.sig_type {
> ```

yes, that's it, this will only clone in case of `Delegated` sig. and doing `*self` should be cheap here.

akaladarshi: @LesnyRumcajs Thanks for pointing it out, I think you're right it doesn't make sense at all. I will update it as per my last comment.

LesnyRumcajs (approved):
LGTM, thanks!

PR #5267, opened 2025-02-11
ci: nix flake

lemmih:
## Summary of changes

- Add nix flake for building Forest. This is similar to a Dockerfile, but completely deterministic.

Closes

## Change checklist


lemmih:
I'll configure a github workflow in a separate PR. Right now, we do not have any nix caching, and building Forest from scratch takes a while.

flake.nix
```diff
@@ -0,0 +1,110 @@
...
+      cargoArtifacts = craneLib.buildDepsOnly commonArgs;
+
+      # Build the actual crate itself, reusing the dependency
+      # artifacts from above.
+      forest = craneLib.buildPackage (commonArgs
+        // {
+          inherit cargoArtifacts;
+          # Set HOME for Go compilation
+          preConfigure = ''
+            export HOME=$(mktemp -d)
```
hanabi1224, line 71: I believe we only need to set `GOCACHE`, check `go env`

README.nix.md
```diff
@@ -0,0 +1,92 @@
...
+information.
+
+## Building Forest with Flakes
+
+Forest uses Nix flakes for reproducible builds. To build Forest:
+
+1. Clone the Forest repository
+2. Run the following command in the repository root:
+   ```bash
+   nix build
```
hanabi1224, line 31: I got the below errors running `nix build` locally. Do I need to sign in or set a token?

```
warning: unable to download 'https://cache.flakehub.com/nix-cache-info': HTTP error 401

         response body:

         {"code":401,"error":"Unauthorized","message":"Unauthorized."}
```

lemmih: It's just a warning. To use the cache, you have to log in to FlakeHub.

hanabi1224: Thanks, maybe it's worth mentioning in the doc as well to avoid surprises.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,27 @@
...
+      - main
+  push:
+    branches:
+      - main
+
+jobs:
+  build-nix-package:
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, ubuntu-24.04-arm, macos-14]
```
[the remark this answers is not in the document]
hanabi1224: Also, maybe it's better to switch to ubuntu-22.04-arm, following https://github.com/ChainSafe/forest/pull/5261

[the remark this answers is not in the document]
lemmih: Updated the workflow to only build on non-draft PRs.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,29 @@
...
```
LesnyRumcajs, line 16:
```suggestion
        os: [ubuntu-24.04, ubuntu-22.04-arm, macos-14]
```

https://github.com/ChainSafe/forest/pull/5261

.github/workflows/nix.yml
```diff
@@ -0,0 +1,29 @@
+name: nix
```
LesnyRumcajs, line 1:
```suggestion
name: nix

# Cancel workflow if there is a new change to the branch.
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}

```

.github/workflows/nix.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  workflow_dispatch:
+  merge_group:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
```
LesnyRumcajs, line 10:
```suggestion
on:
  workflow_dispatch:
  merge_group:
  pull_request:
    branches:
      - main
    paths-ignore:
      - 'docs/**'
      - '.github/workflows/docs-*.yml'
  push:
    branches:
      - main
    paths-ignore:
      - 'docs/**'
      - '.github/workflows/docs-*.yml'
```

LesnyRumcajs (changes requested):

.github/workflows/nix.yml
lemmih: If we can't afford the minutes then let's not build the nix flake at all.

LesnyRumcajs: We can afford them, but I'm not convinced the usage of MacOS here is justified. Do you expect this build to pass on ARM and not on MacOS, especially given its deterministic setup? In the same way, we don't build and run RPC calibnet tests or unit tests on all worker types.

lemmih: Generating cache artifacts on MacOS is quite useful for me since I do most of my development on MacOS. Perhaps a compromise would be to cache the compiled dependencies and releases of Forest.

LesnyRumcajs: Given your use case, would running the MacOS build only on `main` make sense, or would that cache be insufficient?

lemmih: If we only run on 'main' then we risk red crosses.

LesnyRumcajs: So you expect the MacOS build to potentially fail even if the Linux one is green? I'm asking because my knowledge of `nix` is non-existent and I wouldn't expect such behaviour with, e.g., `docker build`, hence my original concern.

lemmih: There are no guarantees that a build will succeed on MacOS, even if it was successful on Linux. I vote for building dependencies in PRs. This guarantees green checkmarks for `main` (since `main` will always see a populated cache).

LesnyRumcajs: Ok.

.github/workflows/nix.yml
lemmih: The builds do nothing if no source files are modified.

elmattic (approved):

.github/workflows/nix.yml
LesnyRumcajs: @lemmih what do you think?

PR #5286, opened 2025-02-13
fix: use non-resumable downloading for actor bundles to avoid CI test flakyness

hanabi1224:
## Summary of changes

-

Closes

## Change checklist


src/utils/net/download_file.rs
```diff
@@ -13,6 +13,13 @@ use std::{
 };
 use url::Url;
 
+#[derive(Debug, Copy, Clone)]
+pub enum DownloadFileOption {
+    /// Non-resumable
+    Default,
```
LesnyRumcajs, line 19: Let's be explicit. As a reader, I cannot tell what `Default` is without looking into this code.
```suggestion
    NonResumable,
```

LesnyRumcajs (changes requested):

src/utils/net/download_file.rs
```diff
@@ -13,6 +13,13 @@ use std::{
 };
 use url::Url;
 
+#[derive(Debug, Copy, Clone)]
+pub enum DownloadFileOption {
+    /// Non-resumable
```
LesnyRumcajs, line 18:
```suggestion
```

LesnyRumcajs (approved):

src/daemon/bundle.rs
```diff
@@ -87,6 +89,7 @@ pub async fn load_actor_bundles_from_server(
     network: &NetworkChain,
     bundles: &[ActorBundleInfo],
 ) -> anyhow::Result<Vec<Cid>> {
+    let semaphore = Arc::new(Semaphore::new(4));
```
LesnyRumcajs, line 92: Why 4 in particular? And not 3 or 10?

hanabi1224: I think any value between 2-8 is fine, 4 is somewhere in the middle.

lemmih (approved):

src/daemon/bundle.rs
hanabi1224: Larger number is faster but more error-prone. Previously, all tasks run in parallel and the test was quite flaky

hanabi1224: Looking at the time cost, 4 makes it more stable and still fast

LesnyRumcajs: We can try, but I'm not :100: convinced. I'd rather we have a deterministically correct and working solution. This seems like reducing error probability from an unknown chance to another unknown (but likely smaller) chance, while always paying the price of reduced performance (slower startup for fresh nodes).

hanabi1224: We've been using the same logic for RPC snapshot test and it's much more stable, the only notable difference is, actor bundle uses Github assets primarily while RPC snapshot test uses DO space CDN.

hanabi1224: And I believe it's a common practice to limit max concurrent downloads, e.g. the author uses 8 in https://patshaughnessy.net/2020/1/20/downloading-100000-files-using-async-rust

Issue #4929, opened 2024-10-26
Indicate snapshot download in sync status


ansermino:
# Summary

When starting a fresh node`forest-cli sync status` will not return useful information if a snapshot download is in progress. 

<img width="302" alt="Screenshot 2024-10-24 at 6 34 45 PM" src="https://github.com/user-attachments/assets/c161f8fa-d416-4589-823b-4d0f8518dc79">

# Requirements

- Clearly indicate in the sync status if the node is presently downloading a snapshot and not yet syncing the chain via p2p connections
- (Optional) Clearly indicate the progress of the download (and a remaining time estimate) in the response

# Motivation

Getting started running a node can be tough, so some hand-holding is warranted. Users may not understand the purpose of snapshots and believe something is broken. Additionally, being able to query the status of the snapshot download could be useful for remote management.

LesnyRumcajs:
Indicate the download progress in the health check (readiness probe, maybe others)

akaladarshi:
@LesnyRumcajs Can I pick this ?

LesnyRumcajs:
Sure, go ahead. Thanks. @akaladarshi

akaladarshi:
@LesnyRumcajs, I have a question regarding this implementation.

I noticed that Lotus doesn't show any progress indicator for snapshot downloads, so I'd like to discuss a few ideas:

- We could update the existing `Filecoin.SyncState` RPC or add a new one that returns the latest snapshot progress. This would involve a few changes:
  - Modify `Filecoin.SyncState `or create a new RPC to include snapshot progress.
  - Get the current progress by either:
    - Adding a callback or channel to the progress struct [here](https://github.com/ChainSafe/forest/blob/d7315bfdb679767ad75a3adac88d8da3ee8d0f32/src/utils/io/progress_log.rs#L147)
    - Instead of using `tokio::io::copy` to read the file [here](https://github.com/ChainSafe/forest/blob/d7315bfdb679767ad75a3adac88d8da3ee8d0f32/src/utils/net/download_file.rs#L161), manually reading the file and updating the snapshot progress.

- One potential problem with this solution is that we might need to continuously poll the RPC to get updated progress data.
- Another idea is to monitor the temporary file where the data is initially written before it’s copied to the destination. This approach is somewhat naive and might lead to issues.
- I am also exploring whether filtering `logging` events could provide the necessary progress information, though I'm not sure how well this will work.


Just putting these ideas out here.

[...]

PR #5368, opened 2025-03-03
feat(status): Add snapshot progress RPC

akaladarshi:
## Summary of changes

- Create an `Appcontext` in `daemon` to hold common objects
- Create `SnapshotTracket` to track the snapshot progress
- Expose a RPC API `Filecoin.SyncStatus` which reports the snapshot download progress

Closes: #4929 

## Change checklist


akaladarshi:
@LesnyRumcajs @elmattic Please 🙏 can any one of give it a review.

This PR is not fully ready for yet (naming and logging changes and need to remove extra command), just wanted to get some opinions on this, so I can proceed further.

LesnyRumcajs:
I'll let @hanabi1224 or @lemmih take a look; I don't have the capacity currently.

src/utils/io/progress_log.rs
```diff
@@ -183,17 +197,37 @@ impl Progress {
...
+                msg
             );
             self.last_logged = now;
             self.last_logged_items = self.completed_items;
         }
     }
 }
 
+// Derive Debug for Progress, skipping the callback field
+impl std::fmt::Debug for Progress {
```
hanabi1224, line 217: Alternatively, https://crates.io/crates/educe can be used to ignore the `callback` field

src/rpc/methods/sync/types.rs
```diff
@@ -0,0 +1,24 @@
+// Add to src/rpc/types.rs or a similar appropriate location
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use crate::lotus_json::lotus_json_with_self;
+
+#[derive(PartialEq, Default, Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(rename_all = "camelCase")]
```
hanabi1224, line 8:
```suggestion
```

hanabi1224: maybe it's OK to keep it for consistency

src/rpc/methods/sync.rs
```diff
@@ -71,6 +75,26 @@ impl RpcMethod<0> for SyncState {
     }
 }
 
+pub enum SyncStatus {}
+impl RpcMethod<0> for SyncStatus {
+    const NAME: &'static str = "Filecoin.SyncStatus";
```
hanabi1224, line 80:
```suggestion
    const NAME: &'static str = "Forest.SyncStatus";
```

Maybe better to avoid `Filecoin.` prefix since this is not a Filecoin API, what do you think? @LesnyRumcajs @lemmih

akaladarshi: @hanabi1224 I missed it actually, it was mentioned [here](https://github.com/ChainSafe/forest/issues/4929#issuecomment-2684489348) we need to use `Forest.SynsStatus`

src/utils/io/progress_log.rs
akaladarshi: Done.

src/rpc/methods/sync.rs
akaladarshi: I have renamed the RPC to `Forest.SyncSnapshotProgress`.

src/utils/io/progress_log.rs
hanabi1224: Thanks

akaladarshi:
Hey @hanabi1224 can you please approve the request for running the workflows.

Also I have made some changes on top of your last review please do review and let me know what you think 🙏🏼.

hanabi1224:
@akaladarshi Any ideas why some tests keep failing after a few retries? Please let us know if assistance is needed for investigating the failures.

akaladarshi:
> @akaladarshi Any ideas why some tests keep failing after a few retries? Please let us know if assistance is needed for investigating the failures.

The Snapshot test mostly fails because it is not able to download it before timeout.

I tried Devnet Check locally and it mostly failed because it was not able to download proofs and all.

I have seen these two test fail mostly in previous PR'S as well. Will try to dig deep.

akaladarshi:
> @akaladarshi Any ideas why some tests keep failing after a few retries? Please let us know if assistance is needed for investigating the failures.

@hanabi1224 While looking into this issue: I came across this weird behaviour. 

While adding another RPC call [SyncSnapshotProgress](https://github.com/akaladarshi/forest/blob/cda094a1dcac63b7e74bed918e49348d5d427220/src/cli/subcommands/sync_cmd.rs#L103) inside the `Status` subcommand which already has an RPC [SyncState](https://github.com/akaladarshi/forest/blob/cda094a1dcac63b7e74bed918e49348d5d427220/src/cli/subcommands/sync_cmd.rs#L113), the `Devnet Checks` starts failing.

That's the reason why the `Devnet Checks` was working in initial commits, because I added an extra subcommand `Progress` to check the snapshot progress.

Have you noticed this behaviour or can you give some pointers about it?

akaladarshi:
@hanabi1224 I was able to figure out what was causing the issue in `Devnet Check`. 

Details:
In snapshot progress I was using ticker crate same as `forest-cli status sync wait`, but since the snapshot progress could be long so I set the ticker timer to `5 seconds` instead of `1 second`.

Using a 5‑second interval lead to a delayed initial tick and slower loop completion. Although the loop escapes immediately when no snapshot progress is detected, the background thread created by the ticker remains active for up to 5 seconds (not sure about this exactly). When invoked repeatedly via Docker’s frequent healthchecks (forest-cli sync status), these delays accumulate—potentially leading to resource exhaustion or sluggish response times, which can cause Docker to mark the container as unhealthy.

So instead of `ticker crate` I am using `tokio time crate`, resolved the issue
> let mut interval = time::interval(Duration::from_secs(5)))

akaladarshi:
> I tried locally but `forest-cli sync status` reports incorrect status when the node is in sync and snapshot export is not in progress. Any ideas? @akaladarshi
> 
> ```
> forest --chain calibnet --encrypt-keystore false
> ```
> 
> ```
> ➜  lotus git:(master) forest -V
> forest-filecoin 0.24.0+git.1deb3388bf1
> ➜  lotus git:(master) forest-cli -V
> forest-filecoin 0.24.0+git.1deb3388bf1
> ➜  lotus git:(master) forest-cli sync status
> ⏳ Snapshot download not started - node is initializing
> ➜  lotus git:(master)                                          
> ```
Thanks for pointing this out 🙏🏽.

It was because I was checking wrong conditions to asses the `sync completion state`. I have added the correct check.

Also can you verify once are these [condition](https://github.com/akaladarshi/forest/blob/b854fcc149ec5bb9487987a91df1443efa372854/src/cli/subcommands/sync_cmd.rs#L128) correct for verifying the chain syncing status?.

hanabi1224:
Now it reports correct status for a running Forest node, but does not show download progress while downloading a snapshot. Does it work properly for you? @akaladarshi

src/cli/subcommands/sync_cmd.rs
```diff
@@ -123,11 +123,12 @@ impl SyncCommands {
...
                 let height_diff = base_height - target_height;
 
-                // If the sync state is not in the Complete stage and both base and target tipsets are empty,
+                // If the sync state is not in the Complete stage
+                // and both base and target cid's are empty,
                 // the node might be downloading the snapshot.
                 if state.stage() != SyncStage::Complete
-                    && base_cids.is_empty()
-                    && target_cids.is_empty()
+                    && base_cids.eq("[]")
```
hanabi1224, line 130: Let's make `"[]"` a constant

akaladarshi: Instead of returning "[]" can we do something like this:

` fn format_tipset_cids(cids: &str) -> &str {
       if cids == "[]" { "None" } else { cids }
   }`
   
 this way it will be more readable otherwise it's little confusing?

akaladarshi: Basically return an empty string and format it while printing.

hanabi1224: Sounds good

hanabi1224:
With the latest commit, `forest-cli sync status` properly reports the download status. But at the very beginning, the output seems incorrect: `🌳 Snapshot download in progress: ync status`

src/cli/subcommands/sync_cmd.rs
akaladarshi: Done

src/rpc/methods/sync/types.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright 2019-2025 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+// Add to src/rpc/types.rs or a similar appropriate location
```
sudo-shashank, line 3: @akaladarshi what is this for?

akaladarshi: Removed

src/cli/subcommands/sync_cmd.rs
```diff
@@ -150,3 +165,47 @@ impl SyncCommands {
...
+                continue;
+            }
+            SnapshotProgressState::Completed => {
+                write!(
+                    stdout,
+                    "\r{}{}",
+                    anes::ClearLine::All,
+                    anes::MoveCursorUp(1)
+                )?;
+                println!("\n✅ Snapshot download completed! Chain will start syncing shortly (retry sync status command in 5 seconds)...");
```
sudo-shashank, line 202: @akaladarshi it would be better to display sync status once snapshot is downloaded and chain starts syncing instead of asking user to re-run after 5 seconds and exiting

sudo-shashank (changes requested):

src/cli/subcommands/sync_cmd.rs
akaladarshi: I thought about that too but the things is we have to wait for some time before the node actually starts to sync after snapshot downloaded and wasn't sure about how much time we have to wait (actually depends on the node), so I just asked user to wait and retry.

hanabi1224:
LGTM. Thanks @akaladarshi . Will approve once outstanding issues are resolved.

BTW, snapshot import scenario does not seem to be covered by this PR. (importing a snapshot in `copy` mode could take a few minutes. `--import-snapshot [PATH] --import-mode copy`). I'm fine with improving that with a subsequent PR if it's non-trival to update this one.

akaladarshi:
> LGTM. Thanks @akaladarshi . Will approve once outstanding issues are resolved.
> 
> BTW, snapshot import scenario does not seem to be covered by this PR. (importing a snapshot in `copy` mode could take a few minutes. `--import-snapshot [PATH] --import-mode copy`). I'm fine with improving that with a subsequent PR if it's non-trival to update this one.

Thanks @hanabi1224.

Do we really need to show progress while copying a file? 

I'm little bit skeptical about it as **Orignal Issue** was about the `user requires hand holding` and `showing snapshot download` progress while running the node for the first time.

sudo-shashank (approved):
LGTM thanks @akaladarshi

hanabi1224 (approved):

hanabi1224:
@akaladarshi with https://github.com/ChainSafe/forest/pull/5403 the new RPC methods have to be added to `src/tool/subcommands/api_cmd/test_snapshots_ignored.txt` to fix the test failures.

hanabi1224 (approved):

Issue #5581, opened 2025-04-14
Reduce Tipset State Computation Overhead Using Next Epoch Lookup


akaladarshi:
# Summary

When retrieving the tipset state, instead of computing it every time as done [here](https://github.com/ChainSafe/forest/blob/bbc0e7bf213d6dea7180087ae27934f3f4c2e669/src/state_manager/mod.rs#L529), we should try to look it up using the next tipset if it’s available, similar to how [lotus handles it](https://github.com/filecoin-project/lotus/blob/a8999cb164f996785f1b01e6cb020610abaa7caf/chain/stmgr/execute.go#L80) handles it. This optimisation would help reduce the overhead of recomputing the state for every tipset unnecessarily (in some specific cases)

# Completion Criteria
- [ ] Create a function for looking up `tipset state` for a provided tipset from next `tipset` (epoch)
- [ ] Use the function fro specific condition, check [lotus](https://github.com/filecoin-project/lotus/blob/a8999cb164f996785f1b01e6cb020610abaa7caf/chain/stmgr/execute.go#L80)

PR #5587, opened 2025-04-16
feat: add tipset state lookup

akaladarshi:
## Summary of changes

- Adds the tipset state lookup from using the next tipset
- Updates the events and receipt caches

Closes: #5581 

## Change checklist

### DIFF
--- a/src/state_manager/mod.rs
+++ b/src/state_manager/mod.rs
@@ -526,21 +526,54 @@ where
                     tipset.epoch(),
                     tipset.len(),
                 );
+
+                // First try to lookup the state and receipt if not found in the blockstore
+                // compute it
+                if let Some(state_from_child) =
+                    self.try_lookup_state_from_next_tipset(tipset.as_ref())
+                {
+                    trace!("Using state from child tipset for epoch {}", tipset.epoch());
+                    return Ok(state_from_child);
+                }
+
+                trace!("Computing state for tipset at epoch {}", tipset.epoch());
                 let state_output = self
                     .compute_tipset_state(Arc::clone(tipset), NO_CALLBACK, VMTrace::NotTraced)
                     .await?;
+
                 for events_root in state_output.events_roots.iter() {
                     trace!("Indexing events root @{}: {}", tipset.epoch(), events_root);
-
                     self.chain_store().put_index(events_root, key)?;
                 }
+
+                // update the events_cache and receipt_cache
+                self.update_receipt_and_events_cache(key, &state_output);
+
                 let ts_state = state_output.into();
                 trace!("Completed tipset state calculation {:?}", tipset.cids());
-                // We missed the opportunity to update `self.events_cache` and `self.receipt_cache` here, to be refactored.
+
                 Ok(ts_state)
             })
             .await
             .map(StateOutput::from)
+    }
+
+    /// Updates the events and receipts cache for the given tipset.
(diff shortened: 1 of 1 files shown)


src/state_manager/mod.rs
```diff
@@ -1726,6 +1758,46 @@ where
 
         Ok((state_root, invoc_trace))
     }
+
+    /// Attempts to lookup the state and receipt root of the next tipset.
+    /// This is a performance optimization to avoid recomputing the state and receipt root by checking the blockstore.
+    /// It only checks the immediate next epoch, as this is the most likely place to find a child.
+    fn try_lookup_state_from_next_tipset(&self, tipset: &Tipset) -> Option<StateOutputValue> {
+        let epoch = tipset.epoch();
+        let next_epoch = epoch.saturating_add(1);
```
LesnyRumcajs, line 1767: It should be fine to use regular `+` operator. If we ran out of epochs in that type, we have much worse problems.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 ### Changed
 
 - [#5467](https://github.com/ChainSafe/forest/pull/5467) Improve error message for `Filecoin.EthEstimateGas` and `Filecoin.EthCall`.
+- [#5587](https://github.com/ChainSafe/forest/pull/5587) Retrieve state root from child tipset rather than recomputing when possible
```
LesnyRumcajs, line 39:
```suggestion
```

This is not really a user-facing change, so let's omit it from the changelog.

LesnyRumcajs (commented):
In general, looks good. Could we craft some tests for this optimisation? There are some utilities that might be helpful, e.g., `Chain4U` or something that was used in the `chain_follower.rs` tests.

CHANGELOG.md
akaladarshi: Done.

src/state_manager/mod.rs
akaladarshi: Done.

src/state_manager/mod.rs
```diff
@@ -526,23 +526,53 @@ where
...
                 let state_output = self
                     .compute_tipset_state(Arc::clone(tipset), NO_CALLBACK, VMTrace::NotTraced)
                     .await?;
                 for events_root in state_output.events_roots.iter().flatten() {
                     trace!("Indexing events root @{}: {}", tipset.epoch(), events_root);
-
                     self.chain_store().put_index(events_root, key)?;
                 }
+
+                self.update_receipt_and_events_cache(key, &state_output);
```
elmattic, line 547: Do we really need to do this? That means even if we disable the RPC server, we're still going to fill some receipt and events cache, which could get quite large. What do you think?

akaladarshi: Nice catch.

You're correct, cache won't be required if the RPC is disabled.

I guess we can remove it from this PR, we can tackle this separately.

Just putting a thought here:
 Could we do something like initalize cache only if `RPC` is enabled, and if the cache is initalized then only insert the data into it?. @elmattic what do you think?.

elmattic: That said, we could issue a warning upon exit if the cache is not empty and RPC is disabled.

akaladarshi: Lets do one thing, I will remove the `updating cache` part from this PR so we can track it separately.

akaladarshi: Raised an Issue for this: https://github.com/ChainSafe/forest/issues/5626

PR #5624, opened 2025-05-05
Revive `Filecoin.EthTraceFilter` RPC

### LINKED ISSUE #5591: Filecoin.EthTraceFilter response mismatch
## Description

Latest RPC parity test failed. Please [check the logs](https://github.com/ChainSafe/forest/actions/runs/14584063545) for more information.

sudo-shashank:
## Summary of changes

- Fix and enable `Filecoin.EthTraceFilter`: The issue was caused by a null tipset being resolved to the earlier available tipset, which could fall outside the specified range in the filter criteria.

Closes #5591 

## Change checklist


src/rpc/methods/eth.rs
```diff
@@ -872,34 +876,57 @@ fn resolve_predefined_tipset<DB: Blockstore>(
...
+                .chain_index
+                .tipset_by_height(safe_height, head, resolve)?)
+        }
+        ExtPredefined::Finalized => {
+            let finality_height = latest_height - chain.chain_config.policy.chain_finality;
+            Ok(chain
+                .chain_index
+                .tipset_by_height(finality_height, head, resolve)?)
+        }
+        _ => unreachable!(), // All variants are already handled
```
LesnyRumcajs, line 899: Not sure I understand - if all variants are handled, then why do we need this? If they are expected to be handled elsewhere but end up not being handled, then we should return an error and not crash the entire process.

sudo-shashank: we have all the variants covered but still we should not crash if a new variant is added later and not handled
changed to return an error

hanabi1224: > but to future-proof the code

Without the catch-all branch, an unhandled variant should cause a compile-time error. Why do we have to make it a runtime error?

sudo-shashank: except `Safe` and `Finalized` other variants are handled separately via `resolve_predefined_tipset` this was done to reuse the code from `Predefined` in `ExtPredefined`, hence we moved from this compile time check to runtime check

LesnyRumcajs (approved):
Let's add a changelog entry.

hanabi1224 (approved):

PR #5703, opened 2025-06-06
Fix `EthGetBlockReceiptsLimited`

### LINKED ISSUE #5177: `EthGetBlockReceiptsLimited` does not use limit
## Describe the bug

The bug is described here: https://github.com/filecoin-project/lotus/issues/12838

This is merely a tracking issue of the issue above, so that when the issue is resolved in Lotus, we can apply similar fix in Forest.

This way we should not differ from in `api compare` tests.

- OS:
- Branch/commit
- Hardware

elmattic:
## Summary of changes

- Fix `EthGetBlockReceiptsLimited` to properly support limit parameter

Closes #5177

## Change checklist


src/blocks/tipset.rs
```diff
@@ -90,6 +90,17 @@ impl TipsetKey {
             .collect::<Vec<_>>()
             .join(", ")
     }
+
+    /// Formats tipset key to match the Lotus display.
+    pub fn to_lotus(&self) -> String {
```
LesnyRumcajs:
```suggestion
    pub fn format_lotus(&self) -> String {
```
nit: I'd say it's more expressive this way. `to_lotus` suggests some other type.

src/blocks/tipset.rs
```diff
@@ -90,6 +90,17 @@ impl TipsetKey {
...
+        let s = self
+            .to_cids()
+            .into_iter()
+            .map(|cid| cid.to_string())
+            .collect::<Vec<_>>()
+            .join(",");
+        format!("{{{}}}", s)
```
LesnyRumcajs:
```suggestion
        format!("{{{}}}", self.to_cids().into_iter().join(","))
```
Would this work? Might be a bit more efficient by avoiding intermediary `Vec` allocation

src/rpc/methods/eth.rs
```diff
@@ -1488,14 +1488,24 @@ impl RpcMethod<2> for EthGetBlockByNumber {
...
     let ts = tipset_by_block_number_or_hash(
         ctx.chain_store(),
         block_param,
         ResolveNullTipset::TakeOlder,
     )?;
+    if let Some(limit) = limit {
+        if limit > LOOKBACK_NO_LIMIT
+            && ts.epoch() < ctx.chain_store().heaviest_tipset().epoch() - limit
+        {
+            return Err(anyhow::anyhow!(
```
LesnyRumcajs: `bail!`?

elmattic: Unfortunately `bail!` macro doesn't work well with `.into()`.

elmattic: That said, we can change the return type of `get_block_receipts`  and then use:
```
get_block_receipts(&ctx, block_param, None)
            .await
            .map_err(|e| e.into())
```

elmattic: Done.

sudo-shashank (approved):

Issue #4490, opened 2024-07-05
Panic in `MessagePool::select_messages_optimal`


hanabi1224:
## Describe the bug

Logic in `select_messages_optimal` is quite out-of-sync with lotus. We need to revisit and update the logic to get parity.

## To reproduce

1. Send the below payload to `http://localhost:2345/rpc/v0`
```
{"jsonrpc":"2.0","id":0,"method":"Filecoin.MpoolSelect","params":
[[{"/":"bafy2bzaceb6vn533luic5oo7tnvudxy2g7jusjc4m3egjb3ei3tu2lhmdlbvi"},{"/":"bafy2bzaceabqr4736hunl6wu36xs5yqwdjhwbopplwqxykf54ssgbaqaphsq4"},{"/":"bafy2bzacec7zaiykrnemfuqmed7pkiyranuhixkfbv6aaa6rvkwojya5opnuk"}],0.8]
}
```
2. get the below panic
```
thread 'tokio-runtime-worker' panicked at /home/me/git/forest/src/message_pool/msgpool/selection.rs:446:17:
attempt to subtract with overflow
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

forest code (panic in the last line):
```rust
                chains[i].merged = true;
                result.extend(chains[i].msgs.clone());
                random_count += chains[i].msgs.len();
                gas_limit -= chain_gas_limit;
```

## Log output

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
  paste log output...
  ```
</details>

## Environment (please complete the following information):

- OS:
- Branch/commit
- Hardware

akaladarshi:
@hanabi1224 Can I pick this?

akaladarshi:
@hanabi1224 For the request mentioned in the issue:

```
curl` -X POST http://127.0.0.1:2345/rpc/v0 -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":0,"method":"Filecoin.MpoolSelect","params":
[...]

PR #5729, opened 2025-06-12
fix: panics in mpool

LesnyRumcajs:
## Summary of changes

Resolves a few cases of panic in the Forest in mempool select logic. 

- unsigned integer underflow in gas calculations,
- total ordering violation in user-defined comparison method.

Tested on both calibnet and mainnet; calibnet tests added to the CI.

Follow-up changes to the logic will come via https://github.com/ChainSafe/forest/issues/5730.

Closes https://github.com/ChainSafe/forest/issues/4490

## Change checklist

### DIFF
--- a/src/message_pool/msgpool/mod.rs
+++ b/src/message_pool/msgpool/mod.rs
@@ -176,7 +176,7 @@ where
                     chains.invalidate(key);
                     continue 'l;
                 }
-                gas_limit -= m.gas_limit();
+                gas_limit = gas_limit.saturating_sub(m.gas_limit());
                 msgs.push(m.clone());
             }
 
--- a/src/message_pool/msgpool/selection.rs
+++ b/src/message_pool/msgpool/selection.rs
@@ -168,7 +168,7 @@ where
                 if gas_limit < chain_gas_limit {
                     break;
                 }
-                gas_limit -= chain_gas_limit;
+                gas_limit = gas_limit.saturating_sub(chain_gas_limit);
                 if gas_limit < MIN_GAS {
                     break;
                 }
@@ -267,7 +267,7 @@ where
                 }
 
                 result.extend(chains[i].msgs.clone());
-                gas_limit -= chain_gas_limit;
+                gas_limit = gas_limit.saturating_sub(chain_gas_limit);
 
                 // re-sort to account for already merged chains and effective performance
                 // adjustments the sort *must* be stable or we end up getting
@@ -576,7 +576,7 @@ fn merge_and_trim(
         }
 
         if node.gas_limit <= gas_limit {
-            gas_limit -= node.gas_limit;
+            gas_limit = gas_limit.saturating_sub(node.gas_limit);
             result.extend(node.msgs.clone());
             continue;
         }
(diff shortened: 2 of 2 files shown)


scripts/mpool_select_killer.rb
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# This script was used to reproduce multuple panic bugs in Forest's `Filecoin.MpoolSelect` RPC method.
```
elmattic: typo in multiple

hanabi1224 (dismissed):

scripts/mpool_select_killer.rb
akaladarshi, line 1: @LesnyRumcajs Instead of ruby script can we use `shell` or `js`. I'd prefer `shell` because we are already using that for other scripts and also using `rb` increases the number of dependencies for forest (even though it's a small script)

LesnyRumcajs: Ruby is [already](https://github.com/ChainSafe/forest/blob/main/scripts/linters/find_unused_deps.rb) a dependency in Forest. It's likely possible to achieve the same result with shell and UNIX tooling, but it's also likely the result would be more convoluted and complex to understand.

akaladarshi: I'm not much aware about ruby but if you feel it does better job here then we should stick to it and try to use it more.

akaladarshi (commented):
We should also update to `gas_limit = gas_limit.saturating_sub(chain_gas_limit)`
https://github.com/ChainSafe/forest/blob/92ee2fb514ccb8d7b81e0f45aa5d00840906368e/src/message_pool/msgpool/selection.rs#L171

scripts/mpool_select_killer.rb
LesnyRumcajs: My view is that we should:
1. Use Rust when possible, unless it's a total overkill.
2. Use shell for scripting, e.g., tests where the logic flow is trivial and can be easily achieved using Unix tools. Most tests in https://github.com/ChainSafe/forest/tree/main/scripts/tests are a good example of this, where there is very little happening - just commands executed in sequence, occasionally in loops. Bash syntax and AWK|sed|jq gymnastics are tricky to understand and debug.
3. Use Ruby for things that require more logic, and writing them with shell scripts would make it difficult to reason.
4. If additional libraries are required for some testing and they're not available in Ruby, the best option is to go with whatever works with such libraries. For example, we went with JavaScript for benchmarking tests because of k6.

Note on scripting languages - I firmly believe we shouldn't use them for anything complex. For complex things, we should profit from the relative safety of the Rust compiler.

LesnyRumcajs:
> We should also update to `gas_limit = gas_limit.saturating_sub(chain_gas_limit)`
> 
> https://github.com/ChainSafe/forest/blob/92ee2fb514ccb8d7b81e0f45aa5d00840906368e/src/message_pool/msgpool/selection.rs#L171
> 
> Also if there are other places we do the same.

It's an excellent point. I updated it only in the places where it panicked, so it's _likely_ not needed, but better safe than sorry.

Issue #5810, opened 2025-07-03
Support exporting snapshots in new fomat with F3 data


hanabi1224:
# Summary

# Completion Criteria

- [ ]

# Additional Links & Resources

PR #5824, opened 2025-07-09
feat: util for calculating multihash from byte stream

hanabi1224:
## Summary of changes

Part of https://github.com/ChainSafe/forest/issues/5810

- add `digest_byte_stream` for calculating multihash and CID of a large byte stream, e.g. F3 snapshot
- unit test

Closes

## Change checklist

## Summary by CodeRabbit

* **New Features**
  * Added the ability to hash data directly from input streams, allowing users to compute multihashes from large or streamed data sources.

* **Tests**
  * Introduced tests to ensure consistent hashing results between stream-based and buffer-based methods.


src/utils/multihash.rs
```diff
@@ -76,6 +76,106 @@ pub enum MultihashCode {
...
+                    break;
+                }
+                if let Some(b) = buf.get(0..n) {
+                    hasher.update(b);
+                }
+            }
+            Ok(hasher.finalize())
+        }
+
+        Ok(match self {
```
LesnyRumcajs, line 99: nit: do you think it's a good opportunity for a macro to avoid code duplication?

src/utils/multihash.rs
```diff
@@ -76,6 +76,106 @@ pub enum MultihashCode {
...
+            Self::Ripemd256 => {
+                let mut hasher = multihash_codetable::Ripemd256::default();
+                self.wrap(hash(&mut hasher, bytes)?)?
+            }
+            Self::Ripemd320 => {
+                let mut hasher = multihash_codetable::Ripemd320::default();
+                self.wrap(hash(&mut hasher, bytes)?)?
+            }
+            _ => {
+                unimplemented!("{self:?}")
```
LesnyRumcajs: That's a bit scary - we shouldn't crash the entire Forest on a bad reader.

src/utils/multihash.rs
hanabi1224: I don't think a macro would significantly reduce the LoC, but it would hurt readability. What do you think?

src/utils/multihash.rs
```diff
@@ -113,3 +213,31 @@ impl<const S: usize> multihash_derive::Hasher for IdentityHasher<S> {
...
+
+    use super::*;
+    use crate::utils::rand::forest_rng;
+    use rand::RngCore as _;
+
+    #[test]
+    fn test_digest_byte_stream() {
+        use MultihashCode::*;
+
+        let mut bytes = vec![0; 10000];
```
LesnyRumcajs: if we're doing a bit of a fuzzy tests, perhaps it would make sense to randomize the input size as well.

src/utils/multihash.rs
```diff
@@ -76,6 +76,106 @@ pub enum MultihashCode {
...
+            loop {
+                let n = bytes.read(&mut buf)?;
+                if n == 0 {
+                    break;
+                }
+                if let Some(b) = buf.get(0..n) {
+                    hasher.update(b);
+                }
+            }
+            Ok(hasher.finalize())
```
LesnyRumcajs, line 96:
```suggestion
            while let Some(n) = bytes.read(&mut buf).ok() {
                if n == 0 {
                    break;
                }
                hasher.update(&buf[..n]);
            }
            Ok(hasher.finalize())
```
nit: some code golf if you want

src/utils/multihash.rs
LesnyRumcajs: No strong position here, I don't expect this code to be updated with many new algorithms anytime soon in the future. We can leave it as is.

src/utils/multihash.rs
LesnyRumcajs: Or just manually check some edge conditions; input size of zero, 1, 1024. Also, there's an uncovered path with an unknown hashing algorithm.

hanabi1224: `IdentityHasher` is for testing and does not work for input > 64 bytes
```rust
/// Identity hasher with a maximum size.
///
/// # Panics
///
/// Panics if the input is bigger than the maximum size.
/// Ported from <https://github.com/multiformats/rust-multihash/pull/289>
```

src/utils/multihash.rs
hanabi1224: I think the `io::Error` from `.read` should be escalated here.
Then I get
```
warning: irrefutable `while let` pattern
  --> src/utils/multihash.rs:96:19
   |
96 |             while let n = bytes.read(&mut buf)? {
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this pattern will always match, so the loop will never exit
   = help: consider instead using a `loop { ... }` with a `let` inside it
   = note: `#[warn(irrefutable_let_patterns)]` on by default
...
error: slicing may panic
   --> src/utils/multihash.rs:100:32
    |
100 |                 hasher.update(&buf[..n]);
    |                                ^^^^^^^^
    |
    = help: consider using `.get(..n)`or `.get_mut(..n)` instead
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#indexing_slicing
```

hanabi1224: It could be a while loop like below but I'm not sure if it's better than a loop loop, what do you think?
```rust
let mut n = 0;
            while {
                n = bytes.read(&mut buf)?;
                n > 0
            } {
                if let Some(b) = buf.get(0..n) {
                    hasher.update(b);
                }
            }
```

src/utils/multihash.rs
hanabi1224: Added 0, 1, 1024

src/utils/multihash.rs
hanabi1224: changed to anyhow::bail

src/utils/multihash.rs
LesnyRumcajs: Let's leave the original version, then.

src/utils/multihash.rs
LesnyRumcajs: let's add a test case for it for completeness?

hanabi1224: Done.

LesnyRumcajs (approved):

akaladarshi (approved):

PR #5867, opened 2025-07-28
feat: add `--unordered` to `forest-cli snapshot export`

hanabi1224:
## Summary of changes

-

Closes

## Change checklist

### DIFF
--- a/src/ipld/util.rs
+++ b/src/ipld/util.rs
@@ -102,7 +102,7 @@ impl Iterator for DfsIter {
 
 enum Task {
     // Yield the block, don't visit it.
-    Emit(Cid),
+    Emit(Cid, Option<Vec<u8>>),
     // Visit all the elements, recursively.
     Iterate(VecDeque<Cid>),
 }
@@ -192,13 +192,19 @@ impl<DB: Blockstore, T: Borrow<Tipset>, ITER: Iterator<Item = T> + Unpin> Stream
         loop {
             while let Some(task) = this.dfs.front_mut() {
                 match task {
-                    Emit(cid) => {
-                        let cid = *cid;
-                        this.dfs.pop_front();
-                        if let Some(data) = this.db.get(&cid)? {
+                    Emit(_, _) => {
+                        if let Some(Emit(cid, data)) = this.dfs.pop_front() {
+                            let data = if let Some(data) = data {
+                                data
+                            } else if let Some(data) = this.db.get(&cid)? {
+                                data
+                            } else {
+                                return Poll::Ready(Some(Err(anyhow::anyhow!(
+                                    "missing key: {}",
+                                    cid
+                                ))));
+                            };
                             return Poll::Ready(Some(Ok(CarBlock { cid, data })));
-                        } else if fail_on_dead_links {
-                            return Poll::Ready(Some(Err(anyhow::anyhow!("missing key: {}", cid))));
                         }
                     }
                     Iterate(cid_vec) => {
@@ -212,10 +218,11 @@ impl<DB: Blockstore, T: Borrow<Tipset>, ITER: Iterator<Item = T> + Unpin> Stream
                                 if let Some(data) = this.db.get(&cid)? {
                                     if cid.codec() == fvm_ipld_encoding::DAG_CBOR {
                                         let new_values = extract_cids(&data)?;
-                                        cid_vec.reserve(new_values.len());
-
-                                        for v in new_values.into_iter().rev() {
-                                            cid_vec.push_front(v)
+                                        if !new_values.is_empty() {
+                                            cid_vec.reserve(new_values.len());
+                                            for v in new_values.into_iter().rev() {
+                                                cid_vec.push_front(v)
+                                            }
                                         }
                                     }
                                     return Poll::Ready(Some(Ok(CarBlock { cid, data })));
@@ -237,14 +244,15 @@ impl<DB: Blockstore, T: Borrow<Tipset>, ITER: Iterator<Item = T> + Unpin> Stream
             // yield the block without walking the graph it represents.
             if let Some(tipset) = this.tipset_iter.next() {
                 for block in tipset.borrow().block_headers() {
-                    if this.seen.insert(*block.cid()) {
+                    let (cid, data) = block.car_block()?;
+                    if this.seen.insert(cid) {
                         // Make sure we always yield a block otherwise.
-                        this.dfs.push_back(Emit(*block.cid()));
+                        this.dfs.push_back(Emit(*block.cid(), Some(data)));
 
                         if block.epoch == 0 {
                             // The genesis block has some kind of dummy parent that needs to be emitted.
                             for p in &block.parents {
-                                this.dfs.push_back(Emit(p));
+                                this.dfs.push_back(Emit(p, None));
                             }
                         }
 
@@ -287,7 +295,7 @@ pin_project! {
         block_recv_stream: Fuse<flume::r#async::RecvStream<'a, anyhow::Result<CarBlock>>>,
         extract_sender: flume::Sender<Cid>,
         stateroot_limit: ChainEpoch,
-        queue: Vec<Cid>,
+        queue: Vec<(Cid, Option<Vec<u8>>)>,
         fail_on_dead_links: bool,
     }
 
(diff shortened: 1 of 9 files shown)


sudo-shashank (dismissed):

src/ipld/util.rs
```diff
@@ -212,17 +216,17 @@ impl<DB: Blockstore, T: Borrow<Tipset>, ITER: Iterator<Item = T> + Unpin> Stream
                                 if let Some(data) = this.db.get(&cid)? {
                                     if cid.codec() == fvm_ipld_encoding::DAG_CBOR {
                                         let new_values = extract_cids(&data)?;
-                                        cid_vec.reserve(new_values.len());
-
-                                        for v in new_values.into_iter().rev() {
-                                            cid_vec.push_front(v)
+                                        if !new_values.is_empty() {
```
akaladarshi: nit: 

`if new_values.is_empty() { continue }`

src/ipld/util.rs
```diff
@@ -308,33 +318,23 @@ impl<DB, T> UnorderedChainStream<DB, T> {
...
     ITER: Iterator<Item = T> + Unpin + Send + 'static,
 >(
     db: Arc<DB>,
     tipset_iter: ITER,
     stateroot_limit: ChainEpoch,
-) -> UnorderedChainStream<DB, ITER> {
+    fail_on_dead_links: bool,
+) -> UnorderedChainStream<'a, DB, ITER> {
     let (sender, receiver) = flume::bounded(BLOCK_CHANNEL_LIMIT);
     let (extract_sender, extract_receiver) = flume::unbounded();
```
akaladarshi, line 328: Can we also use the `flume::bounded` for extract sender and receiver ?

src/ipld/util.rs
```diff
@@ -401,8 +407,7 @@ impl<DB: Blockstore + Send + Sync + 'static, T: Borrow<Tipset>, ITER: Iterator<I
     ) -> JoinHandle<anyhow::Result<()>> {
         task::spawn(async move {
             let mut handles = JoinSet::new();
-
-            for _ in 0..num_cpus::get() {
+            for _ in 0..num_cpus::get().clamp(1, 8) {
```
akaladarshi, line 405: Do you think it makes sense to accept this as params as well?

Basically allow user to pass the worker count otherwise use current values as default?

src/ipld/util.rs
```diff
@@ -415,7 +420,9 @@ impl<DB: Blockstore + Send + Sync + 'static, T: Borrow<Tipset>, ITER: Iterator<I
                                 if let Some(data) = db.get(&cid)? {
                                     if cid.codec() == fvm_ipld_encoding::DAG_CBOR {
                                         let mut new_values = extract_cids(&data)?;
-                                        cid_vec.append(&mut new_values);
+                                        if !new_values.is_empty() {
+                                            cid_vec.append(&mut new_values);
```
akaladarshi: Refactor it to:

```rust=
let new_values = extract_cids(&data)?;
if !new_values.is_empty() {
    cid_vec.reserve(new_values.len());
    cid_vec.extend(new_values);
}
```

akaladarshi: Do you think `cid_vec` could grow very large since we are also sending the `state_root` ?

Should we have batch processing here?

Something like this:

```rust=
const MAX_CID_QUEUE_SIZE: usize = 1000;

if cid_vec.len() > MAX_CID_QUEUE_SIZE {
    for cid in cid_vec.drain(MAX_CID_QUEUE_SIZE..) {
        let _ = extract_sender.send(cid);
    }
}
```

src/ipld/util.rs
```diff
@@ -448,102 +455,87 @@ impl<DB: Blockstore + Send + Sync + 'static, T: Borrow<Tipset>, ITER: Iterator<I
...
-                            // `block_receiver` are held by worker_handle and their counterparts -
-                            // by the main process. So those are either both functional or both
-                            // closed.
-                        } else if err == TryRecvError::Disconnected {
-                            panic!(
-                                "block_receiver can only be closed after extract_sender is empty"
-                            )
-                        }
+                    } else if this.extract_sender.is_empty() {
+                        this.worker_handle.abort();
```
akaladarshi: Right now we are aborting without checking if the `workers` are still processing the existing load.

Do you think we should make sure that all the workers have finished before aborting it here, we can send signal from the worker side (once all tipset tasks are done) and then abort.

akaladarshi (changes requested):
Some comments otherwise LGTM. Very nice work 👌🏽

akaladarshi:
Also do you thinks it's possible to divide the `UnorderedChainStream` poll_next function into small chunks, since there are lot of `if` `else` blocks there.

Just a suggestion, we can use introduce couple of different structs to handle different tasks:

```rust=
struct WorkerPool {
    handle: JoinHandle<anyhow::Result<()>>,
    work_sender: flume::Sender<WorkItem>,
    shutdown_sender: Option<oneshot::Sender<()>>,
}

enum WorkItem {
    Process(Cid),
    Shutdown,
}
```

src/ipld/util.rs
hanabi1224: This has been done internally in `Vec::append`

https://doc.rust-lang.org/stable/src/alloc/vec/mod.rs.html#2579
```rust
    pub fn append(&mut self, other: &mut Self) {
        unsafe {
            self.append_elements(other.as_slice() as _);
            other.set_len(0);
        }
    }

    /// Appends elements to `self` from other buffer.
    #[cfg(not(no_global_oom_handling))]
    #[inline]
    #[track_caller]
    unsafe fn append_elements(&mut self, other: *const [T]) {
        let count = other.len();
        self.reserve(count);
        let len = self.len();
        unsafe { ptr::copy_nonoverlapping(other as *const T, self.as_mut_ptr().add(len), count) };
        self.len += count;
    }
```

src/ipld/util.rs
hanabi1224: That changes the logic, it always returns sth here

hanabi1224:
Hey @akaladarshi Thanks for your review! Regarding `stream_chain` and `unordered_stream_chain` refactoring, I've extracted the changes into https://github.com/ChainSafe/forest/pull/5868 please take a look. Your ideas on furture refactoring `unordered_stream_chain` sound great but it's out of scope of this PR. This PR tries to revive the dead `unordered_stream_chain` function so that we can test it via `forest-cli`. I'd suggest we make decisions to either further refactor it or remove it after proving its correctness and usefulness. What do you think?

akaladarshi:
@hanabi1224 I didn't realise this just for reviving the dead code. We can hold on to the changes for now, we can make them once we are sure this what we are going with. 

I will approve the PR, we can revisit the comments later.

akaladarshi (dismissed):

src/ipld/util.rs
hanabi1224: Great point. The old logic did not handle the shutdown gracefully indeed. Fixed by closing the `extract_sender` instead of forcing `worker_handle.abort`

src/ipld/util.rs
hanabi1224: Let's tackle this in another PR

src/ipld/util.rs
hanabi1224: I think it's our responsibility to tune this value. It might turn out this `unordered_stream_chain` is not fast and we end up removing it.

src/ipld/util.rs
hanabi1224: `extract_sender` is bounded, and this will likely get stuck. Actually the original code gets stuck on CI and I have to change it to `send_async`

src/ipld/util.rs
```diff
@@ -437,106 +452,109 @@
...
-                            panic!(
-                                "block_receiver can only be closed after extract_sender is empty"
-                            )
-                        }
+                    } else if let Some(extract_sender) = this.extract_sender
+                        && extract_sender.is_empty()
+                    {
+                        // drop the sender to abort the woker task
+                        *this.extract_sender = None;
                     }
```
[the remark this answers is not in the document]
hanabi1224: It's been stablized on rust 1.88 and we are now on 1.89

src/ipld/util.rs
```diff
@@ -437,106 +452,109 @@ impl<DB: Blockstore + Send + Sync + 'static, T: Borrow<Tipset>, ITER: Iterator<I
...
+                if db.has(&cid)? {
+                    send(extract_sender, cid)?;
+                } else if fail_on_dead_links {
+                    queue.push((cid, None));
+                } else {
+                    seen.lock().insert(cid);
+                }
+            }
+            Ok(())
+        }
```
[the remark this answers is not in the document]
hanabi1224: Sender is dropped to gracefully shutdown the worker_handle and end the stream. It's an error if there're more to send.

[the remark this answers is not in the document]
hanabi1224: Fixed.

src/ipld/util.rs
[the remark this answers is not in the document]
hanabi1224: Typo fixed.

PR #6087, opened 2025-09-16
refactor(actor_state): allow easy version upgrades

akaladarshi:
## Summary of changes

- Refactors the actor state so we can upgrade to new version seamlessly by just adding `V18` to the macro.
- Introduced `default_latest_version` method for all actors, so the actor states can return the current latest actor version.
- Instead of hardcoding `V16` use latest default version for actor states while deserialising the lotus json from the data.

Closes: https://github.com/ChainSafe/forest/issues/6050

## Change checklist


src/lotus_json/actors/states/evm_state.rs
```diff
@@ -13,41 +13,166 @@ pub struct EVMStateLotusJson {
...
+                            tombstone_state.into_lotus_json()
+                        }),
+                        transient_data: None,
+                    }
+                }
+
+                fn from_lotus_json(lotus_json: Self::LotusJson) -> Self {
+                    let tombstone = lotus_json.tombstone.map(|tombstone_lotus| {
+                        fil_actor_evm_state::[<v $version>]::Tombstone {
+                            origin: tombstone_lotus.orign.into(),
```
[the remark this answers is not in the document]
akaladarshi: Done

src/lotus_json/entry.rs
```diff
@@ -24,11 +24,8 @@ impl HasLotusJson for Entry {
...
                 "MethodNum": 10
             }),
-            // Create a test entry
-            Entry::V16(fil_actor_cron_state::v16::Entry {
-                receiver: Default::default(),
-                method_num: 0,
-            }),
+            // Create a test entry using the latest version
+            Entry::default_latest_version(Address::default().into(), 0),
         )]
```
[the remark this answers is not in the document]
akaladarshi: Done.

src/lotus_json/transient_data.rs
```diff
@@ -13,85 +13,99 @@ pub struct TransientDataLotusJson {
...
+                            }
+                        }},
+                        Self {
+                            transient_data_state: Default::default(),
+                            transient_data_lifespan: fil_actor_evm_state::[<v $version>]::TransientDataLifespan {
+                                origin: Default::default(),
+                                nonce: Default::default(),
+                            },
+                        },
+                    )]
```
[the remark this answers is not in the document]
akaladarshi: Done.

src/lotus_json/filter_estimate.rs
```diff
@@ -15,33 +16,43 @@ pub struct FilterEstimateLotusJson {
...
-                "Velocity" : "0",
-            }),
-            FilterEstimate::default(),
-        )]
-    }
+                #[cfg(test)]
+                fn snapshots() -> Vec<(serde_json::Value, Self)> {
+                    vec![(
+                        json!({
+                            "PositionEstimate": "0",
```
LesnyRumcajs, line 31: Was this incorrect?

src/shim/actors/builtin/account/mod.rs
```diff
@@ -25,7 +25,7 @@ pub enum State {
 }
 
 impl State {
-    pub fn pubkey_address(&self) -> Address {
+    pub fn pubkey_address(&self) -> fvm_shared2::address::Address {
```
LesnyRumcajs, line 28: Probably outside of this PR, but do we need to have this old type here? From performance perspective, we're mostly calling the latest actors, so we should avoid conversions on those types and make the "old" versions do conversions.

src/lotus_json/filter_estimate.rs
akaladarshi: Yeah it was incorrect, but it doesn't affect anything specifically. the actual field name was correct.

src/shim/actors/builtin/account/mod.rs
akaladarshi: There are couple of places where we are only using the old version of types such as `Address`, but we should remove it and switch to the latest versions. 

I see lot of shims which uses the latest version but not all of the types have the latest version, something like `FilterEstimate` doesn't have shim for the latest version.

akaladarshi:
> function

Actually No, we can't use the `Default` trait because it doesn't accept parameters, and actor states require real data to be constructed meaningful latest state version.

Idea here was instead of hardcoding V16 or V17 in `from_lotus_json` function we should just use latest version with the correct json data.

src/lotus_json/filter_estimate.rs
LesnyRumcajs: Can you help me understand this? If it was incorrect, why wasn't it caught in any tests?

src/shim/actors/builtin/account/mod.rs
LesnyRumcajs: @coderabbitai create an issue to ensure shims are using the latest versions of the data types.

PR #6103, opened 2025-09-22
fix: `eth_getTransactionCount` to return the actual count, not parent

LesnyRumcajs:
## Summary of changes

-

Closes

## Change checklist


elmattic:
> Sadly, no bonus points for tests here.

Is this a breaking change? Have you considered adding an `rpcsnap` file to verify that the RPC method now behaves correctly?

LesnyRumcajs:
> > Sadly, no bonus points for tests here.
> 
> Is this a breaking change? Have you considered adding an `rpcsnap` file to verify that the RPC method now behaves correctly?

I could add it, but the problem is that it'd be difficult to re-create if needed. So I decided not to add it at all.

In general, such tests would be better suited for running on devnet, but we don't have them (yet).

Issue #6113, opened 2025-09-24
Implement `ChainGetFinalizedTipset` API


akaladarshi:
# Summary

- Lotus https://github.com/filecoin-project/lotus/pull/13299 has introduced a new v1 API `ChainGetFinalizedTipset`, forest should support this as well.

PR #6133, opened 2025-10-07
feat(api): implement the `ChainGetFinalizedTipset` API

akaladarshi:
## Summary of changes

- Adds implementation for the `ChainGetFinalizedTipset` API
- Adds snapshot tests for the ChainGetFinalizedTipset API:
  - The test only verifies `basic` check since it is dependent on F3 and the latest tipset.

Closes https://github.com/ChainSafe/forest/issues/6113

## Change checklist

## Summary by CodeRabbit

- New Features
  - Finalized tipset retrieval now prefers the latest network certificate when available, with automatic fallback to consensus finality for reliability.
- Bug Fixes
  - Replaced a stubbed finalized-tipset response with real resolution logic and improved error handling and logging to avoid incorrect or empty results.
- Tests
  - Added an API test for finalized tipset retrieval and expanded snapshot data to cover the new behavior.

### DIFF
--- a/src/rpc/methods/chain.rs
+++ b/src/rpc/methods/chain.rs
@@ -145,7 +145,7 @@ impl RpcMethod<0> for ChainGetFinalizedTipset {
         (): Self::Params,
     ) -> Result<Self::Ok, ServerError> {
         let head = ctx.chain_store().heaviest_tipset();
-        let ec_finality_epoch = head.epoch() - ctx.chain_config().policy.chain_finality;
+        let ec_finality_epoch = (head.epoch() - ctx.chain_config().policy.chain_finality).max(0);
 
         // Either get the f3 finalized tipset or the ec finalized tipset
         match get_f3_finality_tipset(&ctx, ec_finality_epoch).await {


src/rpc/methods/chain.rs
```diff
@@ -140,11 +140,62 @@ impl RpcMethod<0> for ChainGetFinalizedTipset {
...
     type Ok = Tipset;
 
-    async fn handle(_ctx: Ctx<impl Blockstore>, (): Self::Params) -> Result<Self::Ok, ServerError> {
-        Err(ServerError::stubbed_for_openrpc())
+    async fn handle(
+        ctx: Ctx<impl Blockstore + Send + Sync + 'static>,
+        (): Self::Params,
+    ) -> Result<Self::Ok, ServerError> {
+        let head = ctx.chain_store().heaviest_tipset();
+        let ec_finality_epoch = head.epoch() - ctx.chain_config().policy.chain_finality;
```
hanabi1224:
```suggestion
        let ec_finality_epoch = (head.epoch() - ctx.chain_config().policy.chain_finality).max(0);
```

src/rpc/methods/chain.rs
```diff
@@ -140,11 +140,62 @@ impl RpcMethod<0> for ChainGetFinalizedTipset {
...
+
+        // Either get the f3 finalized tipset or the ec finalized tipset
+        match get_f3_finality_tipset(&ctx, ec_finality_epoch).await {
+            Ok(f3_tipset) => {
+                tracing::debug!("Using F3 finalized tipset at epoch {}", f3_tipset.epoch());
+                Ok((*f3_tipset).clone())
```
elmattic, line 154: Could we avoid cloning here?

src/rpc/methods/chain.rs
```diff
@@ -140,11 +140,62 @@ impl RpcMethod<0> for ChainGetFinalizedTipset {
...
+// get f3 finalized tipset based on ec finality epoch
+async fn get_f3_finality_tipset<DB: Blockstore + Sync + Send + 'static>(
+    ctx: &Ctx<DB>,
+    ec_finality_epoch: ChainEpoch,
+) -> Result<Arc<Tipset>> {
+    let f3_finalized_cert = crate::rpc::f3::F3GetLatestCertificate::handle(ctx.clone(), ())
+        .await
+        .map_err(|e| anyhow::anyhow!("Failed to get F3 certificate: {}", e))?;
+
+    let f3_finalized_head = f3_finalized_cert.chain_head();
```
hanabi1224, line 179: `f3_finalized_cert.chain_head()` could be newer than chain head during catchup. Lotus does not seem to handle that case, should we handle it? e.g. marking finalized tipset in `F3.Finalize` RPC method instead.

hanabi1224: We could track this in a separate issue, for now the logic matches Lotus

src/rpc/methods/chain.rs
akaladarshi: I don't think so `f3_tipset` wrapped in `Arc` so we will have to clone only.

src/rpc/methods/chain.rs
akaladarshi: Done.

PR #6165, opened 2025-10-15
fix(gc): schedule automatic GC only when the node is in sync

hanabi1224:
## Summary of changes

This fixes a rare condition when GC kicks in during node catchup and breaks the block validation

- also run `cargo update -p half` to update the yanked [`half@2.7.0`](https://crates.io/crates/half/2.7.0)

Closes

## Change checklist

## Summary by CodeRabbit

- New Features
  - Live sync status is now forwarded at startup so background services act on current node sync state.

- Improvements
  - Snapshot garbage collector only runs when the node is near the network head.
  - Scheduling logs now include network head context.
  - Snapshot export logs report total duration for monitoring.

- Bug Fixes
  - Export failures now reset state to allow re-triggering of export attempts.

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -10,7 +10,7 @@ use crate::blocks::Tipset;
 use crate::chain::HeadChange;
 use crate::chain::index::ResolveNullTipset;
 use crate::chain_sync::network_context::SyncNetworkContext;
-use crate::chain_sync::{ChainFollower, SyncStatusReport};
+use crate::chain_sync::{ChainFollower, SyncStatus};
 use crate::cli_shared::snapshot;
 use crate::cli_shared::{
     chain_path,
@@ -36,7 +36,6 @@ use crate::utils::{proofs_api::ensure_proof_params_downloaded, version::FOREST_V
 use anyhow::{Context as _, bail};
 use dialoguer::theme::ColorfulTheme;
 use futures::{Future, FutureExt, select};
-use parking_lot::RwLock;
 use std::path::Path;
 use std::sync::Arc;
 use std::sync::OnceLock;
@@ -578,7 +577,7 @@ pub(super) async fn start_services(
     opts: &CliOpts,
     mut config: Config,
     shutdown_send: mpsc::Sender<()>,
-    on_app_context_and_db_initialized: impl FnOnce(&AppContext, Arc<RwLock<SyncStatusReport>>),
+    on_app_context_and_db_initialized: impl FnOnce(&AppContext, SyncStatus),
 ) -> anyhow::Result<()> {
     // Cleanup the collector prometheus metrics registry on start
     crate::metrics::reset_collector_registry();
--- a/src/db/gc/snapshot.rs
+++ b/src/db/gc/snapshot.rs
@@ -74,7 +74,7 @@ pub struct SnapshotGarbageCollector<DB> {
     running: AtomicBool,
     blessed_lite_snapshot: RwLock<Option<PathBuf>>,
     db: RwLock<Option<Arc<DB>>>,
-    sync_status: RwLock<Option<Arc<RwLock<crate::chain_sync::SyncStatusReport>>>>,
+    sync_status: RwLock<Option<crate::chain_sync::SyncStatus>>,
     // On mainnet, it takes ~50MiB-200MiB RAM, depending on the time cost of snapshot export
     memory_db: RwLock<Option<HashMap<Cid, Vec<u8>>>>,
     memory_db_head_key: RwLock<Option<TipsetKey>>,
@@ -134,7 +134,7 @@ where
         *self.car_db_head_epoch.write() = Some(epoch);
     }
 
-    pub fn set_sync_status(&self, sync_status: Arc<RwLock<crate::chain_sync::SyncStatusReport>>) {
+    pub fn set_sync_status(&self, sync_status: crate::chain_sync::SyncStatus) {
         *self.sync_status.write() = Some(sync_status)
     }
 
--- a/src/chain_sync/chain_follower.rs
+++ b/src/chain_sync/chain_follower.rs
@@ -15,40 +15,36 @@
 //!
 //! The state machine does not do any network requests or validation. Those are
 //! handled by an external actor.
-use crate::libp2p::hello::HelloRequest;
-use crate::message_pool::MessagePool;
-use crate::message_pool::MpoolRpcProvider;
-use crate::networks::calculate_expected_epoch;
-use crate::shim::clock::ChainEpoch;
-use crate::state_manager::StateManager;
-use crate::utils::misc::env::is_env_truthy;
+
+use super::network_context::SyncNetworkContext;
+use crate::{
+    blocks::{Block, FullTipset, Tipset, TipsetKey},
+    chain::ChainStore,
+    chain_sync::{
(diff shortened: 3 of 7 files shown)


src/db/gc/snapshot.rs
```diff
@@ -170,18 +176,22 @@ where
...
             if !self.running.load(Ordering::Relaxed)
-                && let Some(db) = &*self.db.read()
                 && let Some(car_db_head_epoch) = *self.car_db_head_epoch.read()
-                && let Ok(head_key) = HeaviestTipsetKeyProvider::heaviest_tipset_key(db)
-                && let Ok(head) = Tipset::load_required(db, &head_key)
+                && let Some(sync_status) = &*self.sync_status.read()
             {
-                let head_epoch = head.epoch();
-                if head_epoch - car_db_head_epoch >= snap_gc_interval_epochs
+                const IN_SYNC_EPOCH_THRESHOLD: i64 = 2;
```
hanabi1224: I notice `sync_status.status` does not refect the reality after a long catchup (potential bug?), so checking `head_epoch + IN_SYNC_EPOCH_THRESHOLD >= network_head_epoch` manually here

akaladarshi: We should increase the `IN_SYNC_EPOCH_THRESHOLD` to atleast 10, I think we are using this value in `SYNCED_EPOCH_THRESHOLD` as well?

src/db/gc/snapshot.rs
```diff
@@ -132,6 +134,10 @@ where
         *self.car_db_head_epoch.write() = Some(epoch);
     }
 
+    pub fn set_sync_status(&self, sync_status: Arc<RwLock<crate::chain_sync::SyncStatusReport>>) {
+        *self.sync_status.write() = Some(sync_status)
+    }
+
     pub async fn event_loop(&self) {
         while self.trigger_rx.recv_async().await.is_ok() {
             if self.running.load(Ordering::Relaxed) {
```
akaladarshi, line 143: Unrelated to current changes but, should we reset the `running` bool so we can run GC again? 

```
if let Err(e) = self.export_snapshot().await {
   tracing::warn!("{e}");
   self.running.store(false, Ordering::Relaxed);
}
```

src/daemon/mod.rs
```diff
@@ -576,7 +578,7 @@ pub(super) async fn start_services(
     opts: &CliOpts,
     mut config: Config,
     shutdown_send: mpsc::Sender<()>,
-    on_app_context_and_db_initialized: impl Fn(&AppContext),
+    on_app_context_and_db_initialized: impl Fn(&AppContext, Arc<RwLock<SyncStatusReport>>),
```
akaladarshi: `Fn` -> `FnOnce`

src/db/gc/snapshot.rs
```diff
@@ -74,6 +74,7 @@ pub struct SnapshotGarbageCollector<DB> {
     running: AtomicBool,
     blessed_lite_snapshot: RwLock<Option<PathBuf>>,
     db: RwLock<Option<Arc<DB>>>,
+    sync_status: RwLock<Option<Arc<RwLock<crate::chain_sync::SyncStatusReport>>>>,
```
akaladarshi: Do you think this will be better here?

`RwLock<Option<Arc<RwLock<crate::chain_sync::SyncStatusReport>>>>` -> `OnceLock<Arc<RwLock<crate::chain_sync::SyncStatusReport>>>`

This way we can avoid doing: `&*sync_status.read()`

src/db/gc/snapshot.rs
akaladarshi: This is a potential bug `SyncStatusReport`, I will file an issue for it.

src/db/gc/snapshot.rs
hanabi1224: It's set more than once after every GC

src/db/gc/snapshot.rs
akaladarshi: This is because there we are using time diff only:

```rust
let time_diff = now_ts.saturating_sub(heaviest.min_timestamp());


if time_diff < seconds_per_epoch as u64 * SYNCED_EPOCH_THRESHOLD {
 NodeSyncStatus::Synced
} else {
  NodeSyncStatus::Syncing
}
```

But you can use 

`if sync_status.is_synced() && sync_status.active_forks.is_empty()`

src/db/gc/snapshot.rs
akaladarshi: Yeah, I missed this GC restarts the node.

src/db/gc/snapshot.rs
hanabi1224: Ah, now I know why `forest-cli sync wait` exits early. I can check `sync_status.status == NodeSyncStatus::Synced` here.

BTW, why `SYNCED_EPOCH_THRESHOLD` has to be 10(5min)? Could we change it to a smaller value like 2(1min)?

src/daemon/mod.rs
hanabi1224: Fixed.

src/db/gc/snapshot.rs
akaladarshi: No, This is not a fixed value we just decided that because we didn't want the node to be too far behind.

As for setting the value 2, this is good for ideal condition but the if nodes are slow or something 2 would be very less, it might actually never reach `Synced` status.

src/db/gc/snapshot.rs
hanabi1224: Good catch! Fixed.

src/db/gc/snapshot.rs
hanabi1224: Fixed!

hanabi1224: Fixed.

src/db/gc/snapshot.rs
elmattic: Why do we need two `RwLock`s here?

elmattic: Could we simplify this type?

src/db/gc/snapshot.rs
```diff
@@ -170,18 +177,22 @@ where
...
-                && let Ok(head_key) = HeaviestTipsetKeyProvider::heaviest_tipset_key(db)
-                && let Ok(head) = Tipset::load_required(db, &head_key)
+                && let Some(sync_status) = &*self.sync_status.read()
             {
-                let head_epoch = head.epoch();
-                if head_epoch - car_db_head_epoch >= snap_gc_interval_epochs
+                let sync_status = &*sync_status.read();
+                let network_head_epoch = sync_status.network_head_epoch;
+                let head_epoch = sync_status.current_head_epoch;
+                if head_epoch > 0
```
elmattic: nit: this test condition is quite a beast, maybe it's time to break it down into functions and add some comments.

elmattic (approved):

src/db/gc/snapshot.rs
hanabi1224: I've added comment to each condition

src/db/gc/snapshot.rs
hanabi1224: @elmattic simplified type by adding `pub type SyncStatus = Arc<RwLock<SyncStatusReport>>;`

akaladarshi (approved):
LGTM. Nice Work!

Issue #6170, opened 2025-10-20
Extend Eth and Filecoin Basic API's to support V2


akaladarshi:
# Summary

Right now Forest only supports `V1` and `V2` API's, we need to expand to V2 API's

# Completion Criteria

Implement Forest Basic API's (which doesn't rely on tipset or F3 API's)

- EthBasic
  - [x] EthChainId
  - [x] EthProtocolVersion
  - [x] EthSyncing
  - [x] EthAccounts
  - [x] Web3ClientVersion
  - [x] NetListening
  - [x] NetVersion
- EthSend
  - [x] EthSendRawTransaction
- Filecoin
  - [ ] EthAddressToFilecoinAddress
- Chain
  - [ ] ChainGetTipSet
- EthTrace
  - [ ] EthTraceTransaction
- EthTransaction
  - [ ] EthGetTransactionByBlockHashAndIndex
  - [ ] EthGetTransactionHashByCid
  - [ ] EthGetMessageCidByTransactionHash
  - [ ] EthGetTransactionReceiptLimited
  - [ ] EthBlockNumber
  - [ ] EthGetBlockByHash
  - [ ] EthGetTransactionByHash
  - [ ] EthGetTransactionByHashLimited
  - [ ] EthGetTransactionReceipt
- EthGas
  - [ ] EthGasPrice
  - [ ] EthMaxPriorityFeePerGas
- EthEvents
  - [ ] EthNewBlockFilter
  - [ ] EthNewPendingTransactionFilter
  - [ ] EthUninstallFilter
  - [ ] EthGetFilterChanges
  - [ ] EthGetFilterLogs
  - [ ] EthSubscribe
  - [ ] EthUnsubscribe

# Additional Links & Resources

PR #6171, opened 2025-10-20
feat: enable V2 API support for basic Eth and Filecoin RPC methods

akaladarshi:
## Summary of changes

- Enable `V2` API support for all the Eth and Filecoin API's which doesn't rely on the F3 Finality
- Added temporary `ApiPaths::all_with_v2` so we can rollout incremental V2 support and later on delete the `all_with_v2` and use `all` direcly.

Closes: https://github.com/ChainSafe/forest/issues/6170

## Change checklist

## Summary by CodeRabbit

* **New Features**
  * Enabled V2 API route support for many Ethereum RPC methods and Net endpoints (expanded V2 coverage).

* **Chores**
  * Updated routing configuration so numerous RPC methods use the V2 route set to support incremental migration.
  * Added a temporary helper to expose V2 routes during the migration.

* **Documentation**
  * Updated changelog noting V2 route availability for affected RPC methods.

### DIFF
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -756,7 +756,7 @@ impl RpcMethod<0> for EthBlockNumber {
     const NAME: &'static str = "Filecoin.EthBlockNumber";
     const NAME_ALIAS: Option<&'static str> = Some("eth_blockNumber");
     const PARAM_NAMES: [&'static str; 0] = [];
-    const API_PATHS: BitFlags<ApiPaths> = ApiPaths::all();
+    const API_PATHS: BitFlags<ApiPaths> = ApiPaths::all_with_v2();
     const PERMISSION: Permission = Permission::Read;
 
     type Params = ();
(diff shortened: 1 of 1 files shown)


src/rpc/reflect/mod.rs
```diff
@@ -134,6 +134,13 @@ impl ApiPaths {
         make_bitflags!(Self::{ V0 | V1 })
     }
 
+    // Remove this helper once all RPC methods are migrated to V2.
+    // We're incrementally migrating methods to V2 support. Once complete,
+    // update all() to include V2 and remove this temporary helper.
+    pub const fn with_v2() -> BitFlags<Self> {
```
LesnyRumcajs: The method name is a bit misleading. When calling `ApiPaths::with_v2` I'd expect **only** the `v2` to be set.

```suggestion
    pub const fn all_with_v2() -> BitFlags<Self> {
```

For the logic - it'd be cleaner to call `all` and just set `v2`.

akaladarshi: Done.

LesnyRumcajs (approved):

Issue #6400, opened 2026-01-09
`eth_subscribe` returns wrong data structure for `newHeads`


LesnyRumcajs:
## Describe the bug

In the `eth_subscribe` method, with `newHeads`, Forest returns incorrect data structure. This causes errors with tooling such as Filecoin Subway (see the issue where it was discovered https://github.com/ChainSafe/forest/issues/6334) 

Forest:

```
❯ websocat ws://127.0.0.1:2345/rpc/v1
{"jsonrpc":"2.0","id":1,"method":"eth_subscribe","params":["newHeads"]}
{"jsonrpc":"2.0","id":1,"result":"0xaa047d05aff3e4d6b65aace000dfb863eca9b077c5bc78920bdd799a8b267493"}
[...]

PR #6402, opened 2026-01-09
fix: use ethereum block in `eth_subscribe` `newHeads`

LesnyRumcajs:
## Summary of changes

- changed the data structure returned by the `newHeads` response to Ethereum block (it was an array of tipset headers (???)). `filecoin-subway` doesn't complain anymore.

Closes https://github.com/ChainSafe/forest/issues/6400

## Change checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscription for newHeads now delivers Ethereum-compatible block payloads and continues streaming if individual conversions fail.

* **New Features**
  * Added option to receive full transaction objects or only transaction hashes in block responses.

* **Compatibility**
  * Subscription payload shape updated to reflect Ethereum block format.

### DIFF
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -540,6 +540,12 @@ pub enum TxInfo {
     Full,
 }
 
+impl From<bool> for TxInfo {
+    fn from(full: bool) -> Self {
+        if full { TxInfo::Full } else { TxInfo::Hash }
+    }
+}
+
 impl Block {
     pub fn new(has_transactions: bool, tipset_len: usize) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


src/rpc/methods/eth.rs
```diff
@@ -1637,7 +1732,12 @@ impl RpcMethod<2> for EthGetBlockByNumberV2 {
     ) -> Result<Self::Ok, ServerError> {
         let ts = tipset_by_block_number_or_hash_v2(&ctx, block_param, ResolveNullTipset::TakeOlder)
             .await?;
-        let block = block_from_filecoin_tipset(ctx, ts, full_tx_info).await?;
+        let tx_info = if full_tx_info {
+            TxInfo::Full
+        } else {
+            TxInfo::Hash
+        };
```
sudo-shashank: we can add an impl `From<bool> for TxInfo` to avoid doing this multiple times

LesnyRumcajs: I'd prefer to be more verbose. That said, I noticed there's some existing code I didn't know of that could be reused.

LesnyRumcajs: Actually, after some consideration and looking into other usages, I thin kit's a good idea.

PR #6421, opened 2026-01-14
fix(rpc): backfill tipset messages from p2p network in a few RPC methods

hanabi1224:
## Summary of changes

- backfill tipset messages from p2p network when `FOREST_RPC_BACKFILL_FULL_TIPSET_FROM_NETWORK=1`

Closes

## Change checklist

## Summary by CodeRabbit

* **Refactor**
  * Removed an optional in-memory tipset reconstruction path; RPC handlers now use a single async tipset resolution flow.

* **New Features**
  * Added FOREST_RPC_BACKFILL_FULL_TIPSET_FROM_NETWORK env var to optionally backfill full tipsets from the network.

* **API**
  * Public surface expanded with a centralized tipset retrieval helper for consistent usage across RPCs.

### DIFF
--- a/src/rpc/methods/chain.rs
+++ b/src/rpc/methods/chain.rs
@@ -18,7 +18,10 @@ use crate::lotus_json::{HasLotusJson, LotusJson, lotus_json_with_self};
 #[cfg(test)]
 use crate::lotus_json::{assert_all_snapshots, assert_unchanged_via_json};
 use crate::message::{ChainMessage, SignedMessage};
-use crate::rpc::eth::{EthLog, eth_logs_with_filter, types::ApiHeaders, types::EthFilterSpec};
+use crate::rpc::eth::Block as EthBlock;
+use crate::rpc::eth::{
+    EthLog, TxInfo, eth_logs_with_filter, types::ApiHeaders, types::EthFilterSpec,
+};
 use crate::rpc::f3::F3ExportLatestSnapshot;
 use crate::rpc::types::*;
 use crate::rpc::{ApiPaths, Ctx, EthEventHandler, Permission, RpcMethod, ServerError};
@@ -76,8 +79,8 @@ static CHAIN_EXPORT_LOCK: LazyLock<Mutex<Option<CancellationToken>>> =
 ///
 /// Spawns an internal `tokio` task that can be aborted anytime via the returned `JoinHandle`,
 /// allowing manual cleanup if needed.
-pub(crate) fn new_heads<DB: Blockstore>(
-    data: &crate::rpc::RPCState<DB>,
+pub(crate) fn new_heads<DB: Blockstore + Send + Sync + 'static>(
+    data: Ctx<DB>,
 ) -> (Subscriber<ApiHeaders>, JoinHandle<()>) {
     let (sender, receiver) = broadcast::channel(HEAD_CHANNEL_CAPACITY);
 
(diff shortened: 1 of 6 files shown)


src/rpc/methods/chain.rs
```diff
@@ -355,13 +356,13 @@ impl RpcMethod<1> for ChainGetMessagesInTipset {
     type Ok = Vec<ApiMessage>;
 
     async fn handle(
-        ctx: Ctx<impl Blockstore>,
+        ctx: Ctx<impl Blockstore + Send + Sync + 'static>,
         (ApiTipsetKey(tipset_key),): Self::Params,
     ) -> Result<Self::Ok, ServerError> {
         let tipset = ctx
             .chain_store()
             .load_required_tipset_or_heaviest(&tipset_key)?;
```
akaladarshi, line 378: Small doubt: Is it possible here that `tipset_key` is available in the store but the messages (full-tipset) is not available?

hanabi1224: Yes. Messages are periodically garbage collected.

akaladarshi: So the `tipset` is not garbage collected 👍🏼.

hanabi1224: Block headers are the spine of the chain that are never garbage collected

LesnyRumcajs (approved):

docs/docs/users/reference/env_variables.md
```diff
@@ -53,6 +53,7 @@ process.
...
 | `FOREST_JWT_DISABLE_EXP_VALIDATION`                       | 1 or true                        | empty                                          | 1                                                             | Whether or not to disable JWT expiration validation                                                                   |
 | `FOREST_ETH_BLOCK_CACHE_SIZE`                             | positive integer                 | 500                                            | 1                                                             | The size of Eth block cache                                                                                           |
+| `FOREST_RPC_BACKFILL_FULL_TIPSET_FROM_NETWORK`            | 1 or true                        | false                                          | 1                                                             | Whether or not to backfill full tipsets from the p2p network                                                          |
```
LesnyRumcajs, line 56: Actually, let's add some warning that enabling this might lead to excessive disk and bandwidth usage.

Also, let's add a CHANGELOG entry.

hanabi1224: Fixed.

src/rpc/methods/chain.rs
```diff
@@ -1301,13 +1303,26 @@ pub(crate) fn chain_notify<DB: Blockstore>(
...
+async fn load_api_messages_from_tipset<DB: Blockstore + Send + Sync + 'static>(
+    ctx: &crate::rpc::RPCState<DB>,
+    tipset_keys: &TipsetKey,
 ) -> Result<Vec<ApiMessage>, ServerError> {
-    let full_tipset = tipset
-        .fill_from_blockstore(store)
-        .context("Failed to load full tipset")?;
+    static SHOULD_BACKFILL: LazyLock<bool> =
+        LazyLock::new(|| is_env_truthy("FOREST_RPC_BACKFILL_FULL_TIPSET_FROM_NETWORK"));
+    let full_tipset = if *SHOULD_BACKFILL {
```
LesnyRumcajs, line 1332: Wouldn't this cause a warning on every call?

hanabi1224: It's intended. I feel printing the warning only on the first call is less visible. What do you think?

LesnyRumcajs: @hanabi1224 I don't think we need this. If the user opted-in and set this variable then we should assume they read documentation and the warning on resource exhaustion. On top of that, if they made such decision consciously then the warning would just annoy them.

hanabi1224: @LesnyRumcajs fixed.

PR #6423, opened 2026-01-15
Implement expensive fork check

### LINKED ISSUE #5965: Investigate expensive fork handling in Forest
## Describe the bug

Forest does not currently handle "expensive forks" when executing messages, which is implemented in Lotus. Without this check, RPC calls can become very slow if an expensive migration is triggered.

Forest should return an `ErrExpensiveFork` error in these cases, similar to Lotus, to prevent performance degradation.

Lotus uses `HasExpensiveForkBetween` to check if executing tipsets between specified heights would trigger an expensive migration, excluding migrations at the target height.

Forest needs to implement similar functionality.

## To reproduce

1. Run Forest with RPC enabled
2. Create and send a request that will trigger such expensive fork
3. Observe slowdown

## Log output

<details>
  <summary>Log Output</summary>
  
  ```Paste log output here
  paste log output...
  ```
</details>

## Expected behaviour

For all RPC methods affected by this issue (an exhaustive list should be established first), we should trigger an error instead of performing the expensive migration.

It is up to the implementer to determine which network upgrades can be considered fast enough to avoid being included in the expensive fork list. Likely candidates would be upgrades that take no longer than 100ms.

- OS:
- Branch/commit
- Hardware

Look for `Expensive: true` in https://github.com/filecoin-project/lotus/blob/master/chain/consensus/filcns/upgrades.go

`HasExpensiveForkBetween`  function: https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/forks.go#L230-L237

sudo-shashank:
## Summary of changes

- Implement `has_expensive_fork_between` to check if executing tipsets between specified heights would trigger an expensive migration, excluding migrations at the target height.
- Return `ErrExpensiveFork` error , matching Lotus behavior, to prevent performance degradation.

Closes #5965 

## Change checklist

## Summary by CodeRabbit

* **New Features**
  * Safeguards to block message execution when an expensive state migration is detected between chain epochs.
  * Added expanded public checks to query expensive migration heights and detect expensive forks between epochs.
  * Introduced a distinct public error variant to surface expensive-fork conditions.

* **Tests**
  * Added unit tests validating detection of expensive migrations between epochs.


src/rpc/methods/eth.rs
```diff
@@ -2017,6 +2018,18 @@ async fn apply_message<DB>(
...
 {
+    if let Some(tipset) = tipset.as_ref()
+        && tipset.epoch() > 0
+    {
+        let parent_ts = Tipset::load_required(ctx.store(), tipset.parents())
+            .context("failed to load parent tipset")?;
+
+        ensure!(
+            !ctx.state_manager
+                .has_expensive_fork_between(parent_ts.epoch(), tipset.epoch() + 1),
```
hanabi1224, line 2029: Why would it be 2 epochs in the range?

sudo-shashank: i guess lotus does this to early detect an expensive fork

PR #6453, opened 2026-01-20
fix: reduce repetitive code in enum match blocks

hanabi1224:
## Summary of changes

-

Closes

## Change checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for an additional virtual machine backend to expand execution platform options.

* **Refactor**
  * Enabled delegated enum conversions across interpreter and actor components to reduce duplication and simplify multi-version handling.
  * Simplified address handling by removing unnecessary conversions in several resolution and serialization paths.

* **Chores**
  * Added a build dependency to enable delegation tooling and bumped a dev-dependency version.


src/interpreter/vm.rs
```diff
@@ -272,11 +274,7 @@ where
 
     /// Flush stores in VM and return state root.
     pub fn flush(&mut self) -> anyhow::Result<Cid> {
-        match self {
-            VM::VM2(fvm_executor) => Ok(fvm_executor.flush()?),
-            VM::VM3(fvm_executor) => Ok(fvm_executor.flush()?),
-            VM::VM4(fvm_executor) => Ok(fvm_executor.flush()?),
-        }
+        Ok(delegate_vm!(self.flush()?))
```
LesnyRumcajs, line 277: to my understanding, we can avoid the declarative macro via `#[delegate_impl]`, no?

hanabi1224: Tried `#[delegate_impl]` but it does not compile. According to the doc, `#[delegate_impl]` is used for trait impls.
https://docs.rs/spire_enum/1.0.0/spire_enum/index.html#2-delegate_impl-inherenttrait-impl-attribute-macro

LesnyRumcajs: Hm, indeed. I wonder if `VM` should've been a trait. Something to consider in the future, but outside of the scope of this PR.

src/shim/actors/builtin/account/mod.rs
```diff
@@ -25,19 +27,8 @@ pub enum State {
 }
 
 impl State {
-    pub fn pubkey_address(&self) -> fvm_shared2::address::Address {
-        match self {
-            State::V8(st) => st.address,
-            State::V9(st) => st.address,
-            State::V10(st) => from_address_v3_to_v2(st.address),
-            State::V11(st) => from_address_v3_to_v2(st.address),
```
LesnyRumcajs, line 33: How does it do it now? What's the error message in case of a fiasco?

hanabi1224: Sorry, I don't quite get it. This method has no error propogation.

LesnyRumcajs: It doesn't, but it panics with a certain string.
```rust
pub fn from_address_v3_to_v2(addr: AddressV3) -> AddressV2 {
    AddressV2::from_bytes(&addr.to_bytes())
        .expect("Couldn't convert between FVM3 and FVM2 addresses.")
}
```
Given the conversion is not trivial (one has to call `from_bytes` etc. and check if it worked), my question is what is generated under the hood.

hanabi1224: No conversion happens here, plz note that the `fn` return type has been changed to `shim::Address`. Previously, all addresses are converted to v2, then converted to `shim::Address` [here](https://github.com/ChainSafe/forest/pull/6453/changes#diff-073482dba9a79baee64b56de64d5001ab1c8fd9782629655dba400bdac5d957aL388)

LesnyRumcajs: Ah okay, I missed the signature change. LGTM.

PR #6455, opened 2026-01-21
feat(rpc): implement duplicate key detection for json-rpc requests

### LINKED ISSUE #6424: Strict map parsing in RPC params
# Summary

Forest silently ignores duplicate entries of same key in params payload like:
```json
"params": [ [ { "/":"bafy2bzacea43254b5x6c4l22ynpjfoct5qvabbbk2abcfspfcjkiltivrlyqi", "/":"bafy2bzacea4viqyaozpfk57lnemwufryb76llxzmebxc7it2rnssqz2ljdl6a", "/":"bafy2bzaceav6j67epppz5ib55v5ty26dhkq4jinbsizq2olb3azbzxvfmc73o"  } ]
```
Note two definitions for the same key `/`. Which CID is the one that survives? It's `serde_json` specific. Forest reports no issues. This should have been a parsing or validation error; in theory it's a correct JSON but in practice it'd be nice to inform the user that it's probably not what they want.

For reference, a correct payload that should have been provided instead:
```json
"params": [ [ { "/":"bafy2bzacea43254b5x6c4l22ynpjfoct5qvabbbk2abcfspfcjkiltivrlyqi" }, { "/":"bafy2bzacea4viqyaozpfk57lnemwufryb76llxzmebxc7it2rnssqz2ljdl6a" }, { "/":"bafy2bzaceav6j67epppz5ib55v5ty26dhkq4jinbsizq2olb3azbzxvfmc73o"  } ]
```

One thing to take into consideration is potential performance degradation; this has to be verified. A straightforward implementation would create a hashset when deserializing the map; this is clearly some overhead which might not be worth it (I'd assume most calls have correct input). Perhaps strict validation could live under some environmental variable, either opt-in or opt-out.

# Completion Criteria

- [ ] Implement strict validation of maps in JSON-RPC parsing; perhaps there's something in `serde_json` already.
- [ ] Assert if performance degradation is minimal; if not, introduce an opt-in or opt-out option for these kinds of checks.

Original command with faulty params:
```sh
printf '{ "jsonrpc": "2.0", "id":1, "method": "Filecoin.ChainGetMessagesInTipset", "params": [ [ { "/":"bafy2bzacea43254b5x6c4l22ynpjfoct5qvabbbk2abcfspfcjkiltivrlyqi", "/":"bafy2bzacea4viqyaozpfk57lnemwufryb76llxzmebxc7it2rnssqz2ljdl6a", "/":"bafy2bzaceav6j67epppz5ib55v5ty26dhkq4jinbsizq2olb3azbzxvfmc73o"     } ]  ] }' | curl http://127.0.0.1:1234/rpc/v0 -sH "Content-Type: application/json" -d@/dev/stdin | jq '.result[].Cid."/"'
```

Original thread: https://filecoinproject.slack.com/archives/C029LPZ5N73/p1768385012689069

Dyslex7c:
## Summary of changes

- This PR introduces an opt-in validation to detect duplicate keys in JSON-RPC request payloads and addresses the issue where malformed requests are accepted which may cause unexpected behaviour.

Due to lack of available disk space in my device, I'm unable to run the node (with snapshots downloaded) and perform an E2E test to validate its working. But I have added integration tests and made sure that the core feature is working. Let me know if anything.

Performance constraints are preserved since it's an opt-in feature. There is zero performance overhead when it is disabled and a single-pass O(n) validation when enabled.

Closes #6424 

## Change checklist

## Summary by CodeRabbit

* **New Features**
  * Optional strict JSON validation for RPC requests to detect duplicate object keys. When enabled, invalid calls, notifications, and batch entries return JSON‑RPC errors (failing batch entries are replaced with error results). Toggle via FOREST_STRICT_JSON.

* **Documentation**
  * Environment variables reference updated with FOREST_STRICT_JSON, examples, and default behavior.

* **Chores**
  * Added JSON parsing tooling dependency and updated project dictionary entries to support validation.


src/rpc/json_validator.rs
```diff
@@ -0,0 +1,183 @@
...
+#[cfg(not(test))]
+use std::sync::LazyLock;
+
+pub const STRICT_JSON_ENV: &str = "FOREST_STRICT_JSON";
+
+#[inline]
+pub fn is_strict_mode() -> bool {
+    #[cfg(test)]
+    {
+        std::env::var(STRICT_JSON_ENV)
```
hanabi1224: `crate::utils::misc::env::is_env_truthy(STRICT_JSON_ENV)`

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,183 @@
...
+            .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
+            .unwrap_or(false)
+    }
+    
+    #[cfg(not(test))]
+    {
+        static STRICT_MODE: LazyLock<bool> = LazyLock::new(|| {
+            std::env::var(STRICT_JSON_ENV)
```
hanabi1224: `crate::utils::misc::env::is_env_truthy(STRICT_JSON_ENV)`

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,80 @@
...
+use jsonrpsee::server::middleware::rpc::RpcServiceT;
+use jsonrpsee::types::ErrorObject;
+use jsonrpsee::MethodResponse;
+use tower::Layer;
+
+use super::json_validator;
+
+/// stateless jsonrpsee layer for validating JSON-RPC requests
+#[derive(Clone, Default)]
+pub(super) struct ValidationLayer;
```
hanabi1224: I'd rename it to `JsonValidationLayer`

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        Either::Left(self.service.call(req))
+    }
+
+    fn batch<'a>(
+        &self,
+        batch: jsonrpsee::core::middleware::Batch<'a>,
+    ) -> impl Future<Output = Self::BatchResponse> + Send + 'a {
+        self.service.batch(batch)
```
hanabi1224: We need to enable the check in `batch` as well

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    fn notification<'a>(
+        &self,
+        n: jsonrpsee::core::middleware::Notification<'a>,
+    ) -> impl Future<Output = Self::NotificationResponse> + Send + 'a {
+        self.service.notification(n)
```
hanabi1224: We need to enable the check in `notification` as well

src/lib.rs
```diff
@@ -64,7 +64,7 @@ mod message;
 mod message_pool;
 mod metrics;
 mod networks;
-mod rpc;
+pub mod rpc;
```
hanabi1224: Why change this?

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+/// validates JSON for duplicate keys by parsing at the token level.
+pub fn validate_json_for_duplicates(json_str: &str) -> Result<(), String> {
+    if !is_strict_mode() {
+        return Ok(());
+    }
+
+    let mut object_stack: Vec<HashSet<String>> = Vec::new();
```
hanabi1224: Is it feasible to utilize some json parsing / tokenizing crate? e.g. https://docs.rs/justjson/latest/justjson/struct.Object.html#deref-methods-Vec%3CEntry%3C'a%3E%3E

> Object uses a Vec of Entry types to store its entries.

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,183 @@
...
+//!
+//! serde_json automatically deduplicates keys at parse time using a "last-write-wins" strategy
+//! This means JSON like `{"/":"cid1", "/":"cid2"}` will keep only the last value, which can lead to unexpected behaviour in RPC calls. 
+
+use ahash::HashSet;
+
+#[cfg(not(test))]
+use std::sync::LazyLock;
+
+pub const STRICT_JSON_ENV: &str = "FOREST_STRICT_JSON";
```
hanabi1224, line 12: Please also update `docs/docs/users/reference/env_variables.md`

hanabi1224 (changes requested):
@Dyslex7c thanks for your contribution! Left a few comments.

src/lib.rs
Dyslex7c: So we can import it to our integration tests. But on a second thought we might not need the integration tests as the local tests inside `json_validator.rs` should be sufficient. Let me know and I'll remove it

src/rpc/json_validator.rs
Dyslex7c: yea I guess we can use this crate

src/lib.rs
hanabi1224: Let's revert this until we need it in an integration test.

Dyslex7c:
@hanabi1224 I've addressed all the comments and made changes accordingly. Let me know if anything

src/lib.rs
Dyslex7c: sure I've removed it

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,133 @@
...
+        let service = self.service.clone();
+        
+        async move {
+            if !Self::validation_enabled() {
+                return service.notification(n).await;
+            }
+
+            let params_str = match n.params() {
+                Some(p) => p.as_ref().get(),
+                None => "[]",
```
hanabi1224: Let's avoid validating `[]` in `None` branch

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,133 @@
...
+        async move {
+            if !Self::validation_enabled() {
+                return service.batch(batch).await;
+            }
+
+            for entry in batch.iter_mut() {
+                if let Ok(batch_entry) = entry {
+                    let params_str = match batch_entry.params() {
+                        Some(p) => p.as_ref().get(),
+                        None => "[]",
```
hanabi1224: Let's avoid validating [] in None branch

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,133 @@
...
+                        match batch_entry {
+                            jsonrpsee::core::middleware::BatchEntry::Call(req) => {
+                                let err = ErrorObject::owned(-32600, e, None::<()>);
+                                let err_entry = jsonrpsee::core::middleware::BatchEntryErr::new(
+                                    req.id().clone(),
+                                    err,
+                                );
+                                *entry = Err(err_entry);
+                            }
+                            jsonrpsee::core::middleware::BatchEntry::Notification(_) => {}
```
hanabi1224: `Notification` handling should not be left empty

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,133 @@
...
+                        }
+                    }
+                }
+            }
+
+            service.batch(batch).await
```
hanabi1224, line 108: I don't think it's logically correct to still call `service.batch(batch).await` here

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,121 @@
...
+//! This means JSON like `{"/":"cid1", "/":"cid2"}` will keep only the last value, which can lead to unexpected behaviour in RPC calls. 
+
+use ahash::HashSet;
+use justjson::Value;
+
+pub const STRICT_JSON_ENV: &str = "FOREST_STRICT_JSON";
+
+#[inline]
+pub fn is_strict_mode() -> bool {
+    crate::utils::misc::env::is_env_truthy(STRICT_JSON_ENV)
```
hanabi1224: I'd keep the `LazyLock` to avoid checking the env var repeatedly.

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,121 @@
...
+            Value::Array(arr) => {
+                for item in arr.iter() {
+                    check_value(item)?;
+                }
+                Ok(())
+            }
+            _ => Ok(())
+        }
+    }
+    let value = Value::from_json(json_str)
```
hanabi1224: I'd forgive invalid JSON here and let `serde_json` make the final call later as `justjson` is less widely used than `serde_json` and might be buggy.

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,183 @@
...
+    where
+        F: FnOnce(),
+    {
+        if enabled {
+            unsafe { std::env::set_var(STRICT_JSON_ENV, "1"); }
+        } else {
+            unsafe { std::env::remove_var(STRICT_JSON_ENV); }
+        }
+        f();
+        unsafe { std::env::remove_var(STRICT_JSON_ENV); }
```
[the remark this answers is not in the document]
hanabi1224: This sounds valid @Dyslex7c

src/rpc/validation_layer.rs
Dyslex7c: how else should we pass the batch then?

src/rpc/validation_layer.rs
Dyslex7c: is it okay to use empty string instead?

hanabi1224: Just skip validation in this case

src/rpc/validation_layer.rs
hanabi1224: Sorry, I mistook the usage. Nvm, `service.batch(batch).await` should be called here

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,151 @@
...
+                    check_value(item)?;
+                }
+                Ok(())
+            }
+            _ => Ok(()),
+        }
+    }
+    let value = match Value::from_json(json_str) {
+        Ok(v) => v,
+        Err(_) => return Ok(()),
```
sudo-shashank, line 72: lets add some comment on why we are ignoring other parse issue

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[test]
+    fn test_strict_mode_disabled() {
+        with_strict_mode(false, || { // should pass with strict mode disabled
+            let json = r#"{"/":"cid1", "/":"cid2"}"#;
+            assert!(validate_json_for_duplicates(json).is_ok());
+        });
+    }
+
+    #[test]
+    fn test_original_issue_case() {
```
sudo-shashank: the test name is very vague to me, also you can serialise the test to avoid potential flakiness as we toggling`with_strict_mode` between tests.

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,138 @@
...
+        req: jsonrpsee::types::Request<'a>,
+    ) -> impl Future<Output = Self::MethodResponse> + Send + 'a {
+        if !Self::validation_enabled() {
+            return Either::Left(self.service.call(req));
+        }
+
+        if let Err(e) =
+            json_validator::validate_json_for_duplicates(req.params().as_str().unwrap_or("[]"))
+        {
+            let err = ErrorObject::owned(-32600, e, None::<()>);
```
sudo-shashank: use some constants here for error code 32600

src/rpc/validation_layer.rs
```diff
@@ -0,0 +1,138 @@
+// Copyright 2019-2026 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use futures::future::Either;
+use jsonrpsee::MethodResponse;
+use jsonrpsee::server::middleware::rpc::RpcServiceT;
```
sudo-shashank, line 7:
```suggestion
use jsonrpsee::server::middleware::{BatchEntry, BatchEntryErr, rpc::RpcServiceT};
```

sudo-shashank (changes requested):

src/rpc/json_validator.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+            ':' => {}
+            ',' => {
+                if !object_stack.is_empty() {
+                    expecting_key = true;
+                }
+            }
+            c if c.is_whitespace() => {}
+            _ => {}
+        }
```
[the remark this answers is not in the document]
hanabi1224: Changed JSON parsing approach

sudo-shashank:
@Dyslex7c all commits must have verified signatures, also please check if all review comments are addressed and resolved.

Dyslex7c:
> @Dyslex7c all commits must have verified signatures, also please check if all review comments are addressed and resolved.

All review comments have been addressed and resolved, let me know if anything

sudo-shashank:
@Dyslex7c All comments must be resolved.

Dyslex7c:
> @Dyslex7c All comments must be resolved.

okay done

PR #6886, opened 2026-04-09
feat: tipset prechecks

### LINKED ISSUE #1914: Investigate potential vulnerability in ChainMuxer (database flood)
**Issue summary**

As per @aakoshh thread (thank you!) in #fil-forest-dev Slack, this should be investigated further as part of hardening Forest and we should assess the severity and potential solutions.

> [@adlrocha](https://filecoinproject.slack.com/team/U01AML47QJH)
> another thing that appears like vulnerability to me is that once the node enters into Follow mode in the ChainMuxer it will listen to pubsub messages about blocks, and when it gets them it first downloads all the messages via Bitswap, which as a side effect are saved to the database.While this happens, the node doesn't process any other P2P message.Then the full block is validated without its history, the stateless checks I mentioned. And finally if the tipset is not as heavy as our head, it's dropped. Only then do we resume with the next message.It seems like an adversary could send us blocks that cause huge downloads that never get deleted from the IPLD store, of blocks that aren't even connected to the chain. I don't see any penalty, since it could just be considered an old block.(I was just pondering what is safe to be gossipped at all)

LesnyRumcajs:
## Summary of changes

- adds prechecks (similar to what Lotus does, but more lightweight) before grabbing entire tipsets and putting them into DB. This doesn't 100% mitigate the issue but makes it more difficult to achieve. In our case, it should be _good enough_.

Closes https://github.com/ChainSafe/forest/issues/1914

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **New Features**
  * Pre-fetch validation for gossip blocks to reject duplicates, bad blocks, out-of-range epochs, and other invalid blocks before fetching.
  * New labeled metric to track rejection reasons for gossip blocks.
  * Reduced redundant processing of duplicate/seen blocks to improve network and persistence efficiency and logging.
* **Tests**
  * Added unit tests covering validation paths and edge cases.

### DIFF
--- a/src/chain_sync/bad_block_cache.rs
+++ b/src/chain_sync/bad_block_cache.rs
@@ -47,7 +47,7 @@ impl BadBlockCache {
 
 /// Thread-safe LRU cache for tracking recently seen gossip block CIDs.
 /// Used to de-duplicate gossip blocks before expensive message fetching.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct SeenBlockCache {
     cache: SizeTrackingLruCache<get_size::CidWrapper, ()>,
 }
--- a/src/chain_sync/bad_block_cache.rs
+++ b/src/chain_sync/bad_block_cache.rs
@@ -8,6 +8,10 @@ use nonzero_ext::nonzero;
 
 use crate::utils::{cache::SizeTrackingLruCache, get_size};
 
+/// Default capacity for CID caches (32768 entries).
+/// That's about 4 MiB.
+const DEFAULT_CID_CACHE_CAPACITY: NonZeroUsize = nonzero!(1usize << 15);
+
 /// Thread-safe cache for tracking bad blocks.
 /// This cache is checked before validating a block, to ensure no duplicate
 /// work.
@@ -18,7 +22,7 @@ pub struct BadBlockCache {
 
 impl Default for BadBlockCache {
     fn default() -> Self {
-        Self::new(nonzero!(1usize << 15))
+        Self::new(DEFAULT_CID_CACHE_CAPACITY)
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/chain_sync/validation.rs
```diff
@@ -145,19 +165,196 @@ impl TipsetValidator<'_> {
...
+
+    fn check_duplicate(
+        cid: Cid,
+        seen_block_cache: &SeenBlockCache,
+    ) -> Result<(), GossipBlockRejectReason> {
+        if seen_block_cache.test_and_insert(&cid) {
+            return Err(GossipBlockRejectReason::DuplicateBlock(cid));
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
LesnyRumcajs: Good bot

src/chain_sync/bad_block_cache.rs
```diff
@@ -40,3 +44,30 @@ impl BadBlockCache {
         self.cache.peek_cloned(&(*c).into())
     }
 }
+
+/// Thread-safe LRU cache for tracking recently seen gossip block CIDs.
+/// Used to de-duplicate gossip blocks before expensive message fetching.
+#[derive(Debug)]
```
hanabi1224:
```suggestion
#[derive(Debug, Clone)]
```

src/chain_sync/chain_follower.rs
```diff
@@ -146,6 +147,7 @@ pub async fn chain_follower<DB: Blockstore + Sync + Send + 'static>(
         stateless_mode,
     )));
     let tasks: Arc<Mutex<HashSet<SyncTask>>> = Arc::new(Mutex::new(HashSet::default()));
+    let seen_block_cache = Arc::new(SeenBlockCache::default());
```
hanabi1224: It's ok to clone `SeenBlockCache` without wrapping it with `Arc`

hanabi1224 (dismissed):

src/chain_sync/chain_follower.rs
LesnyRumcajs: good point

hanabi1224 (approved):

PR #6889, opened 2026-04-10
refactor: no Arc wrapper around ChainIndex and ZstdFrameCache

hanabi1224:
## Summary of changes

-

Closes

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **Refactor**
  * Simplified ownership and cloning across indexing, state, and CAR caching to reduce overhead.
  * Introduced a lightweight shallow-clone mechanism so shared data is copied more cheaply.
  * Cache bookkeeping changed to share size counters to reduce memory churn.
  * Concurrency paths updated to use shallow cloning for spawned tasks.
  * No user-facing behavioral changes; functionality and APIs remain effectively unchanged.

### DIFF
--- a/src/db/car/mod.rs
+++ b/src/db/car/mod.rs
@@ -16,12 +16,12 @@ use positioned_io::{ReadAt, Size};
 use std::{
     num::NonZeroUsize,
     sync::{
-        LazyLock,
+        Arc, LazyLock,
         atomic::{AtomicUsize, Ordering},
     },
 };
 
-use crate::utils::{cache::SizeTrackingLruCache, get_size::CidWrapper};
+use crate::utils::{ShallowClone, cache::SizeTrackingLruCache, get_size::CidWrapper};
 
 pub trait RandomAccessFileReader: ReadAt + Size + Send + Sync + 'static {}
 impl<X: ReadAt + Size + Send + Sync + 'static> RandomAccessFileReader for X {}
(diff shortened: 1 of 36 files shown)


src/db/car/mod.rs
```diff
@@ -52,11 +52,12 @@ pub static ZSTD_FRAME_CACHE_DEFAULT_MAX_SIZE: LazyLock<usize> = LazyLock::new(||
     256 * 1024 * 1024
 });
 
+#[derive(Clone)]
```
LesnyRumcajs: This is somehow debatable. A `Clone` subjectively implies an _independent copy_; here it's kind of a weird hybrid.
```
a = cache
b = a.clone
a.current_size += 1
assert a.current_size == b.current_size // shared ownership
assert a == b // essentially same
a.max_size += 1
assert a.max_size == b.current_size // no longer a == b
```
And is `lru` also a deep or shallow clone? Can't say without going into the implementation.

My suggestion is to **not** have such ambiguity and make the code easily understandable.

For example, you could, e.g., rename it to `ZstdFrameCacheInner` (private) and make `ZstdFrameCAche { inner: Arc<ZstdFrameCacheInner> }`. It's immediately clear that it's a shallow copy and a clone just gives a handle. 

Does it make sense?

hanabi1224: How about introducing a ShallowClone trait instead of deriving Clone?

LesnyRumcajs: Yeah, either shallow clone or `Handle`. That was actually my first thought, but I don't think it's an established practice in the Rust ecosystem; I think we'd be more idiomatic with the `inner` approach, which we already have in some places in the code already.

hanabi1224: The problem I see with `inner` approach is that, it's not flexible for hierarchical shallow-copy-able types, e.g. in the below example, `BarInner` could contain a shallow-copy-able field `foo` that is not needed to be wrapped around `Arc`
```
struct Foo(Arc<FooInner>);

struct Bar(Arc<BarInner>);

struct BarInner {
    a: i64,
    foo: Foo,
}
```

LesnyRumcajs: Up to you; at the very least we need an explicit admonition.

hanabi1224: Fixed. There should be more types that could implement `ShallowClone`, let's do that in subsequent PRs

src/utils/shallow_clone.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2019-2026 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::sync::Arc;
+
+pub trait ShallowClone {
+    fn shallow_clone(&self) -> Self;
+}
```
LesnyRumcajs, line 18: Let's nicely document what this trait and its method are supposed to do (and why). It's obvious to us, might not be so obvious for a random contributor.

hanabi1224: Fixed.

LesnyRumcajs (approved):

PR #6983, opened 2026-04-30
Refactor calibnet wallet test

sudo-shashank:
## Summary of changes

- Refactor calibnet basic and delegated wallet bash scripts check to rust integration test.

Closes #6850 

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **Tests**
  * Replaced bash wallet checks with a Rust integration test suite covering export/import, funding, transfers, delegated-wallet flows, address mapping, on-chain polling, and shared test helpers plus preloaded-wallet setup.

* **Chores**
  * Consolidated CI wallet verification into a single workflow job and added a dedicated test task to run the integration suite; removed the previous delegated-wallet bash runner.


tests/calibnet_wallet.rs
```diff
@@ -0,0 +1,204 @@
...
+    .await?;
+    let msg_cid = cid_from_lotus_json_result(&result)?;
+    poll_until_state_search_msg(&msg_cid).await
+}
+
+#[tokio::test]
+#[ignore = "requires a running calibnet Forest daemon"]
+async fn send_to_local_filecoin_address() -> anyhow::Result<()> {
+    let target = wallet(&["new"])?;
+    let _ = send_from(PRELOADED_ADDRESS, &target, FIL_AMT, true)?;
```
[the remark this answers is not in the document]
sudo-shashank: we have `send_to_local_filecoin_address` and `send_to_remote_filecoin_address` also `send_from` can switch between send from remote or local by setting the `remote` param

mise.toml
```diff
@@ -206,6 +206,19 @@ mise task run test:nextest ${usage_profile?}
...
+description = "Run calibnet wallet integration tests."
+shell = "bash -c"
+usage = '''
+arg "<preloaded_key>" help="Hex-encoded preloaded calibnet wallet key (e.g. $CALIBNET_WALLET secret)"
+'''
+run = '''
+set -euxo pipefail
+source ./scripts/tests/harness.sh
+forest_wallet_init "${usage_preloaded_key?}"
+cargo test --profile quick-test --test calibnet_wallet -- --ignored --nocapture
```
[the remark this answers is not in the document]
sudo-shashank: with the revamped nonce calculation logic in forest it does not need to be sequential

tests/calibnet_wallet.rs
```diff
@@ -0,0 +1,191 @@
...
+#[tokio::test]
+#[ignore = "requires a running calibnet Forest daemon"]
+async fn send_to_filecoin_address(#[case] backend: Backend) -> anyhow::Result<()> {
+    let target = wallet(backend, &["new"])?;
+    let msg = send_from(&PRELOADED_ADDRESS, &target, FIL_AMT, backend)?;
+    eprintln!("send to {target} ({}) msg: {msg}", backend.label());
+    let funded = poll_until_funded(&target, backend).await?;
+    eprintln!("{target} funded balance: {funded}");
+    Ok(())
+}
```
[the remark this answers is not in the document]
sudo-shashank: same as above: with the revamped nonce calculation logic in forest it does not need to be sequential

LesnyRumcajs (changes requested):
The test themselves shouldn't use the `?` shorthand; in case of a failure the exact line which failed won't propagate and will make things more difficult to debug. `unwrap` should be used there and the tests shouldn't return `Result`.

tests/common/calibnet_wallet_helpers.rs
```diff
@@ -0,0 +1,304 @@
+// Copyright 2019-2026 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+//! Helpers for the calibnet wallet integration tests in
+//! [`tests/calibnet_wallet.rs`](../../calibnet_wallet.rs).
+
+#![allow(dead_code)]
```
LesnyRumcajs: why?

tests/common/calibnet_wallet_helpers.rs
```diff
@@ -0,0 +1,304 @@
...
+use std::time::Duration;
+
+use anyhow::{Context as _, bail};
+use parking_lot::Mutex;
+use serde_json::{Value, json};
+use tempfile::NamedTempFile;
+
+/// Funded preloaded address from env `PRELOADED_ADDRESS` (see `forest_wallet_init`).
+pub static PRELOADED_ADDRESS: LazyLock<String> = LazyLock::new(|| {
+    std::env::var("PRELOADED_ADDRESS")
```
LesnyRumcajs: What's the convention for environment variables in Forest?

tests/common/calibnet_wallet_helpers.rs
```diff
@@ -0,0 +1,304 @@
...
+use std::process::Command;
+use std::sync::LazyLock;
+use std::time::Duration;
+
+use anyhow::{Context as _, bail};
+use parking_lot::Mutex;
+use serde_json::{Value, json};
+use tempfile::NamedTempFile;
+
+/// Funded preloaded address from env `PRELOADED_ADDRESS` (see `forest_wallet_init`).
```
LesnyRumcajs: Links in the doc comments should point to actual code so that it's reported when a link is broken.

tests/common/calibnet_wallet_helpers.rs
```diff
@@ -0,0 +1,304 @@
...
+pub static PRELOADED_ADDRESS: LazyLock<String> = LazyLock::new(|| {
+    std::env::var("PRELOADED_ADDRESS")
+        .ok()
+        .map(|s| s.trim().to_owned())
+        .filter(|s| !s.is_empty())
+        .expect("PRELOADED_ADDRESS must be set")
+});
+
+pub const FIL_AMT: &str = "500 atto FIL";
```
LesnyRumcajs, line 25: What's this?

tests/calibnet_wallet.rs
```diff
@@ -0,0 +1,194 @@
...
+    };
+    assert!(
+        observed != baseline,
+        "{target} balance unchanged after delegated --remote-wallet send: {observed}",
+    );
+}
+
+/// Delegated signer: create once on local, fund locally, mirror to remote
+/// for tests that query or sign.
+async fn funded_delegated_addr() -> &'static str {
```
LesnyRumcajs: shouldn't this be in the helpers?

mise.toml
```diff
@@ -206,6 +206,19 @@ mise task run test:nextest ${usage_profile?}
...
+description = "Run calibnet wallet integration tests."
+shell = "bash -c"
+usage = '''
+arg "<preloaded_key>" help="Hex-encoded preloaded calibnet wallet key (e.g. $CALIBNET_WALLET secret)"
+'''
+run = '''
+set -euo pipefail
+source ./scripts/tests/harness.sh
+forest_wallet_init "${usage_preloaded_key?}"
+cargo test --profile quick-test --test calibnet_wallet -- --ignored --nocapture
```
LesnyRumcajs: I completely miss your usage of `ignored` in these tests. Please elaborate.

tests/common/calibnet_wallet_helpers.rs
sudo-shashank: not required now, cleaned up

tests/common/calibnet_wallet_helpers.rs
sudo-shashank: Changed to `FOREST_TEST_PRELOADED_ADDRESS`

tests/common/calibnet_wallet_helpers.rs
sudo-shashank: `forest_wallet_init` lives in [scripts/tests/harness.sh](https://github.com/ChainSafe/forest/blob/main/scripts/tests/harness.sh) so we cannot link it directly

tests/common/calibnet_wallet_helpers.rs
sudo-shashank: small transfer amount used in wallet test, added a description

tests/calibnet_wallet.rs
sudo-shashank: yes moved

mise.toml
sudo-shashank: since these test require a running calibnet node, `ignored` is used so that these test doesn't get executed as a part of the default test suite

LesnyRumcajs: It's not a great use case for `ignored`. You should either use a feature flag or an environmental variable if feature flag introduce significantly longer compile times.

mise.toml
```diff
@@ -211,6 +211,19 @@ mise task run test:nextest ${usage_profile?}
...
 
+[tasks."test:wallet"]
+description = "Run calibnet wallet integration tests."
+shell = "bash -c"
+usage = '''
+arg "<preloaded_key>" help="Hex-encoded preloaded calibnet wallet key (e.g. $CALIBNET_WALLET secret)"
+'''
+run = '''
+set -euo pipefail
+source ./scripts/tests/harness.sh
```
LesnyRumcajs, line 222: Does the daemon run still after the test? If so, what if there are two consecutive tests?

tests/common/calibnet_wallet_helpers.rs
```diff
@@ -0,0 +1,334 @@
...
+
+/// Test amount to be transferred between accounts in wallet tests.
+pub const FIL_AMT: &str = "500 atto FIL";
+/// Sentinel `forest-wallet balance --exact-balance` returns for an unfunded address.
+pub const FIL_ZERO: &str = "0 FIL";
+/// Amount to seed a freshly-created delegated wallet.
+pub const DELEGATE_FUND_AMT: &str = "3 micro FIL";
+
+pub const POLL_RETRIES: usize = 20;
+pub const POLL_DELAY: Duration = Duration::from_secs(30);
```
LesnyRumcajs: This is wasteful, making us wait for 29 seconds in worst case scenario where a condition ends up true just after the last retry. For a test that's supposed to be run often, that's long.

We should poll frequently (even sub-second) up to a specified timeout.

mise.toml
sudo-shashank: No it shutdowns after the test

tests/common/calibnet_wallet_helpers.rs
sudo-shashank: Fixed

tests/common/calibnet_wallet_helpers.rs
```diff
@@ -0,0 +1,335 @@
...
+
+/// Test amount to be transferred between accounts in wallet tests.
+pub const FIL_AMT: &str = "500 atto FIL";
+/// Sentinel `forest-wallet balance --exact-balance` returns for an unfunded address.
+pub const FIL_ZERO: &str = "0 FIL";
+/// Amount to seed a freshly-created delegated wallet.
+pub const DELEGATE_FUND_AMT: &str = "3 micro FIL";
+
+/// Maximum time to wait for a polled condition before failing the test.
+pub const POLL_TIMEOUT: Duration = Duration::from_secs(60);
```
LesnyRumcajs: are you sure 1 minute is enough and would not make the job flaky? the original bash script had it set to 10 minutes

tests/calibnet_wallet.rs
```diff
@@ -0,0 +1,157 @@
+// Copyright 2019-2026 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+//! Calibnet wallet integration tests. Each `#[tokio::test]` is `#[ignore]`
```
LesnyRumcajs: leftover

mise.toml
```diff
@@ -211,6 +211,19 @@ mise task run test:nextest ${usage_profile?}
...
+forest_wallet_init "${usage_preloaded_key?}"
+cargo test --profile quick-test --features calibnet-wallet-integration --test calibnet_wallet -- --nocapture
```
LesnyRumcajs, line 224: The tests are run in parallel, no? Wouldn't this introduce some nonce races? It might be safer to run wallet tests on a single thread.

LesnyRumcajs (changes requested):

tests/calibnet_wallet.rs
sudo-shashank: done

tests/common/calibnet_wallet_helpers.rs
sudo-shashank: increased it to 5 mins with 1 sec interval

mise.toml
sudo-shashank: after the nonce calculation fix we should not run into any nonce race conditions

LesnyRumcajs (approved):

PR #7172, opened 2026-06-11
fix: return reverted field of logs set to true if tipset was reverted

### LINKED ISSUE #7096: eth_subscribe `logs`: emit reorg-removed (`removed: true`) logs in subscription
# Summary

Forest's `eth_subscribe("logs")` subscription does not emit reorg-removed logs. On a chain reorg, geth, reth, **and Lotus** all re-emit the logs of reverted tipsets/blocks with `removed: true` so subscribers can roll back state. Forest currently drops reverted tipsets entirely and only emits logs for applied tipsets.

The root cause is that **Forest has no reverted-event tracking**. [EthEventHandler::collect_events](https://github.com/ChainSafe/forest/blob/0e7f413d42b81c12487b984f08314591ca8ce49c/src/rpc/methods/eth/filter/mod.rs#L406) recomputes events from tipset state on every query and hardcodes `reverted: false`. `CollectedEvent.reverted` is therefore always `false`, and `EthLog.removed` (`src/rpc/methods/eth.rs`, `removed: event.reverted`) is consequently always `false`.

# How Lotus does it (reference design)

Lotus maintains `reverted` as a first-class property, two converging paths both feeding `CollectedEvent.Reverted` → `EthLog.Removed` (`Removed: ev.Reverted`, `node/impl/eth/events.go`):

- **Persisted index (ChainIndexer / SQLite).**
- **Live subscription push.** An in-memory `eventFilter` registered as a head-change observer builds `CollectedEvent{ Reverted: revert }` from whether the head-change was an apply (`false`) or revert (`true`)

# Completion Criteria
- [ ] `eth_subscribe("logs")` re-emits the logs of reverted tipsets with `removed: true` on reorg (reverts emitted before applies).
- [ ] Test coverage for the reorg removed-log path.

akaladarshi:
## Summary of changes

- Read the head change (apply and reverted) and fetch the logs live for both
- Refactor the existing logs matching

Closes https://github.com/ChainSafe/forest/issues/7096

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `eth_subscribe` log delivery during reorgs so reorg-reverted logs are re-emitted with `removed: true`, before logs from the replacing chain head.
  * Improved installed log filters (`eth_getFilterLogs` / `eth_getFilterChanges`) to return only newly observed events by tracking per-filter event delivery positions.
* **Tests**
  * Expanded end-to-end `eth_subscribe` coverage into a filter-matrix validating address/topic wildcard behavior (including trailing wildcards) and correct inclusion/exclusion across concurrent subscriptions.

### DIFF
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -35,7 +35,10 @@ use crate::rpc::{
     error::ServerError,
     eth::{
         errors::EthErrors,
-        filter::{SkipEvent, event::EventFilter, mempool::MempoolFilter, tipset::TipSetFilter},
+        filter::{
+            EventRevertStatus, SkipEvent, event::EventFilter, mempool::MempoolFilter,
+            tipset::TipSetFilter,
+        },
         utils::decode_revert_reason,
     },
     methods::chain::ChainGetTipSetV2,
@@ -1454,17 +1457,30 @@ pub async fn eth_logs_for_block_and_transaction(
     eth_filter_logs_from_events(ctx, &events)
 }
 
-pub async fn eth_logs_with_filter(
+/// Collects the Ethereum logs of the message tipset whose receipts live in `receipt_ts`
+/// (its child), marking them as removed when the head change that surfaced them was a reorg
+/// revert. The receipt tipset is passed explicitly because reverted tipsets can no longer be
+/// resolved through the canonical chain.
+pub(in crate::rpc) async fn eth_logs_for_head_change(
     ctx: &Ctx,
-    ts: &Tipset,
-    spec: Option<EthFilterSpec>,
+    receipt_ts: &Tipset,
+    revert_status: EventRevertStatus,
 ) -> anyhow::Result<Vec<EthLog>> {
+    let msg_ts = ctx
+        .chain_index()
+        .load_required_tipset(receipt_ts.parents())?;
+    let executed_ts = ctx
+        .state_manager
+        .load_executed_tipset_with_receipt(&msg_ts, receipt_ts)
+        .await?;
     let mut events = vec![];
-    EthEventHandler::collect_events(
+    EthEventHandler::collect_events_from_messages(
         &ctx.state_manager,
-        ts,
-        spec.as_ref(),
+        &msg_ts,
+        &executed_ts.executed_messages,
+        None::<&ParsedFilter>,
         SkipEvent::OnUnresolvedAddress,
+        revert_status,
         &mut events,
     )
     .await?;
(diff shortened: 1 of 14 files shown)


src/rpc/methods/eth.rs
```diff
@@ -3289,6 +3315,54 @@ impl RpcMethod<1> for EthGetLogs {
     }
 }
 
+/// Shared implementation of `eth_getFilterLogs` / `eth_getFilterChanges` for installed event
+/// filters: collects the filter's full result set from the canonical chain, returns only the
+/// events that were not present in the previous poll.
```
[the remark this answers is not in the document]
akaladarshi: `poll_event_filter` is supposed to return all the logs that it has seen since the last call to the `poll_event_filter`  `eth_getFilterChanges` returns the changes since the previous call to the `eth_getFilterChanges`.

`seen_positions` is there to track what logs were seen in the last call. Please verify once again and then report.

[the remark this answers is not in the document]
akaladarshi: What you're quoting is what ethereum does but the lotus does something else it calls the `e.filterStore.Get(ctx, types.FilterID(id))` to get the filter using filter id for both (`eth_getFilterLogs`, `eth_getFilterChanges`) and drains the `collectedEvents` from the `EventFilter`.

Can you please take a look at it once again and verify what lotus does , since we need to make parity with the lotus not ethereum.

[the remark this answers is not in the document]
akaladarshi: Create a new issue in forest for this so i can check if this is a bug in lotus or not and we can track if this is the intended behaviour or not.

src/state_manager/state_computation.rs
```diff
@@ -47,7 +45,43 @@ impl StateManager {
             StateRecomputePolicy::Disallowed
         };
 
-        self.load_executed_tipset_with_cache(ts, policy).await
+        // https://github.com/ChainSafe/forest/issues/7118
```
LesnyRumcajs: that issue is closed, no?

akaladarshi: Removed, It was left over from the rebase conflict.

src/tool/subcommands/api_cmd/stateful_tests.rs
```diff
@@ -567,67 +568,298 @@ fn eth_subscribe_pending_transactions(tx: TestTransaction) -> RpcTestScenario {
...
     RpcTestScenario::basic(move |client| {
         let tx = tx.clone();
         async move {
-            let filter = json!({
-                "address": [],
-                "topics": [tx.topic.to_string()],
-            });
-            let (mut ws_stream, subscription_id) =
-                open_eth_subscription(&client, SubscriptionKind::Logs, Some(filter)).await?;
+            let contract = EthAddress::from_filecoin_address(&tx.to)?;
```
LesnyRumcajs, line 662: this test is huge and it's unclear to me what it does. Tests must be easy to follow, either with clear code, e.g., by extracting certain chunks to separate functions or at least with meaningful, commented blocks. I also see there are tons of magical timeouts, each with different value. Why? How long is this method supposed to run, realistically?

akaladarshi: It basically test the response of `EthSubscription` API for `Logs` under the different kind of filters.

`filter_all`, `filter_wrong`, `filter_topic`, `filter_wild`, `filter_constrained`.

But yeah test got too complicated, will refactor it.

LesnyRumcajs (approved):

PR #7283, opened 2026-07-02
feat(api-compare): report tested API versions in conformance report

### LINKED ISSUE #6509: Monthly conformance reports should report API version
# Summary

Right now conformance reports contain only the API version; same goes for the output for `api compare`. The report should contain all version categories, i.e., `v0`, `v1`, `v2`. Alternatively, to avoid bloat, there could be an additional column listing tested API versions.

# Completion Criteria

- [ ] Conformance report contain API versions tested

0xDevNinja:
## Summary of changes

The conformance/`api compare` report gave no indication of which API version (`v0`/`v1`/`v2`) a method was tested under. This adds that information.

- Track the API path of every executed test per method in `ReportBuilder`.
- Expose it as a `tested_api_versions` field in the JSON report (omitted when empty, so existing consumers stay compatible).
- Add an "API Versions" column to the printed summary table.

I went with both the JSON field and the summary column to cover the two options mentioned in the issue.

Closes #6509

The version set is a `BTreeSet<ApiPaths>` so the output is deterministic (`v0, v1, v2`) regardless of test execution order. Added a unit test for the formatting.

I didn't add a CHANGELOG entry since `forest-tool api compare` is a developer/CI conformance tool rather than a node-operator-facing feature, but happy to add one if you'd prefer.

## Change checklist

- [x] I have made corresponding changes to the documentation. All new code adheres to the team's documentation standards,
- [x] I have made sure the CHANGELOG is up-to-date. All user-facing changes should be reflected in this document.

### Outside contributions

- [x] I have read and agree to the CONTRIBUTING document.
- [x] I have read and agree to the AI Policy document. I understand that failure to comply with the guidelines will lead to rejection of the pull request.

## Summary by CodeRabbit

* **New Features**
  * API test reports now show which API versions were used for each RPC method.
  * The summary output includes a new “API Versions” column for easier review.
  * Saved report data now preserves API version usage per method for later inspection.

### DIFF
--- a/src/tool/subcommands/api_cmd/report.rs
+++ b/src/tool/subcommands/api_cmd/report.rs
@@ -3,13 +3,14 @@
 
 use super::ReportMode;
 use crate::prelude::*;
-use crate::rpc::{self, FilterList, Permission};
+use crate::rpc::{self, ApiPaths, FilterList, Permission};
 use crate::tool::subcommands::api_cmd::api_compare_tests::TestSummary;
 use ahash::{HashMap, HashSet};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_with::{DisplayFromStr, DurationMilliSeconds, DurationSeconds, serde_as};
 use similar::{ChangeTag, TextDiff};
+use std::collections::BTreeSet;
 use std::path::Path;
 use std::time::{Duration, Instant};
 use tabled::{builder::Builder, settings::Style};
@@ -123,6 +124,10 @@ struct MethodReport {
     /// Current testing status
     status: MethodTestStatus,
 
+    /// API versions this method was exercised under
+    #[serde(skip_serializing_if = "BTreeSet::is_empty")]
+    api_versions: BTreeSet<ApiPaths>,
+
     // Performance metrics (always included)
     #[serde(skip_serializing_if = "Option::is_none")]
     performance: Option<PerformanceMetrics>,
@@ -177,6 +182,7 @@ impl ReportBuilder {
                     } else {
                         MethodTestStatus::NotTested
                     },
+                    api_versions: BTreeSet::new(),
                     performance: None,
                     success_test_params: vec![],
                     failed_test_params: vec![],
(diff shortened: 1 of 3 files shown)


src/tool/subcommands/api_cmd/report.rs
```diff
@@ -123,6 +124,10 @@ struct MethodReport {
     /// Current testing status
     status: MethodTestStatus,
 
+    /// API versions (`v0`/`v1`/`v2`) this method was exercised under
+    #[serde(skip_serializing_if = "BTreeSet::is_empty")]
+    tested_api_versions: BTreeSet<ApiPaths>,
```
akaladarshi: The name can just be `api_versions`

src/tool/subcommands/api_cmd/report.rs
```diff
@@ -412,6 +422,24 @@ impl ReportBuilder {
     }
 }
 
+/// Human-readable label for an API version, e.g. `v0`/`v1`/`v2`.
+fn api_path_label(api_path: ApiPaths) -> &'static str {
```
akaladarshi: We don't need any helper as `ApiPaths` can just use the `strum::Display`.

src/tool/subcommands/api_cmd/report.rs
```diff
@@ -339,6 +348,7 @@ impl ReportBuilder {
                         method_name.as_str(),
                         &format!("{success_count}/{total_count}"),
                         &format!("{success_count}/{total_count}"),
+                        &format_api_versions(&report.tested_api_versions),
```
akaladarshi: Once we have the `strum::Display` we can just use the `report.tested_api_versions.iter().join(", ")`

akaladarshi (changes requested):

0xDevNinja:
Thanks! Addressed all three:

- Renamed the field to `api_versions`.
- Derived `strum::Display` on `ApiPaths` and dropped the helper.
- The summary column now just does `api_versions.iter().join(", ")`.

akaladarshi (approved):
Thanks, LGTM.

PR #7347, opened 2026-07-13
test(lotus_json): assert LotusJson<Option<T>> is nullable in schema

### LINKED ISSUE #4331: Ensure `LotusJson<Option<T>>` is marked as `optional` in the OpenRPC spec
Does this fix an encountered issue? Perhaps a simple test is in order to verify that `LotusJson<Option<T>>` encodes/decodes the way we expect. It's a couple of lines of code, but goes a long way, especially if this code is ever touched by somebody else. 

I leave this up to you, non-blocking.

_Originally posted by @ruseinov in https://github.com/ChainSafe/forest/issues/4330#issuecomment-2113427560_

0xDevNinja:
## Summary

Closes #4331.

`LotusJson<Option<T>>` delegates its schema to `Option<T::LotusJson>`, which `schemars` already renders as nullable (`"type": [..., "null"]` / a `null` branch in `anyOf`). This adds a small regression test locking that behaviour in, as suggested in the issue — verifying that:

- `LotusJson<Option<T>>` produces a schema that permits `null` (i.e. is optional in the OpenRPC spec), and
- the non-optional `LotusJson<T>` counterpart does **not** permit `null`.

The helper `schema_allows_null` accepts both the `"type"` union form (used for primitives/arrays) and the `anyOf`/`oneOf` form (used for `$ref` object types), so the test stays robust across inner types.

No production code changes — test only.

## Reference tests

`cargo test -p forest-filecoin --lib lotus_json::opt::` — 3 passed (existing snapshot + quickcheck, plus the new `optional_is_nullable_in_schema`).

Issue originally raised by @ruseinov in https://github.com/ChainSafe/forest/issues/4330#issuecomment-2113427560.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage to ensure optional Lotus JSON values accept `null` in generated JSON/OpenRPC schemas.
  * Verified non-optional Lotus JSON values do not permit `null` in the corresponding schemas.

### DIFF
--- a/src/lotus_json/opt.rs
+++ b/src/lotus_json/opt.rs
@@ -37,3 +37,41 @@ fn shapshots() {
 fn quickcheck(val: Option<::cid::Cid>) {
     assert_unchanged_via_json(val)
 }
+
+/// Regression test for <https://github.com/ChainSafe/forest/issues/4331>.
+///
+/// `LotusJson<Option<T>>` must be rendered as an optional value that permits
+/// `null` in the generated JSON/OpenRPC schema, so consumers know the field
+/// may be absent. The non-optional `LotusJson<T>` counterpart must not permit
+/// `null`.
+#[test]
+fn optional_is_nullable_in_schema() {
+    let optional = schemars::schema_for!(LotusJson<Option<::cid::Cid>>);
+    assert!(
+        schema_allows_null(optional.as_value()),
+        "LotusJson<Option<T>> schema must permit null: {optional:?}"
+    );
+
+    let required = schemars::schema_for!(LotusJson<::cid::Cid>);
+    assert!(
+        !schema_allows_null(required.as_value()),
+        "LotusJson<T> schema must not permit null: {required:?}"
+    );
+}
+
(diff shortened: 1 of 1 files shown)


hanabi1224:
LGTM, thanks! There is an outstanding lint issue to fix tho.

sudo-shashank (changes requested):
`lint:spellcheck` failing

0xDevNinja:
Good catch, thanks. Fixed the spellcheck failure by rewording the doc comment to avoid the word that tripped the dictionary. Should be green now.

sudo-shashank (changes requested):
@0xDevNinja `lint:clippy` check is failing

0xDevNinja:
Thanks, fixed. The `collapsible_match` lint was on my helper in `opt.rs`, so I reworked it to compute the result without the nested if. Verified locally with clippy on the lib and test targets plus the unit tests, all green now.

sudo-shashank (approved):
LGTM

PR #7382, opened 2026-07-17
chore: update the chain follower logs to report validation

### LINKED ISSUE #5499: The ChainFollower should report when tipset validation will start
# Summary

When catching up to HEAD, Forest will print out the max number of tipsets there are left to download. This number is not all that useful, though, since there will always be a few forks with lots of outstanding tipsets. It would be better to report the minimum number of tipsets to be downloaded before validation can proceed. This number is independent of the number of forks.

# Requirements

# Motivation

akaladarshi:
## Summary of changes

- Instead of the `max` forks epoch remaining to start validation, now log shows `min` epochs remaining to start validation
- Update the  chain follower logs to report the remaining tipset to start validation instead of remaining downloads

Closes https://github.com/ChainSafe/forest/issues/5499

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chain synchronization progress reporting while catching up to the latest chain head by showing the heaviest epoch, expected head, and the computed difference.
  * Progress messages now include clearer stage-based status (e.g., validating tipsets, remaining header fetches, or waiting for tipsets).
  * Log updates are emitted only when the node is significantly behind, reducing unnecessary log noise during normal operation.

### DIFF
--- a/src/chain_sync/chain_follower.rs
+++ b/src/chain_sync/chain_follower.rs
@@ -401,9 +401,17 @@ async fn chain_follower(
                 }
 
                 let (_, forks) = state_machine.lock().tasks();
+                let is_validating = forks
+                    .iter()
+                    .any(|fork| fork.stage == ForkSyncStage::ValidatingTipsets);
+                let is_fetching = forks
+                    .iter()
+                    .any(|fork| fork.stage == ForkSyncStage::FetchingHeaders);
                 // The fork closest to the validated head connects first, so its
-                // gap is how many epochs remain before tipset validation can start.
-                let until_validation = forks
+                // gap is roughly how many header fetches remain before tipset
+                // validation can start. Forks fetching at or below the validated
+                // head (fork-point search) have no meaningful gap.
+                let remaining_headers = forks
                     .iter()
                     .filter(|fork| fork.stage == ForkSyncStage::FetchingHeaders)
                     .filter_map(|fork| {
@@ -411,16 +419,20 @@ async fn chain_follower(
                         (gap > 0).then_some(gap)
                     })
                     .min();
-                let is_validating = forks
-                    .iter()
-                    .any(|fork| fork.stage == ForkSyncStage::ValidatingTipsets);
 
                 let status = if is_validating {
                     ", validating tipsets".to_string()
-                } else if let Some(gap) = until_validation {
-                    format!(", validation starts in ~{gap} epochs")
+                } else if let Some(remaining) = remaining_headers {
+                    format!(", fetching headers (~{remaining})")
+                } else if is_fetching {
+                    // All live fetches are at or below the validated head:
+                    // fork-point searches, fetching a competing branch's
+                    // ancestry down to where it forked off our chain.
+                    ", fetching headers".to_string()
                 } else {
-                    String::new()
+                    // waiting for peers to (re)seed tipsets (startup, no usable peers, or the buffer
+                    // was just cleared after a failed validation).
+                    ", waiting for tipsets".to_string()
                 };
                 info!(
                     "Catching up to HEAD: {heaviest_epoch}{} -> {expected_head} (diff: {diff}){status}",
--- a/src/chain_sync/chain_follower.rs
+++ b/src/chain_sync/chain_follower.rs
@@ -376,52 +376,56 @@ async fn chain_follower(
         }
     });
 
-    // Periodically report progress if there are any tipsets left to be fetched.
-    // Once we're in steady-state (i.e. caught up to HEAD) and there are no
-    // active forks, this will not report anything.
+    // Periodically report progress while catching up to HEAD. Once we're in
+    // steady-state (i.e. caught up to HEAD), and there are
+    // no active forks, this will not report anything.
     set.spawn({
         let state_manager = state_manager.shallow_clone();
         let state_machine = state_machine.clone();
         async move {
             loop {
                 tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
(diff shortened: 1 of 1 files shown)


sudo-shashank:
No green checkmark

sudo-shashank (dismissed):

src/chain_sync/chain_follower.rs
```diff
@@ -376,52 +376,56 @@ async fn chain_follower(
...
+                    })
+                    .min();
+                let is_validating = forks
+                    .iter()
+                    .any(|fork| fork.stage == ForkSyncStage::ValidatingTipsets);
+
+                let status = if is_validating {
+                    ", validating tipsets".to_string()
+                } else if let Some(gap) = until_validation {
+                    format!(", validation starts in ~{gap} epochs")
```
LesnyRumcajs: I don't think this communicates what the node is doing correctly. If I'd be reading this as a user and see "validation starts in ~10 epochs" I'd expect it to start in 5 minutes. It's more of a "fetching X headers", no?

src/chain_sync/chain_follower.rs
```diff
@@ -376,52 +376,56 @@ async fn chain_follower(
...
+                } else {
+                    String::new()
```
LesnyRumcajs: So, what is this situation?

src/chain_sync/chain_follower.rs
akaladarshi: I guess you're right, wording can be more correct here.

Do you think this makes more sense?
`Catching up to HEAD: X[cid] -> Y (diff: 6775), fetching headers, validation starts in ~299 epochs`

src/chain_sync/chain_follower.rs
akaladarshi: ~~String::new() is not going to be printed because we have this: `if diff <= 10 {
                    continue;
                }`~~

akaladarshi: I guess we should not be printing anything in this case, since catching up head is only supposed to be for `Fetching Header` or `Validating Tipset`

src/chain_sync/chain_follower.rs
LesnyRumcajs: no, `validation starts in...` in my view implies clock time duration. Maybe `fetching headers (number)` would be the simplest option

src/chain_sync/chain_follower.rs
LesnyRumcajs: Yeah, but what is this state of Forest? It's not validating and there's no gap yet it's not synced?

akaladarshi: This state is basically about node is not getting any tipset, waiting for peer to send tipset.
- It can be either because validation failed and we have marked exisiting tipsets as bad
- Or we can get this, simply because the peer are not sending us the tipset, or there is some delay in receiving more tipset.

 So this should also be another status = `waiting for tipset`

src/chain_sync/chain_follower.rs
```diff
@@ -376,52 +376,68 @@ async fn chain_follower(
...
+                let remaining_headers = forks
+                    .iter()
+                    .filter(|fork| fork.stage == ForkSyncStage::FetchingHeaders)
+                    .filter_map(|fork| {
+                        let gap = fork.target_sync_epoch_start - fork.validated_chain_head_epoch;
+                        (gap > 0).then_some(gap)
+                    })
+                    .min();
+
+                let status = if is_validating {
```
LesnyRumcajs: It's obviously not a big deal because these logs aren't emitted very often, but you might be interested in using a `Cow` to avoid allocations for static strings.

src/chain_sync/chain_follower.rs
```diff
@@ -376,52 +376,68 @@ async fn chain_follower(
...
+                    ", validating tipsets".to_string()
+                } else if let Some(remaining) = remaining_headers {
+                    format!(", fetching headers (~{remaining})")
```
LesnyRumcajs: why `~`? we know the exact value, no?

src/chain_sync/chain_follower.rs
```diff
@@ -376,52 +376,68 @@ async fn chain_follower(
...
+                    continue;
                 }
+
+                let (_, forks) = state_machine.lock().tasks();
+                let is_validating = forks
+                    .iter()
+                    .any(|fork| fork.stage == ForkSyncStage::ValidatingTipsets);
+                let is_fetching = forks
+                    .iter()
+                    .any(|fork| fork.stage == ForkSyncStage::FetchingHeaders);
```
LesnyRumcajs: This seems kind of off or wasteful. Please revisit the logic. Imagine you have two forks, one doing validations, and another fetching headers. You kind of authoritatively decided only display only the first one (I'm not saying it's bad, it just need to be conscious) if it's present and skip the second one but you still do all the computation (however cheap) for the latter.

src/chain_sync/chain_follower.rs
akaladarshi: There might be null rounds in between and also the forks as well, which could make the value shrink. At least that was the thought behind it.

src/chain_sync/chain_follower.rs
akaladarshi: Done.

src/chain_sync/chain_follower.rs
akaladarshi: Refactored it. Now we are using the `sync_status` to avoid loading the state machine as well. 

Also left lot comments there because after certain point of time it get difficult to remember the reasoning behind the logs.

LesnyRumcajs (approved):
LGTM

PR #7419, opened 2026-07-27
feat(eth): allow eth_call and eth_estimateGas from contract and non-existent senders

### LINKED ISSUE #7394: Allow `eth_call` and `eth_estimateGas` from contract and non-existent senders
# Summary

Basically port https://github.com/filecoin-project/lotus/pull/13724

Make sure to review that PR carefully and implement the changes into idiomatic Rust (🌲 edition) code, not just translate to Golang, especially potentially slow bits (we have a couple of exclusive tools in our arsenal)

# Completion Criteria

- [ ] Port the implementation of https://github.com/filecoin-project/lotus/pull/13724 to Forest (including tests)
- [ ] Benchmark the changes against Lotus - they **must** be at least as performant out of the box.

0xDevNinja:
## Summary

Ports filecoin-project/lotus#13724 to Forest. Closes #7394.

Ethereum tooling assumes `eth_call` and `eth_estimateGas` work from senders that Filecoin's explicit path rejects: Safe's protocol-kit issues `eth_call` with `from == to` for EIP-1271 checks, dApps estimate gas before an account is funded, and viem defaults `from` to the connected wallet even for reads. Forest ran both methods on the FVM's explicit path, which requires the sender to be an existing account-type actor, so all of the above failed.

This mirrors Geth (and Lotus) by skipping sender validation for these two methods only:

- **Sender does not exist on chain:** apply an implicit zero-value send to the address so the FVM instantiates an ephemeral placeholder actor, then run the message on the explicit path so gas accounting matches a real first send.
- **Sender exists but is an EVM contract:** run the message on the implicit path, which skips the account-type / nonce / balance checks.

Everything runs on the VM's in-memory buffered state; nothing is persisted.

## Design notes (Forest specifics)

- Forest routes both `eth_call` and `eth_estimateGas` through `StateManager::call_with_gas` (explicit apply), so both cases are gated there behind a new `skip_sender_validation` flag. The strict public API is unchanged (thin wrappers), and `Filecoin.GasEstimateGasLimit` / the other Filecoin RPCs keep strict validation — the skip path is scoped to the two eth methods.
- `eth_call` tries the strict path first and retries on the skip path when the receipt is `SysErrSenderInvalid` (contract sender) or the call returns the new typed `Error::SenderValidationFailed` (missing sender).
- `eth_estimateGas` detects a contract or missing sender up front (via the sender actor's code) and estimates on the skip path: initial estimate, overestimation multiplier, then the usual gas search.

## Known deficiency (inherited from Lotus)

As documented in the upstream PR, estimates from *contract* senders miss the on-chain inclusion cost on the implicit path; the 1.25x overestimation multiplier covers it for realistic calls. The proper fix needs a new FVM `ApplyKind` and a ref-fvm/ffi release.

## Tests

Added RPC conformance cases (`eth_call` / `eth_estimateGas` from a non-existent sender, V1 + V2). Rejections are tolerated identically across nodes so the cases stay stable against a Lotus that predates the equivalent change.

## Validation status

- `cargo check` / `clippy --all-targets -D warnings` (modulo the pre-existing generated `_go_bindings.rs` cast lints) / `cargo fmt` all clean.
- The completion criteria call for a gas benchmark against Lotus. That needs the RPC conformance harness against a Lotus build carrying lotus#13724, which is why this is a **draft** — I would like the conformance CI to run before I call the parity guaranteed. Happy to add contract-sender fixtures or adjust the approach based on what you would prefer to see benchmarked.

### DIFF
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -1875,6 +1875,15 @@ async fn eth_estimate_gas(
     // gas estimation actually run.
     msg.gas_limit = 0;
 
+    // Geth allows eth_estimateGas from senders that Filecoin's explicit path rejects. An
+    // EVM-contract sender fails the FVM's account-type check, and a sender that does not
+    // exist on chain has no nonce to read. Detect both up front and estimate on the
+    // skip-sender-validation path, which applies the message implicitly (contract sender)
+    // or creates an ephemeral placeholder first (non-existent sender).
+    if sender_needs_skip_validation(ctx, msg.from, &tipset).await {
+        return eth_estimate_gas_skip_sender(ctx, msg, tipset).await;
+    }
+
     match gas::estimate_message_gas(ctx, msg.clone(), None, tipset.key().clone().into()).await {
         Err(server_err) => {
             // On failure, GasEstimateMessageGas doesn't actually return the invocation result,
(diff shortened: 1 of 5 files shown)


sudo-shashank:
@0xDevNinja, Thanks for your interest! 
This issue hasn't been assigned yet. Please wait for a maintainer to assign it or give you the go-ahead before you start writing code.

For more details, please see the contribution guidelines: https://github.com/ChainSafe/forest/blob/main/CONTRIBUTING.md#i-want-to-contribute

PR #7435, opened 2026-08-03
fix(eth): allow eth_call and eth_estimateGas from contract and non-existent senders

### LINKED ISSUE #7394: Allow `eth_call` and `eth_estimateGas` from contract and non-existent senders

sudo-shashank:
## Summary of changes

- Allow `eth_call` and `eth_estimateGas` from contract and non-existent addresses via a new skip-sender-validation path, matching Lotus/Geth including tests.
- Port changes from https://github.com/filecoin-project/lotus/pull/13724

Closes #7394 

## Existing-sender: Forest main (baseline) vs forest-384e029ff
No regression.
### eth_call
| | Forest main (baseline) | forest-384e029ff | delta |
|---|---:|---:|---|
| P50 | 0.706 ms | 0.679 ms | −3.8% |
| P99 | 1.015 ms | 0.956 ms | −5.8% |
| Avg | 0.714 ms | 0.684 ms | −4.2% |
### eth_estimateGas
| | Forest main (baseline) | forest-384e029ff | delta |
|---|---:|---:|---|
| P50 | 1.383 ms | 1.354 ms | −2.2% |
| P99 | 1.832 ms | 1.716 ms | −6.3% |
| Avg | 1.388 ms | 1.358 ms | −2.1% |
## forest-384e029ff: existing-sender vs missing-sender
Missing-sender is slower by ~0.4 ms at P50.
### eth_call
| | Existing-sender | Missing-sender | delta |
|---|---:|---:|---|
| P50 | 0.679 ms | 1.093 ms | +60.9% (+0.414 ms) |
| P99 | 0.956 ms | 1.443 ms | +51.0% |
| Avg | 0.684 ms | 1.091 ms | +59.5% |
### eth_estimateGas
| | Existing-sender | Missing-sender | delta |
|---|---:|---:|---|
| P50 | 1.354 ms | 1.733 ms | +28.0% (+0.379 ms) |
| P99 | 1.716 ms | 2.207 ms | +28.6% |
| Avg | 1.358 ms | 1.740 ms | +28.1% |

## Change checklist

### Outside contributions

## Bug Fixes
* Improved `eth_call` and `eth_estimateGas` behavior across sender types, validation modes, contract creation, and insufficient-funds scenarios.
* Added consistent Ethereum out-of-gas and execution-revert errors with preserved revert details.
* Improved gas estimation and sender simulation handling.
* Fixed trace results for self-destructed contracts.

## Tests
* Expanded API parity coverage for sender validation, gas estimation, and reverting initialization.

## Documentation
* Updated the unreleased changelog with RPC compatibility fixes.


src/tool/subcommands/api_cmd/test_snapshots.txt
LesnyRumcajs, line 1: 1. What's the grand total size of all new snapshots?
2. What are those snapshots doing? Is each one of them testing something distinct or is there some overlap?

sudo-shashank: 1. Each file size is ~5 Mb so the total size of all the new snapshots is around 120 Mb.
2. Yes, distinct test cases but the same set is executed for both `v1` and `v2` api paths. I'll add some tags to each file for better clarity.

LesnyRumcajs: are v1/v2 paths any different in practice for these methods?

sudo-shashank: No, i'll keep one set and remove other

src/tool/subcommands/api_cmd/api_compare_tests.rs
```diff
@@ -1643,6 +1644,120 @@ fn eth_call_api_err_tests(epoch: ChainEpoch) -> Vec<RpcTest> {
     tests
 }
 
+const SKIP_SENDER_CONTRACT: &str = "0x0c1d86d34e469770339b53613f3a2343accd62cb";
+const SKIP_SENDER_CALLDATA: &str =
+    "0xf8b2cb4f000000000000000000000000CbfF24DED1CE6B53712078759233Ac8f91ea71B6";
```
LesnyRumcajs: Those needs some commenting.

src/tool/subcommands/api_cmd/api_compare_tests.rs
```diff
@@ -1643,6 +1644,120 @@ fn eth_call_api_err_tests(epoch: ChainEpoch) -> Vec<RpcTest> {
...
+
+fn eth_skip_sender_tests(epoch: ChainEpoch) -> anyhow::Result<Vec<RpcTest>> {
+    let mut tests = eth_skip_sender_success_tests(epoch)?;
+    tests.extend(eth_skip_sender_insufficient_funds_tests(epoch)?);
+    tests.extend(eth_skip_sender_create_reject_tests(epoch)?);
+    tests.extend(eth_skip_sender_block_param_tests(epoch)?);
+    Ok(tests)
+}
+
+fn eth_skip_sender_cases(
```
LesnyRumcajs: It's not exactly clear what it does. Please add some comments.

src/tool/subcommands/api_cmd/api_compare_tests.rs
```diff
@@ -1643,6 +1644,120 @@ fn eth_call_api_err_tests(epoch: ChainEpoch) -> Vec<RpcTest> {
...
+    epoch: ChainEpoch,
+    policy: PolicyOnRejected,
+    messages: impl IntoIterator<Item = EthCallMessage>,
+) -> anyhow::Result<Vec<RpcTest>> {
+    let mut tests = Vec::new();
```
LesnyRumcajs: nit: not a huge deal given perf is not crucial on this path, but you could avoid reallocating - you already know the capacity the vec should have.
```suggestion
    let mut tests = Vec::with_capacity(2 * messages.len());
```

LesnyRumcajs:
@sudo-shashank Did you run your changes against local CC review? I ran it on this PR and it surfaced some potential issues.

```
Correctness:
1. eth.rs:2064 — out-of-gas trace check is dead. Some(_expected_exit_code) is a fresh pattern binding, so it matches any exit code, not SYS_OUT_OF_GAS. A genuinely-reverting eth_estimateGas can be pushed into gas_search and return a bogus near-block-limit estimate instead of the revert error. This is the most important one — and it directly qualifies the parity agent's "point 5 correct" verdict.
2. message_simulation.rs:195 — skip-retry never fires for ID-mapped senders. The Enforce branch propagates an untyped resolve error, so eth_call from a non-existent 0xff..<id> sender errors out where Lotus (which checks GetActor first) succeeds. The sibling gas.rs path types it correctly, so the two are inconsistent.
3. gas.rs:281 — lost diagnostics. Direct Filecoin.GasEstimate* callers now get a bare "sender validation failed" instead of the exit-code + reason main surfaced.

Plausible:
4. gas.rs:225 — map_err(|_| …) swallows the real error. A transient blockstore/state read failure gets misclassified as a sender-validation problem and silently falls into the skip-sender path with a fabricated placeholder.

Cleanups: a redundant tipset+actor load on the hot EOA path (eth.rs:1875), a duplicated overestimation/cap formula that can drift (eth.rs:1974), and unconditional tipset+msg clones for a rarely-fired retry (eth.rs:1997).
```

Are those plausible?

sudo-shashank:
> @sudo-shashank Did you run your changes against local CC review? I ran it on this PR and it surfaced some potential issues.
> 
> ```
> Correctness:
> 1. eth.rs:2064 — out-of-gas trace check is dead. Some(_expected_exit_code) is a fresh pattern binding, so it matches any exit code, not SYS_OUT_OF_GAS. A genuinely-reverting eth_estimateGas can be pushed into gas_search and return a bogus near-block-limit estimate instead of the revert error. This is the most important one — and it directly qualifies the parity agent's "point 5 correct" verdict.
> 2. message_simulation.rs:195 — skip-retry never fires for ID-mapped senders. The Enforce branch propagates an untyped resolve error, so eth_call from a non-existent 0xff..<id> sender errors out where Lotus (which checks GetActor first) succeeds. The sibling gas.rs path types it correctly, so the two are inconsistent.
> 3. gas.rs:281 — lost diagnostics. Direct Filecoin.GasEstimate* callers now get a bare "sender validation failed" instead of the exit-code + reason main surfaced.
> 
> Plausible:
> 4. gas.rs:225 — map_err(|_| …) swallows the real error. A transient blockstore/state read failure gets misclassified as a sender-validation problem and silently falls into the skip-sender path with a fabricated placeholder.
> 
> Cleanups: a redundant tipset+actor load on the hot EOA path (eth.rs:1875), a duplicated overestimation/cap formula that can drift (eth.rs:1974), and unconditional tipset+msg clones for a rarely-fired retry (eth.rs:1997).
> ```
> 
> Are those plausible?

Yes

LesnyRumcajs:
> Yes

Yes what? Which issues were correctly flagged and fixed, and which did you discard?

sudo-shashank:
> > Yes
> 
> Yes what? Which issues were correctly flagged and fixed, and which did you discard?

1. eth.rs:2064 — out-of-gas trace check is dead is also valid but not covered here as related changes are drafted in https://github.com/ChainSafe/forest/pull/7473. Reviewing the changes there.

2. message_simulation.rs was valid and fixed, `Enforce` maps to `Error::SenderValidationFailed` so the skip-retry fires for ID-mapped senders.

3. gas.rs:281 — lost diagnostics was also valid and fixed

4. gas.rs:225 — map_err(|_| …) swallows the real error, this is also valid but for now we are only preserving the error msg. Still looking into transient read failure getting misclassified.

5. Cleanups are done.

LesnyRumcajs:
> Still looking into transient read failure getting misclassified.

Ok, in this case if you're still looking into it, please put the PR to draft.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@
 
 ### Fixed
 
+- [#7394](https://github.com/ChainSafe/forest/issues/7394): `eth_call` and `eth_estimateGas` now accept a `from` address that is an EVM contract or that doesn't exist on chain. Ports ([filecoin-project/lotus#13724](https://github.com/filecoin-project/lotus/pull/13724)).
```
akaladarshi: I don't think we need to tell the users which PR we have ported. We can simply mention `follows lotus` or `same as lotus ` or don't mentioning anything at all would also be fine.

src/rpc/methods/eth.rs
```diff
@@ -1882,17 +1887,25 @@ async fn eth_estimate_gas(
     // gas estimation actually run.
     msg.gas_limit = 0;
 
+    if sender_validation_for(ctx, &msg.from, &tipset) == SenderValidation::Skip {
```
akaladarshi: We can directly use the `sender_validation_for_actor`.

src/state_manager/message_simulation.rs
```diff
@@ -281,3 +352,32 @@ impl StateManager {
...
+
+        assert_eq!(msg.from, Address::SYSTEM_ACTOR);
+        assert_eq!(msg.to, sender);
+        assert_eq!(msg.method_num, METHOD_SEND);
+        assert_eq!(msg.gas_limit, IMPLICIT_MESSAGE_GAS_LIMIT as u64);
+        assert!(msg.value.is_zero());
+        assert!(msg.gas_fee_cap.is_zero());
+        assert!(msg.gas_premium.is_zero());
+    }
+}
```
LesnyRumcajs:
```suggestion
```

I think it's fair to assume that field assignment in Rust works, and there's no need to test it.

src/state_manager/message_simulation.rs
LesnyRumcajs, line 1: So, just to understand it, right now when we are simulating messages, we're writing some fake executions to the blockstore? If so, it's not great as it can spam the database (not a HUGE deal, we have GC, but I'd rather avoid garbage rather than deal with it when possible). If so, we'd need some mechanism to NOT include them.

It's likely a pre-existing issue so no need to do anything in this PR - this will be a bigger change. Just confirm I understand it correctly.

src/rpc/methods/gas.rs
```diff
@@ -578,6 +626,23 @@ mod tests {
         assert_eq!(msg.gas_premium(), TokenAmount::from_atto(100));
     }
 
+    #[test]
+    fn test_overestimate_applies_the_multiplier_and_truncates() {
```
LesnyRumcajs: I don't see how valuable this test is really. I don't see a need to validate `(gas as f64 * multiplier) as u64` works as expected (unless it's just to win 1 line of coverage, but I'd assume the callers of this method already do it).

src/rpc/methods/eth.rs
```diff
@@ -1906,12 +1919,63 @@ async fn eth_estimate_gas(
...
         Ok(gassed_msg) => {
-            let expected_gas = eth_gas_search(ctx, gassed_msg, &tipset.key().into()).await?;
+            let expected_gas =
+                eth_gas_search(ctx, gassed_msg, &tipset, SenderValidation::Enforce).await?;
             Ok(expected_gas.into())
         }
     }
 }
 
+fn is_sender_validation_failure(err: &Error) -> bool {
```
LesnyRumcajs: 1. You extracted this one but do the downcast thing with others; why? 
2. This (and other occurrences) can be done more neatly with `matches!(err.downcast_ref(), Some(StateManagerError::SenderValidationFailed(_)))`. Likely removing the need for a separate function.

src/state_manager/message_simulation.rs
```diff
@@ -262,13 +290,56 @@ impl StateManager {
...
+                {
+                    Some(actor) => actor,
+                    None => match sender_validation {
+                        SenderValidation::Enforce => {
+                            return Err(Error::SenderValidationFailed(format!(
+                                "sender {} not found on chain",
+                                message.from()
+                            )));
+                        }
+                        SenderValidation::Skip => {
```
akaladarshi: Please some detailed comment explaining about why are we using placeholder and what is it's purpose, otherwise we loose context on this.

Actually I think creating a separate function out of this and providing explanation there would be more clear and reduce clutter here.

Something like 
```Rust
let (from_actor, apply) = sender_for_simulation(&mut vm, message.from(), sender_validation?:
let (ret, duration) = match apply {                                                                                                                                                                                                
      SenderApply::Explicit => vm.apply_message(&message)?,                                                                                                                                                                                               
      SenderApply::Implicit => vm.apply_implicit_message(message.message())?,                 
```

src/state_manager/message_simulation.rs
```diff
@@ -182,26 +182,53 @@ impl StateManager {
...
+                        let TipsetState { state_root, .. } = self.load_tipset_state(&ts).await?;
+                        return Err(if self.get_actor(&msg.from, state_root)?.is_some() {
+                            e.context("could not resolve key")
+                        } else {
+                            Error::SenderValidationFailed(format!(
+                                "sender {} not found on chain",
+                                msg.from
+                            ))
+                            .into()
+                        });
```
LesnyRumcajs: Is this for some error parity with Lotus? This seems like a pretty expensive operation just to pick between two error messages.

LesnyRumcajs (commented):
what's the coverage for the code you added? As in, are there any arms that are not exercised?

src/rpc/methods/gas.rs
```diff
@@ -292,16 +320,36 @@ impl RpcMethod<3> for GasEstimateMessageGas {
...
+pub fn overestimate_gas_limit(data: &Ctx, gas: u64) -> u64 {
+    overestimate(gas, data.mpool.gas_limit_overestimation())
+}
+
+/// As [`overestimate_gas_limit`], clamped to the block gas limit.
+pub fn overestimate_gas_limit_capped(data: &Ctx, gas: u64) -> u64 {
+    overestimate_gas_limit(data, gas).min(BLOCK_GAS_LIMIT)
+}
+
+fn overestimate(gas: u64, multiplier: f64) -> u64 {
```
akaladarshi: These are extra function which are not needed, it is just doing a multiplication, not related to gas at all.

src/rpc/methods/eth.rs
```diff
@@ -1906,12 +1919,63 @@ async fn eth_estimate_gas(
             Err(err.context("failed to estimate gas").into())
         }
```
akaladarshi: This whole part and `eth_estimate_gas_skip_sender` is the same, we can combine the method and reduce the duplication.

src/state_manager/message_simulation.rs
```diff
@@ -182,26 +182,53 @@ impl StateManager {
...
+                    }
```
akaladarshi: This whole `Err` block is not actually needed, since `call_with_gas` already does this.

It can simply be 

```rust
 SenderValidation::Enforce => self                                                                                                                           
          .resolve_to_deterministic_address(msg.from, &ts)                                                                                                        
          .await                                                                                                                                                  
          .context("could not resolve key")?    
```

src/rpc/methods/gas.rs
```diff
@@ -251,22 +262,39 @@ impl GasEstimateGasLimit {
...
-        anyhow::ensure!(
-            apply_ret.exit_code().is_success(),
+        if apply_ret.exit_code().is_success() {
+            return Ok(apply_ret.gas_used() as i64);
+        }
+        if sender_validation == SenderValidation::Enforce
+            && apply_ret.exit_code() == fvm_shared4::error::ExitCode::SYS_SENDER_INVALID
+        {
+            return Err(crate::state_manager::Error::SenderValidationFailed(format!(
+                "message execution failed (exit=[{}], vm error=[{}])",
```
akaladarshi: The error here would differ from lotus, lotus emits `exit=[SysErrSenderInvalid(1)]` not just the value (1) we can use this to get the correct error:
`let exit_code = crate::shim::error::ExitCode::from(apply_ret.exit_code())`

src/state_manager/message_simulation.rs
```diff
@@ -262,13 +290,56 @@ impl StateManager {
...
+                                })?
+                        }
+                    },
+                };
 
                 message.set_sequence(from_actor.sequence);
-                let (ret, duration) = vm.apply_message(&message)?;
+                let (ret, duration) = match (sender_validation, sender_created) {
+                    // An existing non-account sender needs the implicit path, which skips the
+                    // account-type, nonce and balance checks.
```
akaladarshi: Please include details about the `no inclusion cost` in the comment.

src/rpc/methods/eth.rs
```diff
@@ -3890,10 +4005,11 @@ impl RpcMethod<3> for EthTraceCall {
         let (invoke_result, post_state_root) = ctx
             .state_manager
             .apply_on_state_with_gas(
-                Some(ts.shallow_clone()),
-                msg.clone(),
+                Some(&ts),
+                &msg,
                 VMFlush::Flush,
                 VMTrace::Traced,
+                SenderValidation::Enforce,
```
akaladarshi, line 3904: Not sure but want to confirm here.

@LesnyRumcajs  Do we care about skipping validation in the `EthTraceCall`, lotus doesn't have this so worth thinking about it.

src/state_manager/mod.rs
```diff
@@ -215,6 +215,14 @@ pub enum VMFlush {
     Skip,
 }
 
+/// Controls whether the FVM sender checks are enforced when simulating a message.
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+pub enum SenderValidation {
+    #[default]
+    Enforce,
+    Skip,
```
akaladarshi, line 245: Worth adding this `Control flow only: callers use it to retry with skip-sender-validation`. 

Basically to enforce that this not the general behaviour and will be changed once the FVM supports this.

src/rpc/methods/gas.rs
```diff
@@ -251,22 +262,39 @@ impl GasEstimateGasLimit {
...
+        msg: Message,
+        curr_ts: &Tipset,
+        sender_validation: SenderValidation,
+    ) -> Result<i64> {
+        let (apply_ret, ..) = Self::measure_gas_used(data, msg, curr_ts, sender_validation)
             .await
             .context("gas estimation failed")?;
-        anyhow::ensure!(
-            apply_ret.exit_code().is_success(),
+        if apply_ret.exit_code().is_success() {
```
akaladarshi: nit: just store the  value `apply_ret.exit_code()` once.

akaladarshi (changes requested):

src/state_manager/message_simulation.rs
sudo-shashank: Yes, makes sense

src/state_manager/message_simulation.rs
sudo-shashank: Yes, it's something pre-existing

src/rpc/methods/gas.rs
sudo-shashank: removed

src/rpc/methods/eth.rs
sudo-shashank: No need for this extraction better to inline like you suggested

src/state_manager/message_simulation.rs
sudo-shashank: Yeah, it was for an error parity with lotus. `call_with_gas` already handles it, so we can drop this

CHANGELOG.md
sudo-shashank: We can do `matching Lotus`. I kept the link so it was clear where this was coming from. 

I'll remove the link

src/state_manager/message_simulation.rs
akaladarshi: Are we actually hitting the database?

I mean we are doing the `VMFlush::Skip` aren't we or does it have different purpose?

src/rpc/methods/eth.rs
sudo-shashank: done

src/state_manager/message_simulation.rs
sudo-shashank: added

src/state_manager/message_simulation.rs
sudo-shashank: We hit db only when we flush. `eth_call` / `eth_estimateGas` use `VMFlush::Skip` so they never hit the db. `call_with_gas`(VMFlush::Flush) and `call_raw_blocking` hit the DB which is pre-existing.

src/rpc/methods/gas.rs
sudo-shashank: Cleaned up

src/rpc/methods/eth.rs
sudo-shashank: extracted error recovery part

src/state_manager/message_simulation.rs
sudo-shashank: done

src/state_manager/message_simulation.rs
sudo-shashank: included

src/state_manager/mod.rs
sudo-shashank: added

src/rpc/methods/gas.rs
sudo-shashank: correct, fixed

src/state_manager/message_simulation.rs
akaladarshi: `call_with_gas` only Flushes when there is `EthTraceCall` API. Which is required because we need post state change to track the diff.
So generally all the calls that doesn't need to flush anything aren't flushing anything in DB only the required ones are.

src/rpc/methods/eth.rs
```diff
@@ -1903,36 +1909,89 @@ async fn eth_estimate_gas(
...
+async fn recover_estimate_gas_error(
+    ctx: &Ctx,
+    mut msg: Message,
+    tipset: &Tipset,
+    estimate_err: anyhow::Error,
+) -> ServerError {
+    msg.set_gas_limit(BLOCK_GAS_LIMIT);
+    if let Err(e) = apply_message(ctx, Some(tipset), &msg).await
+        && matches!(e.downcast_ref(), Some(EthErrors::ExecutionReverted { .. }))
+    {
```
akaladarshi, line 1984: I think this condition is incorrect.
If `apply_message` didn't return an `ExecutionReverted` then?

src/rpc/methods/eth.rs
LesnyRumcajs: I think we do? I guess this could be tackled as a separate issue, though.

src/rpc/methods/eth.rs
sudo-shashank: then we return the original estimate error and the condition here matches [lotus](https://github.com/filecoin-project/lotus/blob/master/node/impl/eth/gas.go#L264-L272)

src/rpc/methods/eth.rs
sudo-shashank: @coderabbitai create a new issue to track this

src/rpc/methods/eth.rs
akaladarshi: So the thing is if the `estimate_message_gas` can also returns the `ExecutionReverted` then we don't need to `apply_message` again [here](https://github.com/filecoin-project/lotus/blob/master/node/impl/eth/gas.go#L239-L244).

So here is another thing that was missed `estimate_gas_limit` is also supposed to return the `ExecutionReverted` [here](https://github.com/filecoin-project/lotus/blob/master/node/impl/gasutils/gasutils.go#L272-L278), if it returns then the `estimate_message_gas` itself has to capture it without calling the `apply_message`.

src/rpc/methods/gas.rs
```diff
@@ -251,22 +262,40 @@ impl GasEstimateGasLimit {
...
+            && exit_code == fvm_shared4::error::ExitCode::SYS_SENDER_INVALID
+        {
+            let exit_code = crate::shim::error::ExitCode::from(exit_code);
+            return Err(crate::state_manager::Error::SenderValidationFailed(format!(
+                "message execution failed (exit=[{exit_code}], vm error=[{}])",
+                apply_ret.failure_info().unwrap_or_default()
+            ))
+            .into());
+        }
+        anyhow::bail!(
```
akaladarshi: Based on the above comment, Here we need to also return the `execution reverted` current error is wrong.

src/rpc/methods/eth.rs
sudo-shashank: Right, Fixed
`estimate_gas_limit` returns `ExecutionReverted`, and `eth_estimate_gas` returns it as-is without `apply_message`.

src/rpc/methods/gas.rs
sudo-shashank: fixed

src/rpc/methods/eth.rs
akaladarshi: Now, since you're covering everything under the execution reverted [Out of Gas ](https://github.com/filecoin-project/lotus/blob/533aaa91772d55879d5898a0a49f75f51b671ab9/node/impl/gasutils/gasutils.go#L267) will be lost in between.

Also now the `apply_message` is not getting executed, after the `estimate_message_gas`. So it will be missed as well.

I will suggest please take a look at the code flow of lotus once again and verify if all the errors and other internal things are handled correctly.

sudo-shashank: Looking into this, I'll verify it and check Lotus flow again

sudo-shashank: Good catch I missed it, this needs to be handled correctly

src/rpc/methods/gas.rs
```diff
@@ -251,22 +264,41 @@ impl GasEstimateGasLimit {
...
+        {
+            let exit_code = crate::shim::error::ExitCode::from(exit_code);
+            return Err(crate::state_manager::Error::SenderValidationFailed(format!(
+                "message execution failed (exit=[{exit_code}], vm error=[{}])",
+                apply_ret.failure_info().unwrap_or_default()
+            ))
+            .into());
+        }
+        let vm_error = apply_ret.failure_info().unwrap_or_default();
+        let (data, reason) = decode_revert_reason(apply_ret.return_data());
```
akaladarshi: `data` shadows the original param `data: &Ctx`

src/rpc/methods/gas.rs
```diff
@@ -251,22 +264,41 @@ impl GasEstimateGasLimit {
...
+        Err(EthErrors::execution_reverted(exit_code.into(), &reason, &vm_error, &data).into())
```
akaladarshi: Instead of making the `decode_revert_reason` pub, you can directly use the `execution_reverted` inside the `EthErrors` constructor. 

```rust
pub fn execution_reverted_from_result(                                                                       
          exit_code: impl Into<ExitCode>,                                                                          
          return_data: RawBytes,                                                                                   
          vm_error: &str,                                                                                          
      ) -> Self {                                                                                                  
          let (data, reason) = decode_revert_reason(return_data);                                                  
          Self::execution_reverted(exit_code.into(), &reason, vm_error, &data)                                     
      }      
```

akaladarshi (changes requested):
Couple of comments otherwise, LGTM

src/rpc/methods/gas.rs
sudo-shashank: Done

src/rpc/methods/gas.rs
sudo-shashank: Done

akaladarshi (approved):

PR #7427, opened 2026-07-30
fix `MinerGetBaseInfo` returns the correct `BeaconEntries`

EclesioMeloJunior:
## Summary of changes

After the quicknet upgrade Forest was returning the incorrect set of `BeaconEntries` as it was hard-coded to return a single entry, while when a null-round happens a Filecoin block should be able to cover every epoch since its parent and one beacon entry per covered epoch

Changes introduced in this pull request:

- `src/beacon/drand.rs`: basicaly on the quicknet "unchained" branch fetch the `max_beacon_round` from `(parent_epoch + 1)` till `epoch`

## Reference issue to close (if applicable)

Closes https://github.com/ChainSafe/forest/issues/7412

## Change checklist

- [x] I have performed a self-review of my own code,
- [ ] I have made corresponding changes to the documentation. All new code adheres to the team's [documentation standards](https://github.com/ChainSafe/forest/wiki/Documentation-practices),
- [x] I have added tests that prove my fix is effective or that my feature works (if possible),
- [x] I have made sure the [CHANGELOG][1] is up-to-date. All user-facing changes should be reflected in this document.

### Outside contributions

- [ ] This pull request is based on an issue that a maintainer has accepted (see [Before Opening a Pull Request][4]).
- [ ] I have read and agree to the [CONTRIBUTING][2] document.
- [ ] I have read and agree to the [AI Policy][3] document. I understand that failure to comply with the guidelines will lead to rejection of the pull request.

[1]: https://github.com/ChainSafe/forest/blob/main/CHANGELOG.md
[2]: https://github.com/ChainSafe/forest/blob/main/CONTRIBUTING.md
[3]: https://github.com/ChainSafe/forest/blob/main/AI_POLICY.md
[4]: https://github.com/ChainSafe/forest/blob/main/CONTRIBUTING.md#before-opening-a-pull-request

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Quicknet beacon handling across multiple Filecoin epochs, including epochs containing null beacon rounds.
  * Ensured beacon entries are generated correctly for every covered epoch.
  * Improved consistency when processing beacon data across supported network configurations.

* **Tests**
  * Added coverage for Quicknet scenarios with and without null beacon rounds.

* **Documentation**
  * Updated the unreleased changelog with details of the Quicknet beacon handling improvement.

PR #7441, opened 2026-08-04
fix(blocks): validate all beacon entries for all covered epochs on quicknet 

### LINKED ISSUE #7413: `validate_block_drand` skips per-epoch round checks for unchained beacons
> [!NOTE]
> This was AI found and seems correct, but do your own checks. Regardless of that, a test will be useful.

> [!TIP]
> It might be best to fix this together with https://github.com/ChainSafe/forest/issues/7412, as those are pretty much related and address the same class of bug.

**Location:** [`src/blocks/header.rs#L172-L188`](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/blocks/header.rs#L172-L188), in [`validate_block_drand` (`#L99-L191`)](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/blocks/header.rs#L99-L191), backed by [`verify_entries` (`src/beacon/drand.rs#L259-L324`)](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/beacon/drand.rs#L259-L324).

### What Forest checks (unchained)

1. the **last** entry's round equals `max_beacon_round_for_epoch(epoch)` ([`header.rs#L172`](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/blocks/header.rs#L172)), and
2. every entry's BLS signature is valid for *its own claimed round* ([`verify_entries` unchained path `#L265-L285`](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/beacon/drand.rs#L265-L285), via `message_unchained(entry.round())`).

### What Lotus additionally checks

[`ValidateBlockValues` (`chain/beacon/beacon.go#L92-L100`)](https://github.com/filecoin-project/lotus/blob/46b92c357134c248c6ee3be633fa18e364033311/chain/beacon/beacon.go#L92-L100):

```go
if !currBeacon.IsChained() {
    for i, e := range h.BeaconEntries {
        correctRound := currBeacon.MaxBeaconRoundForEpoch(nv, parentEpoch+abi.ChainEpoch(i)+1)
        if e.Round != correctRound {
            return xerrors.Errorf("unexpected beacon round %d, expected %d for epoch %d", ...)
        }
    }
}
```

Entry `i` must be at exactly the round for epoch `parentEpoch+i+1`. Combined with the last-round check, this pins both the **count** (`= epoch − parentEpoch`) and the **exact round of every entry**.

### The gap

Forest enforces only the last entry's round plus per-entry signature validity — not that intermediate entries sit at the correct per-epoch rounds, nor that the count matches the covered-epoch span. Since drand signatures for all rounds are public, wrong-but-real entries (each individually BLS-valid, last one at `max_round`) pass Forest and fail Lotus.

The rest of Forest's block validation does **not** compensate: election proof, ticket, and winning-PoSt validation all draw randomness from [`header.beacon_entries.last()`](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/fil_cns/validation.rs#L259) only ([ticket `#L313`](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/fil_cns/validation.rs#L313), [winning PoSt `#L350`](https://github.com/ChainSafe/forest/blob/910cbf11f65cb95dbcc83df6a4f739266f1a621e/src/fil_cns/validation.rs#L350)). Nothing iterates the list or constrains its count/positions. So `validate_block_drand` is the sole structural constraint, and the divergence is reachable at the validation layer. It is one-directional: Forest's checks are a strict subset of Lotus's, so Forest never rejects a Lotus-valid block on these grounds — it only accepts extra ones.

### Reachability

Two triggers produce a block Forest accepts and mainnet rejects:

- **Adversarial (Bug 2 alone):** a miner with block-producing power wins a round legitimately (so election/ticket/PoSt validate off the correct `.last()` entry), then pads the header with extra valid-but-wrong-round entries. Forest accepts; Lotus rejects.
- **Honest (Bug 1 ∘ Bug 2):** a miner using Forest's `MinerGetBaseInfo` after a null round gets a one-entry base (Bug 1) where the network requires one per covered epoch. Forest validates it (last entry is `max_round(E)`, signature valid); Lotus rejects it (positional loop expects `max_round(E−1)` at index 0). No malice required.

### Exploitability / severity

Low in Forest's current reality:

- **Forest is not a block producer**, so the honest trigger only fires if someone mines through Forest on mainnet — effectively nobody today. Latent, not live.
- **The adversarial path is economically irrational.** It needs real mainnet storage power, and a miner produces one block per election win — publishing the malformed block means forfeiting the block reward of the valid one you could have produced instead. Cost per attempt is a lost block reward; direct gain is zero (no theft, no double-spend).
- **Impact is transient and Forest-only.** It affects only Forest nodes (a network minority), and the malformed block's fork gets no follow-on weight, so Forest reorgs onto the heavier mainnet chain and drops it. Concrete damage: Forest serves a divergent tipset (wrong messages/receipts/state over RPC) for the affected height, then self-corrects.

Not embargo-worthy. Two things would raise it: (1) Forest becoming a mining backend, at which point Bug 1 ∘ Bug 2 is a free, recurring, honest self-split after every null round; (2) F3 finalizing Forest's divergent view before the reorg — **unverified here**, worth a check by whoever takes the fix, since it could turn "transient" into "sticky".

### Fix

In the unchained validation path, add Lotus's positional loop — assert `beacon_entries.len() == epoch − parent_epoch` and that entry `i`'s round equals `max_beacon_round_for_epoch(parent_epoch + i + 1)`. This is the same per-epoch-round rule as Bug 1, applied on the validation side rather than the production side — fix both together behind a shared helper. Compare against Lotus's read-side equivalent, [`getBeaconEntryV3` (`chain/rand/rand.go#L146`)](https://github.com/filecoin-project/lotus/blob/46b92c357134c248c6ee3be633fa18e364033311/chain/rand/rand.go#L146).

EclesioMeloJunior:
## Summary of changes

This PR fixes the `RawBlockHeader::validate_block_drand` method by checking correctly, for quicknet, if the current block header contains the beacon entries from `parent_epoch + 1` till `self.epoch`, meaning that a block built after `N` null-epoch should have a beacon entry per skipped epoch

- `src/blocks/header.rs`: `validate_block_drand` now checks, for unchained beacons only, that the header carries exactly `epoch - parent_epoch` entries and that entry `i` is at `max_beacon_round_for_epoch(parent_epoch + i + 1)`

Closes #7413

- Companion PR #7427 fixes the producing side, where `Filecoin.MinerGetBaseInfo` returned too few entries for the same reason. This PR is the validating side.

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Quicknet block validation to require complete beacon coverage for every epoch between the parent and new block.
  * Beacon entries must appear in ascending epoch order and match each epoch’s maximum beacon round.
  * Blocks with missing, extra, incorrectly ordered, or mismatched beacon entries are now rejected with clear validation errors.
  * Validation now correctly handles epochs with null beacon rounds, improving reliability for affected blocks.


src/blocks/header.rs
```diff
@@ -180,6 +180,48 @@ impl RawBlockHeader {
             ));
         }
 
+        // An unchained beacon carries no link between consecutive entries, so nothing
+        // but these checks ties them to the epochs they are supposed to cover. A block
+        // covers every epoch since its parent - normally just its own, but null rounds
+        // can happen and it must carry exactly one entry per covered epoch, in
+        // ascending order.
+        if curr_beacon.network().is_unchained() {
```
LesnyRumcajs, line 184: Given it's a consensus-affecting change, can we have a source to Lotus code with this logic?

src/blocks/header.rs
```diff
@@ -180,6 +180,48 @@ impl RawBlockHeader {
...
+                    )
+                    .into(),
+                ));
+            }
+
+            // Lengths match, so `zip` pairs every entry with the epoch it must cover.
+            for (beacon_entry, lookup_epoch) in self
+                .beacon_entries
+                .iter()
+                .zip((parent_epoch + 1)..=self.epoch)
```
LesnyRumcajs: nit: I think you could simplify this via `izip!` https://docs.rs/itertools/latest/itertools/macro.izip.html

and end up with, e.g., `... in izip!(self.beacon_entries.iter(), (parent_epoch+1)..)`.

src/blocks/header.rs
```diff
@@ -432,4 +477,42 @@ mod tests {
             *FILECOIN_GENESIS_CID
         );
     }
+
+    #[tokio::test]
```
LesnyRumcajs, line 526: do we have a failing case?

src/blocks/header.rs
```diff
@@ -180,6 +180,48 @@ impl RawBlockHeader {
...
+                            "expected max round for epoch {} to be {}, got: {}",
+                            lookup_epoch,
+                            expected_round,
+                            beacon_entry.round(),
+                        )
+                        .into(),
+                    ));
+                }
+            }
+        }
```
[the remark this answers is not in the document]
EclesioMeloJunior: fixed #7427

src/blocks/header.rs
EclesioMeloJunior: Addressed at [0096657](https://github.com/ChainSafe/forest/pull/7441/commits/0096657517f609d29c182778c1657e4e18df569f)

src/blocks/header.rs
EclesioMeloJunior: Addressed at [0096657](https://github.com/ChainSafe/forest/pull/7441/commits/0096657517f609d29c182778c1657e4e18df569f)

src/blocks/header.rs
EclesioMeloJunior: Addressed at [89ac8fa](https://github.com/ChainSafe/forest/pull/7441/commits/89ac8fa3c8249817078eafe5cb6a9fce108637c1)

src/blocks/header.rs
```diff
@@ -201,10 +203,8 @@ impl RawBlockHeader {
             }
 
             // Lengths match, so `zip` pairs every entry with the epoch it must cover.
-            for (beacon_entry, lookup_epoch) in self
-                .beacon_entries
-                .iter()
-                .zip((parent_epoch + 1)..=self.epoch)
+            for (beacon_entry, lookup_epoch) in izip!(
+                self.beacon_entries.iter(), (parent_epoch+1)..=self.epoch)
```
LesnyRumcajs: Didn't you already assert the lenghts match? In this case the `=self.epoch` shouldn't be needed, no?

EclesioMeloJunior: yeah, there is a length check, that is indeed unneeded!

EclesioMeloJunior: Addressed at [8e4fb8a](https://github.com/ChainSafe/forest/pull/7441/commits/8e4fb8a2d070cd18ebf9d880033a7e051c198816)

LesnyRumcajs:
nit: there are some [partially covered lines](https://app.codecov.io/gh/ChainSafe/forest/pull/7441?src=pr&el=tree) in the coverage you might want to cover (they're fairly trivial to reason about, so not a huge deal).

I'll approve once CI is green.

src/blocks/header.rs
```diff
@@ -180,6 +181,47 @@ impl RawBlockHeader {
...
+            if found_entries != covered_epochs {
+                return Err(Error::Validation(
+                    format!(
+                        "expected {covered_epochs} beacon entries for epochs {}..={}, got: {found_entries}",
+                        parent_epoch + 1,
+                        self.epoch,
+                    )
+                    .into(),
+                ));
+            }
```
LesnyRumcajs: I wonder if it's even needed. There's no Lotus equivalent for this. Wouldn't this logic be exercised via the loop below?

src/blocks/header.rs
```diff
@@ -432,4 +477,87 @@ mod tests {
             *FILECOIN_GENESIS_CID
         );
     }
+
+    #[tokio::test]
+    async fn validate_beacon_entries_on_quicknet() {
+        // (acse name, parent epoch, its beacon round, block epoch, rounds the header must carry, expect fail)
```
LesnyRumcajs:
```suggestion
        // (case name, parent epoch, its beacon round, block epoch, rounds the header must carry, expect fail)
```

src/blocks/header.rs
EclesioMeloJunior: yeah, lotus enforce it via 2 checks, ensure the last beacon entry matches the `self.epoch` max round & the loop ensures it covers since the `parent + 1` epoch till `self.epoch`, we do the `last == max` and the loop, so no need to compare the length

src/blocks/header.rs
LesnyRumcajs: btw you might be interested in some `rstest` applications we're doing in some other tests.

src/blocks/header.rs
```diff
@@ -466,86 +466,98 @@ mod tests {
...
+        accepted: false,
+    })]
+    // the block's own entry is missing, so the last entry isn't at `max_round`
+    #[case::null_round_missing_own_entry(BeaconEntriesCase {
+        parent_epoch: 6216198,
+        prev_round: 30662982,
+        epoch: 6216200,
+        rounds: vec![30662992],
+        accepted: false,
+    })]
```
LesnyRumcajs, line 511:
```rust
#[case::extra_trailing_entry(BeaconEntriesCase {
    parent_epoch: 6216199,
    prev_round: 30662992,
    epoch: 6216200,
    rounds: vec![30663002, 30663002],
    accepted: false,
})]
#[case::null_round_extra_trailing_entry(BeaconEntriesCase {
    parent_epoch: 6216198,
    prev_round: 30662982,
    epoch: 6216200,
    rounds: vec![30662992, 30663002, 30663002],
    accepted: false,
})]
```

Would these pass? If not, why?

EclesioMeloJunior: they failed 😢 , investigated and the bounding the `izip!` to `self.epoch` `izip!(entries, (parent_epoch + 1)..=self.epoch)` obfuscates the trailing entry, while in lotus they go over all entries (which would catch these extra entry and validate which would cause a fail) 

The fix here is just to remove the bound so we will validate all entry and it's expected round

EclesioMeloJunior: thanks for the catch! addressed at [6c46c3b](https://github.com/ChainSafe/forest/pull/7441/commits/6c46c3b43929584d51e1dab816c004400019d735)

src/blocks/header.rs
```diff
@@ -180,6 +181,37 @@ impl RawBlockHeader {
             ));
         }
 
+        // An unchained beacon carries no link between consecutive entries, so nothing
```
LesnyRumcajs: I'd trim those comments down, pretty long comments for such short code.

src/blocks/header.rs
```diff
@@ -180,6 +181,37 @@ impl RawBlockHeader {
...
+        // ascending order.
+        // ref: https://github.com/filecoin-project/lotus/blob/27abf0f16a7f2a83305910f3c2a1844764d20b75/chain/beacon/beacon.go#L95
+        if curr_beacon.network().is_unchained() {
+            // We already made sure that the last beacon entry matches the current block
+            // epoch, and this loop covers a possible gap between the parent block and
+            // the current one. The epoch range is deliberately unbounded so that every
+            // entry is checked: an entry past `self.epoch` has no epoch to cover and is
+            // rejected against the round of the epoch that would follow this block.
+            for (beacon_entry, lookup_epoch) in
+                izip!(self.beacon_entries.iter(), (parent_epoch + 1)..)
```
LesnyRumcajs: nit: at this point I think it makes sense to just go with standard `.enumerate()` which everyone knows

src/blocks/header.rs
```diff
@@ -432,4 +468,114 @@ mod tests {
...
+    })]
+    // a duplicated final entry: one more entry than the covered range
+    #[case::extra_trailing_entry(BeaconEntriesCase {
+        parent_epoch: 6216199,
+        prev_round: 30662992,
+        epoch: 6216200,
+        rounds: vec![30663002, 30663002],
+        accepted: false,
+    })]
+    #[case::null_round_extra_trailing_entry(BeaconEntriesCase {
```
LesnyRumcajs, line 519: nit: it's the only one lacking comment, but subjectively those comments are not that useful - test cases names are already pretty self-explanatory

LesnyRumcajs (dismissed):
LGTM

LesnyRumcajs:
I had to do a bump commit, the CI seemed to got stuck due to yesterday's GH outage.

LesnyRumcajs (approved):
We should limit the drand relay calls in tests, but that's a precedent not introduced here. Otherwise LGTM

PR #7459, opened 2026-08-06
fix: limit possible underflows/overflows

### LINKED ISSUE #7448: Clear out unsound overflows/underflows
# Summary

With the current profile we're building with, Forest will panic (and then abort) on overflow/underflow. I'm inclined to keep it for now for correctness purposes. (better fail fast rather than continue with some garbage data).

We need some safeguards, especially where users or peers are providing input.

# Completion Criteria

- [ ] Iterate over possible places where a user-supplied underflow/overflow might happen
- [ ] harden those
- [ ] there might be cases where the logic was blindly copied from Lotus which would do likely wrapping overflows. In case such thing is not sound, this must be reported to FilOz as well.

LesnyRumcajs:
## Summary of changes

- limits possible underflows and overflows + adds tests to improve coverage around those
- most of those are pretty benign (we kind of assume drand HTTP relay won't be malicious but it might have a bad day...), but it's good to deal with those proactively
- extracted hex helper to reduce code duplication around prefixed hex-encoded values.

Closes https://github.com/ChainSafe/forest/issues/7448

## Change checklist

### Outside contributions

## Bug Fixes
- Improved validation for negative, invalid, and out-of-range epochs across chain, RPC, state, and beacon operations.
- Prevented arithmetic overflows and underflows in timestamps, lookbacks, leases, checkpoints, and range calculations.
- Improved beacon traversal, round validation, and error reporting.
- Added safer handling for malformed hexadecimal values, invalid database offsets, and unsafe export ranges.
- Prevented invalid sector durations and strengthened block-range and participation-lease validation.
- Added clearer errors for invalid heights and malformed requests.

### DIFF
--- a/src/beacon/drand.rs
+++ b/src/beacon/drand.rs
@@ -87,8 +87,7 @@ impl BeaconSchedule {
             let (pb_epoch, _) = self.beacon_for_epoch(parent_epoch)?;
             if cb_epoch != pb_epoch {
                 // Fork logic, take entries from the last two rounds of the new beacon.
-                let round = curr_beacon.max_beacon_round_for_epoch(network_version, epoch);
-
+                let round = curr_beacon.max_beacon_round_for_epoch(network_version, epoch)?;
                 let out = vec![
                     curr_beacon.entry(round - 1).await?,
                     curr_beacon.entry(round).await?,
@@ -97,7 +96,7 @@ impl BeaconSchedule {
             }
         }
 
-        let max_round = curr_beacon.max_beacon_round_for_epoch(network_version, epoch);
+        let max_round = curr_beacon.max_beacon_round_for_epoch(network_version, epoch)?;
         // We don't expect this to ever be the case
         if max_round == prev.round() {
             tracing::warn!(
@@ -117,18 +116,19 @@ impl BeaconSchedule {
 
         let mut out = Vec::with_capacity(2);
         if curr_beacon.network().is_unchained() {
-            for covered_epoch in (parent_epoch + 1)..=epoch {
-                let round = curr_beacon.max_beacon_round_for_epoch(network_version, covered_epoch);
+            // Newest-first, so a large gap fails on its first unavailable round:
+            // <https://github.com/filecoin-project/lotus/blob/v1.35.1/chain/beacon/beacon.go#L152>
+            for covered_epoch in (parent_epoch + 1..=epoch).rev() {
+                let round =
+                    curr_beacon.max_beacon_round_for_epoch(network_version, covered_epoch)?;
                 out.push(curr_beacon.entry(round).await?);
             }
+            out.reverse();
             Ok(out)
         } else {
-            let mut cur = max_round;
-            while cur > prev_round {
-                // Push all entries from rounds elapsed since the last chain epoch.
-                let entry = curr_beacon.entry(cur).await?;
-                cur = entry.round() - 1;
-                out.push(entry);
+            // Rounds elapsed since the last chain epoch, newest-first as above.
+            for round in (prev_round + 1..=max_round).rev() {
+                out.push(curr_beacon.entry(round).await?);
             }
             out.reverse();
             Ok(out)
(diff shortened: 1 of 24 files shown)


src/rpc/methods/eth/filter/mod.rs
```diff
@@ -681,6 +681,9 @@ fn parse_block_range(
         max_height >= 0 || max_height == -1,
         "max_height requested is less than 0"
     );
+    // Unlike `max_height`, `-1` is not a sentinel here: "latest" resolves to `heaviest` above,
+    // which leaves the `min_height == -1` branch below inert, as it is in Lotus.
+    ensure!(min_height >= 0, "min_height requested is less than 0");
```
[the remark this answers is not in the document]
LesnyRumcajs: unreachable

src/beacon/drand.rs
```diff
@@ -87,8 +87,7 @@ impl BeaconSchedule {
             let (pb_epoch, _) = self.beacon_for_epoch(parent_epoch)?;
             if cb_epoch != pb_epoch {
                 // Fork logic, take entries from the last two rounds of the new beacon.
-                let round = curr_beacon.max_beacon_round_for_epoch(network_version, epoch);
-
+                let round = curr_beacon.max_beacon_round_for_epoch(network_version, epoch)?;
```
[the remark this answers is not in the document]
LesnyRumcajs: unreachable

src/daemon/db_util.rs
```diff
@@ -654,12 +654,28 @@ async fn process_ts(
...
+        match (to, n_tipsets) {
+            (Some(to), None) => {
+                anyhow::ensure!(to >= 0, "'to' must not be negative, got {to}.");
+                Ok(Self::To(to))
+            }
+            (None, Some(n)) => Ok(Self::NumTipsets(n)),
+            (None, None) => anyhow::bail!("You must provide either 'to' or 'n_tipsets'."),
+            (Some(_), Some(_)) => anyhow::bail!("'to' and 'n_tipsets' are mutually exclusive."),
+        }
+    }
```
EclesioMeloJunior, line 676: maybe have two constructors?

```rs
pub fn new_with_chain_epoch(to: ChainEpoch) -> Self {}

pub fn new_with_tipsets(n_tipsets: usize) -> Self {}
```

also is it possible to enforce the constraint `'to' must not be negative` at ChainEpoch type? so we don't need that extra check here once we make sure ChainEpoch is always a positive integer

src/utils/encoding/hex.rs
```diff
@@ -27,6 +29,24 @@ pub fn encode_prefixed(data: impl AsRef<[u8]>) -> String {
...
+    T: num_traits::Num,
+    <T as num_traits::Num>::FromStrRadixErr: std::fmt::Display,
+{
+    let digits = input
+        .strip_prefix("0x")
+        .with_context(|| format!("not a 0x-prefixed hex integer: {input}"))?;
+    anyhow::ensure!(
+        !digits.starts_with(['+', '-']),
+        "signed hex integer: {input}"
+    );
```
EclesioMeloJunior, line 46: can we drop this `ensure!` as any invalid sign would be catch by `T::from_str_radix`? or `0x-ae1d` is a valid hex number that `T::from_str_radix` allows?

src/daemon/db_util.rs
LesnyRumcajs: Look at the usage - the goal of this is to simplify it. With two c-tors we'd push handling some of this logic to callers.

`ChainEpoch` is an external type we're using (and alias to some signed type AFAIR). We're also using everywhere so any change around this type would be become a massive PR on its own. I also recall some logic that actually expects this to be negative, e.g., network upgrade schedule.

src/utils/encoding/hex.rs
LesnyRumcajs: `from_str_radix` doesn't bat an eye

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e8f26d7e3934f5ec10464851f3757f99

LesnyRumcajs: Note `0xdeadbeef` would fail, but not because of any `+/-` - we strip the `0x` for a reason here.

EclesioMeloJunior (approved):

PR #7498, opened 2026-08-17
feat: improve `eth_getBlockReceipts` perf

LesnyRumcajs:
## Summary of changes

- improves event collection from quadratic to linear
- less message re-hashing
- checked correctness against riba dataset manually (mainnet 6277900–6279799 (~1900 epochs, 2026-08-14))

```
┌──────────┬─────────────────┬────────┬────────┬───────────┬────────────────┬─────────┬─────────┬──────────┐
│   net    │      range      │ blocks │ max rx │ NEW total │ BASELINE total │ speedup │ new/blk │ base/blk │
├──────────┼─────────────────┼────────┼────────┼───────────┼────────────────┼─────────┼─────────┼──────────┤
│ calibnet │ 3986400–3986637 │ 235    │ 73     │ 773 ms    │ 1303 ms        │ 1.69×   │ 3.29 ms │ 5.54 ms  │
├──────────┼─────────────────┼────────┼────────┼───────────┼────────────────┼─────────┼─────────┼──────────┤
│ mainnet  │ 6279600–6279838 │ 237    │ 27     │ 15 898 ms │ 30 611 ms      │ 1.93×   │ 67.1 ms │ 129.2 ms │
└──────────┴─────────────────┴────────┴────────┴───────────┴────────────────┴─────────┴─────────┴──────────┘

Per-large-block hot path (calibnet, 65–73 rx, warm 50 iters): new 2.7 ms vs baseline ~11 ms ≈ 4×.
```
so on average almost 2x improvement, bad case 4x (and in case of bigger blocks it only gets better).

Closes

## Change checklist

### Outside contributions

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Ethereum transaction receipts to include event logs more reliably.
  - Corrected event-log handling for reverted transactions.
  - Improved consistency of logs across block receipts and individual transaction lookups.
  - Improved receipt accuracy when multiple transactions emit logs within the same block.
  - Added clearer handling when event data is unavailable for the current chain head.
  - Improved event ordering and transaction association across messages, including messages without events.

### DIFF
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -1608,10 +1608,13 @@ async fn get_block_receipts(
     let mut eth_receipts = Vec::with_capacity(executed_messages.len());
     for (
         i,
-        ExecutedMessage {
-            message, receipt, ..
-        },
-    ) in executed_messages.iter().enumerate()
+        (
+            ExecutedMessage {
+                message, receipt, ..
+            },
+            logs,
+        ),
+    ) in executed_messages.iter().zip(logs_by_msg).enumerate()
     {
         let tx = new_eth_tx(
             ctx,
@@ -1622,10 +1625,6 @@ async fn get_block_receipts(
             i as u64,
         )?;
 
-        let logs = logs_by_msg
-            .get_mut(i)
-            .map(std::mem::take)
-            .unwrap_or_default();
         let receipt = new_eth_tx_receipt(&ts_ref, &tx, receipt, logs)?;
         eth_receipts.push(receipt);
     }
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -1606,16 +1606,10 @@ async fn get_block_receipts(
 
     let block_cid = ts_key.cid()?;
     let mut eth_receipts = Vec::with_capacity(executed_messages.len());
-    for (
-        i,
-        (
-            ExecutedMessage {
-                message, receipt, ..
-            },
-            logs,
-        ),
-    ) in executed_messages.iter().zip(logs_by_msg).enumerate()
-    {
+    for (i, (executed, logs)) in executed_messages.iter().zip(logs_by_msg).enumerate() {
+        let ExecutedMessage {
+            message, receipt, ..
+        } = executed;
         let tx = new_eth_tx(
             ctx,
             &state_tree,
--- a/src/rpc/methods/eth.rs
+++ b/src/rpc/methods/eth.rs
@@ -1353,12 +1353,11 @@ fn new_eth_tx(
     })
 }
 
-async fn new_eth_tx_receipt(
-    ctx: &Ctx,
+fn new_eth_tx_receipt(
     tipset: &Tipset,
     tx: &ApiEthTx,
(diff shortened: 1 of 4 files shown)


src/rpc/methods/eth.rs
```diff
@@ -1581,6 +1584,27 @@ async fn get_block_receipts(
     // Load the state tree
     let state_tree = ctx.state_manager.get_state_tree(&state_root)?;
 
+    // Collect the whole tipset's logs in a single pass and scatter them back to their messages.
+    let heaviest_epoch = ctx.chain_store().heaviest_tipset().epoch();
+    if ts_ref.epoch() >= heaviest_epoch
```
akaladarshi, line 1589: This `(ts_ref.epoch() >= heaviest_epoch)` seems to be the common condition for the `EventsNotYetAvailable` error.
We can make a method out of it and use it everywhere
```
fn ensure_events_available(ctx: &Ctx, epoch: ChainEpoch) -> anyhow::Result<()> {
    if epoch >= ctx.chain_store().heaviest_tipset().epoch() {
        return Err(EthErrors::EventsNotYetAvailable.into());
    }
    Ok(())
}
```

src/rpc/methods/eth.rs
akaladarshi: We can use below to avoid another iteration on `logs_by_msg`
```suggestion
    executed_messages.iter().zip(logs_by_msg).enumerate()
```

akaladarshi (changes requested):

akaladarshi:
@LesnyRumcajs Also if possible can we also include some tests to make sure the indexing of transaction is correct in `collect_events_from_messages` (`transaction_index == msg_idx == executed_message`) because we rely on it a lot.

I mean it is correct we already know that but just adding a test if possible might make it regression free.

src/rpc/methods/eth.rs
LesnyRumcajs: Please look closer.

src/rpc/methods/eth.rs
LesnyRumcajs: It doesn't avoid any iteration, just some trivial lookups. Regardless, good change, if anything then just for iterator combo.

src/rpc/methods/eth.rs
akaladarshi: I might be missing something here the whole suggestion here was.

```rust
    if executed_messages
        .iter()
        .any(|em| em.receipt.events_root().is_some())
    {
        ensure_events_available(ctx, ts_ref.epoch())?;
    }

```

Anyways this was just a suggestion to avoid writing same condition for error multiple times.

LesnyRumcajs: Do you see why it's objectively worse from performance POV?

LesnyRumcajs:
> @LesnyRumcajs Also if possible can we also include some tests to make sure the indexing of transaction is correct in `collect_events_from_messages` (`transaction_index == msg_idx == executed_message`) because we rely on it a lot.
> 
> I mean it is correct we already know that but just adding a test if possible might make it regression free.

Yeah, it's a good invariant to lock.

akaladarshi (approved):

PR #7502, opened 2026-08-18
fix(rpc): more fixes in filecoin chain notify API

### LINKED ISSUE #5795: Test subscription-based APIs
# Summary

We want to validate that subscription APIs are working correctly.

Right now, our api-compare tool isn’t sufficient to properly test these APIs. We would like to create simple integration scripts that run WebSocket clients connecting directly to the node.

## Filecoin.ChainNotify

First, we want to test that the channel mechanism itself is functioning correctly.
We can use the existing data-flow diagram as a reference to test different message payloads for creating and closing channels.
It’s very important to ensure that multiple channels can be active simultaneously.

Then, we need to test the different types of notifications that `ChainNotify` can emit:
- "current": the current head tipset
- "apply": a new tipset added to the chain
- "revert": a chain reorganization occurred, and a tipset was reverted

To validate the responses:
- Mine new blocks using the devnet/miner.
- Force a reorg by stopping one miner and modifying weight rules.
- Simulate null rounds by briefly disabling mining.

We can start by writing integration tests on calibnet to cover channel creation/closing.

For the specific types of notifications, we can choose the most practical testing approach. Same for the scripting language (hint: Ruby is a good pick).

## eth_subscribe / eth_unsubscribe

Similarly, we want to test opening and closing subscriptions, and then focus on testing the different event types:
- "newHeads"
- "newPendingTransactions"
- "logs" (with and without filter parameters)

We need to ensure that performance scales well with the number of WebSocket clients (around 100s).

## ✅ Note

We should ensure that #5793 is completed first, as we might want to leverage the internal `EthEventHandler` type to factor out and reuse filtering logic across these endpoints.

Also, these scripts should ideally be written to also work against Lotus, the reference implementation.
This way, we can first validate correctness on Lotus, then run them against Forest to ensure compatibility.

# Completion Criteria

All integration tests and unit tests (if necessary) are passing.

akaladarshi:
## Summary of changes

- Removing the closed subscription from the registry

Partially Closes https://github.com/ChainSafe/forest/issues/5795

## Change checklist

### Outside contributions

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved RPC subscription cleanup when channels close or sources shut down.
  - Reusing a request ID now properly closes the previous subscription.
  - Subscription cancellation works immediately, including before acceptance and when authorization filters are enabled.
  - Enforced per-connection subscription limits.
  - Serialization failures no longer stop later notifications from being delivered.
  - Improved handling of lagged subscriptions and added warnings for slow chain-notification subscribers.

- **Documentation**
  - Updated the unreleased changelog with RPC subscription fixes.

### DIFF
--- a/src/rpc/channel.rs
+++ b/src/rpc/channel.rs
@@ -65,7 +65,7 @@ use jsonrpsee::{
     ConnectionId, MethodResponse, MethodSink,
     server::{
         IntoSubscriptionCloseResponse, MethodCallback, Methods, RegisterMethodError,
-        ResponsePayload,
+        ResponsePayload, SubscriptionPermit,
     },
     types::{ErrorObjectOwned, Id, Params, error::ErrorCode},
 };
@@ -83,34 +83,61 @@ pub const CANCEL_METHOD_NAME: &str = "xrpc.cancel";
 
 pub type ChannelId = u64;
 
-/// Type-alias for subscribers.
+/// Type-alias for subscribers: request id -> (connection sink, unsubscribe
+/// receiver, channel id).
+///
+/// Dropping the receiver is what closes the channel — it fires the matching
+/// [`IsUnsubscribed`] sender, which wakes the channel's pump.
 pub type Subscribers =
     Arc<Mutex<HashMap<(ConnectionId, Id<'static>), (MethodSink, mpsc::Receiver<()>, ChannelId)>>>;
 
-/// Represents a single subscription that is waiting to be accepted or rejected.
-///
-/// If this is dropped without calling `PendingSubscription::reject` or `PendingSubscriptionSink::accept`
-/// a default error is sent out as response to the subscription call.
-///
-/// Thus, if you want a customized error message then `PendingSubscription::reject` must be called.
+/// Owns one channel's entry in the subscriber registry and removes it on drop,
+/// unless a newer channel has since reused the same request id.
 #[derive(Debug)]
-#[must_use = "PendingSubscriptionSink does nothing unless `accept` or `reject` is called"]
+pub(crate) struct ChannelRegistration {
+    subscribers: Subscribers,
+    connection_id: ConnectionId,
+    request_id: Id<'static>,
+    channel_id: ChannelId,
+}
+
+impl Drop for ChannelRegistration {
+    fn drop(&mut self) {
+        let key = (self.connection_id, self.request_id.clone());
+        let mut subscribers = self.subscribers.lock();
+        if subscribers
+            .get(&key)
+            .is_some_and(|(_, _, channel_id)| *channel_id == self.channel_id)
+        {
+            subscribers.remove(&key);
+        }
+    }
+}
+
+/// Represents a single subscription that is waiting to be accepted.
+///
+/// Dropping this without calling [`PendingSubscriptionSink::accept`] answers the
+/// subscribe call with an internal error and removes the channel's registry
+/// entry.
+#[derive(Debug)]
+#[must_use = "PendingSubscriptionSink does nothing unless `accept` is called"]
 pub struct PendingSubscriptionSink {
     /// Sink.
     pub(crate) inner: MethodSink,
     /// `MethodCallback`.
     pub(crate) method: &'static str,
-    /// Shared Mutex of subscriptions for this method.
-    pub(crate) subscribers: Subscribers,
+    /// Owns this channel's registry entry until the pump exits.
+    pub(crate) registration: ChannelRegistration,
     /// ID of the `subscription call` (i.e. not the same as subscription id) which is used
     /// to reply to subscription method call and must only be used once.
     pub(crate) id: Id<'static>,
     /// Sender to answer the subscribe call.
     pub(crate) subscribe: oneshot::Sender<MethodResponse>,
-    /// Channel identifier.
-    pub(crate) channel_id: ChannelId,
-    /// Connection identifier.
-    pub(crate) connection_id: ConnectionId,
+    /// Fires once this channel's registry entry is dropped.
+    pub(crate) unsubscribe: IsUnsubscribed,
+    /// Held until the pump exits; dropping it early would hand the slot back
+    /// and make `max_subscriptions_per_connection` a no-op for channels.
+    pub(crate) _permit: SubscriptionPermit,
 }
 
 impl PendingSubscriptionSink {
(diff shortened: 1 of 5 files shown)


src/rpc/channel.rs
```diff
@@ -121,13 +166,16 @@ impl PendingSubscriptionSink {
...
             self.id,
             ResponsePayload::success_borrowed(&channel_id),
             self.inner.max_response_size() as usize,
         );
-        let success = response.is_success();
+        if !response.is_success() {
+            panic!(
+                "The subscription response was too big; adjust the `max_response_size` or change Subscription ID generation"
+            );
+        }
```
LesnyRumcajs, line 160: I understand this is an existing behavior, but crashing the node when the response is a bit too big is a bit dramatic, isn't it?

LesnyRumcajs (commented):
I'm not really an expert in the JSON-RPC subscription mechanism and WS, so I employed AI to check some things, especially compatibility with `go-jsonrpc` (which majority of the Filecoin ecosystem uses). Please check.


```
Headline: xrpc.cancel still does not work for any go-jsonrpc client, three independent reasons

(a) id-less form never dispatched. go-jsonrpc builds the cancel with ID unset (client.go:379-384, websocket.go:381-385) and request.ID is json:"id,omitempty" (handler.go:56), so the wire frame is a JSON-RPC notification. jsonrpsee RpcService::notification (jsonrpsee-server-0.26.0/src/middleware/rpc.rs:190-195) returns MethodResponse::notification() and never reaches a MethodCallback. No Forest layer converts notifications to calls (src/rpc/mod.rs:665-672). The module doc admits this ("the id-less notification form is currently ignored") while the CHANGELOG reads as if cancel now works. The filter exemption and the register-before-accept race fix are both unreachable for real clients: the dispatch gap, not the filter list, is what blocks cancels.

(b) the reply is the wrong frame kind. close_channel_response (channel.rs:311) emits {"jsonrpc":"2.0","id":null,"result":{"jsonrpc":"2.0","method":"xrpc.ch.close","params":[N]}}, a response with a notification-shaped object nested in result. go-jsonrpc's handleFrame switches on frame.Method; that frame has none, so it lands in handleResponse, never handleChanClose. The client's chanHandlers[N] entry and its Go channel are never released for the life of the websocket. Lotus always sends the bare notification, including after a cancel: cancelling the handler ctx closes the handler's channel, and the select loop then writes request{ID: nil, // notification, Method: chClose} (websocket.go:303-315). Forest's pump instead breaks on the sink.closed() arm and sends nothing. The PR's new "Close path (a)" diagram documents this divergence as if it were the protocol.

(c) id convention is inconsistent. Success uses Id::Null, the error path echoes the request id (channel.rs:379). A spec-compliant client keying pending calls by id never resolves the success case. Tracked as #4453, but this PR rewrote exactly that doc block and added assertions locking it in.

Fixing (b) is cheap and is the one that actually unbreaks Lotus/Curio teardown: send the bare xrpc.ch.close from the pump's unsubscribe arm. (a) needs a notification hook in a layer.

Medium

Reject list also bypassed. filter_layer.rs:50: method_name == CANCEL_METHOD_NAME || self.filter_list.authorize(..) short-circuits before the reject half. FilterList::authorize is "allow-empty-or-matched AND not rejected", with substring matching (filter_list.rs:22-26), so !xrpc.cancel (and even !xrpc) is silently ignored, no log. Keep the reject check in the condition; there is no is_rejected accessor yet, so one has to be added.

Serialization-failure skip contradicts the #5795 sibling entry. channel.rs:429: a batch that fails to serialize is now dropped and the channel stays open, 12 lines above a RecvError::Lagged arm that closes the channel specifically to avoid undetectable gaps, and one CHANGELOG bullet above "Filecoin.ChainNotify now closes the subscription channel ... instead of silently dropping head changes". go-jsonrpc's continue (websocket.go:320-323) is faithful parity, but for ChainNotify a lost apply/revert batch is exactly the failure #5795 set out to kill. A persistently unserializable payload also produces one tracing::error! per head change forever. Was skip-vs-close a deliberate choice for ChainNotify, or just parity copied from a different payload class? Also missing CHANGELOG entries for this flip, for 10 -> 16, and for the new slow-subscriber warning.

Cancel-after-server-close now errors. Previously the registry entry outlived the pump, so a client tearing down after xrpc.ch.close got a success. Now ChannelRegistration::drop removes it and the same cancel gets channel not found, codified by the new source_closed_sends_bare_close assertion. go-jsonrpc's cancelCtx (websocket.go:416-441) ignores an unknown id silently and never responds. Unconditional-cancel-on-teardown is a normal pattern, so this surfaces a new spurious error.

Low

- Displaced close races its own pump (channel.rs:512). insert drops the old unsubscribe receiver and a detached send_close runs concurrently with the old pump, whose tokio::select! picks randomly among ready arms. With queued broadcast messages, xrpc.ch.val for channel N can be written after xrpc.ch.close for N on the same sink; go-jsonrpc has already deleted the handler and logs handler N not found per stray frame. Send the close from the pump's unsubscribe arm instead.
- Double close window (channel.rs:504). Pump does send_close, break, debug!, then drops the registration. A same-id subscribe in that window displaces the stale entry and spawns a second send_close for the already-dead channel id.
- Forwarder task outlives the pump (chain.rs:1739). It parks on head_changes_rx.recv_async() and only notices the dead receiver on the next sender.send. On a stalled or header-syncing node, every connect/disconnect cycle accumulates a task plus a flume receiver the publisher must fan out to. Pre-existing, but the CHANGELOG says "no longer leak node resources".
- Duplicate mechanism (chain.rs:61). subscribe_head_changes_bounded exists for precisely "consumers driven by an untrusted RPC client's read rate" (chain_store.rs:349-355) and eth newHeads uses it (pubsub.rs:119). ChainNotify keeps the unbounded subscription and stacks a second bespoke scheme on top (16-slot broadcast, lag close, hand-rolled len() > 5 warn). Two things to keep in sync, and the bounded of the new warning.
- Permit fix untested (channel.rs:158). Methods::raw_json_request injects mock_subscription_permit() (jsonrpsee-core-0.26.0/src/server/rpc_module.rs:358), so nothing in the module asserts the max_subscriptions_per_connection claim. Only _permit plus a comment stand between a future cleanup and silently reverting it.
- PendingSubscriptionSink.id duplicates registration.request_id (lines 499 and 522 both clone the same Id). One source of truth would do.
- ChannelRegistration::drop does get then remove (two hashes; entry/Occupied does it in one) and drops the removed tuple while ke the subscribe path (504) and cancel handler (360) which both release first.
- v1.36.2 lotus link is likely dead (chain.rs:60). No such tag in the local clone (latest v1.36.1), and the repo pins v1.36.0 / release/v1.36.1 elsewhere. Content is right: at v1.36.1, store.go:295 is make(chan []*api.HeadChange, 16), warn at :329 with > 5 and byte-identical text.
- CHANGELOG: stray blank line inserted between the two ### Added bullets; five separate #5795 bullets could collapse.
- SLOW_SUBSCRIBER_BACKLOG doc: "forest" -> "Forest".

Style, your personal rules rather than repo rules, so raise only if you want to: the new comments include several 6-7 line narrative paragraphs that explain the alternative and the decision (lines 489-494, 505-511), the _permit rationale is duplicated verbatim at 156-158 and 232-234, and four added comment lines use em-dashes (78, 107, 739, 867).
```

PR #7529, opened 2026-08-21
chore(perf): optimize `eth_feeHistory`

LesnyRumcajs:
## Summary of changes

- avoid grabbing event logs for fee history (it's not needed - we only need receipts)
- avoid wasteful `load_child_tipset` - the maximum block walk we allow for this method is 1024, so there would be 1023 completely wasteful blocking tasks just to get message receipt, which we already have, given we walk from newest to oldest.
- this nicely improves the methods questionable performance (especially around P99 - ~500 ms), while doing less work. A naive alternative was to parallelize the entire walk, but it'd just increase resource usage and possibly exhaust blocking task pool.

Closes

```
┌────────────┬──────────────────────┬────────────────────┬───────┐
│ blockCount │ baseline (no-thread) │     threading      │  win  │
├────────────┼──────────────────────┼────────────────────┼───────┤
│ 500        │ 0.221s median        │ 0.024s             │ ~9.2× │
├────────────┼──────────────────────┼────────────────────┼───────┤
│ 1024       │ 0.441s (0.587 max)   │ 0.051s (0.215 max) │ ~8.6× │
└────────────┴──────────────────────┴────────────────────┴───────┘
```

We might be doing similar silly thing elsewhere. To be investigated.

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **Performance**
  * Improved fee history retrieval by using available message receipts more efficiently.
  * Reduced unnecessary processing when retrieving contract bytecode and calculating gas reward data.

* **Reliability**
  * Added fallback handling when receipt data is unavailable.
  * Added validation to ensure messages and receipts remain correctly ordered and matched.
  * Improved consistency when retrieving receipts across related tipsets.

### DIFF
--- a/src/state_manager/mod.rs
+++ b/src/state_manager/mod.rs
@@ -92,7 +92,8 @@ impl GetSize for ExecutedMessage {
     }
 }
 
-/// A tipset's messages paired with their execution receipts.
+/// A tipset's messages paired with their execution receipts. The variants are an internal detail of
+/// how the pairs were loaded; consumers iterate them via [`Self::iter`].
 pub enum TipsetMessageReceipts {
     /// Backed by a full executed tipset (a cache hit, or the fallback loader).
     Executed(Arc<Vec<ExecutedMessage>>),
--- a/src/state_manager/mod.rs
+++ b/src/state_manager/mod.rs
@@ -54,6 +54,7 @@ use crate::utils::cache::SizeTrackingCache;
 use crate::utils::get_size::GetSize;
 use anyhow::Context as _;
 use chain_rand::ChainRand;
+use itertools::Either;
 use nonzero_ext::nonzero;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -88,6 +89,26 @@ impl GetSize for ExecutedMessage {
                 + self.events.get_heap_size_with_tracker(&mut tracker).0,
             tracker,
         )
+    }
+}
+
+/// A tipset's messages paired with their execution receipts.
+pub enum TipsetMessageReceipts {
+    /// Backed by a full executed tipset (a cache hit, or the fallback loader).
+    Executed(Arc<Vec<ExecutedMessage>>),
+    /// Receipts read directly from the child tipset, paired with the tipset's messages (no events).
+    Stored(Arc<Vec<ChainMessage>>, Vec<Receipt>),
+}
+
+impl TipsetMessageReceipts {
+    /// Iterates the (message, receipt) pairs in index order, by reference.
+    pub fn iter(&self) -> impl Iterator<Item = (&ChainMessage, &Receipt)> + '_ {
+        match self {
+            Self::Executed(messages) => {
+                Either::Left(messages.iter().map(|em| (&em.message, &em.receipt)))
+            }
+            Self::Stored(messages, receipts) => Either::Right(messages.iter().zip(receipts.iter())),
+        }
     }
 }
 
--- a/src/state_manager/state_computation.rs
+++ b/src/state_manager/state_computation.rs
@@ -75,6 +75,45 @@ impl StateManager {
             .await
     }
 
+    /// Returns `ts`'s messages paired with their execution receipts, without loading events.
+    /// `receipt_root` is `ts`'s message-receipt root (the child tipset's `parent_message_receipts`)
+    /// when the caller already knows it, avoiding a `load_child_tipset` lookup; `None` resolves it.
+    pub async fn tipset_message_receipts(
+        &self,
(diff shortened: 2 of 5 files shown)


src/state_manager/state_computation.rs
```diff
@@ -75,6 +75,45 @@ impl StateManager {
...
+                messages.len(),
+                receipts.len()
+            );
+            return Ok(TipsetMessageReceipts::Stored(messages, receipts));
+        }
+        Ok(TipsetMessageReceipts::Executed(
+            self.load_executed_tipset_for_rpc(ts)
+                .await?
+                .executed_messages,
+        ))
```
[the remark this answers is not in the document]
LesnyRumcajs: no, it's dense by construction

src/state_manager/state_computation.rs
```diff
@@ -75,6 +75,45 @@ impl StateManager {
...
+            && let Ok(receipts) = Receipt::get_receipts(self.cs.db(), root)
+        {
+            let messages = self.chain_store().messages_for_tipset(ts)?;
+            anyhow::ensure!(
+                messages.len() == receipts.len(),
+                "mismatching message and receipt counts ({} messages, {} receipts)",
+                messages.len(),
+                receipts.len()
+            );
+            return Ok(TipsetMessageReceipts::Stored(messages, receipts));
```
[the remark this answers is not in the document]
LesnyRumcajs: no, it's dense by construction

src/state_manager/state_computation.rs
```diff
@@ -75,6 +75,45 @@ impl StateManager {
             .await
     }
 
+    /// Returns `ts`'s messages paired with their execution receipts, without loading events.
+    /// `receipt_root` is `ts`'s message-receipt root (the child tipset's `parent_message_receipts`)
+    /// when the caller already knows it, avoiding a `load_child_tipset` lookup; `None` resolves it.
+    pub async fn tipset_message_receipts(
+        &self,
+        ts: &Tipset,
+        receipt_root: Option<Cid>,
```
akaladarshi: Small suggestion: If we pass the `receipt_root` as the child tipset we should be able to do, to avoid fork.

```Rust=
anyhow::ensure!(
       msg_ts.key() == receipt_ts.parents(),
       "message tipset should be the parent of message receipt tipset"
);
```

This is specific pattern i have seen in many places to avoid forks, but can be skipped since we are getting the child state root directly from the parent tipset but still worth considering.

src/state_manager/mod.rs
```diff
@@ -91,6 +92,26 @@ impl GetSize for ExecutedMessage {
     }
 }
 
+/// A tipset's messages paired with their execution receipts.
+pub enum TipsetMessageReceipts {
+    /// Backed by a full executed tipset (a cache hit, or the fallback loader).
+    Executed(Arc<Vec<ExecutedMessage>>),
```
akaladarshi, line 99: Not sure if these names are really worth having. I mean does it really matter if the messages and receipts are `Executed` or loaded (`Stored`) to the outside consumer?

It creates kind of confusion since they are public, either make them private or just name something else.

akaladarshi (commented):
I think we can add more test directly testing the `tipset_message_receipts` using the `chain4u!` macro.

src/state_manager/mod.rs
LesnyRumcajs: I don't really get your concern. The names are needed to somehow distinguish between different variants. We want to have the variant to easily iterate the messages and receipts without headaches with borrowing. The enum needs to be public to be a returned type. We could have another, inner enum type inside for total encapsulation but I think that's over-engineering.

akaladarshi: My only concern here is that, the naming suggest that consumer of the enum should care if message is `Executed` or `Stored` which is not the case, these variant are only there to show if message and receipt parallely accessible or fetched from the `ExecutedMessage`.

Nothing pressing, was just stating the confusion i got while reading it, nothing else.

LesnyRumcajs: How about I throw a comment that those variants are an internal thing so the reader stops reading? :)

src/state_manager/state_computation.rs
LesnyRumcajs: Hm, yeah, it makes sense.

LesnyRumcajs:
> I think we can add more test directly testing the `tipset_message_receipts` using the `chain4u!` macro.

@akaladarshi I opted out of `chain4u` due to some of its limitations that'd bloat the test. I think it's more understandable without doing the autowiring.

I also moved some test utils out of "prod" modules so they don't pollute it.

akaladarshi (approved):

PR #7532, opened 2026-08-23
fix:(drand): `SizeTrackingCache` cache name per drand network name

EclesioMeloJunior:
## Summary of changes

- While adding metrics to track the Drand HTTP relay usage noticed that the field `verified_beacons` is always initialized with `SizeTrackingCache::new_with_metrics("verified_beacons", CACHE_SIZE)`, so any instance will create/export the metrics using that `verified_beacons`. 
- The problem is that we call `DrandBeacon::new`  for every point and for mainnet there are 3 points and for calibnet we have 2 points, and that expose the metrics 3 times with the same naming
- This PR solves that by appending the `config.network_type` to the cache name which avoids the metrics naming conflict

N/A

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate `cache_verified_beacons_*` metrics caused by multiple drand schedule points.
  * Monitoring now reports these metrics only for the active beacon, preventing duplicate metric families.
  * Historical beacons continue to cache verified entries without generating redundant metrics.

* **Documentation**
  * Clarified the changelog entry describing the metric behavior and caching changes.


src/beacon/drand.rs
```diff
@@ -271,7 +272,10 @@ impl DrandBeacon {
             drand_gen_time: config.chain_info.genesis_time as u64,
             fil_round_time: interval,
             fil_gen_time: genesis_ts,
-            verified_beacons: SizeTrackingCache::new_with_metrics("verified_beacons", CACHE_SIZE),
+            verified_beacons: SizeTrackingCache::new_with_metrics(
```
LesnyRumcajs: While it's technically correct, it's kind of not pragmatic. Incentinet and mainnet are a thing of the past and we should expect it to always show zero, with the exception of someone serving some really old archival data.

Prometheus metrics are scraped every X seconds, then persisted. We should keep only metrics useful to operators and ourselves.

LesnyRumcajs (commented):
What happens if someone turns on caching for another drand network variant?

src/beacon/drand.rs
```diff
@@ -271,12 +274,19 @@ impl DrandBeacon {
             drand_gen_time: config.chain_info.genesis_time as u64,
             fil_round_time: interval,
             fil_gen_time: genesis_ts,
-            verified_beacons: SizeTrackingCache::new_with_metrics("verified_beacons", CACHE_SIZE),
+            verified_beacons: config.use_cache.then(|| {
+                SizeTrackingCache::new_with_metrics(
+                    format!("verified_beacons_{}", config.network_type),
+                    CACHE_SIZE,
+                )
+            }),
```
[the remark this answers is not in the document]
EclesioMeloJunior: @LesnyRumcajs should we place any check to avoid the same variant be present in the points array more than once?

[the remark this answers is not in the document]
LesnyRumcajs: This wouldn't mitigate the issue in practice; metrics can be registered outside of this `SizeTrackingCache`, so a sanity check that we're not adding a duplicate metric should be on the highest level possible.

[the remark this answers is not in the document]
LesnyRumcajs: But I'd say it's outside of the scope of this PR and should be done separately.

src/beacon/drand.rs
```diff
@@ -252,8 +255,8 @@ pub struct DrandBeacon {
     fil_gen_time: u64,
     fil_round_time: u64,
 
-    /// Keeps track of verified beacon entries.
-    verified_beacons: SizeTrackingCache<u64, Arc<BeaconEntry>>,
+    /// Keeps track of verified beacon entries. `None` when `use_cache` is unset.
+    verified_beacons: Option<SizeTrackingCache<u64, Arc<BeaconEntry>>>,
```
LesnyRumcajs: I think we kind of misunderstood each other. The cache should still be there, we just shouldn't care about metrics around it.

src/beacon/drand.rs
```diff
@@ -271,7 +271,11 @@ impl DrandBeacon {
             drand_gen_time: config.chain_info.genesis_time as u64,
             fil_round_time: interval,
             fil_gen_time: genesis_ts,
-            verified_beacons: SizeTrackingCache::new_with_metrics("verified_beacons", CACHE_SIZE),
+            verified_beacons: if config.network_type.is_unchained() {
```
LesnyRumcajs: This is still a brittle condition. The exactly same bug would be necromanced if tomorrow someone introduces `lightspeednet` that is also unchained by chance, e.g., `static LIGHTSPEEDNET: LazyLock<DrandBeacon> = LazyLock::new(new_beacon_lightspeednet);` - the metric would be duplicated.

IMO a design that wouldn't be brittle is:
1. an additional parameter to the c-tor, either a positional one or via the `DrandConfig` if its serialized form isn't somehow important. This forces the instance creator to decide whether the metrics should be enabled or not.
2. The `verified_beacons` cache should have the name in it. I guess it should be fine to put the `DrandNetwork` value there and not expect another network with exactly the same name to be created.

An extra step (outside of this PR) would be to add a global hashset of identifiers and check if a duplicate is not getting introduced; the memory overhead should be minimal but at least we could get warnings if something like this happens.

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@
 
 - [#7537](https://github.com/ChainSafe/forest/pull/7537): Self-destructed EVM contracts no longer report stale state in the trace RPCs. `Forest.EthTraceCall` (`trace_call`) `stateDiff` and `Forest.EthDebugTraceTransaction` (`debug_traceTransaction`) `prestateTracer` now report a zero nonce and empty storage for them, matching `Filecoin.EthGetTransactionCount`, `Filecoin.EthGetCode`, `Filecoin.EthGetStorageAt` and Lotus. Previously a contract that self-destructed during the traced message showed no storage change at all.
 
+- [#7414](https://github.com/ChainSafe/forest/issues/7414): Forest no longer exports duplicate `cache_verified_beacons_*` metric families. Every drand point in a network's schedule registered its own collector under that one name (three of them on mainnet), which most Prometheus parsers reject. Only the active drand beacon reports these metrics now; the historical ones still cache, they are just not observed.
```
LesnyRumcajs:
```suggestion
- [#7414](https://github.com/ChainSafe/forest/issues/7414): Forest no longer exports duplicate `cache_verified_beacons_*` metric families.
```

Let's skip the narrative. The first part is questionable (I don't recall scrapers rejecting our metrics, at most silently ignoring some of them) . The second is an internal detail.

LesnyRumcajs (approved):

Issue #7414, opened 2026-07-24
fetch beacon entries from gossipsub


LesnyRumcajs:
# Summary

See spike at https://github.com/ChainSafe/forest/issues/7364

Right now, when we need beacon entries, we query the https relay; we could avoid our reliance on that by using the dedicated gossipsub topic and feeding it into a cache.

We might also that would somehow resubscribe to the topic in case we don't get queries for longer period. 

# Completion Criteria

- [ ] Feed beacon entries from gossipsub to the drand cache.
- [ ] Test that it works correctly, e.g., drop drand https relays and ensure recent queries to `MinerGetBaseInfo` work fine without it.
- [ ] Watchdog - if we don't get updates, we must resubscribe (workaround)
- [ ] Metrics around the new drand entries cache
- [ ] Metrics around the HTTP drand usage

# Additional Links & Resources

yashksaini-coder:
@LesnyRumcajs Hi I would like to take on this issue. And is there is a dedicated communication platform you use, like discord, or slack which I can join to keep sharing updates or ask queries ?

LesnyRumcajs:
@yashksaini-coder Thanks for your interest, I'd rather this to be done internally. For future, you can check out docs for channels we're hanging around.

PR #7533, opened 2026-08-23
feat(drand): include drand metrics around HTTP retrieval

EclesioMeloJunior:
## Summary of changes

- Include HTTP metrics around Drand entropy retrieval

- Relates to https://github.com/ChainSafe/forest/pull/7532
- Relates to https://github.com/ChainSafe/forest/issues/7414 (Metrics around the HTTP drand usage)

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **New Features**
  * Added a metric counting drand rounds fetched over HTTP rather than served from the in-memory cache.
  * The metric tracks each HTTP fetch attempt after a verified-cache miss, including retries and fallback servers.

* **Documentation**
  * Updated the unreleased changelog with details about the drand HTTP fetch metric.


src/beacon/metrics.rs
LesnyRumcajs, line 1: My point is kind of similar to the one in https://github.com/ChainSafe/forest/pull/7532/changes#r3842118238

We must not bloat the metrics. I see value only in the number of HTTP calls given we want to limit them. Errors, duration... this is somehow a debug information.

src/beacon/tests/drand.rs
```diff
@@ -308,3 +308,25 @@ async fn beacon_entries_for_block_covers_null_rounds_quicknet() {
         );
     }
 }
+
+/// The counter tracks rounds that had to leave the process: a cache miss increments it,
+/// a hit does not. Compared as a delta so the test does not depend on what else ran.
+#[tokio::test]
+async fn http_fetch_metric_counts_only_cache_misses() {
```
LesnyRumcajs: Is there a data race in this test?

EclesioMeloJunior: removed the test as it was relying on a global `DRAND_HTTP_FETCH_TOTAL` that can be affected by other tests, also the cache is being exercised by other tests

src/beacon/drand.rs
```diff
@@ -426,6 +427,7 @@ impl Beacon for DrandBeacon {
             entry.round()
         );
         self.cache_fetched_entry(&entry);
+        metrics::DRAND_HTTP_FETCH_TOTAL.inc();
```
LesnyRumcajs: The placement of this metric is incorrect. Why?

EclesioMeloJunior: sure, we should count when cache miss we needed to go through http, so I am going to move it right after the cache miss

src/beacon/drand.rs
```diff
@@ -374,6 +375,9 @@ impl Beacon for DrandBeacon {
             return Ok(Arc::unwrap_or_clone(cached_entry));
         }
 
+        // counting when cache miss and needed to go fetch via HTTP
+        metrics::DRAND_HTTP_FETCH_TOTAL.inc();
```
LesnyRumcajs: The placement of this metric is not ideal; it's correct-ish (it'd still miss hits in case of, e.g., retries due to bad signature) but only if you analyze the logic in the following lines. This wouldn't really be need if you'd put right next to the place where the fetch is actually happening; right now below line 390.

LesnyRumcajs (approved):

PR #7545, opened 2026-08-26
test(rpc): cover StateMinerCreationDeposit initial-pledge calc

### LINKED ISSUE #7503: Add test coverage for StateMinerCreationDeposit network activation and calculation
## Summary

Add focused test coverage for `Filecoin.StateMinerCreationDeposit` and the related initial-pledge refactoring.

## Rationale

The RPC calculation is required for Curio miner integration. Existing API parity and snapshot tests verify the RPC wiring. Add deterministic calculation tests to verify network-version activation and network-specific behavior.

## Affected areas

- `src/rpc/methods/state.rs`
- RPC/API comparison tests in `src/tool/subcommands/api_cmd/`
- Test fixtures or vectors used by builtin-actors or Lotus, if compatible and available

## Required work

- Add tests for the pre-V27 behavior, where `Filecoin.StateMinerCreationDeposit` returns zero.
- Add tests at and after the V27 activation epoch.
- Cover mainnet and calibnet network configuration and their applicable activation epochs.
- Verify the creation-deposit result derives from one-tenth of minimum consensus power.
- Prefer authoritative vectors or fixtures from Lotus or builtin-actors when they can be imported or reproduced without changing production behavior.
- Keep the current API parity and snapshot tests. Add unit or integration tests that validate calculation inputs and outputs directly.

## Acceptance criteria

- The test suite covers pre-activation, activation-boundary, and post-activation cases.
- The suite covers both mainnet and calibnet configuration paths.
- Expected token amounts are asserted from deterministic fixtures or authoritative vectors.
- If Lotus or builtin-actors vectors cannot be used, document the reason in the test or issue discussion and use equivalent deterministic fixtures.

## Backlinks

- Originating PR: https://github.com/ChainSafe/forest/pull/7495
- Requesting comment: https://github.com/ChainSafe/forest/pull/7495#issuecomment-5329314817
- Requested by: @LesnyRumcajs

0xDevNinja:
## Summary of changes

Adds synthetic (snapshot-free) test coverage for `Filecoin.StateMinerCreationDeposit` and the extracted `compute_initial_pledge_for_power`, as requested in #7503. Tests are deterministic and network-free, per the guidance to keep them synthetic rather than snapshot-based.

- New `creation_deposit_tests` module in `src/rpc/methods/state.rs`:
  - `creation_deposit_is_zero_before_v27` — the handler returns a zero deposit before network version 27 without reading state.
  - `creation_deposit_positive_at_v27` — the handler computes a positive deposit at v27 from hand-built actor state.
  - `compute_initial_pledge_with_active_ramp` / `compute_initial_pledge_without_ramp` — cover both branches of the pledge-ramp selection (`ramp_start_epoch > 0` and `== 0`).
- The tests build a state tree with the power, reward, burnt-funds and reserve actors that the calculation and the circulating-supply computation read, using the real v18 actor code CIDs from the embedded bundle metadata (`ACTOR_BUNDLES_METADATA`).
- Small enabling refactor: `compute_initial_pledge_for_power` now takes `&StateManager` instead of `&Ctx`, so the calculation can be exercised without constructing a full RPC context. The two call sites pass `&ctx.state_manager`; no behaviour change.

Between the four tests, `compute_initial_pledge_for_power` and the `StateMinerCreationDeposit` handler (both the pre-v27 gate and the v27 path) are fully exercised.

Closes #7503

Test-only plus a no-op signature refactor, so no CHANGELOG entry.

## Change checklist

### Outside contributions

- [x] This pull request is based on an issue that a maintainer has accepted (see [Before Opening a Pull Request][4]).
- [x] I have read and agree to the [CONTRIBUTING][2] document.
- [x] I have read and agree to the [AI Policy][3] document. I understand that failure to comply with the guidelines will lead to rejection of the pull request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initial pledge calculations across network conditions, including active and inactive ramp periods.
  * Corrected miner creation deposit calculations for network versions before and at network version 27.
  * Improved calculation reliability by using current network state, circulating supply, rewards, and network parameters.
  * Added clearer error reporting when required state information cannot be read.


src/rpc/methods/state.rs
```diff
@@ -3565,3 +3566,260 @@ mod tests {
...
+    use cid::Cid;
+    use fvm_ipld_blockstore::Blockstore;
+    use std::sync::Arc;
+
+    // 1 FIL in atto.
+    const FIL: i128 = 1_000_000_000_000_000_000;
+
+    fn filter_estimate(
+        position: i128,
+    ) -> fil_actors_shared::v18::builtin::reward::smooth::FilterEstimate {
```
LesnyRumcajs: Let's not use specific versions of actors/fvm directly. Ideally, only shims would would deal with those.

src/rpc/methods/state.rs
```diff
@@ -3565,3 +3566,260 @@ mod tests {
...
+        let reward_state = reward::State::default_latest_version(
+            StoragePower::from(0),
+            StoragePower::from(0),
+            0,
+            StoragePower::from(1u64 << 50),
+            fvm_shared4::econ::TokenAmount::from_atto(0),
+            filter_estimate(smoothed),      // this_epoch_reward_smoothed
+            StoragePower::from(1u64 << 50), // this_epoch_baseline_power (> 0, avoids div-by-zero)
+            0,
+            fvm_shared4::econ::TokenAmount::from_atto(1_000_000 * FIL), // total_storage_power_reward -> fil_mined
```
LesnyRumcajs: You can just use `from_whole`.

LesnyRumcajs (changes requested):
Thanks. Let's move it to a separate file, e.g., `state_tests.rs` as to not bloat this massive file any further.

Let's also address all the CodeRabbit issues and make sure the CI is passing.

EclesioMeloJunior:
@0xDevNinja any updates regarding the comments/suggestions?

src/rpc/methods/state.rs
0xDevNinja: Moved everything onto the shim now. All the token amounts go through `TokenAmount::from_whole` and convert at the builder boundary. The one type I could not avoid naming is the `FilterEstimate`, because `power`/`reward::State::default_latest_version` take the actor native `fil_actors_shared` `FilterEstimate` directly in their signatures, so any caller building inputs for them has to construct that type. Happy to add a shim side constructor for it if you would prefer.

src/rpc/methods/state.rs
0xDevNinja: Done, switched to `from_whole`.

0xDevNinja:
Thanks both, and sorry for the delay. Pushed a revision addressing the review:

- Tests moved to a dedicated `state_tests.rs` so `state.rs` does not grow further.
- Amounts built through the `TokenAmount` shim (`from_whole`) instead of naming fvm/actor versions. The only remaining version specific type is the `FilterEstimate` that `default_latest_version` requires in its signature (noted inline).
- Added `anyhow` context to the state reads in `compute_initial_pledge_for_power`.
- The tests now reach network version 27 through calibnet’s real activation epoch (one past GoldenWeek) and a promoted head, rather than overriding `genesis_network`. The pre v27 case stays on mainnet at genesis.
- Pledge values are pinned and deterministic. The mid ramp case comes out at about half the fully ramped case, so a change that ignored the ramp parameters would fail.

All four tests pass locally with fmt and clippy clean.

EclesioMeloJunior (approved):

src/rpc/methods/state.rs
```diff
@@ -3407,23 +3410,29 @@ impl RpcMethod<1> for StateMinerCreationDeposit {
...
+        .context("loading power actor state for initial pledge")?;
+    let reward_state: reward::State = state_tree
+        .get_actor_state()
+        .context("loading reward actor state for initial pledge")?;
+
+    let circ_supply = state_manager
         .genesis_info()
-        .get_vm_circulating_supply_detailed_with_state_tree(ts.epoch(), &state_tree)?;
+        .get_vm_circulating_supply_detailed_with_state_tree(ts.epoch(), &state_tree)
+        .context("computing circulating supply for initial pledge")?;
```
LesnyRumcajs, line 3435: The error messages mustn't be modified. Adding context makes this

https://github.com/ChainSafe/forest/blob/a80b3f8f6534c777909d2ef3f851e7ea029c0536/src/rpc/error.rs#L117-L118

completely override the error message. So now instead of a `Actor state doesn't exist in store` we get something else (arguably not as helpful).

src/rpc/methods/state_tests.rs
```diff
@@ -0,0 +1,284 @@
...
+use fil_actors_shared::v18::builtin::reward::smooth::FilterEstimate;
+use fvm_ipld_blockstore::Blockstore;
+use num::BigInt;
+use std::str::FromStr as _;
+use std::sync::Arc;
+
+// First calibnet epoch on network version 27 (the deposit calculation): one
+// past the GoldenWeek upgrade height. Testing here keeps the real upgrade
+// schedule rather than overriding the genesis network version.
+const CALIBNET_V27_EPOCH: i64 = 3_007_295;
```
LesnyRumcajs, line 37: Is the activation boundary covered?

src/rpc/methods/state_tests.rs
```diff
@@ -0,0 +1,284 @@
...
+        StoragePower::from(1u128 << 80), // this_epoch_baseline_power
+        0,
+        (&TokenAmount::from_whole(1_000_000)).into(), // total_storage_power_reward -> fil_mined
+        (&TokenAmount::from_whole(0)).into(),
+        (&TokenAmount::from_whole(0)).into(),
+    );
+
+    let meta = ACTOR_BUNDLES_METADATA
+        .values()
+        .find(|m| m.actor_major_version().ok() == Some(18))
```
LesnyRumcajs, line 96: incorrect

src/rpc/methods/state_tests.rs
```diff
@@ -0,0 +1,284 @@
...
+        position: BigInt::from(position),
+        velocity: BigInt::from(0),
+    }
+}
+
+// Builds a state tree with the power, reward, burnt-funds and reserve actors
+// that `compute_initial_pledge_for_power` and the circulating-supply
+// calculation read, and returns its root. `ramp_start_epoch` selects the
+// pledge-ramp branch under test.
+fn build_state_root<S: Blockstore + ShallowClone>(
```
LesnyRumcajs, line 53: Incorrect actor version & bundle handling.

LesnyRumcajs (commented):
@0xDevNinja Thank you for you contribution, but after reviewing your PR I decided this needs to be handled internally. There are too many traps to do it correctly.

Issue #7553, opened 2026-08-26
[NV29] fil-actor-states update and release


LesnyRumcajs:
# Summary

`fil-actor-states` repo is updated to reflect the release of `builtin-actors`. Missing diffs must be applied. Note that initially this might be based off a not-yet-merged branch.

# Completion Criteria

- [ ] All diffs between the last release are addressed
- [ ] Release is cut and consumed by Forest - make sure all enums and delegations are updated. Hopefully, most of them will be enforced by the compiler requiring exhaustive match, but there might be some places where it isn't, e.g., some of our declarative macros where we just list the numbers for actors). To be considered how to automate it if such cases happen.

# Additional Links & Resources

PR #7593, opened 2026-09-03
feat: integrate v19 actor

akaladarshi:
## Summary of changes

- add v19 support for all the actors
- Bump the state version and default latest to v19

Closes Part of https://github.com/ChainSafe/forest/issues/7553

## Change checklist

### Outside contributions

## Summary by CodeRabbit

* **New Features**
  * Added support for Filecoin actors version 19 across actor state handling, RPC methods, runtime policies, and state loading.
  * Extended Lotus JSON conversion for version 19 actor states, parameters, entries, bytecode hashes, and transient data.
  * Added version 19 support for miner sector upgrades, reward streams, distributions, weighting, shares, and claims.
  * Updated market handling for version 19 state fields and deal operations.
* **Bug Fixes**
  * Improved compatibility when deserializing newer reward and market state data.


akaladarshi:
I believe this [pending_allocation](https://github.com/ChainSafe/forest/blob/175196167d3e6cfe307f62d7835473f7b4345728/src/tool/subcommands/api_cmd/api_compare_tests.rs#L1042) test will fail for new version since it has been deprecated.

LesnyRumcajs (commented):
A couple of things, not necessarily to do in this PR but worth considering at least as a separate issue.

1. There's a lot of this `impl_evm_delegate_call_params_lotus_json` dance; if one misses a particular version, is there a compilation error? If not, can we make it so that it's somehow an exhaustive match? It's not a huge deal if it can't be done easily (let's not contrive a monster abstraction to do that), a simple human-based grep can find those. Just something to consider.
2. Are there places where actor version is "leaking"? Ideally, there wouldn't be, e.g., `fil_actor_XYZ_state::v19::` outside of shims. It might be idealistic, of coruse, but each place where it isn't need to have a solid justification why it's not. In some places, it might be sufficient to have a public `latest` export from the shim, so that modifications outside of it are minimal.
3. I see there is still some duplication, e.g., 
```rust
          State::V17(st) => Ok(DealStates::V17(V16AsActorError::context_code(
                V17DealMetaArray::load(&st.states, store),
                FVM4ExitCode::USR_ILLEGAL_STATE,
                "failed to load deal state array",
            )?)),
            State::V18(st) => Ok(DealStates::V18(V18AsActorError::context_code(
                V18DealMetaArray::load(&st.states, store),
                FVM4ExitCode::USR_ILLEGAL_STATE,
                "failed to load deal state array",
            )?)),
            State::V19(st) => Ok(DealStates::V19(V19AsActorError::context_code(
                V19DealMetaArray::load(&st.states, store),
                FVM4ExitCode::USR_ILLEGAL_STATE,
                "failed to load deal state array",
            )?)),
```
(which is also incorrect, see `DealStates::V17(V16AsActorErro`). This could be deduplicated with some macros or perhaps the delegate mechanism from `ambassador`.
4. I see things like `impl TryFrom<&fil_actor_market_state::v19::DealProposal> for DealProposal` that do a bunch of cloning. I'm not sure where we use it, but it might be worth considering having an owned version of it to avoid re-allocations and redundant ops. I might be missing something but I'd assume we mostly want to upgrade one struct to another and then throw the other one away.

Also, stating the obvious, but this PR mustn't be merged until the shape of the v19 actor state is known and a proper `fil-actor-states` is released; we cannot have git references in `Cargo.toml`, as it'd block any (potential emergency) releases

src/shim/actors/builtin/market/ext/state.rs
```diff
@@ -133,6 +133,9 @@ impl MarketStateExt for market::State {
                     Ok(())
                 })?;
             }
+            Self::V19(_) => {
+                anyhow::bail!("unsupported from actors v19");
```
LesnyRumcajs, line 137: Is this the expected behavior? As in, should an RPC method that somewhere along the way invokes `get_allocations_for_pending_deals` always fail from NV29 onwards? What's the behaviour in Lotus?

src/shim/actors/builtin/market/mod.rs
```diff
@@ -703,6 +728,34 @@ impl TryFrom<&fil_actor_market_state::v18::DealProposal> for DealProposal {
     }
 }
 
+impl TryFrom<&fil_actor_market_state::v19::DealProposal> for DealProposal {
```
LesnyRumcajs, line 731: There's lots of this `TryFrom<ref to new version> for Shim` that does a lot of cloning. My naive assumption is that normally we should just consume the original (compiler will tell).

It's an existing issue with the codebase based on my feeling how it's used, not a deep research. To be investigated. @coderabbitai create an issue.

src/shim/actors/builtin/market/ext/state.rs
akaladarshi: Yeah solstice deprecated the FIL+ so i guess we don't need allocations anymore :

https://github.com/filecoin-project/lotus/blob/74da8af2d595eca68153d43f5c610147f988adc5/chain/actors/builtin/market/v19.go#L391

LesnyRumcajs (commented):
LGTM outside of one comment around potential RPC failure.

LesnyRumcajs (approved):

PR #7594, opened 2026-09-03
Drop `--recompute` flag on index backfill

sudo-shashank:
## Summary of changes

- Dropped `--recompute` and always recompute and backfill as much as possible, exiting non-zero on error.

Closes

## Change checklist

### Outside contributions

## Summary by CodeRabbit

- **Changes**
  - Updated `forest-cli index backfill` so `--recompute` defaults to `true` and can be used without specifying a value.
  - Backfills now report when they finish and exit with an error status if tipsets were skipped.

- **Documentation**
  - Updated the CLI reference to describe the new `--recompute` default and usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
 
 ### Changed
 
+- [#7594](https://github.com/ChainSafe/forest/pull/7594): `forest-cli index backfill` now defaults `--recompute` to true and exits with an error if any tipsets were skipped.
+
 ### Removed
 
 ### Fixed
--- a/docs/docs/users/reference/cli.md
+++ b/docs/docs/users/reference/cli.md
@@ -835,8 +835,8 @@ Options:
       --n-tipsets <N_TIPSETS>
           Number of tipsets to back-fill
 
-      --recompute
-          Recompute missing tipset state (expensive) instead of skipping it; tipsets that still can't be computed are skipped and reported rather than aborting the run
+      --recompute [<RECOMPUTE>]
+          Recompute missing tipset state (expensive) instead of skipping it; tipsets that still can't be computed are skipped and reported rather than aborting the run [default: true]
 
       --allow-near-head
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,8 @@
 
 ### Changed
 
+- [#7594](https://github.com/ChainSafe/forest/pull/7594): Dropped `--recompute` from `forest-cli index backfill`; always recompute and backfill as much as possible, exiting non-zero on error.
```
LesnyRumcajs: This is a breaking change; users that were applying the `--recompute` would find their command failing after this.

LesnyRumcajs (commented):
1. We can avoid a breaking change by making the `--recompute` default to `true`. This way, we also allow the consumer to not do it, for any reason, e.g., mentioned sync starving.
2. Does the backfill now fail or not on a missing receipt? I think that was the main issue; it silently failed with return code zero, thus it flew under @EclesioMeloJunior radar.

sudo-shashank:
> 1. We can avoid a breaking change by making the `--recompute` default to `true`. This way, we also allow the consumer to not do it, for any reason, e.g., mentioned sync starving.
> 2. Does the backfill now fail or not on a missing receipt? I think that was the main issue; it silently failed with return code zero, thus it flew under @EclesioMeloJunior radar.

- If we want to allow the consumer to not do it then its better to make `--recompute` default to `true`, will make this change.
- Yes it will fail now with a non-zero exit code.

EclesioMeloJunior:
Currently the external rpc checks are failing but this branch (`shashank/backfill-default-recompute`) does not has the fix that asserts the compute tipset roots.

CHANGELOG.md
sudo-shashank: made `--recompute` default to true

sudo-shashank:
> Currently the external rpc checks are failing but this branch (`shashank/backfill-default-recompute`) does not has the fix that asserts the compute tipset roots.

merged the upstream changes, all checks are passing now

EclesioMeloJunior (approved):

PR #7625, opened 2026-09-15
fix: return eth hash from eth_sendRawTransaction

wjmelements:
Ethereum jsonrpc expects eth_sendRawTransaction to return keccak256 of the raw transaction bytes.
Returning a different hash breaks some tooling, for example `forge script` (https://github.com/filecoin-project/solstice/issues/57).
#### Changes
* return ethereum-style transaction hash from ethereum jsonrpc method eth_sendRawTransaction 
* add test

## Summary by CodeRabbit

- **Bug Fixes**
  - Ethereum raw transaction submissions now return Ethereum transaction hashes instead of Filecoin message CIDs.
  - Submissions report success only after transaction-hash mappings are stored successfully.
  - Errors encountered while saving transaction-hash mappings are now returned instead of ignored.
  - Transaction-hash handling and related error reporting are more reliable.

- **Tests**
  - Expanded test configuration support for validating transaction submission and hash-mapping behavior.


src/rpc/methods/eth.rs
```diff
@@ -3209,9 +3218,18 @@ impl RpcMethod<1> for EthSendRawTransactionUntrusted {
...
         let cid = ctx.mpool.push_untrusted(smsg).await?;
-        Ok(cid.into())
+        // Index immediately; process_signed_messages also indexes it once mined.
+        if let Err(e) =
+            ctx.chain_store()
+                .put_mapping(tx_hash, cid, chrono::Utc::now().timestamp() as u64)
+        {
+            tracing::error!("error inserting eth tx mapping: {e}");
+        }
+        Ok(tx_hash)
```
LesnyRumcajs, line 3238: Kind of unrelated to this PR, but those methods are almost exactly the same. Worth factoring out.

LesnyRumcajs (changes requested):
Thanks for finding this. Real gap. Couple of things need addressing:

1. Test is not testing the fix.
2. (optional) refactor shared logic for deduplication and ease of testing the above
3. changelog entry (user-facing fix)
4. (optional here, if not we can address it on a follow-up) the gap comes from lack of parity checks via https://github.com/ChainSafe/forest/blob/main/src/tool/subcommands/api_cmd/api_compare_tests.rs

src/rpc/methods/eth.rs
```diff
@@ -4577,6 +4595,57 @@ mod test {
...
+        let tx = EthTx::from(tx_args);
+
+        // What EthSendRawTransaction must return.
+        let tx_hash: EthHash = tx.eth_hash().unwrap().into();
+        assert_eq!(
+            &format!("{tx_hash}"),
+            "0x9f2e70d5737c6b798eccea14895893fb48091ab3c59d0fe95508dc7efdae2e5f"
+        );
+
+        // What it returned before the fix.
```
LesnyRumcajs: You fixed the `EthSendRawTransaction*::handle`; this test checks `get_signed_message`; not a regression test for `EthSendRawTransaction`. This will happily pass without the fix.

You might want to factor out 

```rust
       let tx_args = parse_eth_transaction(&raw_tx.0)?;
        // Eth tx hash != Filecoin message CID (see EthHash::to_cid's note).
        let tx_hash: EthHash = tx_args.eth_hash()?.into();
        let smsg = tx_args.get_signed_message(ctx.chain_config().eth_chain_id)?;
```
to a shared function and test against it (thus avoiding the dance with the context)

LesnyRumcajs (commented):
LGTM, as mentioned earlier, let's have a changelog entry for this fix. Thanks!

LesnyRumcajs (approved):
LGTM, thanks!

EclesioMeloJunior (approved):

Issue #7449, opened 2026-08-05
Redundant CBOR encoding of messages during block validation


LesnyRumcajs:
> [!NOTE]
> This was AI found and seems correct, but do your own checks. Regardless of that, a test will be useful.

## Summary

Forest re-encodes every message in a block to CBOR at least three times while validating it, and
throws each encoding away. The root cause is that `Cid::from_cbor_blake2b256` materialises the whole
encoding into a `Vec<u8>` purely to feed a hasher, and neither `Message` nor `SignedMessage` caches
its CID.

No consensus risk in fixing this: the changes below alter only *how* the bytes are produced, never
what they are.

## Where the encodings happen

`src/utils/cid/mod.rs`, the root of the problem:

```rust
fn from_cbor_blake2b256<S: serde::ser::Serialize>(obj: &S) -> Result<Cid, Error> {
    let bytes = fvm_ipld_encoding::to_vec(obj)?;          // full Vec allocation
    Ok(Self::from_cbor_encoded_raw_bytes_blake2b256(&bytes))  // hash it, drop it
}
```

`Message::cid()` (`src/shim/message.rs:188`) and `SignedMessage::cid()`
(`src/message/signed_message.rs:83`) both route through it. Neither memoises.

Per **BLS** message, inside one `check_block_messages` call
(`src/chain_sync/tipset_syncer.rs`):

| # | Site | Fate of the bytes |
| - | ---- | ----------------- |
| 1 | `tipset_syncer.rs:397` — `m.cid().to_bytes()`, for the BLS aggregate check | hashed, dropped |
| 2 | `tipset_syncer.rs:424` — `to_vec(msg)?.len()`, for `price_list.on_chain_message` | length read, dropped |
| 3 | `tipset_syncer.rs:511` → `TipsetValidator::compute_msg_root` → `Cid::from_cbor_blake2b256` | hashed, dropped |

Per **SECP** message, three as well, in different shapes:

| # | Site | What is encoded |
| - | ---- | --------------- |
| 1 | `tipset_syncer.rs:493` — `check_msg(msg.message(), ..)` → `to_vec` | the inner unsigned `Message` |
| 2 | `src/shim/crypto.rs:160` — `msg.message().cid().to_bytes()` inside `authenticate_msg` | the inner `Message` again |
| 3 | `tipset_syncer.rs:511` → `compute_msg_root` | the full `SignedMessage` |

[...]

PR #7634, opened 2026-09-17
Remove redundant CBOR allocations

sudo-shashank:
## Summary of changes

- `utils::encoding::encoded_len` replaces `to_vec(x)?.len()`.
- `encoded_len` is `30-50`% faster with allocation free.
- `Cid::from_cbor_blake2b256` streams into the Blake2b hasher instead of buffering into a `Vec` and hashing that.
- `Cid::from_cbor_blake2b256` is slightly (3-6%) faster for messages with payloads, neutral for empty ones.
- Benchmark numbers:

`encoded_len(&Message)` vs `to_vec(&Message)?.len():

| `params` bytes | buffered | streamed | delta |
|---|---|---|---|
| 0 | 207.1 ns | 141.0 ns | −31.9% |
| 256 | 244.5 ns | 141.5 ns | −42.1% |
| 4096 | 295.3 ns | 141.4 ns | −52.1% |

`Cid::from_cbor_blake2b256(&Message)`

| `params` bytes | buffered | streamed | delta |
|---|---|---|---|
| 0 | 313.4 ns | 313.0 ns | −0.2% |
| 256 | 522.5 ns | 489.6 ns | −6.3% |
| 4096 | 3.041 µs | 2.952 µs | −2.9% |

Part of #7449 

## Change checklist

### Outside contributions

## Summary by CodeRabbit

- **Performance**
  - Reduced temporary memory usage when calculating encoded message sizes, validating messages, exchanging chain data, and generating content identifiers.
  - Improved CID generation by streaming serialized data instead of buffering it in memory.

- **Bug Fixes**
  - Preserved message-size limits while improving validation accuracy.
  - Ensured content identifiers consistently reflect the correct signed or unsigned message data.

- **Tests**
  - Expanded coverage for message sizing, signature variants, encoding, and CID generation.


src/utils/cid/mod.rs
```diff
@@ -14,8 +15,12 @@ pub trait CidCborExt {
...
     /// This matches [`abi.CidBuilder`](https://github.com/filecoin-project/go-state-types/blob/master/abi/cid.go#L49) in go
     fn from_cbor_blake2b256<S: serde::ser::Serialize>(obj: &S) -> Result<Cid, Error> {
-        let bytes = fvm_ipld_encoding::to_vec(obj)?;
-        Ok(Self::from_cbor_encoded_raw_bytes_blake2b256(&bytes))
+        let mut hasher = multihash_codetable::Blake2b256::default();
+        fvm_ipld_encoding::to_writer(&mut hasher, obj)?;
+        let digest = MultihashCode::Blake2b256
+            .wrap(hasher.finalize())
+            .expect("BLAKE2b-256 digest is 32 bytes, within the multihash allocation");
+        Ok(Cid::new_v1(fvm_ipld_encoding::DAG_CBOR, digest))
```
LesnyRumcajs, line 23: Is this faster than the previous implementation? Do you have a benchmark? I mean, it looks like it should be faster but I'd love some numbers.

src/utils/encoding/mod.rs
```diff
@@ -40,6 +40,27 @@ pub fn from_slice_with_fallback<'a, T: serde::de::Deserialize<'a>>(
...
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+/// Byte length of the `DAG_CBOR` encoding of `value`, without allocating it.
+pub fn encoded_len<T: serde::Serialize>(value: &T) -> Result<usize, fvm_ipld_encoding::Error> {
+    let mut sink = CountingSink(0);
+    fvm_ipld_encoding::to_writer(&mut sink, value)?;
```
LesnyRumcajs, line 60: > now constant-time

How is this constant time? It still has to go through the encoding process which depends on the object size; to my understanding you're just not allocating anything.

LesnyRumcajs (commented):
I think it's an improvement but you removed redundant CBOR allocations and not redundant CBOR encoding. Encoding cost, even without allocations, is still there. We can work on that in a follow-up, but the linked issue shouldn't be closed.

src/utils/encoding/mod.rs
sudo-shashank: Correct only avoiding allocation, fixed pr desc

src/utils/cid/mod.rs
sudo-shashank: i'll add in the pr desc

LesnyRumcajs: Ok, ping me once it's there. I don't get notifications on edits.

EclesioMeloJunior (dismissed):

src/utils/encoding/mod.rs
```diff
@@ -40,6 +40,27 @@ pub fn from_slice_with_fallback<'a, T: serde::de::Deserialize<'a>>(
...
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+/// Byte length of the `DAG_CBOR` encoding of `value`, without allocating it.
+pub fn encoded_len<T: serde::Serialize>(value: &T) -> Result<usize, fvm_ipld_encoding::Error> {
```
LesnyRumcajs: `encoded_len` kind of suggests that's it's a no-op, as in, it was already calculated somewhere and you're just extracting the value.

I suggest a more prescriptive name, e.g., `calc_encoded_len`.

src/utils/encoding/mod.rs
```diff
@@ -40,6 +40,27 @@ pub fn from_slice_with_fallback<'a, T: serde::de::Deserialize<'a>>(
...
+impl std::io::Write for CountingSink {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.0 += buf.len();
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
```
LesnyRumcajs, line 62:
```suggestion
/// `io::Write` that discards the bytes and only tracks how many were written.
#[derive(Default)]
struct CountingSink(usize);

impl std::io::Write for CountingSink {
    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
        self.0 += buf.len();
        Ok(buf.len())
    }

    fn write_vectored(&mut self, bufs: &[std::io::IoSlice<'_>]) -> std::io::Result<usize> {
        let n: usize = bufs.iter().map(|b| b.len()).sum();
        self.0 += n;
        Ok(n)
    }

    fn is_write_vectored(&self) -> bool {
        true
    }

    fn flush(&mut self) -> std::io::Result<()> {
        Ok(())
    }
}
```

If we're writing a sink, we might as well do it in full https://doc.rust-lang.org/src/std/io/util.rs.html#301-304

LesnyRumcajs: This _can_ potentially increase performance by avoiding some syscalls, if the input is scattered across different buffers

src/utils/encoding/mod.rs
sudo-shashank: Done

src/utils/encoding/mod.rs
sudo-shashank: Done

LesnyRumcajs (approved):

PR #7644, opened 2026-09-21
Remove redundant cbor encoding

### LINKED ISSUE #7449: Redundant CBOR encoding of messages during block validation

sudo-shashank:
## Summary of changes

- `check_block_messages` encodes each message once and reuses the CID and the byte length.
- BLS goes from 3 encodes to 1.
- SECP and delegated encode the inner message once.

Closes #7449 

## Change checklist

### Outside contributions

## Bug Fixes

- Improved message validation by using consistent encoded lengths and identifiers for gas checks and message-root construction.
- Strengthened Secp256k1, BLS, and delegated-signature authentication to reject messages with incorrect identifiers.
- Improved reliability of message encoding and size calculations, including streamed serialization support.

## Tests

- Added coverage for signature authentication, message identifiers, and encoded data length calculations.


src/chain_sync/tipset_syncer.rs
```diff
@@ -491,7 +494,12 @@ async fn check_block_messages(
                 "Network version must be at least NV23 for legacy Ethereum transactions".to_owned(),
             ));
         }
-        check_msg(msg.message(), &mut account_sequences, &tree).map_err(|e| {
+        let message = msg.message();
+        let msg_cid = message.cid();
+        let encoded_len = msg
+            .chain_length()
+            .expect("message serialization is infallible");
```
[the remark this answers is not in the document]
sudo-shashank: check again

src/chain_sync/tipset_syncer.rs
```diff
@@ -491,7 +495,10 @@ async fn check_block_messages(
                 "Network version must be at least NV23 for legacy Ethereum transactions".to_owned(),
             ));
         }
-        check_msg(msg.message(), &mut account_sequences, &tree).map_err(|e| {
+        let (cid, encoded_len) =
+            cid_and_encoded_len(msg).expect("message serialization is infallible");
```
LesnyRumcajs, line 499: I don't trust this. If it's infallible, then why does the function even returns a `Result`? Mind that messages are user-supplied so they can and will be malicious.

src/utils/cid/mod.rs
```diff
@@ -34,6 +35,19 @@ pub trait CidCborExt {
...
+    let mut writer = CountingWriter::new(multihash_codetable::Blake2b256::default());
+    fvm_ipld_encoding::to_writer(&mut writer, obj)?;
+    let digest = MultihashCode::Blake2b256
+        .wrap(writer.inner.finalize())
+        .expect("BLAKE2b-256 digest is 32 bytes, within the multihash allocation");
+    Ok((
+        Cid::new_v1(fvm_ipld_encoding::DAG_CBOR, digest),
+        writer.written,
+    ))
+}
```
LesnyRumcajs, line 49: this seems to be almost an exact duplicate of the one you introduced earlier. Can we do some delegation instead?

src/chain_sync/tipset_syncer.rs
```diff
@@ -387,26 +387,28 @@ async fn check_block_messages(
...
         // Do the initial loop here
         // check block message and signatures in them
         let mut pub_keys = Vec::with_capacity(block.bls_msgs().len());
-        let mut cids = Vec::with_capacity(block.bls_msgs().len());
         let db = state_manager.db();
         for m in block.bls_msgs() {
             let pk = StateManager::get_bls_public_key(db, m.from, *base_tipset.parent_state())?;
             pub_keys.push(pk);
-            cids.push(m.cid().to_bytes());
+            bls_info.push(cid_and_encoded_len(m).expect("message serialization is infallible"));
```
LesnyRumcajs, line 399: I don't feel comfortable with panicking on user-provided input.

LesnyRumcajs (changes requested):
I'm not sure this is the way to go. It introduces a bit of complexity and manual handling to avoid redundancy. IMO it should be done via memoization, reducing any mental overhead.

I noticed also a couple of more severe bugs (pre-existing) that should be handled separately.

Let's hold it off for a bit.

